PR #2827, opened 2019-06-05
Improve ContextManager class of sniffer module

gzshilu:

- [X] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
This problem is caused by the threadlocal variable。In most cases, a thread in  application server does not exist Independent. It does exist in thread-pool. The threads in the thread pool are reused, so if the value in thedaoLocal is not removed in time, it happens. After a new request, it may get an old value。
At SpringMvc plugins, `org.apache.skywalking.apm.plugin.spring.mvc.commons.interceptorStackDepth` has never been removed, so this bug was be happened.

**·How to reproduce ?**

1.  `Remove tomcat plugin (undertow, jetty-server)`
2.  Restart my application
3.  Repeat the request for the same interface multiple times with postman or jmeter.

You will see that the operation-name of current trace segment is the method name, not the request url. And EntrySpan becomes LocalSpan. I think this is unreasonable. Because I have removed all the application container plugins, spring mvc is the first interceptor , so EntrySpan is Right ,not LocalSpan。

- How to fix?
I think it's more troublesome to fix this problem in spring mvc plugins. If you remove `StackDepth` there, It may generate new problems. so I think modify ContextManager.stopSpan() is best way.Also remove `RUNTIME_CONTEXT` while removing `CONTEXT` like this:
```java

    public static void stopSpan(AbstractSpan span) {
        if (get().stopSpan(span)) {
            CONTEXT.remove();
         // add
            RUNTIME_CONTEXT.remove();  
        }
    }

```

___

This modification has been verified locally.


ascrutae:
This PR will affect the other plugin which uses RuntimeContext. The principle of clear `RunTimeContext` is who stores, who cleans up.  So I don't agree with clear it in `stopSpan` method.

and It seems like the spring plugins had a memory leak.  I think this issue should be fixed in the Spring plugin, not in the core jar.

wu-sheng:
RunningContext only closes when TracingContext close. Can't do like this.

gzshilu:
Thanks for your advice @ascrutae 。But , CONTEXT has been removed, meaning that TraceContext has been removed from ThreadLocal, so I think that there is no need to keep RUNTIME-CONTEXT. My thoughts are consistent with you. @wu-sheng

ascrutae:
Many scenarios may cause the issue. I’m not clear about your situation, The situation that the memory leak is that the trace isn't finished, and the `StackDepth` don’t remove from thread local. And the issue maybe caused by two thread had to share the StackDepth variable.


Original Message
Sender:小水怪notifications@github.com
Recipient:apache/skywalkingskywalking@noreply.github.com
Cc:Zhang" zhangxin@apache.org; Mentionmention@noreply.github.com
Date:Wednesday, Jun 5, 2019 16:27
Subject:Re: [apache/skywalking] Improve ContextManager class of sniffermodule (#2827)


Thanks for your advice @ascrutae 。But , CONTEXT has been removed, meaning that TraceContext has been removed from ThreadLocal, so I think that there is no need to keep RUNTIME-CONTEXT. My thoughts are consistent with you. @wu-sheng

wu-sheng:
I am not following, what is the original issue? Is the TracingContext, RunningContext or something else. This change should not need in any case. 

@ascrutae running context is threadlocal based, there is no thread share scenario.

ascrutae:
I had check the code and found that theStackDepth had store in RuntimeContext andthere is no thread share scenario in Spring plugin.


Original Message
Sender:吴晟 Wu Shengnotifications@github.com
Recipient:apache/skywalkingskywalking@noreply.github.com
Cc:Zhang" zhangxin@apache.org; Mentionmention@noreply.github.com
Date:Wednesday, Jun 5, 2019 16:55
Subject:Re: [apache/skywalking] Improve ContextManager class of sniffermodule (#2827)


I am not following, what is the original issue? Is the TracingContext, RunningContext or something else. This change should not need in any case.
@ascrutae running context is threadlocal based, there is no thread share scenario.

gzshilu:
>coveralls :  Many scenarios may cause the issue. I’m not clear about your situation, The situation that the memory leak is that the trace isn't finished, and the `StackDepth` don’t remove from thread local. And the issue maybe caused by two thread had to share the StackDepth variable.

1.  The code of I added in this PR is after the traceContext be finished.
2.  `StackDepth` in springmvc plugin had store in thread local, and also you later discovered it.
3.   you do not need to worry share thread in threadLocal variable.

> wu-sheng: RunningContext only closes when TracingContext close. Can't do like this.

this is original code of  `skywalking/master`:
``` java
 //ContextManager.stopSpan() : 
    public static void stopSpan(AbstractSpan span) {
        if (get().stopSpan(span)) {
            CONTEXT.remove();
        }
    }

// TraceContext().stopSpan(span) : 
    public boolean stopSpan(AbstractSpan span) {
        AbstractSpan lastSpan = peek();
        if (lastSpan == span) {
            if (lastSpan instanceof AbstractTracingSpan) {
                AbstractTracingSpan toFinishSpan = (AbstractTracingSpan)lastSpan;
                if (toFinishSpan.finish(segment)) {
                    pop();
                }
            } else {
                pop();
            }
        } else {
            throw new IllegalStateException("Stopping the unexpected span = " + span);
        }

        if (checkFinishConditions()) {
            finish();
        }

        return activeSpanStack.isEmpty();
    }

// checkFinishConditions(): 
private boolean checkFinishConditions() {
        if (isRunningInAsyncMode) {
            asyncFinishLock.lock();
        }
        try {
            if (activeSpanStack.isEmpty() && asyncSpanCounter.get() == 0) {
                return true;
            }
        } finally {
            if (isRunningInAsyncMode) {
                asyncFinishLock.unlock();
            }
        }
        return false;
    }

```
**Read this original sourceCode and you will find that CONTEXT.remove() is executed after TraceContext finished，**if  RUNTIME_CONTEXT.remove() after CONTEXT.remove(),I don't think there has any problem, and add this line is required. 
 I searched globally for used by RUNTIME_CONTEXT in project, it never be  removed**.It has a risk of out of memory.** Theadlocal is best to  execute remove() after use completed in current Thread. I believe that if there is no RUNTIME_CONTEXT be added in this place, there will be other business bug , but we have not found it now.

by my changes, I tested a day on my team’s project yesterday，did not find any questions for this change, and it solved the spring plugin's bug of this PR point(we don't use tomcat、undertow or jetty-server plugins,because there is an imperfection about display of operation-name of  restful-api 。).

I didn't want to reply because this PR was closed. I reply it  because I think the technology should be stricter.

Thanks for listening. @wu-sheng @ascrutae

wu-sheng:
Thanks for reply. I will dig a little more about, reopening again.

In existing design, running context should be managed by plugin, but if it is for safer, I am om with that.

wu-sheng:
@gzshilu @ascrutae I am going to accept this pull request, but please remember, this is just a fail-safe. The plugin still should fix the leak issue.

> At SpringMvc plugins, org.apache.skywalking.apm.plugin.spring.mvc.commons.interceptorStackDepth has never been removed, so this bug was be happened.

@ascrutae This bug should be fixed.

ascrutae:
@gzshilu Could you post me the demo to reproduce the issue?

gzshilu:
Thanks for follow-up .   I don't know how to provide demo .   but it is easy to reproduce，first of all , comment `RUNTIME_CONTEXT.remove()`  in ContextManager on master branch. 
Then proceed as follows ：

>  1. Remove tomcat plugin (undertow, jetty-server)
>  2. Restart my application
>  3. Repeat the request for the same interface multiple times with postman or jmeter.

ascrutae:
Restart my application
That application is I want to , Could you send it to me ?


Original Message
Sender:小水怪notifications@github.com
Recipient:apache/skywalkingskywalking@noreply.github.com
Cc:Zhang" zhangxin@apache.org; Mentionmention@noreply.github.com
Date:Monday, Jun 10, 2019 10:36
Subject:Re: [apache/skywalking] Improve ContextManager class of sniffermodule (#2827)


Thanks for follow-up . I don't know how to provide demo . but it is easy to reproduce，first of all , comment RUNTIME_CONTEXT.remove() in ContextManager on master branch.
 Then proceed as follows ：
Remove tomcat plugin (undertow, jetty-server) Restart my application Repeat the request for the same interface multiple times with postman or jmeter.

gzshilu:
Any application can, just need it to use the interface defined by the spring mvc annotation.I tested it in our internal project,  so I am sorry. and skywalking is non-intrusive, I think that it's no related to the application, but related to the plugins itself.

ascrutae:
I just want to make sense that this issue if it caused by specify application, As your said any application can cause this issue, I will try to fix it.


Original Message
Sender:小水怪notifications@github.com
Recipient:apache/skywalkingskywalking@noreply.github.com
Cc:Zhang" zhangxin@apache.org; Mentionmention@noreply.github.com
Date:Monday, Jun 10, 2019 10:49
Subject:Re: [apache/skywalking] Improve ContextManager class of sniffermodule (#2827)


Any application can, just need it to use the interface defined by the spring mvc annotation.I tested it in our internal project, so I am sorry. and skywalking is non-intrusive, I think that it's no related to the application, but related to the plugins itself.

gzshilu:
Yes, I can understanding you so much.Optative, I can write another demo application for you, but I don't have time now, but please believe me, really no related the application, I have tested multiple microservices, has the same problem, EntrySpan becomes LocalSpan. 
Also, Thanks for providing such a nice project for the open-source community, which will bring great help to our project.

PR #2828, opened 2019-06-05
Optimize SegmentParseV2::parse code logic

zhaoyuguang:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance
___
Passing my test `parseBinarySegment(upstreamSegment)` is a time-consuming and stable method that does not reduce time-consuming because of multiple calls.
image


wu-sheng:
Why the parse method called multiple time? I think this is not true.

zhaoyuguang:
With debug, `bufferData.getV2Segment() == null` is true so parse method called multiple time.

wu-sheng:
Get your point. I am thinking more, if this is the case, why don't we just get v2 segment after if null block?

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/parser/SegmentParseV2.java
```diff
@@ -82,10 +82,11 @@ public boolean parse(BufferData<UpstreamSegment> bufferData, SegmentSource sourc
 
             List<UniqueId> traceIds = upstreamSegment.getGlobalTraceIdsList();
 
+            SegmentObject segmentObject = parseBinarySegment(upstreamSegment);
```
wu-sheng: `SegmentParseV2#parse` will be invoked multiple times, due to `SegmentParseV2#Producer#call` triggered by `DataStreamReader#L159-L161`. So, the `#parseBinarySegment` should be called only when `bufferData.getV2Segment() == null` from my perspective

zhaoyuguang: After the debug, as you said, read from the file and go back to this judgment.

PR #2835, opened 2019-06-06
Refactor ConfigWatcherRegister to only read interested configs

### LINKED ISSUE #2834: Refactor method `ConfigWatcherRegister#readConfig`, add config keys as parameter to only read interested configs
I think you need to change this interface method, let's add config keys as parameter? Then we could avoid config keys in providet. 

Could you send a new pull request about this first?

_Originally posted by @wu-sheng in https://github.com/apache/skywalking/pull/2817_

kezhenxu94:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues
resolve #2834

___
Specify interested config keys and only read configs of those keys from configuration center


wu-sheng:
Thanks for adjusting this fast.

PR #2840, opened 2019-06-08
Support exporter runs in increment and total modes

wu-sheng:
In prev, the exporter always exports total data of metrics, now, it supports all metrics, which give the exporter provide much more freedom. 

Also, two modes provided.
1. Increment(New), metrics data of this bulk.
1. Total(Old), metrics data of current status, including prev persistent data.

### DIFF
--- a/oap-server/exporter/src/main/java/org/apache/skywalking/oap/server/exporter/provider/grpc/GRPCExporter.java
+++ b/oap-server/exporter/src/main/java/org/apache/skywalking/oap/server/exporter/provider/grpc/GRPCExporter.java
@@ -26,7 +26,7 @@ import lombok.*;
 import org.apache.skywalking.apm.commons.datacarrier.DataCarrier;
 import org.apache.skywalking.apm.commons.datacarrier.consumer.IConsumer;
 import org.apache.skywalking.oap.server.core.analysis.metrics.*;
-import org.apache.skywalking.oap.server.core.exporter.MetricValuesExportService;
+import org.apache.skywalking.oap.server.core.exporter.*;
 import org.apache.skywalking.oap.server.exporter.grpc.*;
 import org.apache.skywalking.oap.server.exporter.provider.MetricFormatter;
 import org.apache.skywalking.oap.server.library.client.grpc.GRPCClient;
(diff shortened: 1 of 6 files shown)


oap-server/exporter/src/main/java/org/apache/skywalking/oap/server/exporter/provider/grpc/GRPCExporter.java
```diff
@@ -56,9 +56,15 @@ public GRPCExporter(GRPCExporterSetting setting) {
         subscriptionSet = new HashSet<>();
     }
 
-    @Override public void export(MetricsMetaInfo meta, Metrics metrics) {
-        if (subscriptionSet.size() == 0 || subscriptionSet.contains(meta.getMetricsName())) {
-            exportBuffer.produce(new ExportData(meta, metrics));
+    @Override public void export(ExportEvent event) {
+        if (ExportEvent.EventType.TOTAL.equals(event.getType())) {
```
kezhenxu94: `EventType` is enum type so we can just replace `.equals` with `==`? As you said in another comment (I don't exactly remember which one), it reduce a unnecessary call stack frame

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/exporter/MetricValuesExportService.java
```diff
@@ -27,5 +26,10 @@
  * @author wusheng
  */
 public interface MetricValuesExportService extends Service {
-    void export(MetricsMetaInfo meta, Metrics metrics);
+    /**
+     * This method is sync-mode export, the performance effects the persistence result. Queue mode is high recommended.
```
kezhenxu94: high recommended => highly recommended?

oap-server/exporter/src/main/java/org/apache/skywalking/oap/server/exporter/provider/grpc/GRPCExporter.java
wu-sheng: I said this for `equal` PR, not merged. For enum, `==` is OK :)

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/exporter/MetricValuesExportService.java
wu-sheng: +1

PR #2846, opened 2019-06-10
Add config to control tracing sql parameters in MySQL agent

### LINKED ISSUE #2587: Support database sql statement parameter value(optional)
component: mysql-connector-java
db.type: sql
db.statement is correctly show but without parameter value.

kezhenxu94:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
resolves #2587 

___

Add a config key to control whether to collect the SQL parameters.

Here is one point need to discuss:

there're methods like `java.sql.PreparedStatement#setBinaryStream(int, java.io.InputStream, long)`, `java.sql.PreparedStatement#setBlob(int, java.io.InputStream, long)` and similar methods in `PreparedStatement`, should we collect them as well? The `InputStream` may be large and hard to display in tags, I would like to only collect their length and display something like `InputStream[length]` or anyone has better idea?

cc @vision-ken

here is the span detail:
image


wu-sheng:
>there're methods like java.sql.PreparedStatement#setBinaryStream(int, java.io.InputStream, long), java.sql.PreparedStatement#setBlob(int, java.io.InputStream, long) and similar methods in PreparedStatement, should we collect them as well? The InputStream may be large and hard to display in tags, I would like to only collect their length and display something like InputStream[length] or anyone has better idea?

You could simply ignore them. I don't think we need to do this with such high payload. For `InputStream[length]`, I am not sure whether it is worth to do.
Also, based on this, your parameter tag value is not perfect. Thinking about these
1. If one parameter is ignored somehow, parameter list would match, unless you keep a `?` or `InputStream` or something to be a placeholder.
1. If the parameter includes `,`, what happens?

wu-sheng:
Also, active this in MySQL integration test case, and modify the test case, please.

kezhenxu94:
> For `InputStream[length]`, I am not sure whether it is worth to do.

The reason why I'd like to do this is to make it a placeholder, as you said:

> If one parameter is ignored somehow, parameter list would match, unless you keep a ? or InputStream or something to be a placeholder.

keeping a `?` as placeholder is fine for me.

> If the parameter includes `,`, what happens?

quoting the parameter with `"` seems to be not perfect either, I'll think about this

ascrutae:
Should we limit the length of string parameters?  The performance of Agent will be reduced if the values of SQL parameter are too large

kezhenxu94:
> Should we limit the length of string parameters? The performance of Agent will be reduced if the values of SQL parameter are too large

You mean truncating the final parameter string? 

If users active this config but not get the complete parameters, it would be useless sometimes (the useful parameters may be truncated).

I'd rather suggest the users not to active this config when the parameters are large/long, or just active during debugging.

It's disabled by default, so the users must pay for the performance themselves if we clearly documented this.

Just my own opinion, further discussion are much welcomed. @ascrutae

wu-sheng:
> Should we limit the length of string parameters? The performance of Agent will be reduced if the values of SQL parameter are too large

This could be argued too :) But for performance concern, I think @kezhenxu94 you could set a max parameter length config too.

wu-sheng:
> quoting the parameter with " seems to be not perfect either, I'll think about this

Normally, I would say JSON Array should a good idea, but in agent, I think we don't have JSON lib today.

kezhenxu94:
> > quoting the parameter with " seems to be not perfect either, I'll think about this
> 
> Normally, I would say JSON Array should a good idea, but in agent, I think we don't have JSON lib today.

There seems to be GSON already

wu-sheng:
> There seems to be GSON already

If so, using it is a better idea. What do you think?

kezhenxu94:
> > There seems to be GSON already
> 
> If so, using it is a better idea. What do you think?

I'm with it

kezhenxu94:
> > Should we limit the length of string parameters? The performance of Agent will be reduced if the values of SQL parameter are too large
> 
> This could be argued too :) But for performance concern, I think @kezhenxu94 you could set a max parameter length config too.

Hi @ascrutae , as suggested, I've truncate the sql parameters if it's long. now it works like this:

image

As @wu-sheng suggested, I've made another config `SQL_PARAMETERS_MAX_LENGTH` to make it possible for users who are perfectly sure they want to keep the complete parameters, thanks @wu-sheng :-)

wu-sheng:
OK. I am OK with this PR at high level. @IanCao @ascrutae Please do the code receivews.

@kezhenxu94 Let's make the test cases work.

apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/define/StatementEnhanceInfos.java
```diff
@@ -31,11 +34,13 @@
     private ConnectionInfo connectionInfo;
     private String statementName;
     private String sql;
+    private Map<Integer, Object> parametersKeyedByIndex;
 
     public StatementEnhanceInfos(ConnectionInfo connectionInfo, String sql, String statementName) {
         this.connectionInfo = connectionInfo;
         this.sql = sql;
         this.statementName = statementName;
+        this.parametersKeyedByIndex = new TreeMap<Integer, Object>();
```
wu-sheng: 1. Why treeMap?
2. Don't create the map if no parameter tracing. Map creation is a high payload. You could test 😄

apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementExecuteMethodsInterceptor.java
```diff
@@ -29,29 +32,59 @@
...
 
     @Override
     public final void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments,
         Class<?>[] argumentsTypes,
         MethodInterceptResult result) throws Throwable {
         StatementEnhanceInfos cacheObject = (StatementEnhanceInfos)objInst.getSkyWalkingDynamicField();
         ConnectionInfo connectInfo = cacheObject.getConnectionInfo();
+
+        final String methodName = method.getName();
+        final boolean traceSqlParameters = Config.Plugin.MySQL.TRACE_SQL_PARAMETERS;
```
wu-sheng: Why need this line?

apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/define/StatementEnhanceInfos.java
wu-sheng: Treemap to keep order? User may not using the 1-n order. You need to bw careful.

kezhenxu94: 1. We need to keep the parameters' order, because user may invoke `ps.setString(2, "val2");` before `ps.setString(1, "val1")`, and we need to print `val2, val1`.
2. Got it, will improve this later

apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementExecuteMethodsInterceptor.java
kezhenxu94: Oh we don't this when `methodName` is under consideration

apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/define/StatementEnhanceInfos.java
wu-sheng: > We need to keep the parameters' order, because user may invoke ps.setString(2, "val2"); before ps.setString(1, "val1"), and we need to print val2, val1.

No, we don't. Think in this way, dev or ops team doesn't know the codes, and saw `select xx from yyy where a=? and b=?` and you put `val2(i = 1), val1(i = 2)`, how they think?

wu-sheng: I think people thought `a = val2 and b = val1`, right? But the reality is `a = val1 and b = val2` right?
Tracing is not really all about codes, it is for observability, a.k.a. understandable for the system.

kezhenxu94: > > We need to keep the parameters' order, because user may invoke ps.setString(2, "val2"); before ps.setString(1, "val1"), and we need to print val2, val1.
> 
> No, we don't. Think in this way, dev or ops team doesn't know the codes, and saw `select xx from yyy where a=? and b=?` and you put `val2(i = 1), val1(i = 2)`, how they think?

by saying "keep the parameters' order", I didn't mean to keep the **order of the code**, I meant keep the **index order of the code**, and since the `index` is the key of the `TreeSet`, therefore the values are sorted by the keys (index here),

wu-sheng: Then why don't use an Array list? Look like much more efficiency, right?
 ```
list = new ArrayList;
list,checkRange(i)
list.set(i, value)
```

Comment (kezhenxu94) [reply apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/define/StatementEnhanceInfos.java]: we don't know the parameters' size in advance, `checkRange` manually may need to enlarge the capacity, and copying elements, I'm not sure which one is better, did I miss anything about `ArrayList`?

Comment (wu-sheng) [reply apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/define/StatementEnhanceInfos.java]: First of all, even in today's agent codes, we have an unsolved issue. In many practices, introduced by Google, ArrayList has better performance than LinkedList, even with enlarge the capacity and copy elements. We can build a benchmark. Clearly, LinkedList should similar or better than TreeMap in complexity, right? Also, the benchmark is welcome.

Besides above, let's think in this way. ArrayList with 20 as initial size, then how many of parameters a SQL will have? 40 - 60 is a big one, right? Especially in micro-service and less Oracle today.
The ArrayList will enlarge twice mostly, others just set, I think that should be much quicker.

Let's do multiple tests to confirm this.

Comment (kezhenxu94) [reply apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/define/StatementEnhanceInfos.java]: make sense, I'll try to do some benchmark tests to make this clear and for future reference

**Comment by kezhenxu94** [comment]
> > There seems to be GSON already
> 
> If so, using it is a better idea. What do you think?

Turns out that we can use the Gson library because of the same reason as #2871

File: .gitmodules [hunk 3]
```diff
@@ -1,9 +1,12 @@
...
 [submodule "oap-server/server-query-plugin/query-graphql-plugin/src/main/resources/query-protocol"]
 	path = oap-server/server-query-plugin/query-graphql-plugin/src/main/resources/query-protocol
-	url = https://github.com/apache/skywalking-query-protocol.git
+	url = file:///data/developer/git/java/skywalking/oap-server/server-query-plugin/query-graphql-plugin/src/main/resources/query-protocol
+	#url = https://github.com/apache/skywalking-query-protocol.git
 [submodule "skywalking-ui"]
 	path = skywalking-ui
-	url = https://github.com/apache/skywalking-rocketbot-ui.git
+	#url = https://github.com/apache/skywalking-rocketbot-ui.git
+	url = file:///data/developer/git/java/skywalking/skywalking-ui
```
wu-sheng: This is unexpected changes

kezhenxu94: > This is unexpected changes

Yes, for local testing, will revert and recheck, once done, I'll ping you, thanks :)

apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/define/StatementEnhanceInfos.java
kezhenxu94: Benchmark and result are here: https://github.com/apache/skywalking/pull/2846/files#diff-c79b0eaced528877516f9cbf5a4ddb7c

kezhenxu94:
> > > There seems to be GSON already
> > 
> > 
> > If so, using it is a better idea. What do you think?
> 
> Turns out that we cannot use the Gson library because of the same reason as #2871

@wu-sheng I have to give up the decision (using Gson) what we discussed above, now I just concat the parameters using `,` and prefixed with `[` suffixed with `]`, perhaps we could find a better way, any suggestion, thanks?

kezhenxu94:
I tested by setting the `plugin.mysql.trace_sql_parameters=true` in the config file and run locally:

[INFO] ---------------------------Here is the test report context -------------------------------------

# Test Report - 2019-06-15 19:03

- tests  : **1 passed**. **0 failed**
- branch name : **[gh/2587](https://github.com/apache/incubator-skywalking/tree/gh/2587)**
- commit id : **[346a61f2656651ed40271797528b7c375387eb80](https://github.com/apache/incubator-skywalking/commit/346a61f2656651ed40271797528b7
5387eb80)**
- cases branch: **[master](https://github.com/SkywalkingTest/skywalking-autotest-scenarios/tree/master)**
- cases commit id: **[6844ab3003a30f21cb7a634509d192b756b2393f](https://github.com/SkywalkingTest/skywalking-autotest-scenarios/commit/6844ab3
a30f21cb7a634509d192b756b2393f)**

## Cases List

| Case description | Status | Cases|
|:-----|:-----:|:-----:|
|Mysql| **1 passed. 0 failed**| [click me](#mysql) |

## Mysql

### 
|  Version     | Passed | Failed|
|:------------- |:-------:|:-----:|
| 8.0.15  | :heavy_check_mark:||


[INFO] -----------------------------------end the report-----------------------------------------------

apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementExecuteMethodsInterceptor.java
```diff
@@ -30,28 +31,55 @@
...
+        }
+
         /**
          * For avoid NPE. In this particular case, Execute sql inside the {@link com.mysql.jdbc.ConnectionImpl} constructor,
          * before the interceptor sets the connectionInfo.
          *
          * @see JDBCDriverInterceptor#afterMethod(EnhancedInstance, Method, Object[], Class[], Object)
          */
-        if (connectInfo != null) {
+        if (connectInfo != null && !PS_SETTERS.contains(methodName) && !PS_IGNORED_SETTERS.contains(methodName)) {
```
wu-sheng: This should be in `else if` and remove `!PS_SETTERS.contains(methodName) && !PS_IGNORED_SETTERS.contains(methodName)`, right?

apm-sniffer/apm-test-tools/src/test/java/org/apache/skywalking/apm/plugin/ArbitrarySetTest.java
```diff
@@ -0,0 +1,238 @@
...
+                newList.addAll(list);
+                newList.addAll(oldSize, Collections.nCopies(newSize - oldSize, PLACEHOLDER));
+                list = newList;
+            }
+            list.set(i, i);
+        }
+    }
+
+    @Benchmark
+    public void array() {
```
wu-sheng: You should use a LinkedList as compare testing.

apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementExecuteMethodsInterceptor.java
```diff
@@ -78,4 +108,39 @@ public final Object afterMethod(EnhancedInstance objInst, Method method, Object[
...
+            if (DISPLAYABLE_TYPES.contains(parameterType)) {
+                return parameter;
+            }
+        } else if (PS_SETTERS.contains(methodName)) {
+            return allArguments[1];
+        }
+        return SQL_PARAMETER_PLACEHOLDER;
+    }
+
+    private String buildParameterString(Object[] parameters) {
```
wu-sheng: Why don't use GSON Array `#toString`?

kezhenxu94: We cannot use Gson because the target application don't necessary include it

wu-sheng: Yes, ignore this.

apm-sniffer/apm-test-tools/src/test/java/org/apache/skywalking/apm/plugin/ArbitrarySetTest.java
kezhenxu94: Added

kezhenxu94:
> ```
> /**
>  * Benchmark                     Mode  Cnt     Score     Error   Units
>  * ArbitrarySetTest.array       thrpt   10  2360.500 ± 138.279  ops/ms
>  * ArbitrarySetTest.arrayList   thrpt   10  1080.005 ± 225.897  ops/ms
>  * ArbitrarySetTest.linkedList  thrpt   10   188.007 ±  11.739  ops/ms
>  * ArbitrarySetTest.treeMap     thrpt   10   214.384 ±  27.816  ops/ms
>  *
>  */
> ```
> 
> 
> Based on this, let's use `array` to do this? Agree?

Surely I agree, I have already used `array` to rewrite after the benchmark shows it's performant

apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementExecuteMethodsInterceptor.java
```diff
@@ -30,28 +31,55 @@
...
     @Override
     public final void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments,
         Class<?>[] argumentsTypes,
         MethodInterceptResult result) throws Throwable {
         StatementEnhanceInfos cacheObject = (StatementEnhanceInfos)objInst.getSkyWalkingDynamicField();
         ConnectionInfo connectInfo = cacheObject.getConnectionInfo();
+
+        final String methodName = method.getName();
+
+        if (PS_SETTERS.contains(methodName) || PS_IGNORED_SETTERS.contains(methodName)) {
```
wu-sheng: You should use tracing parameter config to judge first, for performance when it is off.

apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementExecuteMethodsInterceptor.java
wu-sheng: You should use tracing parameter config to judge first, for performance when it is off.

apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementExecuteMethodsInterceptor.java
```diff
@@ -61,7 +89,9 @@ public final Object afterMethod(EnhancedInstance objInst, Method method, Object[
         Class<?>[] argumentsTypes,
         Object ret) throws Throwable {
         StatementEnhanceInfos cacheObject = (StatementEnhanceInfos)objInst.getSkyWalkingDynamicField();
-        if (cacheObject.getConnectionInfo() != null) {
+        ConnectionInfo connectionInfo = cacheObject.getConnectionInfo();
+        String methodName = method.getName();
+        if (connectionInfo != null && !PS_SETTERS.contains(methodName) && !PS_IGNORED_SETTERS.contains(methodName)) {
```
wu-sheng: You should use tracing parameter config to judge first, for performance when it is off.

apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/Constants.java
```diff
@@ -29,4 +40,62 @@
     public static final String SET_CATALOG_INTERCEPTOR = "org.apache.skywalking.apm.plugin.jdbc.mysql.SetCatalogInterceptor";
     public static final String STATEMENT_EXECUTE_METHODS_INTERCEPTOR = "org.apache.skywalking.apm.plugin.jdbc.mysql.StatementExecuteMethodsInterceptor";
     public static final String DRIVER_CONNECT_INTERCEPTOR = "org.apache.skywalking.apm.plugin.jdbc.mysql.DriverConnectInterceptor";
+
+    public static final StringTag SQL_PARAMETERS = new StringTag("db.sql.parameters");
+    public static final String SQL_PARAMETER_PLACEHOLDER = "?";
+    public static final Set<String> PS_SETTERS = new HashSet<String>(Arrays.asList(
```
wu-sheng: I still have performance concerns for these naming match. You should set a explicit instrumentation for get parameter, rather than using string match.
Instrumentation name match filter happens in startup amd class loading stage.

apm-sniffer/apm-sdk-plugin/mysql-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/v6/define/PreparedStatementInstrumentation.java
```diff
@@ -53,10 +56,19 @@
...
                         .or(named("executeLargeUpdate"));
+                    if (Config.Plugin.MySQL.TRACE_SQL_PARAMETERS) {
+                        for (String setter : PS_SETTERS) {
+                            matcher = matcher.or(named(setter));
+                        }
+                        for (String setter : PS_IGNORED_SETTERS) {
+                            matcher = matcher.or(named(setter));
+                        }
+                    }
+                    return matcher;
```
wu-sheng: Same to
>I still have performance concerns for these naming match. You should set a explicit instrumentation for get parameter, rather than using string match.
Instrumentation name match filter happens in startup amd class loading stage.

kezhenxu94:
Hi @wu-sheng here I've rewrote the instrumentations for the sake of performance, at least it behaviors as if it doesn't exist when `trace_sql_parameters=false`

Thanks for kindly reminding me the concerns of performance and the suggestions.

apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementSetterMethodsInterceptor.java
```diff
@@ -0,0 +1,80 @@
...
+    private Object getDisplayedParameter(final Method method, final Class<?>[] argumentTypes, final Object[] allArguments) {
+        final String methodName = method.getName();
+        if ("setNull".equals(methodName)) {
+            return "null";
+        }
+        if ("setObject".equals(methodName)) {
+            final Object parameter = allArguments[0];
+            final Class<?> parameterType = argumentTypes[1];
+
+            if (DISPLAYABLE_TYPES.contains(parameterType)) {
```
wu-sheng: For higher performance, you should totally avoid contains, especially contain a long list and high frequency op. Separate it to different interceptors.

And notice, these inteceptor should be placed in jdbc common, you would notice other jdbc driver plugin could reuse all these, right? Only instrumentation definitions are mysql related.

apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/JDBCPreparedStatementSetterInstanceMethodsInterceptPoint.java
```diff
@@ -0,0 +1,69 @@
...
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.none;
+import static org.apache.skywalking.apm.plugin.jdbc.define.Constants.PS_IGNORABLE_SETTERS;
+import static org.apache.skywalking.apm.plugin.jdbc.define.Constants.PS_SETTERS;
+
+/**
+ * @author kezhenxu94
+ */
+public class JDBCPreparedStatementSetterInstanceMethodsInterceptPoint implements InstanceMethodsInterceptPoint {
```
wu-sheng: Could you rename these, such as  `PreparedStatementSetterInstanceMethodsInterceptPointOfJDBCInstrumentation`. Because only `Instrumentation` suffix could be checked by CI scripts of instrumentation definition.

kezhenxu94: Yes sure

wu-sheng: Or short version `PSSetterDefinitionOfJDBCInstrumentation`.

apm-sniffer/apm-sdk-plugin/mysql-8.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/v8/define/PreparedStatementNullSetterInstrumentation.java
```diff
@@ -24,14 +24,15 @@
 import org.apache.skywalking.apm.agent.core.plugin.match.ClassMatch;
 import org.apache.skywalking.apm.plugin.jdbc.JDBCPreparedStatementNullSetterInstanceMethodsInterceptPoint;
 
-import static org.apache.skywalking.apm.agent.core.plugin.match.NameMatch.byName;
+import static org.apache.skywalking.apm.agent.core.plugin.match.MultiClassNameMatch.byMultiClassMatch;
 
 /**
  * @author kezhenxu94
  */
 public class PreparedStatementNullSetterInstrumentation extends AbstractMysqlInstrumentation {
```
wu-sheng: Why not extend PreparedStatementInstrumentation? Just override `getInstanceMethodsInterceptPoints`. Should be much easier and safer? Agree.

kezhenxu94: make sense, save me a lot of codes, thanks

kezhenxu94:
> LGTM. Could you run the test case locally or ask @IanCao to run the cases on our server.
> 
> I think it would be safer to recheck the test cases. @kezhenxu94 If you want to run server, you need to tell @IanCao which cases are needed to run. All MySQL cases, I guess?

@IanCao Could you please run the tests with the original test code and agent code of this PR? I'll run the modified version locally and post the result here

wu-sheng:
> Could you please run the tests with the original test code and agent code of this PR? I'll run the modified version locally and post the result here

Our tests will use these PR codes, not original codes. @kezhenxu94

kezhenxu94:
> > Could you please run the tests with the original test code and agent code of this PR? I'll run the modified version locally and post the result here
> 
> Our tests will use these PR codes, not original codes. @kezhenxu94

What I meant is to ask @IanCao to use the original **test codes** to test these PR codes (with the config key off, which is by default), and in order to test the codes with the config key on, I have to modify some **test codes**, and I'll use the modified test codes to test these PR code

wu-sheng:
@kezhenxu94 Look like most passed. One fails, I assume it is just accidental by current test tool, cc @ascrutae .

wu-sheng:
@kezhenxu94 I am waiting for your test results of `tracing parameters on` to make these merged.

kezhenxu94:
> @kezhenxu94 I am waiting for your test results of `tracing parameters on` to make these merged.

Tests are still running locally, poor local machine 😢 once done, I'll post here

wu-sheng:
> Tests are still running locally, poor local machine 😢 once done, I'll post here

Got it. @ascrutae Let's make this work in next month, then we could have a much more friendly PR procedure.

kezhenxu94:
# Test Report - 2019-06-19 11:52

- tests  : **53 passed**. **0 failed**
- branch name : **[gh/2587](https://github.com/apache/incubator-skywalking/tree/gh/2587)**
- commit id : **[148f12125a00cf198e0f75881e4a648d7e55c9dc](https://github.com/apache/incubator-skywalking/commit/148f12125a00cf198e0f75881e4a648d7e55c9dc)**
- cases branch: **[master](https://github.com/SkywalkingTest/skywalking-autotest-scenarios/tree/master)**
- cases commit id: **[277e06f4b49a9e8e9bf61fe61bc23910b6975d6c](https://github.com/SkywalkingTest/skywalking-autotest-scenarios/commit/277e06f4b49a9e8e9bf61fe61bc23910b6975d6c)**

## Cases List

| Case description | Status | Cases|
|:-----|:-----:|:-----:|
|Mysql| **53 passed. 0 failed**| [click me](#mysql) |

## Mysql

### 
|  Version     | Passed | Failed|
|:------------- |:-------:|:-----:|
| 5.1.10  | :heavy_check_mark:||
| 5.1.11  | :heavy_check_mark:||
| 5.1.12  | :heavy_check_mark:||
| 5.1.13  | :heavy_check_mark:||
| 5.1.14  | :heavy_check_mark:||
| 5.1.15  | :heavy_check_mark:||
| 5.1.16  | :heavy_check_mark:||
| 5.1.17  | :heavy_check_mark:||
| 5.1.18  | :heavy_check_mark:||
| 5.1.19  | :heavy_check_mark:||
| 5.1.2  | :heavy_check_mark:||
| 5.1.20  | :heavy_check_mark:||
| 5.1.21  | :heavy_check_mark:||
| 5.1.22  | :heavy_check_mark:||
| 5.1.23  | :heavy_check_mark:||
| 5.1.24  | :heavy_check_mark:||
| 5.1.25  | :heavy_check_mark:||
| 5.1.26  | :heavy_check_mark:||
| 5.1.27  | :heavy_check_mark:||
| 5.1.28  | :heavy_check_mark:||
| 5.1.29  | :heavy_check_mark:||
| 5.1.3  | :heavy_check_mark:||
| 5.1.30  | :heavy_check_mark:||
| 5.1.31  | :heavy_check_mark:||
| 5.1.32  | :heavy_check_mark:||
| 5.1.33  | :heavy_check_mark:||
| 5.1.34  | :heavy_check_mark:||
| 5.1.35  | :heavy_check_mark:||
| 5.1.36  | :heavy_check_mark:||
| 5.1.37  | :heavy_check_mark:||
| 5.1.38  | :heavy_check_mark:||
| 5.1.39  | :heavy_check_mark:||
| 5.1.4  | :heavy_check_mark:||
| 5.1.40  | :heavy_check_mark:||
| 5.1.41  | :heavy_check_mark:||
| 5.1.42  | :heavy_check_mark:||
| 5.1.43  | :heavy_check_mark:||
| 5.1.44  | :heavy_check_mark:||
| 5.1.45  | :heavy_check_mark:||
| 5.1.5  | :heavy_check_mark:||
| 5.1.6  | :heavy_check_mark:||
| 5.1.8  | :heavy_check_mark:||
| 5.1.9  | :heavy_check_mark:||
| 6.0.2  | :heavy_check_mark:||
| 6.0.3  | :heavy_check_mark:||
| 6.0.4  | :heavy_check_mark:||
| 6.0.5  | :heavy_check_mark:||
| 6.0.6  | :heavy_check_mark:||
| 8.0.11  | :heavy_check_mark:||
| 8.0.12  | :heavy_check_mark:||
| 8.0.13  | :heavy_check_mark:||
| 8.0.14  | :heavy_check_mark:||
| 8.0.15  | :heavy_check_mark:||

And FYI, my test codes are [here](https://github.com/kezhenxu94/agent-auto-integration-testcases/commit/ae3ff07aa02e9a08431fb180ffd19d7635640c5a)

Issue #2501, opened 2019-04-21
[Agent] Provide plugin for Solr-7.x


dmsolr:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [x] Requirement
- [ ] Feature or performance improvement

___
- it will provide a sniffer plugin include solrj and solr.

wu-sheng:
Welcome the contribution for the new plugin.

dmsolr:
Coming soon. :-)

wu-sheng:
Please notice this, https://github.com/SkyAPMTest/agent-auto-integration-testcases. Automatic tests are highly recommended to provide with your plugin pull request before we do code review. That test is the strong proof to help the project team accepts your new plugin.

dmsolr:
I get it. thanks for your tips.

dmsolr:
how difficult about auto-testcases.  : (

wu-sheng:
Which part is different for you? Do you face any issue? Cc @ascrutae

dmsolr:
I don't have any experience on docker. but it is ok, I think.

wu-sheng:
Yes. Docker is required. But basically, it reduced the difficulties because of docker.

PR #2853, opened 2019-06-11
Provide Solr-7.x-plugin

dmsolr:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#2501


dmsolr:
The situation on the server side is more complicated than SolrJ. 
I need some advice `how to name operationName` and other. 

It will put the trace id into Solr MDC. Then Solr logs it in SLOW logger or where you want to by Log4j configuration file. 

image
image

dmsolr:
@wu-sheng yes, it works in Solr Server, likes ElasticSearch server-side.

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -53,7 +53,8 @@
     * [Xmemcached](https://github.com/killme2008/xmemcached) 2.x
   * [Elasticsearch](https://github.com/elastic/elasticsearch)
     * [transport-client](https://github.com/elastic/elasticsearch/tree/master/client/transport) 5.2.x-5.6.x
-  * [SolrJ](https://lucene.apache.org/solr) 7.0.0-7.7.1
+  * [Solr](https://github.com/apache/lucene-solr) 7.0.0-7.7.1
+    * [SolrJ](https://github.com/apache/lucene-solr/tree/master/solr/solrj) 7.0.0-7.7.1
```
wu-sheng: Add Solr Server and versions as a subtype here.

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -53,7 +53,8 @@
     * [Xmemcached](https://github.com/killme2008/xmemcached) 2.x
   * [Elasticsearch](https://github.com/elastic/elasticsearch)
     * [transport-client](https://github.com/elastic/elasticsearch/tree/master/client/transport) 5.2.x-5.6.x
-  * [SolrJ](https://lucene.apache.org/solr) 7.0.0-7.7.1
+  * [Solr](https://github.com/apache/lucene-solr) 7.0.0-7.7.1
```
wu-sheng: Remove the version.

wu-sheng:
Could you explain more about how many spans created for each request? And what are they?

dmsolr:
Here is a screenshot of solr collection. Obviously, it consists of 2 shards with 2 replicas per shard. 

image

In query, we omit  the replicas. One shard become a coordinator, when it receives a query from user. Because it is has a part of data in collection, It launch a request and send to others shards. Then coordinator have to merge the result that comes from all shards. In real scenario, it has serveral stages in usually.

image
It seems like above picture.

In update, the documents need to be routed and send to corresponding shard. Then shard will copy it to they replicas.

dmsolr:
Solr is a distributed database.

dmsolr:
@kezhenxu94， Do you have anything else for the explaination？

wu-sheng:
OK. I think it is clear enough for the Solr server. Then let's talk about how this goes in topology and trace.
1. Does topology show Solr server cluster as a single one nodes? Like other distributed system.
1. Do you plan to show RPC between Solr server nodes? 

I think the decision is based on the answers of the above questions.

dmsolr:
@wu-sheng , Sorry, I have a mistake that I configured the `agent.service_name` in smae name.
In fact, they work in cluster. 
By the way, Solr still uses http. So I define the SpanLayer as `http`.

image

image

image

wu-sheng:
@TinyAllen I noticed in this case, there are instances of one service calling each others, the topology shows a little strange, please take a look.

wu-sheng:
@dmsolr If the client SolrJ is instrumented, do the topology and metrics still make sense? I notice you are identifying the different shard as different service? Then which peer is SolrJ using?

Such as Solr server nodes are `10.5.1.23:1000`(node1), `10.5.1.53:1000`(node2). What is SolrJ plugin peer? I am assuming something like `10.5.1.23:1000,10.5.1.53:1000`? If so, the topology can't show perfectly, because backend can't tell `10.5.1.23:1000,10.5.1.53:1000` is alias name of node1 or node2.

kezhenxu94:
> Then which peer is SolrJ using?

> Such as Solr server nodes are 10.5.1.23:1000(node1), 10.5.1.53:1000(node2). What is SolrJ plugin peer? 

@wu-sheng The peer of SolrJ Plugin may not be always the same since there're something like 'load balancing' in the request. The SolrJ would contact the zookeeper first to get a Solr node and then send the request to that node (afterwards the request may require communication between Solr nodes such as aggregation, which is none of the SolrJ's buniess), the request may not always be routed to the same node. Am I right? @dmsolr

dmsolr:
Great, @kezhenxu94 .
Communication between nodes use SolrJ withou SolrJ-plugin.
You mean there is a problem when SolrJ-plugin works with Solr-plugin in one node, right. @wu-sheng

apm-sniffer/apm-sdk-plugin/solr-7.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/solr/search/QueryStageTracerInterceptor.java
```diff
@@ -0,0 +1,70 @@
...
+            if (snapshot != null) {
+                ContextManager.continued(snapshot);
+            }
+        }
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                              Object ret) throws Throwable {
+        if (method.getName().equals("finishStage")) {
```
IanCao: i'm a bit confused about `if (method.getName().equals("distributedProcess"))` in before method and `if (method.getName().equals("finishStage")) {` in after method

this two method name should be same.but this logic judgment is not

apm-sniffer/apm-sdk-plugin/solr-7.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/solr/shard/ShardRequestInterceptor.java
```diff
@@ -0,0 +1,110 @@
...
+        SolrRequest<?> request = (SolrRequest<?>) allArguments[0];
+        if (!request.getPath().startsWith(SKIP_ADMIN_PREF) && ContextManager.isActive()) {
+            ContextManager.stopSpan();
+        }
+        return ret;
+    }
+
+    @Override
+    public void handleMethodException(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Throwable t) {
+        if (ContextManager.isActive()) {
```
IanCao: `!request.getPath().startsWith(SKIP_ADMIN_PREF)` should be here

apm-sniffer/apm-sdk-plugin/solr-7.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/solr/search/QueryStageTracerInterceptor.java
dmsolr: This span start at `distributedProcess()` and stop at `finishStage()`. 
But I'm pretty sure they always come in pairs.

apm-sniffer/apm-sdk-plugin/solr-7.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/solr/shard/ShardRequestInterceptor.java
dmsolr: Yes, I notice that. Maybe it is not required. In usually, there is no active span which matches this condition here. (Unless other interceptors create.)

That there is a active span in context, not matter of who created, logs the exception when it is in the method with catching exception. I think.

wu-sheng:
> You mean there is a problem when SolrJ-plugin works with Solr-plugin in one node, right.

Could you test about SolrJ plugin works with one node Solr server and multiple nodes cluster? Especially for Topology. I highly doubt the topology will make sense if you do so when APP with Solr cluster.

dmsolr:
I have tried it seems like ok.

image

wu-sheng:
What is the user accessing all nodes? Maybe because you are using browser access Solr server?

dmsolr:
Yes, I accessed all the nodes through the browser. 
Accessing Solr via browser only connects to one node.
But SolrJ can access all nodes in a request, expecially writting data. This is because the node to which the document should be written is usually known. If partition rule is given.

wu-sheng:
OK, if no peer looks like `ip1:port1,ip2:port2...` in SolrJ plugin, I think that is fine.

wu-sheng:
cc https://github.com/apache/skywalking/pull/2871

wu-sheng:
SolrJ plugin has been reverted, have to, because it breaks the agent.

asfgit:
FAILURE

asfgit:
https://builds.apache.org/job/skywalking-e2e/11/

dmsolr:
> ## Cases List
> 
> | Case description | Status | Cases|
> |:-----|:-----:|:-----:|
> |solr| **12 passed. 0 failed**| [click me](#solr) |
>

wu-sheng:
You need to fix conflict first.

dmsolr:
The PR has delayed for long time, so I works in a new branch. This branch lose some commits. Could I close this and open another PR?

dmsolr:
move to #3002

PR #2859, opened 2019-06-12
Add FAQ for ElasticSearch server side setting

wu-sheng:
This FAQ is based on community end-user feedback. FYI @kezhenxu94


docs/en/FAQ/ES-Server-FAQ.md
```diff
@@ -0,0 +1,21 @@
...
+
+You could add following config to `elasticsearch.yml`, set the value based on your env.
+```yml
+# In tracing scenario, consider to set more than this at least.
+thread_pool.bulk.queue_size: 1000
+
+thread_pool.bulk.size: 16
+thread_pool.index.queue_size: 500
+thread_pool.write.queuq_size: 500
+index.max_result_window: 1000000
```
peng-yongsheng: Why suggest the users setting their result window to a huge value?

wu-sheng: I think sure feedback about query trace failure. That is all. Do you think we should remove this?

PR #2866, opened 2019-06-13
 fix potential NullPointerException

tzy1316106836:

- [x] Bug fix
___
### Bug fix
- Bug description.
the inappropriate use of equals() 
- How to fix?
fix the use of equals(),fix potential NullPointerException


wu-sheng:
@IanCao CI is failing mostly because Circle CI change. I and @kezhenxu94 will make it back online soon.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/gc/GCProvider.java
```diff
@@ -63,7 +63,7 @@ private GCMetricAccessor findByBeanName(String name) {
         } else if (name.indexOf("G1") > -1) {
             // G1 collector ( -XX:+UseG1GC )
             return new G1GCModule(beans);
-        } else if (name.equals("MarkSweepCompact")) {
+        } else if ("MarkSweepCompact".equals(name)) {
```
wu-sheng: When the name is NULL?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/HierarchyMatch.java
```diff
@@ -95,7 +95,7 @@ private void matchHierarchyClass(TypeDescription.Generic clazz, List<String> par
         }
 
         TypeDescription.Generic superClazz = clazz.getSuperClass();
-        if (superClazz != null && !clazz.getTypeName().equals("java.lang.Object")) {
+        if (superClazz != null && !"java.lang.Object".equals(clazz.getTypeName())) {
```
wu-sheng: When class type name is NULL?

tzy1316106836: in this case, the class type is not null, but  equqls() was usually used in this style

wu-sheng:
First of all, this PR is not fixing potential NPE from my current understanding. For coding style, in these cases(your changes), I could only say it is harmless.

PR #2864, opened 2019-06-12
Integrate Jenkins CI

kezhenxu94:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

https://github.com/apache/skywalking/pull/2838#issuecomment-500182253

PR #2870, opened 2019-06-13
Adjust pom plugin position and add docs

kezhenxu94:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance
- [x] polish/documentation

- Related issues

#2864


docs/en/guides/How-to-build.md
```diff
@@ -142,5 +142,24 @@ a oap image with name `bar/oap:foo`, run the following command
 ```
 5. Run `./mvnw compile -Dmaven.test.skip=true`
 6. Run `maven update`. Must remove the clean projects item before maven update(This will be clear the proto conversion Java file generated by the complie)
-7. Run `./mvnw compile` complie collector-remote-grpc-provider and apm-protocol
+7. Run `./mvnw compile` compile collector-remote-grpc-provider and apm-protocol
 8. Refresh project
+
+## Write necessary tests and run them
```
wu-sheng: Add this to here https://github.com/apache/skywalking/blob/master/docs/en/guides/README.md#for-code-developer . Make sense?

kezhenxu94: Added

wu-sheng: I think we don't need this paragraph here. Build source doesn't mean write codes and contribute to upstream.

docs/en/guides/How-to-build.md
```diff
@@ -142,5 +142,24 @@ a oap image with name `bar/oap:foo`, run the following command
...
+
+After setting up the environment and writing your codes, to make it more easily accepted by SkyWalking project, you'll
```
wu-sheng: in order to make it

docs/en/guides/How-to-build.md
wu-sheng: I suggest move all to there

kezhenxu94: Got it, I misunderstood your meaning, I thought you're asking me to put the link there, what a embarrass :-)

PR #2871, opened 2019-06-13
Revert SolrJ plugin.

wu-sheng:
@dmsolr I have to do this emergency revert. This plugin breaks agent, because
1. Static import, `import org.apache.http.client.methods.HttpUriRequest;` This will break the agent when this doesn't exist.
1. `CloseableHttpClient` should not be instrumented by SolrJ plugin, it should be httpClient plugin's responsibility.


wu-sheng:
@IanCao Please notice this in other review.

dmsolr:
I am so sorry to heard that. I am going to recheck and remove instrumentation of CloseableHttpClient.

wu-sheng:
@dmsolr we are working on more automatically tests requirements and environments to make this smoothly next time.

Instrumentation is one thing. And please don't import the dependency class. Check other plugins, use string name.

dmsolr:
I got it. Thank you!

wu-sheng:
@dmsolr According to our offline discussion, I am looking forward to your script to avoid this happens in the future.

PR #2873, opened 2019-06-14
Fix no stream register.

wu-sheng:
Metrics and Inventory stream should register to **StreamDataMapping**. Miss this in current version.


wu-sheng:
@peng-yongsheng Also, I am adding stream class list order back, to avoid potential conflict in different JVM class scan mechanism.

PR #2838, opened 2019-06-08
Improve plugin performance

kezhenxu94:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues

fixes #2837 

___
### New feature or improvement

- Describe the details and related test reports.

replace `String.format` with `+` operator

PR #2874, opened 2019-06-14
Rewrite AtomicRangeInteger for higher performance

lkxiaolou:
- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

### improvement
- The function AtomicRangeInteger.getAndIncrement() was rewrote,  performs higher performance in many threads scene.  From benchmark it improves about 200% ~ 300% performance in 128 threads.
- It produce less blank element than the old one, and stay the same in other sides.


kezhenxu94:
Hi @lkxiaolou , technicality speaking, what you provided is not a benchmark test. Please refer to #2838 and provide one to verify it

lkxiaolou:
> Learn `jmh`, and write benchmark test if you want to propose performance improvement.

thanks, I have provided

lkxiaolou:
> Hi @lkxiaolou , technicality speaking, what you provided is not a benchmark test. Please refer to #2838 and provide one to verify it

thanks, I have provided

apm-commons/apm-datacarrier/src/test/java/org/apache/skywalking/apm/commons/datacarrier/common/AtomicRangeIntegerTest.java
```diff
@@ -39,4 +48,95 @@ public void testGetAndIncrement() {
...
+        ATOMIC_NEW.getAndIncrement();
+    }
+
+    public static void main(String[] args) throws RunnerException {
+        Options opt = new OptionsBuilder()
+                .include(AtomicRangeIntegerTest.class.getSimpleName())
+                .forks(1)
+                .threads(4)
+                .output("/tmp/jmh.log")
+                .warmupIterations(0)
```
wu-sheng: Add warmup to make the result more stable, at least 3.

apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/common/AtomicRangeInteger.java
```diff
@@ -37,7 +37,7 @@ public AtomicRangeInteger(int startValue, int maxValue) {
         this.endValue = maxValue - 1;
     }
 
-    public final int getAndIncrement() {
+    public final int oriGetAndIncrement() {
```
wu-sheng: This method should move to test, as the original method and test baseline.

apm-commons/apm-datacarrier/src/test/java/org/apache/skywalking/apm/commons/datacarrier/common/AtomicRangeIntegerTest.java
lkxiaolou: I have added, and add syncIterations false.

apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/common/AtomicRangeInteger.java
```diff
@@ -49,6 +49,18 @@ public final int getAndIncrement() {
         return current;
     }
 
+    public final int getAndIncrement() {
+        int next;
+        do {
+            next = this.value.incrementAndGet();
```
wu-sheng: From my understanding, you are using the `conflict abandon mechanism`, trade off more throughput. Because this line invokes, when comareAndSet fail, you can't get the index in this loop.

Please share with me, why you think this is better?  I think this is two sides of the coin
1. **[OLD]**, lower throughput, higher cache queue usage rate, less blank element.
1. **[NEW]**, opposite.

lkxiaolou: The conflict in the New implementation only happens on the edge of the buffer, so the loop count in  New implementation is less than the Old one for the edge-off after abandoning conflict. It will produce  less blank element than the old one.
I can't understand what is "higher cache queue usage rate" in this case, could you please explain it ?

lkxiaolou: > From my understanding, you are using the `conflict abandon mechanism`, trade off more throughput. Because this line invokes, when comareAndSet fail, you can't get the index in this loop.
> 
> Please share with me, why you think this is better? I think this is two sides of the coin
> 
> 1. **[OLD]**, lower throughput, higher cache queue usage rate, less blank element.
> 2. **[NEW]**, opposite.
The conflict in the New implementation only happens on the edge of the buffer, so the loop count in New implementation is less than the Old one for the edge-off after abandoning conflict. It will produce less blank element than the old one.
I can't understand what is "higher cache queue usage rate" in this case, could you please explain it ?

wu-sheng: Sorry, my fault, I misread the codes. Yes, the conflict only happens at the edge.

wu-sheng:
@kezhenxu94 @ascrutae @IanCao @candyleer @zhaoyuguang  Review opinion?

I want to hear more from yours. This PR includes only several lines, but effects agent and OAP server. So, we need multiple check.

wu-sheng:
> I have tested the function performance (in the unit test), which may has 5 - 10 times or more improve

@lkxiaolou Could you provide an official comment here. From benchmark, it improves a lot(surely), but not 5-10 times. Yes?

kezhenxu94:
I notice that both the old and new versions don't work well in concurrent scenario, but the new version is more obviously to be discovered:
    
```java
    int next;
    do {
      next = this.value.incrementAndGet(); // <-- here is the point, it always increment the `value`
      if (next > endValue && this.value.compareAndSet(next, startValue)) {
        return endValue;
      }
    } while (next > endValue);    
```
    
in the case of `org.apache.skywalking.apm.commons.datacarrier.buffer.Buffer`, it may leave some "black holes" in the queue.
though the class `AtomicRangeInteger` is only used in `org.apache.skywalking.apm.commons.datacarrier.buffer.Buffer`, and it handles the `null` values carefully, once the `AtomicRangeInteger` is used in other more scenarios, it may be a bug somehow, I am not sure whether this is possible to happen.
    
> Bellow is the test codes, which doesn't necessary always print 5000, sometimes are much less than 5000:

```java
    @Test
    public void testConcurrency() throws InterruptedException {
        int count = 5000;
        final AtomicRangeInteger integer = new AtomicRangeInteger(0, count);
        final CountDownLatch latch = new CountDownLatch(1);
        final CountDownLatch mainLatch = new CountDownLatch(count);
        final Set<Integer> seen = new HashSet<Integer>();
        final ExecutorService executor = Executors.newCachedThreadPool();
        for (int i = 0; i < count; i++) {
            executor.submit(new Runnable() {
                @Override public void run() {
                    try {
                        latch.await();
                        int i = integer.getAndIncrement();
                        seen.add(i);
                        mainLatch.countDown();
                        e.printStackTrace();
                    }
                }
            });
        }
        latch.countDown();
        mainLatch.await();
        System.out.println("seen.size() = " + seen.size());
        executor.shutdown();
    }
```

wu-sheng:
> though the class AtomicRangeInteger is only used in org.apache.skywalking.apm.commons.datacarrier.buffer.Buffer, and it handles the null values carefully, once the AtomicRangeInteger is used in other more scenarios, it may be a bug somehow, I am not sure whether this is possible to happen.

Yes. That is true, including your other words. This AtomicInteger is designed for DataCarrier queue specifically. Tradeoff increment strictly with higher performance, or higher in the new version.

lkxiaolou:
@kezhenxu94 
you should use a thread-safe set
`final Set<Integer> seen = Collections.synchronizedSet(new HashSet<Integer>());`

kezhenxu94:
> @kezhenxu94
> you should use a thread-safe set
> `final Set<Integer> seen = Collections.synchronizedSet(new HashSet<Integer>());`

uh ha, yes, sorry for my mistake

wu-sheng:
For the record, this OP improvement has 3x better than the existing one.

``` 
  /**
     * # JMH version: 1.21
     * # VM version: JDK 1.8.0_111, Java HotSpot(TM) 64-Bit Server VM, 25.111-b14
     * # VM invoker: /Library/Java/JavaVirtualMachines/jdk1.8.0_111.jdk/Contents/Home/jre/bin/java
     * # VM options: -Dfile.encoding=UTF-8
     * # Warmup: 3 iterations, 10 s each
     * # Measurement: 5 iterations, 10 s each
     * # Timeout: 10 min per iteration
     * # Threads: 128 threads, ***WARNING: Synchronize iterations are disabled!***
     * # Benchmark mode: Throughput, ops/time
     * # Benchmark: org.apache.skywalking.apm.commons.datacarrier.common.AtomicRangeIntegerTest.testNewGetAndIncrementPerformance
     *
     * # Run progress: 0.00% complete, ETA 00:02:40
     * # Fork: 1 of 1
     * # Warmup Iteration   1: 41358475.014 ops/s
     * # Warmup Iteration   2: 40973232.064 ops/s
     * # Warmup Iteration   3: 41310422.853 ops/s
     * Iteration   1: 41557782.370 ops/s
     * Iteration   2: 42723032.686 ops/s
     * Iteration   3: 41957321.407 ops/s
     * Iteration   4: 40708422.580 ops/s
     * Iteration   5: 40424870.574 ops/s
     *
     *
     * Result "org.apache.skywalking.apm.commons.datacarrier.common.AtomicRangeIntegerTest.testNewGetAndIncrementPerformance":
     *   41474285.923 ±(99.9%) 3595500.827 ops/s [Average]
     *   (min, avg, max) = (40424870.574, 41474285.923, 42723032.686), stdev = 933740.147
     *   CI (99.9%): [37878785.096, 45069786.751] (assumes normal distribution)
     *
     *
     * # JMH version: 1.21
     * # VM version: JDK 1.8.0_111, Java HotSpot(TM) 64-Bit Server VM, 25.111-b14
     * # VM invoker: /Library/Java/JavaVirtualMachines/jdk1.8.0_111.jdk/Contents/Home/jre/bin/java
     * # VM options: -Dfile.encoding=UTF-8
     * # Warmup: 3 iterations, 10 s each
     * # Measurement: 5 iterations, 10 s each
     * # Timeout: 10 min per iteration
     * # Threads: 128 threads, ***WARNING: Synchronize iterations are disabled!***
     * # Benchmark mode: Throughput, ops/time
     * # Benchmark: org.apache.skywalking.apm.commons.datacarrier.common.AtomicRangeIntegerTest.testOriGetAndIncrementPerformance
     *
     * # Run progress: 50.00% complete, ETA 00:01:25
     * # Fork: 1 of 1
     * # Warmup Iteration   1: 14169937.124 ops/s
     * # Warmup Iteration   2: 14087015.239 ops/s
     * # Warmup Iteration   3: 13955313.979 ops/s
     * Iteration   1: 13984516.590 ops/s
     * Iteration   2: 13913174.492 ops/s
     * Iteration   3: 13824113.805 ops/s
     * Iteration   4: 14886990.831 ops/s
     * Iteration   5: 14388283.816 ops/s
     *
     *
     * Result "org.apache.skywalking.apm.commons.datacarrier.common.AtomicRangeIntegerTest.testOriGetAndIncrementPerformance":
     *   14199415.907 ±(99.9%) 1697559.657 ops/s [Average]
     *   (min, avg, max) = (13824113.805, 14199415.907, 14886990.831), stdev = 440850.852
     *   CI (99.9%): [12501856.250, 15896975.564] (assumes normal distribution)
     *
     *
     * # Run complete. Total time: 00:02:51
     *
     * REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
     * why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
     * experiments, perform baseline and negative tests that provide experimental control, make sure
     * the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
     * Do not assume the numbers tell you what you want them to tell.
     *
     * Benchmark                                                  Mode  Cnt         Score         Error  Units
     * AtomicRangeIntegerTest.testNewGetAndIncrementPerformance  thrpt    5  41474285.923 ± 3595500.827  ops/s
     * AtomicRangeIntegerTest.testOriGetAndIncrementPerformance  thrpt    5  14199415.907 ± 1697559.657  ops/s
     */
```

lkxiaolou:
> @lkxiaolou For making merge safer, could you add a benchmark in multiple thread case? Let's formalize a result of that.
> 
> A case like, use 6 threads to get 10k+ from a 1000 size of range concurrently. Let's see the performance look like.

Is that means I should provide a benchmark with certain threads, buffer size, run times on the new implementation ? I had provided a benchmark with 128 threads, 100 buffer size, 5 x 10s run time

wu-sheng:
@lkxiaolou Please confirm, from the benchmark result, 3x improvement not 5+. Right?

lkxiaolou:
> @lkxiaolou Please confirm, from the benchmark result, 3x improvement not 5+. Right?

It depends on the buffer size and threads count, I make a test with 10000 buffer size and 1024 threads get 5+improvement.
Theoretically, the bigger buffer size and the more threads make the more improvement, at least 3x.

`# JMH version: 1.21
# VM version: JDK 1.8.0_111, Java HotSpot(TM) 64-Bit Server VM, 25.111-b14
# VM invoker: /Library/Java/JavaVirtualMachines/jdk1.8.0_111.jdk/Contents/Home/jre/bin/java
# VM options: -Dfile.encoding=UTF-8
# Warmup: 3 iterations, 10 s each
# Measurement: 5 iterations, 1000 ms each, 1000 calls per op
# Threads: 1024 threads, ***WARNING: Synchronize iterations are disabled!***
# Benchmark: org.apache.skywalking.apm.commons.datacarrier.common.AtomicRangeIntegerTest.testNewGetAndIncrementPerformance

# Run progress: 0.00% complete, ETA 00:01:10
# Fork: 1 of 1
# Warmup Iteration   1: 2078993549.108 ops/s
# Warmup Iteration   2: 6408026002.297 ops/s
# Warmup Iteration   3: 5455960135.636 ops/s
Iteration   1: 165330.972 ops/s
Iteration   2: 95420.526 ops/s
Iteration   3: 173340.089 ops/s
Iteration   4: 94214.669 ops/s
Iteration   5: 261606.088 ops/s


Result "org.apache.skywalking.apm.commons.datacarrier.common.AtomicRangeIntegerTest.testNewGetAndIncrementPerformance":
  157982.469 ±(99.9%) 265443.465 ops/s [Average]
  (min, avg, max) = (94214.669, 157982.469, 261606.088), stdev = 68934.825
  CI (99.9%): [≈ 0, 423425.934] (assumes normal distribution)


# JMH version: 1.21
# VM version: JDK 1.8.0_111, Java HotSpot(TM) 64-Bit Server VM, 25.111-b14
# VM invoker: /Library/Java/JavaVirtualMachines/jdk1.8.0_111.jdk/Contents/Home/jre/bin/java
# VM options: -Dfile.encoding=UTF-8
# Warmup: 3 iterations, 10 s each
# Measurement: 5 iterations, 1000 ms each, 1000 calls per op
# Threads: 1024 threads, ***WARNING: Synchronize iterations are disabled!***
# Benchmark: org.apache.skywalking.apm.commons.datacarrier.common.AtomicRangeIntegerTest.testOriGetAndIncrementPerformance

# Run progress: 50.00% complete, ETA 00:00:43
# Fork: 1 of 1
# Warmup Iteration   1: 3345520433.355 ops/s
# Warmup Iteration   2: 4823749249.799 ops/s
# Warmup Iteration   3: 6676209367.819 ops/s
Iteration   1: 34885.954 ops/s
Iteration   2: 33128.147 ops/s
Iteration   3: 31300.865 ops/s
Iteration   4: 19188.602 ops/s
Iteration   5: 34664.282 ops/s


Result "org.apache.skywalking.apm.commons.datacarrier.common.AtomicRangeIntegerTest.testOriGetAndIncrementPerformance":
  30633.570 ±(99.9%) 25249.253 ops/s [Average]
  (min, avg, max) = (19188.602, 30633.570, 34885.954), stdev = 6557.151
  CI (99.9%): [5384.317, 55882.822] (assumes normal distribution)


# Run complete. Total time: 00:01:26

REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
experiments, perform baseline and negative tests that provide experimental control, make sure
the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
Do not assume the numbers tell you what you want them to tell.

Benchmark                                                  Mode  Cnt       Score        Error  Units
AtomicRangeIntegerTest.testNewGetAndIncrementPerformance  thrpt    5  157982.469 ± 265443.465  ops/s
AtomicRangeIntegerTest.testOriGetAndIncrementPerformance  thrpt    5   30633.570 ±  25249.253  ops/s
`

wu-sheng:
@lkxiaolou Merged. Thanks for providing this.

wu-sheng:
@lkxiaolou You are on the SkyWalking official twitter today :) https://twitter.com/ASFSkyWalking/status/1139771184210313219

lkxiaolou:
> @lkxiaolou You are on the SkyWalking official twitter today :) https://twitter.com/ASFSkyWalking/status/1139771184210313219

thanks

PR #2881, opened 2019-06-15
Remove imported classes and add ci scripts for plugin import check.

dmsolr:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

Imported 3rd-party classes unless JDK or ByteBuddy maybe trigger agent crash. So I remove it. And provide a shell to check it. But the script is very weak.


PluginImportedCheck.sh
```diff
@@ -0,0 +1,10 @@
+plugin_dir=$1
+for dir in `ls "./apm-sniffer/$plugin_dir/"`; do
+	for f in `find ./apm-sniffer/$plugin_dir/$dir -name *Instrumentation.java `; do
+		NUM=`head -400 $f | grep import |grep -v net.bytebuddy. | grep -v org.apache.skywalking. |grep -v java.| wc -l`;
```
kezhenxu94: `grep import` => `grep ^import` should be more accurately

Jenkinsfile
```diff
@@ -56,7 +56,15 @@ pipeline {
                                 sh 'ls'
                                 sh 'git status'
                             }
-                        }    
+                        }
+
+                        stage('Check 3rd-party classes imported') {
```
wu-sheng: Named as Check agent plugin....

apm-sniffer/apm-sdk-plugin/motan-plugin/src/main/java/org/apache/skywalking/apm/plugin/motan/define/MotanConsumerInstrumentation.java
```diff
@@ -41,7 +40,7 @@
      */
     private static final String CONSTRUCTOR_INTERCEPT_CLASS = "org.apache.skywalking.apm.plugin.motan.MotanConsumerInterceptor";
     /**
-     * Class that intercept {@link com.weibo.api.motan.rpc.AbstractProvider#call(Request)}.
+     * Class that intercept {@link com.weibo.api.motan.rpc.AbstractProvider#call(com.weibo.api.motan.rpc.Request)}.
```
wu-sheng: Don't use @link, use iteral string only.

wu-sheng:
Grant shell to execute right, run commit the shell

apm-sniffer/apm-sdk-plugin/motan-plugin/src/main/java/org/apache/skywalking/apm/plugin/motan/define/MotanConsumerInstrumentation.java
kezhenxu94: > Don't use @link, use iteral string only.

Use `{@code ..}` is also acceptable?

wu-sheng: I think both are safe for class loader. I recommend remove link because this class wouldn't exist in agent.

PluginImportedCheck.sh
dmsolr: Yes

tools/check/agent/plugin/PluginImportedCheck.sh
```diff
@@ -0,0 +1,30 @@
...
+# specific language governing permissions and limitations
+# under the License.
+# ----------------------------------------------------------------------------
+
+plugin_dir=$1
+for dir in `ls "./apm-sniffer/$plugin_dir/"`; do
+	for f in `find ./apm-sniffer/$plugin_dir/$dir -name *Instrumentation.java `; do
+		NUM=`head -400 $f | grep ^import |grep -v net.bytebuddy. | grep -v org.apache.skywalking. |grep -v java.| wc -l`;
+		if [ $NUM -gt 0 ] ; then
+			echo "Plugin: $dir, never import any class unless JDK and ByteBuddy!";
```
wu-sheng: Change to, only allow to import JDK and ByteBuddy classes

tools/check/agent/plugin/PluginImportedCheck.sh
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+# ----------------------------------------------------------------------------
+
+plugin_dir=$1
+for dir in `ls "./apm-sniffer/$plugin_dir/"`; do
```
wu-sheng: Is this path still working after you moved it?

tools/check/agent/plugin/PluginImportedCheck.sh
```diff
@@ -0,0 +1,30 @@
...
+	for f in `find ./apm-sniffer/$plugin_dir/$dir -name *Instrumentation.java `; do
+		NUM=`head -400 $f | grep ^import |grep -v net.bytebuddy. | grep -v org.apache.skywalking. |grep -v java.| wc -l`;
```
wu-sheng: I think we should output the class file name and path to log. We could show explicitly which file has illegal import.

wu-sheng: I mean output all instrumentation class list.

kezhenxu94: The regexp is not perfect, I tried that I can import `import javassist.ClassMap;` and the script did not catch this.

tools/check/agent/plugin/PluginImportedCheck.sh
```diff
@@ -0,0 +1,32 @@
...
+# specific language governing permissions and limitations
+# under the License.
+# ----------------------------------------------------------------------------
+
+plugin_dir=$1
+for dir in `ls "./apm-sniffer/$plugin_dir/"`; do
+	echo "Scanning $dir"
+	for f in `find ./apm-sniffer/$plugin_dir/$dir -name *Instrumentation.java `; do
+		NUM=`head -400 $f | grep ^import |grep -Ev "^import\s+(static\s+)*net.bytebuddy\\." \
+			| grep -Ev "^import\s+(static\s+)*org.apache.skywalking\\." |grep -Ev "^import\s+(static\s+)*java(x)*\\." | wc -l`
```
kezhenxu94: I don't think `javax` is allowed here, there are so many third-party libraries whose name contains `javax`
image

dmsolr: I am not sure whether we need javax.sql.

wu-sheng: We don't need javax

wu-sheng: In instrumentation.

dmsolr: I got it. thx.

wu-sheng: OK. I will wait your new commit.

wu-sheng:
@dmsolr Thanks for providing this valuable scripts in CI. @ascrutae I am merging this CI scripts, which could very helpful for our further plugin review.

PR #2888, opened 2019-06-17
Restore solrj-plugin

dmsolr:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#2871 
#2730

### DIFF
--- a/oap-server/server-core/src/test/resources/component-libraries.yml
+++ b/oap-server/server-core/src/test/resources/component-libraries.yml
@@ -198,6 +198,12 @@ spring-cloud-gateway:
 RESTEasy:
   id: 62
   languages: Java
+SolrJ:
+  id: 63
+  languages: Java
+Solr:
+  id: 64
+  languages: Java
 
 # .NET/.NET Core components
 # [3000, 4000) for C#/.NET only
@@ -295,3 +301,4 @@ Component-Server-Mappings:
   Pomelo.EntityFrameworkCore.MySql: Mysql
   Npgsql.EntityFrameworkCore.PostgreSQL: PostgreSQL
   transport-client: Elasticsearch
+  SolrJ: Solr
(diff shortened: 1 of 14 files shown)


dmsolr:
> ## Cases List
> 
> | Case description | Status | Cases|
> |:-----|:-----:|:-----:|
> |solrj| **12 passed. 0 failed**| [click me](#solrj) |
>

wu-sheng:
@IanCao Could you re-run this, and we could merge this agian.

dmsolr:
> inline

Sorry, @IanCao . I don't really understand what you mean.

wu-sheng:
@IanCao I think you miss your comments.

wu-sheng:
Here is the [test report](http://github.com/SkywalkingTest/agent-integration-test-report/tree/master/2019/6/dmsolr/testReport-master-2019-06-17-17-07.md) and [validate logs](http://35.227.88.23:49001/job/skywalking-agent-test/137/artifact/workspace/logs/validate_2019-06-17-17-07.log)

oap-server/server-core/src/test/resources/component-libraries.yml
```diff
@@ -198,6 +198,12 @@ spring-cloud-gateway:
 RESTEasy:
   id: 62
   languages: Java
+SolrJ:
+  id: 63
+  languages: Java
+Solr:
```
kezhenxu94: Is this what you will have in the other PR #2853 ? If so, leave it to where it should be

oap-server/server-starter/src/main/resources/component-libraries.yml
```diff
@@ -216,6 +216,13 @@ spring-cloud-gateway:
 RESTEasy:
   id: 62
   languages: Java
+SolrJ:
+  id: 63
+  languages: Java
+Solr:
```
kezhenxu94: Is this what you will have in the other PR #2853 ? If so, leave it to where it should be

oap-server/server-core/src/test/resources/component-libraries.yml
wu-sheng: This is for both server plugin and server mapping. So it makes sense to be here from I known.

PR #2891, opened 2019-06-17
fix cannot use 'System.out' through configuration.

dmsolr:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
According documents, we can use `System.out` to output logs through set `logging.dir` as null or blank string.


dmsolr:
I think logging to Console is still useful, especially when testing under Windows. 
So I think we need it. And set the default is FileWriter. 
This is just my opinion.

PR #2908, opened 2019-06-19
Check disallowed imports in agent instrumentation classes

kezhenxu94:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance
- [x] Enhancement

- Related issues

As described in #2871 , we merged a PR that breaks the agent by importing external classes, and a script to check that is provided in #2881 , here I'm proposing another way to check this before our contributors commit and push their codes, by leveraging the checkstyle plugin, if the contributors have a checkstyle plugin installed in their IDE, it would show an error when writing codes, before submitting, which would save some time.

image


kezhenxu94:
> Check style module is being removed, right?

Yes, the module is just used for holding the `checkstyle.xml` file, and the file can be simply placed in the directory. After removing it, changing the checkstyle.xml doesn't require a deployment of that module.

> And do I still need that scripts to check? Could we remove that?

Removing it or not is both ok to me as long as it does no harm

wu-sheng:
If you can remove, remove. Because we need consistent configuration in one place. With this PR, it shows in both places.

kezhenxu94:
> If you can remove, remove. Because we need consistent configuration in one place. With this PR, it shows in both places.

done

dmsolr:
You are so fast @kezhenxu94 . I'm going to do it.

PR #2917, opened 2019-06-20
Introduce dead line settings from gRPC document.

wu-sheng:
We have listened about at some bad network environment, people report gRPC not responding sometimes.
Following gRPC document of the deadline, I think this fits for us. https://grpc.io/blog/deadlines/

10 seconds is the value I set for 
1. Register, JVM reporter and Segment reporters at agent side.
1. Remote client and gRPC exporter at OAP side.

I don't set any sever side timeout, because all of gRPC clients don't require the response.

Any suggestion?

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMService.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMService.java
@@ -151,7 +151,7 @@ public class JVMService implements BootService, Runnable {
         public void statusChanged(GRPCChannelStatus status) {
             if (GRPCChannelStatus.CONNECTED.equals(status)) {
                 Channel channel = ServiceManager.INSTANCE.findService(GRPCChannelManager.class).getChannel();
-                stub = JVMMetricReportServiceGrpc.newBlockingStub(channel);
+                stub = JVMMetricReportServiceGrpc.newBlockingStub(channel).withDeadlineAfter(10, TimeUnit.SECONDS);
             }
             this.status = status;
         }
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/ServiceAndEndpointRegisterClient.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/ServiceAndEndpointRegisterClient.java
@@ -65,8 +65,8 @@ public class ServiceAndEndpointRegisterClient implements BootService, Runnable, 
     public void statusChanged(GRPCChannelStatus status) {
         if (GRPCChannelStatus.CONNECTED.equals(status)) {
             Channel channel = ServiceManager.INSTANCE.findService(GRPCChannelManager.class).getChannel();
-            registerBlockingStub = RegisterGrpc.newBlockingStub(channel);
-            serviceInstancePingStub = ServiceInstancePingGrpc.newBlockingStub(channel);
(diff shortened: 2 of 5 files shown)


wu-sheng:
This is the error log when deadline/timeout happens at the client, even in stream mode. Exception accepts in `#onError` method.
```
io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED: deadline exceeded after 969455382ns
	at io.grpc.Status.asRuntimeException(Status.java:526)
	at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:434)
	at io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:39)
	at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
	at io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
	at io.grpc.internal.CensusStatsModule$StatsClientInterceptor$1$1.onClose(CensusStatsModule.java:678)
	at io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:39)
	at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
	at io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
	at io.grpc.internal.CensusTracingModule$TracingClientInterceptor$1$1.onClose(CensusTracingModule.java:403)
	at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:459)
	at io.grpc.internal.ClientCallImpl.access$300(ClientCallImpl.java:63)
	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl.close(ClientCallImpl.java:546)
	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl.access$600(ClientCallImpl.java:467)
```

peng-yongsheng:
The document you gave have description is 
- `By default this deadline is a very large number, dependent on the language implementation.`
- `This sets the deadline to 100ms from when the client RPC is set to when the response is picked up by the client.`

But from the error log you gave is that `deadline exceeded after 969455382ns`.

 **Important thing: 969455382ns == 900ms == 0.9s**
We don't set the deadline before this PR, and our agent and OAP server are written with Java language, so the deadline should the default value of 100ms.

Question: 
- I'm very curious about why the deadline exceeded is 900ms instead of 100ms?
- Why you set the deadline to 10s? The document said: `In general, when you don’t set a deadline, resources will be held for all in-flight requests, and all requests can potentially reach the maximum timeout. This puts the service at risk of running out of resources, like memory, which would increase the latency of the service, or could crash the entire process in the worst case.` I think the 10s deadline is too large because the agent sends the segments to the server-side every 30s. No more time for the next round. I suggest make it smaller is best. **5s**?

@wu-sheng

wu-sheng:
> Important thing: 969455382ns == 900ms == 0.9s

Where do you get this? 

From my local test, without setting a deadline, it doesn't end in 0.9s. I tried to wait 20 seconds in stream mode, still no timeout response.

peng-yongsheng:
> > Important thing: 969455382ns == 900ms == 0.9s
> 
> Where do you get this?
> 
> From my local test, without setting a deadline, it doesn't end in 0.9s. I tried to wait 20 seconds in stream mode, still no timeout response.

Here it is: `io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED: deadline exceeded after 969455382ns`.

wu-sheng:
> Here it is: io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED: deadline exceeded after 969455382ns.

This is the result after I set the parameter, not before. I think you misunderstand.

Issue #2926, opened 2019-06-21
java.lang.OutOfMemoryError: Java heap space


SummerOfServenteen:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS and JRE?
SkyWalking：6.1.0,jre:1.8
- Which company or project?
jingling
- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.

___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

kezhenxu94:
@SummerOfServenteen Can you provide more information?

wu-sheng:
@kezhenxu94 We only set version if we plan to accept PR. Before that, we discuss.

SummerOfServenteen:
hello.When we use skyWalking ,it cause Java heap space。after analyze the dump file ,we find it's conflict with Sentinel. 
Here is the detail.

![image](https://user-images.githubusercontent.com/31895647/59915229-bb2a9200-944e-11e9-945b-7c0fb19c6e0e.png)
![image](https://user-images.githubusercontent.com/31895647/59915288-dac1ba80-944e-11e9-8ba3-845a6d6b7322.png)
![image](https://user-images.githubusercontent.com/31895647/59915318-edd48a80-944e-11e9-9dbd-d19f2bdcf66b.png)
![image](https://user-images.githubusercontent.com/31895647/59915331-f62cc580-944e-11e9-81a0-724721955ffb.png)
Finally,i find it's method of org.apache.skywalking.apm.agent.core.context.util.ThrowableTransformer.convert2String that cause outOfMemory .

Code analyzing...
![image](https://user-images.githubusercontent.com/31895647/59915716-e3ff5700-944f-11e9-8ed0-da89195e24b1.png)
![image](https://user-images.githubusercontent.com/31895647/59915738-ee215580-944f-11e9-8792-d7d0639e7e77.png)

Because of Sentinel had overWrite the method of  fillInStackTrace(),so the stackTrace of size is 0,
the return value of printStackElement is always false.Then the method of convert2String is in Infinite loop.

My idea is rewrite the method of printStackElement like this :
![image](https://user-images.githubusercontent.com/31895647/59916195-15c4ed80-9451-11e9-977e-98be9975a40e.png)

[...]

PR #2928, opened 2019-06-21
Update ThrowableTransformer.java

SummerOfServenteen:
to avoid Infinite loop

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
fix java.lang.OutOfMemoryError
- How to fix?
if stackTrace.length == 0 ,then return true
___

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/ThrowableTransformer.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/ThrowableTransformer.java
@@ -58,7 +58,7 @@ public enum ThrowableTransformer {
     }
 
     private boolean printStackElement(StackTraceElement[] stackTrace, AppendListener printListener) {
-        if(stackTrace.length == 0){
+        if (stackTrace.length == 0) {
             return true;
         }
         for (StackTraceElement traceElement : stackTrace) {
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/ThrowableTransformer.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/ThrowableTransformer.java
@@ -58,6 +58,9 @@ public enum ThrowableTransformer {
     }
 
     private boolean printStackElement(StackTraceElement[] stackTrace, AppendListener printListener) {
+        if(stackTrace.length == 0){
+            return true;
+        }
         for (StackTraceElement traceElement : stackTrace) {
             printListener.append("at " + traceElement + LINE_SEPARATOR);
             if (printListener.overMaxLength()) {


dmsolr:
I don't think it's because of this. I run below test with `-Xmx100m`.
```java
@Test
    public void testOOM() {
        ThrowableTransformer instance = ThrowableTransformer.INSTANCE;

        while(true) {

            Exception e = new Exception();
            for (int level=0; level<100; level++) {
                e = new Exception("level : " + level, e);
            }

            instance.convert2String(e, 100);
        }
    }
```

It can run for a long time. And it is normal for GC.

By the way, I think your code cannot work in expected.

SummerOfServenteen:
@dmsolr , because e.getStackTrace().length>0 in you code,but the com.alibaba.csp.sentinel.slots.block.flow.FlowException.getStackTrace().length=0

wu-sheng:
That is whay I said in the issue, this looks like a bug for them. Could you share why stack length is 0?

dmsolr:
if stackTrace is empty, the exception message is appended only. right? 
This is not cause of oom, i think.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/ThrowableTransformer.java
```diff
@@ -58,6 +58,9 @@ private String printExceptionInfo(Throwable causeException) {
     }
 
     private boolean printStackElement(StackTraceElement[] stackTrace, AppendListener printListener) {
+        if (stackTrace.length == 0) {
+            return true;
```
wu-sheng: Why return `true` could void OOM? If I am reading code right, this root cause is `Exception` is parent endless loop?

wu-sheng:
Ref #2926

wu-sheng:
@SummerOfServenteen Issue submitted to Sentinel community directly. https://github.com/alibaba/Sentinel/issues/856

wu-sheng:
Also twitter this, https://twitter.com/wusheng1108/status/1142058624308502528, asking for help from Java experts

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/ThrowableTransformer.java
SummerOfServenteen: it seems cause by he has override fillInStackTrace
image,but i don't know what's the mean of this.

SummerOfServenteen: Because when SentinelRpcException happens,printListener.overMaxLength() is never be invoked,then the value of overMaxLength is always false.so it cause endless loop.
image

wu-sheng:
I will raise a better one.

PR #2930, opened 2019-06-21
One more time for heigher performance

lkxiaolou:
- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

AtomicRangeInteger v1 is the original implementation.
AtomicRangeInteger v2 is rewrote by me which impoves 3x~10x performance,.
And today I'm wrote AtomicRangeInteger v3 which improves about 30% performance to the v2.

The V3 algorithms is the same as v2, and I'm learning that AtomicRangeInteger V2 _value_ is a false-sharing with the _startValue_ and _endValue_, in the v3 implementation it was disabled the false-sharing with a 64 byte array to store the _value_ , so get the _startValue_ or _endValue_ can use the CPU's L1 cache line.


wu-sheng:
Please fix conflict and let's discuss this. Interesting.

lkxiaolou:
> Please fix conflict and let's discuss this. Interesting.

done

apm-commons/apm-datacarrier/src/test/java/org/apache/skywalking/apm/commons/datacarrier/common/AtomicRangeIntegerV2.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+
+package org.apache.skywalking.apm.commons.datacarrier.common;
+
+import java.io.Serializable;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Created by wusheng on 2016/10/25.
```
wu-sheng: You are the creator since V2, I am not :)

wu-sheng: And move V1 and V2 to test folders only.

lkxiaolou: > And move V1 and V2 to test folders only.

yes,  in the test folders already

lkxiaolou: > You are the creator since V2, I am not :)

uha, fixed

wu-sheng:
> The V3 algorithms is the same as v2, and I'm learning that AtomicRangeInteger V2 value is a false-sharing with the startValue and endValue, in the v3 implementation it was disabled the false-sharing with a 64 byte array to store the value , so get the startValue or endValue can use the CPU's L1 cache line.

First, the CI is still failing, please fix. Then could you point me to someplace to learn about `how is this making sure that benefit from the L1 cache`? I have read the test result and saw the increase. But honestly, this is out of my knowledge, so I need to learn and review. So basically, I need yours helps to get me through this.

@kezhenxu94 @JaredTan95 @hanahmily @IanCao or any other @apache/skywalking-committers if you have this field knowledge, please review and help me on this.

apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/common/AtomicRangeInteger.java
```diff
@@ -18,54 +18,106 @@
 
 
 package org.apache.skywalking.apm.commons.datacarrier.common;
+import sun.misc.Unsafe;
```
wu-sheng: Is this safe for all JREs? Including OpenJDK and new versions? I remember Oracle changed the package name once.

lkxiaolou: Yes,  Unsafe maybe change in the future, I have changed the code, just  a small change on the v2. 
This implementation also used the L1 cache, so it's very fast.
Disable the false-sharing reference from the Disruptor, Disruptor's Sequence also use placeholders to full a cache line so that the increase value changes doesn't affect others cache.

see [Sequence](https://github.com/LMAX-Exchange/disruptor/blob/master/src/main/java/com/lmax/disruptor/Sequence.java)

apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/common/AtomicRangeInteger.java
```diff
@@ -18,23 +18,32 @@
...
+
     private int startValue;
     private int endValue;
 
     public AtomicRangeInteger(int startValue, int maxValue) {
         this.value = new AtomicInteger(startValue);
         this.startValue = startValue;
         this.endValue = maxValue - 1;
+        leftPlaceholders = new int[15];
+        rightPlaceholders = new int[15];
```
wu-sheng: I don't find the place to use `leftPlaceholders` and `rightPlaceholders`. And why the length is 15?

wu-sheng: I think I need a little more explanation about this.

lkxiaolou: sorry, I have make some tests to find my first version is right. so I'm roll back the code.
Use an array ensure the value load in CPU's cache and cached in one line without other variable.

We can learn the Disruptor how to do this in this article

[https://tech.meituan.com/2016/11/18/disruptor.html](https://tech.meituan.com/2016/11/18/disruptor.html)

The Unsafe is used in the Disruptor, and it worked well. So we don't worry about Unsafe will change in the future.

kezhenxu94: > I don't find the place to use `leftPlaceholders` and `rightPlaceholders`. And why the length is 15?

It's indeed not used in the code because it's only declared to pad the cache line (64 bytes per line ), 15 * 4 = 60,  plus the value to occupy the entire cache line to dismiss the false-sharing, am I right ? @lkxiaolou but why do we need padding right? Did I miss anything ?

kezhenxu94: > sorry, I have make some tests to find my first version is right. so I'm roll back the code.
> Use an array ensure the value load in CPU's cache and cached in one line without other variable.
> 
> We can learn the Disruptor how to do this in this article
> 
> https://tech.meituan.com/2016/11/18/disruptor.html
> 
> The Unsafe is used in the Disruptor, and it worked well. So we don't worry about Unsafe will change in the future.

@wu-sheng the unsafe is also used in netty too, I don't think it's a concern herr

lkxiaolou: > > I don't find the place to use `leftPlaceholders` and `rightPlaceholders`. And why the length is 15?
> 
> It's indeed not used in the code because it's only declared to pad the cache line (64 bytes per line ), 15 * 4 = 60, plus the value to occupy the entire cache line to dismiss the false-sharing, am I right ? @lkxiaolou but why do we need padding right? Did I miss anything ?

The value may cached  anywhere in one line,.
 We must ensure one cache line cached the value and can not cache other variables.
 So suppose that cache the value in the end slot of the cache line, we need 15 left and 15 right.

kezhenxu94: > > > I don't find the place to use `leftPlaceholders` and `rightPlaceholders`. And why the length is 15?
> > 
> > 
> > It's indeed not used in the code because it's only declared to pad the cache line (64 bytes per line ), 15 * 4 = 60, plus the value to occupy the entire cache line to dismiss the false-sharing, am I right ? @lkxiaolou but why do we need padding right? Did I miss anything ?
> 
> The value may cached anywhere in one line,.
> We must ensure one cache line cached the value and can not cache other variables.
> So suppose that cache the value in the end slot of the cache line or start slot of the cache line, we need 15 left and 15 right.

Got it, really interesting

apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/common/AtomicRangeInteger.java
```diff
@@ -18,54 +18,108 @@
...
+        return UNSAFE.compareAndSwapInt(values, VALUE_BYTE_OFFSET, expect, update);
+    }
+
+    private final int incrementAndGet() {
+        return UNSAFE.getAndAddInt(values, VALUE_BYTE_OFFSET, 1) + 1;
+    }
+
     public final int get() {
-        return this.value.get();
+        return this.values[VALUES_OFFSET];
```
wu-sheng: This value never changed. You saved start value, not current value. I know this is never used, but still not right.

lkxiaolou: Current value has changed by the Unsafe native function, see compareAndSet and incrementAndGet.

wu-sheng:
I have a quick read about line cache, fail share and disruptor articles.
Now I could understand why this could be quicker. But I am not sure all codes, still a little hard about static block.
The thing I am still thinking about is the sun.unsafe incompatible risk. I was facing disruptor incompatible issue in IBM j9vm.

I will go to Shanghai KubeCon to give a talk. So, may be a little late.

lkxiaolou:
> I have a quick read about line cache, fail share and disruptor articles.
> Now I could understand why this could be quicker. But I am not sure all codes, still a little hard about static block.
> The thing I am still thinking about is the sun.unsafe incompatible risk. I was facing disruptor incompatible issue in IBM j9vm.
> 
> I will go to Shanghai KubeCon to give a talk. So, may be a little late.

Yes, sun.unsafe is not safe, I'm thinking about a better solution.
In java 8 we can use @Contended to solve the false-sharing problem, but in java 7 or lower version may got a compile error.

lkxiaolou:
@wu-sheng I'm pushed the code  using an AtomicIntegerArray with 31 continuous integer instead of AtomicInteger to solve the false-sharing problem and it didn't use sun.unsafe directly.
So I think it's a better solution, what's your opinion?

asfgit:
FAILURE

asfgit:
https://builds.apache.org/job/skywalking-e2e/7/

asfgit:
https://builds.apache.org/job/skywalking-e2e/3/

asfgit:
https://builds.apache.org/job/skywalking-ci-it/58/

wu-sheng:
Add an English blog ref about false-sharing, https://dzone.com/articles/false-sharing

apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/common/AtomicRangeInteger.java
```diff
@@ -20,28 +20,32 @@
...
+    private AtomicIntegerArray values;
+
+    private static final int VALUE_OFFSET = 15;
+
     private int startValue;
     private int endValue;
 
     public AtomicRangeInteger(int startValue, int maxValue) {
-        this.value = new AtomicInteger(startValue);
+        this.values = new AtomicIntegerArray(31);
```
wu-sheng: Cache line should be 64 bytes in 64-bits OS. Then as `int[] array`.
1. 12 bytes(in OS 64-bits) for the array ref
1. 4 bytes(int) * 13(size)

Am I missing anything? Why set 31? And why set `VALUE_OFFSET` to 15? Compatible for 32-bit OS？

kezhenxu94: > * 12 bytes(in OS 64-bits) for the array ref

The memory address of the array reference isn't necessarily next to the 31 `int`s, the reference can be allocated at another memory block far away from the 31 `int`s, so it's unnecessary to take it into consideration.

> 4 bytes(int) * 13(size)

I think you had missed the same thing as I had, see https://github.com/apache/skywalking/pull/2930#discussion_r296446602

consider the cache line as a sliding window (of size 64 in 64-bits OS and size 32 in 32-bits OS), we must ensure that no matter where the window slides to, there must be only one `value` that will change in the window, and the remainings are `placeholders` that are never modified, to ensure that the cache line only contains our desired `value` that is changing.

```text
              15 * 4 = 60
                                                                                              15 * 4 = 60
+-------------------------------------------------+                    +----------------------------------------------------------+
|                                                 |     (4 bytes)      |                                                          |
+-------------------------------------------------+       value        +----------------------------------------------------------+



 +--------------------------------------------------------------------+
 |                                                                    |
 +--------------------------------------------------------------------+
                         cache line(sliding window) 64 

                                                +---------------------------->
                                                               slide


```

wu-sheng:
@lkxiaolou As 6.2.0 should release soon, could we wait and put this in 6.3.0 just for safe? Right now, I think this should be good.

lkxiaolou:
> @lkxiaolou As 6.2.0 should release soon, could we wait and put this in 6.3.0 just for safe? Right now, I think this should be good.

okay

PR #2931, opened 2019-06-21
Fix OOM by empty stack of exception.

wu-sheng:
First, I provide more readable codes for #2926 fix.
Then, as people are asking for bootstrap class enhancing, so I provided this as an optional mechanism. I am still hesitating whether we should accept plugin for bootstrap class, such as `java.net.URLConnection`

@ascrutae @candyleer We should test by using this to enhance some class in `rt.jar`. Please verify.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/ThrowableTransformer.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/ThrowableTransformer.java
@@ -37,7 +37,7 @@ public enum ThrowableTransformer {
         while (causeException != null || depth == 0) {
             stackMessage.append(printExceptionInfo(causeException));
 
-            boolean isLookDeeper = printStackElement(throwable.getStackTrace(), new AppendListener() {
+            boolean isLookDeeper = printStackElement(causeException.getStackTrace(), new AppendListener() {
                 public void append(String value) {
                     stackMessage.append(value);
                 }
@@ -51,7 +51,7 @@ public enum ThrowableTransformer {
                 break;
             }
 
-            causeException = throwable.getCause();
+            causeException = causeException.getCause();
             depth--;
         }
 
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/ThrowableTransformer.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/ThrowableTransformer.java
@@ -34,7 +34,7 @@ public enum ThrowableTransformer {
         Throwable causeException = throwable;
 
         int depth = CAUSE_EXCEPTION_DEPTH;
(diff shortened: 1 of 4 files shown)


wu-sheng:
@IanCao @zhaoyuguang @JaredTan95 I hope you have time to verify.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/ThrowableTransformer.java
```diff
@@ -30,10 +32,12 @@
...
         final StringBuilder stackMessage = new StringBuilder();
         Throwable causeException = throwable;
-        while (causeException != null) {
+
+        int depth = CAUSE_EXCEPTION_DEPTH;
+        while (causeException != null || depth == 0) {
             stackMessage.append(printExceptionInfo(causeException));
 
-            boolean overMaxLength = printStackElement(throwable.getStackTrace(), new AppendListener() {
+            boolean isLookDeeper = printStackElement(causeException.getStackTrace(), new AppendListener() {
```
wu-sheng: @Arugal @SummerOfServenteen I think there are two places originally wrong, which also trigger #2926

Thank @Arugal to catch this.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/ThrowableTransformer.java
```diff
@@ -43,11 +47,12 @@ public boolean overMaxLength() {
                 }
             });
 
-            if (overMaxLength) {
+            if (isLookDeeper) {
                 break;
             }
 
-            causeException = throwable.getCause();
+            causeException = causeException.getCause();
```
wu-sheng: The other one triggers #2926

wu-sheng: FYI @ascrutae This is the root cause and being here for a long time. `OverSize` check avoids OOM in most cases only.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/ThrowableTransformer.java
SummerOfServenteen: maybe depth != 0

SummerOfServenteen: maybe ( causeException != null  && depth != 0)

wu-sheng: Yes. Done.

wu-sheng:
Bootstrap loader is more complex than this. I will remove it from this PR.

Issue #2933, opened 2019-06-23
OAP server catches the index no found exception when the time at 0:00.


peng-yongsheng:
In this edition, we offered a new important feature that makes the Elasticsearch storage implementation has time series ability. It will automatically create a new index when the date turns to tomorrow. But, the logic had something wrong. First of all, the OAP server found the metrics in the index by its ID. But, the next day index is not created. So, we can find that there are some exceptions about index not found in the Elasticsearch server's log.

PR #2934, opened 2019-06-23
Use ids query method instead of get method for metrics persistence.

peng-yongsheng:
**Performance test report:**

4 million rows in the index, those two methods executed 200 thousand times each.
- The total duration of the get method:  42737 million seconds.
- The total duration of the ids query method: 103436 million seconds.

**Conclusion**
Although the ids query method is slower than the get method, but it's not too much. So, I think it's acceptable.

#2933


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/MetricsEsDAO.java
```diff
@@ -42,9 +42,9 @@
 
     @Override public Metrics get(Model model, Metrics metrics) throws IOException {
         String modelName = TimeSeriesUtils.timeSeries(model, metrics.getTimeBucket());
```
wu-sheng: This looks like still reading from the physical index, I think we should read from alias name, right?

asfgit:
FAILURE
 No test results found.

asfgit:
https://builds.apache.org/job/skywalking-e2e/2/

PR #2961, opened 2019-06-27
Update UI, revert #2952 and fix JVM CPU usage again

wu-sheng:


wu-sheng:
I prefer to wait for @TinyAllen to fix other p99 graph bugs.

PR #2962, opened 2019-06-28
Allow alarm rules and webhooks to be dynamically configured

### LINKED ISSUE #2843: Can alarm rule be dynamically modified?

- [x] Question or discussion
- [ ] Bug
- [ ] Requirement

___
### Question
- What do you want to know?
Can alarm rule be dynamically modified?

kezhenxu94:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

fix #2843


oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/AlarmModuleProvider.java
```diff
@@ -49,19 +53,26 @@
         }
         RulesReader reader = new RulesReader(applicationReader);
         Rules rules = reader.readRules();
+        alarmRulesWatcher = new AlarmRulesWatcher(rules, this, this::reload);
+    }
+
+    private void reload(final Rules rules) {
```
wu-sheng: This `reload` is a hijack to existing code. `ModuleProvider#registerServiceImplementation` didn't expect runtime change in the design. So you could see `Map<Class<? extends Service>, Service> services` for service register, which is not thread-safe.

We need to reconsider this, even whether do we need to approve the service hot-switch

wu-sheng:
Close for now. Open it when you are ready

PR #2963, opened 2019-06-28
Add more logs for tracking elasticsearch validation error, e.g. [Data…

yijiuping:
…Carrier.REGISTER_L2.BulkConsumePool.0.Thread] ERROR [] - Validation Failed: 1: id is too long, must be no longer than 512 bytes but was: 685;

We have got a lot of error logs like: org.elasticsearch.action.ActionRequestValidationException: Validation Failed: 1: id is too long, must be no longer than 512 bytes but was: 685;

Without the detailed log, we cannot do improvement for the endpoint names or find out whether SkyWalking has a bug or not.


yijiuping:
The validation exception was not catch by this pull request.

PR #2973, opened 2019-07-01
Support etcd configuration.

wayilau:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

Support etcd configuration.


oap-server/server-configuration/configuration-etcd/pom.xml
```diff
@@ -39,10 +39,76 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.apache.skywalking</groupId>
-            <artifactId>cluster-etcd-plugin</artifactId>
-            <version>${project.version}</version>
+            <groupId>org.projectlombok</groupId>
```
wu-sheng: etcd4j is existing in dependency management. You should not add it here.

wu-sheng:
Revert by #2974

oap-server/server-configuration/configuration-etcd/pom.xml
wayilau: ok

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdServerSettings.java
```diff
@@ -0,0 +1,44 @@
...
+
+package org.apache.skywalking.oap.server.configuration.etcd;
+
+import lombok.Getter;
+import org.apache.skywalking.oap.server.library.module.ModuleConfig;
+
+/**
+ * @author Alan Lau
+ */
+public class EtcdServerSettings extends ModuleConfig {
```
kezhenxu94:
```suggestion
@ToString
public class EtcdServerSettings extends ModuleConfig {
```

since you're using lombok `@Getter`, `@ToString` should free you from writing boilerplate code (`toString` below)

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdUtils.java
```diff
@@ -0,0 +1,74 @@
...
+    public EtcdUtils() {
+    }
+
+    public static List<URI> parse(EtcdServerSettings settings) throws ModuleStartException {
+        List<URI> uris = new ArrayList<>();
+        try {
+            logger.info("etcd settings is {}", settings);
+            List<Address> addressList = ConnectUtils.parse(settings.getServerAddr());
+            for (Address address : addressList) {
+                uris.add(URI.create(new StringBuilder("http://").append(address.getHost()).append(":").append(address.getPort()).toString()));
```
kezhenxu94: string concatenation is easy-prone, I'll suggest using `java.net.URI#URI(java.lang.String, java.lang.String, java.lang.String, int, java.lang.String, java.lang.String, java.lang.String)`.

kezhenxu94: not a `etcd` expert, perhaps you need to deal with the [authorization](https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/authentication.md), I'm not sure

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdUtils.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+
+    public static List<URI> parseProp(Properties properties) throws ModuleStartException {
+        List<URI> uris = new ArrayList<>();
+        try {
+            logger.info("etcd server addr is {}", properties);
+            List<Address> addressList = ConnectUtils.parse(properties.getProperty("serverAddr"));
+            for (Address address : addressList) {
+                uris.add(URI.create(new StringBuilder("http://").append(address.getHost()).append(":").append(address.getPort()).toString()));
```
kezhenxu94: same here

oap-server/server-configuration/configuration-etcd/src/test/java/org/apache/skywalking/oap/server/configuration/etcd/ITEtcdConfigurationTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+        provider =
+            (EtcdConfigurationTestProvider)moduleManager
+                .find(EtcdConfigurationTestModule.NAME)
+                .provider();
+
+        assertNotNull(provider);
+    }
+
+    @Test(timeout = 2000000)
```
kezhenxu94: `2000000 ms ~= 33 minutes`, does it take so long in reality? if not I'll suggest choosing a reasonable threshold, otherwise the build job will hang for a long time

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdServerSettings.java
wayilau: Thanks for you review. I got it.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdUtils.java
wayilau: while open the tls, may be needed authorization.

oap-server/server-configuration/configuration-etcd/src/test/java/org/apache/skywalking/oap/server/configuration/etcd/ITEtcdConfigurationTest.java
wayilau: yes, i forget to change this. i am working on this.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
```diff
@@ -0,0 +1,103 @@
...
+    @Override public ConfigTable readConfig(Set<String> keys) {
+
+        if (client == null) {
+            try {
+                client = new EtcdClient(EtcdUtils.parse(settings).toArray(new URI[] {}));
+            } catch (ModuleStartException e) {
+                logger.error(e.getMessage(), e);
+            }
+
+            EtcdResponsePromise promise;
```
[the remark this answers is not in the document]
wayilau: ok, i will add the type.

oap-server/server-configuration/configuration-etcd/pom.xml
wayilau: here, i should add the dependency, or it will compile failed.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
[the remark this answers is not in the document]
wayilau: done

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdUtils.java
wayilau: I have modified it.

wayilau:
it has conflicts. @wu-sheng @kezhenxu94 , i have no rights to merge.

oap-server/server-configuration/configuration-etcd/pom.xml
kezhenxu94: @wayilau he means the `projectlombok`, it's included in `oap-server/pom.xml` already am I right @wu-sheng

kezhenxu94:
@wayilau and please resolve the conflicts

wayilau:
> @wayilau and please resolve the conflicts
ok, let me check.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
```diff
@@ -0,0 +1,103 @@
...
+
+        final ConfigTable table = new ConfigTable();
+        keys.forEach(registryKey -> {
+            String key = new StringBuilder("/").append(settings.getGroup()).append("/").append(registryKey).toString();
+            try {
+                EtcdResponsePromise<EtcdKeysResponse> promise = client.get(key).send();
+                EtcdKeysResponse response = promise.get();
+                table.add(new ConfigTable.ConfigItem(getRealKey(key, settings.getGroup()), response.getNode().getValue()));
+            } catch (EtcdException e) {
+                if (e.getErrorCode() == 100) {
```
kezhenxu94: don't use magic number :) use `mousio.etcd4j.responses.EtcdErrorCode#KeyNotFound` instead

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private EtcdServerSettings settings;
+
+    private EtcdClient client;
+
+    @Override public ConfigTable readConfig(Set<String> keys) {
+
+        if (client == null) {
+            try {
+                client = new EtcdClient(EtcdUtils.parse(settings).toArray(new URI[] {}));
```
kezhenxu94: may need some synchronization and double check here? `readConfig` is invoked in a scheduled executor at fixed rate (`org.apache.skywalking.oap.server.configuration.api.ConfigWatcherRegister#start`), in an extreme situation, there may be two or more threads that run into this piece of code simultaneously. I'm not sure whether it's worthy to lazy initialize the `EtcdClient` or not.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
wayilau: ok

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
wayilau: I saw only one thread do the sync operation. @kezhenxu94  so i didn't add the sync. am i wrong ? if so, i will modify this.

wayilau:
the e2e tests failed. @kezhenxu94 @wu-sheng

oap-server/server-configuration/configuration-etcd/pom.xml
kezhenxu94: @wayilau can you try to remove the `projectlombok` dependency? I think it's included in the `oap-server/pom.xml`, other codes look good to me

oap-server/server-configuration/configuration-etcd/pom.xml
```diff
@@ -0,0 +1,214 @@
...
+        <dependency>
+            <groupId>com.fasterxml.jackson.module</groupId>
+            <artifactId>jackson-module-afterburner</artifactId>
+            <version>2.9.5</version>
+        </dependency>
+        <dependency>
+            <groupId>org.yaml</groupId>
+            <artifactId>snakeyaml</artifactId>
+            <version>1.18</version>
+        </dependency>
```
wu-sheng: For all dependencies, we have etcd cluster management implementation, we should have all the dependencies we need, why we add more here? What is the difference?

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
```diff
@@ -0,0 +1,104 @@
...
+                });
+
+            } catch (IOException e) {
+                logger.error(e.getMessage(), e);
+            }
+        }
+
+        final ConfigTable table = new ConfigTable();
+        keys.forEach(registryKey -> {
+            String key = new StringBuilder("/").append(settings.getGroup()).append("/").append(registryKey).toString();
```
wu-sheng: Use `+` instead of `#append`. There is compiler optimization there, no need to use this.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
```diff
@@ -0,0 +1,104 @@
...
+        });
+
+        return table;
+    }
+
+    public EtcdConfigWatcherRegister(EtcdServerSettings settings) {
+        this.settings = settings;
+    }
+
+    private void onDataValueChanged() {
```
wu-sheng: What is this empty method for?

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigurationProvider.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.skywalking.oap.server.library.module.ModuleStartException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author Alan Lau
+ */
+public class EtcdConfigurationProvider extends AbstractConfigurationProvider {
+
+    private final static Logger LOG = LoggerFactory.getLogger(EtcdConfigurationProvider.class);
```
wu-sheng: LOG -> logger

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdServerSettings.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+@ToString
+@Getter
+@Setter
+public class EtcdServerSettings extends ModuleConfig {
+
+    private String clusterName = "default";
+    private String serverAddr;
+    private String group;
+    private int period = 5;
```
wu-sheng: default is 5, unit? If unit=second, I think it is too short for timeout, try 30 for configuration center, because in most case, people don't treat performance and latency very seriously for it.

wu-sheng: For the config
```yaml
#    period : 60 # Unit seconds, sync period. Default fetch every 60 seconds.
```

Why don't use 60 in codes?

oap-server/server-configuration/configuration-etcd/pom.xml
wu-sheng: FYI @kezhenxu94 This is the most important thing, we should treat dependency very serious, accept reasonable and necessary new dependencies.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdServerSettings.java
wayilau: ok, i will fix this.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
wayilau: Actually, it is unused now. Maybe use to monitor the value changed some time for some thing. But i didn't impl that at this time.

wu-sheng: This is a private method. So you write it by yourself. No one asks for this.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
```diff
@@ -0,0 +1,104 @@
...
+            } catch (ModuleStartException e) {
+                logger.error(e.getMessage(), e);
+            }
+
+            EtcdResponsePromise<EtcdKeysResponse> promise;
+            try {
+                promise = client.get(settings.getClusterName()).waitForChange().send();
+                promise.addListener(responsePromise -> {
+                    onDataValueChanged();
+                });
```
wu-sheng: This looks like asking for change every time, is this a good way in etcd? Or is there a change to register and listening for change notification always.

wayilau: Yes, it is a problem for everytime make a waitForChange. I will fix this.

wayilau: As we have executor to sync config every 60 sec, so i need not add a monitor for value changed. I just remove the waitForChange logic. what's your idea ? @wu-sheng 
Another way i add a map to store the waitForChange for every key, everytime i check the key is exists, remove which when the change occurs.  but i think is not needed.  also the logic is to complex. I also can use the timeout, but here, I think it is needed.

oap-server/server-configuration/configuration-etcd/pom.xml
wayilau: > For all dependencies, we have etcd cluster management implementation, we should have all the dependencies we need, why we add more here? What is the difference?

You mean i import the cluster-etcd-impl here for compile ? am i right ?

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigurationProvider.java
wayilau: fixed

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
wayilau: removed.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
wayilau: fixed

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
wu-sheng: I am not an etcd user, I can't tell much detail. My concern and question is only, do query for etcd every 60 sec, is it a high payload when oap is a 20-30 nodes cluster.

wu-sheng: In zk implementation, I think it set up a listener. Yes? @kezhenxu94 
Anyway, you could make the decision, and we could wait for user feedback.

oap-server/server-configuration/configuration-etcd/pom.xml
wu-sheng: I mean, etcd dependency version is already there. Take a look at etcd cluster manager module

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
wu-sheng: Where is the timeout for etcd server access?
Current codes, with more configuration items, etcd server will pay more payload to response all items every time. 
Take a look at @kezhenxu94 's pr about set alarm rule dynamicly. That PR is ready to be merged, but it tells you, etcd server needs to response while alarm rule everytime. I am not sure it is the right way.

kezhenxu94: > In zk implementation, I think it set up a listener. Yes? @kezhenxu94
> Anyway, you could make the decision, and we could wait for user feedback.

It uses `PathChildrenCache`, whose JavaDoc says that it **watch the ZK path**, so technically **YES**, so do the Apollo and Nacos implementations, we use watcher/listener whenever possible.

> My concern and question is only, do query for etcd every 60 sec, is it a high payload when oap is a 20-30 nodes cluster

That's really a concern, even when there're just one or two oap nodes, but the configuration items/contents are large (although we have only several config items, but I believe it definitely grows in the future). So @wayilau please, take this into consideration, and yet, you will make the decision as @wu-sheng said

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdServerSettings.java
kezhenxu94: @wayilau I saw that you changed the default value of `EtcdServerSettings#period` to 30, but in the `application.yml`, it's still:

```yml
#  etcd:
#    period : 60 # Unit seconds, sync period. Default fetch every 60 seconds.
```

it's quite misleading, please keep consistent

oap-server/server-configuration/configuration-etcd/pom.xml
wayilau: ok

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdServerSettings.java
wayilau: ok

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
wayilau: ok, i will find a better way to use waitForChange to achieve this.

wayilau:
@kezhenxu94 @wu-sheng 
I have update the listener for etcd configuration.

oap-server/server-configuration/configuration-etcd/pom.xml
wu-sheng: You misunderstand my suggestions, You should pop etcd dependency up to oap-server/pom.xml/

oap-server/server-configuration/configuration-etcd/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.skywalking</groupId>
+            <artifactId>configuration-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.skywalking</groupId>
+            <artifactId>library-client</artifactId>
```
wu-sheng: What is this for?

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
```diff
@@ -0,0 +1,148 @@
...
+import mousio.etcd4j.responses.EtcdKeysResponse;
+import org.apache.skywalking.oap.server.configuration.api.ConfigTable;
+import org.apache.skywalking.oap.server.configuration.api.ConfigWatcherRegister;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author Alan Lau
+ */
+public class EtcdConfigWatcherRegister extends ConfigWatcherRegister {
```
wu-sheng: Add comments to class, fields and methods.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdServerSettings.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+@ToString
+@Getter
+@Setter
+public class EtcdServerSettings extends ModuleConfig {
+
+    private String clusterName = "default";
+    private String serverAddr;
+    private String group;
+    private int period = 60;
```
wu-sheng: Comments are required here. Such as what is group? What is period unit?

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdUtils.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.Properties;
+import org.apache.skywalking.oap.server.library.util.Address;
+import org.apache.skywalking.oap.server.library.util.ConnectUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author Alan Lau
+ */
+public class EtcdUtils {
```
wu-sheng: Comments are required here.

wu-sheng:
@kezhenxu94 @JaredTan95 Please review the codes.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
```diff
@@ -0,0 +1,148 @@
...
+        });
+    }
+
+    private void onDataValueChanged(ResponsePromise<EtcdKeysResponse> promise) {
+        try {
+            EtcdKeysResponse.EtcdNode node = promise.get().getNode();
+            String dataId = getRealKey(node.getKey(), settings.getGroup());
+            String value = node.getValue();
+            if (logger.isInfoEnabled()) {
+                logger.info("Nacos config changed: {}: {}", dataId, node.getValue());
```
kezhenxu94: Not "Nacos" :)

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
```diff
@@ -0,0 +1,148 @@
...
+            }
+
+            configItemKeyedByName.put(dataId, Optional.ofNullable(value));
+        } catch (Exception e) {
+            if (e instanceof EtcdException) {
+                if (EtcdErrorCode.KeyNotFound == ((EtcdException)e).errorCode) {
+                    return;
+                }
+            }
+            throw new EtcdConfigException("wait for value chanaged fail", e);
```
kezhenxu94: typo: **chanaged**

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
```diff
@@ -0,0 +1,148 @@
...
+            String value = node.getValue();
+            if (logger.isInfoEnabled()) {
+                logger.info("Nacos config changed: {}: {}", dataId, node.getValue());
+            }
+
+            configItemKeyedByName.put(dataId, Optional.ofNullable(value));
+        } catch (Exception e) {
+            if (e instanceof EtcdException) {
+                if (EtcdErrorCode.KeyNotFound == ((EtcdException)e).errorCode) {
+                    return;
```
kezhenxu94: simply ignore `EtcdErrorCode.KeyNotFound`? do we need something like `configItemKeyedByName.put(dataId, Optional.empty());`

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
wayilau: yes,  i will change.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
wayilau: ok

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
wayilau: As exception is so much that it doesn't say that we need put null to the table. we do this only we found that no value found. This is my idea. what do you think ? @kezhenxu94

kezhenxu94: > As exception is so much that it doesn't say that we need put null to the table. we do this only we found that no value found. This is my idea. what do you think ? @kezhenxu94

I'm not sure, I was thinking that etcd would throw such exception (`KeyNotFound`) when we are getting a deleted key, the exception name `KeyNotFound` made me think that, it's just my personal thought and may be wrong, please ignore if you had confirmed, thanks :)

kezhenxu94: @wayilau Ping me if you are sure and I'll mark this as resolved :)

oap-server/server-configuration/configuration-etcd/pom.xml
wayilau: I see.

oap-server/pom.xml
```diff
@@ -83,7 +83,7 @@
         <curator-test.version>2.12.0</curator-test.version>
         <etcd4j.version>2.17.0</etcd4j.version>
         <etcd.version>v3.2.3</etcd.version>
-        <netty.version>4.0.27.Final</netty.version>
+        <netty.version>4.1.27.Final</netty.version>
```
wu-sheng: Why are you changing netty version? Something incompatible?

wayilau: Because etcd4j use 4.1.32 but grpc use 4.1.27. so i change the version, 
Keep them consistent. then need not add other netty jar to dist.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
wayilau: ok

oap-server/server-configuration/configuration-etcd/pom.xml
wayilau: deleted.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdServerSettings.java
wayilau: yes, i do.

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
wayilau: @kezhenxu94 You are right, we need set empty for this when key not found. It is say that key is not exists for the first time or be deleted in the later. For other exception, we just throw, do not change the table.

wayilau: @kezhenxu94 You are right, we shoud add empty when key was first time created or when it is deleted in later time. I have changed this.

wayilau:
@kezhenxu94 It's ready to review now.

oap-server/pom.xml
```diff
@@ -378,11 +382,75 @@
...
+            <dependency>
+                <groupId>io.netty</groupId>
+                <artifactId>netty-codec-http</artifactId>
+                <version>${netty.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>com.fasterxml.jackson.module</groupId>
+                <artifactId>jackson-module-afterburner</artifactId>
+                <version>${jackson-module-afterburner.version}</version>
```
wu-sheng: Where are the licenses of these? I think you haven't updated the dist LICENSE file.

wayilau: @wu-sheng  I have added .

wu-sheng:
I am rerunning CI and e2e. @kezhenxu94 @JaredTan95 I hope you two have time to review and give the result.

oap-server/server-configuration/configuration-etcd/pom.xml
```diff
@@ -0,0 +1,178 @@
...
+                                <id>add-default-properties</id>
+                                <phase>initialize</phase>
+                                <goals>
+                                    <goal>execute</goal>
+                                </goals>
+                                <configuration>
+                                    <providerSelection>2.0</providerSelection>
+                                    <source>
+                                        project.properties.setProperty('etcd.host', 'localhost')
+                                        project.properties.setProperty('etcd.port', '2379')
```
kezhenxu94: don't set `etc.port` here statically, which would cause conflict when multiple build jobs are running on the same Jenkins node, the port is dynamically determined by `docker-maven-plugin` (line 100), and is available as a system property(line 156-163)

oap-server/server-configuration/configuration-etcd/src/test/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegisterTest.java
```diff
@@ -0,0 +1,131 @@
...
+        when(mockSettings.getGroup()).thenReturn(group);
+        mockStatic(EtcdUtils.class);
+
+        List<URI> uris = mock(List.class);
+        when(EtcdUtils.parse(any())).thenReturn(uris);
+
+        final EtcdClient client = PowerMockito.mock(EtcdClient.class);
+        whenNew(EtcdClient.class).withAnyArguments().thenReturn(client);
+
+        URI uri = new URI("http://localhost:2379");
```
kezhenxu94: And you should get the port from system properties, which is dynamically set by `docker-maven-plugin`, as for the host, it doesn't matter

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
```diff
@@ -0,0 +1,162 @@
...
+                    configItemKeyedByName.put(key, Optional.empty());
+                    return;
+                }
+            }
+            throw new EtcdConfigException("wait for value changed fail", e);
+        }
+    }
+
+    /**
+     * get real key in etcd cluster which is removed "/${group}" from the key retrived from etcd.
```
kezhenxu94: typo: retrived

oap-server/server-configuration/configuration-etcd/src/test/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegisterTest.java
wayilau: ok

oap-server/server-configuration/configuration-etcd/src/main/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegister.java
wayilau: Fixed

oap-server/server-configuration/configuration-etcd/pom.xml
wayilau: yes.

PR #2987, opened 2019-07-02
Adding deadline to gRPC client

### LINKED ISSUE #2972: Add deadline to all gRPC including stream and block stubs
There are following gRPC communications
1. Agent register/JVM/segment report
1. OAP internal communication
1. OAP Exporter gRPC client
1. OAP Dynamic Configuration Service gRPC client.

hanahmily:
Adding deadline to gRPC client

 - Set up 10 seconds deadline after gRPC client sending
 - The duration of deadline contains three segments: connecting,
   request and response

For blocking stub, I just set the deadline before invoking service. For
bi-streaming stub, I found all of them are used as a blocking style,
that after getting streaming response client stub just close current
streaming. Base on the above reality, I pick the same way as blocking one.

Fixed #2972


hanahmily:
I take a look at your original PR which is different from this one.

The correct approach is to set deadline just before invoking service
method. If we request steaming 10 times, 10 deadlines should be set to each
request.

Your approach was to set deadline at stub initial that is the wrong way of
using deadline

PR #2992, opened 2019-07-03
Optimising docker image size.

JaredTan95:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

In my opinion, I think the final image is too big.
If there are no special requirements for the docker image, we can optimize it in some way.

Before:
```bash
skywalking/ui                               latest      733MB
skywalking/oap                              latest     834MB
```

After:
```bash
skywalking/ui                               latest     83.4MB
skywalking/oap                              latest    184MB
```


wu-sheng:
What is the difference between two basic images?

wu-sheng:
@TinyAllen I think you should consider adding an e2e stage(parallel) to check the image we built in dev stage.

wu-sheng:
@JaredTan95 I think you should consider adding an e2e stage(parallel) to check the image we built in dev stage.

JaredTan95:
Closed, temporarily. Others need to be done.

hanahmily:
Seems like there’s no `bash` in it. entrypoint script can not be executed.

PR #3004, opened 2019-07-04
Improve RegisterPersistentWorker by bulk persistence to reduce cluster blocking

lt964709769:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance
___
We found that when skywalking block in RegisterPersistentWorker, the grpc threads will be blocked, and then the entire cluster will not accept requests. 

Here are some fixes:
1. Added LRU cache to reduce query quantity
2. Update and insert are aggregated and updated with bulk


wu-sheng:
Could you indicate what is the difference between your LRU and existing Map.
Such as following, this should have guaranteed there is no duplicated.
```java
    private void onWork(RegisterSource registerSource) {
        if (!sources.containsKey(registerSource)) {
            sources.put(registerSource, registerSource);
        } else {
            sources.get(registerSource).combine(registerSource);
        }
```

lt964709769:
> Could you indicate what is the difference between your LRU and existing Map.
> Such as following, this should have guaranteed there is no duplicated.
> 
> ```java
>     private void onWork(RegisterSource registerSource) {
>         if (!sources.containsKey(registerSource)) {
>             sources.put(registerSource, registerSource);
>         } else {
>             sources.get(registerSource).combine(registerSource);
>         }
> ```

This Map is used to combine the registration information of this batch. 
In this improvement, LRU can be access across batches. It is used to cache the registration information to reduce access to data sources.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/register/worker/RegisterPersistentWorker.java
```diff
@@ -80,43 +92,80 @@ private void onWork(RegisterSource registerSource) {
...
-
         if (sources.size() > 1000 || registerSource.getEndOfBatchContext().isEndOfBatch()) {
             sources.values().forEach(source -> {
+                RegisterSource cacheSource = null;
                 try {
-                    RegisterSource dbSource = registerDAO.get(modelName, source.id());
+                    RegisterSource dbSource = this.getRegisterSourceCache(source.id());
                     if (Objects.nonNull(dbSource)) {
                         if (dbSource.combine(source)) {
-                            registerDAO.forceUpdate(modelName, dbSource);
```
dmsolr: forceUpdate means update immediately. I am not sure that it can be change to batch. Because the data will be delay to visible to others.

wu-sheng:
@lt964709769  LRU could cause dirty read. Because the oap node doing register could change because cluster size change in some case.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/register/worker/RegisterPersistentWorker.java
lt964709769: I've used the IMMEDIATE RefreshPolicy in my code, and you can find it in ElasticSearchClient.

dmsolr: I means cache could cause dirty read in cluster mode. As @wu-sheng say.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/register/worker/RegisterPersistentWorker.java
```diff
@@ -93,78 +91,49 @@ private void onWork(RegisterSource registerSource) {
...
-                        if ((sequence = registerLockDAO.getId(scopeId, source)) != Const.NONE) {
-                            try {
-                                dbSource = this.getRegisterSourceCache(source.id());
-                                if (Objects.nonNull(dbSource)) {
-                                    if (dbSource.combine(source)) {
-                                        batchPersistenList.add(registerDAO.prepareBatchUpdate(modelName, dbSource));
-                                        cacheSource = dbSource;
+            try {
+                List<String> ids = sources.values().stream().map(value -> value.id()).collect(Collectors.toList());
+                final Map<String, RegisterSource> map = registerDAO.batchGet(modelName, ids.toArray(new String[0]));
```
dmsolr: We don't need to convert array, right?

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/h2/dao/H2RegisterDAO.java
```diff
@@ -47,6 +49,18 @@ public H2RegisterDAO(JDBCHikariCPClient h2Client, StorageBuilder<RegisterSource>
         return (RegisterSource)getByID(h2Client, modelName, id, storageBuilder);
     }
 
+    @Override
+    public Map<String, RegisterSource> batchGet(String modelName, String... ids) throws IOException {
+        Map<String, RegisterSource> map = new HashMap<>(ids.length);
+        for (String id : ids) {
+            RegisterSource registerSource = this.get(modelName, id);
```
dmsolr: We can do real batch queries as ES do here.

wu-sheng: @peng-yongsheng what do you think?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/register/worker/RegisterPersistentWorker.java
lt964709769: This problem has now been modified.

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/h2/dao/H2RegisterDAO.java
lt964709769: Now, this has been modified to be a real batch query.

wu-sheng:
@peng-yongsheng I think you need to review this. RegisterDAO interface even is being changed in this PR. I think we need to be careful about this change.

peng-yongsheng:
@lt964709769 Please put the test report before and after modification in this PR. Is this the real problem? How many inventories need to register in 20ms or end of the datacarrier batch?

```
sources.size() > 1000 || registerSource.getEndOfBatchContext().isEndOfBatch();
```

peng-yongsheng:
@lt964709769  If you say, the heartbeat of the inventory update causes the gRPC block, I think its the real problem. Inventory register just occurs at the moment of the new application started and all the applications keep beating all the time.

lt964709769:
@peng-yongsheng Yes,The heartbeat is the main problem. Currently, there are 96 updates of service_inventory, 399 updates of service_instance_inventory, 93 updates of endpoint_inventory, and 17 updates of network_address_inventory every 3 seconds in our production environment. The 3 seconds param can be change to 1 min may alleviate the problem,but our system may scale to 10 times larger. This is why we do the optimization.

wu-sheng:
> Yes,The heartbeat is the main problem. Currently, there are 96 updates of service_inventory, 399 updates of service_instance_inventory, 93 updates of endpoint_inventory, and 17 updates of network_address_inventory every 3 seconds in our production environment. The 3 seconds param can be change to 1 min may alleviate the problem,but our system may scale to 10 times larger. This is why we do the optimization.
Before the use of bulk, the online registration service was often blocked, which eventually led to the collapse of the whole cluster. After the use of bulk, the registration service ran stably.

The heartbeats of 400 instances look like not large scale, but it depends on the ElasticSearch performance.

peng-yongsheng:
> @peng-yongsheng Yes,The heartbeat is the main problem. Currently, there are 96 updates of service_inventory, 399 updates of service_instance_inventory, 93 updates of endpoint_inventory, and 17 updates of network_address_inventory every 3 seconds in our production environment. The 3 seconds param can be change to 1 min may alleviate the problem,but our system may scale to 10 times larger. This is why we do the optimization.
> Before the use of bulk, the online registration service was often blocked, which eventually led to the collapse of the whole cluster. After the use of bulk, the registration service ran stably.

Extending the interval between heartbeats is the best solution. The register is a core competence, I am not very sure about the bulk. Maybe it is no problem, but I don't think we have the necessary to risk fixing this because it's not the main problem.

lt964709769:
All right. What is the role of the heartbeat in skywalking?

wu-sheng:
> All right. What is the role of the heartbeat in skywalking?

Keep services shown in the UI.

PR #3016, opened 2019-07-06
Add e2e cluster tests

kezhenxu94:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

For quick feedback and discussion, chores will be done once settled,

failure is expected until the docker image is pushed to dockerhub


kezhenxu94:
)

test/e2e/e2e-cluster/test-runner/src/docker/clusterize.awk
```diff
@@ -0,0 +1,92 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#!/usr/bin/awk -f
+
+BEGIN {
```
wu-sheng: Please format this file.

test/e2e/e2e-single-service/src/test/java/org/apache/skywalking/e2e/SampleVerificationITCase.java
```diff
@@ -111,7 +100,7 @@ public void verify() throws Exception {
         );
         assertThat(responseEntity.getStatusCode()).isEqualTo(HttpStatus.OK);
 
-        Thread.sleep(5000);
+        Thread.sleep(20000);
```
wu-sheng: I think we could do a re-try here. Do it every 20s, but we could try 5 times, in case of unexpected failure. What do you think?

wu-sheng:
FYI @hanahmily @ascrutae @peng-yongsheng @JaredTan95

Issue #3014, opened 2019-07-05
org.apache.skywalking.oap.server.storage.plugin.elasticsearch.base.HistoryDeleteEsDAO 的 57行 retrievalIndexByAliases 方法已经把 namespace 加上了 72行的 deleteIndex 方法又加了一遍namespace


NaOH-nagao:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS and JRE?

- Which company or project?

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.

___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

NaOH-nagao:
org.apache.skywalking.oap.server.storage.plugin.elasticsearch.base.HistoryDeleteEsDAO 的 57行 retrievalIndexByAliases 方法已经把 namespace 加上了 72行的 deleteIndex 方法又加了一遍namespace
6.2.0 版本

wu-sheng:
We need issue and pull request in English. Could you change it?

Arugal:
Don't quite understand what the problem is
https://github.com/apache/skywalking/blob/b21db17148df12f34a13a2b1693e009e11e8d7a2/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/HistoryDeleteEsDAO.java#L49-L80

wu-sheng:
From my understanding, #retrievalIndexByAliases method return value, should not add namespace as prefix, because deleteIndex will add automatically.

Also, in this change, make sure TimeSeriesUtils.indexTimeSeries still work.

PR #3017, opened 2019-07-06
fix deleteIndex repeat append namespace

Arugal:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
#3014
- How to fix?
#formatIndexName optional
___


oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -146,7 +146,13 @@ public JsonObject getIndex(String indexName) throws IOException {
     }
 
     public boolean deleteIndex(String indexName) throws IOException {
-        indexName = formatIndexName(indexName);
+        return deleteIndex(indexName, true);
+    }
+
+    public boolean deleteIndex(String indexName, boolean formatIndexName) throws IOException {
```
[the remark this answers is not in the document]
Arugal: > I suggest you move the whole code block inside `ElasticSearchClient` as method `#deleteTimeSeriesData(modelName, timeBefore)`. Then the delete codes without `formatIndexName` could ba placed inside method directly.
> 
> ```java
> List<String> indexes = client.retrievalIndexByAliases(model.getName());
> 
>             List<String> prepareDeleteIndexes = new ArrayList<>();
>             for (String index : indexes) {
>                 long timeSeries = TimeSeriesUtils.indexTimeSeries(index);
>                 if (timeBefore >= timeSeries) {
>                     prepareDeleteIndexes.add(index);
>                 }
>             }
> 
>             if (indexes.size() == prepareDeleteIndexes.size()) {
>                 client.createIndex(TimeSeriesUtils.timeSeries(model));
>             }
> 
>             for (String prepareDeleteIndex : prepareDeleteIndexes) {
>                 client.deleteIndex(prepareDeleteIndex);
>             }
> ```
> 
> I think don't mixe these two kinds of OP(s) together should be much better.

This code block needs to use ```TimeSeriesUtils```, but ```TimeSeriesUtils``` is located in``` storage-elasticsearch-plugin```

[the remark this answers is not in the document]
Arugal: or delete the ```namespace``` in the ```#retrievalIndexByAliases```return result.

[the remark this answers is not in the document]
wu-sheng: Then build List<ESTimeSeriesIndex>, in ESIndex, keep namespace and name in two fields. Then add a new delete method accepting ESTimeSeriesIndex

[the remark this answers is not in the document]
Arugal: done

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -292,7 +300,14 @@ public int delete(String indexName, String timeBucketColumnName, long endTimeBuc
 
     public String formatIndexName(String indexName) {
         if (StringUtils.isNotEmpty(namespace)) {
-            return namespace + "_" + indexName;
+            return namespacePrefix + indexName;
+        }
+        return indexName;
+    }
+
+    public String undoFormatIndexName(String indexName) {
```
wu-sheng: This should be private.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -292,7 +300,14 @@ public int delete(String indexName, String timeBucketColumnName, long endTimeBuc
...
+        if (StringUtils.isNotEmpty(namespace) && indexName.startsWith(namespacePrefix)) {
```
wu-sheng: If `!startWith(namespace`, then you should raise an exception.

oap-server/server-library/library-client/src/test/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ITElasticSearchClient.java
```diff
@@ -183,4 +183,14 @@ public void bulk() throws InterruptedException {
         bulkProcessor.flush();
         bulkProcessor.awaitClose(2, TimeUnit.SECONDS);
     }
+
+    @Test
+    public void undoFormatIndexName() {
```
wu-sheng: This is an IT test, Integration Test. You should test `retrievalIndexByAliases` with real Elastic, create indexes with namespace, then test `#retrievalIndexByAliases` result

wu-sheng: Take an example with other tests in this class.

Arugal:
done

oap-server/server-library/library-client/src/test/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ITElasticSearchClient.java
```diff
@@ -51,7 +52,10 @@
     @Before
     public void before() throws IOException {
         final String esAddress = System.getProperty("elastic.search.address");
-        client = new ElasticSearchClient(esAddress, "", "test", "test");
+        final String esNamespace = System.getProperty("elastic.search.namespace", "");
+        final String esUser = System.getProperty("elastic.search.user", "test");
+        final String esPassword = System.getProperty("elastic.search.password", "test");
```
wu-sheng: One question, who will set these properties?

wu-sheng: If not, then,  namespace feature hasn't been tested.

oap-server/server-library/library-client/src/test/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ITElasticSearchClient.java
```diff
@@ -183,4 +187,51 @@ public void bulk() throws InterruptedException {
...
         bulkProcessor.awaitClose(2, TimeUnit.SECONDS);
     }
+
+    @Test
+    public void timeSeriesOperate() throws IOException {
+        String indexName = "test_time_series_operate";
+        String timeSeriesIndexName = indexName + "-2019";
+
+        try {
+            if (client.isExistsTemplate(indexName)) {
```
wu-sheng: Why these is an chance existing? If other case doesn't remove, set a finally there.

oap-server/server-library/library-client/src/test/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ITElasticSearchClient.java
```diff
@@ -183,4 +187,51 @@ public void bulk() throws InterruptedException {
...
+
+            XContentBuilder builder = XContentFactory.jsonBuilder().startObject()
+                .field("name", "pengys")
+                .endObject();
+            client.forceInsert(timeSeriesIndexName, "testid", builder);
+
+            List<ElasticSearchTimeSeriesIndex> indexes = client.retrievalIndexByAliases(indexName);
+            Assert.assertEquals(1, indexes.size());
+            ElasticSearchTimeSeriesIndex index = indexes.get(0);
+            Assert.assertEquals(index.getNamespace(), System.getProperty("elastic.search.namespace", ""));
```
wu-sheng: Same question where does the property come from?

oap-server/server-library/library-client/src/test/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ITElasticSearchClient.java
Arugal: Can I use the ```maven-failsafe-plugin```, as follows
```
                    <plugin>
                        <groupId>org.apache.maven.plugins</groupId>
                        <artifactId>maven-failsafe-plugin</artifactId>
                        <configuration>
                            <systemPropertyVariables>
                                <elastic.search.address>
                                    ${docker.hostname}:${es-port}
                                </elastic.search.address>
                                <elastic.search.namespace>
                                    test
                                </elastic.search.namespace>
                            </systemPropertyVariables>
                        </configuration>
                        <executions>
                            <execution>
                                <goals>
                                    <goal>integration-test</goal>
                                    <goal>verify</goal>
                                </goals>
                            </execution>
                        </executions>
                    </plugin>
```

oap-server/server-library/library-client/src/test/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ITElasticSearchClient.java
Arugal: Because my local tests don't create a new es container every time, I delete it

oap-server/server-library/library-client/src/test/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ITElasticSearchClient.java
Arugal: After adding ```namespace```,```getIndex``` has the same problem, which I will modify together
https://github.com/apache/skywalking/blob/b21db17148df12f34a13a2b1693e009e11e8d7a2/oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java#L138-L146

wu-sheng: System properties are used for uncertain variables, because of docker env. You don't need this, actually, you could add it to test codes directly, which will be more clear.

wu-sheng: Also, @Arugal `#getIndex` is not used in codes anymore, we only use it in IT test, let's move it to ITElasticSearchClient? That will make us has fewer codes concern of index name.

Arugal: > Also, @Arugal `#getIndex` is not used in codes anymore, we only use it in IT test, let's move it to ITElasticSearchClient? That will make us has fewer codes concern of index name.

```#getIndex``` needs to invoker ```RestHighLevelClient#getLowLevelClient``` ,unless we can get a ```RestHighLevelClient``` instance via ```ElasticSearchClient```

Comment (Arugal) [reply oap-server/server-library/library-client/src/test/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ITElasticSearchClient.java]: > System properties are used for uncertain variables, because of docker env. You don't need this, actually, you could add it to test codes directly, which will be more clear.

done

File: oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java [hunk 8]
```diff
@@ -292,7 +300,34 @@ public int delete(String indexName, String timeBucketColumnName, long endTimeBuc
...
-            return namespace + "_" + indexName;
+            return namespacePrefix + indexName;
+        }
+        return indexName;
+    }
+
+    private JsonObject undoFormatIndexName(JsonObject index) {
+        if (StringUtils.isNotEmpty(namespace) && index != null && index.size() > 0) {
+            Set<Map.Entry<String, JsonElement>> entrySet = index.entrySet();
+            for (Map.Entry<String, JsonElement> entry : entrySet) {
```
wu-sheng: Change this to `index.foreach -> `

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -119,20 +123,20 @@ public boolean createIndex(String indexName, JsonObject settings, JsonObject map
...
+            Set<String> keySet = responseJson.keySet();
+            if (!keySet.isEmpty()) {
+                return keySet.stream().map(this::undoFormatIndexName).map(index -> new ElasticSearchTimeSeriesIndex(namespace, index)).collect(Collectors.toList());
+            }
         }
-        return indexes;
+        return Collections.EMPTY_LIST;
     }
 
     public JsonObject getIndex(String indexName) throws IOException {
```
wu-sheng: `getIndex` method needs to move to `ITElasticSearchClient` with `undoFormatIndexName(JsonObject index)` method.

oap-server/server-library/library-client/src/test/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ITElasticSearchClientOfNamespace.java
```diff
@@ -0,0 +1,30 @@
...
+
+package org.apache.skywalking.oap.server.library.client.elasticsearch;
+
+/**
+ * @author: zhangwei
+ */
+public class ITElasticSearchClientOfNamespace extends ITElasticSearchClient {
+
+
+    public ITElasticSearchClientOfNamespace() {
```
wu-sheng: Has this run? I haven't used this kind of test before. Please confirm in the log.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -292,7 +300,34 @@ public int delete(String indexName, String timeBucketColumnName, long endTimeBuc
 
     public String formatIndexName(String indexName) {
         if (StringUtils.isNotEmpty(namespace)) {
-            return namespace + "_" + indexName;
+            return namespacePrefix + indexName;
+        }
+        return indexName;
+    }
+
+    private JsonObject undoFormatIndexName(JsonObject index) {
```
wu-sheng: We don't need this method, this should move to test class only. Right?

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
Arugal: done

oap-server/server-library/library-client/src/test/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ITElasticSearchClientOfNamespace.java
Arugal: [This is my local compile log](https://github.com/Arugal/skywalking-install-log/blob/master/install.txt)

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
Arugal: done

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
Arugal: done

Arugal:
> You miss some suggestions from last review.

Do I have any unfinished suggestions now?

wu-sheng:
getIndex method is still there. I think no ine uses it, should remove, right?

oap-server/server-library/library-client/src/test/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ITElasticSearchClient.java
Arugal: > > Also, @Arugal `#getIndex` is not used in codes anymore, we only use it in IT test, let's move it to ITElasticSearchClient? That will make us has fewer codes concern of index name.
> 
> `#getIndex` needs to invoker `RestHighLevelClient#getLowLevelClient` ,unless we can get a `RestHighLevelClient` instance via `ElasticSearchClient`



> getIndex method is still there. I think no ine uses it, should remove, right?

Can I do what I said above?

wu-sheng:
> I'm not suggest fixed it by this way. You can just add a new delete method without format. It is a very simply way.

@peng-yongsheng That kind of delete is very hard to understand from API use perspective. When to use delete with prefix, when don't? That is also very hard to review.
I hope you have a better suggestion, but a new delete with string just without format inside, looks not good.

peng-yongsheng:
This 

> > I'm not suggest fixed it by this way. You can just add a new delete method without format. It is a very simply way.
> 
> @peng-yongsheng That kind of delete is very hard to understand from API use perspective. When to use delete with prefix, when don't? That is also very hard to review.
> I hope you have a better suggestion, but a new delete with string just without format inside, looks not good.

Although it is hard to understand, it is the best way. This is the implementation of the Elasticsearch.
Before you modify it, some things you must know.
- Template index_patterns 
- Template aliases
- Template retrieval Index by aliases
Those things are difficult to understand. Retrieval index by aliases, return the index name with the namespace, then substring it from the index name, then join together the namespace in the delete method. This is a strange operation. Unless we don't use aliases to retrieval Index.

wu-sheng:
@peng-yongsheng If you just want an easy way, please consider putting two good method names. Because time series indexes exist, we can't set these methods as `#deleteByIndexName` and `#deleteByModelName`. 
In time series case, the model `A` has indexes including `A-20190913`,`A-20190914`, etc., but we can't call `#deleteByModelName(A)`(I know we don't use in this way). But still, show the API is hard named.

peng-yongsheng:
> @peng-yongsheng If you just want an easy way, please consider putting two good method names. Because time series indexes exist, we can't set these methods as `#deleteByIndexName` and `#deleteByModelName`.
> In time series case, the model `A` has indexes including `A-20190913`,`A-20190914`, etc., but we can't call `#deleteByModelName(A)`(I know we don't use in this way). But still, show the API is hard named.

Agree with my two hands.

Arugal:
> > @peng-yongsheng If you just want an easy way, please consider putting two good method names. Because time series indexes exist, we can't set these methods as `#deleteByIndexName` and `#deleteByModelName`.
> > In time series case, the model `A` has indexes including `A-20190913`,`A-20190914`, etc., but we can't call `#deleteByModelName(A)`(I know we don't use in this way). But still, show the API is hard named.
> 
> Agree with my two hands.

Now, I only think ```ElasticSearchTimeSeriesIndex ``` redundant

wu-sheng:
@peng-yongsheng @dmsolr Then this is my suggestion
1. Change `public boolean deleteIndex(String indexName) throws IOException` to `public boolean deleteByModelName(String noTimeSeriesMode) throws IOException`, also add comment to describe the argument.
1. Add `public boolean deleteByIndexName(String indexName) throws IOException`, do delete OP w/o format.

> Now, I only think ElasticSearchTimeSeriesIndex redundant

Yes. But basically, we are just talking about naming style.

For `getIndex`, if you need it in ITTest, add a new method in ITTest.

peng-yongsheng:
> > > @peng-yongsheng If you just want an easy way, please consider putting two good method names. Because time series indexes exist, we can't set these methods as `#deleteByIndexName` and `#deleteByModelName`.
> > > In time series case, the model `A` has indexes including `A-20190913`,`A-20190914`, etc., but we can't call `#deleteByModelName(A)`(I know we don't use in this way). But still, show the API is hard named.
> > 
> > 
> > Agree with my two hands.
> 
> Now, I only think `ElasticSearchTimeSeriesIndex ` redundant

Yes, If you use it, it's not the only place that needs it. It is a strategy that how to manage the namespace. Not just for the delete method. So, I do not suggest you use this way. I had thought it at the time of I program the time series implementation. At that time, I had an ideal same as you. But, If we must keep the overall style of the code. I found there are too many codes that need to modify.

peng-yongsheng:
> > > @peng-yongsheng If you just want an easy way, please consider putting two good method names. Because time series indexes exist, we can't set these methods as `#deleteByIndexName` and `#deleteByModelName`.
> > > In time series case, the model `A` has indexes including `A-20190913`,`A-20190914`, etc., but we can't call `#deleteByModelName(A)`(I know we don't use in this way). But still, show the API is hard named.
> > 
> > 
> > Agree with my two hands.
> 
> Now, I only think `ElasticSearchTimeSeriesIndex ` redundant

@Arugal 
If you interest in how to manage the namespace. I think we can talk about providing a new class named  ModelName with at least two attributes. Namespace and model name. And a method to build the final name, and a method to split it. Fixed this PR, think it in the another PR.

Arugal:
> If you interest in how to manage the namespace. I think we can talk about providing a new class named ModelName with at least two attributes. Namespace and model name. And a method to build the final name and a method to split it.

Yes

>  Fixed this PR, think it in the another PR.

Do I need to make any modification to this PR

peng-yongsheng:
> > If you interest in how to manage the namespace. I think we can talk about providing a new class named ModelName with at least two attributes. Namespace and model name. And a method to build the final name and a method to split it.
> 
> Yes, but wouldn't it be better to know the namespace with less code
No problem, you can new an issue to discuss it. 

> > Fixed this PR, think it in the another PR.
> 
> Do I need to make any modification to this PR
Yes, just provide those two methods, '#deleteByIndexName' and '#deleteByModelName'. Rollback other changed files.

wu-sheng:
I still recommend to remove getIndex method, only keep it in test case. Don't revert that.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -57,13 +60,15 @@
     public static final String TYPE = "type";
     private final String clusterNodes;
     private final String namespace;
+    private final String namespacePrefix;
     private final String user;
     private final String password;
-    private RestHighLevelClient client;
+    @Getter(value = AccessLevel.PACKAGE) private RestHighLevelClient client;
```
wu-sheng: Use whitebox get from powermock to get this.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -57,13 +58,15 @@
...
     private final String namespace;
+    private final String namespacePrefix;
     private final String user;
     private final String password;
     private RestHighLevelClient client;
 
     public ElasticSearchClient(String clusterNodes, String namespace, String user, String password) {
         this.clusterNodes = clusterNodes;
         this.namespace = namespace;
+        this.namespacePrefix = namespace + "_";
```
peng-yongsheng: I think this.namespacePrefix attribute is no value because we have a common method named formatIndexName.

Arugal: done

wu-sheng:
ASF jenkins break, we are waiting.

PR #2766, opened 2019-05-27
Replace add with addAll, reduce the invoke time a little.

Arugal:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

PR #3021, opened 2019-07-08
Using ArryList instead of LinkedList

dmsolr:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

___

Arylist has better performance than LinkedList.

Benchmark | Mode | Cnt | Score | Error | Units
---|---|---|---|---|----
DmBenchmark.testArray | thrpt | 5 | 18941.928 | ± 1594.049 | ops/s
DmBenchmark.testLinked | thrpt  |  5 | 12825.072 | ±  128.101 | ops/s

And to make ConsumerThread have one ArrayList to reduce create ArrayList Object and grow.

### DIFF
--- a/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumerThread.java
+++ b/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumerThread.java
@@ -19,9 +19,10 @@
 
 package org.apache.skywalking.apm.commons.datacarrier.consumer;
 
-import java.util.LinkedList;
-import java.util.List;
 import org.apache.skywalking.apm.commons.datacarrier.buffer.Buffer;
+
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * Created by wusheng on 2016/10/25.
@@ -36,7 +37,7 @@ public class ConsumerThread<T> extends Thread {
         super(threadName);
         this.consumer = consumer;
         running = false;
-        dataSources = new LinkedList<DataSource>();
+        dataSources = new ArrayList<DataSource>(1);
         this.consumeCycle = consumeCycle;
     }
 
@@ -64,34 +65,30 @@ public class ConsumerThread<T> extends Thread {
     public void run() {
         running = true;
 
+        final List<T> consumeList = new ArrayList<T>();
         while (running) {
-            boolean hasData = consume();
+            boolean hasData = consume(consumeList);
 
             if (!hasData) {
                 try {
                     Thread.sleep(consumeCycle);
                 } catch (InterruptedException e) {
                 }
+            } else {
+                consumeList.clear();
             }
         }
 
         // consumer thread is going to stop
         // consume the last time
-        consume();
+        consume(consumeList);
 
         consumer.onExit();
     }
 
-    private boolean consume() {
-        boolean hasData = false;
-        LinkedList<T> consumeList = new LinkedList<T>();
+    private boolean consume(List<T> consumeList) {
         for (DataSource dataSource : dataSources) {
-            LinkedList<T> data = dataSource.obtain();
-            if (data.size() == 0) {
-                continue;
-            }
-            consumeList.addAll(data);
-            hasData = true;
+            dataSource.obtain(consumeList);
         }
 
         if (consumeList.size() > 0) {
@@ -101,7 +98,7 @@ public class ConsumerThread<T> extends Thread {
                 consumer.onError(consumeList, t);
             }
         }
-        return hasData;
+        return !consumeList.isEmpty();
     }
 
     void shutdown() {
@@ -122,8 +119,8 @@ public class ConsumerThread<T> extends Thread {
             this.end = end;
         }
 
-        LinkedList<T> obtain() {
-            return sourceBuffer.obtain(start, end);
+        void obtain(List<T> consumeList) {
+            sourceBuffer.obtain(consumeList, start, end);
         }
     }
 }
--- a/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/buffer/Buffer.java
+++ b/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/buffer/Buffer.java
@@ -18,9 +18,11 @@
 
 package org.apache.skywalking.apm.commons.datacarrier.buffer;
 
-import java.util.*;
 import org.apache.skywalking.apm.commons.datacarrier.callback.QueueBlockingCallback;
 import org.apache.skywalking.apm.commons.datacarrier.common.AtomicRangeInteger;
+
+import java.util.LinkedList;
+import java.util.List;
 
 /**
  * Created by wusheng on 2016/10/25.
@@ -79,19 +81,17 @@ public class Buffer<T> {
         return buffer.length;
     }
 
-    public LinkedList<T> obtain() {
-        return this.obtain(0, buffer.length);
+    public void obtain(List<T> consumeList) {
+        this.obtain(consumeList, 0, buffer.length);
     }
 
-    public LinkedList<T> obtain(int start, int end) {
-        LinkedList<T> result = new LinkedList<T>();
+    public void obtain(List<T> consumeList, int start, int end) {
         for (int i = start; i < end; i++) {
             if (buffer[i] != null) {
-                result.add((T)buffer[i]);
+                consumeList.add((T)buffer[i]);
                 buffer[i] = null;
             }
         }
-        return result;
     }
 
 }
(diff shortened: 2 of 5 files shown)


apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumerThread.java
```diff
@@ -36,7 +37,7 @@
         super(threadName);
         this.consumer = consumer;
         running = false;
-        dataSources = new LinkedList<DataSource>();
+        dataSources = new ArrayList<DataSource>(1);
```
wu-sheng: ArrayList size 1? Why?

dmsolr: Because I think the default value(10) is large. It will only grow at initialization. Could you give some suggestion?

wu-sheng: This is why we use LinkedList. From several Google articles, we should always use ArrayList(no LinkedList)

But in this case, we can't be sure which is the right size for ArrayList, and how to be sure ArrayList is better in all scenario.

I used to think about use ArrayList in TracingContext, another core using LinkedList today. But I find out it is hard to make a reasonable decision.

wu-sheng: In DataCarrier Consumer, there are two
1. ConsumerThread(agent used today)
1. MultipleChannelsConsumer(OAP used today)

In the queue, it is much easier than TracingContext, you need to tests the performance in different initial sizes and different max sizes.

Such as in `MultipleChannelsConsumer`
1. Init=10, Expected list size=40000. In 16CPU VM, 32 Threads deal with 100 queues,  meaning max list size = 10000 * 4(queue size). Read `MultipleChannelsConsumer`
1. Init=1, Max data size=40000
1. Init = 40000, Max = 40000
1. Init = 8000, Max = 40000

Another optimization option is, hold the consumer list, just set the element to `null`, to avoid the size extension, but cost more GC potentially.

wu-sheng: Anyway, if you want to do this, I think we should work on a test blog first, work on different scenarios, then back to this PR. If this makes sense to you.

cc @zhaoyuguang @kezhenxu94 @ascrutae @JaredTan95   in case you are interested in this core optimization.

dmsolr: Thanks for your comments. I will think more and provide more completely benchmark cases later.
Welcome to join.

wu-sheng: Feel free to create a new repo for set up benchmarks, if that is easier to create  benchmarks before change codes.

peng-yongsheng: I had tested the performance of the ArrayList and LinkedList. But the report was lost. I just remember that LinkedList is faster than ArrayList if the size is below 5w.
A very important thing you must know is Datacarrier sends it's data to the consumer 20ms. So, if the queue size more than 5w, that means the TPS is 5w * (1000ms / 20ms) = 250w. This is a not reachable TPS.

@dmsolr

wu-sheng:
+1 for mutiple and real OAP and agent scenarios.

dmsolr:
Thank both of you.
I ran two scenarios. 
1. run benchmark tests with ArrayList and LinkedList. 
2. run simulate real-world testing.

The first of all, I need to explain two about this changes.
1. In constructor, I use ArrayList to store DataSource and the ArrayList size is from 1. This is why we just use it at initialization. And its length must less 100. So ArrayList or LinkedList is amlost the same here.
2. In  ComsumerThread#consume(), I use singleton ArrayList instance to avoid recreate and continuely grow. that means Array grownth is approaching to bufferSize(in MultiConsumerThread, it is less several times than bufferSize). And then it never grow.

Second, How do I design test case for simulate real-world testing.
Using `io.dropwizard.metrics` to trace GC and elapsed time. So I create a DataCarrier with 4-channels and 1000/2000/4000-buffers. And then I add 32-thread producer for continuely produce and 32-ConsumerThread for consume.

image

image

Here is the result: (It ran in Aliyun VM 2core, 4G)
## before changing.

* channelSize = 4
* bufferSize = 2000
* numThreads = 32
* run 3-mins

============================================================

-- Gauges ----------------------------------------------------------------------
PS-MarkSweep.count
             value = 0
PS-MarkSweep.time
             value = 0
PS-Scavenge.count
             value = 332
PS-Scavenge.time
             value = 477

-- Timers ----------------------------------------------------------------------
ConsumerThread.1
             count = 613190
         mean rate = 3404.58 calls/second
     1-minute rate = 3408.83 calls/second
     5-minute rate = 3390.47 calls/second
    15-minute rate = 3382.55 calls/second
               min = 0.00 milliseconds
               max = 2.46 milliseconds
              mean = 0.01 milliseconds
            stddev = 0.12 milliseconds
            median = 0.00 milliseconds
              75% <= 0.01 milliseconds
              95% <= 0.02 milliseconds
              98% <= 0.03 milliseconds
              99% <= 0.03 milliseconds
            99.9% <= 2.46 milliseconds

## after changing.

* channelSize = 4
* bufferSize = 2000
* numThreads = 32
* run 3-mins

============================================================
-- Gauges ----------------------------------------------------------------------
PS-MarkSweep.count
             value = 0
PS-MarkSweep.time
             value = 0
PS-Scavenge.count
             value = 107
PS-Scavenge.time
             value = 131

-- Timers ----------------------------------------------------------------------
ConsumerThread.1
             count = 621289
         mean rate = 3449.42 calls/second
     1-minute rate = 3439.23 calls/second
     5-minute rate = 3431.61 calls/second
    15-minute rate = 3423.44 calls/second
               min = 0.00 milliseconds
               max = 0.02 milliseconds
              mean = 0.00 milliseconds
            stddev = 0.00 milliseconds
            median = 0.00 milliseconds
              75% <= 0.00 milliseconds
              95% <= 0.00 milliseconds
              98% <= 0.01 milliseconds
              99% <= 0.01 milliseconds
            99.9% <= 0.02 milliseconds

dmsolr:
I copy the module to new repo. I push the simulate tests codes into there.
https://github.com/dmsolr/apm-datacarrier-benchmark

wu-sheng:
Please use jmh to run benchmark, it shows gc. ops and latency. We have many benchmarks like this.

dmsolr:
Those are the code and the result of benchmark tests. Although it is not too strict.

image

image


Thank @kezhenxu94 for your suggestion.

dmsolr:
Thank @kezhenxu94 for your suggestion and help.
Those are the code and the result of benchmark tests. Although it is not too strict.

image

image

Arugal:
[This is a good demonstration of JMH](https://github.com/apache/skywalking/pull/2766#issuecomment-496240702)

peng-yongsheng:
> Thank @kezhenxu94 for your suggestion and help.
> Those are the code and the result of benchmark tests. Although it is not too strict.
> 
> image
> 
> image

Hi, @dmsolr 
From the implementation of the ArrayList and LinkedList in JDK, the LinkedList create node object to store data and reference to the next node, LinkedList creates much more node instance. So,  LinkedList will collect more garbage than ArrayList. 
What's the purpose of comparing this?

dmsolr:
Hi @peng-yongsheng ,
We want to know the effect of large arrays on GC. Garbage collector need to check all elements of array every time. That will cost more CPU and make GC taking more time. But we are not sure how many it will cost. So we do this test.

peng-yongsheng:
> Hi @peng-yongsheng ,
> We want to know the effect of large arrays on GC. Garbage collector need to check all elements of array every time. That will cost more CPU and make GC taking more time. But we are not sure how many it will cost. So we do this test.

@dmsolr Maybe it's better to do an OAP server comparison. Mock the same amount segment and post into the OAP server, then compare the GC report between before and after modification. This can let us know the performance in a real scenario.

dmsolr:
Hi @peng-yongsheng , thanks for your comment. 
OAP server is complex so that we cannot do benchmark. But I can test this like [DataCarrier Tests](https://github.com/apache/skywalking/pull/3021#issuecomment-509896369) .

In my idea, DataCarrier tests may represent something.

wu-sheng:
Let's work on DataCarrier tests, then we could discuss more. Anyway, this is a complex proposal, should we close this PR, and work on an issue to discuss tests and results?

dmsolr:
OK. I close this PR and open a new issue later.

Issue #3033, opened 2019-07-10
6.2.0 DataTTLKeeperTimer do not delete the expired records in the table segment


shinelon:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Bug
- Which version of SkyWalking, OS and JRE?
SkyWalking 6.2.0
storage Mysql 5.7.16
jdk 1.8.0_77

- What happen?
DataTTLKeeperTimer do not delete the expired records in the table segment.

the column time_bucket of table segment is second unit
but GeneralStorageTTL.java do not has SecondTTLCalculator

debug the 6.2.0 code : 

sql param unit is minute
table column unit is seconds

___
### Requirement or improvement
- 
add switch case `Second` to the GeneralStorageTTL.java can fix this bug??
Will it cause another bug?

case Second:
                return new SecondTTLCalculator();

wu-sheng:
We are deleting index rather than delete record today.

wu-sheng:
Are you talking about bug in Mysql storage?

wu-sheng:
Could you test locally and submit a pull request?

wu-sheng:
Cc @kezhenxu94 ttl e2e may be needed too. But a little complex, I think

@peng-yongsheng Your refator of TTL looks like breaking mysql

yangvipguang:
类似问题：
skywalking: 6.2 GA
ES  6.8.1

我这边部署完成后，一开始页面数据都正常。过一会再打开页面就没数据了。查看日志时间点正好是如下日志：2019-07-12 15:26:33,020 - org.apache.skywalking.oap.server.core.storage.ttl.DataTTLKeeperTimer - 61 [pool-17-thread-1] INFO  [] - Beginning to remove expired metrics from the storage.
猜测原因： 删除了当前的metrics
疑问：
是否因为配置问题还是本身bug


暂时解决方案就是：关闭配置文件中的自动删除后，没有出现这种问题，系统时间是正常的。

shook2012:
[...]

PR #3038, opened 2019-07-10
Try to fix MySQL TTL failure.

wu-sheng:
Following #3033, please re-test and confirm.


wu-sheng:
@shinelon  Could you check this branch?

Issue #3055, opened 2019-07-12
ContextManager.activeSpan() API usage error


zhaoyuguang:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [x] Question or discussion
- [ ] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- 
Look the code. such as:
![image](https://user-images.githubusercontent.com/10150229/61099499-1d4a4600-a495-11e9-9773-f417d6621d44.png)
![image](https://user-images.githubusercontent.com/10150229/61099508-29360800-a495-11e9-9f47-d8416577fc24.png)
The code uses `ContextManager.activeSpan()` to determine if there is an active span in the current thread, which will cause NPE when no active span. The correct approach should be `ContextManager.isActive()` 

I am right?

wu-sheng:
Yes. I think so.

PR #3057, opened 2019-07-12
Fix 3055

zhaoyuguang:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
Fix https://github.com/apache/skywalking/issues/3055

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManager.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManager.java
@@ -157,6 +157,10 @@ public class ContextManager implements BootService {
         return get().awaitFinishAsync();
     }
 
+    /**
+     * If not sure has the active span, use this method, will be cause NPE when has no active span,
+     * use ContextManager::isActive method to determine whether there has the active span.
+     */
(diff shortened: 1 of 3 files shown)


wu-sheng:
+1 for code scan. I also think, we should check all `ContextManager#activeSpan` in plugin, and decide whether we should `@Deprecated` or delete. Let's check that and talk.

dmsolr:
As I know, `ContextManager#activeSpan()` is used for `handleMethodException`

dmsolr:
I think it cannot be deleted. Because it is useful sometimes.

wu-sheng:
From the current failures, I am thinking about, is this right using `ContextManager#activeSpan()` even for logging error? It could cause an NPE too, right?

dmsolr:
Not NPE. It throws 'No active span.' exception when `ContextManager#activeSpan()` called.

wu-sheng:
> Not NPE. It throws 'No active span.' exception when ContextManager#activeSpan() called.

But the exception should not happen in agent, it costs too much.

dmsolr:
I entirely agree with you. Maybe I have a mistake. I mean we need `ContextManager#activeSpan()` in other scenario.

zhaoyuguang:
I scanned the code and found only the two places, which are obvious usage errors.

PR #3065, opened 2019-07-12
Propose a API change : AbstractTracerContext#asyncStop and AsyncSpan#asyncFinish return boolean

qxo:
change API AbstractTracerContext#asyncStop and AsyncSpan#asyncFinish return boolean
    return true when in asyncStop do invoke finished

目的是提供能力用来判断调用是否成功调用了fishished方法

原因如下：
  在调用了prepareForAsync后，而后如果是最后代码点调用asyncStop因目前asyncStop实现法中和stopSpan方法中pop last span逻辑而致没法触发fishished方法
如此时再调用stopSpan因asyncSpanCounter==0可以触发fishished方法
但如果调用asyncStop方法内已经触发fishished方法，而后直接再调用stopSpan则会再次触发fishished方法
同时也不能直接调用stopSpan(因之前调用prepareForAsync,asyncSpanCounter不为0且isRunningInAsyncMode=true,故没法触发fishished方法）
由于实际代码存在两种情况，需要调整API返回调后的结果以后判断

之后就这样调用：
```
if (!span.asyncFinish() ){
    ContextManager.stopSpan(span);
}
```

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AbstractTracerContext.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AbstractTracerContext.java
@@ -113,6 +113,7 @@ public interface AbstractTracerContext {
      * The given span could be stopped officially.
      *
      * @param span to be stopped.
+     * @return true when finished 
      */
-    void asyncStop(AsyncSpan span);
+    boolean asyncStop(AsyncSpan span);
 }
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AsyncSpan.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AsyncSpan.java
@@ -50,7 +50,7 @@ public interface AsyncSpan {
      *
      * The execution times of {@link #prepareForAsync} and {@link #asyncFinish()} must match.
      *
-     * @return the current span
+     * @return true when the current span is finished
      */
-    AbstractSpan asyncFinish();
+    boolean asyncFinish();
 }
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/IgnoredTracerContext.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/IgnoredTracerContext.java
@@ -97,8 +97,8 @@ public class IgnoredTracerContext implements AbstractTracerContext {
         return this;
     }
 
-    @Override public void asyncStop(AsyncSpan span) {
-
+    @Override public boolean asyncStop(AsyncSpan span) {
+        return stackDepth == 0;
     }
 
     public static class ListenerManager {
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/TracingContext.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/TracingContext.java
@@ -423,12 +423,14 @@ public class TracingContext implements AbstractTracerContext {
         return this;
     }
 
-    @Override public void asyncStop(AsyncSpan span) {
+    @Override public boolean asyncStop(AsyncSpan span) {
         asyncSpanCounter.addAndGet(-1);
 
         if (checkFinishConditions()) {
             finish();
+            return true;
         }
+        return false;
     }
 
     private boolean checkFinishConditions() {
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/AbstractTracingSpan.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/AbstractTracingSpan.java
@@ -327,13 +327,12 @@ public abstract class AbstractTracingSpan implements AbstractSpan {
         return this;
     }
 
-    @Override public AbstractSpan asyncFinish() {
+    @Override public boolean asyncFinish() {
         if (!isInAsyncMode) {
             throw new RuntimeException("Span is not in async mode, please use '#prepareForAsync' to active.");
         }
 
         this.endTime = System.currentTimeMillis();
-        context.asyncStop(this);
-        return this;
+        return context.asyncStop(this);
     }
 }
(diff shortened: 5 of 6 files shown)


kezhenxu94:
Please translate into English and then we'll discuss and merge (if possible)

wu-sheng:
No matter Chinese or English, this API proposal is not right. Every tracing API have change to miss closing, but that the responsibilities of plugin. There is no need to do this check. You should know as plugin developer. If close is un clear, then should not use async, choose capture and continued.

qxo:
@wu-sheng  where the right example of asyncFinish?  I try to fix the vertx plugin, but can not fix the logic...

可否给个样例

wu-sheng:
How to finish it is plugin responsibility. You could use dynamic field or anything you like to coordinate it. But basically, I don't know vertx, so I need you to make it clear why you can't be sure about when to close it, and why you try to finish it duplicatedly.

I am just saying, your proposal doesn't change anything, because think in this way, if there is two async spans, what happens? Is this API change still helpful? I think it isn't, no one should check the context in this way.

It is same when you called span#close twice, you close other plugin span.

qxo:
@wu-sheng 
1.  just one span
2. span.prepareForAsync()  then  span.asyncFinish() will not finished the span
   async code not work as the [async-span-apis doc](https://github.com/apache/skywalking/blob/master/docs/en/guides/Java-Plugin-Development-Guide.md#async-span-apis)
3. sorrry for my poor english
我还是用中文表述：
   虽然之前给出的API用例是不正确，但这个API的改除了插件要重新才能使用编译外没其他影响
   可能是API使用不当，也可能是API内部实施逻辑有欠缺，今天我再验证一下
没找到对应的API的使用样例，有的话麻烦提供一下：）

Question:
1. maybe just one span or maybe not, how to code the plugin on the last code ? should I use span.prepareForAsync()  then  span.asyncFinish() ?  

2. should [asyncStop](
https://github.com/apache/skywalking/blob/535878054313d94e00f81c6a647bac2f92559207/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/TracingContext.java#L426) the pop logic same as [stopSpan](https://github.com/apache/skywalking/blob/535878054313d94e00f81c6a647bac2f92559207/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/TracingContext.java#L390)

qxo:
```
import org.apache.skywalking.apm.agent.core.conf.RemoteDownstreamConfig;
import org.apache.skywalking.apm.agent.core.context.ContextCarrier;
import org.apache.skywalking.apm.agent.core.context.ContextManager;
import org.apache.skywalking.apm.agent.core.context.tag.Tags;
import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
import org.apache.skywalking.apm.agent.core.context.trace.SpanLayer;
import org.apache.skywalking.apm.agent.core.dictionary.DictionaryUtil;
import org.apache.skywalking.apm.network.trace.component.ComponentsDefine;

public class TestAsyncStop {

    public static void main(String[] args) throws Exception {
        long start = System.currentTimeMillis();
        // wait for skywalking agent ready
        while (true) {
            Thread.sleep(200l);
            if (RemoteDownstreamConfig.Agent.SERVICE_ID != DictionaryUtil.nullValue()
                    && RemoteDownstreamConfig.Agent.SERVICE_INSTANCE_ID != DictionaryUtil.nullValue()) {
                System.out.println("skywalking is ready:" + (System.currentTimeMillis() - start) + " ms");
                break;
            }
        }

        ContextCarrier contextCarrier = new ContextCarrier();
        final String opName = "/abc123";
        AbstractSpan span = ContextManager.createEntrySpan(opName, contextCarrier);

        span.setComponent(ComponentsDefine.VERTX);
        SpanLayer.asHttp(span);

        Tags.HTTP.METHOD.set(span, "GET");
        Tags.URL.set(span, opName);
        span.prepareForAsync();
        Thread.sleep(1l);
        span.asyncFinish();
        try {// wait for skywalking to finish
            Thread.sleep(200l);
            // TODO Auto-generated catch block
            e.printStackTrace();
        }
    }
 }
```
please fix the code:)

wu-sheng:
I am not following, your descriptions are not consistent. I don't understand how you want to do. But your proposal to return the count, the reason I am arguing, is we should not do that, it is a wrong design.

For 1, The document tolds you very clear, prepareAsync first, then finish at somewhere else. What is your question?
For 2, simple answer is no. 

I think you haven't understood what is async finish. Async finish is only for make span finished out of current tracing context, and keep tag and duration right.

qxo:
In the same tracing context, should not use prepareAsync ?
API should have sample use case test code  :)

API强壮性还有待提高： API设计上应该防止不当使用， A plugin的bug却影响了B plugin的功能:(

wu-sheng:
You can use in same thread for sure. I am keep talking, you look like just guessing, and use a main to test, rather than read codes.

We wouldn't do the thinga you said, check how you use API. Agent is a hijack and black tech, we are focusing on performance as always highest priority. So, we are not other library you used. In real world, no matter what API you do, you could effect other plugin, if you know what is distributed tracing.

wu-sheng:
You are entering a wrong direction, and trying to ask the wrong question.

wu-sheng:
The true is, you could make connection leak if you use the connection pool in wrong way, and crash application. Nothing could stop you to do that.

Also, the thing you said about the check, is even not performance related, it is just wrong. You are solving a question and bringing more to the core.

qxo:
Then what's wrong in 'TestAsyncStop' and how to using the api right?
Any sample code ?

This is current asyncStop ： 
```
    @Override public void asyncStop(AsyncSpan span) {
        asyncSpanCounter.addAndGet(-1);

        if (checkFinishConditions()) {
            finish();
        }
    }

    private boolean checkFinishConditions() {
        if (isRunningInAsyncMode) {
            asyncFinishLock.lock();
        }
        try {
            if (activeSpanStack.isEmpty() && (!isRunningInAsyncMode || asyncSpanCounter.get() == 0)) {
                return true;
            }
        } finally {
            if (isRunningInAsyncMode) {
                asyncFinishLock.unlock();
            }
        }
        return false;
    }
```
after prepareForAsync, isRunningInAsyncMode=true 
and in asyncStop activeSpanStack must be empty then will invoke "finish"
ony in "stopSpan" could make(pop) activeSpanStack  empty
so before "asyncStop"  we must invoke "stopSpan" first ?

要想在同一上下文调用asyncStop前必须先调用stopSpan？
这是我实测的结果，不知是否和测试相符？

代调用顺序如下调整可以发送数据到skywalking系统
但总感觉如下调用有点问题
```
       span.prepareForAsync();
        Thread.sleep(3l);
        ContextManager.stopSpan(span);
        Thread.sleep(1l);
        span.asyncFinish();  
```

另外关于API 健壮性：我主要指定的是目前实现一个span可以被多调次asyncStop, 如能限制一个span只能调用一次stop(或只有第一次调用能有附加影响:能改程序状态)，那就是基上能防止：“因plugin A的bug而导致之后的plugin B功能不正常" --  不过，这是另外一个问题

wu-sheng:
I wouldn't do more discussion, unless yoi change this to English. Sorry. 
This is unfriendly to others.

PR #3073, opened 2019-07-14
fixed for asyncFinishLock: avoid call finish more than once

qxo:

- [ *] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
fixed for asyncFinishLock: avoid call finish more than once
- How to fix?
call "finish"  after asyncFinishLock lock and before unlock
___


qxo:
1. set breakpoint before call checkFinishConditions and inside checkFinishConditions, then try in debug mode,asyncStop in other thread(pause before checkFinishConditions ), let stopSpan pop make activeSpanStack to empty first,...
2. It's hard reproduct without codeline debug
3. I'll do the demo if have the time
4. In my point of view: The method(checkFinishConditions) code do not change program(object) state,the asyncFinishLock is useless if it's not a problem.
===>  理论上一个方法内部没改状态且程序又依赖于其返回值，方法内部加锁是不正确的使用方式：因为多线程在同时请求checkFinishConditions时就会出问题
如有不对，请指:)

wu-sheng:
I think I get your point of issue, but if you are talking about `#asyncStop` and `#stopSpan` called concurrently, then your solution is still not working, due to the same reason **no status change**.

I think this problem only could be solved by adding a field `#running` in `TracingContext`, default true in the constructor. check `running==true` before finish condition check, set it to false if check passed. Then add your change, move finish into it.
```java
// rename checkFinishConditions
private void finish () {
     ... lock
     if (running) {
            if (activeSpanStack.isEmpty() && (!isRunningInAsyncMode || asyncSpanCounter.get() == 0)) {
                  running = false;
                  _doFinish(); // rename old finish method
            }
    }
   .... unlock
}
```

qxo:
Yep, my fix still have issue.Please fix it :)

wu-sheng:
> Yep, my fix still have issue.Please fix it :)

By you or me? Do you want me to raise to new PR?

PR #3075, opened 2019-07-14
Fix wrong logger names

kezhenxu94:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

Logger name differentiate from the real class may be troublesome when debugging, and we have several pull requests that try to correct some of them (#3054), but not all of them.

So I just wrote [a simple shell](https://gist.github.com/kezhenxu94/806af6bbc944b93f7aa6e0b18ce5f3f5) script to find out all such cases and fix them once and for all.

Since we are using `lombok`, to avoid such problem in the future, I'd recommend using `@Slf4j` annotation in future PRs when possible.


wu-sheng:
> Since we are using lombok, to avoid such problem in the future, I'd recommend using @slf4j annotation in future PRs when possible.

@kezhenxu94 You could change this in OAP side, I am just saying, the check may be still necessary.

wu-sheng:
Are we going to merge this first?

kezhenxu94:
> Are we going to merge this first?

I've just fixed those in agent side, and can be merged first now.

As for the check in CI, let me work on it in another PR

PR #3090, opened 2019-07-15
Fixed a mistake that inject v1 segment producer into v2 standardizati…

peng-yongsheng:
…on worker.

- [X] Bug fix
- [ ] New feature provided
- [ ] Improve performance


wu-sheng:
Just for the record, this bug wouldn't cause any issue, because of protobuf compatible design, and due to we don't change field sequence, but just field name, which is not used in protobuf serialization.

wu-sheng:
Look like Jenkins has some issues, let's wait for a moment.

PR #3092, opened 2019-07-16
Fix bug conflicts when containers start simultaneously

kezhenxu94:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

The e2e containers' names of single mode and cluster mode
would conflict when they start up simultaneously in Jenkins job.


kezhenxu94:
The e2e cluster test Jenkins jobs failed occasionally due to two main reasons, one of which is caused by container name conflicts, which should be fixed by this patch, the other is caused by timing out, and is under investigation

Issue #3064, opened 2019-07-12
To improve DataCarrier 


dmsolr:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [x] Feature or performance improvement

___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

This issue traces whether to use ArrayList instead of LinkedList. We have worked something on #3021  before this. Now we need to discuss how to make sure which is the better.

wu-sheng:
Hi, I am back from travell, so we could work on this now.

For your prev tests
![image](https://user-images.githubusercontent.com/5441976/61172239-1a924280-a5b4-11e9-9074-f85828611ec6.png)
1. You set consume size to 4k, from my prev calculation, it should at least 40k, is it your mistake or mine?
1. You didn't clear the link after each consume cycle, I assume it is a bug? If you don't add, the link size keeps increasing.

How about improving this as the first step?

dmsolr:
1. It is my mistake. 
2. It is recreated every time. That is why make the test more seem like reality. (I know it out of benchmark)

wu-sheng:
> It is recreated every time. That is why make the test more seem like reality. (I know it out of benchmark)

You are testing different list size, why is it more real?

dmsolr:
The max capacity is 40k with default initial capacity. It showed that the ArrayList seem like better in ops. And the reused ListedList is better at GC.

```
Benchmark                                                                   Mode  Cnt           Score          Error   Units
    LinkedArrayBenchmark.testArray                                             thrpt    5           4.226 ±        0.141   ops/s
    LinkedArrayBenchmark.testArray:·gc.alloc.rate                              thrpt    5        3688.588 ±      123.565  MB/sec
    LinkedArrayBenchmark.testArray:·gc.alloc.rate.norm                         thrpt    5   960569530.094 ±        0.412    B/op
    LinkedArrayBenchmark.testArray:·gc.churn.PS_Eden_Space                     thrpt    5        3689.467 ±      122.160  MB/sec
    LinkedArrayBenchmark.testArray:·gc.churn.PS_Eden_Space.norm                thrpt    5   960800434.530 ±  6919898.689    B/op
    LinkedArrayBenchmark.testArray:·gc.churn.PS_Survivor_Space                 thrpt    5           2.414 ±        0.083  MB/sec
[...]

PR #3093, opened 2019-07-16
Using ArrayList to instead of LinkedList

dmsolr:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues
#3064 
#3021 
___

According to #3064, we can learn ArrayList has ~3x faster than LinkedList.

### DIFF
--- a/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumerThread.java
+++ b/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumerThread.java
@@ -67,15 +67,11 @@ public class ConsumerThread<T> extends Thread {
 
         final List<T> consumeList = new ArrayList<T>(1500);
         while (running) {
-            boolean hasData = consume(consumeList);
-
-            if (!hasData) {
+            if (consume(consumeList)) {
                 try {
                     Thread.sleep(consumeCycle);
                 } catch (InterruptedException e) {
                 }
-            } else {
-                consumeList.clear();
             }
         }
 
@@ -97,8 +93,10 @@ public class ConsumerThread<T> extends Thread {
             } catch (Throwable t) {
                 consumer.onError(consumeList, t);
             }
+            consumeList.clear();
+            return true;
         }
-        return !consumeList.isEmpty();
+        return false;
     }
 
     void shutdown() {
--- a/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumerThread.java
+++ b/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumerThread.java
@@ -92,8 +92,9 @@ public class ConsumerThread<T> extends Thread {
                 consumer.consume(consumeList);
             } catch (Throwable t) {
                 consumer.onError(consumeList, t);
+            } finally {
+                consumeList.clear();
             }
-            consumeList.clear();
             return true;
         }
         return false;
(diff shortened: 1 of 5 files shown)


wu-sheng:
Fix CI and test first

apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumerThread.java
```diff
@@ -77,31 +77,26 @@ public void run() {
...
         }
 
-        if (consumeList.size() > 0) {
+        if (!consumeList.isEmpty()) {
             try {
                 consumer.consume(consumeList);
             } catch (Throwable t) {
                 consumer.onError(consumeList, t);
             }
+            consumeList.clear();
```
wu-sheng: All clear need in final. In case `#onError` fails.

apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/MultipleChannelsConsumer.java
```diff
@@ -55,34 +59,33 @@ public void run() {
...
         }
 
-        if (hasData = consumeList.size() > 0) {
+        if (!consumeList.isEmpty()) {
             try {
                 target.consumer.consume(consumeList);
             } catch (Throwable t) {
                 target.consumer.onError(consumeList, t);
             }
+            consumeList.clear();
```
wu-sheng: Same as above. Keep clear in final.

apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumerThread.java
dmsolr: I totatly accept this suggestion. In actually, None of error will out of the throwable. So it is safe, I think.

apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/MultipleChannelsConsumer.java
```diff
@@ -43,10 +46,11 @@ public MultipleChannelsConsumer(String threadName, long consumeCycle) {
     public void run() {
         running = true;
 
+        final List consumeList = new ArrayList(2000);
         while (running) {
             boolean hasData = false;
             for (Group target : consumeTargets) {
-                hasData = hasData || consume(target);
+                hasData |= consume(target, consumeList);
```
wu-sheng: You change `||` to `|`

dmsolr: They are the same here, right?

kezhenxu94: > They are the same here, right?

Not realy, it's `short circuit` operation (`consume` won't be called if `hasData == true`) before your changes and the `consume` would always be called after your changes

wu-sheng: > Not realy, it's short circuit operation (consume won't be called if hasData == true) before your changes and the consume would always be called after your changes

Let's make it clear, it should consume even hasData. If old codes didn't, that is a bug by me  :( I was not expecting this before. 
Let's remove the `|` and `||`, change it to normal codes 
```java
boolean consumeResult = consume(target);
hasData = hasData || consumeResult;
```

dmsolr: I got it.

wu-sheng: Checked. Old codes did `short circuit`, which is wrong. 
We should consume all group in one loop
I think this will improve performance.

@kezhenxu94 Thanks for pointing this out.

dmsolr: I have a test, `||` is shor circuit.

Issue #3096, opened 2019-07-16
redis-cluster too many nodes cause  too long index id in elasticsearch


carlvine500:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [-] Bug

___
### Bug
- Which version of SkyWalking, OS and JRE?
6.2.0 linux java8
- Which company or project?
oap
- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.
create a redis-cluster with lots of instance,such as :
172.31.88.62:8300;172.31.88.50:8300;172.31.88.62:8301;172.31.88.50:8301;172.31.88.62:8302;172.31.88.50:8302;172.31.88.62:8303;172.31.88.50:8303;172.31.88.62:8304;172.31.88.50:8304;172.31.88.62:8305;172.31.88.50:8305;172.31.88.62:8306;172.31.88.50:8306;172.31.88.62:8307;172.31.88.62:8308;172.31.88.50:8308;172.31.88.62:8309;172.31.88.62:8315;172.31.88.73:8316;172.31.88.73:8317;172.31.88.73:8318;172.31.88.73:8319;172.31.88.73:8300;172.31.88.73:8301;172.31.88.73:8302;172.31.88.73:8303;172.31.88.73:8304;172.31.88.73:8305;172.31.88.73:8306;172.31.88.73:8307;172.31.88.73:8308;172.31.88.73:8309;172.31.88.73:8315;172.31.88.50:8315;172.31.

and cause exception :  
```
2019-07-16 16:48:48,227 - org.apache.skywalking.oap.server.core.register.worker.RegisterPersistentWorker - 106 [DataCarrier.REGISTER_L2.BulkConsumePool.0.Thread] ERROR [] - Validation Failed: 1: id is too long, must be no longer than 512 bytes but was: 774;
org.elasticsearch.action.ActionRequestValidationException: Validation Failed: 1: id is too long, must be no longer than 512 bytes but was: 774;
        at org.elasticsearch.action.ValidateActions.addValidationError(ValidateActions.java:26) ~[elasticsearch-6.3.2.jar:6.3.2]
        at org.elasticsearch.action.index.IndexRequest.validate(IndexRequest.java:183) ~[elasticsearch-6.3.2.jar:6.3.2]
        at org.elasticsearch.client.RestHighLevelClient.performRequest(RestHighLevelClient.java:515) ~[elasticsearch-rest-high-level-client-6.3.2.jar:6.3.2]
[...]

PR #3105, opened 2019-07-17
Provide a configuration to control the length of the peer field

zhaoyuguang:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
https://github.com/apache/skywalking/issues/3096

- How to fix?
Add  Peer max length configuration, Used it in the interception of Jedis, Redission, and Lettuce cluster configuration.

if `agent.peer_max_length=5` 
topo like this
image
trace like this
image


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -93,6 +93,11 @@
          * How depth the agent goes, when log cause exceptions.
          */
         public static int CAUSE_EXCEPTION_DEPTH = 5;
+
+        /**
+         * Peer max length, Used in the interception of Jedis, Redission, and Lettuce cluster configuration.
+         */
+        public static int PEER_MAX_LENGTH = 200;
```
wu-sheng: Move this to Plugin, this is not a part of core API, right? And don't put plugin name in comments, you wouldn't know when someone add it to somewhere else, considering `PeerUtil` is a core API

zhaoyuguang:
> Is this PR ready? If so, change the title, and don't use issue number as the title, which is hard to read from PR list view.

OK, The PR is ready.

wu-sheng:
@zhaoyuguang Follow the review, please.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
zhaoyuguang: I modified the comments and MD documentation
Moved the `PEER_MAX_LENGTH ` configured location
Modified the name of the class and changed to the enum tool class
Please review code again

zhaoyuguang:
I fix the review issue, Please review code again @wu-sheng

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/PeerFormat.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.skywalking.apm.agent.core.conf.Config;
+import org.apache.skywalking.apm.util.StringUtil;
+
+/**
+ * @author zhaoyuguang
+ */
+
+public enum PeerFormat {
+
+    INSTANCE;
```
wu-sheng: There is no need to set in singleton. Just keep it as normal class(new one every time) or static method,

zhaoyuguang: Thanks for guidance, this does look better.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/PeerFormat.java
```diff
@@ -25,13 +25,11 @@
...
-
-    INSTANCE;
+public class PeerFormat {
 
     private static final String ABBR = "...";
 
-    public String Shorten(String original) {
+    public static String shorten(String original) {
         if (!StringUtil.isEmpty(original) && original.length() > Config.Plugin.PEER_MAX_LENGTH) {
             return original.substring(0, Config.Plugin.PEER_MAX_LENGTH - 3) + ABBR;
```
kezhenxu94: What if users set `Config.Plugin.PEER_MAX_LENGTH` to less than `3` (intentionally or by mistake) 😄

zhaoyuguang: Can cause unexpected exceptions, the size of the prompt on the document needs to exceed 3 and the judgment of increasing the length of `PEER_MAX_LENGTH ` is less than 3, less than 3 will be invalid and the default value is 200.

zhaoyuguang: image
I found another way, is this better?

PR #3106, opened 2019-07-17
Use CopyOnWriteArrayList instead of recreating manually

kezhenxu94:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

https://github.com/apache/skywalking/blob/8faa3564037570fca198ab9f70534f3d359a9d40/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/MultipleChannelsConsumer.java#L99-L109

the comment `// Recreate the new list to avoid change list while the list is used in consuming.` reminds me of the `CopyOnWrite` data structure that is perfectly suitable to this scenario, although the method `addNewTarget` is not frequently invoked, I did a simple benchmark out of curiosity.

This patch also fixes that the non-atomic operation `+=` on the `volatile` variable `size`

### DIFF
--- a/apm-commons/apm-datacarrier/src/test/java/org/apache/skywalking/apm/commons/datacarrier/CowVsManualCopyBenchmark.java
+++ b/apm-commons/apm-datacarrier/src/test/java/org/apache/skywalking/apm/commons/datacarrier/CowVsManualCopyBenchmark.java
@@ -1,0 +1,72 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.commons.datacarrier;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.Param;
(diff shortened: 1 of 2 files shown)


apm-commons/apm-datacarrier/src/test/java/org/apache/skywalking/apm/commons/datacarrier/CowVsManualCopyBenchmark.java
```diff
@@ -0,0 +1,72 @@
...
+        Options opt = new OptionsBuilder()
+            .include(CowVsManualCopyBenchmark.class.getSimpleName())
+            .mode(Mode.Throughput)
+            .forks(1)
+            .warmupIterations(4)
+            .measurementIterations(4)
+            .build();
+
+        new Runner(opt).run();
+    }
```
wu-sheng: You should post benchmark result here.

kezhenxu94:
FYI @wu-sheng , I'm recently considering whether we should create a new module/directory to place all the benchmark tests like other projects, the reasons are:

1. we have some benchmarks that are independent to any existed module, such as `LinkedArrayBenchmark`, `StringFormatBenchmarkTest`, IMHO they should be placed in an independent module, otherwise it's hard for new-comers to refer to and find out whether we already have that benchmark or not

2. some of the existed benchmarks are annotated with `@Test`, and are executed in maven test phase (I think), which is not necessary and time consuming, gathering them together may be easy to manage and remove `@Test`

wu-sheng:
@kezhenxu94 How about we create WIKI pages for that? And link to the PRs?

wu-sheng:
We have this. https://cwiki.apache.org/confluence/display/SKYWALKING/Home 

I could grant committers access right. Then all the acceptable performance improvement PRs, should have a benchmark list there.

kezhenxu94:
> @kezhenxu94 How about we create WIKI pages for that? And link to the PRs?

That's a great alternative, it would be easy for new-comers who are interested in performance optimization and can refer to the wiki

dmsolr:
Out of this patch.
ArrayList has worse performance because ArrayList uses `add()` one by one. It would be better if using `addAll()` or `new ArrayList<>(collections)`.

wu-sheng:
We accept a lot of performance improvements these days, but we don't need to focus on that too much.

wu-sheng:
Like I said in some other PRs, we don't require people to change their coding style, due to that is personal interests.

It is only acceptable when causing runtime performance issue.

dmsolr:
Yes. In this case, it is more importance to fix `size`. And I say that it is out of this issue. Just for talking.

Issue #3098, opened 2019-07-16
Change OAL script from compile level to runtime compile level


wu-sheng:
Hi

I am doing a challenge refactor work. Today, our oal script needs to rebuild/repackage the whole project for any update, which is not friendly to end-users, especially don't need to change docker image when OAL change.

I am going to keep as original design, OAL script should be performance-oriented and compile oriented. I wouldn't use any DSL(such as groovy) because of performance concern, rather than, I plan to use javaassist to generate metrics and dispatcher classes dynamically, and notify the core.

Fortunately, I am familiar with class bytecode things due to write the agent core. So, the dark magic thing will come to OAL too. 

@apache/skywalking-committers This PR will be large but could be guaranteed by e2e tests, also related to old OAL tool mostly. Also, the codes will become less readable as I am beginning to use java code manipulate tech, like this

```java
        CtClass parentMetricsClass = null;
        try {
            parentMetricsClass = classPool.get(METRICS_FUNCTION_PACKAGE + metricsStmt.getMetricsClassName());
        } catch (NotFoundException e) {
            logger.error("Can't find parent class for " + metricsStmt.getMetricsName() + ".", e);
            throw new OALCompileException(e.getMessage(), e);
        }
        CtClass metricsClass = classPool.makeClass(metricsClassName(metricsStmt), parentMetricsClass);
        try {
            metricsClass.addInterface(classPool.get(WITH_METADATA_INTERFACE));
        } catch (NotFoundException e) {
            logger.error("Can't find WithMetadata interface for " + metricsStmt.getMetricsName() + ".", e);
            throw new OALCompileException(e.getMessage(), e);
        }

        ClassFile metricsClassClassFile = metricsClass.getClassFile();
        ConstPool constPool = metricsClassClassFile.getConstPool();
```

PR #3109, opened 2019-07-18
All NEW OAL Runtime Engine

wu-sheng:
Hi @apache/skywalking-committers and all 

This PR brings all I talked in #3098, now
1. OAL scripts are loaded in bootstrap stage
1. OAL is still at compile level. But not by JDK, instead, by OALRuntime to generate binary codes. This is the black magic, I have tried my best to keep codes as simple as possible, and keep using freemaker templates for readability.
1. All metrics process flows are the same as before

I am going to add `metrics name` in the remote protocol after this PR merged, to replace the ID, because, at bootstrap level, OAL scripts may different somehow in different instances. We should use the literal string name in the internal distributed aggregation and print warning log when receiving unexpected metrics name. cc @peng-yongsheng


wu-sheng:
ASF Jenkins H35 keeps failing

wu-sheng:
@kezhenxu94 Delete workspace keeping failure.

PR #3115, opened 2019-07-18
Make #finish more stable in async APIs

wu-sheng:
Fix #3073, avoid the `#finish` called in the concurrency case. 

FYI @BFergerson We missed this scenario last time.


wu-sheng:
cc @qxo I am fixing the async API issue.

Issue #2975, opened 2019-07-01
SpringMVC plugin not support servlet-api-2.5


JoshShaw:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Bug
- SkyWalking 6.1.0
- OS CentOS
- JRE 1.8

- What happen?
A springMVC webapp project is running with a Web Container. In my case, it is running with Weblogic11g which is implemented based on servlet-api-2.5. I had review the source code of springmvc plugin and found out it is implemented based on servlet-api-3.0.1.
Therefore, when `org.apache.skywalking.apm.plugin.spring.mvc.commons.interceptor.AbstractMethodInterceptor.afterMethod ` calls `HttpServletResponse.getStatus()`, it will throw NoSuchMethodException.

wu-sheng:
I think we need a `witness class` to set two interceptors for different servlet-api versions.

candyleer:
A simple way is to add a judgement for `HttpServletResponse. getStatus ` if it is exists ,I can solve this.

wu-sheng:
> A simple way is to add a judgement for HttpServletResponse. getStatus if it is exists ,I can solve this.

What kind of judgement? Doing a refactor checker?

candyleer:
I mean just add a `if`,such as 
![image](https://user-images.githubusercontent.com/11530760/61388341-5269f480-a8ea-11e9-8e53-0179b28b4be2.png)
cause for servlet below version 3 just not have `getStatus` method which we used ,this method will effect `springMvc`,`tomcat`,`jetty`.

So in servlet2.5,we just ignore the status judge and stop span .

wu-sheng:
OK, that is a tricky thing to JVM, using a fast fail to acoid the issue. I guess it should work.

Please test it, if pass as expected, submit the PR.

wu-sheng:
Fixed.

PR #3120, opened 2019-07-19
Support servlet 2.5 for Jetty,Tomcat,SpringMvc

candyleer:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
#2975
___
### Bug fix
- Bug description.

- How to fix?

___


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/MethodUtil.java
```diff
@@ -42,4 +43,16 @@ public static String generateOperationName(Method method) {
         operationName.append(")");
         return operationName.toString();
     }
+
+
+    public static boolean isGetStatusExist(ClassLoader classLoader) {
```
wu-sheng: As an util, the class name should be parameter, right?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/MethodUtil.java
```diff
@@ -42,4 +43,21 @@ public static String generateOperationName(Method method) {
         operationName.append(")");
         return operationName.toString();
     }
+
+
+    public static boolean isMethodExist(ClassLoader classLoader, String className, String methodName, Class<?>... parameterTypes) {
+        try {
+            Class<?> httpServletResponse = Class.forName(className, true, classLoader);
```
wu-sheng: Wrong variable name

wu-sheng: Parameter type should be String if class name is string.

candyleer: Yes,I forgot update

PR #3068, opened 2019-07-13
improve ContextManager.stopSpan performance: call ThreadLocal only once

qxo:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [ ] New feature provided
- [ *] Improve performance

___
### New feature or improvement
For the [JMH benchmark test](https://gitlab.com/qxo1/Common-JMH-benchmark/-/jobs/251025992) :
On gitlab CI(java 8): call ThreadLocal once vs twice fast over 10%
My T480 java 8 : fast over 40%; java 11: fast over 50%;

[The JMH benchmark code ](https://github.com/qxo/Common-JMH-benchmark/blob/master/src/main/java/test/ThreadLocalGetXTimeBenchmark.java)

PR #3122, opened 2019-07-19
improve MethodsAroundInterceptor API:add MethodAroundContext over the Interceptor …

qxo:

## Why submit this pull request?
- New feature provided
- Improve performance

## Related issues
___
-  MethodsAroundInterceptor should have MethodAroundContext  over the Interceptor 
so we can pass current span,then we change `ContextManager#stopSpan` to `ContextManager#stopSpan(span)` -- for speed up and fail-fast
ref: https://github.com/apache/skywalking/pull/3068

static MethodsAroundInterceptor also need to be support

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/AbstractInstanceMethodsAroundInterceptor.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/AbstractInstanceMethodsAroundInterceptor.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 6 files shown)


wu-sheng:
Why do we need to change this? How much performance does it bring to the agents?

I am not just asking benchmark, I am asking about in the real scenario tests

qxo:
1. The change TraceAnnotationMethodInterceptor  in this pr is the use case 
2. After change of TraceAnnotationMethodInterceptor  the code more strong (fail-fast)
and it could improve the performance  if we grid of the invoke  for ThreadLocal 
It's another pr I want to do:)

wu-sheng:
I do know it is related to performance. My question is how much. 

Change API will break many things, including document etc. especially we have many plugins today. I have found any real performance issue by using ThreadLocal. Actually, most of plugins wouldn't use this.

qxo:
> I do know it is related to performance. My question is how much.
> 
> Change API will break many things, including document etc. especially we have many plugins today. I have found any real performance issue by using ThreadLocal. Actually, most of plugins wouldn't use this.

The change of API is for the future, and it's compatible with the exist plugin (  without change the plugin ).

wu-sheng:
All these changes, including #3124 , need to be proved necessary. Also need to have multiple PMC/committors support that. I would not make the call here.

You should know, as OSS project, we only add or change API when it is necessary, especially these are all open APIs. You are bringing the extra maintain payload to the community.

Personally, I would support this change.

-1 from me.

cc @apache/skywalking-committers if anyone who wants this, please `+1 explicitly. And explain why and what are the scenarios.

wu-sheng:
Ask committer and PMC, id they support, open issue to discuss. We could not change core API directly.

qxo:
I agree the API should not change if necessary
On my point of view, the API should  change  for the future 

This PR try to  make the API strong: pass span over MethodAroundContext 
 and then call `ContextManager#stopSpan(span)`   -- the fail-fast version of stopSpan

combine with #3124( "span carrry TracingContext" , reduce call ThreadLocal  if not required ) will improve  the performance  additionally

wu-sheng:
Yes, I agree the two PRs will improve the performance, but we need to be careful about that. SkyWalking today is making the project more stable by several version releases. So, if we noticed serious performance concerns, we prefer to keep unchanged. Also, we haven't received the report, our agent core(not perfect :) for sure) causes the performance issue to target applications.

PR #3128, opened 2019-07-20
OAP internal RemoteService protocol change and code refactor

wu-sheng:
Remote nextWorkerId, streamDataId in OAP internal RemoteService, instead, I add **nextWorkerName**. Because, as OAL scripts have been moved in runtime, most likely, OAL scripts may be different in different OAP instances, which are wrong. But in old codes, this error can't identify clearly, due to we use the bootstrap sequence as integer IDs to do aggregations across OAP instances. It is hard to tell why the data structure of metrics or inventory is not matching.

By using string, OAL variable alias, it is clear and easy unique at global, when OAL scripts are not same, OAP gRPC handler could give a very clear warning.
> logger.warn("Work name [{}] not found. Check OAL script, make sure they are same in the whole cluster.", nextWorkerName);


wu-sheng:
FYI @hanahmily this warning should be noticed.

Issue #2977, opened 2019-07-01
Improve backend operability


hanahmily:
There're some feedbacks from the community about how to monitor or debug online. So I create this issue to trace a series of PRs to improve operability.

 - [ ] self-monitoring, monitoring/alerting key metrics of OAP and Storage.
 - [ ] searching metadata(service, instance and endpoint) without duration, that help operator to find when/in which time duration the data is available.

If anyone has some thoughts, I'm open to your new ideas about this topic, thanks.

wu-sheng:
I think we have metrics for OAP self-observability. Which metrics are you proposing?

hanahmily:
Now I supported prometheus exporter format, for o11y available, we should deploy prometheus server with oap backend.
Using another o11y tool to observe a o11y tool is little wired, and especially their capability overlaps.

So my plan is fetching current self-metrcs + storage-metrics(es cpu,mem,gc etc) to oap receiver, then use graphql to query
a time series result.



Hongtao Gao

Apache SkyWalking && Apache ShardingSphere
Twitter, @hanahmily

wu-sheng:
So, you are proposing an ES storage implementation for metrics? and query extension for that?

wu-sheng:
Do you want to set up a new OAP to do that? Or just using the current OAP to storage and read?

hanahmily:
yes, I intend to extend ES storage to support new metrics that don't belong to OAL.

but current query schema is enough for those metrics, well, I’m going to proposal some new query
for metadata though.

hanahmily:
I hope they will be part of current OAP.

wu-sheng:
Please be advised, the telemetry is a very basic module, even core depends on it, so you can't call storage module services.

hanahmily:
Got it. That means if I invoke storage in telemetry, maybe there will be a cycle dependency, right?

So my current idea is to create a new receiver, that pull the metrics from telemetry through local network or api.

Can it work, or not?

wu-sheng:
@hanahmily I am moving this to 6.5 again, as it is not updated recently.

PR #3140, opened 2019-07-22
Self observability of OAP backend

hanahmily:
These changes related to #2977 

 * Adding new telemetry called `so11y` derived from prometheus, and a new service to collect metrics from the exporter.
 * Adding a new receiver to collect these metrics and generate metrics related to JVM.

### DIFF
--- a/docker/oap/docker-entrypoint.sh
+++ b/docker/oap/docker-entrypoint.sh
@@ -256,6 +256,8 @@ receiver-jvm:
   default:
 receiver-clr:
   default:
+receiver-so11y:
+  default:
 service-mesh:
   default:
     bufferPath: \${SW_SERVICE_MESH_BUFFER_PATH:../mesh-buffer/}  # Path to trace buffer files, suggest to use absolute path
@@ -270,9 +272,10 @@ query:
 alarm:
   default:
 telemetry:
-  prometheus:
-    host: \${SW_TELEMETRY_PROMETHEUS_HOST:0.0.0.0}
-    port: \${SW_TELEMETRY_PROMETHEUS_PORT:1234}
+  so11y:
+    prometheusExporterEnabled: \${SW_TELEMETRY_SO11Y_PROMETHEUS_ENABLED:true}
+    prometheusExporterHost: \${SW_TELEMETRY_PROMETHEUS_HOST:0.0.0.0}
+    prometheusExporterPort: \${SW_TELEMETRY_PROMETHEUS_PORT:1234}
 EOT
     # generate configuration
     case ${SW_CONFIGURATION} in
--- a/oap-server/server-starter/src/main/resources/application.yml
+++ b/oap-server/server-starter/src/main/resources/application.yml
@@ -63,28 +63,28 @@ core:
     dayMetricsDataTTL: ${SW_CORE_DAY_METRIC_DATA_TTL:45} # Unit is day
     monthMetricsDataTTL: ${SW_CORE_MONTH_METRIC_DATA_TTL:18} # Unit is month
 storage:
-#  elasticsearch:
-#    nameSpace: ${SW_NAMESPACE:""}
-#    clusterNodes: ${SW_STORAGE_ES_CLUSTER_NODES:localhost:9200}
-#    user: ${SW_ES_USER:""}
-#    password: ${SW_ES_PASSWORD:""}
-#    indexShardsNumber: ${SW_STORAGE_ES_INDEX_SHARDS_NUMBER:2}
-#    indexReplicasNumber: ${SW_STORAGE_ES_INDEX_REPLICAS_NUMBER:0}
-#    # Those data TTL settings will override the same settings in core module.
-#    recordDataTTL: ${SW_STORAGE_ES_RECORD_DATA_TTL:7} # Unit is day
-#    otherMetricsDataTTL: ${SW_STORAGE_ES_OTHER_METRIC_DATA_TTL:45} # Unit is day
-#    monthMetricsDataTTL: ${SW_STORAGE_ES_MONTH_METRIC_DATA_TTL:18} # Unit is month
-#    # Batch process setting, refer to https://www.elastic.co/guide/en/elasticsearch/client/java-api/5.5/java-docs-bulk-processor.html
-#    bulkActions: ${SW_STORAGE_ES_BULK_ACTIONS:1000} # Execute the bulk every 1000 requests
-#    flushInterval: ${SW_STORAGE_ES_FLUSH_INTERVAL:10} # flush the bulk every 10 seconds whatever the number of requests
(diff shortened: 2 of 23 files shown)


oap-server/server-starter/src/main/resources/application.yml
```diff
@@ -63,28 +63,28 @@ core:
...
+#    bulkActions: ${SW_STORAGE_ES_BULK_ACTIONS:1000} # Execute the bulk every 1000 requests
+#    flushInterval: ${SW_STORAGE_ES_FLUSH_INTERVAL:10} # flush the bulk every 10 seconds whatever the number of requests
+#    concurrentRequests: ${SW_STORAGE_ES_CONCURRENT_REQUESTS:2} # the number of concurrent requests
+#    metadataQueryMaxSize: ${SW_STORAGE_ES_QUERY_MAX_SIZE:5000}
+#    segmentQueryMaxSize: ${SW_STORAGE_ES_QUERY_SEGMENT_SIZE:200}
+  h2:
+    driver: ${SW_STORAGE_H2_DRIVER:org.h2.jdbcx.JdbcDataSource}
+    url: ${SW_STORAGE_H2_URL:jdbc:h2:mem:skywalking-oap-db}
+    user: ${SW_STORAGE_H2_USER:sa}
+    metadataQueryMaxSize: ${SW_STORAGE_H2_QUERY_MAX_SIZE:5000}
```
wu-sheng: You should not submit this change. Dev mode has been set in ES as storage, as 1st priority storange implementation today.

oap-server/server-telemetry/telemetry-api/src/main/java/org/apache/skywalking/oap/server/telemetry/api/Metrics.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Metrics define a metric and all its samples.
+ *
+ * @author gaohongtao
+ */
+@AllArgsConstructor
+@EqualsAndHashCode
+@ToString
+public class Metrics {
```
wu-sheng: This is `org.apache.skywalking.oap.server.core.analysis.metrics.Metrics` in core, could you consider rename it?

wu-sheng:
@hanahmily Your commit break e2e(single and cluster modes) somehow, could you check?

oap-server/server-starter/src/main/resources/application.yml
hanahmily: done

oap-server/server-telemetry/telemetry-api/src/main/java/org/apache/skywalking/oap/server/telemetry/api/Metrics.java
hanahmily: done. rename to `MetricFamily`

hanahmily:
> @hanahmily Does `So11Receiver` collect metrics from its own OAP metrics? Look like that way. Also, you still keep Prometheus port open, right?

Exactly. `So11Telemtry` provides an API to supply metrics and still listen to promethues port.

oap-server/server-telemetry/telemetry-so11y/src/main/java/org/apache/skywalking/oap/server/telemetry/so11y/So11yConfig.java
```diff
@@ -0,0 +1,38 @@
...
+import lombok.Getter;
+import lombok.Setter;
+import org.apache.skywalking.oap.server.library.module.ModuleConfig;
+
+/**
+ * The module configuration of self observability.
+ */
+@Setter
+@Getter
+public class So11yConfig extends ModuleConfig {
```
wu-sheng: According to `telemetry-prometheus`, this should extends the config from that module, right?

oap-server/server-starter/src/main/resources/application.yml
```diff
@@ -103,6 +103,8 @@ receiver-jvm:
   default:
 receiver-clr:
   default:
+receiver-so11y:
+  default:
```
wu-sheng: This receiver should be closed, due to `telemetry/so11y` closed.

oap-server/server-telemetry/telemetry-api/src/main/java/org/apache/skywalking/oap/server/telemetry/none/NoneTelemetryProvider.java
```diff
@@ -43,6 +44,7 @@
 
     @Override public void prepare() throws ServiceNotProvidedException, ModuleStartException {
         this.registerServiceImplementation(MetricsCreator.class, new MetricsCreatorNoop());
+        this.registerServiceImplementation(MetricsCollector.class, new MetricsCollectorNoop());
```
wu-sheng: Where is the change of `PrometheusTelemetryProvider`? I think this new service is also missing there, which will cause that module is unable to startup.

wu-sheng: From the codes, look like you are not going to abandon that module because you added a new one.

hanahmily:
> Besides the comments, you miss the document changes about receiver and backend-telemetry。

done.

oap-server/server-telemetry/telemetry-so11y/src/main/java/org/apache/skywalking/oap/server/telemetry/so11y/So11yConfig.java
hanahmily: No, `telemetry-so11y` is combining `telemetry-promtheus` instead of extending it.

oap-server/server-starter/src/main/resources/application.yml
hanahmily: done

oap-server/server-telemetry/telemetry-api/src/main/java/org/apache/skywalking/oap/server/telemetry/none/NoneTelemetryProvider.java
hanahmily: Exactly.
You could check out the docs I added to find it out.

docker/oap/docker-entrypoint.sh
```diff
@@ -256,6 +256,8 @@ receiver-jvm:
   default:
 receiver-clr:
   default:
+receiver-so11y:
```
peng-yongsheng: What's the meaning of **so11y**?

hanahmily: self(s) obervability(o11y)

Issue #3142, opened 2019-07-22
Java agent cannot trace gRPC stream service


aiyanbo:
- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement
___
### Bug
- Which version of SkyWalking, OS and JRE?
SkyWaling: 6.2.0
OS: Mac OS X 10.14.4
JRE: 1.8.0_181
gRPC: 1.22.1

- What happen?
When application invoke gRPC stream service failed, trace cannot work and log will output this:

- beforeMethod
```
java.lang.ClassCastException: org.apache.skywalking.apm.plugin.grpc.v1.StreamCallClientInterceptor cannot be cast to org.apache.skywalking.apm.plugin.grpc.v1.BlockingCallClientInterceptor
	at org.apache.skywalking.apm.plugin.grpc.v1.AsyncUnaryRequestCallCallInterceptor.beforeMethod(AsyncUnaryRequestCallCallInterceptor.java:39)
	at org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.StaticMethodsInterWithOverrideArgs.intercept(StaticMethodsInterWithOverrideArgs.java:75)
	at io.grpc.stub.ClientCalls.asyncUnaryRequestCall(ClientCalls.java)
	at io.grpc.stub.ClientCalls.asyncUnaryRequestCall(ClientCalls.java:267)
	at io.grpc.stub.ClientCalls.asyncServerStreamingCall(ClientCalls.java:75)
``` 

- afterMethod

```
java.lang.NullPointerException
	at org.apache.skywalking.apm.agent.core.context.ContextManager.activeSpan(ContextManager.java:161)
	at org.apache.skywalking.apm.agent.core.context.ContextManager.stopSpan(ContextManager.java:165)
	at org.apache.skywalking.apm.plugin.grpc.v1.AsyncUnaryRequestCallCallInterceptor.afterMethod(AsyncUnaryRequestCallCallInterceptor.java:50)
	at org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.StaticMethodsInterWithOverrideArgs.intercept(StaticMethodsInterWithOverrideArgs.java:96)
	at io.grpc.stub.ClientCalls.asyncUnaryRequestCall(ClientCalls.java)
	at io.grpc.stub.ClientCalls.asyncUnaryRequestCall(ClientCalls.java:267)
[...]

PR #3145, opened 2019-07-22
Fixed java agent cannot trace gRPC stream service

aiyanbo:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/3142
___
### Bug fix
- Bug description.
`org.apache.skywalking.apm.plugin.grpc.v1.GRPCClientInterceptor.interceptCall` returned the `org.apache.skywalking.apm.plugin.grpc.v1.StreamCallClientInterceptor` when gRPC method type is not `UNARY`, But `org.apache.skywalking.apm.plugin.grpc.v1.AsyncUnaryRequestCallCallInterceptor#L39` cast to `org.apache.skywalking.apm.plugin.grpc.v1.BlockingCallClientInterceptor`

- How to fix?
Extract interface `org.apache.skywalking.apm.plugin.grpc.v1.CallClientInterceptor` and `org.apache.skywalking.apm.plugin.grpc.v1.StreamCallClientInterceptor` & `org.apache.skywalking.apm.plugin.grpc.v1.BlockingCallClientInterceptor` implements it.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/AsyncUnaryRequestCallCallInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/AsyncUnaryRequestCallCallInterceptor.java
@@ -36,7 +36,7 @@ import static org.apache.skywalking.apm.plugin.grpc.v1.OperationNameFormatUtil.f
 public class AsyncUnaryRequestCallCallInterceptor implements StaticMethodsAroundInterceptor {
     @Override public void beforeMethod(Class clazz, Method method, Object[] allArguments, Class<?>[] parameterTypes,
         MethodInterceptResult result) {
-        BlockingCallClientInterceptor originClientCall = (BlockingCallClientInterceptor)allArguments[0];
+        CallClientInterceptor originClientCall = (CallClientInterceptor)allArguments[0];
         Channel channel = originClientCall.getChannel();
         MethodDescriptor methodDescriptor = originClientCall.getMethodDescriptor();
 
--- a/apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/StreamCallClientInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/StreamCallClientInterceptor.java
@@ -132,10 +132,10 @@ public class StreamCallClientInterceptor extends ForwardingClientCall.SimpleForw
                 } else {
                     AbstractSpan abstractSpan = ContextManager.createLocalSpan(operationPrefix + STREAM_RESPONSE_OBSERVER_ON_COMPLETE_OPERATION_NAME);
                 }
+                delegate().onClose(status, trailers);
                 if (Objects.nonNull(contextSnapshot)) {
                     ContextManager.continued(contextSnapshot);
                 }
-                delegate().onClose(status, trailers);
             } catch (Throwable t) {
                 ContextManager.activeSpan().errorOccurred().log(t);
             } finally {
--- a/apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/StreamCallClientInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/StreamCallClientInterceptor.java
@@ -127,7 +127,7 @@ public class StreamCallClientInterceptor extends ForwardingClientCall.SimpleForw
             try {
                 if (!status.isOk()) {
                     AbstractSpan abstractSpan = ContextManager.createLocalSpan(operationPrefix + STREAM_RESPONSE_OBSERVER_ON_ERROR_OPERATION_NAME);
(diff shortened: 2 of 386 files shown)


wu-sheng:
Is this still in WIP status?

aiyanbo:
@wu-sheng Yes, it still has a bug: When the server stream service has en error, it will be eat exception and the span will be success.

wu-sheng:
> Yes, it still has a bug: When the server stream service has en error, it will be eat exception and the span will be success.

Try `ContextManager#awaitFinishAsync` and hold the span ref, you could tag and finish it in another thread. That API did exist when gRPC plugin built.

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/StreamCallClientInterceptor.java
```diff
@@ -95,18 +95,25 @@ protected CallListener(Listener delegate, ContextSnapshot contextSnapshot) {
...
         }
 
-        @Override public void onMessage(Object message) {
+        @Override
+        public void onMessage(Object message) {
             try {
                 ContextManager.createLocalSpan(operationPrefix + STREAM_RESPONSE_OBSERVER_ON_NEXT_OPERATION_NAME);
                 ContextManager.continued(contextSnapshot);
+            } catch (Throwable t) {
+                // ignore trace error
```
wu-sheng: Why trace has error?

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/StreamCallClientInterceptor.java
```diff
@@ -124,8 +132,12 @@ protected CallListener(Listener delegate, ContextSnapshot contextSnapshot) {
                 } else {
                     AbstractSpan abstractSpan = ContextManager.createLocalSpan(operationPrefix + STREAM_RESPONSE_OBSERVER_ON_COMPLETE_OPERATION_NAME);
                 }
-                delegate().onClose(status, trailers);
                 ContextManager.continued(contextSnapshot);
+            } catch (Throwable t) {
+                // ignore trace error
```
wu-sheng: Same above.

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/StreamCallClientInterceptor.java
aiyanbo: ContextManager.continued has status check, It is maybe throws `IllegalArgumentException`, So using `try-catch` block avoid trace error break normal logic.

wu-sheng: No, you should not do that. You should check the `contextSnapshot != null`. Exception throw/catch is a heavy OP, it costs more CPU than you thought.

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/StreamCallClientInterceptor.java
```diff
@@ -81,18 +97,23 @@ protected CallListener(Listener delegate, ContextSnapshot contextSnapshot) {
...
             delegate().onHeaders(headers);
         }
 
-        @Override public void onMessage(Object message) {
+        @Override
+        public void onMessage(Object message) {
             try {
                 ContextManager.createLocalSpan(operationPrefix + STREAM_RESPONSE_OBSERVER_ON_NEXT_OPERATION_NAME);
-                ContextManager.continued(contextSnapshot);
+                if (Objects.nonNull(contextSnapshot)) {
```
wu-sheng: When the contextSnapshot is null?

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/AsyncUnaryRequestCallCallInterceptor.java
```diff
@@ -36,7 +36,7 @@
 public class AsyncUnaryRequestCallCallInterceptor implements StaticMethodsAroundInterceptor {
     @Override public void beforeMethod(Class clazz, Method method, Object[] allArguments, Class<?>[] parameterTypes,
         MethodInterceptResult result) {
-        BlockingCallClientInterceptor originClientCall = (BlockingCallClientInterceptor)allArguments[0];
+        CallClientInterceptor originClientCall = (CallClientInterceptor)allArguments[0];
```
wu-sheng: Is this the key of the fix? Because of the incompatible type cast?

aiyanbo: Yes

wu-sheng: @ascrutae Please review as you are the author.

Issue #2932, opened 2019-06-22
Try to make compatible update for adding supports of classes enhancement loaded by bootstrap classloader


wu-sheng:
In #2931 I used to plan to add **supports of classes enhancement loaded by bootstrap classloader**. But because of bootstrap classloader and application classloader conflicts, it looks like more complex than expected.

I will work with @ascrutae beginning with this demo codes, https://gist.github.com/ascrutae/30d27b55d191c2b38a07b68be4e56def.

Try to add this feature with a plugin(Maybe `HTTPUrlConnection` plugin) in 6.3.0. Anyone interested is welcome to join.

Right now, the idea is, try to use application classloader to run agent plugin interceptor codes in the bootstrap class interceptor.

I hope I could provide a very compatible and consistent way to write the plugin. I have a plan and will post here about the updates.

@candyleer @IanCao @zhaoyuguang @kezhenxu94 Welcome to join.

candyleer:
I think this feature implement should be:
1.seperate the `agent-core.jar` and `agent.jar`.
2.make the `agent-core.jar` loaded by bootstrap classloader ( cause the `agent-core.jar ` should be loaded by bootstrap classloader and `agent.jar` should be loaded by application classloader.).
3.Initialize the `Agent` by relfection in application classloader.
4.write some plugin to test

I am trying to implement the first version.

wu-sheng:
@candyleer I am writing the POC version, look like it is working :) But the side effect is because of classloader issue, the interceptor has been a static method, and using bytebuddy APIs(maybe could be improved, not sure). You could see my codes at `bootstrap-cl` branch.

POC is still underway. Welcome to join.

wu-sheng:
I have successfully added the dynamic field, but due to classloader issue(AppClassLoader and BootstrapClassLoader both loaded it), it can't be used in the interceptor, causing cast exception `Can't cast EnhancedInstance to EnhancedInstance. But still could use refactor to use if it is necessary, sometimes, I assume yes.

candyleer:
OK.

wu-sheng:
@candyleer Even separate the jar(s), you still have the classloader issue to fix, because you need to access the same `TracingContext` and `ContextManager` :)

wu-sheng:
@candyleer New codebase is there. Please review

SkyWalkingRobot:
[...]

PR #3152, opened 2019-07-23
Set up the bootstrap instrumentation plugin framework

wu-sheng:
Hi, All

This PR is following #2932, in these days, I work on building the bootstrap instrumentation plugin framework by not breaking any existing plugin. Look like it works.

For plugin developer, one line required only in plugin define.
```java
     // This is the only change for plugin developer
    @Override public boolean isBootstrapInstrumentation() {
        return true;
    }
```


wu-sheng:
If you want to understand how I do this, follow `BootstrapInstrumentBoost` class, it will link the things together. Also, plugin interceptor point definitions have been as public, as `BootstrapInstrumentBoost` needs to access it.

Also, I add a demo `jre-HttpUrlConnection` plugin for demo, and instrumen the following application codes.
```java
import java.io.BufferedReader;
import java.io.IOException;
import java.io.InputStreamReader;
import java.net.HttpURLConnection;
import java.net.URL;

public class Main {
    public static void main(String[] args) throws IOException {
        URL obj = new URL("http://www.baidu.com");
        HttpURLConnection con = (HttpURLConnection)obj.openConnection();

        con.setRequestMethod("GET");

        int responseCode = con.getResponseCode();
        System.out.println("Response Code : " + responseCode);

        BufferedReader in = new BufferedReader(
            new InputStreamReader(con.getInputStream()));
        String inputLine;
        StringBuffer response = new StringBuffer();

        while ((inputLine = in.readLine()) != null) {
            response.append(inputLine);
        }
        in.close();

        //print result
        System.out.println(response.toString());
    }
}
```

Like the existing plugins, the bootstrap intrumentation only requires you to declare it in plugin define, such as following codes I included in this PR
```java
/**
 * @author wusheng
 */
public class HttpUrlConnectionInstrumentation extends ClassEnhancePluginDefine {
    private static String CLASS_NAME = "java.net.HttpURLConnection";

     // This is the only change for plugin developer
    @Override public boolean isBootstrapInstrumentation() {
        return true;
    }

    @Override protected ClassMatch enhanceClass() {
        return byName(CLASS_NAME);
    }

    @Override public InstanceMethodsInterceptPoint[] getInstanceMethodsInterceptPoints() {
        return new InstanceMethodsInterceptPoint[] {
            new InstanceMethodsInterceptPoint() {
                @Override public ElementMatcher<MethodDescription> getMethodsMatcher() {
                    return named("setRequestMethod");
                }

                @Override public String getMethodsInterceptor() {
                    return "org.apache.skywalking.apm.plugin.jre.httpurlconnection.Interceptor";
                }

                @Override public boolean isOverrideArgs() {
                    return false;
                }
            }
        };
    }
}
```

Also, I haven't finished construct and static method plugin, but they are very similar, as we have this, they are simple. I will push those later.

I hope you could test this new mechanism in other demo codes, Spring Boot and Tomcat especially, in case I miss anything.

@ascrutae Please lead the tests.

wu-sheng:
This PR is ready to review and test. I have used it to instrument URL and HttpURLConnection successfully. But just a demo code, of course. 

Part of my test codes are posted here, https://gist.github.com/wu-sheng/25a8877971db089e196ada0fe4395813

In this PR, there is no real plugin. But further pull request is welcome.

candyleer:
I have tested in `bootstrap class` and `3rd part class`,both work well.
I have a question,the classes in `HIGH_PRIORITY_CLASSES` still will be loaded by two classloader,right? 

but I debug found, eg.`EnhancedInstance` is loaded by bootstrap classloader in bootstrap interceptor (generated by `InstanceMethodInterTemplate`),and  in normal `InstMethodsInter` it is loaded by application classloader .is it right? 

I think `EnhancedInstance` will be find in bootstrap classloader first .

wu-sheng:
> I have tested in bootstrap class and 3rd part class,both work well.
I have a question,the classes in HIGH_PRIORITY_CLASSES still will be loaded by two classloader,right?
but I debug found, eg.EnhancedInstance is loaded by bootstrap classloader in bootstrap interceptor (generated by InstanceMethodInterTemplate),and in normal InstMethodsInter it is loaded by application classloader .is it right?
I think EnhancedInstance will be find in bootstrap classloader first .

You do the tests quickly :) Thanks. 
To your question, from my understanding, when bootstrap instrumentation actived, EnhancedInstance should only be loaded in Bootstrap classloader. Otherwise, you will face `Class Cast Exception`, because you can't cat Bootstrap's EnhancedInstance to ApplicationClassLoader's EnhancedInstance.

Load the generated class to class loader w/o initialization should be safe. That is why I use `#injectRaw` rather than `#inject`.

This is my guess and understanding from test results, I used to try `#inject`, which cause that cast exception. :)

wu-sheng:
@candyleer If you are OK, please approve this PR.

kezhenxu94:
> @candyleer If you are OK, please approve this PR.

is this file (`debug/org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.bootstrap.Interceptor_internal.class`) committed by mistake?

wu-sheng:
> is this file (debug/org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.bootstrap.Interceptor_internal.class) committed by mistake?

Yes. Removing it.

wu-sheng:
@candyleer Have you tested Tomcat? And JDK 6?

candyleer:
> @candyleer Have you tested Tomcat? And JDK 6?

I have test Tomcat .but JDK 6 not test ,maybe tomorrow I can run a test manually.

candyleer:
It seems not support JDK6 for the error:
image

wu-sheng:
Look like javassist has an error.

candyleer:
I have tested ,javassist v3.21.0-GA is the last version support jdk6,and we can work well by using it

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/bootstrap/template/ConstructorInterTemplate.java
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @param obj target class instance.
+     * @param allArguments all constructor arguments
+     */
+    @RuntimeType
+    public static void intercept(@This Object obj,
+        @AllArguments Object[] allArguments) {
+        try {
+            EnhancedInstance targetObject = (EnhancedInstance)obj;
+
```
candyleer: should call `prepare()`

apm-sniffer/apm-agent-core/pom.xml
```diff
@@ -37,6 +38,7 @@
         <grpc.version>1.14.0</grpc.version>
         <guava.version>20.0</guava.version>
         <bytebuddy.version>1.9.2</bytebuddy.version>
+        <javaassist.version>3.23.2-GA</javaassist.version>
```
candyleer: 3.21.0-GA support JDK6

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/bootstrap/template/ConstructorInterTemplate.java
```diff
@@ -0,0 +1,85 @@
...
+        } catch (Throwable t) {
+            LOGGER.error("ConstructorInter failure.", t);
+        }
+    }
+
+    /**
+     * Prepare the context. Link to the agent core in AppClassLoader.
+     */
+    private static void prepare() {
+        if (INTERCEPTOR == null) {
```
candyleer: the `INTERCPETOR` Initialization shoud be synchronized?

apm-sniffer/apm-agent-core/pom.xml
wu-sheng: OK. I am doing downgrade

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/bootstrap/template/ConstructorInterTemplate.java
wu-sheng: Don't' have to. Create multiple instances is still safe. This could just avoid creating too many instances, and also avoid the lock.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/bootstrap/template/ConstructorInterTemplate.java
wu-sheng: Yes.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/bootstrap/template/ConstructorInterTemplate.java
candyleer: Yes

wu-sheng:
@candyleer Updated.

wu-sheng:
@ascrutae Please run our autotests, let's see whether this breaks anything.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/bootstrap/BootstrapPluginLogBridge.java
```diff
@@ -0,0 +1,91 @@
...
+
+import org.apache.skywalking.apm.agent.core.logging.api.ILog;
+import org.apache.skywalking.apm.agent.core.logging.api.LogManager;
+
+/**
+ * The log bridge makes the ILog accessible inside bootstrap classloader, especially for internal interceptor.
+ *
+ * @author wusheng
+ */
+public class BootstrapPluginLogBridge implements IBootstrapLog {
```
candyleer: the method in this class should be implemented, or the log will be `NoopLog`.

wu-sheng: Yes.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/bootstrap/BootstrapPluginLogBridge.java
```diff
@@ -0,0 +1,91 @@
...
+    private BootstrapPluginLogBridge(String clazz) {
+        logger = LogManager.getLogger(clazz);
+    }
+
+    @Override public void info(String format) {
+        logger.info(format);
+    }
+
+    @Override public void info(String format, Object... arguments) {
+        logger.info(format, arguments);
```
wu-sheng: @candyleer Could you test this log method? I am little concerned about arguments(Object...) could work or not. I haven't the test codes at my hand now.

wu-sheng: You could simple call log("", arg1, arg2) in InstanceMethodInterTemplate, and see whether log works well?

candyleer: OK ,I just tested info(String format)

wu-sheng: Make sure you use args list, no matter which method you test.

candyleer: Test passed

wu-sheng:
From the test report, it is same as before. After CI pass, I will merge this to the core. The new plugin is welcome, especially for HttpUrlConnection.

PR #3162, opened 2019-07-24
Simplify the PxxMetrics and ThermodynamicMetrics to improve performance

peng-yongsheng:

- [ ] Bug fix
- [ ] New feature provided
- [X] Improve performance

The dispatchers will create new objects for metrics, that is too much. so we need a light-weight object.


oap-server/oal-rt/src/main/resources/code-templates/metrics/deserialize.ftl
```diff
@@ -16,12 +16,12 @@ public void deserialize(org.apache.skywalking.oap.server.core.remote.grpc.proto.
     </#list>
 
     <#list serializeFields.intLongValuePairListFields as field>
-        setDetailGroup(new org.apache.skywalking.oap.server.core.analysis.metrics.IntKeyLongValueArray(30));
+        setDetailGroup(new org.apache.skywalking.oap.server.core.analysis.metrics.IntKeyLongValueHashMap(30));
```
wu-sheng: You need to change field names in analysis core.

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/parser/PersistenceColumns.java
```diff
@@ -43,7 +43,7 @@ public void addIntField(String fieldName) {
         intFields.add(new PersistenceField(fieldName));
     }
 
-    public void addIntLongValuePairelistField(String fieldName) {
+    public void addIntKeyLongValueHashMapField(String fieldName) {
         intLongValuePairListFields.add(new PersistenceField(fieldName));
```
wu-sheng: This field name should be changed too.

PR #3164, opened 2019-07-25
Revert parallelism in CI IT build

kezhenxu94:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

A recent build job shows that the `javadoc:javadoc` is causing issues, we don't host JavaDoc pages and this task is a waste of time/resources, so I'm just removing it

[image](https://builds.apache.org/job/skywalking-ci-it/641/flowGraphTable/)


wu-sheng:
We must build. Apache release policy. Doc must be built successfully.

wu-sheng:
@kezhenxu94 Check the error log, look like even build fails occasionally.

PR #3169, opened 2019-07-26
A type word fix

wu-sheng:
Incorrectly change by refactoring of IDEA.


kezhenxu94:
The e2e failed due to the missing `docker` command in Jenkins node H14, I'll ask the infra folks for help in Slack

wu-sheng:
Sure. This PR is easy, We could ignore for now.

PR #2409, opened 2019-03-27
Support agent reset command

ascrutae:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

PR #3170, opened 2019-07-26
Support upgrade backend w/o rebooting agents

### LINKED ISSUE #3072: Support upgrade backend w/o rebooting agents
# Support upgrade backend w/o rebooting agents
2019 - 07 - 14 
## Background
SkyWalking backend version upgrade usually includes persistence structure changes, due to performance improvements or new feature added. But, today, SkyWalking agent caches service id, service instance id, and endpoint mapping cache, all of these caches are designed for reducing network/storage payload/cost. So these have to be kept.
But, we realized the agents rebooting is a challenge to most end users. Because rebooting agents mean rebooting business applications, which is a limited operation in the OPS team.

## Design doc
https://docs.google.com/document/d/1mlOmydJyICuqwGveem260dJEWxAE-ssMOy8gzXm33Yo/edit?usp=sharing

## Committer
@kezhenxu94 

## Reviewer team
@wu-sheng @ascrutae @peng-yongsheng and others who have interests.

kezhenxu94:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

resolve #3072 

most of the codes are taken from #2409 , thanks to @ascrutae and hope you get some time to review this, thanks :)


kezhenxu94:
@wu-sheng I'm putting this into milestone 6.3.0 in case that I can finish the e2e test within a week, if that's not the case, I'll just move it into next version

wu-sheng:
Is this PR ready to review or just a PR?

kezhenxu94:
> Is this PR ready to review or just a PR?

It's ready to review, I'm preparing e2e tests

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/GRPCChannelManager.java
```diff
@@ -88,24 +88,22 @@ public void run() {
                 String server = "";
                 try {
                     int index = Math.abs(random.nextInt()) % grpcServers.size();
-                    if (index != selectedIdx) {
-                        selectedIdx = index;
+                    selectedIdx = index;
```
wu-sheng: Why removed this if?

oap-server/server-receiver-plugin/skywalking-register-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/register/provider/handler/v6/grpc/ServiceInstancePingServiceHandler.java
```diff
@@ -57,11 +61,15 @@ public ServiceInstancePingServiceHandler(ModuleManager moduleManager) {
         ServiceInstanceInventory serviceInstanceInventory = serviceInstanceInventoryCache.get(serviceInstanceId);
         if (Objects.nonNull(serviceInstanceInventory)) {
             serviceInventoryRegister.heartbeat(serviceInstanceInventory.getServiceId(), heartBeatTime);
+            responseObserver.onNext(Commands.getDefaultInstance());
         } else {
-            logger.warn("Can't found service by service instance id from cache, service instance id is: {}", serviceInstanceId);
+            logger.warn("Can't found service by service instance id from cache," +
+                " service instance id is: {}, will send a reset command to agent side", serviceInstanceId);
+            final Command resetCommand = new ServiceResetCommand(UUID.randomUUID().toString()).serialize().build();
```
wu-sheng: Let's send a Command ID generation method.

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/parser/SegmentParseV2.java
```diff
@@ -87,6 +92,12 @@ public boolean parse(BufferData<UpstreamSegment> bufferData, SegmentSource sourc
             }
             SegmentObject segmentObject = bufferData.getV2Segment();
 
+            final int serviceInstanceId = segmentObject.getServiceInstanceId();
+            if (serviceInstanceInventoryCache.get(serviceInstanceId) == null) {
+                logger.warn("Cannot recognize service instance id [{}] from cache, segment will be ignored", serviceInstanceId);
```
wu-sheng: We need this check in segment parser. Because data could come from file buffer.

wu-sheng: Also JVM and register services are check required. Otherwise NPE happens.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/GRPCChannelManager.java
kezhenxu94: Most of the components (such as `ServiceAndEndpointRegisterClient`) depend on the status change callback `org.apache.skywalking.apm.agent.core.remote.GRPCChannelListener#statusChanged` (triggered by `notify` line 106), if there is only one OAP node and it's restarting, the status would be `DISCONNECTED` and never change back to `CONNECTED`, all the components depends on the status won't work since then

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/commands/CommandService.java
```diff
@@ -0,0 +1,117 @@
...
+    public void receiveCommand(Commands commands) {
+        for (Command command : commands.getCommandsList()) {
+            try {
+                BaseCommand baseCommand = CommandDeserializer.deserialize(command);
+
+                if (isCommandExecuted(baseCommand)) {
+                    continue;
+                }
+
+                if (ServiceResetCommand.NAME.equals(baseCommand.getCommand())) {
```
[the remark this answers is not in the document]
kezhenxu94: I'm thinking that whether we should put every command executor as a `BootService` or not, it's fine if we have just a few commands, but if we have many commands, the command executors would explode accordingly and I'm not sure whether it's a good idea to register so many `BootService` once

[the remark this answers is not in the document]
wu-sheng: I am not saying every executor is a service. I am saying, service should be the only one to provide command executor. Think in this way, which command should execute by a random class in the agent? My guess is `should not`. Command means a change order, change means someone is controlling something, so the change should be done inside a controller, right?

The expectation from me is, service could register itself as one or several command executor.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/GRPCChannelManager.java
wu-sheng: OK. That is a bug. The reason was trying to this is, the gRPC will reconnect directly, so if the address has only one, should wait for gRPC to reconnect. We may need to change this code better.

kezhenxu94: Thanks for clarifying this, I try another way to avoid recreating a new channel every time (take the advantage of GRPC's auto-reconnectivity) when there is only one server, by checking the existed channel's state.

kezhenxu94:
Hi @wu-sheng , most of the review comments (if not all) are addressed, really appreciate the professional opinions. Hope I fully got your points, thanks again

oap-server/server-receiver-plugin/skywalking-register-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/register/provider/handler/v6/grpc/ServiceInstancePingServiceHandler.java
```diff
@@ -57,11 +61,21 @@ public ServiceInstancePingServiceHandler(ModuleManager moduleManager) {
...
         if (Objects.nonNull(serviceInstanceInventory)) {
             serviceInventoryRegister.heartbeat(serviceInstanceInventory.getServiceId(), heartBeatTime);
+            responseObserver.onNext(Commands.getDefaultInstance());
         } else {
-            logger.warn("Can't found service by service instance id from cache, service instance id is: {}", serviceInstanceId);
+            logger.warn("Can't found service by service instance id from cache," +
+                " service instance id is: {}, will send a reset command to agent side", serviceInstanceId);
+            final String serialNumber = generateSerialNumber(request);
+            final Command resetCommand = new ServiceResetCommand(serialNumber).serialize().build();
+            final Commands nextCommands = Commands.newBuilder().addCommands(resetCommand).build();
```
wu-sheng: Command should be generated in the server-core module, please add a new service there in `org.apache.skywalking.oap.server.core.command`.

oap-server/server-receiver-plugin/skywalking-register-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/register/provider/handler/v6/grpc/ServiceInstancePingServiceHandler.java
```diff
@@ -57,11 +61,21 @@ public ServiceInstancePingServiceHandler(ModuleManager moduleManager) {
...
+            responseObserver.onNext(nextCommands);
         }
 
-        responseObserver.onNext(Commands.getDefaultInstance());
         responseObserver.onCompleted();
     }
+
+    // Simply generate a uuid without taking care of the {@code request}
+    private String generateSerialNumber(final ServiceInstancePingPkg request) {
```
wu-sheng: Following above, this method should be moved into core too.

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/h2/dao/H2NetworkAddressInventoryCacheDAO.java
```diff
@@ -63,7 +63,7 @@ public H2NetworkAddressInventoryCacheDAO(JDBCHikariCPClient h2Client) {
                 try (ResultSet resultSet = h2Client.executeQuery(connection, sql.toString(), lastUpdateTime)) {
                     NetworkAddressInventory addressInventory;
                     do {
-                        addressInventory = (NetworkAddressInventory)toStorageData(resultSet, NetworkAddressInventory.INDEX_NAME, new ServiceInventory.Builder());
+                        addressInventory = (NetworkAddressInventory)toStorageData(resultSet, NetworkAddressInventory.INDEX_NAME, new NetworkAddressInventory.Builder());
```
wu-sheng: This bug should be merged quickly. Could you send another PR to fix?

kezhenxu94: > This bug should be merged quickly. Could you send another PR to fix?

Yes, I'll send a fix PR soon

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/GRPCChannelManager.java
kezhenxu94: Fixed in #3181 and I'll mark it as resolved

wu-sheng:
Have you finished all changes?

kezhenxu94:
> Have you finished all changes?

Yes, and I tested locally, passed

wu-sheng:
I will do code review this afternoon, I think we should add an e2e case, which runs OAP in standalone mode, and try to erase the storage, and reboot backend.

kezhenxu94:
> I will do code review this afternoon

thanks

> I think we should add an e2e case, which runs OAP in standalone mode, and try to erase the storage, and reboot backend.

Yes, I'm recently trying to find a way to restart a container(OAP) in unit test code, will add the test once I make it

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/parser/SegmentParseV2.java
```diff
@@ -87,6 +92,13 @@ public boolean parse(BufferData<UpstreamSegment> bufferData, SegmentSource sourc
             }
             SegmentObject segmentObject = bufferData.getV2Segment();
 
+            // Recheck in case that the segment comes from file buffer
+            final int serviceInstanceId = segmentObject.getServiceInstanceId();
+            if (serviceInstanceInventoryCache.get(serviceInstanceId) == null) {
+                logger.warn("Cannot recognize service instance id [{}] from cache, segment will be ignored", serviceInstanceId);
+                return false;
```
wu-sheng: If you want to ignore, you should return `true`. Check `DataStreamReader#read`

kezhenxu94:
Hi @wu-sheng , e2e test is added, let's wait for the result and if it passes (it passes locally) I'll "Update branch" to trigger it again to see the stability, I think this pr is ready now

wu-sheng:
One question, why the case is so quickly, AFAIK, the agent cooldown should be 10 mins in default.

I saw the `table: alarm_record does not exist` log, which could represent the rebooting, but the time doesn't make sense. Please verify again.

test/e2e/e2e-agent-reboot/pom.xml
```diff
@@ -84,6 +84,7 @@
                             <run>
                                 <env>
                                     <INSTRUMENTED_SERVICE>${project.build.finalName}.jar</INSTRUMENTED_SERVICE>
+                                    <INSTRUMENTED_SERVICE_OPTS>-DSW_AGENT_COOL_DOWN_THRESHOLD=1</INSTRUMENTED_SERVICE_OPTS>
```
wu-sheng: How this opt works? If you use `-D`, then it should be `skywalking.agent. cool_down_threshold`, right?

Your own document says so too
> `agent.cool_down_threshold `|How long should the agent wait (in minute) before re-registering to the OAP server after receiving reset command.|`10`|

wu-sheng:
@heyanlong @liuhaoyang @ascrutae Please notice this. The reboot command has been supported in OAP and Java agent. Please consider supporting this in PHP, .NetCore and Nodejs.

PR #3176, opened 2019-07-28
Remove javassist from agent dependency

wu-sheng:
Through @raphw helps in these days, I got the ideas of how to use ByteBuddy only to do bootstrap instrumentation. So, I remove the javassist from the agent side dependency, and from my tests, all features are still there.

New in this PR, add several byte buddy annotation classes(used by us) into the bootstrap class loader.

@candyleer As you are doing the instrumentation of jdk HTTP Connection, please verify.

FYI All @apache/skywalking-committers, javassist will be gone from the agent. Only depend on ByteBuddy again, at least we wouldn't need to have concerns about that, javassist was no longer supported in JDK6 for at least a year ago.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/bootstrap/BootstrapInstrumentBoost.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/bootstrap/BootstrapInstrumentBoost.java
@@ -26,22 +26,27 @@ import java.lang.instrument.Instrumentation;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import javassist.ClassPool;
-import javassist.CtClass;
-import javassist.CtField;
+import net.bytebuddy.ByteBuddy;
 import net.bytebuddy.agent.builder.AgentBuilder;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.dynamic.ClassFileLocator;
+import net.bytebuddy.dynamic.DynamicType;
 import net.bytebuddy.dynamic.loading.ClassInjector;
+import net.bytebuddy.pool.TypePool;
 import org.apache.skywalking.apm.agent.core.boot.AgentPackageNotFoundException;
 import org.apache.skywalking.apm.agent.core.boot.AgentPackagePath;
 import org.apache.skywalking.apm.agent.core.logging.api.ILog;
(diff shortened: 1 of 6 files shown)


candyleer:
wow， i will verify today

wu-sheng:
Any update? @candyleer

candyleer:
wait a moment ,I am verifying now

candyleer:
verified .works well

wu-sheng:
Thanks. Please approve and merge.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/bootstrap/BootstrapInstrumentBoost.java
```diff
@@ -160,29 +173,49 @@ private static boolean prepareJREInstrumentation(PluginFinder pluginFinder,
...
+            DynamicType.Unloaded interceptorType = new ByteBuddy()
+                .redefine(templateTypeDescription, ClassFileLocator.ForClassLoader.of(BootstrapInstrumentBoost.class.getClassLoader()))
+                .name(internalInterceptorName)
+                .field(named("TARGET_INTERCEPTOR")).value(methodsInterceptor)
+                .make();
+
+            classesTypeMap.put(internalInterceptorName, interceptorType.getBytes());
+
+            try {
+                File debuggingClassesRootPath = null;
```
candyleer: wirte debug class when Config.Agent.IS_OPEN_DEBUGGING_CLASS is true?

wu-sheng: Yes. It exists before, I just move it to support this new part.

candyleer: Yeah,I mean I think it is not necessary to write the class to disk every time and It seems duplicated in this place.
image

wu-sheng: I will remove the duplicated ones.

wu-sheng: I removed the duplicated file write and map#put. Please recheck, if it is OK now, please approve again.

PR #3178, opened 2019-07-28
Fix bug that {Minute,Hour}TTLCalculator use wrong time unit

kezhenxu94:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance


kezhenxu94:
> In today's ES storage, the ES Hour/Minute level metrics are deleted through day level TTL.

I know that, and the `EsHourTTLCalculator` truncates the date format to `yyyyMMdd`. But the calculation is wrong IMHO:

https://github.com/apache/skywalking/blob/5036e8334301322d49973005cd98513c17b0f28f/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/ttl/EsHourTTLCalculator.java#L30

For example, `timeBefore` 20190728 18:00:00 with `DataTTLConfig#hourMetricsDataTTL=3` should return `20190728`, but it actually returns `20190725`, which I think will keep more old data than expected

wu-sheng:
No, the design is `20190725`. The unit is **DAY** when we are talking about ES storage TTL. Read this, https://github.com/apache/skywalking/blob/master/docs/en/setup/backend/ttl.md#elasticsearch-6-storage-ttl

PR #3181, opened 2019-07-29
Fix bug: state not updated when GRPC auto reconnect to the same server

kezhenxu94:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

___
### Bug fix
- Bug description.
Most of the components (such as ServiceAndEndpointRegisterClient) depend on the status change callback org.apache.skywalking.apm.agent.core.remote.GRPCChannelListener#statusChanged (triggered by notify line 106), if there is only one OAP node and it's restarting, the status would be DISCONNECTED and never change back to CONNECTED, all the components depends on the status won't work since then

- How to fix?
avoid recreating a new channel every time (yet take the advantage of GRPC's auto-reconnectivity) when there is only one server, by checking the existed channel's state.

___

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/GRPCChannel.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/GRPCChannel.java
@@ -19,6 +19,7 @@
 package org.apache.skywalking.apm.agent.core.remote;
 
 import io.grpc.Channel;
+import io.grpc.ConnectivityState;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 import io.grpc.netty.NettyChannelBuilder;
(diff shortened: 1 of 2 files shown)


wu-sheng:
@kezhenxu94 there is a db inventory bug needs to fix.

kezhenxu94:
> @kezhenxu94 there is a db inventory bug needs to fix.

Do you mean this one?

https://github.com/apache/skywalking/blob/16b82365067632d291421aad0a6ac14fc025bcae/oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/h2/dao/H2NetworkAddressInventoryCacheDAO.java#L66

wu-sheng:
Yes. It is a bug, right?

kezhenxu94:
> Yes. It is a bug, right?

Yes it is, I'll send an individual pull request in case that #3170 can't go into milestone 6.3.0

wu-sheng:
I think #3170 most likely could be merged into 6.3.0. There are not tests need to be done. Should not be too hard.

PR #3183, opened 2019-07-29
Fix bug: NetworkAddressInventory uses wrong builder to convert data

kezhenxu94:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

NetworkAddressInventory uses wrong builder to convert data, in some cases it would throw NPE


wu-sheng:
@kezhenxu94 I think we miss this check in the e2e, right? So this bug flows in.

wu-sheng:
cc @peng-yongsheng

wu-sheng:
cc @peng-yongsheng

wu-sheng:
Originally from #3170

PR #3171, opened 2019-07-26
Add jdk http plugin

candyleer:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#2709
___
### Bug fix
- Bug description.

- How to fix?

___
### New feature or improvement
- Describe the details and related test reports.

PR #3194, opened 2019-07-31
Make bootstrap instrumentation available in JDK9 - 11

wu-sheng:
Hi @candyleer 

By following the discussion at https://github.com/raphw/byte-buddy/issues/697 and PR #3171, I have done some adjustments. Now from my local test, the bootstrap instrumentation is available in OpenJDK 8 - 11, due to 8 is available, I assume JDK 6/7 are working too.

The issue remains today that, JDK 12 is still an issue because UnSafe injection is not working in JDK12. I have left a message to @raphw, let's see what is the right move. 

I think we may need two way, which should be OK. But at least, it is better for most widely used JDK, especially 9 and 11.


candyleer:
Got it

wu-sheng:
`bootstrapJarTmp` folder and document related to it have been removed permanently, as we use unsafe injection in this PR.

candyleer:
LGTM support jdk 6-11

PR #3199, opened 2019-08-01
Make auto instrumentation works when Module definition existing

wu-sheng:
If the user has defined the `module-info.java`, the instrumentation doesn't work today. Even, through previous PRs, we could do instrumentation w/o `module-info.java` running in JDK 9 - 12.

I am refactoring codes first, then submit a question at ByteBuddy https://github.com/raphw/byte-buddy/issues/699. Let's see whether we could get a solution, and how to change codes.

BTW, Disruptor is also not working when `module-info.java`. Potentially, we need to remove the dependency. For now, I just remove the usage temporary at local and try to make ByteBuddy works first. 


wu-sheng:
Waiting for new ByteBuddy 1.10.1 release.

wu-sheng:
@candyleer You could read this PR now. And notice the document I added. I am waiting for @raphw 's new release and confirmation(Take a look at our discussion issue at ByteBuddy). If you want to try this, you need to install ByteBuddy master branch by yourself and change the dependency to 1.10.1-SNAPSHOT.

wu-sheng:
@candyleer 

I did following changes

1. Adopt the latest bytebudy changes(snapshot only), commit https://github.com/raphw/byte-buddy/commit/b60971156403ed8fba6124b8e424f0749917594a. 
2. Remove disruptor by ArrayBlockingQueue, because Disruptor can't work in both JDK 8(1) and 9-12, if we don't add `--add-modules jdk.unsupported`
3. Doc update. Remove following
~~Since Java 9, JDK introduced a new level of abstraction above packages, formally known as the Java Platform Module System (JPMS), or “Modules” for short.
When you are using this JDK feature, add `--add-modules jdk.unsupported` to the JVM argument, 
if your application hasn't require `jdk.unsupported`, such as log output showing `Caused by: java.lang.ClassNotFoundException: sun.misc.Unsafe`.~~

wu-sheng:
I just active CI again, because bytebuddy just did a release 1.10.1

wu-sheng:
@candyleer Please test and review. 

@apache/skywalking-committers This is ready to merge.

wu-sheng:
@candyleer Any update?

candyleer:
I test this branch in `java module` from jdk 9-12 ,tradition no module application from jdk6-12,all works well.amazing.

candyleer:
for the logger,I think we can use `logback` as the agent default logger instead of custom queue.

wu-sheng:
I still have concerns about binding a 3rd party logging framework. One side, too big and may be conflicts with existing system config. The other hand, such as log4j2 including disruptor too.

PR #3200, opened 2019-08-01
fixed for okhttp 4.0.x AsyncCall

qxo:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
fixed for okhttp 4.0.x AsyncCall
- How to fix?

___

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/okhttp-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/okhttp/v3/define/AsyncCallInstrumentation.java
+++ b/apm-sniffer/apm-sdk-plugin/okhttp-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/okhttp/v3/define/AsyncCallInstrumentation.java
@@ -49,7 +49,7 @@ public class AsyncCallInstrumentation extends ClassInstanceMethodsEnhancePluginD
         return new InstanceMethodsInterceptPoint[] {
             new InstanceMethodsInterceptPoint() {
                 @Override public ElementMatcher<MethodDescription> getMethodsMatcher() {
-                    return named("execute");
+                    return named("execute").or(ElementMatchers.named("run").and(ElementMatchers.not(ElementMatchers.isFinal())));
                 }
 
                 @Override public String getMethodsInterceptor() {


apm-sniffer/apm-sdk-plugin/okhttp-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/okhttp/v3/define/AsyncCallInstrumentation.java
```diff
@@ -49,7 +49,7 @@
         return new InstanceMethodsInterceptPoint[] {
             new InstanceMethodsInterceptPoint() {
                 @Override public ElementMatcher<MethodDescription> getMethodsMatcher() {
-                    return named("execute");
+                    return named("execute").or(ElementMatchers.named("run").and(ElementMatchers.not(ElementMatchers.isFinal())));
```
wu-sheng: Which method is `run` w/ `final`? Could you point out the source and reason?

qxo: https://github.com/square/okhttp/blob/02958768ed9f19f78fa3060ea22c8025c0232242/okhttp/src/main/java/okhttp3/RealCall.kt#L131
okhttp 4.0.x AyncCall without execute method, only have "run" without "final" ( okhttp 3.x run is final"

wu-sheng:
If you are talking about OKHttp 4, then you need this PR CI works, and add the test cases to https://github.com/SkyAPMTest/agent-auto-integration-testcases. We have cases in OKHttp 3.x, you need to add 4.x

wu-sheng:
Also, in most cases, plugins for different versions require viteness class for keeping code safe.

wu-sheng:
The response is required, otherwise, we have to close the pull request.

wu-sheng:
No response in 4 days, closing

qxo:
no time for respone this PR
welcome other maintainers continue this work:)

Issue #2641, opened 2019-05-09
Zipkin receiver wrongly deduced whether the spans are compressed (gzipped) or not


tmichnik1981:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [*] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS and JRE?

SkyWalking: 6.1.0
JRE: Jdk8
OS: win10

zipkin client libs:
io.zipkin.zipkin2:zipkin:2.11.10
io.zipkin.reporter2:zipkin-reporter:2.7.13

- Which company or project?
ING Bank Śląski SA 

- What happen?

Zipkin allows to decide whether we want to send compressed or uncopressed spans.

It turned out that  zipkin-receiver-plugin was not able to figure out that the spans were compressed.  In other words compressed data was interpreted as uncompressed which 
led to exception. 

```
 java.lang.IllegalArgumentException: Malformed reading List<Span> from json
	at zipkin2.internal.JsonCodec.exceptionReading(JsonCodec.java:240) ~[zipkin-2.9.1.jar:?]
	at zipkin2.internal.JsonCodec.readList(JsonCodec.java:155) ~[zipkin-2.9.1.jar:?]
	at zipkin2.codec.SpanBytesDecoder$3.decodeList(SpanBytesDecoder.java:107) ~[zipkin-2.9.1.jar:?]
	at zipkin2.codec.SpanBytesDecoder.decodeList(SpanBytesDecoder.java:151) ~[zipkin-2.9.1.jar:?]
	at zipkin2.codec.SpanBytesDecoder$3.decodeList(SpanBytesDecoder.java:118) ~[zipkin-2.9.1.jar:?]
	at org.apache.skywalking.oap.server.receiver.zipkin.handler.SpanProcessor.convert(SpanProcessor.java:58) ~[zipkin-receiver-plugin-6.1.0.jar:6.1.0]
	at org.apache.skywalking.oap.server.receiver.zipkin.handler.SpanV2JettyHandler.doPost(SpanV2JettyHandler.java:70) [zipkin-receiver-plugin-6.1.0.jar:6.1.0]
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:707) [javax.servlet-api-3.1.0.jar:3.1.0]
[...]

PR #3208, opened 2019-08-02
zipkin receive issue fix for Not in Gzip format and span.kind is null

credit2go-admin:
1. Correct HTTP Header to fix "Not in GZIP format" when receive zipkin data

2. Add default value to Span.kind to Server as per Span.kind maybe null.

- [ X] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/2641
___
### Bug fix
- Bug description.
We use nginx with opentracing lib to send zipkin format data to Skywalking APM.
We found that Not in GZIP format issue and the default value of span.kind is null that cause nullpoint exception.

- How to fix?
After reference src code  of zipkin (https://github.com/openzipkin/zipkin with tag 2.9.1) that is same version of current zipkin lib referenced.  We change to Header value to "Content-Encoding" to solved GZIP issue.
And we add default value of span.kind with SERVER, so that with SW_RECEIVER_ZIPKIN_NEED_ANALYSIS=false mode, zipkin data can be show in trace.
___

### DIFF
--- a/oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/analysis/transform/SegmentBuilder.java
+++ b/oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/analysis/transform/SegmentBuilder.java
@@ -145,7 +145,8 @@ public class SegmentBuilder {
         if (!isSegmentRoot && parentSegmentSpan != null) {
             spanBuilder.setParentSpanId(parentSegmentSpan.getSpanId());
         }
-        Span.Kind kind = span.kind();
+        //Add default Span.kind to SERVER in case span.kind is null
+        Span.Kind kind = span.kind() == null ? Span.Kind.SERVER : span.kind();
         String opName = Strings.isNullOrEmpty(span.name()) ? "-" : span.name();
         spanBuilder.setOperationName(opName);
         ClientSideSpan clientSideSpan;
--- a/oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/handler/SpanProcessor.java
+++ b/oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/handler/SpanProcessor.java
@@ -69,7 +69,7 @@ public class SpanProcessor {
     private InputStream getInputStream(HttpServletRequest request) throws IOException {
         InputStream requestInStream;
 
-        String headEncoding = request.getHeader("accept-encoding");
+        String headEncoding = request.getHeader("Content-Encoding");
         if (headEncoding != null && (headEncoding.indexOf("gzip") != -1)) {
             requestInStream = new GZIPInputStream(request.getInputStream());
         } else {
--- a/oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/trace/SpanForward.java
+++ b/oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/trace/SpanForward.java
@@ -73,7 +73,8 @@ public class SpanForward {
             }
 
             String spanName = span.name();
-            Span.Kind kind = span.kind();
+            //Add default Span.kind to SERVER in case span.kind is null
+            Span.Kind kind = span.kind() == null ? Span.Kind.SERVER : span.kind();
             switch (kind) {
                 case SERVER:
                 case CONSUMER:
--- a/oap-server/server-storage-plugin/storage-zipkin-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/zipkin/elasticsearch/ZipkinTraceQueryEsDAO.java
+++ b/oap-server/server-storage-plugin/storage-zipkin-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/zipkin/elasticsearch/ZipkinTraceQueryEsDAO.java
@@ -186,7 +186,8 @@ public class ZipkinTraceQueryEsDAO extends EsDAO implements ITraceQueryDAO {
             swSpan.setParentSpanId(-1);
             swSpan.setSegmentSpanId(span.id());
             swSpan.setSegmentId(span.id());
-            Span.Kind kind = span.kind();
+            //Add default Span.kind to SERVER in case span.kind is null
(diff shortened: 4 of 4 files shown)


oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/handler/SpanProcessor.java
```diff
@@ -69,7 +69,7 @@ void convert(ZipkinReceiverConfig config, SpanBytesDecoder decoder, HttpServletR
     private InputStream getInputStream(HttpServletRequest request) throws IOException {
         InputStream requestInStream;
 
-        String headEncoding = request.getHeader("accept-encoding");
+        String headEncoding = request.getHeader("Content-Encoding");
```
wu-sheng: @adriancole which is the right name?

oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/analysis/transform/SegmentBuilder.java
```diff
@@ -145,7 +145,8 @@ private void scanSpansFromRoot(SpanObjectV2.Builder parentSegmentSpan, Span pare
         if (!isSegmentRoot && parentSegmentSpan != null) {
             spanBuilder.setParentSpanId(parentSegmentSpan.getSpanId());
         }
-        Span.Kind kind = span.kind();
+        //Add default Span.kind to SERVER in case span.kind is null
+        Span.Kind kind = span.kind() == null ? Span.Kind.SERVER : span.kind();
```
wu-sheng: Why set server as default?

wu-sheng:
No analysis mode works. Because w/ unexpected kind missing, we are hard to analysis.

oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/handler/SpanProcessor.java
adriancole: content-encoding is for the incoming payload. there's no format defined for the response sent back to the client as we just send status 202. iotw, accept-encoding isn't that relevant.

oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/analysis/transform/SegmentBuilder.java
adriancole: that doesn't make sense to me.

credit2go-admin: span.kind() got null value. This cause null point exception.
If skip  span.kind is null value, no zipkin trace data will display. So we just force it to SERVER to continue handle zipkin data process.

credit2go-admin:
when we use nginx opentrace module ( https://github.com/opentracing-contrib/nginx-opentracing ) with No analysis mode. The span.Kind got null value. It cause zipkin data store failed. You may use docker image opentracing-contrib/nginx-opentracing:latest to check this issue.

oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/analysis/transform/SegmentBuilder.java
wu-sheng: It triggers NPE, yes. But the default value is wrong. If no kind, I would say it is just a local span.

oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/handler/SpanProcessor.java
wu-sheng: @adriancole I forgot where this code from about accept-encoding. But I remember I copied from somewhere.

So this change is right fix, right?

oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/analysis/transform/SegmentBuilder.java
adriancole: yep exactly. local span is lack of kind. kind is not a required field. all fields are not required except trace and span ID

oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/handler/SpanProcessor.java
adriancole: maybe rename headEncoding to requestEncoding to make clear, but yeah.

It is common bug to flip these

oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/analysis/transform/SegmentBuilder.java
wu-sheng: @credit2go-admin Could you change the variable name too?

oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/handler/SpanProcessor.java
wu-sheng: @credit2go-admin Could you change the variable name too?

wu-sheng:
@credit2go-admin Update? We need you to follow the review.

wu-sheng:
@credit2go-admin This PR has no update in 6 days, we need to follow the review, otherwise, I have to close this.

wu-sheng:
No update in half month. Have to close.

PR #3211, opened 2019-08-03
Increase test service latency to help it to be counted

kezhenxu94:
The current latency of the test service is too slow, causing it's not counted and
the query result is 0, this patch increases the latency intentionally by sleeping
1 second

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance


wu-sheng:
@kezhenxu94 Accepted. Please approve the PR.

PR #3213, opened 2019-08-03
apm-agent-core PluginDefine(skywalking-plugin.def) support String argument: add a feature for reuse some ClassEnhancePluginDefine

qxo:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___

apm-agent-core PluginDefine(skywalking-plugin.def) support String argument: add a feature for reuse some ClassEnhancePluginDefine, which cloud reduce duplicate code of plugins

such as EnhancedInstanceInstrumentation:
```
vertx-core-3.x.enhancedInstanceClasses=io.vertx.core.impl.EventLoopContext;io.vertx.core.impl.WorkerContext
```
or
```
okhttp-3.x.enhancedInstanceClasses=okhttp3:Request,Response
```
or
```
okhttp-3.x=org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstanceInstrumentation:okhttp3:Request,Response
```
@apache/skywalking-committers

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/PluginBootstrap.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/PluginBootstrap.java
@@ -66,12 +66,11 @@ public class PluginBootstrap {
         for (PluginDefine pluginDefine : pluginClassList) {
             try {
                 logger.debug("loading plugin class {}.", pluginDefine.getDefineClass());
-                AbstractClassEnhancePluginDefine plugin =
-                    (AbstractClassEnhancePluginDefine)Class.forName(pluginDefine.getDefineClass(),
-                        true,
-                        AgentClassLoader.getDefault())
-                        .newInstance();
-                plugins.add(plugin);
+                final Class<?> cls = Class.forName(pluginDefine.getDefineClass(),
+                    true, AgentClassLoader.getDefault());
+                final String arg = pluginDefine.getConstructorArgument();
+                final Object plugin =  arg == null ? cls.newInstance() : cls.getConstructor(String.class).newInstance(arg);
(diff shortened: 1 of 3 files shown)


wu-sheng:
I am not following. I need more clear use case.

wu-sheng:
We have intercept point definitions already. Why do you want to change API?

I noticed you submitted many pull requests, which are all pointing to change APIs. They don't make sense to me. All these things, could be done by current API easily.

wu-sheng:
You need to understand open source community API basic principles
1. Have clear usage, and as easy as possible.
1. Once it is set, don't change it unless have too.
1. Performance is just perspective, not has to be best. 
1. Changing API is mostly accepted when you are core maintainer,whcih means you could support the project in the future, because every new API means payload to core maintainer team. If everyone could add one, and leave, the core team is impossible to maintainer. So all changes must be in deeply discussion, not just for anyone's convenience.
1. Coding style is not the reason to change APIs. Feature and use case are.

qxo:
I agree these principles
But the change of the API is for the future ...

The goal of this pr try to improve code reuse -- 增加代码重用度，减少插件中的重复代码
skywalking-plugin.def如支持通过定义传入构建器参数，则只用少数几个通用的AbstractClassEnhancePluginDefine子类，然后各插件可通过声明式定义来实现相同功能，重用代码逻辑:)

wu-sheng:
Also, in many open source projects, including SkyWalking. We don't do design for the future, unless it is a real scenario which we can't support.

I rejected the changes because I don't see the necessary of new APIs. Intercept point APIs are lower the maintain payload when plugin core refactor. DI is better than hierarchy always. Also, how could you support hierarchy enhance? By interface or super class?

Open source APIs need to be consistent abd clear. Your design looks like for internal developers, try to make it easy.

PR #3216, opened 2019-08-05
GC time metrics should be total time.

wu-sheng:
Use `sum` to replace `longAvg` in GC time calculation. The `longAvg` OP makes the metrics meaning the avg GC time per report period.

This bug is reported from the community. FYI @apache/skywalking-committers


kezhenxu94:
the unit test `TraceSegmentServiceClientTest.testSendTraceSegmentWithException` is not stable after #3181 , it failed two times, I'll try to find out why it failed sometimes

wu-sheng:
I added a fix for another bug. Prometheus telemetry provider is down.

PR #3219, opened 2019-08-05
Log4j2 support AsyncRoot print tid

evanxuhe:

- [ ] New feature provided

___
As we know, many companys choose log4j2 because of  good performance using async.
Therefore it's a huge limitation for skywalking , which doesn't support log4j2 async print tid,
like a gateway app.

#### New feature
Log4j2 support AsyncRoot print tid 
Just refer to the idea from  Logback AsyncAppenderRoot

#### Test App
I have build a simple SpringMVC web app to test log4j2   **Root** and **AsyncRoot** individually.
And it works well as the photo
image

##### 1. Is it ok to delete 2 classes that intercepts  Log4jOutputAppender before? ↓
```Log4j2OutputAppenderActivation```  ```PrintTraceIdInterceptor```(code pasted to the end)
I am not quite sure why adds Log4j2OutputAppender, which seems redundant. And i notice it is designed as a static method, is that from the idea that avoids OOM ?
What's more I just delete these 2 files,  because all i need is the format method in TraceIdConverter, whose 1st param is the enhanced Log4jLogEvent that carrys tid info. I am not sure if it obeys your origin design. Waiting for your reply.

#### 2. How to design test?
I am not quite sure how to design the test scenerio for this. :)

```java
@Plugin(name = "TraceIdConverter", category = "Converter")
@ConverterKeys({"traceId"})
public class TraceIdConverter extends LogEventPatternConverter {

    /**
     * Constructs an instance of LoggingEventPatternConverter.
     *
     * @param name name of converter.
     * @param style CSS style for output.
     */
    protected TraceIdConverter(String name, String style) {
        super(name, style);
    }

    public static TraceIdConverter newInstance(String[] options) {
        return new TraceIdConverter("traceId", "traceId");
    }

    @Override
    public void format(LogEvent event, StringBuilder toAppendTo) {
        Log4j2OutputAppender.append(toAppendTo);
    }
}
```
```java
public class Log4j2OutputAppender {
    /**
     * As default, append "TID: N/A" to the output message,
     * if sky-walking agent in active mode, append the real traceId in the recent Context, if existed, or empty String.
     *
     * @param toAppendTo origin output message.
     */
    public static void append(StringBuilder toAppendTo) {
        toAppendTo.append("TID: N/A");
    }
}
```


evanxuhe:
What's more, I comment the filter rule about org.apache.logging. 
I noticed that agent use log4j2 too and relavant instrument is ignored. I simply comment this filter.
Is it OK?
```java
            .ignore(
                nameStartsWith("net.bytebuddy.")	                nameStartsWith("net.bytebuddy.")
                    .or(nameStartsWith("org.slf4j."))	                    .or(nameStartsWith("org.slf4j."))
                    .or(nameStartsWith("org.apache.logging."))	//                    .or(nameStartsWith("org.apache.logging."))
                    .or(nameStartsWith("org.groovy."))	                    .or(nameStartsWith("org.groovy."))
                    .or(nameContains("javassist"))	                    .or(nameContains("javassist"))
                    .or(nameContains(".asm."))	                    .or(nameContains(".asm."))
```

wu-sheng:
1. If you are doing the same thing and covered the old cases, it is OK to delete the existing one.
1. I noticed, there is no document update, is there any change about usage?
1. The agent doesn't use log4j or any other log framework.
1. Why do you comment `.or(nameStartsWith("org.apache.logging."))`? I don't think we are going to instrument log framework, do you?

evanxuhe:
1. OK
2. no change in usage
3. OK
4. Because I intercept  ```callAppenders``` in  the ```org.apache.logging.log4j.core.async.AsyncLoggerConfig ``` of log4j2 to achieve the goal. Therefore. I comment it  to  make instrument on org.apache.logging.log4j work. I am not sure why it is abondend before and wanna to make sure that comment will not affect other components work normally .:smile:

wu-sheng:
For (4), because log4j is complex, it sometimes causes issues in agent startup, so we ignore it before. I am not sure whether we will face the issue now. Which versions of log4j2 have you tested? I suggest you to do including some old versions.

As you said, log component plugin is very hard to test, which mostly is based on trust, we can't do the plugin tests as we did for others.

evanxuhe:
|  Version   | Root  |  AsyncRoot   |
|  ----  | ----  |----  |
| 2.12.0  | :heavy_check_mark: |  :x: |
| 2.11.1  | :heavy_check_mark: |  :x: |
| 2.11.0  | :heavy_check_mark: |  :heavy_check_mark: |
| 2.8.2  |  :heavy_check_mark: |  :heavy_check_mark: |
| 2.4  |  :heavy_check_mark: |  :heavy_check_mark: |
| 2.3  |  :heavy_check_mark: |  :heavy_check_mark: |
| 2.1  |  :heavy_check_mark: |  :heavy_check_mark: |
| 2.0+  |  :x: |  :x: |

1. What's more, for latest version 2.11.1+, async root will not work. I will find out reason lately
1. **While for most common cases, 2.1<=version<=2.11.0,root and asyncroot works well**
1. for 2.0.0, I think it's too old, without the help of log4j-jul(minimum version2.1.0), it cant integrate with spring log framework even. Therefore, it will throw a exception if use 2.0 in springboot app.

Finally, I paste the version distribution of log4j2 as base for consideration
image

wu-sheng:
That is another reason, we design the way, which doesn't instrument log framework internal class. It is unstable, but the log requires stable. 

Another optional you could consider, it is doing log public API instrumentation, such as `logger#debug`, add the trace id into the log message.

evanxuhe:
> Another optional you could consider, it is doing log public API instrumentation, such as `logger#debug`, add the trace id into the log message.

So do you mean, I cound try to intercpet another class, some common log API, which is not included in "org.apache.logging."? Ok I will try to find this solution. While, I doubt it may be not ideal as this solution cause it has to intercept common API rather than some relavant API, which more targeted and clear.

wu-sheng:
> not included in "org.apache.logging."

I am not talking about this. Just for the internal API interception, which you have found the incompatible issues.

wu-sheng:
Update? No update in 10 days.

evanxuhe:
Sorry  i am busy with work this  week because of personal business
 i will try to fix it  before next tuesday:）

evanxuhe:
I reopen a new PR  see https://github.com/apache/skywalking/pull/3284/commits
finally it can supports all version above 2.0

kezhenxu94:
moved to #3284

PR #3221, opened 2019-08-06
Add dependency licenses check

kezhenxu94:

- [x] Bug fix
- [x] New feature provided
- [ ] Improve performance


apm-dist/release-docs/NOTICE
```diff
@@ -892,3 +892,12 @@ the 'license' directory of the distribution file, for the license terms of the
...
 ------
+
+========================================================================
+Freemarker NOTICE
+========================================================================
+Apache FreeMarker
+Copyright 2015-2018 The Apache Software Foundation
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
```
wu-sheng: Apache notice does not need to add. We are in the same foundation, read the first lines of this file.

oap-server/server-starter/src/main/resources/official_analysis.oal
```diff
@@ -109,8 +109,8 @@ envoy_parent_connections_used = from(EnvoyInstanceMetric.value).filter(metricNam
...
-// disable(endpoint_relation_server_side);
-// disable(top_n_database_statement);
-// disable(zipkin_span);
-// disable(jaeger_span);
\ No newline at end of file
+disable(segment);
+disable(endpoint_relation_server_side);
+disable(top_n_database_statement);
+disable(zipkin_span);
+disable(jaeger_span);
```
wu-sheng: Keep commented.

tools/dependencies/check-LICENSE.sh
```diff
@@ -0,0 +1,32 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# List all modules(jars) that belong to the SkyWalking itself, these will be ignored
+# when checking the dependency licenses
+./mvnw -Pbackend -Dexec.executable='echo' -Dexec.args='${project.artifactId}-${project.version}.jar' exec:exec -q > self-modules.txt
```
wu-sheng: Why mvnw again?

tools/dependencies/check-LICENSE.sh
```diff
@@ -0,0 +1,32 @@
...
+
+ls dist/apache-skywalking-apm-bin/oap-libs > all-dependencies.txt
+
+# Exclude all self modules(jars) to generate all third-party dependencies
+grep -vf self-modules.txt all-dependencies.txt > third-party-dependencies.txt
+
+# Compare the third-party dependencies with known dependencies, expect that
+# all third-party dependencies are KNOWN and the exit code of the command is 0,
+# otherwise we should add its license to LICENSE file and add the dependency to known-dependencies.txt
+diff -U0 tools/dependencies/known-dependencies.txt third-party-dependencies.txt
```
wu-sheng: You need to consider order. Maybe different in different OS.

kezhenxu94: > You need to consider order. Maybe different in different OS.

I sort the dependencies when generating the files

tools/dependencies/check-LICENSE.sh
kezhenxu94: Not that we didn't package again here, we just list all the modules in the backend

PR #3225, opened 2019-08-07
Modify the thread pool of grpc server.

yantaowu:

- [ ] Bug fix
- [ ] New feature provided
- [✔️] Improve performance

- Related issues
Have discussed by email
___
### Bug fix
- Bug description.

- How to fix?

___


oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/grpc/GRPCServer.java
```diff
@@ -90,12 +93,26 @@ public String serverClassify() {
     @Override
     public void initialize() {
         InetSocketAddress address = new InetSocketAddress(host, port);
-
+        ArrayBlockingQueue blockingQueue = new ArrayBlockingQueue(10000);
+        ExecutorService executor = new ThreadPoolExecutor(500, 500, 60,
```
wu-sheng: All these parameters require configurable, and default values should be reasonable, such as calculated from CPU core number. Such as take a look at `MetricsAggregateWorker#L52`
```JAVA
BulkConsumePool.Creator creator = new BulkConsumePool.Creator(name, BulkConsumePool.Creator.recommendMaxSize() * 2, 20);
```

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/grpc/GRPCServer.java
```diff
@@ -90,12 +93,26 @@ public String serverClassify() {
...
     }
 
+    static class CustomRejectedExecutionHandler implements RejectedExecutionHandler {
+        private final AtomicInteger rejectNumber = new AtomicInteger(1);
+
+        @Override
+        public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
+            if (rejectNumber.getAndIncrement() == 100) {
+                logger.warn("Grpc server thread pool is full, rejecting the task");
+                rejectNumber.set(1);
```
wu-sheng: If the rejectNumber is just for logging, no need to keep it so accurate. A simple int w/ volatile is enough. `AtomicInteger` causes unnecessary CPU.

yantaowu: good idea

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/grpc/GRPCServer.java
```diff
@@ -90,12 +103,22 @@ public String serverClassify() {
     @Override
     public void initialize() {
         InetSocketAddress address = new InetSocketAddress(host, port);
-
+        ArrayBlockingQueue blockingQueue = new ArrayBlockingQueue(threadPoolQueueSize);
```
wu-sheng: Is this queue used by gRPC in default?

yantaowu: GRPC uses SyncronousQueue by defalult.

PR #3236, opened 2019-08-09
support elasticsearch address configuration with http scheme.

JaredTan95:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance


JaredTan95:
> Why put http here? Are you trying to support https?

yes

wu-sheng:
Put an optional config in es storage provider. This change breaks existing user config, but get little benifit.

Add a config named protocol, use http as default value.

wu-sheng:
Wait for e2e

Issue #3238, opened 2019-08-09
restart k8s oap cluster make some agents disconnect with oap


qlql489:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [x] Question or discussion
- [ ] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?
we set oap with k8s cluster use 3 nodes,  after access more agent  we want to increase to 5 nodes ,but  aflter restart oap service ,a lots of agent log show error, they can't connect to opa, aflter restart agent , it goes right,   the error  logs show:

WARN 2019-08-09 20:50:40:484 DataCarrier.DEFAULT.Consumser.0.Thread GRPCStreamServiceStatus :  Collector traceSegment service doesn't response in 100955 seconds.

when restart oap  we need to  restart all agents?   or  what is the safe way to expand oap server's number
___
### Bug
- Which version of SkyWalking, OS and JRE?
6.2.0
- Which company or project?

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.

___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

wu-sheng:
Could you test 6.3.0 first? There is a bug in 6.2.0-, fixed by @kezhenxu94

wu-sheng:
` Collector traceSegment service doesn't response ` also means, server-side is not active.

PR #3240, opened 2019-08-10
Improve agent grpc auto reconnect

mm23504570:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
Question about https://github.com/apache/skywalking/issues/3238. 
In k8s. We often use Service to support  The ip of pod has change when we rolling-update. So the grpc auto-connect not work.
- How to fix?
 We should call the channel.getState() with param true to make a connection to the server.  
___

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
@@ -100,6 +100,13 @@ public class Config {
          * after receiving reset command
          */
         public static int COOL_DOWN_THRESHOLD = 10;
+
+        /**
+         * Max wait grpc channel auto connect count for same channel.
+         * If count of check grpc channel status more than this number.
+         * The channel check will call channel.getState(true) to requestConnection.
+         */
+        public static long MAX_WAIT_CHANNEL_AUTO_CONNECT_COUNT = 1;
     }
 
     public static class Collector {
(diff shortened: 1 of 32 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -100,6 +100,13 @@
          * after receiving reset command
          */
         public static int COOL_DOWN_THRESHOLD = 10;
+
+        /**
+         * Max wait grpc channel auto connect count for same channel.
+         * If count of check grpc channel status more than this number.
+         * The channel check will call channel.getState(true) to requestConnection.
+         */
+        public static long MAX_WAIT_CHANNEL_AUTO_CONNECT_COUNT = 1;
```
wu-sheng: Default should be 10.

apm-sniffer/config/agent.config
```diff
@@ -46,3 +46,6 @@ logging.file_name=${SW_LOGGING_FILE_NAME:skywalking-api.log}
 
 # Logging level
 logging.level=${SW_LOGGING_LEVEL:DEBUG}
+
+# Max wait grpc channel auto connect count for same channel.
+agent.max_wait_channel_auto_connect_count=${SW_AGENT_MAX_WAIT_CHANNEL_AUTO_CONNECT_COUNT:1}
```
wu-sheng: Most users don't need this. Remove it.

wu-sheng:
FYI @mm23504570 Besides your issue, this PR fixed the reconnection issue when Nginx responded `go away`.

PR #3247, opened 2019-08-11
support tracing any jdbc by delegate DataSource

qxo:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___
  support tracing any jdbc by delegate DataSource
    require getUrl method in target DataSource
    ie: subclass of org.springframework.jdbc.datasource.DelegatingDataSource add getUrl
    ```
    import javax.sql.DataSource;
    
    import org.springframework.jdbc.datasource.DelegatingDataSource;
    
    import net.zoneland.jdbc.ISmartDataSourceConfig;
    
    public class CustomDelegatingDataSource extends DelegatingDataSource {
    
        private String url;
    
        public void setUrl(String url) {
            this.url = url;
        }
    
        public CustomDelegatingDataSource(DataSource targetDataSource) {
            super(targetDataSource);
        }
    
        public final String getUrl() {
            return url;
        }
    }
    
    ```
    then:
    1. app using delegate DataSource, ie:CustomDelegatingDataSource
    2. start app using jvm properties: `-Dskywalking_datasource_intercept_class=xxx.CustomDelegatingDataSource`
    
    instrument origin DataSource (ie: druid) may not work, cause: DataSourceInstrumentation will wrap the #getConnection, and inside connection pool way using the origin connection
    so we instrument  delegated DtaSource


wu-sheng:
Why we do this? Targeting real driver is much better. You do not require application level change, unless necessary. 

I could share with you, tracing any jdbc is not working. We havr tried, but fail. Because there are many jdbc object type casts in real world. Delegation breaks the type cast

qxo:
It's work for me on oracle database
And I found there no plugin for oracle jdbc

wu-sheng:
We have oracle plugin list in doc. It is not an Apache plugin, because of license. Many oracle cases don't fit, we have faced before.

qxo:
where is it ? 
https://github.com/SkyAPM/java-plugin-extensions/blob/master/oracle-10.x-plugin  ?

wu-sheng:
Yes. For some case. But for your PR, I just want you know, this delegation mode isn't working for many use cases.

PR #3250, opened 2019-08-12
Enhance agent logger.

### LINKED ISSUE #3231: Support configurable agent-config to config logging format and output where

- [x] Question or discussion
- [ ] Bug
- [ ] Requirement

___
### Question
- What do you want to know?
I wish config log format and always output to system.out.   Not output a file. 

___
### Bug
6.3.0. k8s
- Which company or project?
ShuiDi Company
- What happen?

___
- Can we provider the configurable entry?  like logback or log4j2?

mm23504570:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
Resolves #3231

___
- Configurable logger format with logger.pattern 
- Configurable where output with logger.dir


wu-sheng:
You need to make CI(s) and e2e passed. Then others will begin to review.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/WriterFactory.java
```diff
@@ -40,4 +40,8 @@ public static IWriter getLogWriter() {
             return SystemOutWriter.INSTANCE;
         }
     }
+
+    private static boolean useSystemOut() {
+        return "system.out".equals(Config.Logging.DIR);
```
wu-sheng: This is not a good one. `Config.Logging.Output` should be a better choice, value options are `file` and `console`

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/PatternLogger.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.skywalking.apm.util.StringUtil;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Properties;
+
+/**
+ * @author alvin
+ */
+public class PatternLogger extends EasyLogger {
```
wu-sheng: Why do we need two? Do you have performance concern? Because I have for the pattern mode.

One further question, who will be interested in setting this pattern?

mm23504570: 1. I think the EasyLogger is a default choice. The default is concat Strings.
String format(LogLevel level, String message, Throwable t) {
        return StringUtil.join(' ', level.name(),
            new SimpleDateFormat("yyyy-MM-dd HH:mm:ss:SSS").format(new Date()),
            Thread.currentThread().getName(),
            targetClass,
            ": ",
            message,
            t == null ? "" : format(t)
        );
    }
```
It has hard to changeing the order. You mentioned pattern means '{}'? But how can i change the above the order?
Use `%{level} %{timestamp} %{thread} %{class} : %{msg} %{throwable}` this method to express the format of the log is more clear.

2. 
* If we use system.out to println log. It will be print together with agented Progress. Some log-collector collect log need use same logformat.
* More flexible. We can log something which we want.

File: apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/WriterFactory.java (hunk 1 above)
Comment (mm23504570) [reply apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/WriterFactory.java]: OK. Got it.

File: apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/PatternLogger.java [hunk 3]
```diff
@@ -0,0 +1,77 @@
...
+        this.pattern = pattern;
+    }
+
+    @Override
+    String format(LogLevel level, String message, Throwable t) {
+        Properties props = buildContext(level, message, t);
+        return PropertyPlaceholderHelper.LOGGER.replacePlaceholders(getPattern(), props);
+    }
+
+    private Properties buildContext(LogLevel level, String message, Throwable t) {
```
wu-sheng: You should test performance by the benchmark. @kezhenxu94 Do you remember a PR to give an example?

kezhenxu94: Yes, @mm23504570 please refer to https://github.com/apache/skywalking/pull/2766#issuecomment-496240702 for how to write a benchmark show the performance result

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/WriterFactory.java
```diff
@@ -27,7 +27,7 @@
 
 public class WriterFactory {
     public static IWriter getLogWriter() {
-        if (SnifferConfigInitializer.isInitCompleted() && AgentPackagePath.isPathFound()) {
+        if (SnifferConfigInitializer.isInitCompleted() && AgentPackagePath.isPathFound() && !useConsole()) {
```
wu-sheng: Static config, `!useConsole()`, should be first.

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -82,7 +82,9 @@ property key | Description | Default |
 `collector.backend_service`|Collector SkyWalking trace receiver service addresses.|`127.0.0.1:11800`|
 `logging.level`|The log level. Default is debug.|`DEBUG`|
 `logging.file_name`|Log file name.|`skywalking-api.log`|
+`logging.output`| Log output. Default is FILE. Use CONSOLE means output to stdout. |`FILE`|
 `logging.dir`|Log files directory. Default is blank string, means, use "system.out" to output logs.|`""`|
+`logging.pattern `|logging format. Default is blank string, means, use EasyLogger who pattern is `%level %timestamp %thread %class : %msg %throwable` |`""`|
```
wu-sheng: Add document about pattern options. Are these(%level %timestamp %thread %class : %msg %throwable) all options? If so, make it clear in the document.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -169,6 +170,16 @@
...
+
+        /**
+         * The log output. Default is FILE.
+         */
+        public static LogOutput OUTPUT = LogOutput.FILE;
+
+        /**
+         * The log patten. Default is "", means use EasyLogger.
+         */
+        public static String PATTERN = "";
```
wu-sheng: Default should be the pattern in the document.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/PatternLogger.java
```diff
@@ -1,51 +1,86 @@
 /*
- * Licensed to the Apache Software Foundation (ASF) under one or more
- * contributor license agreements.  See the NOTICE file distributed with
- * this work for additional information regarding copyright ownership.
- * The ASF licenses this file to You under the Apache License, Version 2.0
- * (the "License"); you may not use this file except in compliance with
- * the License.  You may obtain a copy of the License at
+ * Copyright 2011 the original author or authors.
```
wu-sheng: The license is not right. This should be Apache foundation header.

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -82,7 +82,9 @@ property key | Description | Default |
 `collector.backend_service`|Collector SkyWalking trace receiver service addresses.|`127.0.0.1:11800`|
 `logging.level`|The log level. Default is debug.|`DEBUG`|
 `logging.file_name`|Log file name.|`skywalking-api.log`|
+`logging.output`| Log output. Default is FILE. Use CONSOLE means output to stdout. |`FILE`|
 `logging.dir`|Log files directory. Default is blank string, means, use "system.out" to output logs.|`""`|
+`logging.pattern `|logging format. Default is blank string, means, use EasyLogger who pattern is `%level %timestamp %thread %class : %msg %throwable` . There are `%level %timestamp %thread %class %msg %throwable %agent_name` converters.  |`""`|
```
wu-sheng: Default is not empty. You have default value in Config class, right?

wu-sheng: There is no EasyLogger

wu-sheng: We need the explanations of each converter here too.

apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/logging/core/PatternLoggerTest.java
```diff
@@ -1,59 +1,67 @@
 /*
- * Licensed to the Apache Software Foundation (ASF) under one or more
- * contributor license agreements.  See the NOTICE file distributed with
- * this work for additional information regarding copyright ownership.
- * The ASF licenses this file to You under the Apache License, Version 2.0
- * (the "License"); you may not use this file except in compliance with
- * the License.  You may obtain a copy of the License at
+ * Copyright 2011 the original author or authors.
```
wu-sheng: The license header is still not right. @kezhenxu94 Look like our class header check is unavailable to test? Becuase CI passed in this definitely wrong case.

apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/logging/core/WriterFactoryTest.java
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright 2011 the original author or authors.
```
wu-sheng: Same wrong header too.

apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/logging/core/PatternLoggerTest.java
kezhenxu94: > The license header is still not right. @kezhenxu94 Look like our class header check is unavailable to test? Becuase CI passed in this definitely wrong case.

I noticed that this morning. It is the `apache-rat` plugin that is responsible to check the license header, but it didn't complain about anything, and I searched around to find that this form of license header is valid in some projects (that are not ASF TLP).

I'll try to make the check strictly, by using the checkstyle header, which is obviously not working now

wu-sheng: rat is just checking the Apache acceptable header, so this one is good.

But as Apache source codes, I think we need our header style check to make it strictly. Thanks for taking care of this. Expect your PR.

Issue #3252, opened 2019-08-13
ContextManager.activeSpan NullPointerException 


yangxb2010000:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS and JRE?
V6.2.0
- Which company or project?

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.
Reproducing Steps:
1. when more than the config **agent.span_limit_per_segment** spans were created,  NoopExitSpan will be used for following remote call.
2. call CreateExitSpan and StopSpan in multi Interceptors for one remote call operation. For Example: Java grpc Agent(AsyncUnaryRequestCallCallInterceptor and BlockingCallInterceptor will intercept grpc client call twice.)
3. the last StopSpan operation in this tracesegment will throw Exception, as shown:
```bash
ERROR io.grpc.internal.SerializingExecutor - Exception while executing runnable io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1HalfClosed@5de15ec
java.lang.NullPointerException: null
  at org.apache.skywalking.apm.agent.core.context.ContextManager.activeSpan(ContextManager.java:161)
  at org.apache.skywalking.apm.agent.core.context.ContextManager.stopSpan(ContextManager.java:165)
  at org.apache.skywalking.apm.plugin.babyfs.grpc.v1.CallServerInterceptor$ServerCallListener.onHalfClose(CallServerInterceptor.java:173)
  at io.grpc.internal.ServerCallImpl$ServerStreamListenerImpl.halfClosed(ServerCallImpl.java:283)
  at io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1HalfClosed.runInContext(ServerImpl.java:710)
  at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
  at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:123)
  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
  at java.lang.Thread.run(Thread.java:745)
```

___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

PR #3253, opened 2019-08-13
Master fix noop exit span bug

yangxb2010000:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/3252
___
### Bug fix
- Bug description.
when Call CreateExitSpan And StopSpan twice in one remote call. In fact, only one NoopExitSpan will be created. but stopSpan twice will pop two spans. So the last stopSpan operation will throw NullPointerException
- How to fix?
when stop NoopExitSpan, check finish before pop.
___


wu-sheng:
If stop twice, that is a bug. This fix is not right. You need to figure out whether
1. Should create two noop span

Or
2. Should stop span once.

wu-sheng:
`StackBasedTracingSpan` is only for performance consideration when tracing is active.

PR #3243, opened 2019-08-10
Webflux plugin optimization

zhaoyuguang:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/3242

### New feature or improvement

Support for more versions
Support for @controller and Functional endpoints
More complete collection of links in high concurrent scenarios

PR #3257, opened 2019-08-13
V6.4.0 snapshot fix noop exit span bug

yangxb2010000:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
when Call CreateExitSpan And StopSpan twice in one remote call. In fact, only one NoopExitSpan will be created. but stopSpan twice will pop two spans. So the last stopSpan operation will throw NullPointerException
- How to fix?
check isLimitMechanismWorking first, if true, create NoopExitSpan, push to stack and return.
___


wu-sheng:
Ref #3243 3252

Issue #3256, opened 2019-08-13
How can I configure ACK in zookeeper-cluster


xiangyanggong:
Please answer these questions before submitting your issue.

___
### Question
In my Env, ZK was set ACL, if I don't configure it, I can get this error： "KeeperErrorcode=NoAuth for /skywalking" when starting skywalking. I checked document, but I can't find corresponding entry in:

zookeeper:
  nameSpace: ${SW_NAMESPACE:""}
  hostPort: ${SW_CLUSTER_ZK_HOST_PORT:localhost:2181}
  #Retry Policy
  baseSleepTimeMs: ${SW_CLUSTER_ZK_SLEEP_TIME:1000} # initial amount of time to wait between retries
  maxRetries: ${SW_CLUSTER_ZK_MAX_RETRIES:3} # max number of times to retry
  internalComHost: 172.10.4.10
  internalComPort: 11800

### Bug
- Which version of SkyWalking, OS and JRE?
V6.2.0

dmsolr:
As I know, we don't support this now.

wu-sheng:
We don't support. If you want, submit pull request w/ explanation.

PR #3273, opened 2019-08-15
Support Zookeeper ACL

dmsolr:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#3256 

Support Cluster-Zookeeper-Plugin config ACL.


oap-server/server-cluster-plugin/cluster-zookeeper-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/zookeeper/ClusterModuleZookeeperProvider.java
```diff
@@ -61,7 +79,39 @@ public ClusterModuleZookeeperProvider() {
...
-        client = CuratorFrameworkFactory.newClient(config.getHostPort(), retryPolicy);
+
+        CuratorFrameworkFactory.Builder builder = CuratorFrameworkFactory.builder()
+            .retryPolicy(retryPolicy)
+            .connectString(config.getHostPort());
+        if (config.isEnableACL()) {
+            final List<ACL> acls = Lists.newArrayList();
+
+            String authInfo = config.getAuth();
+            if ("digest".equals(config.getSchema())) {
```
wu-sheng: I am not familiar w/ ACL, but why is `digest` schema special?

dmsolr: When used in the ACL the expression will be the username:base64 encoded SHA1 password digest. But when used in auth the expression must not encode.

The other schemas, such as `IP`, it doesn't have this problem.

oap-server/server-cluster-plugin/cluster-zookeeper-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/zookeeper/ClusterModuleZookeeperProvider.java
```diff
@@ -61,7 +79,39 @@ public ClusterModuleZookeeperProvider() {
...
+                try {
+                    authInfo = DigestAuthenticationProvider.generateDigest(authInfo);
+                } catch (NoSuchAlgorithmException e) {
+                    logger.error(e.getMessage(), e);
```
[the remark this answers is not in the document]
dmsolr: Yes, i will update later.

oap-server/server-cluster-plugin/cluster-zookeeper-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/zookeeper/ClusterModuleZookeeperProvider.java
wu-sheng: Is this the official recommended way? Look like strange that, a project like Zookeeper providing such strange style of API.

dmsolr: It's not good. I will try others.

dmsolr: The Zookeeper Document said:
> digest uses a username:password string to generate MD5 hash which is then used as an ACL ID identity. Authentication is done by sending the username:password in clear text. When used in the ACL the expression will be the username:base64 encoded SHA1 password digest.

Most projects only provided `digest` schema, such as Solr.

wu-sheng:
What is `ip` schema? Just a literal random String? @kezhenxu94

kezhenxu94:
> What is `ip` schema? Just a literal random String? @kezhenxu94

`ip` schema designates which ip (or ip range) can access the nodes, but in my test, if I pre-create the nodes (say `/skywalking/remote`) and set the ACL to `ip` of my machine (in zookeeper side), and set the `enableACL` to `true` and `schema` to `ip` (in SkyWalking side), it failed to connect to zk, after changing the `enableACL` to `false`, it works, which I think is a little strange

dmsolr:
We use Zookeeper ACL that has 2 steps, 
1. set ACL for `zNode`.
2. authorization, like `login`.

`IP` schema, we need not `authorization` step. But I will do those steps in this pr. @kezhenxu94

dmsolr:
I think whether we support `digest` only. Because `digest` is more widely used, and it is more clear.

dmsolr:
How do you think? @wu-sheng @kezhenxu94

wu-sheng:
@dmsolr One mode is good for me. Make sure it is clear at document.

PR #3274, opened 2019-08-15
Check license header strictly

kezhenxu94:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

The `apache-rat` plugin checks the license headers, but it checks them with a loose manner, there is scenario like https://github.com/apache/skywalking/pull/3250#discussion_r313712568 that is not valid for ASF projects, this patch will check the license headers more strictly by using the header checker module of `checkstyle` plugin


apm-checkstyle/checkStyle.xml
```diff
@@ -110,4 +110,11 @@
             <property name="path" value="apm-sniffer/(apm-sdk-plugin|optional-plugins)/.+/src/main/.+Instrumentation.java$" />
         </module>
     </module>
+
+    <module name="RegexpHeader">
+        <property name="headerFile" value="${checkstyle.header.file}"/>
+        <property name="multiLines" value="16"/>
+        <property name="fileExtensions" value="java"/>
```
wu-sheng: I think header check is required in shell, bat, yaml, xml, etc.

apm-checkstyle/checkStyle.xml
```diff
@@ -110,4 +110,11 @@
             <property name="path" value="apm-sniffer/(apm-sdk-plugin|optional-plugins)/.+/src/main/.+Instrumentation.java$" />
         </module>
     </module>
+
+    <module name="RegexpHeader">
```
kezhenxu94: We're using `RegexpHeader` checker module here because the existed license headers vary, some of them have more spaces between sentences or have more blank lines, which I don't think is a problem, so I just use `RegexpHeader` and ignore blank lines or spaces, like line 16 @wu-sheng

wu-sheng: This is not a problem. I am just asking about your only check `.java`.

PR #3283, opened 2019-08-16
Module config set override is not working

wu-sheng:
Due to our module core is using refactor to set value, the `setter` is not working. I just change it to getter.


wu-sheng:
Update UI submodule too.

PR #3287, opened 2019-08-19
Add cached writer.

wayilau:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

I found that every time we get log, we need exe whole getLogWriter.
Add cached writer. so we need not call whole method always.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/WriterFactory.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/WriterFactory.java
@@ -27,12 +27,12 @@ import org.apache.skywalking.apm.util.StringUtil;
 
 public class WriterFactory {
 
-    private static volatile IWriter writer;
+    private static volatile IWriter WRITER;
 
     public static IWriter getLogWriter() {
 
-        if (writer != null) {
-            return writer;
+        if (WRITER != null) {
+            return WRITER;
         }
 
         if (!useConsole() && SnifferConfigInitializer.isInitCompleted() && AgentPackagePath.isPathFound()) {
@@ -43,12 +43,12 @@ public class WriterFactory {
                     e.printStackTrace();
                 }
             }
-            writer = FileWriter.get();
+            WRITER = FileWriter.get();
         } else {
-            writer = SystemOutWriter.INSTANCE;
+            WRITER = SystemOutWriter.INSTANCE;
         }
 
-        return writer;
+        return WRITER;
     }
 
     private static boolean useConsole() {
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/WriterFactory.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/WriterFactory.java
@@ -25,33 +25,35 @@ import org.apache.skywalking.apm.agent.core.conf.Config;
 import org.apache.skywalking.apm.agent.core.conf.SnifferConfigInitializer;
 import org.apache.skywalking.apm.util.StringUtil;
 
+/**
+ * @author Alan Lau
+ */
 public class WriterFactory {
 
     private static volatile IWriter WRITER;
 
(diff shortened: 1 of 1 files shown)


wu-sheng:
At today, provide a static writer pointer may be safe for now. I am not sure about the future yet. Do you face any real issue about this?

wayilau:
> At today, provide a static writer pointer may be safe for now. I am not sure about the future yet. Do you face any real issue about this?

Actually not. It is just my idea from code point.

wayilau:
> `org.apache.skywalking.apm.agent.core.conf.Config.Logging#OUTPUT` is modifiable at runtime and the real writer may change, after your patch the modification of `Config.Logging#OUTPUT` takes no effect.
> 
> Although I think we hardly ever change the `Config.Logging#OUTPUT` at runtime, but the unit tests do this in order to verify, so I think you should either modify the unit test accordingly, or keep everthing unchanged, WDYT @wu-sheng

@kezhenxu94 I see.  But why we change the output type at runtime? is there a real scene?

kezhenxu94:
> > `org.apache.skywalking.apm.agent.core.conf.Config.Logging#OUTPUT` is modifiable at runtime and the real writer may change, after your patch the modification of `Config.Logging#OUTPUT` takes no effect.
> > Although I think we hardly ever change the `Config.Logging#OUTPUT` at runtime, but the unit tests do this in order to verify, so I think you should either modify the unit test accordingly, or keep everthing unchanged, WDYT @wu-sheng
> 
> @kezhenxu94 I see. But why we change the output type at runtime? is there a real scene?

I think the answer is NO, but you have to make the CI pass, it changes the output type at runtime, just for verification :)

wayilau:
> > > `org.apache.skywalking.apm.agent.core.conf.Config.Logging#OUTPUT` is modifiable at runtime and the real writer may change, after your patch the modification of `Config.Logging#OUTPUT` takes no effect.
> > > Although I think we hardly ever change the `Config.Logging#OUTPUT` at runtime, but the unit tests do this in order to verify, so I think you should either modify the unit test accordingly, or keep everthing unchanged, WDYT @wu-sheng
> > 
> > 
> > @kezhenxu94 I see. But why we change the output type at runtime? is there a real scene?
> 
> I think the answer is NO, but you have to make the CI pass at least, it changes the output type at runtime, just for verification :)

ok

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/WriterFactory.java
```diff
@@ -25,23 +25,37 @@
 import org.apache.skywalking.apm.agent.core.conf.SnifferConfigInitializer;
 import org.apache.skywalking.apm.util.StringUtil;
 
+/**
+ * @author Alan Lau
+ */
 public class WriterFactory {
+
+    private static volatile IWriter WRITER;
```
wu-sheng: This variable is not going to be changed in concurrency case, right? Why need `volatile`?

wayilau: As logger.info may be called by many threads, so we give volatile may reduce call the check method.  @wu-sheng

wu-sheng: Most application has thread pool, such as Tomcat. Even cache mechanism makes the check method invoked more than once, it is harmless. But if you add `volatile `, you will force every single access to real memory rather than cache. I am seeing this as an unexpected cost.

wayilau: ok, i've checked the false sharing doc. You're right. i will remove this.

kezhenxu94:
@wu-sheng `PatternLoggerTest` this is one of the unstable unit tests AFAIK, other unstable tests like `TraceSegmentServiceClientTest` are all singletons, so I think the reason why they fail sometimes is the states they hold are changed somewhere else (in other unit tests), I'll just re-run the CI here and will get some time to dig more into this

Issue #3263, opened 2019-08-14
recordDataTTL do not work


xiongrl:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
recordDataTTL notwork！
-  storage:
      elasticsearch:
  	recordDataTTL: 45

```java
   RecordStreamProcessor.create(){
     Model model = modelSetter.putIfAbsent(recordClass, stream.scopeId(),
                new Storage(stream.name(), true, true, Downsampling.Second), true);
  }

 public class ElasticsearchStorageTTL implements StorageTTL {
    @Override public TTLCalculator calculator(Downsampling downsampling) {
        switch (downsampling) {
            case Month:
                return new MonthTTLCalculator();
            case Hour:
                return new EsHourTTLCalculator();
            case Minute:
                return new EsMinuteTTLCalculator();
            default:
                return new DayTTLCalculator();
        }
    }
}  
public class DayTTLCalculator implements TTLCalculator {

    @Override public long timeBefore(DateTime currentTime, DataTTLConfig dataTTLConfig) {
        return Long.valueOf(currentTime.plusDays(0 - dataTTLConfig.getDayMetricsDataTTL()).toString("yyyyMMdd"));
    }
}
```

xiongrl:
commit

wu-sheng:
What is your problem? TTL of ES is day related.

xiongrl:
storage:
   elasticsearch:
       recordDataTTL: 45

trace data was deleted once every two days.

2019-08-13 00:02:03,809 - org.apache.skywalking.oap.server.library.client.elasticsearch.ElasticSearchClient - 161 [pool-17-thread-1] DEBUG [] - delete skywalking_segment-20190811 index finished, isAcknowledged: true

wu-sheng:
OK. Get your point. Are you saying the TTL of the record is using TTL of day metrics, right?

xiongrl:
yes

wu-sheng:
ES and other storages have this same issue. Reporting bug to @peng-yongsheng

wu-sheng:
Fixed. @xiongrl Could you recheck the master branch, and recheck at your env?

PR #3292, opened 2019-08-20
Fix TTL issue.

wu-sheng:
Description in #3263, I am adding new TTL Calculators for record type, in ES and MySQL storages.

W/ this, the TTL should work well in delete record.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/ttl/RecordTTLCalculator.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/ttl/RecordTTLCalculator.java
@@ -29,6 +29,6 @@ import org.joda.time.DateTime;
 public class RecordTTLCalculator implements TTLCalculator {
 
     @Override public long timeBefore(DateTime currentTime, DataTTLConfig dataTTLConfig) {
-        return Long.valueOf(currentTime.plusMinutes(0 - dataTTLConfig.getRecordDataTTL()).toString("yyyyMMddHHmm"));
+        return Long.valueOf(currentTime.plusMinutes(0 - dataTTLConfig.getRecordDataTTL()).toString("yyyyMMddHHmmss"));
     }
 }
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/ttl/RecordTTLCalculator.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/ttl/RecordTTLCalculator.java
@@ -13,32 +13,22 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
  */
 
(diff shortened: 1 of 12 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/ttl/RecordTTLCalculator.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Calculate TTL for record.
+ *
+ * @author wusheng
+ */
+public class RecordTTLCalculator implements TTLCalculator {
+
+    @Override public long timeBefore(DateTime currentTime, DataTTLConfig dataTTLConfig) {
+        return Long.valueOf(currentTime.plusMinutes(0 - dataTTLConfig.getRecordDataTTL()).toString("yyyyMMddHHmm"));
```
peng-yongsheng: Segment time bucket is the second time format.

wu-sheng: Fixed, w/ wrong in EsRecordTTLCalculator too, we should use `#plusDays` there, right?

peng-yongsheng: Alarm record time bucket is also the second time format.

Alarm
```
record.setStartTime(message.getStartTime());
record.setTimeBucket(TimeBucket.getSecondTimeBucket(message.getStartTime()));
``` 

Segment
```
long timeBucket = TimeBucket.getSecondTimeBucket(segmentCoreInfo.getStartTime());
segment.setTimeBucket(timeBucket);
```

Suggest that, rename the getSecondTimeBucket to getRecordTimeBucket or create an alias name of getRecordTimeBucket, it will make it clear.

peng-yongsheng: > Fixed, w/ wrong in EsRecordTTLCalculator too, we should use `#plusDays` there, right?

En, En.

wu-sheng:
@peng-yongsheng Method rename done.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/ttl/RecordTTLCalculator.java
peng-yongsheng: toString("yyyyMMddHHmm") is wrong, modify to  toString("yyyyMMddHHmmss").

peng-yongsheng: Plus minute is for the config item in application.yml. 
 `recordDataTTL: ${SW_CORE_RECORD_DATA_TTL:90} # Unit is minute`

toString("yyyyMMddHHmmss") if for the database, the `timeBucket` column in database is second data format.

wu-sheng:
@peng-yongsheng Have you refreshed the page? You are referring the `Outdated` codes.

Issue #3298, opened 2019-08-21
Group APIs for the operation name in agent


wu-sheng:
Hi

This issue started a long time ago, and recently raised at #3293.
Operation Name in auto instrumentation agent is unpredictable, some time, it carries parameter in it, due to the parameter included in URI mostly. Such as `/api/checkTicket/tk/{userToken}`.
We solved most of these cases, by leverage the parameter pattern path in framework, such as SpringMVC, Webflux, etc. But it is undetected in RPC client side, such as HTTP restful client.

The core issues of this are 
1. Too many memory costs in the agent.
1. Too many registers happen in the backend, and metrics become meaningless because separated by parameter.

So we need to add some APIs to make the group mechanism works. Due to performance cost, I would suggest this suite of APIs are out of tracing core, instead, require the plugin to call them explicitly.

Please join the discussion here @kezhenxu94 @zhaoyuguang @IanCao @candyleer @JaredTan95 @david1228

wu-sheng:
Who knows the good group lib? Beside regex expression.

SkyWalkingRobot:
Here is the [test report](http://github.com/SkywalkingTest/agent-integration-test-report/tree/master/2019/8/apache/testReport-master-2019-08-27-11-48.md) and [validate logs](http://35.227.88.23:49001/job/skywalking-agent-test/248/artifact/workspace/logs/validate_2019-08-27-11-48.log)

wu-sheng:
Fixed.

PR #3299, opened 2019-08-21
Support op group name in agent config core.

wu-sheng:
Begin to work on #3298, and I begin to add document about this feature. Please review and we could work from here.

# Operation Name Group Rule
Operation Name in auto instrumentation agent is unpredictable, some time, target application carries parameter in it, due to the parameter included in URI mostly.
Those operation name are also as known endpoint name in most cases.
Such as /api/checkTicket/tk/{userToken}.

We solved most of these cases, by leverage the parameter pattern path in framework, such as SpringMVC, Webflux, etc. 
But it is undetected in RPC client side, such as HTTP restful client.
In this case, we have to ask the users to set the group rule manually.

All rules are supported to set through agent.config, system properties and system env, like other agent settings.
- Config format, `plugin.opgroup.`plugin name`.rule[`rule name`]`=pattern expression
Example, `plugin.opgroup.resttemplate.rule[/user/auth/{token}]=/user/auth/*`

We have following plugins supporting operation name group.


kezhenxu94:
I didn't find a better way than regular expressions, the idea looks good to me, but seems it will have a serious performance impact on the agent, besides, I have a couple of questions:

0. Can the normalization be done in backend? What's the pros and cons?
0. The operations will evolve for sure, deleting operations does no harm though, but adding and updating operations (done in `dev` side) requires updating the rules accordingly (usually done in `ops` side), and we don't support dynamical configuration in agent side, it may be hard to mantain the rules in time

wu-sheng:
> Can the normalization be done in backend? What's the pros and cons?

Backend thing still costs the same CPU, same calculation same CPU costs, just different places. I am thinking about tradeoff, regex is powerful, but I don't intend to do so. Just support `*` should be enough.
And make the agent cache harder. Right?

> The operations will evolve for sure, deleting operations does no harm though, but adding and updating operations (done in dev side) requires updating the rules accordingly (usually done in ops side), and we don't support dynamical configuration in agent side, it may be hard to mantain the rules in time

That is another thing, how to downstream the config from OAP, which is not impossible, right? We have many static configurations today, I think this is not a block, just a feature wanted.

@kezhenxu94

wu-sheng:
@kezhenxu94 We have regex based string group util, which gives the following benchmark. Read `StringFormatGroupTest` for more detail. For my understanding, if this works for a single REST client, should be enough. What do you think?

```
/*********************************
     * # JMH version: 1.21
     * # VM version: JDK 1.8.0_91, Java HotSpot(TM) 64-Bit Server VM, 25.91-b14
     * # VM invoker: /Users/wusheng/Documents/applications/jdk1.8.0_91.jdk/Contents/Home/jre/bin/java
     * # VM options: -ea -Didea.test.cyclic.buffer.size=1048576 -javaagent:/Applications/IntelliJ IDEA.app/Contents/lib/idea_rt.jar=54841:/Applications/IntelliJ IDEA.app/Contents/bin -Dfile.encoding=UTF-8
     * # Warmup: <none>
     * # Measurement: 5 iterations, 10 s each
     * # Timeout: 10 min per iteration
     * # Threads: 1 thread, will synchronize iterations
     * # Benchmark mode: Throughput, ops/time
     * # Benchmark: org.apache.skywalking.apm.util.StringFormatGroupTest.test100Rule
     *
     * # Run progress: 0.00% complete, ETA 00:01:40
     * # Fork: 1 of 1
     * Iteration   1: 32016.496 ops/s
     * Iteration   2: 36703.873 ops/s
     * Iteration   3: 37121.543 ops/s
     * Iteration   4: 36898.225 ops/s
     * Iteration   5: 34712.564 ops/s
     *
     *
     * Result "org.apache.skywalking.apm.util.StringFormatGroupTest.test100Rule":
     *   35490.540 ±(99.9%) 8345.368 ops/s [Average]
     *   (min, avg, max) = (32016.496, 35490.540, 37121.543), stdev = 2167.265
     *   CI (99.9%): [27145.173, 43835.908] (assumes normal distribution)
     *
     *
     * # JMH version: 1.21
     * # VM version: JDK 1.8.0_91, Java HotSpot(TM) 64-Bit Server VM, 25.91-b14
     * # VM invoker: /Users/wusheng/Documents/applications/jdk1.8.0_91.jdk/Contents/Home/jre/bin/java
     * # VM options: -ea -Didea.test.cyclic.buffer.size=1048576 -javaagent:/Applications/IntelliJ IDEA.app/Contents/lib/idea_rt.jar=54841:/Applications/IntelliJ IDEA.app/Contents/bin -Dfile.encoding=UTF-8
     * # Warmup: <none>
     * # Measurement: 5 iterations, 10 s each
     * # Timeout: 10 min per iteration
     * # Threads: 1 thread, will synchronize iterations
     * # Benchmark mode: Throughput, ops/time
     * # Benchmark: org.apache.skywalking.apm.util.StringFormatGroupTest.testMatch
     *
     * # Run progress: 50.00% complete, ETA 00:00:50
     * # Fork: 1 of 1
     * Iteration   1: 1137158.205 ops/s
     * Iteration   2: 1192936.161 ops/s
     * Iteration   3: 1218773.403 ops/s
     * Iteration   4: 1222966.452 ops/s
     * Iteration   5: 1235609.354 ops/s
     *
     *
     * Result "org.apache.skywalking.apm.util.StringFormatGroupTest.testMatch":
     *   1201488.715 ±(99.9%) 150813.461 ops/s [Average]
     *   (min, avg, max) = (1137158.205, 1201488.715, 1235609.354), stdev = 39165.777
     *   CI (99.9%): [1050675.254, 1352302.176] (assumes normal distribution)
     *
     *
     * # Run complete. Total time: 00:01:41
     *
     * REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
     * why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
     * experiments, perform baseline and negative tests that provide experimental control, make sure
     * the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
     * Do not assume the numbers tell you what you want them to tell.
     *
     * Benchmark                           Mode  Cnt        Score        Error  Units
     * StringFormatGroupTest.test100Rule  thrpt    5    35490.540 ±   8345.368  ops/s
     * StringFormatGroupTest.testMatch    thrpt    5  1201488.715 ± 150813.461  ops/s
     *
     */
```

wu-sheng:
Document updated w/ new code implementation.

# Operation Name Group Rule
Operation Name in auto instrumentation agent is unpredictable, some time, target application carries parameter in it, due to the parameter included in URI mostly.
Those operation name are also as known endpoint name in most cases.
Such as /api/checkTicket/tk/{userToken}.

We solved most of these cases, by leverage the parameter pattern path in framework, such as SpringMVC, Webflux, etc. 
But it is undetected in RPC client side, such as HTTP restful client.
In this case, we have to ask the users to set the group rule manually.

All rules are supported to set through agent.config, system properties and system env, like other agent settings.
- Config format, `plugin.opgroup.`plugin name`.rule[`rule name`]`=pattern regex expression

Multiple configuration items for a single plugin with different keys are supports, such as
1. plugin.opgroup.resttemplate.rule[/rule1]=/path1
1. plugin.opgroup.resttemplate.rule[/rule2]=/path2
1. plugin.opgroup.resttemplate.rule[/rule2]=/path3

We have following plugins supporting operation name group.

| Plugin | Config Key | Example |
|:----:|:-----:|:----|
|RestTemplate| plugin.opgroup.resttemplate.rule | plugin.opgroup.resttemplate.rule[/user/auth/{token}]=\/user\/auth\/.* |

@ascrutae Please review the Rest Template plugin change and run test for it.

wu-sheng:
@david1228 Could you try this?  Or are you using other plugins for outside API call?

david1228:
> @david1228 Could you try this? Or are you using other plugins for outside API call?

Okay, I tried to test this function

david1228:
@wu-sheng 
It took me some time to test the case, the final test passed.
Through the communication with wu-sheng, some problems are confirmed and solved.

Thanks for your support.
skywalking-ui trace span info as follows:
WX20190826-180929@2x

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/OperationNameFormatService.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Format the operation name based on group rules
+     *
+     * @param definition in the Config
+     * @param opName represents the operation name literal string
+     * @return format string if rule matched or the given opName
+     */
+    public static String formatOperationName(Class<? extends OPGroupDefinition> definition, String opName) {
```
kezhenxu94: this method can be non-static

wu-sheng: Yes. Fixed.

PR #3302, opened 2019-08-22
Add more configuration and explainations.

JaredTan95:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

Show more frequently used configuration.

### DIFF
--- a/apm-sniffer/config/agent.config
+++ b/apm-sniffer/config/agent.config
@@ -18,7 +18,7 @@
 # agent.namespace=${SW_AGENT_NAMESPACE:default-namespace}
 
 # The service name in UI
-agent.service_name=${SW_AGENT_NAME:Your_ApplicationName}
+agent.service_name=${SW_AGENT_NAME:Your_ServiceName}
 
 # The number of sampled traces per 3 seconds
 # Negative number means sample traces as many as possible, most likely 100%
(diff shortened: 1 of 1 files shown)


apm-sniffer/config/agent.config
```diff
@@ -46,3 +46,12 @@ logging.file_name=${SW_LOGGING_FILE_NAME:skywalking-api.log}
 
 # Logging level
 logging.level=${SW_LOGGING_LEVEL:DEBUG}
+
+# Logging dir
+# logging.dir=${SW_LOGGING_DIR:""}
+
+# Logging max_file_size
+# logging.max_file_size=${SW_LOGGING_MAX_FILE_SIZE:300 * 1024 * 1024}
```
kezhenxu94: Can we use an expression `300 * 1024 * 1024` here?

wu-sheng: I don't think so. Why open all this config? I think most users don't need to know this

JaredTan95: > I don't think so. Why open all this config? I think most users don't need to know this

Some users use a log collector and have their own acquisition rules. As a result, some customers need to change the location of the log output or other configurations

JaredTan95:
And I also think, hold more frequently used configuration here, when they see those configurations, the user knows about the agent advanced features.

kezhenxu94:
@JaredTan95 The e2e test uses the default name `Your_ApplicationName`, and it's renamed by you to `Your_ServiceName`, making the tests failed, can you please change the test data as well?

https://github.com/apache/skywalking/blob/49f50a24a4c88fcc9a6c95f1f94d95a8166d5955/test/e2e/e2e-agent-reboot/src/test/resources/expected-data/org.apache.skywalking.e2e.SampleVerificationITCase.services.yml#L20-L25

https://github.com/apache/skywalking/blob/d1a521b705a27e92aa844f03e9d90131e47391cd/test/e2e/e2e-single-service/src/test/resources/expected-data/org.apache.skywalking.e2e.SampleVerificationITCase.services.yml#L21-L25

wu-sheng:
I prefer don't change this. It has been there for a long time.

wu-sheng:
Just revert the service name change should be better.

JaredTan95:
ok

Issue #3296, opened 2019-08-21
MySQL PreparedStatement execute method been enhanced twice


honganan:
I may found a bug for MySQL `PreparedStatement.execute()` method enhancement when I am 
imitating MySQL enhancement In class 
`apm-sniffer/apm-sdk-plugin/mysql-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/v5/define/PreparedStatementInstrumentation.java`
The phenomenon is the interceptor of `execute()` method executed twice.

After study I may known how this happend. When trying to enhance a subclass which inherite a method from it's parent class, Bytebuddy will copy this method to this subclass. In case of the parent class enhanced first, then copy it to subclass and the twice enhancement happened.(refer to:https://github.com/raphw/byte-buddy/issues/622)

In our situation the `PreparedStatementInstrumentation` enhanced `execute()` method in 
`com.mysql.jdbc.PreparedStatement` 
and 
`com.mysql.jdbc.JDBC42PreparedStatement`
But `execute()` method in `JDBC42PreparedStatement` is inherited from `PreparedStatement`.

wu-sheng:
See `DeclaredInstanceMethodsInterceptPoint`. Look like `JDBC42PreparedStatement` need this kind of interceptPoint. Would you submit a pull request to fix this?

cc @candyleer

honganan:
Yes，glad to

PR #3306, opened 2019-08-22
Honganan fix 3296

honganan:
**Bug fix**
#3296 

**How to fix**
As all concerned methods are in parent class com.mysql.jdbc.PreparedStatement, so I think we can just enhance the parent class like MySQL6


wu-sheng:
@ascrutae MySQL 5 test cases seem normal, could you recheck the codes?

PR #3308, opened 2019-08-22
Support gateway without agent

### LINKED ISSUE #3218: Support Gateway w/o agent
Today, we asked for agents in every node, but in reality, we may not have suitable agents, due to language or plugin. So in the backend, we should provide a way to tell the analysis core, there is a gateway or proxy to represent as a load balancer for multiple services.

We should add setting about trace-receiver, also active as dynamic configuration. We should list
1. Proxy service name.
1. IP(host name) list under all proxy service.
1. Generate service and service instance for each service and its IP.
1. Change A->B(when there is no agent in the proxy), to A -> proxy -> B, meaning while `A -> peer` as A to proxy as the client-side and service incoming metrics, `ref -> B` as proxy to B server-side metrics and service incoming metrics.

kezhenxu94:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

this PR resolves #3218


oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/gateway/IGatewayService.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package org.apache.skywalking.oap.server.receiver.trace.gateway;
+
+import org.apache.skywalking.oap.server.library.module.Service;
+
+/**
+ * @author kezhenxu94
+ */
+public interface IGatewayService extends Service {
```
wu-sheng: Why this is a service? Who would use it? Service is for inter-module.

wu-sheng: You could create a class to do all these things, but it is not a module service, actually. receiver usually doesn't include service definition. `ISegmentParserService` service exists because the zipkin receiver is going to use it.

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/module/TraceModule.java
```diff
@@ -32,6 +33,9 @@ public TraceModule() {
     }
 
     @Override public Class[] services() {
-        return new Class[] {ISegmentParserService.class};
+        return new Class[] {
+            ISegmentParserService.class,
+            IGatewayService.class
```
wu-sheng: Remove this service please.

kezhenxu94:
@wu-sheng This PR is ready for review now, all the review comments are addressed, thanks for your review :) , and @JaredTan95 please feel free to comment and review if you are interested in this

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/register/ServiceInventory.java
```diff
@@ -187,7 +194,7 @@ public ServiceInventory getClone() {
         if (serviceInventory.getLastUpdateTime() >= this.getLastUpdateTime()) {
             this.nodeType = serviceInventory.getNodeType();
             setProp(serviceInventory.getProp());
-            if (Const.NONE != serviceInventory.getMappingServiceId()) {
+            if (serviceInventory.isForceResetServiceMapping() || Const.NONE != serviceInventory.getMappingServiceId()) {
```
wu-sheng: `#forceResetServiceMapping` should be combined, otherwise, if some combine merge happens in streaming, rather than memory-db merge, this field lost, as a result, the database would update `mappingServiceId` anymore.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/register/service/IServiceInventoryRegister.java
```diff
@@ -35,5 +35,5 @@
 
     void heartbeat(int serviceId, long heartBeatTime);
 
-    void updateMapping(int serviceId, int mappingServiceId);
+    void updateMapping(int serviceId, int mappingServiceId, boolean force);
```
wu-sheng: I would recommend you to provide a `#resetMapping` method. Because in that case, there is no need to provide `#mappingServiceId`. This method only should be called when you want to mappingServiceId back to 0. In other cases, this update works w/o `force==true`.

This change seems to misguide the API user, if `force` usually means, update even the value is there(not null/0), but here, your meaning is, mappingServiceId could be set as 0 when the mappingServiceId != 0.

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/parser/listener/service/ServiceMappingSpanListener.java
```diff
@@ -58,14 +71,20 @@ private ServiceMappingSpanListener(ModuleManager moduleManager) {
...
+                    ServiceMapping serviceMapping = new ServiceMapping();
+                    serviceMapping.setServiceId(serviceId);
+
+                    if (config.getStaticGatewaysConfig().isAddressConfiguredAsGateway(address)) {
+                        serviceMapping.setMappingServiceId(Const.NONE);
+                        serviceMapping.setForceUpdate(true);
+                    } else {
                         serviceMapping.setMappingServiceId(segmentCoreInfo.getServiceId());
-                        serviceMappings.add(serviceMapping);
+                        serviceMapping.setForceUpdate(false);
```
wu-sheng: Keep `serviceMappings` as normal update, and add a list reset mappings for this case.

oap-server/server-starter/src/main/assembly/gateways.yml
```diff
@@ -0,0 +1,20 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+gateways:
+  - name: proxy0
+    instances:
+      - host: 127.0.0.1 # the host/ip of this gateway instance
+        port: 9099 # the port of this gateway instance, defaults to 80
```
wu-sheng: Why does this file include something?

kezhenxu94: Oops, that must be a mistake 😄

kezhenxu94:
Hi, @wu-sheng the e2e test is included into the e2e-cluster module and it passed in the previous build, and I think this PR is ready to be reviewed/merged, thanks for all the review comments.

docs/en/setup/backend/static-gateways.md
```diff
@@ -0,0 +1,22 @@
+# Static Gateways/Proxies
```
wu-sheng: I prefer to use `Uninstrumented` to replace static, because gateway is always static :)

docs/en/setup/backend/static-gateways.md
```diff
@@ -0,0 +1,22 @@
...
+
+The word "static" here means that the gateways are not automatically/dynamically registered by SkyWalking agent when they are started,
+but configured statically in `gateways.yml` file or via [Dynamic Configuration](dynamic-config.md). The reason why they can't register
```
wu-sheng: `The reason of`, not `why`

docs/en/setup/backend/static-gateways.md
```diff
@@ -0,0 +1,22 @@
...
+to backend automatically is that there're no suitable agent plugins, for example, there is no agent plugins for Nginx, haproxy, etc.
+So in order to reflect the real topology, we provide a way to configure the gateways/proxies manually.
```
wu-sheng: `reflect` -> `visualize`

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/register/service/ServiceInventoryRegister.java
```diff
@@ -132,6 +136,19 @@ private ServiceInventoryCache getServiceInventoryCache() {
         }
     }
 
+    @Override public void resetMapping(int serviceId) {
```
wu-sheng: You should indicate why we have resetMapping. A short introduction version.

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/StaticGatewaysConfig.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import static java.util.Objects.isNull;
+
+/**
+ * @author kezhenxu94
+ */
+@Slf4j
+public class StaticGatewaysConfig extends ConfigChangeWatcher {
```
wu-sheng: Same name change expected here.

docs/en/setup/backend/static-gateways.md
kezhenxu94: I remembered that this is not a grammar error, just like `the person who does...`, `the PR that changes...`, `the place where I live...`, `the way how we get here`, if you prefer I'll change to `The reason of that they can't register...`

kezhenxu94: I search for a random webpage, seems that people say like this too https://www.oxfordlearnersdictionaries.com/definition/english/reason_1

wu-sheng: OK. I just feel a little strange, if it is right, keep it.

kezhenxu94:
I also increase the timeout to wait for ES's readiness, CI failed for no particular reason in the several previous builds because of timeout

docs/en/setup/backend/uninstrumented-gateways.md
```diff
@@ -0,0 +1,22 @@
+# Uninstrumented Gateways/Proxies
+
+The word "static" here means that the gateways are not automatically/dynamically registered by SkyWalking agent when they are started,
```
wu-sheng: You miss `static` here :)

PR #3323, opened 2019-08-26
Add Light4j plugin

tsuilouis:

- [ ] Bug fix
- [X] New feature provided
- [ ] Improve performance

___

This pull request is for a Skywalking plugin for Light4J 1.6.x, which uses Java 8 (Light4J 2.x uses Java 11).

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-1.6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/HandleRequestInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-1.6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/HandleRequestInterceptor.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *
+ */
+
+package org.apache.skywalking.apm.plugin.light4j;
+
+import io.undertow.server.HttpServerExchange;
+import org.apache.skywalking.apm.agent.core.context.ContextManager;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+import org.apache.skywalking.apm.network.trace.component.ComponentsDefine;
+
+import java.lang.reflect.Method;
+
+/**
+ * {@link HandleRequestInterceptor} creates a local span when the {@link com.networknt.handler.LightHttpHandler#handleRequest(HttpServerExchange)}
+ * method executes. {@link com.networknt.handler.LightHttpHandler} is implemented by various middleware and business handlers and
+ * the Light4J framework runs these in succession. Thus when a request is sent, a chain of
+ * {@link org.apache.skywalking.apm.agent.core.context.trace.LocalSpan}s is expected to be produced.
+ *
+ * @author tsuilouis
+ */
+public class HandleRequestInterceptor implements InstanceMethodsAroundInterceptor {
+
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                             MethodInterceptResult result) {
+        String operationName = objInst.getClass().getName() + "." + method.getName();
+        ContextManager.createLocalSpan(operationName)
+                .setComponent(ComponentsDefine.LIGHT_4J);
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                              Object ret) {
+        ContextManager.stopSpan();
+        return ret;
+    }
+
+    @Override
+    public void handleMethodException(EnhancedInstance objInst, Method method, Object[] allArguments,
+                                      Class<?>[] argumentsTypes,
+                                      Throwable t) {
+        ContextManager.activeSpan().errorOccurred().log(t);
+    }
+}
--- a/oap-server/server-core/src/test/resources/component-libraries.yml
+++ b/oap-server/server-core/src/test/resources/component-libraries.yml
@@ -212,6 +212,9 @@ JdkHttp:
 spring-webflux:
   id: 67
   languages: Java
+Light4J:
+  id: 68
+  languages: java
 
 # .NET/.NET Core components
 # [3000, 4000) for C#/.NET only
--- a/apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
+++ b/apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
@@ -130,4 +130,6 @@ public class ComponentsDefine {
 
     public static final OfficialComponent SPRING_WEBFLUX = new OfficialComponent(67, "spring-webflux");
 
+    public static final OfficialComponent LIGHT_4J = new OfficialComponent(68, "Light4J");
+
 }
--- a/apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-1.6.x-plugin/pom.xml
+++ b/apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-1.6.x-plugin/pom.xml
@@ -1,0 +1,45 @@
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
+  ~ contributor license agreements.  See the NOTICE file distributed with
+  ~ this work for additional information regarding copyright ownership.
+  ~ The ASF licenses this file to You under the Apache License, Version 2.0
+  ~ (the "License"); you may not use this file except in compliance with
+  ~ the License.  You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  ~
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>light4j-plugins</artifactId>
+        <groupId>org.apache.skywalking</groupId>
+        <version>6.4.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>apm-light4j-1.6.x-plugin</artifactId>
+    <packaging>jar</packaging>
+
+    <name>light4j-1.6.x-plugin</name>
+    <url>http://maven.apache.org</url>
+
+    <properties>
+        <version.light4j>1.6.7</version.light4j>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.networknt</groupId>
+            <artifactId>handler</artifactId>
+            <version>${version.light4j}</version>
+            <scope>provided</scope>
+        </dependency>
+    </dependencies>
+</project>
--- a/apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-1.6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/define/LightInstrumentation.java
+++ b/apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-1.6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/define/LightInstrumentation.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.plugin.light4j.define;
+
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import net.bytebuddy.matcher.ElementMatchers;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.ConstructorInterceptPoint;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.InstanceMethodsInterceptPoint;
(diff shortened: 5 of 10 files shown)


kezhenxu94:
@tsuilouis if it uses JDK 8, please move it into `optional-plugins` module

wu-sheng:
At SkyWalking, the JDK requirement is your plugin codes, rather than your target lib code level. All plugins activated by default should be able to be loaded(not running) in JVM 1.6.

If you want it to be the default, change the plugin codes, and don't use JDK 1.6+ classes. Otherwise, it should stay in optional, and add document about it, as we have several plugins there already.

kezhenxu94:
I have a quick glance at the code, and it seems that it didn't use JDK6+ features, so it should be safe to stay in `apm-sdk-plugin`, I'll have a detailed look later, thanks @tsuilouis

oap-server/server-core/src/test/resources/component-libraries.yml
```diff
@@ -212,6 +212,9 @@ JdkHttp:
 spring-webflux:
   id: 67
   languages: Java
+Light4J:
+  id: 68
+  languages: java
```
zhaoyuguang: oap-server/server-starter/src/main/resources/component-libraries.yml need add too

apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-1.6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/HandleRequestInterceptor.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @author tsuilouis
+ */
+public class HandleRequestInterceptor implements InstanceMethodsAroundInterceptor {
+
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                             MethodInterceptResult result) {
+        String operationName = objInst.getClass().getName() + "." + method.getName();
+        ContextManager.createLocalSpan(operationName)
```
zhaoyuguang: You add Light4j at `HTTP Server` at `Supported-list.md`,  why this plugin has not entrySpan?

wu-sheng: I think Light4j should exist in `supported-list`. If the undertow plugin will create the entry span, then it makes sense to me.

wu-sheng:
Fix conflict please, Play Framework plugin has been merged, so the ID is being used by it. Remember to change the test cases too.

tsuilouis:
@kezhenxu94 I have made some commits that I think solve the broken trace problem. The UI shows two entries, but I see that the trace IDs are identical (3.33.156...) so that means there is one segment? Please let me know if your thoughts on an alternative approach. 

Also, when I ran my automated test cases locally, I still see the `actualData.yml` contains two segments. Is that normal or am I doing something wrong?

exception_handler
v1_pets

wu-sheng:
If rpc happens, then two segments generated, which show as two rows. When you set two service codes at agent.config, at trace page and topology, you should see them linked.

wu-sheng:
For async, segment will be created because another segment created in new thread.

Another question, why do many spans created?

tsuilouis:
Oh I see, `com.networknt.exception.ExceptionHandler#handleRequest` does execute across threads so then another segment is indeed expected. Thank you!

As for why there are so many spans: based on discussions with @stevehu and other Light4J developers, we would like to see detailed information on the `LightHttpHandler`s that are involved during the life of a HTTP request in a Light4J project, in addition to the tracing from the Undertow plugin. Do you foresee any problems with the number of spans being created?

wu-sheng:
SkyWalking is not just a tracing project, it is an APM, so the data you generated related to metrics and topology. And SkyWalking is designed and used in 100% trace sampling and only reasonable resource cost. So first of all, too many spans are performance concerns. Also, APM is focusing on business code performance, rather than framework. We are assuming framework has no big performance issue, but the duration of span still includes the framework cost, in this case including undertow and light4j.

Then, do you generate entry span for light4j? If yes, it exista with undertow entry span, rather than override it, which cause service request counter incorrectly. +2 for every request. When you create an entry span inside the undertow entry span, the override should happen automatically. Also this override should be able to make the icon/type of topology node showing as a light4j service.


If you really think light4j should add many spans, such as you have cases useful for end user, consider to add a plugin config, and share with us why. 

The ideas is mostly about, we are monitoring performance, rather than telling developer what light4j do and how it forward request, even in tech level, we could, but don't fit production requirement.

Do I make myself clear?

stevehu:
The light-4j is based on the undertow and we have created so many middleware handlers to form handler chains for each individual endpoint. As these middleware handlers can be enabled or disabled through config files, our users would like to know how many middleware handlers were passed through before reaching the business handler. It would be great for the APM to capture the information if possible. If you think SkyWalking is not designed for this, do you happen to know what other options do we have? If we don't do need to capture all middleware handler spans, do you think we should just use the Undertow agent? Thanks a lot for your help.

wu-sheng:
> If you think SkyWalking is not designed for this, do you happen to know what other options do we have? 

We could, just don't recommend always set this in default. Consider providing an option in Config.

> If we don't do need to capture all middleware handler spans, do you think we should just use the Undertow agent? Thanks a lot for your help.

An example is, Spring is also doing handler thing, right? We set SpringMVC `@Controller` as entry span, so when a user uses Tomcat + SpringMVC, the user sees `@Controller` as the only entry span. That is what I mean override. Light4j entry span overrides undertow span.
Undertow agent is not the concern, it should disappear if you set entry span for light4j. 

Anyway, there are only suggestions, the only requirement is, don't make `light4j` and `undertwo` entry spans both exist.

wu-sheng:
@zhaoyuguang Please run this test.

@zhaoyuguang @kezhenxu94 Please continue the review.

zhaoyuguang:
I had triggered the auto tests, let's wait for the result

wu-sheng:
@zhaoyuguang Nothing shows up in the report. Does the test case PR miss something?

zhaoyuguang:
> @zhaoyuguang Nothing shows up in the report. Does the test case PR miss something?

Is my mistake of forgot add `the scenario name`, I had triggered the auto tests again.

wu-sheng:
@tsuilouis This PR is facing CI issue.

> [2019-09-16T02:04:56.507Z] [ERROR] /home/jenkins/jenkins-slave/workspace/skywalking-ci-it/apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/define/LightInstrumentation.java:21:1: Disallowed import - io.undertow.server.HttpServerExchange. [ImportControl]

zhaoyuguang:
@tsuilouis CI has some problem, please fix, may reason is 
```
skywalking-agent-test/workspace/source/skywalking/apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/define/LightInstrumentation.java:21:1: Disallowed import - io.undertow.server.HttpServerExchange. [ImportControl]
Audit done.
```

tsuilouis:
@wu-sheng I wanted to leave a top-level folder in case the light-4j team wanted to add plugins for other light-4j subprojects. I can remove it if you want.

wu-sheng:
> I wanted to leave a top-level folder in case the light-4j team wanted to add plugins for other light-4j subprojects. I can remove it if you want.

If you have plans for that, I am OK to keep this structure.

wu-sheng:
@zhaoyuguang CI passed, I think you could run tests again.

apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/HandleRequestInterceptor.java
```diff
@@ -0,0 +1,141 @@
...
+                CarrierItem next = contextCarrier.items();
+                while (next.hasNext()) {
+                    next = next.next();
+                    next.setHeadValue(headers.getFirst(next.getHeadKey()));
+                }
+
+                AbstractSpan span = ContextManager.createEntrySpan(operationName, contextCarrier);
+                Tags.URL.set(span, exchange.getRequestURL());
+                Tags.HTTP.METHOD.set(span, exchange.getRequestMethod().toString());
+                span.setComponent(ComponentsDefine.LIGHT_4J);
```
wu-sheng: As you have light4j entry span, I think you should have a PR to submit icon showing in topology, in this repo https://github.com/apache/skywalking-rocketbot-ui

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -102,6 +102,7 @@ property key | Description | Default |
 `plugin.solrj.trace_statement`|If true, trace all the query parameters(include deleteByIds and deleteByQuery) in Solr query request, default is false.|`false`|
 `plugin.solrj.trace_ops_params`|If true, trace all the operation parameters in Solr request, default is false.|`false`|
 `plugin.opgroup.*`|Support operation name customize group rules in different plugins. Read [Group rule supported plugins](op_name_group_rule.md)|Not set|
+`plugin.light4j.trace_handler_chain`|If true, trace all middleware/business handlers that are part of the Light4J handler chain for a request.|false|
```
wu-sheng: Please move this above `plugin.opgroup.*`

apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/HandleRequestInterceptor.java
```diff
@@ -0,0 +1,141 @@
...
+
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                             MethodInterceptResult result) {
+        if (isExceptionHandler(objInst)) {
+            HttpServerExchange exchange = (HttpServerExchange) allArguments[0];
+
+            if (exchange.isInIoThread()) {
+
+                String operationName = exchange.getRequestPath() + "@" + exchange.getRequestMethod();
```
wu-sheng: Could you give me an example about this? How does the op name look like?

apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/HandleRequestInterceptor.java
wu-sheng: Should be an icon shows up at the top of the node in the topology to represent this service providing service through Light4j.

apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/HandleRequestInterceptor.java
tsuilouis: The op name looks like "/v1/pets@GET". This is the format that light-4j uses with Jaeger implementation of OpenTracing: https://github.com/networknt/light-4j/blob/master/jaeger-tracing/src/main/java/com/networknt/jaeger/tracing/JaegerHandler.java#L79.

I would like to keep the HTTP verb in the op name, but I am open to changing to the rest of the format.

docs/en/setup/service-agent/java-agent/README.md
tsuilouis: Ok, done.

apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/HandleRequestInterceptor.java
wu-sheng: In existing SpringMVC, the format looks like `{PUT, GET} /a/b/c`, based on `RequestMapping` settings. I would recommend we use the same format in light4j

https://github.com/apache/skywalking/blob/c694a30c5fc66dd576c331738ed3bfde760babfe/apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-commons/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/commons/interceptor/RequestMappingMethodInterceptor.java#L48-L57

apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/HandleRequestInterceptor.java
tsuilouis: Ok, I have submitted the PR at https://github.com/apache/skywalking-rocketbot-ui/pull/157.

wu-sheng: Merged, you need to include the UI submodule update into this PR too.

apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/HandleRequestInterceptor.java
tsuilouis: Just to clarify using your example, if the request sent is `GET /a/b/c` you would like to see  the op name as just `{GET} /a/b/c` and not `{PUT, GET} /a/b/c`, correct?

wu-sheng: In SpringMVC, it is config based, so, if one method could support PUT and GET, then it has {PUT, GET} in OP name. What is the case in Light4j?

stevehu: In spring, one controller can handle multiple requests with different HTTP verbs and I think it makes sense to have the format that supports multiple verbs. Light-4j is targeted to microservices and each endpoint will have only one verb and path combined and list of middleware handlers and business handlers. As we are supporting other tracing product like Jaeger and planning to support others. It makes sense to use the Light-4j endpoint for us as our users are familiar with the path@verb formation. What do you think?

wu-sheng: In today's analysis, we don't care about the format much. I am just thinking to make name format consistently. That is the only reason I asked for `{GET} /a/b/c`.

stevehu: I assume that the op name is part of the span that will be displayed on the UI to identify which endpoint the request is coming from. If this assumption is right, then I think the familiarity of the format from framework users is more important than the format itself.

apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/HandleRequestInterceptor.java
tsuilouis: Ok, I have added a commit with the submodule update at https://github.com/apache/skywalking/pull/3323/commits/efde0fff50ddf44bb78f3e25b6d0ccd6f7c38d2f.

apm-sniffer/apm-sdk-plugin/light4j-plugins/light4j-plugin/src/main/java/org/apache/skywalking/apm/plugin/light4j/HandleRequestInterceptor.java
tsuilouis: I think both points are valid to the question of "What span op name format does the user of Skywalking want to see?". If the user prefers the Skywalking platform's format, then they would expect `{GET} /a/b/c`. Other users may prefer the format to be consistent with respect to the framework they use. Perhaps we should have another config option to let the user decide which format they would like to see? I believe this has been done before with one of the Spring plugins having a flag to pick the format for the operation name. The default can follow the Skywalking format.

wu-sheng: If you are sure your(light4j) users are familiar with the current format(with `@`), I am OK to don't change it to SpringMVC's.

wu-sheng:
@kezhenxu94 Look like our E2E CI facing failure issue. Most of them are timeout or failure. Recently, I think we don't merge anything related to it.

wu-sheng:
Welcome to be 158th contributor of main repo :)

tsuilouis:
Thanks for all your help @wu-sheng! I learned a lot from developing the light4j plugin.

wu-sheng:
@tsuilouis @stevehu The plugin test immigration is on going in these two months, a lot of the plugins have moved into PR related test process. https://github.com/apache/skywalking/issues/3583

Could any of you move the light4j plugin test into the main repo too? The original plugin test env has been shut down. Moving into the main repo could make the light4j plugin checked in every PR.

tsuilouis:
@wu-sheng Ok, understood. I will do that over the next few days and make PR when it is ready.

Issue #2659, opened 2019-05-14
intercept failure  org.apache.skywalking.apm.plugin.spring.mvc.commons.exception.IllegalMethodStackDepthException


appregidc:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Bug
___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS and JRE?
skywalking6.1GA centos
- Which company or project?
springboot +spring cloudt
- What happen?

from skywalkinglogs, it promt to submit bug
[new 2.txt](https://github.com/apache/skywalking/files/3175653/new.2.txt)

___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

wu-sheng:
Could you provide Spring version and your demo codes? We need to reproduce the issue.

appregidc:
only found one in exception log

wu-sheng:
We need codes, because others are using these, and no issue reported. We have test cases, but passed. So, we need you to provide detail, which could tell us how this error happens.

appregidc:
![image](https://user-images.githubusercontent.com/16357470/57674823-dede0980-7652-11e9-8ecb-47b37d837e30.png)

这个代码片段是否有帮助，，从异常中就找到这个地方？或者是否可以我们准备一下环境，您remote调试

IanCao:
> ![image](https://user-images.githubusercontent.com/16357470/57674823-dede0980-7652-11e9-8ecb-47b37d837e30.png)
> 
> 这个代码片段是否有帮助？从异常中就找到这个地方。或者是否可以我们准备一下环境，您remote调试

English only
could u provide a demo project that can run directly?

SummerOfServenteen:
@wu-sheng  ,i also encountered the same exception, but the exception message was different from that of him.
This log is printed too often. I have use sw in 7 applications,but the curious thing about it was that it was only happened in one application.Nevertheless, it does not affect data reporting.The exception was caused by which the value of enhancedClassField_ws was null.
![image](https://user-images.githubusercontent.com/31895647/60158610-0adcd500-9824-11e9-9e05-31498e9ff262.png)
This is the exception message.
[exception.txt](https://github.com/apache/skywalking/files/3328772/exception.txt)

I can't find the root cause,the only temporary solution I can think of is to catch this exception.

SummerOfServenteen:
The code is simple
[...]

PR #3337, opened 2019-08-26
Update ControllerConstructorInterceptor.java

SummerOfServenteen:
catch NoSuchMethodError

- [ ] Bug fix

- Related issues
#2659
___
### Bug fix
- Bug description.
when we useing springVersion='4.1.6.RELEASE',there is no SuchMethod of RequestMapping.path().
in this case,it will throw NoSuchMethodError ,then  faild to  setSkyWalkingDynamicField.
so in then method of org.apache.skywalking.apm.plugin.spring.mvc.commons.interceptor.AbstractMethodInterceptor#beforeMethod ,it will throw NullPointerException.

image
image

image

- How to fix?
to catch NoSuchMethodError
___


wu-sheng:
NoSuchMethod should be processed by witness class, rather than try-catch. You should create another interceptor and instrumentation for the case.

wu-sheng:
Please submit a new pull request by following that path. Appreciated.

PR #3357, opened 2019-08-28
ADD: add operationName length threshold

stalary:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___
Limit the length of the operationName to prevent errors when inserting elasticsearch

### DIFF
--- a/apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringUtil.java
+++ b/apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringUtil.java
@@ -18,6 +18,8 @@
 
 package org.apache.skywalking.apm.util;
 
+import java.util.Arrays;
+
 public final class StringUtil {
     public static boolean isEmpty(String str) {
         return str == null || str.length() == 0;
@@ -61,5 +63,12 @@ public final class StringUtil {
             }
         }
         return true;
+    }
+
+    public static String cut(String str, int threshold) {
+        if (isEmpty(str) || str.getBytes().length <= threshold) {
+            return str;
+        }
+        return new String(Arrays.copyOf(str.getBytes(), threshold));
     }
 }
--- a/apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringUtil.java
+++ b/apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringUtil.java
@@ -18,8 +18,6 @@
 
 package org.apache.skywalking.apm.util;
 
-import java.util.Arrays;
-
 public final class StringUtil {
     public static boolean isEmpty(String str) {
         return str == null || str.length() == 0;
@@ -66,9 +64,9 @@ public final class StringUtil {
     }
 
     public static String cut(String str, int threshold) {
-        if (isEmpty(str) || str.getBytes().length <= threshold) {
+        if (isEmpty(str) || str.length() <= threshold) {
             return str;
         }
-        return new String(Arrays.copyOf(str.getBytes(), threshold));
+        return str.substring(0, threshold);
     }
 }
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManager.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManager.java
@@ -43,7 +43,6 @@ public class ContextManager implements BootService {
     private static ThreadLocal<AbstractTracerContext> CONTEXT = new ThreadLocal<AbstractTracerContext>();
     private static ThreadLocal<RuntimeContext> RUNTIME_CONTEXT = new ThreadLocal<RuntimeContext>();
     private static ContextManagerExtendService EXTEND_SERVICE;
-    private static final int NAME_THRESHOLD = 500;
 
     private static AbstractTracerContext getOrCreate(String operationName, boolean forceSampling) {
         AbstractTracerContext context = CONTEXT.get();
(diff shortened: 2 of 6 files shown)


stalary:
Why is it ok for me to use `./mvnw package -Pagent,dist` compilation locally

stalary:
Why is it ok for me to use `./mvnw package -Pagent,dist` compilation locally，But travis-ci can't pass

wu-sheng:
Travis is not stable, ignore it. `CI` and `e2e` are keys.

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringUtil.java
```diff
@@ -62,4 +64,11 @@ public static boolean substringMatch(CharSequence str, int index, CharSequence s
         }
         return true;
     }
+
+    public static String cut(String str, int threshold) {
+        if (isEmpty(str) || str.getBytes().length <= threshold) {
+            return str;
+        }
+        return new String(Arrays.copyOf(str.getBytes(), threshold));
```
kezhenxu94: Extract `str.getBytes()` to variable?

stalary: ok，I will modify it later.

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringUtil.java
```diff
@@ -62,4 +64,11 @@ public static boolean substringMatch(CharSequence str, int index, CharSequence s
         }
         return true;
     }
+
+    public static String cut(String str, int threshold) {
+        if (isEmpty(str) || str.getBytes().length <= threshold) {
```
wu-sheng: Why don't use `String#Length` directly?

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringUtil.java
wu-sheng: What is the difference between this and `new String(value, beginIndex, subLen)` in `String#substring`?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManager.java
```diff
@@ -41,6 +41,7 @@
     private static ThreadLocal<AbstractTracerContext> CONTEXT = new ThreadLocal<AbstractTracerContext>();
     private static ThreadLocal<RuntimeContext> RUNTIME_CONTEXT = new ThreadLocal<RuntimeContext>();
     private static ContextManagerExtendService EXTEND_SERVICE;
+    private static final int NAME_THRESHOLD = 500;
```
wu-sheng: This should be moved to Config, rather than placing in here.

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringUtil.java
stalary: Because I think there will be a difference in the number of byte occupied by Chinese and English

stalary: `id.getBytes(StandardCharsets.UTF_8).length > 512` used in elasticsearch validate

wu-sheng: We don't recommend Chinese OP name in any case.

stalary: Sorry, I don't know this, I will modify it later.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManager.java
wu-sheng: You forget to delete this?

apm-sniffer/config/agent.config
```diff
@@ -38,6 +38,9 @@ agent.service_name=${SW_AGENT_NAME:Your_ApplicationName}
 # Skywalking team may ask for these files in order to resolve compatible problem.
 # agent.is_open_debugging_class = ${SW_AGENT_OPEN_DEBUG:true}
 
+# The operationName max length
+# agent.operation_name_threshold=${OPERATION_NAME_THRESHOLD:500}
```
wu-sheng: The agent set up document requires to update. Add this, and notice user, > 500 is not recommended.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManager.java
stalary: Sorry, I made a stupid mistake

apm-sniffer/config/agent.config
stalary: ok, I have finished.

kezhenxu94: Is it better to prefixed it with something like `SW_AGENT_`?, we have this prefix in other configs

wu-sheng: Make sense to me, please adjust to the ENV name prefix.

stalary: Sorry, I'm going to change it now

stalary: done

Issue #2720, opened 2019-05-20
Provide plugin for kafka-v2-plugin


Mars626:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [x] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS and JRE?
6.0GA

- Which company or project?

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.
i try to sent message by kafka-clients-2.0.1 ,but fail，because of exception in kafka-clients-2.0.1 jar：
“  if (magic < RecordBatch.MAGIC_VALUE_V2 && headers != null && headers.length > 0)
                throw new IllegalArgumentException("Magic v" + magic + " does not support record headers");” in class MemoryRecordsBuilder
___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

ascrutae:
I am going to provide this plugin

wu-sheng:
@Mars626 If you have interests, contributions are welcome.

Mars626:
> I am going to provide this plugin

thanks for your help

wu-sheng:
@ascrutae I think you should fix this in 6.3.0

wu-sheng:
@zhaoyuguang Could you help on this?

zhaoyuguang:
> @zhaoyuguang Could you help on this?

I am very interested in kafka, but I will try to finish it at 6.4.0, if other people are interested, welcome.

dmsolr:
@zhaoyuguang Do you work in progress that? May I join in?

wu-sheng:
I think he hasn't. Welcome to do. FYI @zhaoyuguang

zhaoyuguang:
I haven't do it yet. Welcome to contribute.

stalary:
I have completed a v-2 plugin that has been used in the test environment. I may consider submitting it in the future

wu-sheng:
@stalary Welcome the contributions. We are closing to 6.4.0 release, if you want, please consider to submit pull request before that.

stalary:
> @stalary Welcome the contributions. We are closing to 6.4.0 release, if you want, please consider to submit pull request before that.

Ok, I will submit it as soon as possible

PR #3369, opened 2019-08-29
[WIP] ADD kafka-v2 plugin

stalary:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/2720
___
### Bug fix
- Bug description.

- How to fix?

___
add kafka-v2 plugin to optional-plugins


stalary:
> 2 basic questions
> 
> 1. Why this is optional? I haven't seen 1.8 required.
> 2. Are you doing this because of v1 confliction? If Yes, read the witness class in instrumentation definition. You could use it to separated the plugins, which could be in default plugins at the same time and active when necessary.



> 2 basic questions
> 
> 1. Why this is optional? I haven't seen 1.8 required.
> 2. Are you doing this because of v1 confliction? If Yes, read the witness class in instrumentation definition. You could use it to separated the plugins, which could be in default plugins at the same time and active when necessary.

Ok, I'm going to make some changes

stalary:
@wu-sheng Can I modify v1's witness class in instrumentation definition  to make the plugin selectively started? Otherwise I can only control whether the v2 plugin starts.

stalary:
@wu-sheng  I have another problem. My local test found that using v1 plugin and kafka-client using 0.11.0.0 also appeared `Magic v1 does not support record headers`.

wu-sheng:
> Can I modify v1's witness class in instrumentation definition to make the plugin selectively started? Otherwise I can only control whether the v2 plugin starts.

You should.

> I have another problem. My local test found that using v1 plugin and kafka-client using 0.11.0.0 also appeared Magic v1 does not support record headers.

I am not familiar with that, from the codes, it seems just server side or client lib doesn't support header

wu-sheng:
Another plugin merged, please fix conflicts

wu-sheng:
Integration test cases are required. Do you know how to do that?

stalary:
> Integration test cases are required. Do you know how to do that?

I know I will continue to improve it tomorrow

wu-sheng:
> I know I will continue to improve it tomorrow

Take your time. And please fix the issues of verifications in old cases, https://github.com/SkyAPMTest/agent-integration-test-report#kafka. They have broken for a while.

stalary:
> > I know I will continue to improve it tomorrow
> 
> Take your time. And please fix the issues of verifications in old cases, https://github.com/SkyAPMTest/agent-integration-test-report#kafka. They have broken for a while.

Okay, got it

dmsolr:
> @wu-sheng I have another problem. My local test found that using v1 plugin and kafka-client using 0.11.0.0 also appeared `Magic v1 does not support record headers`.

Did you connect to V1.0.x or higher using client-0.11.0.0?

stalary:
> Did you connect to V0.10.x or lower using client-0.11.0.0? As far as I know v0.11.0 uses message format V2(batch record).

@dmsolr Oh, that's right, so if judging the version of kafka-clents doesn't determine whether it works or not, should the header be dropped？And `ApiVersions.computeMaxUsableProduceMagic` mean that we can't use headers in produce?

dmsolr:
I am not understand well what is your means.
`ApiVersions` can be witness class, because it is in every versions. The MessageFormatVersion is none of the plugins business.

stalary:
> The MessageFormatVersion is none of the plugins business.

I mean computeMaxUsableProduceMagic will set the magic in the produce message is 0, but the header data needs to be in the magic of 2

stalary:
I think if we delete the header we can support the old version of kafka

dmsolr:
Not really. SW plugins nevenr change the original logic. This is Kafka's businesses. I think.

> I think if we delete the header we can support the old version of kafka

How do we can do that? We can not change the message format. (The `header` is a part of MessageFormat in your words, right?)

stalary:
> Not really. SW plugins nevenr change the original logic. This is Kafka's businesses. I think.
> 
> > I think if we delete the header we can support the old version of kafka
> 
> How do we can do that? We can not change the message format. (The `header` is a part of MessageFormat in your words, right?)

Sorry, I may not have made it clear, but what I meant was to stop manually adding headers to ProducerRecord in the kafka-v1 plugin

wu-sheng:
Server side is not of control, please don't consider to change message body. We just leverage existing header. If server side doesn't support it, so be it, we don't support. Plugin is easy to be removed, we don't need to worry about that. 

The point of plugin, we make sure witness is good in detecting client library version, so our interceptor works as expected, that is all.

stalary:
> Server side is not of control, please don't consider to change message body

I have the same idea, so I want to remove the activity of adding headers manually in the kafka-v1 plugin.

wu-sheng:
After you removed, who covers v1 use case? I haven't read codes, just have context you discuss here. Header is required to make trace continued.

apm-sniffer/apm-sdk-plugin/kafka-v1-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/v1/KafkaProducerInterceptor.java
```diff
@@ -47,22 +47,18 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
...
-        CarrierItem next = contextCarrier.items();
-        while (next.hasNext()) {
-            next = next.next();
-            record.headers().add(next.getHeadKey(), next.getHeadValue().getBytes());
-        }
-
-        EnhancedInstance callbackInstance = (EnhancedInstance)allArguments[1];
-        if (callbackInstance != null) {
+        // kafkaTemplate can lead to ClassCastException
+        if (allArguments[1] instanceof EnhancedInstance) {
```
wu-sheng: Use more explicit instrument definition, including parameter type. Instanceof is not recommend here

apm-sniffer/apm-sdk-plugin/kafka-v1-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/v1/KafkaProducerInterceptor.java
```diff
@@ -47,22 +47,18 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
...
+        Tags.MQ_BROKER.set(activeSpan, (String) objInst.getSkyWalkingDynamicField());
         Tags.MQ_TOPIC.set(activeSpan, topicName);
+        Tags.MQ_QUEUE.set(activeSpan, key);
         SpanLayer.asMQ(activeSpan);
         activeSpan.setComponent(ComponentsDefine.KAFKA_PRODUCER);
 
-        CarrierItem next = contextCarrier.items();
-        while (next.hasNext()) {
-            next = next.next();
-            record.headers().add(next.getHeadKey(), next.getHeadValue().getBytes());
```
wu-sheng: AFAIK, kafka 0.11+ supports header. Why remove this?

wu-sheng: Our online demo is running on kafka.

stalary: I used kafka-clients of version 0.11.0.3 in my local test, but kafka server use 0.10.0, which will cause error reporting. Do we need to support new clients to connect to old servers?

apm-sniffer/apm-sdk-plugin/kafka-v1-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/v1/KafkaProducerInterceptor.java
stalary: Ok, I will adjust it later

apm-sniffer/apm-sdk-plugin/kafka-v1-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/v1/KafkaProducerInterceptor.java
stalary: Currently we are using the old version of kafka service in our production cluster, but we also want to use the kafka plugin

wu-sheng: That is our job. It is Kafka requirement. Nothing more.

If you want, feel free to update document to indicate server version requirement. But it is pointless to make us support less versions tham before. If this change lands, we have broken trace for sure even server is already 0.11+

stalary: Ok, I get it, I'm going to continue to improve the code

dmsolr: > I used kafka-clients of version 0.11.0.3 in my local test, but kafka server use 0.10.0, which will cause error reporting. Do we need to support new clients to connect to old servers?

I think you are mistake. The version has to match with Server and Client. Because Kafka has several MessageFormatVersion. This is the key point.

stalary: > > I used kafka-clients of version 0.11.0.3 in my local test, but kafka server use 0.10.0, which will cause error reporting. Do we need to support new clients to connect to old servers?
> 
> I think you are mistake. The version has to match with Server and Client. Because Kafka has several MessageFormatVersion. This is the key point.

You can rediscover it by using the kafka service before 0.11 and then using the client after 0.11

stalary:
`ERROR 2019-08-30 16:15:14:501 http-nio-19930-exec-3 InstMethodsInter : class[class org.apache.kafka.clients.producer.KafkaProducer] before method[doSend] intercept failure
java.lang.ClassCastException: org.springframework.kafka.core.KafkaTemplate$$Lambda$1286/692688374 cannot be cast to org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance
	at org.apache.skywalking.apm.plugin.kafka.v1.KafkaProducerInterceptor.beforeMethod(KafkaProducerInterceptor.java:66)
	at org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstMethodsInter.intercept(InstMethodsInter.java:82)
	at org.apache.kafka.clients.producer.KafkaProducer.doSend(KafkaProducer.java)
	at org.apache.kafka.clients.producer.KafkaProducer.send(KafkaProducer.java:803)
	at org.springframework.kafka.core.DefaultKafkaProducerFactory$CloseSafeProducer.send(DefaultKafkaProducerFactory.java:444)
	at org.springframework.kafka.core.KafkaTemplate.doSend(KafkaTemplate.java:372)
	at org.springframework.kafka.core.KafkaTemplate.send(KafkaTemplate.java:190)
	at outfox.course.tiku.kafka.Producer.send(Producer.java:38)`

When I use kafka-clients-0.11.0.3 through kafkaTemplate and use callback, it will cause this error. Do you know the reason? I have no idea. @dmsolr

wu-sheng:
@stalary This seems Spring Kafka is adding something new, so it breaks the type cast. Only the SkyWalking instrumented class implement `EnhancedInstance` dynamically. If you want to fix this, that is another issue.

stalary:
> @stalary This seems Spring Kafka is adding something new, so it breaks the type cast. Only the SkyWalking instrumented class implement `EnhancedInstance` dynamically. If you want to fix this, that is another issue.

Ok, now I upgrade the kafka cluster to 0.11.0.1, and use kafka-clients-2.0.1 test. It seems that the kafka-v1 plugin only has problems with consumption interception (2.0 updates the method), but there is no problem with production, so I plan to modify the kafka-v1 plug-in for compatibility, instead of writing kafka-v2 plugin.
Moreover, I think https://github.com/apache/skywalking/issues/2720 is also caused by the old version of kafka cluster, so we need to modify the document to explain it

wu-sheng:
> Ok, now I upgrade the kafka cluster to 0.11.0.1, and use kafka-clients-2.0.1 test. It seems that the kafka-v1 plugin only has problems with consumption interception (2.0 updates the method), but there is no problem with production, so I plan to modify the kafka-v1 plug-in for compatibility, instead of writing kafka-v2 plugin.
Moreover, I think #2720 is also caused by the old version of kafka cluster, so we need to modify the document to explain it

OK, sounds good, take any action you need to do. Let's keep Kafka 2.x tests still there.

dmsolr:
@stalary Do you make sure kafka-v1-plugin support V2?
I work a little on this, I found V1 is basically compatible with V2 without COMSUMER. Maybe we just  need to compate with COMSUER.
In another hand, DockerFile of testcase has some problem need to fix.

stalary:
> Do you make sure kafka-v1-plugin support V2?

@dmsolr  Yes, I went through some local tests and the producers didn't find any problems.

Consumers changed from the original pollOnce to pollForFetches, so I need to modify getMethodsMatcher to make it compatibility, and I plan to enrich the information of tags and add MQ_QUEUE to represent the key in production and the groud.id in consumption, how do you feel?

dmsolr:
> Consumers changed from the original pollOnce to pollForFetches, so I need to modify getMethodsMatcher to make it compatibility, 

Yes, I think so.

> and I plan to enrich the information of tags and add MQ_QUEUE to represent the key in production and the groud.id in consumption, how do you feel?

About adding tags, I think you need to think carelly whether it have to add. Because `group.id` can be append to OperationName. 

MQ_QUEUE means topic? If so, it is contained in OperationName.  and it is a tag as well.

stalary:
> MQ_QUEUE means topic? If so, it is contained in OperationName. and it is a tag as well.

You are right.

wu-sheng:
Operation name relate to endpoint metrics. Please consider carefully. We don't bring too many unnecessary endpoint id

dmsolr:
> Operation name relate to endpoint metrics. Please consider carefully. We don't bring too many unnecessary endpoint id

I got it. Usually, we don't subscribe too many topics by many different `group.id` in one agent. Otherwise, `group.id` is usuful for analysis. I think we could consider doing it.

wu-sheng:
> I got it. Usually, we don't subscribe too many topics by many different group.id in one agent. Otherwise, group.id is usuful for analysis. I think we could consider doing it.

For Kafka, I agree. `queue/groupid` is your proposal?

dmsolr:
Yes.

wu-sheng:
@stalary If you are going to merge 2.x and 1.x, please change 1.x module name/package name, remove `1.x`.

stalary:
@wu-sheng  Ok, I will submit the new PR

wu-sheng:
Replaced by #3390

PR #3370, opened 2019-08-29
support es https ssl config.

JaredTan95:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
 Based pr：https://github.com/apache/skywalking/pull/3236
___
### Bug fix
- Bug description.

- How to fix?

___


docs/en/setup/backend/backend-storage.md
```diff
@@ -54,6 +56,34 @@ storage:
...
+
+```yaml
+storage:
+  elasticsearch:
+    # nameSpace: ${SW_NAMESPACE:""}
+    user: ${SW_ES_USER:""} # User needs to be set when Http Basic authentication is enabled
+    password: ${SW_ES_PASSWORD:""} # Password to be set when Http Basic authentication is enabled
+    clusterNodes: ${SW_STORAGE_ES_CLUSTER_NODES:localhost:443}
+    keyStorePath: ${SW_SW_STORAGE_ES_SSL_JKS_PATH:"../es_keystore.jks"}
+    keyStorePass: ${SW_SW_STORAGE_ES_SSL_JKS_PASS:"dangerous"}
```
hanahmily: seems elasticsearch doesn't support Keystore password according to https://www.elastic.co/guide/en/elasticsearch/reference/6.8/secure-settings.html. But this feature may be added in the future, I agree with your current design, well, just to pick an empty default `password` value would be a better choice though

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -84,30 +95,50 @@
...
         if (StringUtils.isNotBlank(user) && StringUtils.isNotBlank(password)) {
             final CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
             credentialsProvider.setCredentials(AuthScope.ANY, new UsernamePasswordCredentials(user, password));
             builder = RestClient.builder(pairsList.toArray(new HttpHost[0]))
                 .setHttpClientConfigCallback(httpClientBuilder -> httpClientBuilder.setDefaultCredentialsProvider(credentialsProvider));
+
+            if ("https".equals(protocol)) {
+                // more type: https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#KeyStore
+                KeyStore truststore = KeyStore.getInstance("jks");
+                try (InputStream is = Files.newInputStream(Paths.get(keyStorePath))) {
```
hanahmily: We should consider a scenario. When the certification is signed by trusted CA such as VeriSign(not self-signed), RestClient should pick up default SslContext instead of a dedicated one. What I mean is the keyStorePath should be empty in that case.

docs/en/setup/backend/backend-storage.md
```diff
@@ -54,6 +56,34 @@ storage:
...
+example: 
+
+```yaml
+storage:
+  elasticsearch:
+    # nameSpace: ${SW_NAMESPACE:""}
+    user: ${SW_ES_USER:""} # User needs to be set when Http Basic authentication is enabled
+    password: ${SW_ES_PASSWORD:""} # Password to be set when Http Basic authentication is enabled
+    clusterNodes: ${SW_STORAGE_ES_CLUSTER_NODES:localhost:443}
+    keyStorePath: ${SW_SW_STORAGE_ES_SSL_JKS_PATH:"../es_keystore.jks"}
```
hanahmily: s/keyStore/trustStore/ ? Since we are client-side, `trustStore` may be a more reasonable name. What do you think about this?

JaredTan95:
@hanahmily  As we discussed before, we could also set SSL within JDK-JVM-Options, like this:

```bash
-Djavax.net.ssl.keyStoreType=jks
-Djavax.net.ssl.keyStore=path-to/my_keystore.jks
-Djavax.net.ssl.keyStorePassword=password
-Djavax.net.ssl.trustStore=path-to/my_keystore.jks
-Djavax.net.ssl.trustStorePassword=password
```

So we can choose one of these ways, through elasticsearch client code implementation or through JVM options?

wu-sheng:
@JaredTan95 Look like we need a document update at es storage section.

docs/en/setup/backend/backend-storage.md
JaredTan95: > @JaredTan95 Look like we need a document update at es storage section.

You mean this?

wu-sheng:
You prefer to add the JVM things you mentioned in the document, some people may be not familiar with that.

JaredTan95:
> You prefer to add the JVM things you mentioned in the document, some people may be not familiar with that.

What I mean is that, if we use JVM  options, we don't need to change the above code.

wu-sheng:
Then why should we change? Is there any difference between those two ways?

JaredTan95:
As far as I know, there should be no difference. That's why I'm just talking about which way we're going to choose.  @hanahmily  Any suggestions?

wu-sheng:
> As far as I know, there should be no difference. That's why I'm just talking about which way we're going to choose. @hanahmily Any suggestions?

If so, I really prefer to do through JVM options, just update the document. @hanahmily Do you have any particular reason to use our codes to do so?

JaredTan95:
> > As far as I know, there should be no difference. That's why I'm just talking about which way we're going to choose. @hanahmily Any suggestions?
> 
> If so, I really prefer to do through JVM options, just update the document. @hanahmily Do you have any particular reason to use our codes to do so?

I agree, code is really more complex.

wu-sheng:
@JaredTan95 One thing crosses my mind, this is useful if SkyWalking supports multiple HTTP peers, such as ES https, export https, dynamic configuration service in http.

hanahmily:
> @JaredTan95 One thing crosses my mind, this is useful if SkyWalking supports multiple HTTP peers, such as ES https, export https, dynamic configuration service in http.

Exactly, that's why the default JVM SSL setting is not the best options in most practices. For operator/maintainer, if they must update Keystore file in order to add a new SSL peer, that won't be accepted. Not only the risk of failure but also it's not possible. For examples, some Keystore fileis a hardware device.

docs/en/setup/backend/backend-storage.md
hanahmily: No, I prefer to select `trustStorePath` instead of `keyStorePath`

docs/en/setup/backend/backend-storage.md
hanahmily: Why mark this `resolved`?

hanahmily: not refresh my page, ignore this.

wu-sheng:
@hanahmily @JaredTan95 Anything needs to update?

docs/en/setup/backend/backend-storage.md
```diff
@@ -39,6 +39,8 @@ storage:
     # nameSpace: ${SW_NAMESPACE:""}
     # user: ${SW_ES_USER:""} # User needs to be set when Http Basic authentication is enabled
     # password: ${SW_ES_PASSWORD:""} # Password to be set when Http Basic authentication is enabled
+    #keyStorePath: ${SW_SW_STORAGE_ES_SSL_JKS_PATH:""}
+    #keyStorePass: ${SW_SW_STORAGE_ES_SSL_JKS_PASS:""}
```
wu-sheng: These two are still using `keyStore*`.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -84,30 +95,50 @@
...
+    public void connect() throws IOException, KeyStoreException, NoSuchAlgorithmException, KeyManagementException, CertificateException {
         List<HttpHost> pairsList = parseClusterNodes(clusterNodes);
         RestClientBuilder builder;
         if (StringUtils.isNotBlank(user) && StringUtils.isNotBlank(password)) {
             final CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
             credentialsProvider.setCredentials(AuthScope.ANY, new UsernamePasswordCredentials(user, password));
             builder = RestClient.builder(pairsList.toArray(new HttpHost[0]))
                 .setHttpClientConfigCallback(httpClientBuilder -> httpClientBuilder.setDefaultCredentialsProvider(credentialsProvider));
+
+            if ("https".equals(protocol)) {
```
hanahmily: `trustStorePaht` may be null in no-self-signed scenario. so we should update this condition.

hanahmily: hmmm... what I mean is that, if user picks up default ssl,  lines from 127 ~ 137 should be remove.
For instance
```java
if (StringUtils.isNotBlank(trustStorePath)) {
   builder = RestClient.builder(pairsList.toArray(new HttpHost[0]))
                    .setHttpClientConfigCallback(httpClientBuilder -> httpClientBuilder.setDefaultCredentialsProvider(credentialsProvider);
} else {
   KeyStore truststore = KeyStore.getInstance("jks");
    ....
   builder = RestClient.builder(pairsList.toArray(new HttpHost[0]))
                    .setHttpClientConfigCallback(httpClientBuilder -> httpClientBuilder.setDefaultCredentialsProvider(credentialsProvider).setSSLContext(sslContext));
}
```

JaredTan95: Thanks for the review. I've updated it.

PR #3385, opened 2019-08-31
fix springMVC4 noSuchMethod

SummerOfServenteen:

- [ ] Bug fix

- Related issues
#2659
#3337 
___
### Bug fix
- Bug description.

- How to fix?
how about this way.I'm not sure whether I used the witness class correctly.I just test some version of springMVC ,and it works well.please check.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/v4/define/ControllerForLowVersionInstrumentation.java
+++ b/apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/v4/define/ControllerForLowVersionInstrumentation.java
@@ -16,16 +16,21 @@
  *
  */
 
-
 package org.apache.skywalking.apm.plugin.spring.mvc.v4.define;
 
-import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ClassInstanceMethodsEnhancePluginDefine;
+public class ControllerForLowVersionInstrumentation extends AbstractControllerInstrumentation {
+    public static final String WITNESS_CLASSES_LOW_VERSION = "org.springframework.web.context.support.ServletContextPropertyPlaceholderConfigurer";
 
-public abstract class AbstractSpring4Instrumentation extends ClassInstanceMethodsEnhancePluginDefine {
-    public static final String WITHNESS_CLASSES = "org.springframework.cache.interceptor.SimpleKey";
+    public static final String ENHANCE_ANNOTATION = "org.springframework.stereotype.Controller";
 
     @Override
     protected String[] witnessClasses() {
-        return new String[] {WITHNESS_CLASSES, "org.springframework.cache.interceptor.DefaultKeyGenerator"};
+        return new String[]{WITHNESS_CLASSES, "org.springframework.cache.interceptor.DefaultKeyGenerator", WITNESS_CLASSES_LOW_VERSION};
+    }
+
+
+    @Override
+    protected String[] getEnhanceAnnotations() {
+        return new String[]{ENHANCE_ANNOTATION};
     }
 }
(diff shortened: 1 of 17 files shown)


SummerOfServenteen:
if  the version of springMVC is lower than  4.2.0.release ,the ControllerForLowVersionInstrumentation will be used ,ohterwise the ControllerInstrumentation will be used.

wu-sheng:
You don't add it into config file, how it works?

apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/v4/define/ControllerForLowVersionInstrumentation.java
```diff
@@ -0,0 +1,36 @@
...
+package org.apache.skywalking.apm.plugin.spring.mvc.v4.define;
+
+public class ControllerForLowVersionInstrumentation extends AbstractControllerInstrumentation {
+    public static final String WITNESS_CLASSES_LOW_VERSION = "org.springframework.web.context.support.ServletContextPropertyPlaceholderConfigurer";
+
+    public static final String ENHANCE_ANNOTATION = "org.springframework.stereotype.Controller";
+
+    @Override
+    protected String[] witnessClasses() {
+        return new String[]{WITHNESS_CLASSES, "org.springframework.cache.interceptor.DefaultKeyGenerator", WITNESS_CLASSES_LOW_VERSION};
```
wu-sheng: Why `org.springframework.cache.interceptor.DefaultKeyGenerator` exists without class variable? And have `WITNESS_CLASSES_LOW_VERSION` too. Which one is working?

apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/v4/define/ControllerInstrumentation.java
```diff
@@ -21,9 +21,17 @@
...
-    @Override protected String[] getEnhanceAnnotations() {
-        return new String[] {ENHANCE_ANNOTATION};
+    @Override
+    protected String[] getEnhanceAnnotations() {
+        return new String[]{ENHANCE_ANNOTATION};
+    }
+
+    @Override
+    protected String[] witnessClasses() {
+        return new String[]{WITHNESS_CLASSES, "org.springframework.cache.interceptor.DefaultKeyGenerator", WITNESS_CLASSES_HIGH_VERSION};
```
wu-sheng: Same as above

apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/v4/define/ControllerForLowVersionInstrumentation.java
wu-sheng: `org.springframework.cache.interceptor.DefaultKeyGenerator` exist in low and high instrumentation definitions, what is the point?

dmsolr: `ControllerForLowVersionConstructorInterceptor ` dose not seem to be refered. We need to override `getConstructorsInterceptPoints()` here, right?

wu-sheng:
Any update?

SummerOfServenteen:
sorry,I was a little busy with my work.I'll check this problem carefully tomorrow.

SummerOfServenteen:
Should i pr with a new brach?This pr is messy because of my rebase

wu-sheng:
Do the thing you think better. You don't need our permission😄

PR #3408, opened 2019-09-04
fix springMVC4 noSuchMethod

SummerOfServenteen:

- [ ] Bug fix
- Related issues
#2659
#3337
___
### Bug fix
- Bug description.

- How to fix?


wu-sheng:
@ascrutae Please review, I need your opinion about this, due to you build the spring plugin at the beginning.

apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/v4/define/ControllerForLowVersionInstrumentation.java
```diff
@@ -0,0 +1,57 @@
...
+            public String getConstructorInterceptor() {
+                return "org.apache.skywalking.apm.plugin.spring.mvc.v4.ControllerForLowVersionConstructorInterceptor";
+            }
+        };
+        return new ConstructorInterceptPoint[]{constructorInterceptPoint};
+    }
+
+    @Override
+    protected String[] witnessClasses() {
+        return new String[]{WITHNESS_CLASSES, "org.springframework.cache.interceptor.DefaultKeyGenerator", WITNESS_CLASSES_LOW_VERSION};
```
ascrutae: @SummerOfServenteen 
This issue only happened before the version 4.2.0. The `path` method is only supported after version 4.2.0. 

Why does the `org.springframework.cache.interceptor.DefaultKeyGenerator` class add the witness class list? I found that the class also in version 4.2.0

I think that the `org.springframework.web.context.support.ServletContextPropertyPlaceholderConfigurer` class as the witness class will be ok. because of the class removed after version 4.2.0.

SummerOfServenteen: They are retained to distinguish between V3 and V5

ascrutae: The witness class means that the Instrumentation will be active when the witness class is available.  and `ServletContextPropertyPlaceholderConfigurer` is available before version 4.2.0, and it means that`ControllerForLowVersionInstrumentation` will be active before version 4.2.0.  So I am confused that why need to distinguish between V3 and V5？

wu-sheng: @SummerOfServenteen If the target class is not there in v3, then you don't witness class.

wu-sheng:
@SummerOfServenteen Update?

SummerOfServenteen:
image
image
image

wu-sheng:
@ascrutae Please check.

ascrutae:
Got your point, @SummerOfServenteen . The PR looks good to me

PR #3437, opened 2019-09-08
Add TTL E2E test

### LINKED ISSUE #3417: TTL e2e test
We have faced the TTL bugs several times, due to performance optimization and code refactor. I think we should have this e2e. Because TTL is time sensitive, we should use a mock trace generator or mesh receiver protocol, to make sure we could cover enough time range.

If you have any issue, we could discuss here. I tagged this as 6.4.0 for now, if you have limited bandwidth, change it. This is not a feature or bug :)

kezhenxu94:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance
- [x] Tests

- Related issues

closes #3417

### DIFF
--- a/test/e2e/e2e-ttl/e2e-ttl-es/src/test/java/org/apache/skywalking/e2e/StorageTTLITCase.java
+++ b/test/e2e/e2e-ttl/e2e-ttl-es/src/test/java/org/apache/skywalking/e2e/StorageTTLITCase.java
@@ -17,12 +17,12 @@
 
 package org.apache.skywalking.e2e;
 
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
+import io.grpc.internal.DnsNameResolverProvider;
+import io.grpc.netty.NettyChannelBuilder;
+import io.grpc.stub.StreamObserver;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.skywalking.apm.dependencies.io.grpc.ManagedChannel;
-import org.apache.skywalking.apm.dependencies.io.grpc.ManagedChannelBuilder;
-import org.apache.skywalking.apm.dependencies.io.grpc.internal.DnsNameResolverProvider;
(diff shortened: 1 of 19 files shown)


wu-sheng:
Why can't I see the TTL stage in Jenkins page?

wu-sheng:
@kezhenxu94 Please send a notice mail to dev mail list, as we are preparing to add new e2e test case.

test/e2e/e2e-ttl/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+            <version>${skywalking.version}</version>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+    </dependencies>
```
wu-sheng: All receivers and core are not required in my mind, it is a gRPC interface, only protos are required.

test/e2e/e2e-ttl/e2e-ttl-es/src/test/java/org/apache/skywalking/e2e/StorageTTLITCase.java
```diff
@@ -0,0 +1,278 @@
...
+                try {
+                    instanceRespTimeMatcher.verify(serviceMetrics);
+                    valid = true;
+                } catch (Throwable t) {
+                    log.warn("History metrics data are not deleted yet, {}", t.getMessage());
+                    Thread.sleep(10000);
+                }
+            } catch (Throwable t) {
+                log.warn("History metrics data are not deleted yet", t);
+                Thread.sleep(10000);
```
wu-sheng: I think it is more reasonable to set max test times, according to your `SW_CORE_DATA_KEEPER_EXECUTE_PERIOD` has been set to 1(min). Then, I would say 10 times(10 mins) should be good to say success or failure? What do you think?

Issue #3428, opened 2019-09-06
SpringCloudGateway plugin generate same trace id after Exception


yujiaxinlong:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Bug
- Which version of SkyWalking, OS and JRE?
tried on 6.3.0 release and 6.4.0 snapshot
jdk version 1.8.0_211
spring cloud gateway 2.1.2.RELEASE

- Which company or project?
a spring cloud micro-service project

- What happen?
when using skywalking on spring cloud gateway,  if there is fallback page set and the routing thrown an exception, skywalking agent will keeps generating the same trace id for all requests.

like this 
`2019-09-06 21:03:52.377 [reactor-http-nio-2] INFO  c.c.c.gateway.filter.JwtFilter - [TID:35.69.15677744414800043] - checking should file /api/auth/users/loginMgr 
2019-09-06 21:03:53.108 [reactor-http-nio-2] INFO  c.c.c.gateway.filter.JwtFilter - [TID:35.69.15677744414800043] - checking should file /api/auth/users/loginMgr 
2019-09-06 21:03:53.690 [reactor-http-nio-2] INFO  c.c.c.gateway.filter.JwtFilter - [TID:35.69.15677744414800043] - checking should file /api/auth/users/loginMgr 
2019-09-06 21:03:54.269 [reactor-http-nio-2] INFO  c.c.c.gateway.filter.JwtFilter - [TID:35.69.15677744414800043] - checking should file /api/auth/users/loginMgr `

the way to reproduce it:

firstly,  start a spring cloud gateway application with skywalking agent, and a backend service. Let the gateway routes requests by eureka.

secondly, send requests to backend service through spring cloud gateway,  the skywalking agent would work fine.

kill the backend service, and keep sending requests.  A Connection refused Exception will be thrown, fallback  controller get the Exception and response , the trace id is still changing now.

later, the Exception would become SERVICE_UNAVAILABLE  Unable to find instance for xxx,  fallback controller  still get the Exception and response. Then, the trace id will not change anymore, even  after the backend service is restarted.  And all the requests to other service through the gateway get the same trace id too.
___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

This seems a bug caused by springcloudgateway plugin,  maybe other plugin that handle the fallback controller also involved, hope it can be solved ASAP.

[...]

PR #3445, opened 2019-09-09
Gateway compatible with downstream loss

zhaoyuguang:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
resolve https://github.com/apache/skywalking/issues/3428

___
### Bug fix
- Bug description.
When the downstream is lost, the reactor-netty interceptor won't work, but the NettyRoutingFilterInterceptor has do `prepareForAsync` and `stopSpan`
image

- How to fix?
At `AbstractServerResponseMethodInterceptor` if ServerWebExchange has span and ContextManager get == null, use span.asyncFinish() to report it to backend


zhaoyuguang:
I am preparing testcase

zhaoyuguang:
Resolves https://github.com/apache/skywalking/issues/3428

wu-sheng:
@zhaoyuguang Only put `resolve` in the first comment, auto closing will work. I just changed it. You could see it related now.

zhaoyuguang:
> @zhaoyuguang Only put `resolve` in the first comment, auto closing will work. I just changed it. You could see it related now.

Got it.

Issue #2764, opened 2019-05-27
Provide plugin for Apache Pulsar


yangl:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [x] Requirement
- [ ] Feature or performance improvement
___
### Requirement or improvement
- Provide plugin for [Apache Pulsar](https://pulsar.apache.org/).

wu-sheng:
Plugin contribution is welcome.

kezhenxu94:
Provided

PR #3476, opened 2019-09-17
Add pulsar apm plugin

codelipenghui:

- [ ] Bug fix
- [x ] New feature provided
- [ ] Improve performance

- Related issues
 #2764 
___

Add pulsar apm plugin and unit tests, integration test will add soon.

### Local tests result:

image

As the image shows, created 1 producer and 3 consumers with different subscription name(test-1, test-2, test-3).

### Local integration test result:
[INFO] ---------------------------Here is the test report context -------------------------------------

# Test Report - 2019-09-24 11:30

- tests  : **7 passed**. **0 failed**
- branch name : **[pulsar-plugin](https://github.com/apache/incubator-skywalking/tree/pulsar-plugin)**
- commit id : **[88e1ef3eeb2cabc30690c5a42a7f1604e9eff2e7](https://github.com/apache/incubator-skywalking/commit/88e1ef3eeb2cabc30690c5a42a7f1604e9eff2e7)**
- cases branch: **[pulsar-scenario](https://github.com/SkywalkingTest/skywalking-autotest-scenarios/tree/pulsar-scenario)**
- cases commit id: **[a5731397356122ee0868a9b107f7ded28e46dd03](https://github.com/SkywalkingTest/skywalking-autotest-scenarios/commit/a5731397356122ee0868a9b107f7ded28e46dd03)**

## Cases List

| Case description | Status | Cases|
|:-----|:-----:|:-----:|
|Pulsar| **7 passed. 0 failed**| [click me](#pulsar) |

###
|  Version     | Passed | Failed|
|:------------- |:-------:|:-----:|
| 2.3.1  | :heavy_check_mark:||
| 2.3.0  | :heavy_check_mark:||
| 2.4.0  | :heavy_check_mark:||
| 2.4.1  | :heavy_check_mark:||
| 2.3.2  | :heavy_check_mark:||
| 2.2.1  | :heavy_check_mark:||
| 2.2.0  | :heavy_check_mark:||


wu-sheng:
@dmsolr Feel free to join.

wu-sheng:
W/ light4j plugin merged, you need to update the component definition because the ID has been used.

wu-sheng:
After you fix the conflicts and pass tests locally, please post your local test report here.

codelipenghui:
@wu-sheng I have fix the conflicts and past the local test screenshot to the description of this PR, please help take a look the test result screenshot.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarConsumerInterceptor.java
```diff
@@ -0,0 +1,79 @@
...
+                ContextManager.stopSpan();
+            }
+        }
+        return ret;
+    }
+
+    @Override
+    public void handleMethodException(EnhancedInstance objInst, Method method, Object[] allArguments,
+                                      Class<?>[] argumentsTypes, Throwable t) {
+        ContextManager.activeSpan().errorOccurred().log(t);
```
dmsolr: I notice that Span starts after the method, so we will meet NPE when an exception happens before the Span created.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarProducerInterceptor.java
```diff
@@ -0,0 +1,85 @@
...
+                callbackRequiredInfo.setContextSnapshot(snapshot);
+                callbackInstance.setSkyWalkingDynamicField(callbackRequiredInfo);
+            }
+        }
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                              Object ret) throws Throwable {
+        ContextManager.stopSpan();
```
dmsolr: I think we need to ensure SPAN has been created will be better.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/SendCallbackInterceptor.java
```diff
@@ -0,0 +1,68 @@
...
+public class SendCallbackInterceptor implements InstanceMethodsAroundInterceptor {
+
+    private static final String OPERATION_NAME = "Pulsar/Producer/Callback";
+
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                             MethodInterceptResult result) throws Throwable {
+        SendCallbackEnhanceRequiredInfo requiredInfo = (SendCallbackEnhanceRequiredInfo) objInst.getSkyWalkingDynamicField();
+        if (null != requiredInfo.getContextSnapshot()) {
+            AbstractSpan activeSpan = ContextManager.createLocalSpan(OPERATION_NAME);
```
dmsolr: I'm not familiar with Pulsar. In my experience, the LocalSpan can not always refer to ExitSpan, or it will miss directly sometimes.

dmsolr: If the ExitSpan is closed and LocalSpan is still working, it may happen.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarConsumerInterceptor.java
codelipenghui: I think we should move create active span logic to the beforeMethod, by the way, shall we need to catch the IllegalStateException here, because if no active span here will throw the IllegalStateException and i can't find a public method to check activeSpanStack is or not empty.

codelipenghui: I think we should move create active span logic to the beforeMethod, by the way, shall we need to catch the IllegalStateException here, because if no active span here will throw the IllegalStateException and i can't find a public method to check activeSpanStack is or not empty.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarProducerInterceptor.java
codelipenghui: As mentioned above, to catch the IllegalStateException or is their any method to check the span is created

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/SendCallbackInterceptor.java
codelipenghui: In pulsar, i think the span of send callback is always a LocalSpan because in the exit span can guarantee that the callback is complete or complete with exception.

Here is the source of pulsar producer: https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerImpl.java#L294

So, If I don't get it wrong, the SendCallback should create a LocalSpan. If there are any mistakes, please point them out.

codelipenghui:
@dmsolr Thanks for your review, i have move the span creation of consumer to beforeMethod. Please help take a look again.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarConsumerInterceptor.java
dmsolr: I think you dose not get my points.

In usually, We need not catch the `IllegalStateException` and need not check the size of the stack. But we have to make sure that the started and stopped of Span must present in pairs in an interceptor. In other word, which starts a Span must stop it.

dmsolr: Usually, We suggest doing like below.
```
}
```
The key point, we have to recheck by the same conditions as a span start when we wanna stop the span.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/SendCallbackInterceptor.java
dmsolr: I got it. But I am thinking of another thing. In a scenario, `SendCallBack.sendComplete` is block till its parent span closed. At this moment, we call the `ContextManager.continued(requiredInfo.getContextSnapshot())` maybe throw an exception. 
Maybe we can use [Asyn API](https://github.com/apache/skywalking/blob/master/docs/en/guides/Java-Plugin-Development-Guide.md#advanced-apis).(I am not sure.)

apm-sniffer/apm-sdk-plugin/pulsar-plugin/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+
+    <artifactId>apm-pulsar-plugin</artifactId>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>8</source>
+                    <target>8</target>
```
wu-sheng: Do you need JDK8 for plugin? If so, this has to be placed in optional plugin.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/ConsumerConstructorInterceptor.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package org.apache.skywalking.apm.plugin.pulsar;
+
+import org.apache.pulsar.client.impl.PulsarClientImpl;
+import org.apache.pulsar.client.impl.conf.ConsumerConfigurationData;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
+
+public class ConsumerConstructorInterceptor implements InstanceConstructorInterceptor {
```
wu-sheng: Comments please, with all other classes/methods.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarConsumerInterceptor.java
```diff
@@ -0,0 +1,82 @@
...
+        if (allArguments.length > 0) {
+            Message msg = (Message) allArguments[0];
+            if (null != msg) {
+                ContextCarrier contextCarrier = new ContextCarrier();
+                CarrierItem next = contextCarrier.items();
+                while (next.hasNext()) {
+                    next = next.next();
+                    next.setHeadValue(msg.getProperty(next.getHeadKey()));
+                }
+                ContextManager.extract(contextCarrier);
```
wu-sheng: Why extract in `afterMethod` rather than when creating span?

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarProducerInterceptor.java
wu-sheng: I think if logically you could generate the span created in Pulsar plugin, it is OK to just stop span directly.

Because even we do `isActive` check, you could just check the span created by other plugins.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/SendCallbackInterceptor.java
wu-sheng: I am missing the context here. Could you show me the thread workflow for your concern?

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/define/AbstractPulsarInstrumentation.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.plugin.pulsar.define;
+
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ClassInstanceMethodsEnhancePluginDefine;
+
+public abstract class AbstractPulsarInstrumentation extends ClassInstanceMethodsEnhancePluginDefine {
```
wu-sheng: What is an abstract instrumentation for?

apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
```diff
@@ -133,7 +133,12 @@
     public static final OfficialComponent PLAY = new OfficialComponent(68, "Play");
 
     public static final OfficialComponent CASSANDRA_JAVA_DRIVER = new OfficialComponent(69, "cassandra-java-driver");
-
+  
     public static final OfficialComponent LIGHT_4J = new OfficialComponent(71, "Light4J");
 
+    public static final OfficialComponent PULSAR = new OfficialComponent(72, "Pulsar");
```
wu-sheng: Who will use this?

apm-sniffer/apm-sdk-plugin/pulsar-plugin/pom.xml
codelipenghui: Yes, pulsar-client requires JDK1.8, i will move the pulsar plugin to optional plugin

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/define/AbstractPulsarInstrumentation.java
codelipenghui: Will remove it.

apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
codelipenghui: Will remove it.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarConsumerInterceptor.java
codelipenghui: Move it to beforeMethod

dmsolr:
seem like the pulsar-plugin does not need use jdk8, because you don't use any feature of jdk8.

wu-sheng:
> seem like the pulsar-plugin does not need use jdk8, because you don't use any feature of jdk8.

@codelipenghui Please confirm. If so, please change the JDK compile requirement, and this plugin could stay in default, rather than optional. Notice, it is about the plugin source codes, nothing related to Pulsar source code JDK version.

codelipenghui:
@dmsolr @wu-sheng I misunderstood it, will move it back, thanks.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarConsumerInterceptor.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                             MethodInterceptResult result) throws Throwable {
+        if (allArguments.length > 0 && allArguments[0] != null) {
+            ConsumerEnhanceRequiredInfo requiredInfo = (ConsumerEnhanceRequiredInfo) objInst.getSkyWalkingDynamicField();
+            requiredInfo.setStartTime(System.currentTimeMillis());
+            AbstractSpan activeSpan = ContextManager.createEntrySpan(OPERATE_NAME_PREFIX +
+                    requiredInfo.getTopic() + CONSUMER_OPERATE_NAME + requiredInfo.getSubscriptionName(), null)
+                    .start(requiredInfo.getStartTime());
```
wu-sheng: You are putting 
> requiredInfo.setStartTime(System.currentTimeMillis());

and 
> .start(requiredInfo.getStartTime());

In the same method section, the point is? Is `ConsumerEnhanceRequiredInfo#StartTime` really required?

codelipenghui: startTime can remove from `ConsumerEnhanceRequiredInfo`, i moved the section from `afterMethod` to `beforeMethod` but forgot to remove the startTime.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarProducerInterceptor.java
```diff
@@ -0,0 +1,110 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                              Object ret) throws Throwable {
+        if (allArguments.length > 0 && allArguments[0] != null) {
```
wu-sheng: I noticed there is plenty of `if (allArguments.length > 0 && allArguments[0] != null)`. Why do we need this? If we don't want to intercept any arg method, you could use the explicit method arg list in instrumentation definition.

codelipenghui: will update the instance method matcher

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarConsumerInterceptor.java
```diff
@@ -54,12 +54,11 @@
...
     public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
                              MethodInterceptResult result) throws Throwable {
-        if (allArguments.length > 0 && allArguments[0] != null) {
+        if (allArguments[0] != null) {
             ConsumerEnhanceRequiredInfo requiredInfo = (ConsumerEnhanceRequiredInfo) objInst.getSkyWalkingDynamicField();
-            requiredInfo.setStartTime(System.currentTimeMillis());
             AbstractSpan activeSpan = ContextManager.createEntrySpan(OPERATE_NAME_PREFIX +
                     requiredInfo.getTopic() + CONSUMER_OPERATE_NAME + requiredInfo.getSubscriptionName(), null)
-                    .start(requiredInfo.getStartTime());
+                    .start(System.currentTimeMillis());
```
wu-sheng: `#createEntrySpan` has set start time automatically, you could find it in source does.

codelipenghui: Yes, AbstractTracingSpan.start() is already set the startTime

wu-sheng:
This PR should not affect the e2e tests, it seems unstable. I am working with INFRA, hope tests could run in the project sponsor VMs, making it more stable.

wu-sheng:
If this PR is ready at your side, please post your local test result here.

wu-sheng:
@codelipenghui Look like the CI and e2e failure caused by limited resource, es image can't be started in time. I will check this later today. You could focus on your local integration tests only for now.

codelipenghui:
@wu-sheng Ok, thanks. I'm doing the local integration tests now.

apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
```diff
@@ -133,7 +133,10 @@
     public static final OfficialComponent PLAY = new OfficialComponent(68, "Play");
 
     public static final OfficialComponent CASSANDRA_JAVA_DRIVER = new OfficialComponent(69, "cassandra-java-driver");
-
+  
     public static final OfficialComponent LIGHT_4J = new OfficialComponent(71, "Light4J");
 
+    public static final OfficialComponent PULSAR_PRODUCER = new OfficialComponent(73, "pulsar-producer");
```
dmsolr: Is No.72 reserved for others?

wu-sheng: Yes. Definition of pulsar server.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarConsumerInterceptor.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public static final String OPERATE_NAME_PREFIX = "Pulsar/";
+    public static final String CONSUMER_OPERATE_NAME = "/Consumer/";
+
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                             MethodInterceptResult result) throws Throwable {
+        if (allArguments[0] != null) {
+            ConsumerEnhanceRequiredInfo requiredInfo = (ConsumerEnhanceRequiredInfo) objInst.getSkyWalkingDynamicField();
+            AbstractSpan activeSpan = ContextManager.createEntrySpan(OPERATE_NAME_PREFIX +
```
dmsolr: This is quite OK. But I think below better.
···
ContextCarrier contextCarrier = new ContextCarrier();
CarrierItem next = contextCarrier.items();
        while (next.hasNext()) {
    next = next.next();
    next.setHeadValue(request.getHeader(next.getHeadKey()));
}

span = ContextManager.createEntrySpan(“/span/operation/name”, contextCarrier);
···

codelipenghui: Good point, seems can create entry span with carrier.

apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
codelipenghui: I was confused at first, seems missing one of them 😄

codelipenghui:
@wu-sheng I have paste the local integration test result to the description of this PR. Please help review the pulsar scenario integration.

wu-sheng:
@ascrutae Please run tests on our INFRA

wu-sheng:
@codelipenghui Tests are 1 passed, 5 failure. Please check the log, we are running the cases parallel, is that some static port in your case?

codelipenghui:
@wu-sheng Currently, pulsar broker use static port, let me try to fix it.

codelipenghui:
@wu-sheng Currently, pulsar broker using static port, is that the error cause?

wu-sheng:
> Currently, pulsar broker using static port, is that the error cause?

You should bind port by using docker compose. This is the cause.

codelipenghui:
Sorry, i'm lost here. I have bind port by using:

dmsolr:
LGTM

wu-sheng:
Merging

codelipenghui:
@wu-sheng @dmsolr @kezhenxu94 Thanks for your help, very nice experience in Skywalking community.

wu-sheng:
Welcome and look forward your storage proposal.

wu-sheng:
@codelipenghui The plugin test immigration is on going in these two months, a lot of the plugins have moved into PR related test process. #3583

Could any of you move the pulsar plugin test into the main repo too? The original plugin test env has been shut down. Moving into the main repo could make the pulsar plugin checked in every PR.

codelipenghui:
@wu-sheng Thanks for your reminder, will move the pulsar plugin test soon.

PR #3498, opened 2019-09-22
Query server time info

x22x22:

- [ ] Bug fix
- [X] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___

add query server timestamp and timezone info


wu-sheng:
Submodule doesn't update.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetadataQuery.java
```diff
@@ -93,4 +96,13 @@ public EndpointInfo getEndpointInfo(final int endpointId) throws IOException {
     public List<Database> getAllDatabases(final Duration duration) throws IOException {
         return getMetadataQueryService().getAllDatabases();
     }
+
+    public TimeInfo getTimeInfo() {
+        TimeInfo timeInfo = new TimeInfo();
+        SimpleDateFormat timezoneFormat = new SimpleDateFormat("ZZZZZZ");
+        Date date = new Date();
+        timeInfo.setCurrentTimestamp(date.getTime());
```
wu-sheng: Ref https://stackoverflow.com/questions/368094/system-currenttimemillis-vs-new-date-vs-calendar-getinstance-gettime

`System.currentTimeMillis() ` should be faster and good enough. Right?

wu-sheng:
@kezhenxu94 Look like some rule causing this auto labeling failure.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetadataQuery.java
x22x22: how do get timezone?

wu-sheng: Sorry, I miss that.

PR #3510, opened 2019-09-23
Bugfix: endpoint grouping rules may be not unique

kezhenxu94:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance


apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringFormatGroup.java
```diff
@@ -69,7 +71,7 @@ public FormatResult format(String string) {
             '}';
     }
 
-    public class FormatResult {
+    public static class FormatResult {
```
wu-sheng: Why static? Who will initial this class?

kezhenxu94: None, I just think it's a better practice to make it `static` if the inner class doesn't reference the outer class, see [here](https://rules.sonarsource.com/java/tag/performance/RSPEC-2694)

Issue #3512, opened 2019-09-24
"id is not in cache and persistent storage" error


YoungHu:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Bug
- Which version of SkyWalking, OS and JRE?
6.4.0 Linux with JRE 8

- Which company or project?

- What happen?
occur "id is not in cache and persistent storage" error when endpoint_inventory index have more than one record with same sequence.
after check the source code, find the problem at `if (response.getHits().totalHits == 1)`, if want to check whether the result just return one hit or not, we should use `if (response.getHits().getHits().length == 1)`
___
### Requirement or improvement
- check the real hits by use `if (response.getHits().getHits().length == 1)`

wu-sheng:
Read the reply on pull request.

PR #3513, opened 2019-09-24
fix the issue #3512

YoungHu:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
#3512 
___
### Bug fix
- bug description
 see the issue desc

- How to fix?
  see the code change


wu-sheng:
What is the difference? Could you post the document of Elasticsearch?

dmsolr:
The `totalHits` means the number of hit documents. The `getHits().length` means the numbe of recalled documents.

wu-sheng:
Thanks for explanation. I would say both ways should be same in the theory. Because ID should be unique in the global always.

wu-sheng:
Based on the design, the id is unique, then if hint more than one is being sure a bug. And this change could cover it up. I would recommend don't change in this way. But try to find out there are duplicated IDs.

So, I tagged this as won't fix and is closing this.

PR #3525, opened 2019-09-25
Refactor web hook callback to SPI

kezhenxu94:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

we have many issues asking how to set alarm notifications to third party systems, but for now we only support web hook, web hook is absolutely useful, but IMHO (as a SkyWalking user), since we already have a server(s) (OAP) running, why bother to set up another http server just for a web hook and proxy the alarm message to third-party systems.

So here I propose refactoring the `AlarmCallback` to SPI, and make the web hook as one of the implementations (also default implementation), and thus we provide possibilities for our users to provide some third-party system implementations, such as Email, Dingtalk, WeChat, Slack or anything else, although they can't be included in the main repo, they can be placed in the SkyAPM org, WDYT? @wu-sheng

### DIFF
--- a/oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/AlarmModuleProvider.java
+++ b/oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/AlarmModuleProvider.java
@@ -19,6 +19,10 @@
 package org.apache.skywalking.oap.server.core.alarm.provider;
 
 import java.io.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.ServiceLoader;
+
 import org.apache.skywalking.oap.server.core.CoreModule;
 import org.apache.skywalking.oap.server.core.alarm.*;
 import org.apache.skywalking.oap.server.library.module.*;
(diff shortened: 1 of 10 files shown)


wu-sheng:
One major question
How to resolve the extension capability issue based on the container image? You could notice there are plenty of integration solutions are using network protocol to replace SPI, because in Cloud Native, container is not very friendly to SPI mode

And, do you read the latest English blog in our website? There is a DingTalk webhook forwarder open source project.

kezhenxu94:
> One major question
> How to resolve the extension capability issue based on the container image? You could notice there are plenty of integration solutions are using network protocol to replace SPI, because in Cloud Native, container is not very friendly to SPI mode

This patch did not remove the web hook implementation and thus it won't be worse than before, users can still use the web hook **out of box**, but it provides convenience for users that don't use containers.

> 
> And, do you read the latest English blog in our website? There is a DingTalk webhook forwarder open source project.

It's a forwarder, meaning that users should start one more process (maybe more, I'm not sure), with this patch, users have possibilities to just copy and paste a `jar` file(will be provided if this PR gets merged) into the `oap-libs`, and the third-party alarm system is supported with minimum effort.

Are there other concerns?

wu-sheng:
Concern no. But I am still stragling to be open to accept SPI style PR. We are using network integration today. I am afraid people begin to contribute SPI implementation, which cause more dependencices, without clear advantage

Issue #3540, opened 2019-09-28
Oap service start failed


g29times:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [X] Question or discussion
- [ ] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?
Running /bin/startup.sh in default config using H2, log shows started failed, console shows success.
![errorlog.png](http://ww1.sinaimg.cn/large/466e120egy1g7f8mkeh9ej20vw0brmyc.jpg)
___
### Bug
- Which version of SkyWalking, OS and JRE?

- Which company or project?

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.
"org.apache.skywalking.oap.server.library.module.ModuleStartException: ScopeDefine fieldName = All not found."
___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

dmsolr:
`startup.sh` prints `success` that means it has launched the processes. 

According to the message, I think you have to initialize.

kezhenxu94:
@g29times if you're using JDK 8+, it's a known issue, because the recommended JDK version for OAP backend is JDK 8

kezhenxu94:
@g29times if you're interested, please take a look at #3541 and see whether it resolves your problem

Issue #3477, opened 2019-09-17
org.apache.skywalking.oap.server.starter.OAPServerStartUp - 55 [main] ERROR [] - ScopeDefine fieldName = All not found.


liangxp:
- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug

___
### Question：start exception
 `
2019-09-17 16:23:10,069 - org.apache.skywalking.oap.server.starter.OAPServerStartUp - 55 [main] ERROR [] - ScopeDefine fieldName = All not found.
org.apache.skywalking.oap.server.library.module.ModuleStartException: ScopeDefine fieldName = All not found.
	at org.apache.skywalking.oap.server.core.CoreModuleProvider.prepare(CoreModuleProvider.java:95) ~[server-core-6.4.0.jar:6.4.0]
	at org.apache.skywalking.oap.server.library.module.ModuleDefine.prepare(ModuleDefine.java:94) ~[library-module-6.4.0.jar:6.4.0]
	at org.apache.skywalking.oap.server.library.module.ModuleManager.init(ModuleManager.java:52) ~[library-module-6.4.0.jar:6.4.0]
	at org.apache.skywalking.oap.server.starter.OAPServerStartUp.main(OAPServerStartUp.java:43) [server-starter-6.4.0.jar:6.4.0]
Caused by: org.apache.skywalking.oap.server.core.UnexpectedException: ScopeDefine fieldName = All not found.
	at org.apache.skywalking.oap.server.core.source.DefaultScopeDefine.valueOf(DefaultScopeDefine.java:152) ~[server-core-6.4.0.jar:6.4.0]
	at org.apache.skywalking.oal.rt.parser.OALListener.enterSource(OALListener.java:52) ~[oal-rt-6.4.0.jar:6.4.0]
	at org.apache.skywalking.oal.rt.grammar.OALParser$SourceContext.enterRule(OALParser.java:605) ~[oal-grammar-6.4.0.jar:6.4.0]
	at org.antlr.v4.runtime.tree.ParseTreeWalker.enterRule(ParseTreeWalker.java:42) ~[antlr4-runtime-4.7.1.jar:4.7.1]
	at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:25) ~[antlr4-runtime-4.7.1.jar:4.7.1]
[...]

PR #3541, opened 2019-09-28
Upgrade guava version to support higher JDK version

kezhenxu94:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

#3540 and #3477 

https://github.com/apache/skywalking/blob/5b039e9747fac98444b4ec423a5d57524202f286/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/annotation/AnnotationScan.java#L54-L56

the `classpath.getTopLevelClassesRecursive("org.apache.skywalking")` returns empty `set` under JDK 8+, and it turns out to be an issue of Guava, see https://github.com/google/guava/issues/3249#issuecomment-416234877 , this patch upgrades the Guava version to 23.1, and I've verified locally the exception disappear and OAP starts successfully

### DIFF
--- a/apm-dist/release-docs/LICENSE
+++ b/apm-dist/release-docs/LICENSE
@@ -231,7 +231,7 @@ The text of each license is the standard Apache 2.0 license.
     Google: grpc 1.15.1: https://grpc.io/ , Apache 2.0
     Google: gprc-java 1.14.0: https://github.com/grpc/grpc-java, Apache 2.0
     Google: gprc-java 1.15.1: https://github.com/grpc/grpc-java, Apache 2.0
-    Google: guava 20.0: https://github.com/google/guava , Apache 2.0
+    Google: guava 23.1: https://github.com/google/guava , Apache 2.0
     Google: guice 4.1.0: https://github.com/google/guice , Apache 2.0
     Google: gson 2.8.1: https://github.com/google/gson , Apache 2.0
     Google: opencensus-java 0.12.3: https://github.com/census-instrumentation/opencensus-java , Apache 2.0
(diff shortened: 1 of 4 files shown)


wu-sheng:
Maybe CI found the issue.

kezhenxu94:
I've replayed the e2e tests under JDK 1.9 manually (which won't report check status back to this PR), let's wait and see the result, if it passes, I think it should **generally work on 1.9 and above**, https://builds.apache.org/job/skywalking-e2e-2/763/console

wu-sheng:
If you see it passed, reply me here. I am using phone only, today.

kezhenxu94:
@wu-sheng it passes under JDK 9, see the selected/highlighted words below the screenshot
image

wu-sheng:
Cool. How about higher? 11 for example? I am thinking about we need JRE documentation update.

kezhenxu94:
> Cool. How about higher? 11 for example? I am thinking about we need JRE documentation update.

I tested this afternoon locally with OpenJDK 12, it works fine too, so it's highly possible that it also works on JDK 10 and JDK11, do we need to replay this on the infra Jenkins?

wu-sheng:
The infra is setting up our VM. Please submit an issue to track the higher jdk tests.

In this PR, at least, we should update document.

kezhenxu94:
Oops, I've made a mistake that the container in which our e2e runs is based on OpenJDK8, the JDK version in `Jenkinsfile-E2E` is just for compile time only, so we'll need some real tests

wu-sheng:
OK, so we haven't verified in 9+?

kezhenxu94:
> OK, so we haven't verified in 9+?

I verified JDK12 locally, others are not verified yet

kezhenxu94:
We may need much more resources if we want to verify every commit on those versions

wu-sheng:
OK. I think 9 and 11 are LTS versions, right? We could focus on that.

PR #3549, opened 2019-09-29
update official docker repos, moved to apache org

rdjogo:

`docker-compose.yml` refers to no longer existing docker hub repos, as they moved to Apache Docker Hub organization.
Replaced with correct OAP server and UI docker image repo references.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -30,7 +30,7 @@ services:
         soft: -1
         hard: -1
   oap:
-    image: skywalking/oap
+    image: apache/skywalking-oap-server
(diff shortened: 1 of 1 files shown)


wu-sheng:
That docker file is for your local test only. We indicated it in readme.

You could file the docker compose in skywalking-docker repo.

hanahmily:
docker-compose.yml is to run image built by /docker folder located at main repo(github.com/apache/skywalking). the officail docker-compose.yml is released in https://github.com/apache/skywalking-docker/tree/master/6/6.4/compose

Issue #2438, opened 2019-04-03
Increase the test coverage rate


wu-sheng:
Today, SkyWalking's tests are mostly around integration tests, which are benifit for end users. Now we hope through this, people from community could help on Unit Tests.

JUnit and Mockito are provided. This is a good start for everyone wants to contribute codes. When you submit test cases PR, please ref to this issue.

wu-sheng:
Also, in this kind of pull request, we don't limit how many test cases you provide. Even one new test case, which could effectively test one more line, it will be acceptable. Just try your best to make tests more sense.

flycash:
Where can I find the code coverage report?

wu-sheng:
The coverage is here, https://coveralls.io/github/apache/skywalking

wu-sheng:
Update here, we are only just increasing the test case coverage, we are doing more.

- [x] @kezhenxu94 set up the Jenkins of ASF for power test env.
- [ ] @ascrutae lead the agent test in Jenkins jobs of every PR when necessary
- [ ] @kezhenxu94 lead the e2e test, make agent/backend integration test done in every PR.

wu-sheng:
We have e2e tests fro single node w/ H2 and cluster mode w/ ES. We are trying to build more. Including MySQL storage tests, alarm tests, and aggregation accuracy tests.

wu-sheng:
After #3561, we have 30%+ coverage now.

wu-sheng:
~~@ascrutae lead the agent test in Jenkins jobs of every PR when necessary~~

This task is lead by @dmsolr on #3528. Ownership replaced.

wu-sheng:
Agent autotest framework has been provided, and 300+ test cases have been added. Seems working well. Just working for @dmsolr to provide the official documents, rather than guide users to do that one by one.

wu-sheng:
# Status update
All important tests have been provided in the master branch and PR process already. This issue will keep open if any contributor wants to help about the Unit Tests.

PR #3552, opened 2019-09-30
test: add service reset command test case

jsbxyyx:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
#2438 
___
### Bug fix
- Bug description.

- How to fix?

___


wu-sheng:
Please bring more test cases to us. And include more than 10 test cases in one pull request, which makes review easier.

jsbxyyx:
@wu-sheng I want to familiarize myself with the process of the skywalking community.

jsbxyyx:
@wu-sheng I want to get familiar with the submission process first.

wu-sheng:
I know. I just hope you could bring more in next time. This one is waiting for pmc review.

jsbxyyx:
@wu-sheng the java source's liencse can be automatic generated？next time, I submit more test case.

wu-sheng:
Yes. IDEA has license header plugin. And we have a header file in root path.

jsbxyyx:
I want the automatic generated on maven life cycle
```
<plugin>
	<groupId>com.mycila</groupId>
	<artifactId>license-maven-plugin</artifactId>
	<version>${license-maven-plugin.version}</version>
	<executions>
		<execution>
			<phase>generate-sources</phase>
			<goals>
				<goal>remove</goal>
				<goal>format</goal>
			</goals>
		</execution>
	</executions>
	<configuration>
		<quiet>true</quiet>
		<header>${user.dir}/style/copyright</header>
		<includes>
			<include>**/src/main/java/**</include>
			<include>**/src/test/java/**</include>
		</includes>
		<excludes>
			<exclude>**/generated/**</exclude>
		</excludes>
		<strictCheck>true</strictCheck>
		<mapping>
			<java>SLASHSTAR_STYLE</java>
		</mapping>
	</configuration>
</plugin>
```

wu-sheng:
License is an IP declarence. I prefer people take the responsibilities of their pull request codes, rather than add them randomly.

jsbxyyx:
OK, I understand. thanks

kezhenxu94:
automatically generating the license header is not realistic, if you want to automatically generate license header, you don't even need a plugin, the IDEA has such ability to include a license when creating a new class (will attach a screenshot later 
 if you need ), the reason why it's not acceptable to generate license header in maven lifecycle is as follows:
We check the license header in maven lifecycle, and report errors if it doesn't adhere to Apache's policy, but if you generate the license header in maven lifecycle too, it hides the errors and the generated license header won't be included in the source codes, which is dangerous for an Apache project.

apm-protocol/apm-network/src/test/java/org/apache/skywalking/api/network/trace/component/command/ServiceResetCommandTest.java
```diff
@@ -0,0 +1,72 @@
...
+    public void test_serialize() throws Exception {
+        ServiceResetCommand command = new ServiceResetCommand("aa");
+        Command.Builder builder = command.serialize();
+
+        Assert.assertEquals(ServiceResetCommand.NAME, builder.getCommand());
+        Assert.assertEquals("aa", builder.getArgs(0).getValue());
+    }
+
+    @Test
+    public void test_deserialize() throws Exception {
```
kezhenxu94: I suggest splitting this test case into two cases, and the names can be more semantic such as "should_throw_when_serial_num_is_null" or "should_deserialize", which is more easy for future contributors to diagnose why they break the test and how to adjust the case to its expected goal, it's the same for your future contributions. WDYT @wu-sheng

wu-sheng: +1 for making test case more clear. @jsbxyyx  please follow

jsbxyyx: OK, Your suggestion is very good.

jsbxyyx:
@kezhenxu94 I understand your intentions about license.

apm-protocol/apm-network/src/test/java/org/apache/skywalking/api/network/trace/component/command/ServiceResetCommandTest.java
wu-sheng: Hi, one reminder, using CamelCase rather than underline. We should forbit underline in method name. Contribution about this rule is welcome.
 too.

jsbxyyx: How do you see this name? `methodName_DoesWhat_WhenTheseConditions`

kezhenxu94: > How do you see this name? `methodName_DoesWhat_WhenTheseConditions`

Looks good to me but we prefer camel case in the project as @wu-sheng suggests, so please use the camel case style

And if the method name becomes too long or hard to describe, consider using an assert method with `message`, like `org.junit.Assert#assertEquals(java.lang.String, java.lang.Object, java.lang.Object)`, thanks :)

wu-sheng:
I am merging this case, but, @jsbxyyx please keep the test case name as short and clear as possible, and don't use `_` if possible. Seems too long to me.

jsbxyyx:
@wu-sheng ok

PR #3554, opened 2019-10-01
Exclude generated sources from test coverage

kezhenxu94:
Turns out `**/generated-sources/**` doesn't exclude the generated sources correctly, this patch excludes the same sources as `checkstyle` plugin


kezhenxu94:
Expectation is that the coverage should be higher than `7.363%` (current coverage)

wu-sheng:
2% increased.

PR #3564, opened 2019-10-06
Instead of datasource-settings.properties, use application.yml for MySQLStorageProvider

tristaZero:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?
1.Create DynamicModuleConfig
2.Create MySQLStorageConfig
3.Modify ModuleDefine for using ModuleConfiguration.
4.Add all properties of HikariCP to application.yml
5.Delete datasource-settings.properties
___


oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/mysql/MySQLStorageProvider.java
```diff
@@ -66,15 +96,8 @@ public MySQLStorageProvider() {
...
-        Properties settings = new Properties();
-        try {
-            settings.load(ResourceUtils.read("datasource-settings.properties"));
-        } catch (IOException e) {
-            throw new ModuleStartException("load datasource setting file failure.", e);
-        }
-
-        mysqlClient = new JDBCHikariCPClient(settings);
+    @Override public void prepare() throws ServiceNotProvidedException {
+        mysqlClient = new JDBCHikariCPClient(getConfigProperties());
```
kezhenxu94: This is the only benefit of introducing interface `DynamicModuleConfig`, right? I don't think it's worthwhile, WDYT @wu-sheng

wu-sheng: I have replied, basically, agree with @kezhenxuke. 

@tristaZero  If you want a properties field, read my suggestion.

wu-sheng:
@tristaZero read the dynamic configuration in backend advanced feature. You will find out why.

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/mysql/MySQLStorageProvider.java
tristaZero: Actually, maybe `DynamicModuleConfig` is not a good name, but i want to say `Dynamic` means we can make the properties of ModuleConfig diverse. For example, like H2StorageConfig, in current H2StorageConfig, only four properties can be set, no more properties. But there are many other properties for `JDBCHikariCPClient`, but H2StorageConfig can not provide.

So look at the `MySQLStorageConfig` implemeting the interface `DynamicModuleConfig`, which only has one member named `properties`, we can put all the of settings into this member, it means `MySQLStorageConfig` can provide any property for `JDBCHikariCPClient`

In future, we don't need to define some important settings as members of `ModuleConfig` previously. Only by implementing `DynamicModuleConfig`, certain `ModuleConfig`, e.g, `ModuleConfig` or `StorageModuleElasticsearchConfig` can provide any setting for their `ModuleProvider`s.

Moreover i think if `ModuleConfig` becomes an interface, that will be better. But this change is big, i don't want to do it togather. And this issue also needs to disscuss in our community.

tristaZero: @wu-sheng @kez

tristaZero:
> I think, you should make Module configuration injection to support properties or map. Using DynamicModuleConfig has a limit, if you have two properties typd fields.
> 
> And making field type extension should be more elegance.
> Make sense?
> 
> And you should raise a MySQL e2e PR to across verify this.

What sceneria two properties is necessary? Why don't we put the settings from two properties into one properties?

oap-server/server-library/library-module/src/main/java/org/apache/skywalking/oap/server/library/module/DynamicModuleConfig.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package org.apache.skywalking.oap.server.library.module;
+
+import java.util.Properties;
+
+/**
+ * @author panjuan
+ */
+public interface DynamicModuleConfig {
```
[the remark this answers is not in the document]
tristaZero: @kezhenxu94 Hi, not only for MySQLStorageConfig, any `ModuleConfig` which want to own uncertain members or settings can implement `DynamicModuleConfig`. For example, 28 setting items + so many properties of JDBC dataSource for `JDBCHikariCPClient`... It is difficult for us to list all those settings, some of which i never used in certain `ModuleConfig`. We need the members or properties of `ModuleConfig`  become dynamic, no zombie, for in some scenerios, we can not enumerate all the settings or properties.

wu-sheng:
I said this in design level. ModuleConfig injection is one of oap core features. This is field type triggered, so when you want properties field, I prefer this way.

For example, you could support properties to match the yaml map type.

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/mysql/MySQLStorageProvider.java
wu-sheng: I know the implementation ia good enough for this feature, but I think @kezhenxkezhand I are arguing about the whole oap design. And these two belong different module, core and mysql storage provider. 
We should not change core just for one module provider, because it is not worth.

wu-sheng:
This PR could be separated into 3 PRs
1. Config injection supports map and properties type
1. New MySQL config implementation
1. e2e for mysql storage.

wu-sheng:
One remind, once you change the mysql config, you need to change this too.
https://github.com/apache/skywalking/blob/81f4c087b01bee72ddd79b2dfb9c995359b5708e/docker/oap/docker-entrypoint.sh#L134-L140

FYI @kezhenxu94 Do you have any idea, how to make the docker-compose checked in CI?

tristaZero:
> I said this in design level. ModuleConfig injection is one of oap core features. This is field type triggered, so when you want properties field, I prefer this way.
> 
> For example, you could support properties to match the yaml map type.

The properties of one `ModuleConfig`, e.g, `MySQLModuleConfig`, set in application.yml are like this way：
```
storege:
  mysql:
    jdbcUrl: ${SW_JDBC_URL:"jdbc:mysql://localhost:3306/swtest"}
    dataSource.user: ${SW_DATA_SOURCE_USER:root}
    dataSource.password: ${SW_DATA_SOURCE_PASSWORD:root@1234}
    dataSource.cachePrepStmts: ${SW_DATA_SOURCE_CACHE_PREP_STMTS:true}
    dataSource.prepStmtCacheSize: ${SW_DATA_SOURCE_PREP_STMT_CACHE_SQL_SIZE:250}
    dataSource.prepStmtCacheSqlLimit: ${SW_DATA_SOURCE_PREP_STMT_CACHE_SQL_LIMIT:2048}
    dataSource.useServerPrepStmts: ${SW_DATA_SOURCE_USE_SERVER_PREP_STMTS:true}
    metadataQueryMaxSize: ${SW_STORAGE_H2_QUERY_MAX_SIZE:5000}
```
So one properties can contain all the settings below `mysql:`, do we need another properties? Fow what?

wu-sheng:
I am not saying in MySQL provider, you need two or more properties fields. Clearly you don't.
But you should take a look at what packages are you placing your implementation
- **org.apache.skywalking.oap.server.library.module**.DynamicModuleConfig
- **org.apache.skywalking.oap.server.library.module**. ModuleDefine

Those two are in **core** module, which means, being satisfied for MySQL implementation is not enough. That is what I mean in design level. You never know which provider will use two properties, even sometimes they are not open source implementation in the upstream.

So, once we want to support properties type, which makes totally sense to me, we should bring a good and elegance feature in the **core**.

Is this clear for you now?

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/mysql/MySQLStorageProvider.java
tristaZero: Hi, you're right. The interface `DynamicModuleConfig` is a big change for he whole oap design. If it's just for `MySQLModuleConfig`, it is not worthy. My initial thought is that for those `ModuleConfig` made up with certain members can implement interface `ModuleConfig`, while other `ModuleConfig` maybe own uncertain members, e.g, the settings of DataSource JDBC can implement interface `DynamicModuleConfig`.

Actually, if `H2ModuleConfig` and `StorageModuleElasticsearchConfig` can implement interface `DynamicModuleConfig`, then the implementation of their `ModuleProvider` will become easier, for you don't worry about how to enumerate and set value for their incalculable properties. But we can consider this issue later.

In future, some of `ModuleConfig` whose properties can not be enumerated easily can also implement `DynamicModuleConfig`.
@wu-sheng @kezhenxu94

docs/en/setup/backend/backend-storage.md
```diff
@@ -172,10 +178,18 @@ Active TiDB as storage, set storage provider to **mysql**.
...
 storage:
   mysql:
+    properties:
+      jdbcUrl: ${SW_JDBC_URL:"jdbc:mysql://localhost:3306/swtest"}
+      dataSource.user: ${SW_DATA_SOURCE_USER:root}
+      dataSource.password: ${SW_DATA_SOURCE_PASSWORD:root@1234}
+      dataSource.cachePrepStmts: ${SW_DATA_SOURCE_CACHE_PREP_STMTS:true}
+      dataSource.prepStmtCacheSize: ${SW_DATA_SOURCE_PREP_STMT_CACHE_SQL_SIZE:250}
+      dataSource.prepStmtCacheSqlLimit: ${SW_DATA_SOURCE_PREP_STMT_CACHE_SQL_LIMIT:2048}
+      dataSource.useServerPrepStmts: ${SW_DATA_SOURCE_USE_SERVER_PREP_STMTS:true}
```
kezhenxu94: We could mention that although not all of the properties are listed here, users can add them as long as they are supported by Hikaricp

wu-sheng: This comment should keep in thw document.

oap-server/server-starter/src/main/java/org/apache/skywalking/oap/server/starter/config/ApplicationConfigLoader.java
```diff
@@ -61,21 +62,25 @@ private void loadConfig(ApplicationConfiguration configuration) throws ConfigFil
...
                             Properties properties = new Properties();
                             if (propertiesConfig != null) {
-                                propertiesConfig.forEach((key, value) -> {
-                                    properties.put(key, value);
-                                    final Object replaceValue = yaml.load(PropertyPlaceholderHelper.INSTANCE
-                                        .replacePlaceholders(value + "", properties));
-                                    if (replaceValue != null) {
-                                        properties.replace(key, replaceValue);
+                                propertiesConfig.forEach((propertyName, propertyValue) -> {
+                                    if (propertyValue instanceof Map) {
```
wu-sheng: I think Map and Properties are good choices for here.

wu-sheng: And is this covered by unit test?

kezhenxu94:
@tristaZero and it seems that you forgot the reminder [above](https://github.com/apache/skywalking/pull/3564#issuecomment-538744984)

oap-server/server-starter/src/main/java/org/apache/skywalking/oap/server/starter/config/ApplicationConfigLoader.java
tristaZero: Hi, where should i add unit test?

tristaZero:
> Do we have two datasource-settings.properties files? I remember so. Only find one deleted, please confirm.

Thanks for your reminder, i deleted it.

tristaZero:
> @tristaZero and it seems that you forgot the reminder [above](https://github.com/apache/skywalking/pull/3564#issuecomment-538744984)

Thanks, fixed it.

docker/oap/docker-entrypoint.sh
```diff
@@ -136,7 +136,7 @@ generateStorageMySQL() {
     cat <<EOT >> ${var_application_file}
 storage:
   mysql:
-    metadataQueryMaxSize: \${SW_STORAGE_H2_QUERY_MAX_SIZE:5000}
+    metadataQueryMaxSize: \${SW_STORAGE_MYSQL_QUERY_MAX_SIZE:5000}
```
wu-sheng: Why this didn't match `application.yml`?  This is the yml generator written by shell, which will be used in container. If you remove other, than there is no chance to override config by system env, like `SW_STORAGE_MYSQL_QUERY_MAX_SIZE`. You should keep recommended config in application.yml and here. Value is configurable, but structure should be here.

docs/en/setup/backend/backend-storage.md
```diff
@@ -159,11 +159,18 @@ Please download MySQL driver by yourself. Copy the connection driver jar to `oap
...
+    metadataQueryMaxSize: ${SW_STORAGE_MYSQL_QUERY_MAX_SIZE:5000}
 ```
-
-All connection related settings including link url, username and password
-are in `datasource-settings.properties`. 
-This setting file follow [HikariCP](https://github.com/brettwooldridge/HikariCP) connection pool document.
+All connection related settings including link url, username and password are in `application.yml`. 
+Here are some of the settings, please follow [HikariCP](https://github.com/brettwooldridge/HikariCP) connection pool document for all the settings.
```
wu-sheng: @tristaZero You should be familiar with HikariCP config and JDBC settings, right? More professional than me. Please put all the necessary and recommended setting items here. The user will change it later.

oap-server/server-starter/src/main/java/org/apache/skywalking/oap/server/starter/config/ApplicationConfigLoader.java
wu-sheng: What do you mean where? Same module, test folder, same package, ClassName + Test. As usual. We don't have a special rule for this.

tristaZero: You mean create a new unit test class, and write unit class for ...loadConfig()?
@wu-sheng

docker/oap/docker-entrypoint.sh
tristaZero: Hi, It means all the settings here should match with settings for MySQL in `application.yml`?

docs/en/setup/backend/backend-storage.md
tristaZero: So...the setting for dataSource..is not enough, or appropriate? Those settings are originally from `datasource-settings.properties`. From my side, those settings are enough for dataSource, if user need more setting items, he can refer to HikariCP document.

docker/oap/docker-entrypoint.sh
wu-sheng: Yes. Because at here, this script is generating that file.

docs/en/setup/backend/backend-storage.md
wu-sheng: If you think they are enough in most cases, I am good about this too. If anyone needs more, they can add by themselves, or send a pull request to us.

oap-server/server-starter/src/main/java/org/apache/skywalking/oap/server/starter/config/ApplicationConfigLoader.java
wu-sheng: ApplicationConfigLoader doesn't have Unit Test today, please add one.

tristaZero:
> Basically, this PR LGTM. @kezhenxu94 What do you think?
> 
> I expect a new e2e test case added to test MySQL storage, but before doing that, I hope you could separate our e2e into 2 groups rather than 4, even 5(if MySQL added), which cost unpredictable CPU in CI env. Agree?

 Is #3583 the same one with what your mentioned?

wu-sheng:
> Is #3583 the same one with what your mentioned?

That is only plugin test. I mean e2e test, check `test/e2e` folder and this doc
https://github.com/apache/skywalking/blob/master/docs/en/guides/README.md#end-to-end-tests-e2e-for-short

wu-sheng:
I will wait for the unit test for the new properties field, then I think the MySQL e2e could start in another PR.

wu-sheng:
> I'm slightly refactor the e2e running script in my ES7 adaptation PR, as what you expect (2 groups)

Sure. Look forward to your new PR about this feature. I think the community will like this.

oap-server/server-starter/src/main/java/org/apache/skywalking/oap/server/starter/config/ApplicationConfigLoader.java
tristaZero: Finished it.

PR #3567, opened 2019-10-07
Fix some miscellaneous bugs

kezhenxu94:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/expression/EqualMatch.java
```diff
@@ -39,19 +41,19 @@ public boolean match(double left, double right) {
...
 
     public boolean match(Integer left, Integer right) {
-        return left == right;
+        return Objects.equals(left, right);
     }
 
     public boolean match(Long left, Long right) {
-        return left == right;
+        return Objects.equals(left, right);
     }
```
kezhenxu94: `Integer`/`Long` is not the same `Object` when `>128` or so, and should be compared via `equals`, tricky, huh

PR #3570, opened 2019-10-08
test: add oap-server core alarm package test case

jsbxyyx:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
#2438 
___
### Bug fix
- Bug description.

- How to fix?

___


oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/alarm/AlarmRecordTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Test
+    public void builderNPEMap2DataVerify() throws Exception {
+        AlarmRecord.Builder builder = new AlarmRecord.Builder();
+
+        try {
+            builder.map2Data(null);
+        } catch (NullPointerException e) {
+            Assert.assertFalse(false);
+        }
```
kezhenxu94:
```java
        try {
            builder.map2Data(null);
            Assert.fail("Should throw");
        } catch (NullPointerException e) {
        }
```

oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/alarm/AlarmRecordTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+        Map<String, Object> dbMap = new HashMap<>();
+        try {
+            builder.map2Data(dbMap);
+        } catch (NullPointerException e) {
+            Assert.assertFalse(false);
```
kezhenxu94:
```suggestion
            Assert.fail("Should never happen");
```

oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/alarm/AlarmModuleTest.java
```diff
@@ -0,0 +1,42 @@
...
+ * @author jsbxyyx
+ */
+public class AlarmModuleTest {
+
+    @Test
+    public void verifyName() throws Exception {
+        AlarmModule alarmModule = new AlarmModule();
+        String name = alarmModule.name();
+        Assert.assertEquals(AlarmModule.NAME, name);
+    }
```
kezhenxu94: I don't think this is necessary

jsbxyyx:
@kezhenxu94 I thought we were using junit4，if using junit5 then the test case will a lot simple.

kezhenxu94:
> @kezhenxu94 I thought we were using junit4，if using junit5 then the test case will a lot simple.

Even we're using JUnit4, the codes should not necessarily be that complicated and verbose, do we have plan to migrate to JUnit5 @wu-sheng

jsbxyyx:
@kezhenxu94 I was wrong, if using mockito the test case will a lot simple

wu-sheng:
I have no plan for `JUnit5`. What do you expect from it?

wu-sheng:
One thing needs to add, CI with IT tests, we more recommend real Integration tests to make API using in real cases. Many structure codes are not needed to test, such as provider, module definition and module config class. 

One core target, making core better, don't chase the number.

wu-sheng:
For test interests, try to make coverage of receiver(s) better, mock the core module, and give the receiver some input, verify the source flowing into the core. Those are very meaningful tests for your time.

jsbxyyx:
I agree

jsbxyyx:
Which modules have higher priority for use cases

wu-sheng:
Trace receiver

wu-sheng:
Don't set test case to the method one by one. I know some books tell so, but I and many SkyWalking PMC  don't believe that clearly @kezhenxu94 thinks so too. We are doing white box tests for workflow.

jsbxyyx:
@wu-sheng @kezhenxu94 OK, I understand.

jsbxyyx:
> The method names is not as what we discussed in [#3552 (comment)](https://github.com/apache/skywalking/pull/3552#discussion_r329522793)
> 
> I'm ok with this PR @wu-sheng , considering that @jsbxyyx you will contribute more test, please pay special attention to what we've discussed, :) thanks

Before, The conclusion is not use `_` ? Is it like this `methodNameDoesWhatWhenTheseConditions`?

kezhenxu94:
> > The method names is not as what we discussed in [#3552 (comment)](https://github.com/apache/skywalking/pull/3552#discussion_r329522793)
> > I'm ok with this PR @wu-sheng , considering that @jsbxyyx you will contribute more test, please pay special attention to what we've discussed, :) thanks
> 
> Before, the conclusion is do not use `_` ? style is it like this `methodNameDoesWhatWhenTheseConditions`?

Not using `_` is mandatory, as for method names, I'm ok as long as it can tell the purpose of that test case, the most important point is as what @wu-sheng said above

kezhenxu94:
Please open new PR to do the tests, thanks

jsbxyyx:
thanks

PR #3581, opened 2019-10-09
Check package name org.apache.skywalking

kezhenxu94:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

to avoid https://github.com/apache/skywalking/pull/3575#discussion_r332974932


wu-sheng:
The generated cores are being checked. That is why CI fails.

PR #3528, opened 2019-09-26
test/plugin autotest 

dmsolr:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#2905 
___
### New feature or improvement
- Describe the details and related test reports.

This PR works base on #2905 which @ascrutae does. The main purpose is to move the plugin-autotest back main repository.

1. There is a httpclient-scenario only here.
2. I don't add this to Jenkins's stages yet.
3. After this, I will add more scenarios for the tests.

Issue #3583, opened 2019-10-10
Plugin test-cases immigrate plan


dmsolr:
This issue is intended to track to immigrate every plugin test-cases back main repo. The plugin test development guide doc is https://github.com/apache/skywalking/blob/master/docs/en/guides/Plugin-test.md

## done list
- [x] httpclient-4.3.x-scenario
- [x] gateway-scenario
- [x] dubbo-scenario
- [x] solrj-scenario
- [x] spring-async-scenario
- [x] spring-scenario
- [x] jetty-scenario
- [x] kafka-scenario
- [x] canal-scenario
- [x] postgresql-scenario
- [x] sofarpc-scenario
- [x] mongodb-3.x-scenario
- [x] elasticsearch-5.x-scenario
- [x] undertow-routing-scenario
- [x] undertow-scenario
- [x] apm-toolkit-trace-scenario
- [x] ~~dubbox-scenario~~. **Dubbox has been merged into Dubbo officially, the test case is not required.**
- [x] httpasyncclient-scenario
- [x] webflux-scenario
- [x] customize-scenario
- [x] redisson-scenario
- [x] jedis-scenario
- [x] ~~jdk-cross-thread-scenario~~ **This test has been included in apm-toolkit-trace-scenario**.
- [x] feign-scenario
- [x] rabbitmq-scenario
- [x] grpc-scenario
- [x] hystrix-scenario
- [x] okhttp-scenario
- [x] mysql-scenario
- [x] gson-scenario
- [x] pulsar-scenario
- [x] oracle-scenario
- [x] jdk-http-scenario
- [x] cassandra-java-driver-3.x-scenario
- [x] lettuce-scenario
- [x] servicecomb-0.x-scenario
- [x] zookeeper-scenario
- [x] shardingsphere-scenario
- [x] play-scenario
- [x] activemq-scenario
- [x] h2-struts-scenario
- [x] vertx-eventbus-3.x-scenario
- [x] vertx-web-3.x-scenario

## todo list
- [ ] header-scenario
- [ ] light4j-1.6.x-scenario
- [ ] motan-scenario
- [ ] resteasy-scenario
- [ ] rocketMQ-scenario
- [ ] spring-with-qualified-op-name-scenario
- [ ] spring-with-request-method-op-name-scenario

arugal:
I want to be part of it

dmsolr:
1. HttpClient #3579

dmsolr:
> I want to be part of it

[...]

PR #3584, opened 2019-10-10
[test/plugin] support jvm-container

dmsolr:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#3528 
___

It make plugin test-cases immigrated easy!


wu-sheng:
I think we should immigrate an easy test case in this PR to verify. What do you think?

dmsolr:
Yes. I will push a jetty-scenario.

wu-sheng:
Cool. Thanks.

wu-sheng:
Related to #3583

wu-sheng:
Congrats. It passed in 35mins, and cost per VM 30% CPU. Make sense to me. And we could add more to group1(Group2 is already 30min, longest path). Due to its timeout last time, please make timeout to 60s for each case.

Also, let's add multiple node parallel for compose case, to make the test framework structure completed.

PR #3587, opened 2019-10-10
[test/plugin] rename the package of testcases

dmsolr:

- [x] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#3584 

We have to use this package name, `org.apache.skywalking.apm.testcase.httpclient`. 
We don't need to type very long code for profile anymore. (This is @mrproliu 's idea. Thanks)


dmsolr:
FYI plugin contributors.

PR #3597, opened 2019-10-11
up ui

Indifer:
fixbug: Reduce of empty array with no initial value


wu-sheng:
You need to solve the conflict.

Indifer:
> You need to solve the conflict.

No conflict was detected

wu-sheng:
I changed the title, you should follow this pattern next time. Thanks.

Issue #3187, opened 2019-07-29
elk7.2  sky start error


qianglu1989:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ x] Question or discussion

___
### Question
 {"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"Root mapping definition has unsupported parameters:  [type : {type={properties={sequence={type=integer}, heartbeat_time={type=long}, service_id={type=integer}, name={copy_to=name_match, type=keyword}, detect_point={type=integer}, name_match={analyzer=oap_analyzer, type=text}, register_time={type=long}}}}]"}],"type":"mapper_parsing_exception","reason":"Failed to parse mapping [_doc]: Root mapping definition has unsupported parameters:  [type : {type={properties={sequence={type=integer}, heartbeat_time={type=long}, service_id={type=integer}, name={copy_to=name_match, type=keyword}, detect_point={type=integer}, name_match={analyzer=oap_analyzer, type=text}, register_time={type=long}}}}]","caused_by":{"type":"mapper_parsing_exception","reason":"Root mapping definition has unsupported parameters:  [type : {type={properties={sequence={type=integer}, heartbeat_time={type=long}, service_id={type=integer}, name={copy_to=name_match, type=keyword}, detect_point={type=integer}, name_match={analyzer=oap_analyzer, type=text}, register_time={type=long}}}}]"}},"status":400}
		at org.elasticsearch.client.RestClient$SyncResponseListener.get(RestClient.java:705) ~[elasticsearch-rest-client-6.3.2.jar:6.3.2]
		at org.elasticsearch.client.RestClient.performRequest(RestClient.java:235) ~[elasticsearch-rest-client-6.3.2.jar:6.3.2]
		at org.elasticsearch.client.RestClient.performRequest(RestClient.java:198) ~[elasticsearch-rest-client-6.3.2.jar:6.3.2]
		at org.elasticsearch.client.RestHighLevelClient.performRequest(RestHighLevelClient.java:522) ~[elasticsearch-rest-high-level-client-6.3.2.jar:6.3.2]
		at org.elasticsearch.client.RestHighLevelClient.performRequestAndParseEntity(RestHighLevelClient.java:508) ~[elasticsearch-rest-high-level-client-6.3.2.jar:6.3.2]
		at org.elasticsearch.client.IndicesClient.create(IndicesClient.java:99) ~[elasticsearch-rest-high-level-client-6.3.2.jar:6.3.2]
[...]

Issue #3047, opened 2019-07-11
facing issue while connecting collector to elastic 7.1


lmzhizhe:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [Question or discussion ] Question or discussion
- [ Bug] Bug

___
### Question
- What do you want to know?
start error with es7.1,how to resolved it?
___
### Bug
- Which version of SkyWalking, OS and JRE?
6.2 linux centos  jdk 1.8
- Which company or project?
vivo
- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.

````
route allocated: 1 of 10; total allocated: 3 of 30]
2019-07-11 16:18:48,357 - org.apache.http.impl.nio.conn.PoolingNHttpClientConnectionManager - 299 [I/O dispatcher 1] DEBUG [] - Connection [id: http-outgoing-0][route: {}->http://172.20.226.147:9200] can be kept alive indefinitely
2019-07-11 16:18:48,357 - org.apache.http.impl.nio.conn.ManagedNHttpClientConnectionImpl - 154 [I/O dispatcher 1] DEBUG [] - http-outgoing-0 172.20.226.153:12847<->172.20.226.147:9200[ACTIVE][r:r]: Set timeout 0
2019-07-11 16:18:48,357 - org.apache.http.impl.nio.conn.PoolingNHttpClientConnectionManager - 305 [I/O dispatcher 1] DEBUG [] - Connection released: [id: http-outgoing-0][route: {}->http://172.20.226.147:9200][total kept alive: 3; route allocated: 1 of 10; total allocated: 3 of 30]
2019-07-11 16:18:48,357 - org.elasticsearch.client.RestClient - 59 [I/O dispatcher 1] DEBUG [] - request [PUT http://172.20.226.147:9200/_template/alarm_record] returned [HTTP/1.1 400 Bad Request]
2019-07-11 16:18:48,361 - org.apache.http.impl.nio.client.InternalIODispatch - 85 [I/O dispatcher 1] DEBUG [] - http-outgoing-0 [ACTIVE] [content length: 1191; pos: 1191; completed: true]
2019-07-11 16:18:48,361 - org.apache.skywalking.oap.server.starter.OAPServerStartUp - 55 [main] ERROR [] - method [PUT], host [http://172.20.226.147:9200], URI [/_template/alarm_record], status line [HTTP/1.1 400 Bad Request]
[...]

Issue #3000, opened 2019-07-04
Is there a plan to support Elasticsearch  7.X?


guanfengchen:
### Question
- Is there a plan to support Elasticsearch  7.X?

wu-sheng:
Duplicate as before. ElasticSearch 6 is new already. We haven't plan to upgrade in short term

guanfengchen:
你好,
我在github上下了源代码在学习，现在有个问题，不知道你有没有时间可以帮忙一下。
在agent这边，我看到有很多的plugin
这些plugin是怎样加载到代码中，别切在代码上面实现拦截的呢？

发送自 Windows 10 版邮件<https://go.microsoft.com/fwlink/?LinkId=550986>应用

________________________________
发件人: 吴晟 Wu Sheng <notifications@github.com>
发送时间: Thursday, July 4, 2019 1:51:22 PM
收件人: apache/skywalking
抄送: guanfengchen; Author
主题: Re: [apache/skywalking] Is there a plan to support Elasticsearch 7.X? (#3000)


Duplicate as before. ElasticSearch 6 is new already. We haven't plan to upgrade in short term

―

wu-sheng:
All issues in English. And if you are learning open source project, you should read source codes. All features are there

wu-sheng:
We can't support to teach everyone step by step. That is unrealistic.

Issue #2529, opened 2019-04-26
 does skywalking support elasticsearch 7?


imkiven:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [√] Question or discussion

___
### Question
- What do you want to know?
 does skywalking support elasticsearch 7?

wu-sheng:
Plugin or Storage?

imkiven:
> Plugin or Storage?

both or either or neither?

wu-sheng:
We haven't any of them. But
- For the plugin, we welcome the contribution.
- For storage, we can't accept, because 6.3+ is already very new to the community.

hellxz:
Does Skywalking 6.2 supported Elasticsearch 7.x  now? @wu-sheng

wu-sheng:
We have clear documents about version supported

PR #3613, opened 2019-10-13
[Feature] Support ElasticSearch 7 as storage

kezhenxu94:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

There're many users asking for support to ES7, #3047 #3187 #3000 #2529 , this PR tries to make the OAP backend work with ES7, without breaking the capabilities of ES6.

There're some breaking changes in ES7, since we are (and will) still using ES6-java-client, all the efforts are made by lower level REST client of ES.

In this PR, we don't use high level REST client to ES7, which is more convenient and is used in ES6 storage plugin, by using low level client, we can pass some parameters that ES6 doesn't support, such as `if_seq_no`, etc.

Some of the parameters are deprecated in ES7, but according to the upgrade strategy of ES, the deprecated parameters should work at least in all the minor versions under a major version, so there may be warnings in the log, but it does no harm, that's as was expected.

@wu-sheng Although the E2E tests passed (locally, and should be the same in Infra Jenkins), I'll propose to mark this as **Experimental** in the documentation for now, and invite some users to try and if the feedback is positive, we can make it **Stable** in the future version, documentations should be added later after fully discussed.

Anyone who wants to review this PR, please pay special attention to the ES6 storage plugin, this PR should not break the ES6 plugin in any aspect, thanks.

### DIFF
--- a/Jenkinsfile-E2E
+++ b/Jenkinsfile-E2E
@@ -69,9 +69,15 @@ pipeline {
                             }
                         }
 
-                        stage('Agent Reboot Tests') {
+                        stage('TTL ES6 Tests') {
                             steps {
-                                sh 'bash -x test/e2e/run.sh e2e-agent-reboot'
+                                sh 'bash -x test/e2e/run.sh e2e-ttl/e2e-ttl-es'
+                            }
+                        }
+
+                        stage('Cluster Tests (ES7/ZK)') {
+                            steps {
+                                sh 'bash -x test/e2e/run.sh e2e-cluster/test-runner-es7'
                             }
                         }
                     }
@@ -85,9 +91,15 @@ pipeline {
                             }
                         }
 
-                        stage('TTL ES Tests') {
+                        stage('Agent Reboot Tests') {
                             steps {
-                                sh 'bash -x test/e2e/run.sh e2e-ttl/e2e-ttl-es'
+                                sh 'bash -x test/e2e/run.sh e2e-agent-reboot'
+                            }
+                        }
+
+                        stage('TTL ES7 Tests') {
+                            steps {
+                                sh 'bash -x test/e2e/run.sh e2e-ttl/e2e-ttl-es7'
                             }
                         }
                     }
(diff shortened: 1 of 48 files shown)


kezhenxu94:
The e2e codes are almost the same and can be refactor to reuse the ES6 tests, which will be refactored later in another PR, if I can got some time :)

kezhenxu94:
Apply TBD until it's fully discussed and documentations are supplemented

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
```diff
@@ -81,7 +81,7 @@ public MetadataQueryEsDAO(ElasticSearchClient client, int queryMaxSize) {
         sourceBuilder.size(0);
 
         SearchResponse response = getClient().search(ServiceInventory.INDEX_NAME, sourceBuilder);
-        return (int)response.getHits().getTotalHits();
+        return response.getHits().getHits() == null ? 0 : response.getHits().getHits().length;
```
wu-sheng: This could be NPE?

Jenkinsfile-E2E
```diff
@@ -69,9 +69,15 @@ pipeline {
                             }
                         }
 
-                        stage('Agent Reboot Tests') {
+                        stage('TTL ES6 Tests') {
```
wu-sheng: I assumr this means es7 test

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
```diff
@@ -95,7 +95,7 @@ public MetadataQueryEsDAO(ElasticSearchClient client, int queryMaxSize) {
         sourceBuilder.size(0);
 
         SearchResponse response = getClient().search(EndpointInventory.INDEX_NAME, sourceBuilder);
-        return (int)response.getHits().getTotalHits();
+        return response.getHits().getHits() == null ? 0 : response.getHits().getHits().length;
```
dmsolr: Note that `getTotalHits()` and `getHits().length` are not equivalent.

kezhenxu94:
> I want to know whether anyone from community will try this locally and in their test env. This looks very challenge for further maintain works, rather than this tests.
> 
> This must be based on near 100% test coverage(logically) supports, including e2e, IT tests, unit tests. We don't know when it breaks. And very hard to review.

Seems it's unwise to do ES7 adaptation along with ES6 together, closing for now

wu-sheng:
> Seems it's unwise to do ES7 adaptation along with ES6 together, closing for now

If you want this, in my old-time, I am playing class loader game, which is not easy :) or redistributes a new release should be better.

PR #3617, opened 2019-10-14
[test/plugin] support plugin testcases run with optional/bootstrap plugins

dmsolr:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
This is PR for supporting plugin testcases run with optional/bootstrap plugins through configuration.yml.
The optional name of configuration is `RunningMode` which has 3 values, `default`,`with_optional`,`with_bootstap`. The default is `default` when the configuration omitted.

### DIFF
--- a/test/plugin/run.sh
+++ b/test/plugin/run.sh
@@ -146,6 +146,26 @@ if [[ ! -f $supported_version_file ]]; then
     exitWithMessage "cannot found 'support-version.list' in directory ${scenario_name}"
 fi
 
+_agent_home=${agent_home}
+mode=`grep "runningMode" ${scenario_home}/configuration.yml |sed -e "s/\s//g" |awk -F: '{print $2}'`
+if [[ "$mode" == "with_optional" ]]; then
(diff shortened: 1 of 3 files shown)


dmsolr:
FYI, @zhaoyuguang

dmsolr:
> LGTM. How to indicate to add only one/serveral specific optional or bootstrap plugins?

Not support yet. I wonder if this is necessary.

wu-sheng:
Actually, only that make sense. Even more than this one. Think in this way, more and more optional plugins added, one day a new optional contributed, all codes are good. But will add one span to many other cases. Then what do you expect?

dmsolr:
> Actually, only that make sense. Even more than this one. Think in this way, more and more optional plugins added, one day a new optional contributed, all codes are good. But will add one span to many other cases. Then what do you expect?

I got it. I will do this later.

wu-sheng:
> I got it. I will do this later.

Once you have that, I recommend you revert these two options, by providing specific plugin selector options, if you feel making sense too.

Issue #3402, opened 2019-09-03
More reasonable error output


hanahmily:
Hi community
I knew we are familiar with the following errors. But they bother us and our users to find what's going on in oap internal.

Can we have a more proper method to improve it? Want to hear your voice, thanks.

```
ap-deployment-dcd587fb8-bqrlp oap java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
oap-deployment-dcd587fb8-bqrlp oap      at java.util.LinkedList.checkElementIndex(LinkedList.java:555) ~[?:1.8.0_181]
oap-deployment-dcd587fb8-bqrlp oap      at java.util.LinkedList.get(LinkedList.java:476) ~[?:1.8.0_181]
oap-deployment-dcd587fb8-bqrlp oap      at org.apache.skywalking.oap.server.core.remote.selector.ForeverFirstSelector.select(ForeverFirstSelector.java:37) ~[server-core-6.1.0.7.jar:6.1.0.7]
oap-deployment-dcd587fb8-bqrlp oap      at org.apache.skywalking.oap.server.core.remote.RemoteSenderService.send(RemoteSenderService.java:58) ~[server-core-6.1.0.7.jar:6.1.0.7]
oap-deployment-dcd587fb8-bqrlp oap      at org.apache.skywalking.oap.server.core.register.worker.RegisterRemoteWorker.in(RegisterRemoteWorker.java:47) ~[server-core-6.1.0.7.jar:6.1.0.7]
oap-deployment-dcd587fb8-bqrlp oap      at org.apache.skywalking.oap.server.core.register.worker.RegisterRemoteWorker.in(RegisterRemoteWorker.java:32) ~[server-core-6.1.0.7.jar:6.1.0.7]
oap-deployment-dcd587fb8-bqrlp oap      at java.util.HashMap$Values.forEach(HashMap.java:981) [?:1.8.0_181]
oap-deployment-dcd587fb8-bqrlp oap      at org.apache.skywalking.oap.server.core.register.worker.RegisterDistinctWorker.onWork(RegisterDistinctWorker.java:77) [server-core-6.1.0.7.jar:6.1.0.7]
oap-deployment-dcd587fb8-bqrlp oap      at org.apache.skywalking.oap.server.core.register.worker.RegisterDistinctWorker.access$100(RegisterDistinctWorker.java:34) [server-core-6.1.0.7.jar:6.1.0.7]
[...]

PR #3619, opened 2019-10-15
Fix more resonable error

yanickxia:
- [ √ ] Bug fix

- Related issues
#3402 
___
### Bug fix
- Bug description.
return a ref list, maybe clear in other thread
- How to fix?
return a Immutable copy list

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/remote/client/RemoteClientManager.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/remote/client/RemoteClientManager.java
@@ -20,6 +20,17 @@ package org.apache.skywalking.oap.server.core.remote.client;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Sets;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.Setter;
@@ -34,17 +45,6 @@ import org.apache.skywalking.oap.server.telemetry.api.MetricsCreator;
 import org.apache.skywalking.oap.server.telemetry.api.MetricsTag;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
 
 /**
  * This class manages the connections between OAP servers. There is a task schedule that will automatically query a
@@ -64,8 +64,9 @@ public class RemoteClientManager implements Service {
 
     /**
      * Initial the manager for all remote communication clients.
+     *
      * @param moduleDefineHolder for looking up other modules
-     * @param remoteTimeout for cluster internal communication, in second unit.
+     * @param remoteTimeout      for cluster internal communication, in second unit.
      */
     public RemoteClientManager(ModuleDefineHolder moduleDefineHolder, int remoteTimeout) {
         this.moduleDefineHolder = moduleDefineHolder;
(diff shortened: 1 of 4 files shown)


wu-sheng:
@yanickxia I am also new to this issue, so if you find anything or I may say something wrong, please let me know.

wu-sheng:
FYI @kezhenxu94 @dmsolr @zhaoyuguang @IanCao We need more eyes on this, please review(not just PR, the original codes)

yanickxia:
@wu-sheng  if `getRemoteClient` before `getFreeClients().clear();` then 
```java
       List<RemoteClient> clientList = clientManager.getRemoteClient();
        if (clientList.size() == 0) {  
            logger.warn("There is no available remote server for now, ignore the streaming data until the cluster metadata initialized.");
            return;
        }
```
math the if condition, without any other code execute under it. 

i think just guarantee the clientList be not modified after got it.

wu-sheng:
It wouldn't NPE, yes. But still it is not wrong. Because client list is there logically, and cause warning by concurrency access. So it is a code bug.

yanickxia:
> It wouldn't NPE, yes. But still it is not wrong. Because client list is there logically, and cause warning by concurrency access. So it is a code bug.

you are right. maybe never return RemoteClient List during reBuildRemoteClients.

wu-sheng:
Never return is a dangerous, will cause performance unstable. Please consider to always maintain immutable list to outside use, and don't clear it.

yanickxia:
amaze

> Never return is a dangerous, will cause performance unstable. Please consider to always maintain immutable list to outside use, and don't clear it.

i refactor the rebuild logic, but have something question.
1. clientA & clientB is too hard understand, i guess that keep stable memory usage..
2. usingClients just a shadow ref to clientA or ClientB, i think without concurrency race.
3. reBuildRemoteClients not ness synchronized protect it, because i invoke by refresh(), use scheduleWithFixedDelay more easy to understand

wu-sheng:
Could you add a test case for this?

yanickxia:
> Could you add a test case for this?

Add a concurrency test for getRemoteClient while refresh. take a look

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/remote/client/RemoteClientManager.java
```diff
@@ -154,7 +159,7 @@ private void printRemoteClientList() {
     }
 
     public List<RemoteClient> getRemoteClient() {
-        return usingClients;
+        return ImmutableList.copyOf(usingClients);
```
wu-sheng: This is heavy op. UsingClient should be immutable list itself.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/remote/client/RemoteClientManager.java
```diff
@@ -181,29 +186,27 @@ private void switchCurrentClients() {
...
-        tempRemoteClients.forEach((address, action) -> {
-            switch (action) {
-                case Leave:
-                    if (remoteClients.containsKey(address)) {
-                        getFreeClients().add(remoteClients.get(address));
-                    }
+        unChangeAddresses.forEach(createRemoteClient::remove);
+        closeRemoteClient.putAll(createRemoteClient);
+
+        getFreeClients().clear(); //clean free client list, for build a new list
```
wu-sheng: If you still use clear, it is not safe. Still could get a chance to have list ref, than cleared by here. Right?

wu-sheng: I hope we don't use switch OP, because seems unnecessary, as here, no one really close connection. So, we just need a local varible to prepare the list, make it sorted and immutable, than using as usingClient. 

Make sense? Do I miss anything?

dmsolr: The key point is, clear and copyOf are not atomic operations. You need to understand that.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/remote/client/RemoteClientManager.java
yanickxia: i got it, will change it

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/remote/client/RemoteClientManager.java
yanickxia: @dmsolr @wu-sheng `clear` or `copyOf` unnecessary atomic operation, in fact, the only refresh thread invoke this function, the raw code has a smart tips, has clientA & ClientB two linkedList, `getFreeClient()` will got a not use linked list, so you can do everything on this `free  linked list`

wu-sheng:
Only one confusing thing, why coverage rate drops as you have added the test case. @JaredTan95 could you help on recheck?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/remote/client/RemoteClientManager.java
```diff
@@ -181,53 +166,49 @@ private void switchCurrentClients() {
...
+                .filter(closeRemoteClient::containsKey)
+                .forEach(unChangeAddress -> closeRemoteClient.get(unChangeAddress).setAction(Action.Unchanged));
 
-        tempRemoteClients.forEach((address, action) -> {
-            switch (action) {
-                case Leave:
-                    if (remoteClients.containsKey(address)) {
-                        getFreeClients().add(remoteClients.get(address));
-                    }
+        unChangeAddresses.forEach(createRemoteClient::remove);
```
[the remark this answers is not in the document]
yanickxia: done

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/remote/client/RemoteClientManager.java
```diff
@@ -181,53 +166,49 @@ private void switchCurrentClients() {
...
      */
-    private synchronized void reBuildRemoteClients(List<RemoteInstance> remoteInstances) {
-        getFreeClients().clear();
+    private void reBuildRemoteClients(List<RemoteInstance> remoteInstances) {
+        final Map<Address, RemoteClientAction> closeRemoteClient = this.usingClients.stream()
+                .collect(Collectors.toMap(RemoteClient::getAddress, client -> new RemoteClientAction(client, Action.Close)));
 
-        Map<Address, RemoteClient> remoteClients = new HashMap<>();
-        getRemoteClient().forEach(client -> remoteClients.put(client.getAddress(), client));
+        final Map<Address, RemoteClientAction> createRemoteClient = remoteInstances.stream()
```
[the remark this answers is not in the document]
yanickxia: done

PR #3624, opened 2019-10-15
Remove duplicated doc

wu-sheng:
Remove duplicated doc, we have a more clear one now.


wu-sheng:
@dmsolr Solr 7.x seems unstable today. Please take a look.

PR #3653, opened 2019-10-18
[test/plugin]immigrate kafka scenario

arugal:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#3583
___
1. For```nameStartsWith("org.apache.skywalking.").and(not(nameStartsWith("org.apache.skywalking.apm.toolkit.")));``` reasons, I use ```test.org.apache.skywalking``` as the package name.


wu-sheng:
FYI @dmsolr

test/plugin/runner-helper/src/main/resources/docker-compose.template
```diff
@@ -57,9 +57,9 @@ services:
             - ${volume}
         </#list>
         </#if>
-        <#if service.environments??>
+        <#if service.environment??>
         environment:
-        <#list service.environments as environment>
+        <#list service.environment as environment>
```
wu-sheng: What is this change for?

PR #3657, opened 2019-10-19
[CI] Make E2E tests run in multiple JDK versions

### LINKED ISSUE #3608: e2e for JDK 9/11/12
@kezhenxu94 I think you said, JDK 9+(included) should be OK for OAP backend and UI. Let's add them into e2e and make it always available when you upgrade some libs.

kezhenxu94:
closes #3608 

should fail until e2e container PR is merged


kezhenxu94:
> As [SkyAPMTest/e2e-container#5](https://github.com/SkyAPMTest/e2e-container/pull/5) merged, let's add JDK 11 and 12 to the test. Is the agent application running in these JDKs too?

Sounds good to me, which group in Jenkins job to add?

wu-sheng:
How about we keep all JDK related and reboot in group 1, and others are in group 2?

wu-sheng:
Two questions
1. Is the agent application running in these JDKs too?
1. I thought UI is not running? So don't know whether is it working? Am I right?

kezhenxu94:
> Two questions
> 
> 1. Is the agent application running in these JDKs too?
If I remember correctly, at the first step of E2E, we designed to run all the components(OAP, agent, web app) in a single container, so they share the same JDK environment, so **YES**

> 2. I thought UI is not running? So don't know whether is it working? Am I right?
Do you mean web app? UI didn't depends on JDK I think

wu-sheng:
UI didn't depend, I mean webapp. Is it running in the tests? If yes, I think we have tested it. I remembered you said, it is not working back to several days ago.

apm-webapp/pom.xml
```diff
@@ -70,11 +70,6 @@
             <artifactId>spring-boot-configuration-processor</artifactId>
             <version>${spring.boot.version}</version>
         </dependency>
-        <dependency>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-devtools</artifactId>
-            <version>${spring.boot.version}</version>
-        </dependency>
```
kezhenxu94: @wu-sheng  It turns out that this is the reason why I said `webapp` doesn't support JDK 9, it's this module (`devtool`) that doesn't support JDK9, and it's just for development, so removing it does no harm

kezhenxu94:
> UI didn't depend, I mean webapp. Is it running in the tests? If yes, I think we have tested it. I remembered you said, it is not working back to several days ago.

Confirmed, the web app is running in the tests

apm-webapp/pom.xml
wu-sheng: Yes, this should be safe to be removed. This is byte code manipulated tech :) So it needs to do many things to be compatible with various JDKs :P

Issue #3671, opened 2019-10-21
Add   register worker metrics


dominicqi:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [+] Question or discussion
- [ ] Bug
- [ ] Requirement
- [+] Feature or performance improvement

___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS and JRE?

- Which company or project?

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.

___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.
 Could i add a register worker latency metrics ?
  Here is the screenshot in Grafana.
![image](https://user-images.githubusercontent.com/12235378/67191670-98bc7b80-f424-11e9-962c-e45a19fa40c2.png)

wu-sheng:
Make sense to me, if you want, please submit the pull request.

JaredTan95:
added.

PR #3677, opened 2019-10-22
Add register persistent worker latency metrics  #3671

dominicqi:

- [ ] Bug fix
- [+] New feature provided
- [ ] Improve performance

- Related issues
#3671 
___
### Bug fix
- Bug description.

- How to fix?

___
Add metrics for register persistent worker  ,and Grafana dashboard.


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/register/worker/RegisterPersistentWorker.java
```diff
@@ -80,39 +91,43 @@ private void onWork(RegisterSource registerSource) {
...
-                } catch (Throwable t) {
-                    logger.error(t.getMessage(), t);
-                }
-            });
-            sources.clear();
+                });
+                sources.clear();
+            }
+        } catch (IOException e) {
+            logger.error("not except exception happen", e);
```
wu-sheng: Why need this log and catch? And clearly, the sentence is not right.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/register/worker/RegisterPersistentWorker.java
```diff
@@ -80,39 +91,43 @@ private void onWork(RegisterSource registerSource) {
...
-                        }
-                    } else {
-                        int sequence;
-                        if ((sequence = registerLockDAO.getId(scopeId, source)) != Const.NONE) {
-                            try {
-                                dbSource = registerDAO.get(modelName, source.id());
-                                if (Objects.nonNull(dbSource)) {
-                                    if (dbSource.combine(source)) {
-                                        registerDAO.forceUpdate(modelName, dbSource);
+        try (HistogramMetrics.Timer ignored = histogram.createTimer()) {
```
wu-sheng: Why timer called ignored?

docs/en/setup/backend/telemetry/trace-mode-grafana.json
```diff
@@ -1158,14 +1158,112 @@
         "align": false,
         "alignLevel": null
       }
+    },
```
wu-sheng: You just change the trace-mode json, miss the mesh-mode one.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/register/worker/RegisterPersistentWorker.java
dominicqi: i want use the try-with-resource sugar . The Timer#close() throws IOException that  never throw. I will remove the IOException from the close method.

wu-sheng:
@JaredTan95 Is this good?

wu-sheng:
@kezhenxu94 This works as expected. 3 plugin test workload finished in 2m, only source download and compare. I just noticed one place needs to be improved, we can't see the cooperation log.

wu-sheng:
@kezhenxu94 This seems to be the whole log I got, https://builds.apache.org/blue/rest/organizations/jenkins/pipelines/skywalking-agent-test/runs/779/nodes/27/log/?start=0

Do I miss anything?

JaredTan95:
image

image

Looks Good to me.

PR #3679, opened 2019-10-22
[test/plugin] docker prune at cleanup

dmsolr:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

Docker prune at cleanup step.


wu-sheng:
CI is even slower than agent plugin test. I think we could add more cases to current workloads. Agree?

PR #3683, opened 2019-10-22
ServiceInstanceRelation set error id

coki230:
the SourceServiceId set error id which is destServiceId.

-  Bug fix

___
### Bug fix
- Bug description.
ServiceInstanceRelation set error id. the setSourceServiceId set the destServiceId
- How to fix?
set the right id.
___


coki230:
> LGTM. Good catch. This bug didn't show up because we haven't used this yet.

we ues the relationship between the instance, so i discover the bug, BTW the function runs well

wu-sheng:
You are showing metrics between instance? Is the paylaod acceptable?

wu-sheng:
@x22x22 This seems interesting. If we could enhance the topology, I think we should do this in 6.6 too.

coki230:
> You are showing metrics between instance? Is the paylaod acceptable?

yes, may be we don't have many pods, so it's working well so far.

x22x22:
> @x22x22 This seems interesting. If we could enhance the topology, I think we should do this in 6.6 too.

ok

PR #3695, opened 2019-10-23
Add postgresql agent sql query param show 

aderm:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___


apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/define/StatementEnhanceInfos.java
```diff
@@ -55,29 +54,14 @@ public String getStatementName() {
...
-        if (index >= length) {
-            int newSize = Math.max(index + 1, length * 2);
-            Object[] newParameters = new Object[newSize];
-            System.arraycopy(parameters, 0, newParameters, 0, length);
-            Arrays.fill(newParameters, length, newSize, null);
-            parameters = newParameters;
+        // start from 1
+        index--;
+        if (index >= 0) {
+            parameters.set(index, parameter);
```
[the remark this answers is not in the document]
kezhenxu94: and refer to https://github.com/apache/skywalking/pull/2846#discussion_r293768204 for context and reasons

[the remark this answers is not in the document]
aderm: it must be add methed，local test will do tomorrow for pg etc. env at company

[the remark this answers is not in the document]
aderm: @kezhenxu94 thanks for you suggest, i will keep it up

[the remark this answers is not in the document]
kezhenxu94: > it must be add methed，local test will do tomorrow for pg etc. env at company

I think "add" is still unsafe if you track back to what I mention above, users don't necessarily set the parameters sequentially, they may set the 4th parameter and then the 1st,  and 3rd and 2nd, I thought TreeMap is the best when I implement the MySQL version, (see my link above), so please take a look there to see why I finally give up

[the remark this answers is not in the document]
wu-sheng: Then why tests pass? I am assuming the test is not covering accurately.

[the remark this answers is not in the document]
kezhenxu94: > Then why tests pass? I am assuming the test is not covering accurately.

Because the config is disabled by default?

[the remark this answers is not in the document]
wu-sheng: Could be, @aderm as you are adding a new feature, please make it tested.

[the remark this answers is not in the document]
aderm: @wu-sheng ok,  I having change the code , runing the test at the motent.

kezhenxu94:
@aderm ping me when you think it's ready

aderm:
@kezhenxu94 add sql param is done. second i want to move some pg common in sub module postgresql-commons, what do you think?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -273,6 +273,27 @@
             public static int SQL_PARAMETERS_MAX_LENGTH = 512;
         }
 
+        public static class POSTGRESQL {
+            /**
+             * If set to true, the parameters of the sql (typically {@link java.sql.PreparedStatement}) would be
+             * collected.
+             */
+            public static boolean TRACE_SQL_PARAMETERS = true;
```
kezhenxu94: Default to `false`

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -273,6 +273,27 @@
...
+             */
+            public static int SQL_PARAMETERS_MAX_COUNT = 20;
+
+            /**
+             * For the sake of performance, SkyWalking won't save the entire parameters string into the tag, but only
+             * the first {@code SQL_PARAMETERS_MAX_LENGTH} characters.
+             *
+             * Set a negative number to save the complete parameter string to the tag.
+             */
+            public static int SQL_PARAMETERS_MAX_LENGTH = 512;
```
kezhenxu94: Add these configs to documentations, otherwise, nobody can find this feature until they read the source codes

apm-sniffer/apm-sdk-plugin/postgresql-8.x-plugin/pom.xml
```diff
@@ -48,6 +49,11 @@
             <version>${project.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>${guava.version}</version>
+        </dependency>
     </dependencies>
```
kezhenxu94: You **MUST NOT** use third-party dependencies in agent plugin

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
kezhenxu94: And use environment variable to override the config in the test

apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/define/StatementEnhanceInfos.java
```diff
@@ -80,4 +87,19 @@ public void setParameter(int index, final Object parameter) {
...
+    public List<Object> getSortParameters() {
+        List<Object> parameters = new ArrayList<Object>();
+        if (parameterMap.size() > 0) {
+            Iterator<Entry<Integer,Object>> iterator = parameterMap.entrySet().iterator();
+            while (iterator.hasNext()) {
+                parameters.add(iterator.next().getValue());
+            }
+        }
+        return parameters;
+    }
```
kezhenxu94: There is still performance impact when using `TreeMap`, see [comment here](https://github.com/apache/skywalking/pull/2846#pullrequestreview-250196401), the performance is a big concern in agent side, please make a tradeoff

apm-sniffer/apm-sdk-plugin/postgresql-8.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdbc/postgresql/util/ParameterUtil.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public static String format(List<Object> params, int limitCnt, int limitLength) {
+        boolean overLoad = false;
+        int actualCnt = params.size();
+        if (params.size() > limitCnt) {
+            overLoad = true;
+            actualCnt = limitCnt;
+        }
+
+        String paramStr = Joiner.on(",").join(params.subList(0, actualCnt));
```
kezhenxu94: Introduce a third-party dependency (guava) just for one method is not worthy, and again, we **MUST NOT** use third-party dependency unless we shad them

apm-sniffer/apm-sdk-plugin/postgresql-8.x-plugin/pom.xml
wu-sheng: Why need this lib? BTW. You are just logging parameter into the span, seems not a complex work. And clearly, the MySQL plugin has given a demo about how to do it.

apm-sniffer/apm-sdk-plugin/postgresql-8.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdbc/postgresql/util/ParameterUtil.java
aderm: ok

test/plugin/scenarios/postgresql-above9.4.1207-scenario/config/expectedData.yaml
```diff
@@ -53,6 +53,7 @@ segmentItems:
         - {key: "db.type", value: "sql"}
         - {key: "db.instance", value: "postgres"}
         - {key: "db.statement", value: "INSERT INTO test_007(id, value) VALUES(?,?)"}
+        - {key: "db.bind_vars", value: "[1,1]"}
```
kezhenxu94: should be `db.sql.parameters`?

kezhenxu94: And after setting the default value of `TRACE_SQL_PARAMETERS` to `false`, we should pass an environment variable to override it to `true`, otherwise, there is no key `db.sql.parameters`

aderm: how to set?

aderm: in docker setting?

kezhenxu94: seems to be here

https://github.com/apache/skywalking/blob/2e2616d0efc34c29a3d3388e100d948303ad2832/test/plugin/scenarios/postgresql-scenario/bin/startup.sh#L20

you can use something like 

```
java -jar ${agent_opts} -Dskywalking.plugin.postgresql.trace_sql_parameters=true ${home}/../libs/posgresql.......
```

it should work, please try

aderm: ok

aderm:
@kezhenxu94 done

PR #3714, opened 2019-10-25
[CI] Run build jobs on different nodes parallelly

kezhenxu94:
In this patch, I've made it to run multiple jobs parallelly on different Jenkins slave nodes, thus, our build jobs can be distributed more evenly to our VM nodes, and the VM with lower resources can also be scheduled fairly.

In this PR, I also rewrote the Jenkisfile with scripted Pipeline syntax, our previous syntax is DSL, it's much more human readable, but lack of flexibility in our complex scenarios, now we're using scripted syntax, based on Groovy, it provides a programmable ability, and flexibility, to simplify the Jenkinsfile, (from ~200 lines to ~100 lines)

Build time: ~18.5 min down to ~15.5 min, not a big deal, but it takes effects

### DIFF
--- a/Jenkinsfile-E2E
+++ b/Jenkinsfile-E2E
@@ -16,134 +16,105 @@
  *
  */
 
-pipeline {
-    agent {
-        label 'skywalking'
+def version = '1.3'
+
+// Test cases for different JDK versions
+def jdk_cases = ['jdk8', 'jdk9', 'jdk11', 'jdk12'].collect {
+    [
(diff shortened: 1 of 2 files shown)


kezhenxu94:
We don't need to group the jobs anymore, because the jobs are scheduled on different nodes in a smaller granularity (test case), meaning that every test case will take an executor, if there is no more available executor, the test case should be in queue automatically

wu-sheng:
Please move this as a record in WIKI. We may need this someday.

PR #3698, opened 2019-10-24
[test/plugin]  immigrate  undertow scenario

arugal:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#3583
___
### Bug fix
- Bug description.

- How to fix?

___
### New feature or improvement
- Describe the details and related test reports.

PR #3716, opened 2019-10-25
[test/plugin] to check agent register status

dmsolr:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#3698 

1. curl time-out, default 2s
2. visit scenario's entry after agent register success.

### DIFF
--- a/test/plugin/mock-collector/src/main/java/org/apache/skywalking/plugin/test/mockcollector/Main.java
+++ b/test/plugin/mock-collector/src/main/java/org/apache/skywalking/plugin/test/mockcollector/Main.java
@@ -19,11 +19,19 @@ package org.apache.skywalking.plugin.test.mockcollector;
 import io.grpc.netty.NettyServerBuilder;
 import io.netty.channel.local.LocalAddress;
 
+import java.io.IOException;
 import java.net.InetSocketAddress;
 
+import org.apache.skywalking.plugin.test.mockcollector.entity.ValidateData;
 import org.apache.skywalking.plugin.test.mockcollector.service.*;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.servlet.ServletContextHandler;
+import org.eclipse.jetty.servlet.ServletHolder;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 
 public class Main {
     public static void main(String[] args) throws Exception {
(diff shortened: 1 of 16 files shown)


test/plugin/mock-collector/src/main/java/org/apache/skywalking/plugin/test/mockcollector/Main.java
```diff
@@ -47,6 +55,18 @@ public static void main(String[] args) throws Exception {
         String contextPath = "/";
         ServletContextHandler servletContextHandler = new ServletContextHandler(ServletContextHandler.NO_SESSIONS);
         servletContextHandler.setContextPath(contextPath);
+        servletContextHandler.addServlet(new ServletHolder(new HttpServlet() {
+            @Override
+            protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
+                if (ValidateData.INSTANCE.getRegistryItem().getApplications().isEmpty()) {
+                    resp.setStatus(500);
```
arugal: Is there a case where there are multiple ```agent``` in one test case.

test/plugin/scenarios/ehcache-2.x-scenario/config/expectedData.yaml
```diff
@@ -23,26 +23,8 @@ registryItems:
   heartbeat: []
 segmentItems:
   - applicationCode: ehcache-2.x-scenario
-    segmentSize: 2
+    segmentSize: ge 2
```
arugal: It is recommended to use ```gt 1```

dmsolr:
> Can you explain why change `<n>` to `ge <n>` in `expectedData.yml`, does it have side effects?

This is why we modify the `health-check` mechanism. The `healthcheck` API will be called many times when health-check failed.

And then I set `curl` timeout which intends to present `the failure` more easy.

dmsolr:
In fact, I think the validate-tool skips the segment of `health-check`.  HDYT?

kezhenxu94:
> In fact, I think the validate-tool skips the segment of `health-check`. HDYT?

Ignoring the `healthCheck` span sounds good to me, health check is not related to the plugin itself, in fact, E2E ignore the health check too IIRC

test/plugin/mock-collector/src/main/java/org/apache/skywalking/plugin/test/mockcollector/Main.java
wu-sheng: Mutiple agent is possible, but let's avoid it if possible.

wu-sheng:
Please remove the socket.io test for now.

PR #3717, opened 2019-10-25
fix deficient code, "/socket.io/" http segment declare

mrproliu:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- missing test case declare segment from plugin netty-socketio

- How to fix?
- add "/socket.io/" HTTP segment declare
___


mrproliu:
@dmsolr , you can have a look

dmsolr:
> @dmsolr , you can have a look

Hi @mrproliu 
I am sorry to let you know that PR passed because the  Agent-Plugin-Test framework has a bug. Which will be fixed by #3716 .  BTW, you losing some segments in the `expectData.yml` yet.

wu-sheng:
@mrproliu I have asked @dmsolr to remove socket.io test in jenkins file, which will give you 1-2 days to fix and update this PR to add test back

dmsolr:
@wu-sheng @mrproliu , I have remove it. Please add it back after fixing it.

wu-sheng:
Data sync complete, please add the test back and fix the test bug.

mrproliu:
I have added the test case into Jenkinsfile-Agent-Test file, why is it not running on Agent Plugin Tests Workload 1?

wu-sheng:
It is running. The graph only updates after it runs once.

mrproliu:
> It is running. The graph only updates after it runs once.

good, have resolved test case bug. contact me if it has any problem.

wu-sheng:
@dmsolr Please continue.

wu-sheng:
@mrproliu  Please verify the case, the test still fails.

mrproliu:
I'm trying to use `${TOOLS_HOME}/skywalking-validator-tools.jar` check the reason.
throw IOException when case servlet closes HttpServletResponse. 
image

wu-sheng:
You should not use it directly. It need mock collector running AFAIK.

wu-sheng:
I am running the test again to recheck stability

wu-sheng:
@dmsolr Please review

test/plugin/scenarios/netty-socketio-scenario/src/main/webapp/WEB-INF/web.xml
```diff
@@ -22,6 +22,10 @@
          version="3.1">
     <display-name>skywalking-netty-socketio-scenario</display-name>
 
+    <listener>
+        <listener-class>org.apache.skywalking.apm.testcase.netty.socketio.ContextListener</listener-class>
+    </listener>
+
```
dmsolr: I think this will cause unstable. If possible, move it to health-check.

mrproliu: resolved. change to the health check servlet.

PR #3721, opened 2019-10-27
Fix missing fields in scope-definitions.md

wu-sheng:
Doc fix

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -79,10 +79,6 @@ pipeline {
                     }
 
                     post {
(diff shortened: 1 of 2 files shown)


kezhenxu94:
I'm going to make CI skip when **only **/*.md** is modified

wu-sheng:
> I'm going to make CI skip when **only /*.md is modified

Make sense to me.

wu-sheng:
I am holding this PR to verfiy #3722

PR #3722, opened 2019-10-27
[CI] Skip CI when only docs are modified

kezhenxu94:
Skip CI when only docs are modified

### DIFF
--- a/tools/ci/changed.sh
+++ b/tools/ci/changed.sh
@@ -16,43 +16,59 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-patterns=()
-any_of=0
+# Detect files' changed set
+# --includes: includes these files when detecting changed file sets, defaults to ^.*$, meaning all files will be checked
+# --excludes: excludes these files when detecting changed file sets
+# exit with status code 0 if no changed file matches the patterns, otherwise exit with status code non 0
+
+includes=('^.*$')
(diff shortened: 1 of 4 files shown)


tools/ci/changed.sh
```diff
@@ -18,12 +18,18 @@
 
 patterns=()
 any_of=0
+all_of=0
 
+# --any-of: requires any of the changed files matche the given patterns
+# --all-of: requires all of the changed files matche the given patterns
```
wu-sheng: Is this the right name? From your usage, if the all-of matched, the job is not working. But for the any-of, if  matched, the job runs. Seems different logic.

wu-sheng: exclusion seems a better name?

kezhenxu94: this script is nothing related to the Jenkins logic from the file name `tools/ci/changed.sh`, it just determines whether (all of/any of) the files matching the given patterns changed or not, the usage sounds something like `do all of the changed files match the given patterns`, or `does any of the changed files match the given patterns`, `exclusion` seems also ok to me but may be less semantic

kezhenxu94: Jenkins just determine whether to run based on the result of this script

wu-sheng: Gor your point. It is ok to unchange.

PR #3724, opened 2019-10-27
[test/plugin] container run.sh optimize

arugal:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?
Set the ```curl```  visit to ```entryService``` and ```receiveData``` timeout to 1 minute

### DIFF
--- a/test/plugin/containers/jvm-container/docker/run.sh
+++ b/test/plugin/containers/jvm-container/docker/run.sh
@@ -76,11 +76,11 @@ healthCheck http://localhost:12800/status
 healthCheck ${SCENARIO_HEALTH_CHECK_URL}
 
 echo "To visit entry service"
-curl -s --max-time 3 ${SCENARIO_ENTRY_SERVICE}
+curl -s --max-time 60 ${SCENARIO_ENTRY_SERVICE}
 sleep 5
 
 echo "To receive actual data"
-curl -s --max-time 3 http://localhost:12800/receiveData > ${SCENARIO_HOME}/data/actualData.yaml
+curl -s --max-time 60 http://localhost:12800/receiveData > ${SCENARIO_HOME}/data/actualData.yaml
 [[ ! -f ${SCENARIO_HOME}/data/actualData.yaml ]] && exitOnError "${SCENARIO_NAME}-${SCENARIO_VERSION}, 'actualData.yaml' Not Found!"
 
 echo "To validate"
--- a/test/plugin/containers/tomcat-container/docker/run.sh
+++ b/test/plugin/containers/tomcat-container/docker/run.sh
@@ -66,11 +66,11 @@ healthCheck http://localhost:12800/status 10
 healthCheck ${SCENARIO_HEALTH_CHECK_URL}
 
 echo "To visit entry service"
-curl -s --max-time 3 ${SCENARIO_ENTRY_SERVICE}
+curl -s --max-time 60 ${SCENARIO_ENTRY_SERVICE}
 sleep 5
 
 echo "To receive actual data"
-curl -s --max-time 3 http://localhost:12800/receiveData > ${SCENARIO_HOME}/data/actualData.yaml
+curl -s --max-time 60 http://localhost:12800/receiveData > ${SCENARIO_HOME}/data/actualData.yaml
 [[ ! -f ${SCENARIO_HOME}/data/actualData.yaml ]] && exitOnError "${SCENARIO_NAME}-${SCENARIO_VERSION}, 'actualData.yaml' Not Found!"
 
 echo "To validate"


wu-sheng:
Why access so many times?

test/plugin/containers/jvm-container/docker/run.sh
```diff
@@ -76,11 +76,11 @@ healthCheck http://localhost:12800/status
 healthCheck ${SCENARIO_HEALTH_CHECK_URL}
 
 echo "To visit entry service"
-curl -s --max-time 3 ${SCENARIO_ENTRY_SERVICE}
+curl -s --max-time 60 ${SCENARIO_ENTRY_SERVICE}
```
wu-sheng: If access too many, the mock collector may be OOM.

test/plugin/containers/tomcat-container/docker/run.sh
```diff
@@ -66,11 +66,11 @@ healthCheck http://localhost:12800/status 10
 healthCheck ${SCENARIO_HEALTH_CHECK_URL}
 
 echo "To visit entry service"
-curl -s --max-time 3 ${SCENARIO_ENTRY_SERVICE}
+curl -s --max-time 60 ${SCENARIO_ENTRY_SERVICE}
 sleep 5
 
 echo "To receive actual data"
-curl -s --max-time 3 http://localhost:12800/receiveData > ${SCENARIO_HOME}/data/actualData.yaml
+curl -s --max-time 60 http://localhost:12800/receiveData > ${SCENARIO_HOME}/data/actualData.yaml
```
wu-sheng: This is only trying to access receive data, but you don't verify. Why add this?

test/plugin/containers/jvm-container/docker/run.sh
kezhenxu94: this should be timeout, not retry count

arugal:
> The changes don't make sense to me. I assume, you should consider to change verify times, not read time.

There are two scenarios:
1. ```EntryService``` could not respond, it could be a code bug or something, so we need to set a timeout.
2. ```EntryService``` takes 7 seconds or more to respond, but eventually it will respond, it has to do with code or vm, so we need a long timeout.

So what we need is a timeout, not a retry

> This is only trying to access receive data, but you don't verify. Why add this?

I can rollback

wu-sheng:
1. If it is a test code bug, we should fix it. Rather than supporting this in tesr framework.
1. What codes need 3 seconds for test? We have given health check to do preparation, right?
1. We could make longer timeout, but you are adding retry time, which are different.

arugal:
>  1. If it is a test code bug, we should fix it. Rather than supporting this in tesr framework.

Yes, but should we let the ```agent-plugin-tests``` end and show failure instead of remaining pending

> 2. What codes need 3 seconds for test? We have given health check to do preparation, right?

for [kafka-scenario](https://github.com/apache/skywalking/blob/master/test/plugin/scenarios/kafka-scenario/src/main/java/test/org/apache/skywalking/apm/testcase/kafka/controller/CaseController.java), may have a 10-second waiting for ```consumer``` to complete,  but waiting is required, otherwise the use case might fail because of the lack of ```operationNames```

> 3. We could make longer timeout, but you are adding retry time, which are different.

``` -m, --max-time <seconds> Maximum time allowed for the transfer``` that's the help of curl,  what did I miss？

wu-sheng:
I mean, for Kafka, the lack of operation name may not relate to this. 
Do you check why the operation name missing? My guess is the message consuming take more time, rather than the entry service timeout. But the timeout clearly is not for that. Could you explain?

arugal:
> My guess is the message consuming take more time

Yes, so I need to wait for the ```ConsumerThread``` to finish in ```EntryService```

https://github.com/apache/skywalking/blob/7f69a73d1312888a20489e8429dd65f3d10a52b1/test/plugin/scenarios/kafka-scenario/src/main/java/test/org/apache/skywalking/apm/testcase/kafka/controller/CaseController.java#L66-L87

> rather than the entry service timeout. But the timeout clearly is not for that.

I think I've answered that one, why a timeout? why a long timeout, ```kafka-scenario``` is an example of an ```EntryService``` that requires 3 seconds or more

> Do you check why the operation name missing?

I will provide the corresponding test code.

wu-sheng:
I think kafka consume period is configurable, which could be in milliseconds, is that right?  My point is that, it may need more time, but it is not timeout, it should be a validation retry.

wu-sheng:
I am not arguing about the more time thing, it is about how to set more time.

wu-sheng:
@dmsolr Could you help about this?

test/plugin/containers/jvm-container/docker/run.sh
dmsolr: It is still a single thread to visit the mok-collector without retry. Increasing the timeout time why it could improve the success rate. Make tests more stable.
There is no retry but will timeout in 60s. I think this change is necessary.

dmsolr:
I think it is not business about any test cases.
Before #3716, the test-framework visit to any URL without a timeout. But we spot which will block for a long time(serval hours) in #3698 .
I think @arugal means what a 3s is too short to visit `EntryService` so that unstable in lack of resource.

For many cases,  we get the responses immediately, but sometimes it will take more time. I think 30-60s is enough.

wu-sheng:
My question is why accessing `EntryService` needs 30s? Is this the reasonable case for codes? Especially, we are talking about test codes. If we accept this kind of latency, what we expect from the time of running all cases?

wu-sheng:
Personally, I don't believe there is a service, no matter we wrote or another OSS, which requires so long time to respond in such low traffic access.

wu-sheng:
I know @arugal 's meaning, but could you give a real case, and , I would like to discuss it. @dmsolr What codes you met need 30 sec?

dmsolr:
In fact, I don't meet yet. But I think it is no harm.

wu-sheng:
> In fact, I don't meet yet. But I think it is no harm.

Actually, it is. Because we will never know who will contribute test codes need to wait for a long time, and causing our INFRA resource with meaningless.

arugal:
Hi, I may have not expressed clearly, please give me a little time, I will comb through and provide corresponding examples. @wu-sheng @dmsolr @kezhenxu94

wu-sheng:
Sure, take your time.

arugal:
1. This is the Jenkins log screenshot of undertow-scenario submitted earlier, because the bug in the use case blocked the test for 2 hours. 
(这是之前提交```undertow-scenario```时的Jenkins日志截图，因为用例的bug阻塞了测试2小时)
WechatIMG7
there is no possibility of retrying, so we need to timeout
2.  The ```EntryService``` of ```kafka-scenario``` needs to wait for ```ConsumerThread``` to complete before it can respond, which may take up to 10s. If you don't wait or time out by 3 seconds, the mock-collector may not have received all the data yet. Here, EntryService blocking is reasonable in my view (```kafka-scenario```的```EntryService```需要等待ConsumerThread完成才能响应，最长可能等待10s, 如果不等待或者3秒超时，mock-collector可能还没有收到全部的数据，这儿EntryService阻塞我认为是合理的)
https://github.com/apache/skywalking/blob/b3da85b52684834f0e39a02ef6998499db81327a/test/plugin/scenarios/kafka-scenario/src/main/java/test/org/apache/skywalking/apm/testcase/kafka/controller/CaseController.java#L66-L87
https://github.com/apache/skywalking/blob/b3da85b52684834f0e39a02ef6998499db81327a/test/plugin/scenarios/kafka-scenario/src/main/java/test/org/apache/skywalking/apm/testcase/kafka/controller/CaseController.java#L158-L191

3. I modified ```undertow-scenario``` on fork project, ```EntryService``` slept for 8 seconds before responding, and the probability of missing ```operationNames``` was almost 100 percent. However, we changed the timeout time of ```EntryService``` to 30 seconds or 60 seconds, which could be avoided, not affecting normal use cases, but improving stability for special use cases like ```kafka ```
(我修改了fork项目上的undertow-scenario，EntryService休眠8秒才响应，缺少operationNames的概率几乎百之百，但是我们将EntryService的超时时间改到30秒或者60秒就可以避免，对于正常的用例没有影响，对于kafka这样特殊的用例能提高稳定性)

https://github.com/arugal/skywalking/blob/113e0811bb7b091ac7373265cd23dfab4393b386/test/plugin/scenarios/undertow-scenario/config/actualData.yaml#L1-L48

wu-sheng:
First of all, is really kafka consumer requiring 10s? I remember kafka read period could be configurated? Do you try to resolve this?
If we accept 60s, even 10s, then every case could do a long sleep and say, we want to be more stable. But really?

PR #3725, opened 2019-10-28
Refactor:extract common module postgresql-common

aderm:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___
### New feature or improvement
- Describe the details and related test reports.

PR #3727, opened 2019-10-28
Update test pg image version and plugin support list 

aderm:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___
1.update pg image to 10.5-alpine;
2.update test support list.


wu-sheng:
A question related to #3725, this PR seems to have no plugin codes, then why we need PostgreSQL common module? I know it is harmless, but I am assuming you are going to reuse those codes.

wu-sheng:
The versions of the driver seem crazy, do you have recommendation to reduce some? Or @dmsolr @kezhenxu94 the workload 4 seems Inevitable, WDYT?

wu-sheng:
Additional, I think as we have optimized the test running conditions, we may could have a new one. I hope the 4 should be enough to provide all tests running in one hour. We still have a long list of plugins, but most of them have less versions.

aderm:
> A question related to #3725, this PR seems to have no plugin codes, then why we need PostgreSQL common module? I know it is harmless, but I am assuming you are going to reuse those codes.

Maybe I have misunderstood before. pg-8.x-plugin already supports pg10. At present, 8-plugin is already supported. When there is a big pg-jdbc version, it should be put into 9.x-plugin. I think so, everyone thinks it can be Suitable?

wu-sheng:
> Maybe I have misunderstood before. pg-8.x-plugin already supports pg10. At present, 8-plugin is already supported. When there is a big pg-jdbc version, it should be put into 9.x-plugin. I think so, everyone thinks it can be Suitable?

If in that case, we may not need the PostgreSQL common for now, because you don't know the next version, you don't know what they look like, and what plugin do you need to build. So, common is not the real common package :)

wu-sheng:
Is it OK to revert it back?

aderm:
In fact, I have a plan that will increase the intercept extension function of pg. The contents of these extensions will be placed in postgresql-9.x-plugin, but the current priority will be lower.

wu-sheng:
> In fact, I have a plan that will increase the intercept extension function of pg. The contents of these extensions will be placed in postgresql-9.x-plugin, but the current priority will be lower.

If this is an additional feature, especially for 9.x plugin, it is OK to do. It is a separated situation of #3725

kezhenxu94:
> The versions of the driver seem crazy, do you have recommendation to reduce some? Or @dmsolr @kezhenxu94 the workload 4 seems Inevitable, WDYT?

I prefer combining them into one workload, according to my attempt and comment [here](https://github.com/apache/skywalking/pull/3714#issuecomment-546408930), combining them into one and we can still make them run in different nodes(without stashing, every node compiles like what we do now), the benefit is that it will make full use of the resources, auto occupy all available executors whenever possible, and queueing automatically when no more executors available

wu-sheng:
But like I said last time, the 100m dist package is my concern. The task is already fully used. Even we set static allocate rules.

wu-sheng:
And every time, we have to rerun all tasks if only one timeout. Make sense?

kezhenxu94:
> But like I said last time, the 100m dist package is my concern.

"without stashing", what I mean is every node will run compile-test process, like what we are doing now, no networking transmission

> The task is already fully used. Even we set static allocate rules.

I don't think so, it's fully used when all the nodes are busy, but if we have only one PR, there is 3 workloads, occupying 3 vm nodes, and the other 2 vm nodes are idle, after combining them into one workload, no matter how many PR's are opened, we will try to make as many nodes busy as possible

> And every time, we have to rerun all tasks if only one timeout. Make sense?

that's what I would propose, to retry the failed stage/task automatically ( for a certain times, say 3 times), this can be also done within Jenkins itself, without changing test framework

wu-sheng:
Lets open another issue to discuss this. Seems like a longer discussion.

PR #3728, opened 2019-10-28
Revert "Refactor:extract common module postgresql-common"

wu-sheng:
Reverts apache/skywalking#3725

According to https://github.com/apache/skywalking/pull/3727#issuecomment-546852048. This is an unnecessary refactor. Even it is harmless, it don't really support a new version as the PR described before, the new versions of JDBC driver have been supported by existing codes. So I am running rolling back.

FYI @aderm


aderm:
ACK. It is not necessary at present. Can I wait until later if I need to add it?

wu-sheng:
What dou you mean wait?

aderm:
The latest PG version is already PG12, and it is not clear whether current version can support it.

wu-sheng:
You could choose any time you need to add it back. I just mean, we don't need it now.

aderm:
ACK

PR #3742, opened 2019-10-30
Add NetEase to our user list.

wu-sheng:
A new user informed in the discuss channel.


wu-sheng:
@kezhenxu94 All tests skip successfully with the only doc changed.

PR #3748, opened 2019-10-30
Remove two max size 500 limit.

wu-sheng:
Follow #3741

### DIFF
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/cache/NetworkAddressInventoryCacheEsDAO.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/cache/NetworkAddressInventoryCacheEsDAO.java
@@ -84,7 +84,7 @@ public class NetworkAddressInventoryCacheEsDAO extends EsDAO implements INetwork
         try {
             SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
             searchSourceBuilder.query(QueryBuilders.rangeQuery(NetworkAddressInventory.LAST_UPDATE_TIME).gte(lastUpdateTime));
-            searchSourceBuilder.size(500);
+            searchSourceBuilder.size(Integer.MAX_VALUE);
 
             SearchResponse response = getClient().search(NetworkAddressInventory.INDEX_NAME, searchSourceBuilder);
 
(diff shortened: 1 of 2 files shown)


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/cache/NetworkAddressInventoryCacheEsDAO.java
```diff
@@ -84,7 +84,7 @@ public NetworkAddressInventoryCacheEsDAO(ElasticSearchClient client) {
         try {
             SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
             searchSourceBuilder.query(QueryBuilders.rangeQuery(NetworkAddressInventory.LAST_UPDATE_TIME).gte(lastUpdateTime));
-            searchSourceBuilder.size(500);
+            searchSourceBuilder.size(Integer.MAX_VALUE);
```
JaredTan95: Set source-builder's size to `Integer.MAX_VALUE ` may cause query time cost higher.

wu-sheng: @JaredTan95 Yes, for sure. But it is hard to guarantee to load all? If not, the update is not working.

wu-sheng: And every load has this condition
```java
List<NetworkAddressInventory> addressInventories = addressInventoryCacheDAO.loadLastUpdate(System.currentTimeMillis() - 60000);
```

The inventory is being reloaded only the update happens at the last minute. And timer runs every 10s(static config today)

JaredTan95: Yeah, I got your point.

JaredTan95: Considering our scenario here, the impact can be ignored in fact.

PR #3761, opened 2019-10-31
Rabbitmq instrument consumer class correctly to get duration reported…

viswaramamoorthy:
… accurately

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
RabbitMQ consumer duration is reported incorrectly due to instrumenting com.rabbitmq.client.impl.ConsumerDispatcher.handleDelivery method which spins up a thread to deliver message to consuming method

- How to fix?
Instrumenting implementations of com.rabbitmq.client.Consumer to resolve
___


wu-sheng:
Your test should be added through jenkins file. Find the fastest workload and group, add your new case there. And make sure test passed.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/HierarchyMatch.java
```diff
@@ -46,6 +47,11 @@ private HierarchyMatch(String[] parentTypes) {
         this.parentTypes = parentTypes;
     }
 
+    private HierarchyMatch(String[] parentTypes, boolean interfaceOnlyMatch) {
+        this(parentTypes);
+        this.interfaceOnlyMatch = interfaceOnlyMatch;
+    }
+
```
dmsolr: Usually, we don't recommend to change the core API.

apm-sniffer/apm-sdk-plugin/rabbitmq-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rabbitmq/define/RabbitMQConsumerInstrumentation.java
```diff
@@ -71,6 +71,6 @@
 
     @Override
     protected ClassMatch enhanceClass() {
-        return MultiClassNameMatch.byMultiClassMatch(ENHANCE_CLASS_PRODUCER);
+        return HierarchyMatch.byHierarchyMatch(new String[] {ENHANCE_CLASS_PRODUCER}, true);
```
dmsolr: `byHierarchyMatch` doesn't recommend, too. If have to do, I think you should explain and discuss more. :)

test/plugin/scenarios/rabbitmq-scenario/src/main/java/test/apache/skywalking/testcase/rabbitmq/Application.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package test.apache.skywalking.testcase.rabbitmq;
```
dmsolr: Please rename the package to `org.apache.skywalking.apm.testcase.*`

dmsolr:
After that, please add `rabbitmq-scenario` task into the Jenkinsfile.

apm-sniffer/apm-sdk-plugin/rabbitmq-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rabbitmq/define/RabbitMQConsumerInstrumentation.java
wu-sheng: Yes. We don't recommend because of potential side effect. Such as someone implement the interfaces.

test/plugin/scenarios/rabbitmq-scenario/src/main/java/test/apache/skywalking/testcase/rabbitmq/Application.java
kezhenxu94: > Please rename the package to `org.apache.skywalking.apm.testcase.*`

I remember the package name in `test/plugin` should not start with `org.apache.skywalking`?

dmsolr: Usually, We recommend `org.apache.skywalking.apm.testcase.*`. 
But testcase needs to enhance some class, `test.org.apache.skywalking.apm.testcase` is also good. That likes `spring-scenario`.

wu-sheng: > I remember the package name in test/plugin should not start with org.apache.skywalking?

These cases are not following the e2e rules. Because we want to make sure explicit enhance. All classes in `org.apache.skywalking.apm.testcase.*` setup by `SkyWalkingAgent#allSkyWalkingAgentExcludeToolkit`

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/HierarchyMatch.java
viswaramamoorthy: RabbitMQ client pretty much leaves the message handling implementation to a custom class or a framework like Camel's RabbitMQ module or a Spring. We need to discover and instrument implementation classes of com.rabbitmq.client.Consumer

We cannot use HierarchyMtach as is. if we did, in this case, it might discover an interface that extends from com.rabbitmq.client.Consumer which will result in undesired behavior

So, HierarchyMtach is lacking and ideally should have a feature to just match interfaces only.

apm-sniffer/apm-sdk-plugin/rabbitmq-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rabbitmq/define/RabbitMQConsumerInstrumentation.java
viswaramamoorthy: See my note above that explains why we should do this

viswaramamoorthy:
@dmsolr @wu-sheng I have answered why we should make changes to core api. once you review and get back, i can make changes to test case suggestions to get ready for final merge. let me know

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/HierarchyMatch.java
wu-sheng: If an interface extends, then what? You can't intercept an interface, because nothing invoked there. Wha is your point? And what is instrumentation for interface only meaning? How to enhance an interface?

viswaramamoorthy: Follow this...

If I give an interface as input (e.g. com.rabbitmq.client.Consumer), I want classes that are implementing this interface (com.rabbitmq.client.Consumer) to be instrumented. I do not want interfaces extending from com.rabbitmq.client.Consumer instrumented. Is it possible with existing core API?

wu-sheng: Yes, that is `DeclaredInstanceMethodsInterceptPoint` for.

wu-sheng: @viswaramamoorthy Have you tried this?

viswaramamoorthy: I am not how to make it work with DeclaredInstanceMethodsInterceptPoint. I do not know much about the code. 

I have removed my changes in HierarchyMatch class (which you are saying, I should not). I am directly using HierarchyMatch as is and instrumenting RabbitMQ interface and its underlying implementation. It is working,.

Review this. if it is fine, I can do changes recommended for legacy test case migration.

If this is not fine, you can make changes to this PR to fix the issue I have reported. Let me know if you have questions on the issue that I reported.

wu-sheng: Replace `InstanceMethodsInterceptPoint` to `DeclaredInstanceMethodsInterceptPoint` in your instrumentation define. It is an easy change. They you could test again. Then in the hierarchy case, it should workw as your exepected. @viswaramamoorthy

This interceptor point mean this method must be implemented in this method, rather that existing the parent class.

I think you are closing the right one.

wu-sheng:
@viswaramamoorthy  Have you checked? Is the way I proposed working for your inherit scenario?

dmsolr:
@viswaramamoorthy Hi, Please rename the package of testcase. And then add it as a Jenkinsfile task to run.

wu-sheng:
@dmsolr Which workload and group should be added? Or should we start workload 4 for now? @kezhenxu94

wu-sheng:
The plugin test framework doc is ready, #3786. Please take a look.

@viswaramamoorthy After you confirm the latest codes are working for all your cases.
You could move the RabbitMQ test case in another PR. After that is merged, then we recheck this.

kezhenxu94:
> @dmsolr Which workload and group should be added? Or should we start workload 4 for now? @kezhenxu94

Workload4 has been set up, use file name `Jenkinsfile-Agent-Test-4` and ping me, and I'll enable this job,

dmsolr:
@kezhenxu94 Hi, I pushed the Jenkinsfile on #3787 . Please review and enable it.

viswaramamoorthy:
Changes in this PR are verified and works for the bug reported.

I have raised a separate PR for RabbitMQ e2e test case migration with package renamed. it is https://github.com/apache/skywalking/pull/3788

wu-sheng:
@viswaramamoorthy  The test case has been merged, let's work on this one.

PR #3765, opened 2019-11-01
make query max window size configurable.

JaredTan95:

- [x] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

After merged pr https://github.com/apache/skywalking/pull/3748
___
### Bug fix
- Bug description.

Skywalking OAP update`NetworkAddressInventory` and `ServiceInventory` will cause exceptions as follows:

```
Caused by: org.elasticsearch.client.ResponseException: method [POST], host [http://localhost:9200], URI [/network_address_inventory/type/_search?typed_keys=true&ignore_unavailable=false&expand_wildcards=open&allow_no_indices=true&search_type=query_then_fetch&batched_reduce_size=512], status line [HTTP/1.1 500 Internal Server Error]
{"error":{"root_cause":[{"type":"query_phase_execution_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [2147483647]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."}],
``` 

- How to fix?

The reason is that elasticsearch window size is`10000` default. If we do not modify `elasticseach.yml`.
We can adjust this size in configuration and set `10000` as default value.
___


wu-sheng:
Could we make it in the configuration and 10k as default?

PR #3769, opened 2019-11-01
httpasyncclient plugin adpter for ES. if out is exit span, do not propagating context

aderm:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___
In the process of developing the rest-high-level of the ES-6.x plugin, because the underlying API uses the httpasycclient implementation, both plugins are intercepted during a call, so that comparisons will appear on the page display one trace. For the following reasons, consider that if the outer layer is exit span, httpasycclient does not create span.
1. reduce the burden of back-end registration
2. when the outer layer is an exit span, the display of httpasycclient span has no specific meaning.

WechatIMG2
WX20191101-190941@2x


wu-sheng:
You need to explain your PR, it is hard to explain.

apm-sniffer/apm-sdk-plugin/httpasyncclient-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpasyncclient/v4/SessionRequestConstructorInterceptor.java
```diff
@@ -31,7 +31,7 @@
 public class SessionRequestConstructorInterceptor implements InstanceConstructorInterceptor {
     @Override
     public void onConstruct(EnhancedInstance objInst, Object[] allArguments) {
-        if (ContextManager.isActive()) {
+        if (ContextManager.isActive() && !ContextManager.activeSpan().isExit()) {
```
dmsolr: An ES created span is not `ExitSpan`?

apm-sniffer/apm-sdk-plugin/httpasyncclient-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpasyncclient/v4/SessionRequestCompleteInterceptor.java
```diff
@@ -34,21 +34,23 @@
 public class SessionRequestCompleteInterceptor implements InstanceMethodsAroundInterceptor {
 
     public static ThreadLocal<HttpContext> CONTEXT_LOCAL = new ThreadLocal<HttpContext>();
+    public static ThreadLocal<Boolean> CONTEXT_LOCAL_NOT_EXIT = new ThreadLocal<Boolean>();
```
dmsolr: You didn't call `remove()`, but I think it's necessary. Right?

apm-sniffer/apm-sdk-plugin/httpasyncclient-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpasyncclient/v4/SessionRequestConstructorInterceptor.java
aderm: when ES exit span. this do not propagating context

apm-sniffer/apm-sdk-plugin/httpasyncclient-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpasyncclient/v4/SessionRequestCompleteInterceptor.java
aderm: ACK.

apm-sniffer/apm-sdk-plugin/httpasyncclient-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpasyncclient/v4/HttpAsyncRequestExecutorInterceptor.java
```diff
@@ -48,12 +48,14 @@
     @Override
     public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
         HttpContext context = CONTEXT_LOCAL.get();
+        Boolean isOutNotExit = CONTEXT_LOCAL_NOT_EXIT.get();
```
wu-sheng: Why need this? Is this method not in the main context thread?

wu-sheng: `ContextManager.activeSpan().isExit()` should be available if this is in the main thread, and`CONTEXT_LOCAL_NOT_EXIT` is not required.

wu-sheng: Because there is creating exit span in there, so it makes sense only existing in the main thread. You need to make the change very clear, otherwise, hard to review.

aderm: it is not in the main context thread，flag isOutNotExit used to determine if the main thread is not exit span. The child thread only creates an exit span when the main thread is a non-exit span.

aderm: > `ContextManager.activeSpan().isExit()` should be available if this is in the main thread, and`CONTEXT_LOCAL_NOT_EXIT` is not required.

In the main thread, the ContextManager.activeSpan() span in some scenarios is an Entry span, not an exit span, such as the example in the test  httpasyncclint scenario.

wu-sheng: I think you still miss the context. The whole scenario is `ES exit span`->`http exit span`,then only ES span left. 

`ContextManager.activeSpan()` could be
1. No exist, because http client is the first one.
1. Local span or entry span, because some other span exists.
1. Exit span exists, such as ES plugin is working.

So what? We are talking about the async HTTP client logic.  
First, the HTTP async span should start in the main thread, no matter what old codes look like. Then change into the `async` mode, and propagate the span into sub-thread, and tag, log or anything you need to do.
If there is a callback for HTTP client, you could choose to create a new segment and local span or just simply tag because such as it is invoked inside ES client.

aderm: It seems that I have a problem before, the calculated span time should be wrong, and does not include the execution time of httpasycclient. Maybe I understand what you mean, I will try again.

There is a problem, in the asynchronous case, you need to create an exit span in the main thread, trigger prepareForAsync. When ContextManager.activeSpan() is Local span or entry span, you need to create create a new segment and local span in the child thread. At this time, you find that there is a duplicate creation of the exit span in the code, in the main thread. One, then the callback will appear again; don't know if this understanding is correct?

wu-sheng: Sure. Take your time. You only could have only one exit span with the accurate duration, and tags.

aderm:
Httpasyncclient plugin When the external is exit span, there is no need to perform snapshot delivery. This problem has been solved. Reference #3803 , so this PR closed.

PR #3770, opened 2019-11-02
Add change log for 6.5.0

kezhenxu94:
Hi @apache/skywalking-committers , this patch summarizes the non-trivial changes in [milestone 6.5.0](https://github.com/apache/skywalking/milestone/36?closed=1) and will be the release logs in our version 6.5.0, so please check that your non-trivial contributions are included in it, thanks,

And since @x22x22 is still working on the enhancement of UI topology, I'll add it here when his work is merged into UI repo, and then we will merge this if everything is OK

FYI @wu-sheng , I wrote a `go` program [draft_change_log.go](https://gist.github.com/kezhenxu94/90cb6e4c51183fd0fa44a5676bb0f4f7) to help to draft this change log based on the merged pull request in this milestone, and this may be reused in the future releases, though we may still manually adjust something


wu-sheng:
A tool make sense to me. If you want to move it into Apache codebase, let's discuss it how in another thread in mail list. You could propose how to use and how to host in ASF.

kezhenxu94:
> * Should notify through mail list about changelog for this release
> * Why every first letter of the word is upper case? It is unusual in English.

2. I just use the `strings.Title(string)` function in Golang to format the pull request title, does it make sense? I think the words in a title are capitalized in English

kezhenxu94:
> A tool make sense to me. If you want to move it into Apache codebase, let's discuss it how in another thread in mail list. You could propose how to use and how to host in ASF.

Will raise a discussion after I polish the tool

wu-sheng:
I remember changelog is in normal sentence case, from my understanding.

PR #3774, opened 2019-11-03
[hotfix] Remove argument BUILID_ID from run.sh

dmsolr:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

Having to configuration BUILD_ID and depending on Jenkinsfile's NO. is absurd.

### DIFF
--- a/Jenkinsfile-Agent-Test
+++ b/Jenkinsfile-Agent-Test
@@ -64,7 +64,7 @@ pipeline {
                 }
             }
             steps {
-                sh './mvnw -f test/plugin/pom.xml clean package -DskipTests -Dbuild_id=${BUILD_NO} docker:build'
+                sh './mvnw -f test/plugin/pom.xml clean package -DskipTests docker:build'
             }
         }
         stage('Test Cases Report (169)') {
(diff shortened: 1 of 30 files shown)


wu-sheng:
@dmsolr There are many Jenkins files

Jenkinsfile-Agent-Test
```diff
@@ -63,12 +64,12 @@ pipeline {
                 }
             }
             steps {
-                sh './mvnw -f test/plugin/pom.xml clean package -DskipTests -Dbuild_id=wl1_${BUILD_ID} docker:build'
+                sh './mvnw -f test/plugin/pom.xml clean package -DskipTests -Dbuild_id=${BUILD_NO} docker:build'
```
kezhenxu94: I don't think we need `--build_id=wl1_${BUILD_ID}` any more

test/plugin/run.sh
```diff
@@ -20,7 +20,7 @@ home="$(cd "$(dirname $0)"; pwd)"
 scenario_name=""
 parallel_run_size=1
 force_build="off"
-build_id="local"
+build_id="${BUILD_NO:=local}"
```
kezhenxu94: Since you have use the environment variable here

wu-sheng:
@kezhenxu94 Recheck please, if it is OK, let's still merge this into 6.5.0

Issue #3762, opened 2019-11-01
SkyWalking 6.4.0 is incompatible with redisson 3.11.1


czq080:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS and JRE?
SkyWalking6.4.0,linux,jdk1.8
- Which company or project?

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.
**ERROR 2019-10-30 18:30:19:614 redisson-netty-4-3 ConstructorInter : ConstructorInter failure.
java.lang.NoSuchMethodError: org.redisson.client.RedisClientConfig.getAddress()Ljava/net/URI;
        at org.apache.skywalking.apm.plugin.redisson.v3.RedisConnectionMethodInterceptor.onConstruct(RedisConnectionMethodInterceptor.java:109)
        at org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ConstructorInter.intercept(ConstructorInter.java:69)
        at org.redisson.client.RedisConnection.<init>(RedisConnection.java)
        at org.redisson.client.handler.RedisConnectionHandler.createConnection(RedisConnectionHandler.java:36)
        at org.redisson.client.handler.BaseConnectionHandler.channelRegistered(BaseConnectionHandler.java:55)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRegistered(AbstractChannelHandlerContext.java:161)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRegistered(AbstractChannelHandlerContext.java:147)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRegistered(AbstractChannelHandlerContext.java:140)
        at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRegistered(DefaultChannelPipeline.java:1381)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRegistered(AbstractChannelHandlerContext.java:161)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRegistered(AbstractChannelHandlerContext.java:147)
        at io.netty.channel.DefaultChannelPipeline.fireChannelRegistered(DefaultChannelPipeline.java:824)
[...]

PR #3780, opened 2019-11-04
redisson plugin support high version

zhaoyuguang:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
resolve https://github.com/apache/skywalking/issues/3762

Low version（3.6.0）
image

High version（3.11.5）
image

___
- Use reflection to avoid `no such method`, because they are all modified in the construction method, so the performance will not be much lower

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/redisson-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/redisson/v3/util/ClassUtil.java
+++ b/apm-sniffer/apm-sdk-plugin/redisson-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/redisson/v3/util/ClassUtil.java
@@ -15,20 +15,18 @@
  * limitations under the License.
  *
  */
-package org.apache.skywalking.apm.plugin.redisson.v3;
+package org.apache.skywalking.apm.plugin.redisson.v3.util;
 
-import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
-import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
+import java.lang.reflect.Field;
 
 /**
- * RedisClient is the link between RedisConnection and ConnectionManager.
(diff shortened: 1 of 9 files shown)


zhaoyuguang:
resolve https://github.com/apache/skywalking/issues/3762

wu-sheng:
This interceptor is for client management, right? So, even using reflect should not cause a performance concern. Am I right?

apm-sniffer/apm-sdk-plugin/redisson-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/redisson/v3/util/ClassUtil.java
```diff
@@ -0,0 +1,32 @@
...
+package org.apache.skywalking.apm.plugin.redisson.v3.util;
+
+import java.lang.reflect.Field;
+
+/**
+ * @author zhaoyuguang
+ */
+public class ClassUtil {
+
+    public static Object getObjectField(Object obj, String name) throws NoSuchFieldException, IllegalAccessException {
```
wu-sheng: This is a reflect API, we should comment that this method should only be used in low frequency. It should not use in trace context, but just in the metadata preparation stage.

zhaoyuguang:
> This interceptor is for client management, right? So, even using reflect should not cause a performance concern. Am I right?

Yes.

apm-sniffer/apm-sdk-plugin/redisson-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/redisson/v3/util/ClassUtil.java
zhaoyuguang: Thanks, the comment had add

wu-sheng:
Workload 2 has over 200 cases, and group1 is full I think. Later, group2 should have available slots, and workload 1.

zhaoyuguang:
> Workload 2 has over 200 cases, and group1 is full I think. Later, group2 should have available slots, and workload 1.

OK

PR #3781, opened 2019-11-04
[test/plugin] Don't copy all optional plugins by default

dmsolr:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#3617 

1. fix BUILD_NO missing default values in containers/pom.xml.
2. disable copy all optional plugins

### DIFF
--- a/test/plugin/run.sh
+++ b/test/plugin/run.sh
@@ -173,7 +173,7 @@ with_plugins=$(grep "^withPlugins" ${scenario_home}/configuration.yml |sed -e "s
 
 if [[ -n "${running_mode}" ]]; then
     [[ -z "${with_plugins}" ]] && exitWithMessage \
-       "'withPlugins' has required configuration when 'runningMode' was set as 'optional_plugins' or 'bootstrap_plugins'"
+       "'withPlugins' is required configuration when 'runningMode' was set as 'optional_plugins' or 'bootstrap_plugins'"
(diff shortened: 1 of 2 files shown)


test/plugin/run.sh
```diff
@@ -172,9 +172,8 @@ running_mode=$(grep "^runningMode" ${scenario_home}/configuration.yml |sed -e "s
 with_plugins=$(grep "^withPlugins" ${scenario_home}/configuration.yml |sed -e "s/ //g" |awk -F: '{print $2}')
 
 if [[ -n "${running_mode}" ]]; then
-    # [[ -z "${with_plugins}" ]] ; then&& exitWithMessage \
-    #    "'withPlugins' has required configuration when 'runningMode' was set as 'optional_plugins' or 'bootstrap_plugins'"
-    [[ -z "${with_plugins}" ]] && with_plugins="*.jar"
+    [[ -z "${with_plugins}" ]] && exitWithMessage \
+       "'withPlugins' has required configuration when 'runningMode' was set as 'optional_plugins' or 'bootstrap_plugins'"
```
kezhenxu94:
```suggestion
       "'withPlugins' is required configuration when 'runningMode' was set as 'optional_plugins' or 'bootstrap_plugins'"
```

wu-sheng:
> Don't know why we do not support copy all optional plugins

Because it will become an unstable factor when the new optional plugin added in.

PR #3782, opened 2019-11-05
Enhance EtcdCoordinator if disconnected from etcd server for a long time

kuaikuai:

when EtcdCoordinator cannot connect to etcd cluster for a long time, 
the information registered in etcd will expire and refresh() function will not work.

- How to fix?
When refresh() finds that the key does not exist, an exception is thrown, and we catch the exception to rewrite the registration information.

### DIFF
--- a/oap-server/server-cluster-plugin/cluster-etcd-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/etcd/EtcdCoordinator.java
+++ b/oap-server/server-cluster-plugin/cluster-etcd-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/etcd/EtcdCoordinator.java
@@ -34,6 +34,8 @@ import org.apache.skywalking.oap.server.core.cluster.RemoteInstance;
 import org.apache.skywalking.oap.server.core.cluster.ServiceRegisterException;
 import org.apache.skywalking.oap.server.core.remote.client.Address;
 import org.apache.skywalking.oap.server.telemetry.api.TelemetryRelatedContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * @author Alan Lau
@@ -114,7 +116,6 @@ public class EtcdCoordinator implements ClusterRegister, ClusterNodesQuery {
             try {
                 client.refresh(key, KEY_TTL).send().get();
             } catch (Exception e) {
-                logger.error(e.getMessage(), e);
                 try {
                     client.put(key, json).ttl(KEY_TTL).send().get();
                 } catch (Exception ee) {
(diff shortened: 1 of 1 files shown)


wu-sheng:
@kezhenxu94 @JaredTan95  @wayilau Could you recheck this? Look like make sense.

oap-server/server-cluster-plugin/cluster-etcd-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/etcd/EtcdCoordinator.java
```diff
@@ -113,7 +114,13 @@ private void renew(EtcdClient client, String key, String json) {
             try {
                 client.refresh(key, KEY_TTL).send().get();
             } catch (Exception e) {
-
+                try {
+                    client.put(key, json).ttl(KEY_TTL)
+                        .setRetryPolicy(new RetryNTimes(1, 10))
```
wayilau: it looks good to me. Just one point that there is a loop already, I think it needs not another retryPolicy. What about you ?

kezhenxu94: Retry 10 times every 1 millisecond? do we really need that frequently?

oap-server/server-cluster-plugin/cluster-etcd-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/etcd/EtcdCoordinator.java
```diff
@@ -113,7 +114,13 @@ private void renew(EtcdClient client, String key, String json) {
...
+                        .send()
+                        .get();
+                } catch (Exception ee) {
```
kezhenxu94: We'd better not swallow the exception, at least, log a warning message to notify users that the Etcd connection is unstable or something

oap-server/server-cluster-plugin/cluster-etcd-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/etcd/EtcdCoordinator.java
kuaikuai: you are right,  I remove retryPolicy.

oap-server/server-cluster-plugin/cluster-etcd-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/etcd/EtcdCoordinator.java
kuaikuai: ok

wu-sheng:
@kuaikuai Welcome to be 171st contributor

PR #3795, opened 2019-11-06
fix opentracing with tag error is failed

cyejing:

- [x ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
   at use opentracing api to tag "error"  is  failed . because tag not judge ERORR tag and call  errorOccurred()
- How to fix?
  show the code
___

### DIFF
--- a/apm-sniffer/apm-toolkit-activation/apm-toolkit-opentracing-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/opentracing/span/SpanSetTagInterceptor.java
+++ b/apm-sniffer/apm-toolkit-activation/apm-toolkit-opentracing-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/opentracing/span/SpanSetTagInterceptor.java
@@ -43,6 +43,8 @@ public class SpanSetTagInterceptor implements InstanceMethodsAroundInterceptor {
             activeSpan.setComponent(tagValue);
         } else if (Tags.PEER_SERVICE.getKey().equals(tagKey)) {
             activeSpan.setOperationName(tagValue);
+        } else if (Tags.ERROR.getKey().equals(tagKey) && "true".equals(tagValue)) {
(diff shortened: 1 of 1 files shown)


apm-sniffer/apm-toolkit-activation/apm-toolkit-opentracing-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/opentracing/span/SpanSetTagInterceptor.java
```diff
@@ -43,6 +43,8 @@ public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allA
             activeSpan.setComponent(tagValue);
         } else if (Tags.PEER_SERVICE.getKey().equals(tagKey)) {
             activeSpan.setOperationName(tagValue);
+        } else if (Tags.ERROR.getKey().equals(tagKey) && "true".equals(tagValue)) {
```
ascrutae: According [the open-tracing document](https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-tags-table), The type of `ERROR` tag is `bool`,  I think that it is incorrect to judge the type of `tagValue`

cyejing: https://github.com/apache/skywalking/blob/39d1869443a89025d4aacea1b1b825fc37ebb19f/apm-application-toolkit/apm-toolkit-opentracing/src/main/java/org/apache/skywalking/apm/toolkit/opentracing/SkywalkingSpan.java#L98
because the code is ``String.valueOf`` . the tagValue is String

PR #3803, opened 2019-11-07
Add elasticsearch 6.x plugin & elasticsearch 6.x test scenario

aderm:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___


wu-sheng:
We need to fix async http first. What is status of that ?

aderm:
Found that httpasyncclient has nothing to do with es6.x plugin, the test samples can pass normally, the only problem is that there may be problems in httpasyncclient.

wu-sheng:
As aysnc http and es plugin are both default plugin. You must make sure they could work together, that is the default option. So, if anything incompatible in there, should be fixed first. Such as having unexpected exit span or local span with meaningless operation name.

aderm:
When the external is exit span, the local span problem created and it has been fixed in the second commit.

docs/en/guides/Plugin-test.md
```diff
@@ -602,12 +602,13 @@ stage('Run Agent Plugin Tests') {
 Find the button 'detail' of your Workload in the PR page. Enter to the page and get the elapsed time of your task.
 
 ### Workload 1
-#### Group 1 (2164.287s)
+#### Group 1 (2635.00s)
```
dmsolr: placed in workload4-group1 is better

wu-sheng: Choose the fastest group/workload to add your task,  please.

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+
+    <artifactId>apm-elasticsearch-6.x-plugin</artifactId>
+    <packaging>jar</packaging>
+
+    <name>elasticsearch-6.x-plugin</name>
+    <url>http://maven.apache.org</url>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <java.version>1.8</java.version>
```
wu-sheng: Do you need 1.8 for plugin source codes?

docs/en/guides/Plugin-test.md
aderm: ok

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/pom.xml
aderm: Other versions have not been tried, however, ES6+ official has requested that you must follow the java8 version.

aderm:
> supported-list doc should be updated.

Regarding the supported version, the version before 6.7.0, some places are different, not compatible

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/pom.xml
wu-sheng: This is not about the ES 6 requirement, this is about your plugin codes. It is OK plugin complied in 1.6, and lib required 1.8+

wu-sheng:
I mean this doc, https://github.com/apache/skywalking/blob/master/docs/en/setup/service-agent/java-agent/Supported-list.md should be updated in this PR too.

aderm:
Ok, no problem

Jenkinsfile-Agent-Test
```diff
@@ -82,6 +83,12 @@ pipeline {
             parallel {
                 stage('Group1') {
                     stages {
+                        stage('elasticsearch-6.x-scenario 6.7.1-6.8.4 (7)') {
+                            steps {
+                                sh 'bash test/plugin/run.sh --build_id=wl1_${BUILD_ID} elasticsearch-6.x-scenario'
```
dmsolr: we don't need the argument, build_id.

apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
```diff
@@ -143,4 +143,6 @@
     public static final OfficialComponent EHCACHE = new OfficialComponent(75, "Ehcache");
 
     public static final OfficialComponent SOCKET_IO = new OfficialComponent(76, "SocketIO");
+
+    public static final OfficialComponent REST_HIGH_LEVEL_CLIENT = new OfficialComponent(77, "rest-high-level-client");
```
dmsolr: What is it? Why don't we use `TRANSPORT_CLIENT`?

aderm: This is not recommended in the new ES version.

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/pom.xml
aderm: OK, I will delete java.version.

Jenkinsfile-Agent-Test
aderm: ok

apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
wu-sheng: Yes. HTTP Client is recommended.

wu-sheng:
> 1. No comment for codes
1. No unit test cases for interceptors. You should mock parameter to check the interceptors.

Follow this two, others LGTM. @kezhenxu94 @dmsolr @arugal Please recheck.

aderm:
Ok, next will add comments and test the interceptors via mock.

wu-sheng:
When you update this PR as another plugin test has been added.

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/elasticsearch/v6/interceptor/IndicesClientCreateMethodsInterceptor.java
```diff
@@ -0,0 +1,79 @@
...
+            SpanLayer.asDB(span);
+        }
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments,
+        Class<?>[] argumentsTypes, Object ret) throws Throwable {
+        ContextManager.stopSpan();
+        return ret;
+    }
```
kezhenxu94: I notice some of the interceptors have this condition(`IndicesClientCreateMethodsInterceptor`, `IndicesClientDeleteMethodsInterceptor`), and the others don't (`RestHighLevelClientGetMethodsInterceptor`, `RestHighLevelClientIndexMethodsInterceptor`), are there differencies? if this condition is a must, then you may need it in `afterMethod`

test/plugin/scenarios/elasticsearch-6.x-scenario/bin/startup.sh
```diff
@@ -0,0 +1,23 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+home="$(cd "$(dirname $0)"; pwd)"
+
+java -jar ${agent_opts} -Dskywalking.plugin.elasticsearch.trace_dsl=true ${home}/../libs/elasticsearch-6.x-scenario.jar &
+
+# user for more time to collect data
+sleep 5
```
arugal: Hi, are you sure need to ```sleep 5``` ？

test/plugin/scenarios/elasticsearch-6.x-scenario/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.elasticsearch</groupId>
+            <artifactId>elasticsearch</artifactId>
+            <version>${test.framework.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>${lombok.version}</version>
```
arugal: There seems to be no place to use lombok?

aderm:
> When you update this PR as another plugin test has been added.

Probably tomorrow, there is an exam today.

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/elasticsearch/v6/RemotePeerCache.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * @author aderm
+ */
+public class RemotePeerCache {
+
+    private List<String> remotePeerList = Collections.synchronizedList(new ArrayList<String>());
```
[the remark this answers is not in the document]
aderm: Yes, I was thinking about future expansion of the function.

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/elasticsearch/v6/RemotePeerCache.java
```diff
@@ -0,0 +1,46 @@
...
+    private String remotePeers = "";
+
+    public void addRemotePeer(String host, int port) {
+        String hostPort = host + ":" + String.valueOf(port);
```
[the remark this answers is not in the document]
aderm: ACK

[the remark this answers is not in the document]
aderm: This may be different in writing habits, essentially the same.

test/plugin/scenarios/elasticsearch-6.x-scenario/bin/startup.sh
aderm: Yes, if there is no waiting, it will often lead to incomplete collection of spans.

test/plugin/scenarios/elasticsearch-6.x-scenario/pom.xml
aderm: ACK, will be deleted

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/elasticsearch/v6/interceptor/IndicesClientCreateMethodsInterceptor.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.skywalking.apm.plugin.elasticsearch.v6.ESTypeEnum;
+import org.apache.skywalking.apm.plugin.elasticsearch.v6.RestClientEnhanceInfo;
+import org.elasticsearch.client.indices.CreateIndexRequest;
+
+/**
+ * @author aderm
+ */
+public class IndicesClientCreateMethodsInterceptor implements InstanceMethodsAroundInterceptor {
+
+    private static final ILog logger = LogManager.getLogger(IndicesClientCreateMethodsInterceptor.class);
```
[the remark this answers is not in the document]
aderm: ACK, will be deleted

wu-sheng:
> Probably tomorrow, there is an exam today.

Take your time.

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/elasticsearch/v6/interceptor/IndicesClientCreateMethodsInterceptor.java
aderm: Yes, they are not the same. The front is mainly related to index operations. 

>  if this condition is a must, then you may need it in afterMethod
Not understanding the meaning

test/plugin/scenarios/elasticsearch-6.x-scenario/bin/startup.sh
arugal: You should use ```healthCheck``` to make sure that ES is ready, not through sleep.

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/elasticsearch/v6/interceptor/IndicesClientCreateMethodsInterceptor.java
kezhenxu94: > Yes, they are not the same. The front is mainly related to index operations.
> 
> > if this condition is a must, then you may need it in afterMethod
> 
> Not understanding the meaning

I was asking why some of them have condition `if (restClientEnhanceInfo != null) {` and other don't, if the condition is really needed, you'll need the same condition in `afterMethod`

test/plugin/scenarios/elasticsearch-6.x-scenario/bin/startup.sh
wu-sheng: Yes, please follow @arugal guidance. The health check service is designed for you to verify the es is ready. Sleep is not a safe option, as it is affected by CPU payload.

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/elasticsearch/v6/RemotePeerCache.java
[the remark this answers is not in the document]
kezhenxu94: > Yes, I was thinking about future expansion of the function.

add it when you really need it, unused codes simply confuse others

test/plugin/scenarios/elasticsearch-6.x-scenario/bin/startup.sh
arugal: We have fixed the problem that the agent unregistered to make the span incomplete, if ES is not ready, it should be a service error, please check.

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/elasticsearch/v6/interceptor/IndicesClientCreateMethodsInterceptor.java
aderm: ACK, in aftermethod, `if (restClientEnhanceInfo != null)`  will be added.

test/plugin/scenarios/elasticsearch-6.x-scenario/bin/startup.sh
aderm: Sorry, there is a problem with the above explanation. The healthCheck is there. In fact, the reason is not that the ES startup time takes a few seconds, but because six requests are executed at one time in the test service, six segments are generated, and now In the JVM container, the waiting time for collection is 5s. Due to insufficient waiting time, the local information will result in incomplete collection of information, and the subsequent segments are not collected. I can normally collect it by adding a sleep to the wait above. And this local test success.  I just tested it, I removed `sleep 5`, and the test passed.

aderm: Ok, now i will remove `sleep 5`

wu-sheng: @aderm Anything you need to do or wait, put those in your codes and add check logic. Don't use time to verify, it is unstable, we were facing this kind of issue multiple times.

aderm: ok

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/elasticsearch/v6/RemotePeerCache.java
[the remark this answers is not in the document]
aderm: Ack

Jenkinsfile-Agent-Test-4
```diff
@@ -68,7 +68,7 @@ pipeline {
             }
         }
 
-        stage('Test Cases Report (153)') {
+        stage('Test Cases Report (154)') {
```
dmsolr: it should not be 160?

docs/en/guides/Plugin-test.md
arugal: It seems to be inconsistent with jenkinsfile-agent-test-4

docs/en/guides/Plugin-test.md
```diff
@@ -612,12 +612,13 @@ stage('Run Agent Plugin Tests') {
...
 scenario name | versions | elapsed time (sec)
 ---|---|---
-apm-toolkit-trace | 1 | 84.69
-jetty 9.x | 63 | 1970.88
-netty-socketio 1.x | 4 | 108.70
+elasticsearch-6.x-scenario | 7 | 326.00
```
arugal: It seems to be inconsistent with jenkinsfile-agent-test-4

wu-sheng: Yes. @aderm Your test has been added to workload 4, rather than workload 1. Please update the document by following your recent test result.

> elasticsearch-6.x-scenario 6.7.1-6.8.4 (7) . - 4m 33s

wu-sheng:
This PR will be hold, 6.5.0 release is ready to go.

aderm:
Ok, this part is to be remembered, just update it after the test is finished, just right.

Jenkinsfile-Agent-Test-4
aderm: I have counted both sides, both are 154

docs/en/guides/Plugin-test.md
aderm: Ok, this part is to be remembered, just update it after the test is finished, just right.

wu-sheng:
Version is required to update to 6.6.0-SNAPSHOT.

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/elasticsearch/v6/interceptor/IndicesClientCreateMethodsInterceptor.java
```diff
@@ -0,0 +1,84 @@
...
+
+        RestClientEnhanceInfo restClientEnhanceInfo = (RestClientEnhanceInfo) (objInst
+            .getSkyWalkingDynamicField());
+        if (restClientEnhanceInfo != null) {
+            AbstractSpan span = ContextManager
+                .createExitSpan(Constants.CREATE_OPERATOR_NAME,
+                    restClientEnhanceInfo.getPeers());
+            span.setComponent(ComponentsDefine.REST_HIGH_LEVEL_CLIENT);
+
+            Tags.DB_TYPE.set(span, ESTypeEnum.CREATE.getType());
```
dmsolr: Usually, we set it as a type of DB, not operation. So it should be `ElasticSearch`.

wu-sheng: Good catch. Please follow the existing ES plugin did, otherwise, the topology will break.

aderm: ACK

wu-sheng:
@aderm I think you miss your test case change? CI fails.

aderm:
yes, unit test is very useful.

Issue #3771, opened 2019-11-02
[ISTIO-CI] Add OAP+Istio(mixer) to CI pipeline


hanahmily:
Now we build CI workflow for istio mixer and skywalking based on #3718 

## Requirement

Get all dependencies, including `kubectl`, `helm`, and `minikube` . (assigned to @hanahmily )

## Steps

the following steps are some of the critical points for the whole works. 

 1. build a OAP image locally, pick up commit id as the tag. 
 1. bootstrap minikube up, make it access local images. (assigned to @hanahmily )
 1. deploy istio, OAP, and ES. Enable skywalking adapter of the mixer. 
 1. deploy bookings demo. 
 1. access demo until we get some metrics from OAP. 
 1. if there are some errors.  print pods status, pods log and some necessary debug details to STD
 1. stop minikube and clear it. (assigned to @hanahmily )
 1. clear docker images.

@wu-sheng @innerpeacez @kezhenxu94 
Some issues should be confirmed before we start to do it. 
 1. ES will deploy several pods with the current Helm chart. That's not accepted in CI. So I suggest creating a new ES template which is only enabled in CI. 
 1. we don't deploy UI in CI. So we should write some scripts to fetch data through GraphQL.

wu-sheng:
> build a OAP image locally, pick up commit id as the tag.

Pick up jenkins job id should make more sense.

> ES will deploy several pods with the current Helm chart. That's not accepted in CI. So I suggest creating a new ES template which is only enabled in CI.

Agree, we should reduce the unnecessary resource cost.

> we don't deploy UI in CI. So we should write some scripts to fetch data through GraphQL.

Recent e2e has coding to verify by GraphQL, please read codes there.

innerpeacez:
Where is the minikube deployed?

wu-sheng:
> Where is the minikube deployed?

Should be installed by scripts locally on VM.

hanahmily:
Yes. I will fetch a appropriate minikube version, then set some system
settings to bootstrap it.

wu-sheng:
Any update about this? @innerpeacez Any block?

innerpeacez:
Thinking .  I have some issues.

1. How should the new ES templete  exist?
2.CI pipeline refering to Jenkins pipeline?

@hanahmily @wu-sheng

wu-sheng:
@innerpeacez @hanahmily Do we have any update about this?

wu-sheng:
[...]

PR #3836, opened 2019-11-11
Add minikube to CI

hanahmily:
Intend to solve some steps in #3771, including to start minikube and setup tunnel.


hanahmily:
@innerpeacez if this PR is merged, you could start your work of #3771

wu-sheng:
@hanahmily I just update the branch by syncing with the master branch, but it seems you updated the UI by mistake. Could you fix this?

innerpeacez:
> @innerpeacez if this PR is merged, you could start your work of #3771

got it

.github/workflows/istio-mixer-ci.yaml
```diff
@@ -0,0 +1,41 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: istio-mixer-ci
+
+on: [push]
```
wu-sheng: This seems not right. `pull request` should be list here. I think we don't need `push` CI, right?

hanahmily: make sense, but I pick up `push` just for testing. Since `istio-ci` branch is a dev one, I want to leave it alone until it's been merged into master.

PR #3839, opened 2019-11-12
Move to 6.6.0 SNAPSHOT

wu-sheng:
Waiting for 6.5.0 release, after that, the version moves to 6.6.0-SNAPSHOT


wu-sheng:
Tag and repo generated, https://repository.apache.org/content/repositories/orgapacheskywalking-1035/org/apache/skywalking/apache-skywalking-apm/6.5.0/.

This should be ready to merge.

Issue #3844, opened 2019-11-13
What is the order defining for Segment  at  the same level ?


bnubobby:
### Question
What is the order defining for segment  at the same level ?

**My display:**
![image](https://user-images.githubusercontent.com/23736568/68736338-53eac580-061b-11ea-8731-edeeb98941ce.png)

**actually , my calling order is :** 

/home/par 
-> /home/test210 
-> /home/test310
-> /home/test410
-> /home/test1
-> /home/test501

**Isn't it order by start time ?**

**My data** 
`
[...]

PR #3846, opened 2019-11-13
fix the order of segments at the same level

bnubobby:

for the #3844


wu-sheng:
Does this sort work?

wu-sheng:
We have multiple implementation of storage. If you order at database side, you need to change all. H2 and Mysql.

bnubobby:
@wu-sheng  
Does this mean that I need to make the same changes on H2 and MySQL?

wu-sheng:
> Does this mean that I need to make the same changes on H2 and MySQL?

Yes, if you have checked this change makes the UI right, can you confirm?

wu-sheng:
Why close?

PR #3849, opened 2019-11-13
Refactor DataCarrier, support ArrayBlockingQueueBuffer as implementor

wu-sheng:
Follow #3726, I am adding the new implementation for blocking mode, which is mostly used at the OAP server.  In this scenario, OAP is more sensitive about across threads visibility but has less concern about the blocking or lock issue. 

DataCarrier implementation queue still has better performance and less lock time, so we will keep it in the agent, unless we have more evidence that the ArrayBlockingQueueBuffer implementor is also better.


wu-sheng:
FYI @lkxiaolou @arugal

apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/buffer/ArrayBlockingQueueBuffer.java
```diff
@@ -0,0 +1,69 @@
...
+                return queue.offer(data);
+            default:
+                try {
+                    queue.put(data);
+                } catch (InterruptedException e) {
+                    // Ignore the error
+                    return false;
+                }
+        }
+        return true;
```
dmsolr: ArrayBlockingQueueBuffer is only used to `BLOCKING` strategy.

wu-sheng: Yes, for now. Because only backend is using blocking strategy. But I still implement if_possible strategy, maybe need this later. When we have time to run more tests about this on the agent

wu-sheng:
@kezhenxu94 Remove the uesless APIs.

PR #3852, opened 2019-11-14
UI Sync: multiple metrics compare and topology interaction

Fine0830:
One metric to one chart by comparative analysis


wu-sheng:
There are some files in UI don't including the ASF license header.

PR #3853, opened 2019-11-14
fix the order of segments at the same level

bnubobby:
- Bug fix
for  #3844


wu-sheng:
You need to follow the code format. There is a format file in the root path.

bnubobby:
OK, you're the boss

wu-sheng:
You just need to update the branch, don't need to reopen a new pull request every time. CI will check your PR. Only be able to merge after you passed all tests.

bnubobby:
I'm sorry about that,  I am new for PR

bnubobby:
@wu-sheng 
Please submit another pull request to fix.
When I just tested it, it led to another problem.

wu-sheng:
OK. Do you need help? Or you could fix by yourself?

Issue #3837, opened 2019-11-12
Support spring-tx in optional plugin 


zhaoyuguang:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [x] Feature or performance improvement

___
### Requirement or improvement
- During the peak of the transaction, it was found that spring-tx took a long time to get the transaction, but sw does not currently enhance this component, but the enhanced spring-tx will bring a lot of localSpan, as shown below: 
![image](https://user-images.githubusercontent.com/10150229/68658230-06fde500-0570-11ea-8257-49864754c735.png)
So I want to put it in `optional-plugins`.
Move spring-tx from the optional plugin to the plugin, and the monitoring can see how much time spring-tx spent in operating the transaction and whether the spring transaction config is correct in use.

wu-sheng:
Optional is OK if you want.

wu-sheng:
Provided

PR #3857, opened 2019-11-14
Support spring-tx in optional plugin

zhaoyuguang:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

___
https://github.com/apache/skywalking/issues/3837


Jenkinsfile-Agent-Test-2
```diff
@@ -82,6 +82,12 @@ pipeline {
             parallel {
                 stage('Group1') {
                     stages {
+                        stage('spring-tx 4.x+ (10)') {
```
wu-sheng: The total case number is not updated.

zhaoyuguang: I fixed, recheck please

test/plugin/scenarios/spring-tx-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,168 @@
...
+                           Mysql/JDBI/PreparedStatement/executeUpdate, /case/spring-tx-case, /case/healthCheck,
+                           Mysql/JDBI/Connection/close, Mysql/JDBI/Statement/executeQuery]
+  heartbeat: []
+segmentItems:
+  - applicationCode: spring-tx-scenario
+    segmentSize: nq 0
+    segments:
+      - segmentId: not null
+        spans:
+          - operationName: TX/get/test.org.apache.skywalking.apm.testcase.spring.transaction.service.impl.DemoServiceImpl.doBiz
```
dmsolr: I am not familiar with this.  I think `operationName` is a little long. Maybe we can use `t.o.s.a.t.s.t.s.i.DemoServiceImpl.doBiz`.

wu-sheng: If we want to support this, please be an optional config.

zhaoyuguang: Good proposal. I'll finish it.

arugal: Hi, store method name in tags? FYI

wu-sheng: Don't save duplicated inform. Otherwise, there is no point to implement short name. Use config if you want to give user options. But, you need make sure converting to short is quick enough.

arugal:
```yaml
operationName: TX/get
tags: 
 - {key: "method", value:"test.org.apache.skywalking.apm.testcase.spring.transaction.service.impl.DemoServiceImpl.doBiz"}
```
My idea should be like this,  can ignore.

apm-sniffer/optional-plugins/optional-spring-plugins/spring-tx-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/transaction/GetTransactionMethodInterceptor.java
```diff
@@ -0,0 +1,76 @@
...
+import java.lang.reflect.Method;
+
+/**
+ * @author zhaoyuguang
+ */
+public class GetTransactionMethodInterceptor implements InstanceMethodsAroundInterceptor {
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                             MethodInterceptResult result) throws Throwable {
+        TransactionDefinition definition = (TransactionDefinition) allArguments[0];
```
kezhenxu94: According to the [Spring source codes](https://github.com/spring-projects/spring-framework/blob/85471d05871ee8461b5f196ed26ffeb866d76cb2/spring-tx/src/main/java/org/springframework/transaction/support/AbstractPlatformTransactionManager.java#L341-L350), the `definition` is marked as `@Nullable`, and it may be null, so I'm wondering that have you covered this scenario? or should we consider this case? future users may encounter weird issues that are hard to diagnose when this happen

```java
	public final TransactionStatus getTransaction(@Nullable TransactionDefinition definition) throws TransactionException {
		Object transaction = doGetTransaction();


		// Cache debug flag to avoid repeated checks.
		boolean debugEnabled = logger.isDebugEnabled();


		if (definition == null) {
			// Use defaults if no transaction definition given.
			definition = new DefaultTransactionDefinition();
		}
```

wu-sheng: Make sense, @zhaoyuguang we should include this.

test/plugin/scenarios/spring-tx-scenario/config/expectedData.yaml
kezhenxu94: I just rechecked the specification and I am also wondering whether this operation name is **too specific**, as suggested in [the specification](https://github.com/opentracing/specification/blob/master/specification.md#start-a-new-span)

wu-sheng: We don't need to follow the OpenTracing, as it is not a specific at all. I would like to discuss what kind of operation name makes more sense to user when they see our trace view. Could we discuss based on this?

apm-sniffer/optional-plugins/optional-spring-plugins/spring-tx-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/transaction/GetTransactionMethodInterceptor.java
zhaoyuguang: Good proposal. I'll finish it. @kezhenxu94

zhaoyuguang: I  fixed  please recheck.

apm-sniffer/optional-plugins/optional-spring-plugins/spring-tx-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/transaction/GetTransactionMethodInterceptor.java
```diff
@@ -0,0 +1,81 @@
...
+        span.tag(Constants.TAG_SPRING_TRANSACTION_ISOLATION_LEVEL, String.valueOf(definition.getIsolationLevel()));
+        span.tag(Constants.TAG_SPRING_TRANSACTION_PROPAGATION_BEHAVIOR, String.valueOf(definition.getPropagationBehavior()));
+        span.tag(Constants.TAG_SPRING_TRANSACTION_TIMEOUT, String.valueOf(definition.getTimeout()));
+        span.setComponent(ComponentsDefine.SPRING_TX);
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                              Object ret) throws Throwable {
+        ContextManager.stopSpan();
```
kezhenxu94: Need the same if condition here?

PR #3870, opened 2019-11-16
[Feature] Support ElasticSearch 7 as backend storage

kezhenxu94:
Add storage-elasticsearch7-plugin module

## Motivation:

Support ES7 as storage

## Modification:

Add storage-elasticsearch7-plugin, make the compilation passed, add maven profiles to distinguish different version of ES

## Result:

ES7 storage plugin can co-exist with ES6 storage plugin

Most of the codes come from the existed ES6 plugin, and the credits should belong to @peng-yongsheng


wu-sheng:
Before all codes related discussion, a question, as this profile based solution, how should developers run the local test for both ES6 and ES7? I mean in the IDE.

kezhenxu94:
> Before all codes related discussion, a question, as this profile based solution, how should developers run the local test for both ES6 and ES7? I mean in the IDE.

Simply switch the profile in maven tab

image

and I'm not sure if it's possible to achieve the same goal without using profiles, we have to build two distribution package, and have two dependency hierarchies

wu-sheng:
> and I'm not sure if it's possible to achieve the same goal without using profiles, we have to build two distribution package, and have two dependency hierarchies

That is why I proposed two starter module. After you have that, you don't need profile anymore. 
From my understanding, our project repo has been too complex already, if need profile to package different dist, it is very hard for a new developer.

wu-sheng:
Another thing, if you have the `exclude` and `dependency` right in new starter and storage, you could use most of existing es6 storage codes, rather than copy them again. It increases the payload to maintain codes. You should just need to implement the ES7 DAO, if ES API changed.

kezhenxu94:
@wu-sheng  Review comments are addressed, please recheck when you have got some time

oap-server/server-starter-es7/src/main/java/org/apache/skywalking/oap/server/starter/OAPServerStartUp.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.skywalking.oap.server.telemetry.TelemetryModule;
+import org.apache.skywalking.oap.server.telemetry.api.MetricsCreator;
+import org.apache.skywalking.oap.server.telemetry.api.MetricsTag;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author peng-yongsheng
+ */
+public class OAPServerStartUp {
```
wu-sheng: Auther is not right. You could call the old start main func directly, rather than copy codes.

oap-server/server-starter-es7/src/main/resources/application.yml
```diff
@@ -0,0 +1,205 @@
...
+    recordDataTTL: ${SW_CORE_RECORD_DATA_TTL:90} # Unit is minute
+    minuteMetricsDataTTL: ${SW_CORE_MINUTE_METRIC_DATA_TTL:90} # Unit is minute
+    hourMetricsDataTTL: ${SW_CORE_HOUR_METRIC_DATA_TTL:36} # Unit is hour
+    dayMetricsDataTTL: ${SW_CORE_DAY_METRIC_DATA_TTL:45} # Unit is day
+    monthMetricsDataTTL: ${SW_CORE_MONTH_METRIC_DATA_TTL:18} # Unit is month
+    # Cache metric data for 1 minute to reduce database queries, and if the OAP cluster changes within that minute,
+    # the metrics may not be accurate within that minute.
+    enableDatabaseSession: ${SW_CORE_ENABLE_DATABASE_SESSION:true}
+storage:
+  elasticsearch7:
```
wu-sheng: If we don't separate this name, we should be able to use old start module config file, reduce duplicated config files. WDYT?

kezhenxu94: Also tried. Impossible. We include the old starter, reusing the name causes two providers exception

oap-server/server-starter-es7/src/main/java/org/apache/skywalking/oap/server/starter/OAPServerStartUp.java
kezhenxu94: I tried. But I want to reuse the full qualified class name, OAPServerStartUp, which causes infinite recursive calls

oap-server/server-starter-es7/src/main/resources/application.yml
wu-sheng: Because you haven't excluded the es6 storage module, which you should.

oap-server/server-starter-es7/src/main/java/org/apache/skywalking/oap/server/starter/OAPServerStartUp.java
wu-sheng: Because you want to reuse the script? If so, why don't you create a start module, then move the codes into there.

oap-server/server-storage-plugin/storage-elasticsearch7-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch7/query/AggregationQueryEsDAO.java
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.skywalking.oap.server.library.client.elasticsearch.ElasticSearchClient;
+import org.elasticsearch.search.aggregations.AggregationBuilders;
+import org.elasticsearch.search.aggregations.BucketOrder;
+import org.elasticsearch.search.aggregations.bucket.terms.TermsAggregationBuilder;
+
+/**
+ * @author peng-yongsheng
+ * @author kezhenxu94
+ */
+public class AggregationQueryEsDAO extends org.apache.skywalking.oap.server.storage.plugin.elasticsearch.query.AggregationQueryEsDAO {
```
wu-sheng: I suggest you refactor the name of all ES7 DAO(s), to use `AggregationQueryEs7DAO`, in order to avoid full package name.

kezhenxu94: There is no issue of full-qualified-class-name, those `DAO`s for ES7 locate in another different package, named `org.apache.skywalking.oap.server.storage.plugin.elasticsearch7`, that's not the same issue as the class `OAPServerStartUp`

wu-sheng: I know they are in different packages, fundamentally, there is no issue. I suggest this just because seems don't need to depend on the  full name technically, which is also not common in Java from my perspective.

kezhenxu94: make sense and done

wu-sheng:
@kezhenxu94 You need to sync and resolve the new YAML file. Facing conflicts.

dmsolr:
Hi @kezhenxu94 
I found `thread_pool.index.queue_size` is deprecated in es7.4.2. Do we need to update the FAQ of the document?

kezhenxu94:
> Hi @kezhenxu94
> I found `thread_pool.index.queue_size` is deprecated in es7.4.2. Do we need to update the FAQ of the document?

The [FAQ](https://github.com/apache/skywalking/blob/master/docs/en/FAQ/ES-Server-FAQ.md#elasticsearch) is for users who are facing the exceptions, if the `thread_pool.index.queue_size` is not set by default in ES7, which I thing is the case, the users won't reach the FAQ

wu-sheng:
What we recommend for es7? Is write queue size still there? Add these two in document directly?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -97,46 +102,46 @@ public StorageEsInstaller(ModuleManager moduleManager, int indexShardsNumber, in
...
-        setting.addProperty("index.number_of_shards", indexShardsNumber);
-        setting.addProperty("index.number_of_replicas", indexReplicasNumber);
-        setting.addProperty("index.refresh_interval", record ? TimeValue.timeValueSeconds(10).toString() : TimeValue.timeValueSeconds(indexRefreshInterval).toString());
-        setting.addProperty("analysis.analyzer.oap_analyzer.type", "stop");
+    protected Map<String, Object> createSetting(boolean record) {
+        Map<String, Object> setting = new HashMap<>();
+        setting.put("index.number_of_shards", indexShardsNumber);
+        setting.put("index.number_of_replicas", indexReplicasNumber);
+        setting.put("index.refresh_interval", record ? TimeValue.timeValueSeconds(10).toString() : TimeValue.timeValueSeconds(indexRefreshInterval).toString());
+        setting.put("analysis.analyzer.oap_analyzer.type", "stop");
```
dmsolr: setting.put("index.max_result_window", 1000000);

As mentioned in the FAQ document, we set this option default.

wu-sheng:
@dmsolr I think we also need to update the document, another PR for that?

dmsolr:
> @dmsolr I think we also need to update the document, another PR for that?

Yes, I will do that.

docs/en/setup/backend/backend-storage.md
```diff
@@ -92,9 +95,19 @@ storage:
...
+### Recommended ElasticSearch server-side configurations
+You could add following config to `elasticsearch.yml`, set the value based on your env.
+
+```yml
+# In tracing scenario, consider to set more than this at least.
+thread_pool.index.queue_size: 1000 # Only suitable for ElasticSearch 6
+thread_pool.write.queue_size: 1000 # Suitable for ElasticSearch 6 and 7
+
+# When you face query error at trace page, remember to check this.
+index.max_result_window: 1000000 # Only suitable for ElasticSearch 6. For ES 7, set `indexMaxResultWindow` under `storage`-`elasticsearch7` section in application.yml
```
wu-sheng: @dmsolr Is this only usable in 6.3? I remember it is not supported in ES 6.8 at least. @kezhenxu94 Could you recheck?

docs/en/setup/backend/backend-storage.md
```diff
@@ -26,10 +26,15 @@ storage:
     user: sa
 ```
 
-## ElasticSearch 6
-Active ElasticSearch 6 as storage, set storage provider to **elasticsearch**.
+## ElasticSearch
+Active ElasticSearch 6 as storage, set storage provider to **elasticsearch**, active ElasticSearch 7 as storage, set storage provider to **elasticsearch7**.
```
wu-sheng: These two should be point 1 and 2. Such as 

* Active ElasticSearch 6 as storage, set storage provider to **elasticsearch**
* Active ElasticSearch 7 as storage, set storage provider to **elasticsearch7**

In this review page, https://github.com/apache/skywalking/blob/dbe634b47e18b8b3b5c0f7ea98b62778afd705bb/docs/en/setup/backend/backend-storage.md#elasticsearch, it is not very clear.

docs/en/setup/backend/backend-storage.md
```diff
@@ -26,10 +26,15 @@ storage:
...
 
-**Required ElasticSearch 6.3.2 or higher, excepted 7.0.0 or higher. HTTP RestHighLevelClient is used to connect server.**
+**Required ElasticSearch 6.3.2 or higher. HTTP RestHighLevelClient is used to connect server.**
+
+For ElasticSearch 6.3.2 ~ 7.0.0 (excluded), please download the `apache-skywalking-bin.tar.gz` or `apache-skywalking-bin.zip`,
+for ElasticSearch 7.0.0 ~ 8.0.0 (excluded), please download the `apache-skywalking-bin-es7.tar.gz` or `apache-skywalking-bin-es7.zip`.
```
wu-sheng: Same here. I think you should add this following the `Active...` sentence.

docs/en/setup/backend/backend-storage.md
```diff
@@ -64,6 +65,16 @@ storage:
...
 ```
 
+and there're also some configurations that are ES7 specific, as follows:
+
+```yaml
+storage:
+  # Index max result window, for segment deep pagination, usually we don't recommend to scroll too many pages,
+  # instead, give more query criteria (e.g. service id or time range), to narrow the query results.
+  # see https://www.elastic.co/guide/en/elasticsearch/guide/current/pagination.html for more information
+  indexMaxResultWindow: ${SW_STORAGE_ES_INDEX_MAX_RESULT_WINDOW:5000}
```
wu-sheng: You miss one level here, which could cause misguide. And should add comments in his YAML fragment to say, using other parameters above.

wu-sheng:
Merging.

PR #3876, opened 2019-11-17
Update joda lib to fix UTC +13/+14 bug

wu-sheng:
In our demo env, the alert can't be triggered. With @ascrutae's invest, the joda time lib 2.9.9 has this bug, which it can't support UTC +13 and + 14. UTC + 13 is the TZ we used on demo env, which lets the global user could see the metrics, but trigger this bug.

After we tested the 2.10.5, seems the upstream fixed it and released. We are following to update. @ascrutae Please follow up this, and recheck again.


wu-sheng:
Haha, totally forgot.

PR #3887, opened 2019-11-18
[test/plugin] immigrate test hystrix scenario.

aderm:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
refer #3583 
___
### Bug fix
- Bug description.

- How to fix?

___


docs/en/guides/Plugin-test.md
```diff
@@ -647,9 +647,10 @@ elasticsearch 5.x | 3 | 142.15
 
 
 ### Workload 3
-#### Group 1 (2291.912s)
+#### Group 1 (3090.912s)
 scenario name | versions | elapsed time (sec)
 ---|---|---
+hystrix-scenario | 20 | 799.00
```
dmsolr: place it in Workload1 Group2. Because the group1 also takes almost 3000s.

test/plugin/scenarios/hystrix-scenario/src/main/java/test/apache/skywalking/apm/testcase/hystrix/Application.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package test.apache.skywalking.apm.testcase.hystrix;
```
dmsolr: We suggest the package name of plugin test case names `test.org.apache.skywalking.apm.testcase.*`.

test/plugin/scenarios/hystrix-scenario/src/main/java/test/apache/skywalking/apm/testcase/hystrix/controller/CaseController.java
```diff
@@ -0,0 +1,92 @@
...
+@Controller
+@RequestMapping("/case")
+public class CaseController {
+
+    private static final Logger logger = LogManager.getLogger(CaseController.class);
+
+    private static final String SUCCESS = "Success";
+
+    @PostConstruct
+    public void setUp() {
```
dmsolr: I suggest and recommend, it is triggered by health-check. And you need to make sure that it will be called one time.

test/plugin/scenarios/hystrix-scenario/src/main/java/test/apache/skywalking/apm/testcase/hystrix/controller/TestACommand.java
```diff
@@ -0,0 +1,61 @@
...
+                HystrixCommandProperties.Setter()
+                    .withExecutionTimeoutInMilliseconds(1000)
+            )
+        );
+        this.name = name;
+    }
+
+    @Override
+    protected String run() throws Exception {
+        Thread.sleep(2001);
```
dmsolr: Usually, I don't recommend to do that. If have to do, I think it doesn't over 1000ms. (Personal suggestion)

dmsolr: Because it is easy to cause by timeout. WDYT, @arugal

test/plugin/scenarios/hystrix-scenario/src/main/java/test/apache/skywalking/apm/testcase/hystrix/Application.java
aderm: OK

test/plugin/scenarios/hystrix-scenario/src/main/java/test/apache/skywalking/apm/testcase/hystrix/controller/TestACommand.java
aderm: Maybe understand a little bit different, this logic is to trigger the timeout fuse, now the fuse time is 1s

test/plugin/scenarios/hystrix-scenario/src/main/java/test/apache/skywalking/apm/testcase/hystrix/controller/CaseController.java
aderm: This estimate needs to be discussed. This is the code that needs to be executed before the @PostConstruct is executed, and the check is the verification logic that is executed after all the startups are normal.

docs/en/guides/Plugin-test.md
aderm: According to `Plugin-test.md`, workload3 group1 is little faster

aderm: Before I modified, According to `Plugin-test.md`, workload3 group1 is little faster

aderm: Before I modified, According to `Plugin-test.md`, workload3 group1 is little faster

aderm: Before I modified, According to Plugin-test.md, workload3 group1 is little faster than workload1 Group 2

dmsolr: I see. But a workload is waiting for both groups to finish, two groups of workload take the same time will be better.

aderm: if add to workload1 group2，total is：3090.98.  current workload3 group1， total is: 3090.912， is right？

test/plugin/scenarios/hystrix-scenario/src/main/java/test/apache/skywalking/apm/testcase/hystrix/controller/TestACommand.java
dmsolr: I mean that the request(`/hystrix-scenario`) timeout so that test failures. before Hystrix works. So that we will get the unexpected data.

docs/en/guides/Plugin-test.md
wu-sheng: @aderm According to document, workload1 group2 is only 2300s. Much faster than 3000. Where is your data from? 

image

Also verified from the CI process, https://builds.apache.org/blue/organizations/jenkins/skywalking-agent-test/detail/skywalking-agent-test/1297/pipeline/135

test/plugin/scenarios/hystrix-scenario/src/main/java/test/apache/skywalking/apm/testcase/hystrix/controller/TestACommand.java
aderm: This setting does not affect, although sleep 2s in run method, the front end rest request hystrix will returns fallback result when run over 1s.

wu-sheng: Then why sleep 2s?

dmsolr: For triggering fusing. @wu-sheng

wu-sheng: OK, if fusing in the case, then make sense.

aderm: This is because there is a hystrix blown test scene.

docs/en/guides/Plugin-test.md
wu-sheng: OK, make sense. I am good.

PR #3913, opened 2019-11-21
support custom elasticsearch settings in index level.

JaredTan95:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance
___

support custom elasticsearch settings in index level.
such as: 
- translog settings: https://www.elastic.co/guide/en/elasticsearch/reference/7.4/index-modules-translog.html

### DIFF
--- a/docker/oap/docker-entrypoint.sh
+++ b/docker/oap/docker-entrypoint.sh
@@ -119,7 +119,6 @@ cat <<EOT >> ${var_application_file}
     resultWindowMaxSize: \${SW_STORAGE_ES_QUERY_MAX_WINDOW_SIZE:10000}
     metadataQueryMaxSize: \${SW_STORAGE_ES_QUERY_MAX_SIZE:5000}
     segmentQueryMaxSize: \${SW_STORAGE_ES_QUERY_SEGMENT_SIZE:200}
-    topNReportPeriod: ${SW_CORE_TOPN_REPORT_PERIOD:10}
 EOT
 }
 
@@ -306,6 +305,7 @@ core:
     # Cache metric data for 1 minute to reduce database queries, and if the OAP cluster changes within that minute,
     # the metrics may not be accurate within that minute.
     enableDatabaseSession: \${SW_CORE_ENABLE_DATABASE_SESSION:true}
+    topNReportPeriod: \${SW_CORE_TOPN_REPORT_PERIOD:10}
 EOT
 
     # generate storage
--- a/docker/oap/docker-entrypoint.sh
+++ b/docker/oap/docker-entrypoint.sh
@@ -119,6 +119,7 @@ cat <<EOT >> ${var_application_file}
     resultWindowMaxSize: \${SW_STORAGE_ES_QUERY_MAX_WINDOW_SIZE:10000}
     metadataQueryMaxSize: \${SW_STORAGE_ES_QUERY_MAX_SIZE:5000}
     segmentQueryMaxSize: \${SW_STORAGE_ES_QUERY_SEGMENT_SIZE:200}
+    topNReportPeriod: ${SW_CORE_TOPN_REPORT_PERIOD:10}
 EOT
 }
 
--- a/docs/en/setup/backend/backend-storage.md
+++ b/docs/en/setup/backend/backend-storage.md
@@ -63,7 +63,7 @@ storage:
     resultWindowMaxSize: ${SW_STORAGE_ES_QUERY_MAX_WINDOW_SIZE:10000}
     metadataQueryMaxSize: ${SW_STORAGE_ES_QUERY_MAX_SIZE:5000}
     segmentQueryMaxSize: ${SW_STORAGE_ES_QUERY_SEGMENT_SIZE:200}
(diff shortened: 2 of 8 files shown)


oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -119,6 +122,9 @@ storage:
     # instead, give more query criteria (e.g. service id or time range), to narrow the query results.
     # see https://www.elastic.co/guide/en/elasticsearch/guide/current/pagination.html for more information
     indexMaxResultWindow: ${SW_STORAGE_ES_INDEX_MAX_RESULT_WINDOW:5000}
+#    advanced:
+#      index.translog.durability: "async"
+#      index.translog.sync_interval: "120s"
```
wu-sheng: According to the official document, `index.translog.sync_interval` default value is 5s,  `index.translog.durability` is `request`. But you set the different, are these two recommended by you? If so, it should be active configs in default. 

And you miss to update the documentation of storage.

wu-sheng: The default values are referring from https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules-translog.html#_translog_settings

docker/oap/docker-entrypoint.sh
```diff
@@ -119,6 +119,7 @@ cat <<EOT >> ${var_application_file}
     resultWindowMaxSize: \${SW_STORAGE_ES_QUERY_MAX_WINDOW_SIZE:10000}
     metadataQueryMaxSize: \${SW_STORAGE_ES_QUERY_MAX_SIZE:5000}
     segmentQueryMaxSize: \${SW_STORAGE_ES_QUERY_SEGMENT_SIZE:200}
+    topNReportPeriod: ${SW_CORE_TOPN_REPORT_PERIOD:10}
```
kezhenxu94:
```suggestion
    topNReportPeriod: \${SW_CORE_TOPN_REPORT_PERIOD:10}
```

`SW_CORE_`?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -43,14 +42,17 @@
     protected final int indexShardsNumber;
     protected final int indexReplicasNumber;
     protected final int indexRefreshInterval;
+    protected final Map<String,Object> advanced;
     protected final ColumnTypeEsMapping columnTypeEsMapping;
 
-    public StorageEsInstaller(ModuleManager moduleManager, int indexShardsNumber, int indexReplicasNumber, int indexRefreshInterval) {
+    public StorageEsInstaller(ModuleManager moduleManager, int indexShardsNumber, int indexReplicasNumber, int indexRefreshInterval,
```
kezhenxu94: The constructor arguments are becoming too long, consider passing the `config` directly and get the config items inside the constructor, like what the ES7 version does

oap-server/server-bootstrap/src/main/resources/application.yml
kezhenxu94: > According to the official document, `index.translog.sync_interval` default value is 5s, `index.translog.durability` is `request`. But you set the different, are these two recommended by you? If so, it should be active configs in default.
> 
> And you miss to update the documentation of storage.

same question here

docs/en/setup/backend/backend-storage.md
```diff
@@ -103,12 +104,24 @@ storage:
...
+    advanced: ${SW_STORAGE_ES_ADVANCED:""}
 ```
 
 
 ### Data TTL
 TTL in ElasticSearch overrides the settings of core, read [ElasticSearch section in TTL document](ttl.md#elasticsearch-6-storage-ttl)
 
+### Advanced Configurations For Elasticsearch Index
+You can add advanced configurations in `JSON` format to set `elasticsearch index settings`.
+For example, set [translog](https://www.elastic.co/guide/en/elasticsearch/reference/master/index-modules-translog.html) settings:
```
wu-sheng: I think this doc ref will misguide people. From my understanding, all parameters in [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html) should be supported, right?

JaredTan95: Yes.

JaredTan95: Need to add this ref https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html ?

wu-sheng: Then you should ref to the config page in the document, your example could be same. Such as
___
You can add advanced configurations in `JSON` format to set `ElasticSearch index settings` by following [ElasticSearch doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html)

JaredTan95: Thanks, I will add it.

PR #3919, opened 2019-11-22
Extending OAP docker container

hanahmily:
I introduce two new directories in OAP docker container to extend its capabilities.

 * /skywalking/ext-libs: put addition jars in it, OAP will load them into its CLASSPATH.
 * /skywalking/ext-config: put configuration files in it, then they will override files stored in /skywalking/config. If /skywalking/ext-config/appliation.yml exists, `docker-entrypoint.sh` wouldn't generate it anymore.


wu-sheng:
Where do we have the doc about how to use Docker in dev mode? Beside [docker build](https://github.com/apache/skywalking/blob/master/docs/en/guides/How-to-build.md#build-docker-images). [This doc](https://github.com/apache/skywalking/tree/master/docker) seems missing.

hanahmily:
> Where do we have the doc about how to use Docker in dev mode? Beside [docker build](https://github.com/apache/skywalking/blob/master/docs/en/guides/How-to-build.md#build-docker-images). [This doc](https://github.com/apache/skywalking/tree/master/docker) seems missing.

Exactly. We should add a new topic to document all of the things about Docker in the main repo, that might belong to another PR.

wu-sheng:
> Exactly. We should add a new topic to document all of the things about Docker in the main repo, that might belong to another PR.

Could you submit an issue about this? I think this should be fixed in 6.6.0 release.

hanahmily:
> Could you submit an issue about this? I think this should be fixed in 6.6.0 release.

done.  #3920

Issue #3942, opened 2019-11-27
Documents  Error In v6.5 ?


seifeHu:
In 
https://github.com/apache/skywalking/blob/v6.5.0/docs/en/setup/service-agent/java-agent/README.md

logging.dir | Log files directory. Default is blank string, means, use "system.out" to output logs. | ""
-- | -- | --

I fact ,the log file dir use    {jarfile.dir}/../logs to create a skywalking-api.log

wu-sheng:
Yes, I think this an error of document, and `org.apache.skywalking.apm.agent.core.conf.Config#Logging#DIR`. Would you send a pull request to fix this?

seifeHu:
> Yes, I think this an error of document, and `org.apache.skywalking.apm.agent.core.conf.Config#Logging#DIR`. Would you send a pull request to fix this?

ok.

PR #3943, opened 2019-11-27
Document error for agent config file 's config description

seifeHu:
### Bug fix
- Bug description.
https://github.com/apache/skywalking/issues/3942

In
https://github.com/apache/skywalking/blob/v6.5.0/docs/en/setup/service-agent/java-agent/README.md

logging.dir | Log files directory. Default is blank string, means, use "system.out" to output logs. | ""
-- | -- | --

### DIFF
--- a/docs/en/setup/service-agent/java-agent/README.md
+++ b/docs/en/setup/service-agent/java-agent/README.md
@@ -93,7 +93,7 @@ property key | Description | Default |
 `logging.level`|The log level. Default is debug.|`DEBUG`|
 `logging.file_name`|Log file name.|`skywalking-api.log`|
 `logging.output`| Log output. Default is FILE. Use CONSOLE means output to stdout. |`FILE`|
-`logging.dir`|Log files directory. Default is blank string, means, use "system.out" to output logs.|`""`|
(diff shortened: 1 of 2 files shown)


docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -93,7 +93,7 @@ property key | Description | Default |
 `logging.level`|The log level. Default is debug.|`DEBUG`|
 `logging.file_name`|Log file name.|`skywalking-api.log`|
 `logging.output`| Log output. Default is FILE. Use CONSOLE means output to stdout. |`FILE`|
-`logging.dir`|Log files directory. Default is blank string, means, use "system.out" to output logs.|`""`|
+`logging.dir`|Log files directory. Default is blank string, means, use "{theSkywalkingAgentJarDir}/logs  " to output logs. {theSkywalkingAgentJarDir} is the directory where the skywalking agent jar file is located |`""`|
```
seifeHu: the description is an Error .In fact, the log will be output to the  "agentJarDir/logs "

wu-sheng:
The class comment is wrong too, org.apache.skywalking.apm.agent.core.conf.Config#Logging#DIR. Could you fix this too?

seifeHu:
> The class comment is wrong too, org.apache.skywalking.apm.agent.core.conf.Config#Logging#DIR. Could you fix this too?

ok . I pull an other one pull request

wu-sheng:
This one is good, you don't need a new one.

Issue #3891, opened 2019-11-19
spring cloud gateway TraceContext.traceId() Is not equal to micro service tid


dodomao:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [Agent ] Bug

___
### Question
- What do you want to know?
Spring Cloud Gateway TraceContext.traceId() Is not equal to micro service logback tid
I think that is the cause of the TraceCrossThread
Before 6.5.0, the traceid Repeated Generation in spring cloud gateway。Now,the bug is solve，but the traceid is not a GlobalTraceId
___
### Bug
- Which version of SkyWalking, OS and JRE?
SkyWalking 6.5  jdk1.8

- What happen?
spring cloud gateway GlobalFilter:
LOGGER.info(TraceContext.traceId());
[ctor-http-nio-3] c.d.cbhis.his_gateway.filter.GrayFilter  : 3729.128.15741514902390093
Micro service logback:
[TID:3729.128.15741514391310083] 
___
### Requirement or improvement
- How to get the GlobalTraceId in spring cloud gateway GlobalFilter or GatewayFilter.

wu-sheng:
Maybe you could provide your codes? @dodomao 

@zhaoyuguang Please take a look, look like Gateway plugin is still no stable.

dodomao:
Spring Cloud Gateway:
  Spring Boot:'org.springframework.boot' version '2.1.4.RELEASE'
  Spring Cloud:"Greenwich.SR2"
  build.gradle:
      compile 'org.apache.skywalking:apm-toolkit-logback-1.x:6.5.0'
      compile 'org.apache.skywalking:apm-toolkit-trace:6.5.0'
  GlobalFilter:
`@Component
public class LoggerFilter implements GlobalFilter, Ordered {

  
    private static final Logger LOGGER = LoggerFactory.getLogger(LoggerFilter.class);

    @Override
    public Mono<Void> filter(ServerWebExchange exchange, GatewayFilterChain chain) {

        ServerHttpRequest request = exchange.getRequest();
        URI uri = request.getURI();
        HttpHeaders httpHeaders = request.getHeaders();
        String zhiGongXm = "";
        
        if (LOGGER.isDebugEnabled()) {
                StringBuilder stringBuilder = new StringBuilder();
                stringBuilder.append("traceID:'").append(TraceContext.traceId()).append("'");
                LOGGER.debug(stringBuilder.toString());
        }


        String finalZhiGongXm = zhiGongXm;
[...]

Issue #2711, opened 2019-05-18
Provide plugin for spring-cloud-gateway 2.0.x


zhaoyuguang:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [x] Requirement
- [ ] Feature or performance improvement
___
### Requirement or improvement
Provide plugin for spring-cloud-gateway 2.0.x, Which use `reactor-netty` API is different from 2.1.x

ygsama:
I want to know if  SkyWalking-6.1 don't support spring-cloud-gateway 2.1.1 ? 
If it is , I need to wait a few days or do it myself ?

wu-sheng:
The Gateway will be merged today, https://github.com/apache/skywalking/pull/2517

SataQiu:
@wu-sheng @zhaoyuguang 
I found that our spring cloud gateway 2.1.x plugin is named `gateway-2.x-plugin`. 
Is there any ambiguity here? 
Should we use `2.1.x` instead of `2.x` ?
If yes, I would like to do this.

wu-sheng:
I am waiting @zhaoyuguang to make the decision. How does he want to put the new plugin of 2.0.x

zhaoyuguang:
I will try to finish it in next milestone, if you have interests, contributions are welcome @SataQiu. Because I think `2.0.x`  can be finished  in 2.x, so I named it.

wu-sheng:
> Because I think 2.0.x can be finished in 2.x, so I named it.

Make sense. Rename it if necessary, such as 2.0.x is very incompatible with 2.1.x, like using a totally different way to intercept.

zhaoyuguang:
> > Because I think 2.0.x can be finished in 2.x, so I named it.
> 
> Make sense. Rename it if necessary, such as 2.0.x is very incompatible with 2.1.x, like using a totally different way to intercept.

OK

wu-sheng:
@zhaoyuguang If you are trying to do this in 6.3.0, please set the milestone.

PR #3945, opened 2019-11-27
Enhance webflux plugin, related to Spring Gateway plugin too.

### LINKED ISSUE #3805: The webflux-plugin is not working properly

- [ ] Question or discussion
- [X] Bug
- [ ] Requirement

___
### Question
- What do you want to know?
The webflux-plugin is not working properly
because WebFlux is Reactive Programming  ，plugin use ThreadLocal is not working properly

webflux-plugin 并不能正常工作,因为 WebFlux 是响应式编程。 处理的线程会发生变化，插件内部使用ThreadLocal 根本不能获取到原来线程的信息，还会因为未能正确关闭span信息 导致ThreadLocal一直没有释放
___
### Bug

- What happen?
this is debug msg：
image
image
image
webflux-plugins Intercept 3 method ，the thread has changed when the method is executed
___
- Need to deal with it without relying on ThreadLocal

cyejing:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

Resolves #3805 #3985 and #2711


cyejing:
I went to see ``webflux-scenario `` test case.
The operationName it collects is different from what I expected, I think it should be collected like this ``method:path`` ``GET:/testcase/annotation/success``  not ``org.apache.skywalking.apm.testcase.sc.webflux.projectB.controller.TestAnnotationController.success``  . 
Because this is what the web interface should focus on is url and method
Can I add a new scenario case？ For this optional plugin

wu-sheng:
> The operationName it collects is different from what I expected, I think it should be collected like this method:path GET:/testcase/annotation/success not org.apache.skywalking.apm.testcase.sc.webflux.projectB.controller.TestAnnotationController.success .

Class + method name is collected by SpringMVC plugin in most cases. Because of `@RestController`. What do you mean about this?

cyejing:
I mean the newly added plugin collection operationName is ``method: url `` and ``webflux-scenario`` is not the same. 
Together they affect each other
Can I add a ``webflux-reactive-scenario`` case？

wu-sheng:
> Can I add a webflux-reactive-scenario case？

Yes, for new plugin, you should add a new one.

wu-sheng:
@cyejing The CI env shows strange error, so please check this locally, after https://github.com/apache/skywalking/pull/3952 done, this should be tested on GitHub Action directly.

zhaoyuguang:
It looks like this webflux plugin can replace the previous webflux plugin, right? You can delete the previous webflux plugin in this pr. Please note that the gateway may be affected, and operationName you can change, in the previous testcase.

wu-sheng:
@cyejing  Jenkins jobs have been disabled, please add your tasks to GitHub action file, and fix the conflict, remove the Jenkins file. https://github.com/apache/skywalking/blob/master/.github/workflows/plugins-test.yaml

cyejing:
> It looks like this webflux plugin can replace the previous webflux plugin
@zhaoyuguang  almost, But it need to use jdk8, I don't know which module should be suitable, Is currently placed in optional module.``spring cloud gateway`` based using webflux, So the gateway will also work normally, I added the test case of the gateway
@zhaoyuguang Can give me some suggestions for two webflux and gateway, it should be more appropriate to put them together。
@wu-sheng ok，Let me merge the code and handle the conflict

zhaoyuguang:
> it should be more appropriate to put them together。

Does the ` webflux plugin` & `gateway plugin` together in one `optional plugin `? Sound not good to me. First of all, from the perspective of `jar dependency`, micro-services will be divided into two types: `webflux application server` and `routing gateway application server`. Second, the webflux plugin should focus on responding to external requests, while the gateway should focus on routing that is, including external forwarding

cyejing:
@zhaoyuguang  You're right. ``spring cloud gateway`` is divided into two types. receive request and routing request.  receive request is based on webflux .  routing request have two components are optional. 
https://github.com/spring-cloud/spring-cloud-gateway/blob/master/spring-cloud-gateway-core/src/main/java/org/springframework/cloud/gateway/filter/NettyRoutingFilter.java
https://github.com/spring-cloud/spring-cloud-gateway/blob/master/spring-cloud-gateway-core/src/main/java/org/springframework/cloud/gateway/filter/WebClientHttpRoutingFilter.java
Actually only ``reactor.netty.http.client.HttpClient`` is work. ``WebClient`` config is commented
So if you divide by types, Is it possible to add a ``reactor.netty.http.client.HttpClient`` plugin.
What do you think

liangxp:
@wu-sheng @cyejing How long will it take for this commits to be merged into the master branch?This modification looks good

zhaoyuguang:
> Actually only `reactor.netty.http.client.HttpClient` is work. `WebClient` config is commented
> So if you divide by types, Is it possible to add a `reactor.netty.http.client.HttpClient` plugin.
> What do you think

Currently, `reactor.netty.http.client.HttpClient plugin in SkyWalking gateway plugin` is no generality, if can take out the component that was too good

xiaoy00:
> Currently, `reactor.netty.http.client.HttpClient plugin in SkyWalking gateway plugin` is no generality, if can take out the component that was too good
agree, and  `SkyWalking gateway plugin`  will create `gateway own span`  if  no need  routing request and stop at `gateway filter`。  so  take out the component is better

cyejing:
I submit a new pr to handle the gateway, and take out ``HttpClient`` component

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-reactive-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/reactive/DispatcherHandlerHandleMethodInterceptor.java
```diff
@@ -16,62 +16,90 @@
...
+        if (carrier != null && carrier.isValid()) {
+            context.extract(carrier);
+        }
+
+        span.setComponent(ComponentsDefine.SPRING_WEBFLUX);
+        SpanLayer.asHttp(span);
+        Tags.URL.set(span, exchange.getRequest().getURI().toString());
+        HTTP.METHOD.set(span,methodStr);
+
+        instance.setSkyWalkingDynamicField(context);
```
xiaoy00: Why set context , does it useful?   
Currently `gateway plugin` rely on value `span` of `SkyWalkingDynamicField` . 
if it's useful ,maybe need to update `gateway plugin`

wu-sheng:
Cases fail, please recheck.

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-reactive-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/reactive/DispatcherHandlerHandleMethodInterceptor.java
cyejing: because ``Context`` cannot be passed through ``ThreadLocal``. Downstream needs to get ``Context`` from ``SkyWalkingDynamic`` .   Reactive programming, the thread of processing is changing

xiaoy00: > because `Context` cannot be passed through `ThreadLocal`. Downstream needs to get `Context` from `SkyWalkingDynamic` . Reactive programming, the thread of processing is changing
1. yes, it cannot pass, But what do Downstream use to 'Context'.  Maybe  `ContextSnapshot`  is more useful when cross thread 
2. if u do. u should modify `FilteringWebHandlerInterceptor.java`，because it need `span` ,not `Context` Currently。

cyejing: I see it. gateway relies on webflux. Let me try

cyejing:
1. Because jdk8 support is required, the general use of webflux is jdk
2. It may be because of moving the directory. Most of the content is the same.Modified operationName

wu-sheng:
Are your plugin codes using JDK8 class? Could give me an example? The target codes are not related. If Your plugin could be compiled in 1.6, keep it in that way, rather than optional plugin.

cyejing:
> Are your plugin codes using JDK8 class? Could give me an example? The target codes are not related. If Your plugin could be compiled in 1.6, keep it in that way, rather than optional plugin.

for lambda. reactor method parameters are lambda interface. Example ``Mono `` ``Flux ``
Plugin code use:
```suggestion
        return ((Mono) ret).doFinally(s -> {
```

test/plugin/scenarios/webflux-scenario/support-version.list
```diff
@@ -14,21 +14,10 @@
...
-2.0.2.RELEASE
-2.0.3.RELEASE
-2.0.4.RELEASE
-2.0.5.RELEASE
-2.0.6.RELEASE
-2.0.7.RELEASE
-2.0.8.RELEASE
-2.0.9.RELEASE
-2.1.0.RELEASE
 2.1.1.RELEASE
```
wu-sheng: Why remove the versions?

test/plugin/scenarios/webflux-scenario/support-version.list
```diff
@@ -2,7 +2,7 @@
 # or more contributor license agreements.  See the NOTICE file
 # distributed with this work for additional information
 # regarding copyright ownership.  The ASF licenses this file
-# to you under the Apache License, Version 2.0 (the
+# to you under the Apache License, Version
```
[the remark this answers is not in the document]
cyejing: I modified it back

test/plugin/scenarios/webflux-scenario/support-version.list
cyejing: because These versions response status sometimes is null.  Test case requires return status code。I think it should be marked as not supported, the support is not very effective

https://github.com/cyejing/skywalking/blob/28fe378fbb5d8781a27a51b42ee77754713deb97/apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-reactive-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/reactive/DispatcherHandlerHandleMethodInterceptor.java#L93-L100

wu-sheng: Do you mean response in HTTP level? Not just codes level? I think the cases are stable before you change. So in http network level, it should be stable. Your plugin code just fix the core bug. 

I can't tell the relation between your reference codes and remove supposed version.

cyejing: This code is to get httpstatus.  webflux-2.0.x sometimes get httpstatus is null , 2.1.x all right.  in 2.0.x  really want to get httpstatus, only get netty http status through reflection, which is too complicated. 
Can anyone help？

wu-sheng: Is this the official confirmed bug? If yes. Then it is not related to our supported version. If it should be different ways, then you should use witness class to use difference APIs for status code.

Which case are you talking about?

cyejing: you are right, Let me check if there is a problem with the source code

wu-sheng: Let me if you need helps to make decisions. Mostly, we should not reduce the supported versions, our users will be surprised by that.

cyejing: I have found spring-framework issues that fix bugs.
issues: https://github.com/spring-projects/spring-framework/issues/21901
commit: https://github.com/spring-projects/spring-framework/commit/75b1396768079f4044db0fef70dd5f889c1c8931#diff-2beb7fc7f94343be7d8553e833e16113
method: https://github.com/spring-projects/spring-framework/blob/12cb0dfe975e89775e936e06acdfda597a9cab28/spring-web/src/main/java/org/springframework/http/server/reactive/ReactorServerHttpResponse.java#L66-L73

fix in spring-web-v5.1.3.RELEASE  , spring-boot-starter-webflux-v2.1.1.RELEASE

wu-sheng: Then, we don’t care it, right? We could support status in 2.1.1+ and ignore it in 2.0-2.1.0, right? And the test versions should keep unchanged. Any block for this?

cyejing: Yes, 2.0-2.1.0 Can ignore this state. Then how to write unit tests better, because unit tests cannot be written for different versions. Unless unit tests don't check this status and tags

wu-sheng: Then, we could remove these versions, but add comments in this version file, describing 2.0-2.1.0 are supported, but due to the status code bug(including the links), we don’t test them. Do you agree this is better?

cyejing: agree, Let me add comments

test/plugin/scenarios/gateway-scenario/config/expectedData.yaml
```diff
@@ -52,30 +52,26 @@ segmentItems:
...
             spanLayer: Http
             startTime: nq 0
             endTime: nq 0
-            componentId: 61
+            componentId: 67
             componentName: ''
             isError: false
-            spanType: Exit
-            peer: not null
+            spanType: Entry
```
wu-sheng: Where is the exit span?

cyejing: There is something wrong with this place, I will check again

cyejing:
> I noticed, you deleted several ref(s) in the test cases, could you tell me why? Ref is an important point to verify for a plugin like webflux. And as I commentted, no exit span. This is why we used to have two applications, we need to verify exit span(including peer) and entry span with correct ref.

I keep the previous test cases and enhanced

test/plugin/scenarios/webflux-scenario/bin/startup.sh
```diff
@@ -18,7 +18,4 @@
 
 home="$(cd "$(dirname $0)"; pwd)"
 
-java -jar ${agent_opts} "-Dskywalking.agent.service_name=webflux-projectA-scenario" ${home}/../libs/webflux-projectA-scenario.jar &
-sleep 1
-
-java -jar ${agent_opts} "-Dskywalking.agent.service_name=webflux-projectB-scenario" ${home}/../libs/webflux-projectB-scenario.jar &
+java -jar ${agent_opts} ${home}/../libs/webflux-scenario.jar &
```
wu-sheng: You just move two package into one, right?

test/plugin/scenarios/webflux-scenario/config/expectedData.yaml
```diff
@@ -14,30 +14,39 @@
...
   heartbeat: []
 segmentItems:
-  - applicationCode: webflux-projectB-scenario
+  - applicationCode: webflux-scenario
     segmentSize: nq 0
     segments:
       - segmentId: not null
         spans:
-          - operationName: org.apache.skywalking.apm.testcase.sc.webflux.projectB.controller.TestAnnotationController.success
+          - operationName: HEAD:/healthCheck
```
wu-sheng: The healthCheck should not add to the expected data, I know it was there, but it actually should not. Because, we can’t guarantee the agent will collect the health check request, actually. The key of healthCheck is making sure the container is ready to test. Make sense?

test/plugin/scenarios/webflux-scenario/bin/startup.sh
cyejing: Yes, make request test in one

test/plugin/scenarios/webflux-scenario/config/expectedData.yaml
cyejing: Ok, I will delete healthCheck test case

zhaoyuguang:
> 2. It may be because of moving the directory. Most of the content is the same.Modified operationName

Hi, @cyejing Compared with the old testcase, the current testcase lacks at least validation `Filter` and `/testcase/route/{test}`

zhaoyuguang:
In testcase the `operationName: GET:/provider/b/testcase`, 
I think `GET:` is not necessary，
    one reason: tag contain it
    two reason: can make `EndpointNameDictionary` more.
@wu-sheng  @arugal @dmsolr @kezhenxu94 @xiaoy00 what is your opinion.

wu-sheng:
> In testcase the operationName: GET:/provider/b/testcase,
I think GET: is not necessary，

I support to keep `GET` in the OP name prefix. Because in rest scenario, it could be `GET: abc/123` and `PUT: abc/123` for different meanings. It is better to keep them separated in endpoint metrics analysis.

wu-sheng:
> the current testcase lacks at least validation Filter and /testcase/route/{test}

@cyejing Please cover these two important cases.

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-reactive-plugin/pom.xml
```diff
@@ -6,27 +7,28 @@
...
-        <artifactId>spring-plugins</artifactId>
+        <artifactId>optional-spring-plugins</artifactId>
         <groupId>org.apache.skywalking</groupId>
         <version>6.6.0-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
-    <artifactId>apm-webflux-5.x-plugin</artifactId>
-    <packaging>jar</packaging>
+    <artifactId>apm-spring-webflux-5.x-reactive-plugin</artifactId>
```
zhaoyuguang: The plugin named `spring-webflux-5.x-plugin` is enough？ 
May missing tag <packaging> and <name>

wu-sheng: If it has replaced the existing plugin, then it should be named as `apm-spring-webflux-5.x-plugin`

cyejing:
> > 1. It may be because of moving the directory. Most of the content is the same.Modified operationName
> 
> Hi, @cyejing Compared with the old testcase, the current testcase lacks at least validation `Filter` and `/testcase/route/{test}`

Do you mean these urls ``/testcase/route/bad`` and ``/testcase/route/hello`` ？
have these test case.  because move two package into one.  some ``Exit`` Span was removed
``Filter`` I will add

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-reactive-plugin/pom.xml
cyejing: ok I modify to ``apm-spring-webflux-5.x-plugin``

wu-sheng:
`/testcase/route/{test}` means, you could detect the parameter. `{test}` could represent a variable.

cyejing:
> `/testcase/route/{test}` means, you could detect the parameter. `{test}` could represent a variable.

I understand, I will add

wu-sheng:
These two documents are required to update too
1. https://github.com/apache/skywalking/blob/master/docs/en/setup/service-agent/java-agent/Supported-list.md. The WebFlux 5.x should be tagged as option 2, right?
1. WebFlux and WebFlux Reactive should be list in https://github.com/apache/skywalking/blob/master/docs/en/setup/service-agent/java-agent/README.md#optional-plugins . And describe why it is optional.

wu-sheng:
@zhaoyuguang Could you recheck the PR about the points you mentioned?

wu-sheng:
Cloud gateway tests fail. Could you recheck?

cyejing:
> Cloud gateway tests fail. Could you recheck?

 Ha, this problem is as you said sometimes ``healthCheck`` is not exist.
I will remove it

wu-sheng:
Does this PR fix #3891, #3985 and #2711? Or some of them?

cyejing:
> Does this PR fix #3891, #3985 and #2711? Or some of them?

#3985 and #2711 is fix, and the docs should state that the gateway plugin depends on the webflux plugin

#3891 I haven't looked carefully at the gateway code, maybe related to HttpClient, maybe related ThreadLocal.

wu-sheng:
> the docs should state that the gateway plugin depends on the webflux plugin

Yes. Then should change
> Plugin of Spring Cloud Gateway 2.1.x in optional plugin folder. Please only active this plugin when you install agent in Spring Gateway.

to 
> Plugin of Spring Cloud Gateway 2.1.x in optional plugin folder. Please only active this plugin when you install agent in Spring Gateway. `spring-cloud-gateway-2.x-plugin` and `spring-webflux-5.x-plugin` are both required.

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -133,6 +133,7 @@ Now, we have the following known optional plugins.
 * Plugin of Spring Cloud Gateway 2.1.x in optional plugin folder. Please only active this plugin when you install agent in Spring Gateway.
 * Plugin of [Play Framework](https://www.playframework.com/) 2.6+ (JDK 1.8 required & Scala 2.12/2.13) in optional plugin folder. Please only active this plugin when you install agent in [Play Framework](https://www.playframework.com/). 
 * Plugin of Spring Transaction in optional plugin folder. The reason of being optional plugin is, many local span are generated, which also spend more CPU, memory and network.
+* Plugin of Spring Webflux 5.x in optional plugin folder. Agent requires JDK 1.8+,  Please make sure you use JDK1.8, when using this Agent
```
wu-sheng:
```suggestion
* Plugin of Spring Webflux 5.x in the optional plugin folder. Because the plugin requires JDK 1.8+.
```

~~Please make sure you use JDK1.8, when using this Agent~~. We don't need this, once user is using Webflux, they have JDK 1.8+ for sure.

wu-sheng:
> #3985 and #2711 is fix, and the docs should state that the gateway plugin depends on the webflux plugin

I just added the `resolves` for those two bugs in your PRs.

docs/en/setup/service-agent/java-agent/README.md
cyejing: ok

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/reactive/DispatcherHandlerHandleMethodInterceptor.java
```diff
@@ -16,62 +16,93 @@
...
+        while (next.hasNext()) {
+            next = next.next();
+            List<String> header = headers.get(next.getHeadKey());
+            if (header != null && header.size() > 0) {
+                next.setHeadValue(header.get(0));
+            }
+        }
+
+        String pathStr = exchange.getRequest().getPath().value();
+        String methodStr = exchange.getRequest().getMethodValue();
```
zhaoyuguang: if the route is `/test/{var}`, when one request url is `/test/a`, the methodStr is `/test/a`,right?
this is why before plugin, `DispatcherHandlerInvokeHandlerMethodInterceptor ` reset operationName.
in this case the `EndpointNameDictionary` will growth more, so if not support this case, need to explain on `the md file` to everyone know. FYI @wu-sheng

cyejing: you're right,  but this problem exists in web interfaces such as tomcat and undertow. right? Is it possible to have a way. When there is a lot of data, the collection program can combine characteristics similarly.

wu-sheng: I think this should be supported. In the Gateway related plugin, this is super important.

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/reactive/DispatcherHandlerHandleMethodInterceptor.java
```diff
@@ -16,62 +16,93 @@
...
+        String operationName = methodStr + ":" + pathStr;
+
+        AbstractTracerContext context = extendService.createTraceContext(operationName, false);
```
zhaoyuguang: Use ContextManager is not enough? extendService.createTraceContext API  is not very suggested... FYI @wu-sheng

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/reactive/DispatcherHandlerHandleMethodInterceptor.java
cyejing: What should I do? Keep using old operationName？

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/reactive/DispatcherHandlerHandleMethodInterceptor.java
wu-sheng: Only `ContextManager` is open API, good catch. Please change to the right API.

zhaoyuguang:
@cyejing  Friendly ping, Look like you have lots experience in spring reactive, from this testcase, SkyWalking agent not supported Spring reactive WebClient component, If you wanna,  you can support https://github.com/apache/skywalking/issues/3992

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/reactive/DispatcherHandlerHandleMethodInterceptor.java
cyejing: ContextManager Internally use ThreadLocal to save Context.  In this scenario, ThreadLocal cannot be used

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/reactive/DispatcherHandlerHandleMethodInterceptor.java
wu-sheng: Make `/test/{var}` as the operation name, rather than `/test/a`, `/test/b`. The real URL only exists in TAG only.

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/reactive/DispatcherHandlerHandleMethodInterceptor.java
wu-sheng: Let's talk about this more. If you use `extendService`, who will manage the AbstractTracerContext?

wu-sheng: And why ThreadLocal should be avoided.

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/reactive/DispatcherHandlerHandleMethodInterceptor.java
cyejing: I do n’t have a good way, can anyone help. @wu-sheng  @zhaoyuguang

zhaoyuguang: If the spring-web supported by tomcat or undertow should be no problem, if you have this case, please add a issue to others know, thanks.
 
previous webflux plugins didn't write very well at performance, but can work,  you can see `apm-sniffer/apm-sdk-plugin/spring-plugins/webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/DispatcherHandlerHandleResultMethodInterceptor.java`

cyejing: all operationName  use "WEBFLUX.handle"? right?

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/reactive/DispatcherHandlerHandleMethodInterceptor.java
zhaoyuguang: Please say more detail , You want control Context at skyWalkingDynamicField? but when createEntry span ThreadLocal will used.

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/reactive/DispatcherHandlerHandleMethodInterceptor.java
cyejing: https://github.com/apache/skywalking/blob/52de0c6dfad5750239267d7808a5016eb51a5068/apm-sniffer/apm-sdk-plugin/tomcat-7.x-8.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/tomcat78x/TomcatInvokeInterceptor.java#L75   @zhaoyuguang  if tomcat servlet url-pattern is ``/**``  what's going to happen

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/reactive/DispatcherHandlerHandleMethodInterceptor.java
wu-sheng: The reason we suggest and open ContextManager, because the plugin should be able to be aware of other plugins in the same context. Once you created the Context and managed by yourself, span created by other spans can't be linked to the Context you created. Because they can't find your Context.

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/reactive/DispatcherHandlerHandleMethodInterceptor.java
wu-sheng: This will be overridden by SpringMVC plugin if it is activated. We are using Spring Controller pattern today.

wu-sheng:
Such as in your expected data file, https://github.com/apache/skywalking/blob/86c35228991776eb53e208800f5159aa32d6909a/test/plugin/scenarios/webflux-scenario/config/expectedData.yaml

There are always entry spans, how could I link exit and local span in the segments? That is why we are using ContextManager. If you have concerns about thread switching, you should choose

`1. Create a snapshot to propagate.` Or `2. Create an async span to finish in another thread`.

It depends on your scenarios. If the whole thread is going to be used by another HTTP request, choose (1), if a span requires to finish in another thread, choose (2). Or both if you are in those cases.

xiaoy00:
> @cyejing Friendly ping, Look like you have lots experience in spring reactive, from this testcase, SkyWalking agent not supported Spring reactive WebClient component, If you wanna, you can support #3992

it can be support at an other PR is better .  i think.

cyejing:
> 1. Create a snapshot to propagate. Or 2. Create an async span to finish in another thread.

I tried this way, If two threads switch, will produce two spans with very short time spans, but the entire request time is very long.  
My approach is to let the ``Context`` manage the full request. The requested time span is also displayed normally. 
at reactive programming, If you want to rely on ThreadLocal to pass parameters, you will often encounter obstacles.  
so I think we should define a set of APIs suitable for reactive programming

wu-sheng:
For snapshot, yes, this could happen. But when you choose solution 2, you finish the span in another thread, why the duration is not right? I am not following. 

No matter what API for reactive you mean, you can't resolve the issue, the context needs to cooperate with other plugins. So eventually, you need back to the ThreadLocal in business local process. That is why we have `AbstractTracingSpan#prepareForAsync` and `AbstractTracingSpan#asyncFinish`.

wu-sheng:
> I tried this way, If two threads switch, will produce two spans with very short time spans, but the entire request time is very long.

I am giving you two options, not just one.

cyejing:
ok, I keep old operationName "WEBFLUX.handle", and set Span to ``SkyWalkingDynamicField``, so no need to modify the gateway .
then use ``2. Create an async span to finish in another thread``

wu-sheng:
> ok, I keep old operationName "WEBFLUX.handle"

If you can't find the pattern to get the parameter format, this could be better. And you could put the URL in the tag, which is safer place.

wu-sheng:
@zhaoyuguang Please recheck.

test/plugin/scenarios/webflux-scenario/config/expectedData.yaml
```diff
@@ -14,30 +14,37 @@
...
   heartbeat: []
 segmentItems:
-  - applicationCode: webflux-projectB-scenario
+  - applicationCode: webflux-scenario
     segmentSize: nq 0
     segments:
       - segmentId: not null
         spans:
-          - operationName: org.apache.skywalking.apm.testcase.sc.webflux.projectB.controller.TestAnnotationController.success
+          - operationName: WEBFLUX.handle
```
wu-sheng: Why removed the class name as operation name? Could you explain?

test/plugin/scenarios/webflux-scenario/config/expectedData.yaml
```diff
@@ -14,30 +14,37 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 registryItems:
   applications:
-    - {webflux-projectB-scenario: nq 0}
-    - {webflux-projectA-scenario: nq 0}
+    - {webflux-scenario: nq 0}
```
wu-sheng: Why do you remove the 2nd app? It makes this test case can't verify the ref. I think you need to add it back.

test/plugin/scenarios/webflux-scenario/config/expectedData.yaml
wu-sheng: By using the class name(old one), it is not as clear as URL, but at least, it could separate different logic. Today, you posted here are all the same operation names. This plugin is becoming useless.

test/plugin/scenarios/gateway-scenario/config/expectedData.yaml
```diff
@@ -69,13 +69,13 @@ segmentItems:
...
-              - {parentEndpointId: 0, parentEndpoint: GATEWAY/provider_route, networkAddressId: 0,
+              - {parentEndpointId: 0, parentEndpoint: WEBFLUX.handle, networkAddressId: 0,
                  entryEndpointId: 0, refType: CrossThread, parentSpanId: 0, parentTraceSegmentId: not null,
-                 parentServiceInstanceId: not null, networkAddress: '', entryEndpoint: GATEWAY/provider_route,
+                 parentServiceInstanceId: not null, networkAddress: '', entryEndpoint: WEBFLUX.handle,
                  entryServiceInstanceId: not null}
       - segmentId: not null
         spans:
-          - operationName: GATEWAY/provider_route
+          - operationName: WEBFLUX.handle
```
wu-sheng: Reminder only, as it is related to my review in the webflux plugin test case.

zhaoyuguang:
@cyejing hi, What is the status of this pr?

cyejing:
@wu-sheng @zhaoyuguang  I try do it
I would like to know ， Execute in one thread `` span.prepareForAsync();  ContextManager.stopSpan(span);`` , How to modify operationName in another thread . ``setOperationName`` failed.

cyejing:
image
I see no changes allowed

wu-sheng:
> How to modify operationName in another thread . setOperationName failed.

Could you share how fail?

cyejing:
> > How to modify operationName in another thread . setOperationName failed.
> 
> Could you share how fail?

like my code shown  ``stackDepth`` is reduced because ``ContextManager.stopSpan`` .  ``ContextManager.stopSpan`` is executed to clean up the context of the current thread

wu-sheng:
> like my code shown stackDepth is reduced because ContextManager.stopSpan . ContextManager.stopSpan is executed to clean up the context of the current thread

When you have the async span, the `#stop` is only reducing the stackDepth as you said. But you still have span ref, right? You could propagate it to any thread, and operate it. The core will wait for you to do real async finish.

And the issue you posted is true, I didn't realize it before. I think all similar methods in `#EntrySpan` and `#ExitSpan` should add `setxxxInAsync`(work in `isInAsyncMode==true`) to ignore the `currentMaxDepth` and `stackDepth`. At the same time, `setxxx` should not work in `isInAsyncMode==true`. Make sense?

If YES, please submit a new pull request to adjust that. We should add them separately. The changes should begin with `AbstractSpan`.

wu-sheng:
I think to add `setOperationNameInAsync` should be enough, right? I don't have the scenario to do this for other methods.

cyejing:
Yes, Currently, I need ``setOperationNameInAsync ``  to set ``urlPattern`` . 
if so can I add ``setOperationNameInAsync ``  on this PR

wu-sheng:
> if so can I add setOperationNameInAsync on this PR

Please don't. Submit a new pull request to add this. This is a core feature changed due to new API. It should be added into changelog.

cyejing:
ok, understand, Then I stop this PR first.
submit next time

wu-sheng:
> ok, understand, Then I stop this PR first.

Just keep it open. It is fine. After the API change PR merged, we could sync to here automatically.

wu-sheng:
API changes has sync to your branch. You should be good to go.

cyejing:
``OperationName`` use url  pattern, and use OkHttpClient  add webflux exitSpan test case

test/plugin/scenarios/gateway-scenario/config/expectedData.yaml
```diff
@@ -23,7 +23,7 @@ registryItems:
     - {gateway-projectA-scenario: 1}
   operationNames:
     - gateway-projectB-scenario: [/provider/b/healthCheck, /provider/b/testcase]
```
wu-sheng: Existing health check URL should be removed.

test/plugin/scenarios/webflux-scenario/pom.xml
```diff
@@ -23,36 +23,85 @@
...
+    <pluginRepositories>
+        <pluginRepository>
+            <id>spring-snapshots</id>
+            <url>http://repo.spring.io/snapshot</url>
+        </pluginRepository>
+        <pluginRepository>
+            <id>spring-milestones</id>
+            <url>http://repo.spring.io/milestone</url>
+        </pluginRepository>
+    </pluginRepositories>
```
wu-sheng: These repos should not exist, right?

test/plugin/scenarios/webflux-scenario/config/expectedData.yaml
```diff
@@ -51,15 +60,35 @@ segmentItems:
...
+                 entryEndpointId: 0, refType: CrossProcess, parentSpanId: 0, parentTraceSegmentId: not null,
+                 parentServiceInstanceId: not null, networkAddress: 'localhost:8080', entryEndpoint: /testcase/annotation/hello,
                  entryServiceInstanceId: not null}
       - segmentId: not null
         spans:
-          - operationName: org.apache.skywalking.apm.testcase.sc.webflux.projectB.controller.TestAnnotationController.error
+          - operationName: /testcase/annotation/hello
+            operationId: 0
+            parentSpanId: -1
+            spanId: 0
```
wu-sheng: Why are all these spans id=0 and parentSpanId=-1, and no ref?  What is the relationship with other segments?

cyejing:
revert old test case,  and revert test case . please recheck @wu-sheng

wu-sheng:
As #4059 merged, you should notice
1. Local and exit span wouldn't trigger operation name register
1. Parent and entry endpoint in the ref will be -1 when they are belong ing to local span, as they are not endpoint.
1. Test expected data changed. Conflicts need to be resolved.

wu-sheng:
@zhaoyuguang Could you take a look?

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/DispatcherHandlerHandleMethodInterceptor.java
```diff
@@ -18,60 +18,87 @@
...
+                    span.setOperationName(urlPattern);
+                }
+                // fix webflux-2.0.0-2.1.0 version have bug. httpStatus is null. not support
+                if (httpStatus != null) {
+                    Tags.STATUS_CODE.set(span, Integer.toString(httpStatus.value()));
+                    if (httpStatus.isError()) {
+                        span.errorOccurred();
+                    }
+                }
+            } finally {
```
zhaoyuguang: there is easy get exception?
if true, can add 
```
private static final ILog logger = LogManager.getLogger(xxx.class) 
catch exception
   logger.error
```
to avoid print exception to user console log;

wu-sheng: I think this is not for exception catch, this is making sure span finished.

zhaoyuguang:
Except my code review comment, This pr is LGTM

cyejing:
ok, I adjust the code.

arugal:
Hi @cyejing, there are update submodule in this PR, please recheck.

cyejing:
> Hi @cyejing, there are [update submodule](https://github.com/apache/skywalking/pull/3945/files#diff-8693f16a05265e9581e98cb7f3ff817a) in this PR, please recheck.

@arugal  I have revert

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/define/DispatcherHandlerInstrumentation.java
```diff
@@ -29,7 +29,7 @@
 import static org.apache.skywalking.apm.agent.core.plugin.match.NameMatch.byName;
 
 /**
- * @author zhaoyuguang
+ * @author Born
  */
```
arugal: It should be `@author zhaoyuguang Born`, right?

kezhenxu94: > It should be `@author zhaoyuguang Born`, right?

both are fine and used in many open-sourced projects, Spring Framework uses the latter one

cyejing: I didn't notice because the two projects were merged,I will modify it

wu-sheng: @kezhenxu94 I think, @arugal 's point is, `zhaoyuguang` should not be removed :) You could have multiple lines of `@author` or two names, separated by `,`.

kezhenxu94: > @kezhenxu94 I think, @arugal 's point is, `zhaoyuguang` should not be removed :) You could have multiple lines of `@author` or two names, separated by `,`.

Haha, I miss that, sorry about that

arugal: Haha, wo ji hua ming nian qu shang ying yu bu xi ban :)

wu-sheng: @arugal Please use English for all comments.

PR #3948, opened 2019-11-27
Immigrate test mysql scenario.

wayilau:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance
- [x] Tests cases.


wu-sheng:
@wayilau Send a mail to your cmss mailbox, please check.

wayilau:
> @wayilau Send a mail to your cmss mailbox, please check.

@wu-sheng  I received message you just wrote.

test/plugin/scenarios/mysql-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,45 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+registryItems:
+  applications:
+    - {mysql-scenario: nq 0}
+  instances:
+    - {mysql-scenario: 1}
+  operationNames:
+    - mysql-scenario: [/mysql-scenario/case/mysql]
```
dmsolr: Are you missing some operatioName? For example, /JDBI/statement/execute.

Jenkinsfile-Agent-Test-3
```diff
@@ -67,7 +67,7 @@ pipeline {
                 sh './mvnw -f test/plugin/pom.xml clean package -DskipTests docker:build'
             }
         }
-        stage('Test Cases Report (161)') {
+        stage('Test Cases Report (214)') {
```
dmsolr: Could you make it place at Workload2-Group2?

wayilau: I am glad to.

arugal:
Ref #3583

arugal:
Hi, I have a question, what do we not use ```JVM-container```？

test/plugin/scenarios/mysql-scenario/config/expectedData.yaml
wayilau: I will copy it from original source.

wayilau:
> Hi, I have a question, what do we not use `JVM-container`？

original tests cases use tomcat, so i follow. @arugal

wu-sheng:
Jenkins jobs have been disabled, please add your tasks to GitHub action file, and fix the conflict, remove the Jenkins file. https://github.com/apache/skywalking/blob/master/.github/workflows/plugins-test.yaml

wayilau:
> Jenkins jobs have been disabled, please add your tasks to GitHub action file, and fix the conflict, remove the Jenkins file. https://github.com/apache/skywalking/blob/master/.github/workflows/plugins-test.yaml

ok.

wayilau:
> > Hi, I have a question, what do we not use `JVM-container`？
> 
> original tests cases use tomcat, so i follow. @arugal

As  I have moved to JVM container. Beacause tomcat met some problems.

dmsolr:
It is very sad to tell you that you need to update the Action configuration (resolve conflicts) because those are updated after #3956 merged. Maybe you need to move it to another group which is faster.
And the test failed (Tomcat-container doesn't need to pack it as a ZIP package. )

wayilau:
> It is very sad to tell you that you need to update the Action configuration (resolve conflicts) because those are updated after #3956 merged. Maybe you need to move it to another group which is faster.
> And the test failed (Tomcat-container doesn't need to pack it as a ZIP package. )

ok, I am using jVM container. May be i lost some files. i will check then submit again.

wu-sheng:
Jenkins is not there any more. We are working out with Apache why those are still there.

wu-sheng:
You should add MySQL As a separated group in that test case, named As MySQL case like others.

wayilau:
> You should add MySQL As a separated group in that test case, named As MySQL case like others.

ok

wu-sheng:
What is `PluginsTest / .github/workflows/plugins-test.yaml (pull_request)`?

.github/workflows/plugins-test.yaml
```diff
@@ -464,4 +464,30 @@ jobs:
       - name: Build the Docker image
         run: ./mvnw -f test/plugin/pom.xml clean package -DskipTests docker:build -DBUILD_NO=local >/dev/null
       - name: Run kafka 0.11.0.0-2.3.0 (16)
-        run: bash test/plugin/run.sh kafka-scenario
\ No newline at end of file
+        run: bash test/plugin/run.sh kafka-scenario
+
+Mysql:
```
wu-sheng: This is not a correct format, This is a YAML.

kezhenxu94: The indentation is not correct IMO, syntax looks good

dmsolr:
It still fails. Please recheck.

wayilau:
> What is `PluginsTest / .github/workflows/plugins-test.yaml (pull_request)`?
Log tells me that docker startup failed. I don't know why. Could you help me check ? @dmsolr

wu-sheng:
@wayilau I think it was caused by the wrong YAML format. Has been fixed. Now, you need to make your tests passed.

wayilau:
> @wayilau I think it was caused by the wrong YAML format. Has been fixed. Now, you need to make your tests passed.

I think  i know why. I missed some commits.

dmsolr:
It looks like good.
Let's wait for this result. :)

wu-sheng:
@wayilau  I don't get any reply from you about the mail. Are you planning to do?

wayilau:
> @wayilau I don't get any reply from you about the mail. Are you planning to do?

@wu-sheng 
I did not receive any emails  from you yet. So i don't understand ..

wu-sheng:
Could you send a mail to me? I send you a mail based on your GitHub homepage mail, maybe it is incorrect? Mind is correct, you could send a mail to there.

wayilau:
> Could you send a mail to me? I send you a mail based on your GitHub homepage mail, maybe it is incorrect? Mind is correct, you could send a mail to there.

I am sorry, i just saw mail from my  rubbish mail box. May be the mail settings is limit  by the commany.  I will read the mail first. @wu-sheng

Issue #3829, opened 2019-11-11
Service instance dependency and metrics of them


wu-sheng:
We are planning to provide service instance level dependency and metrics of them. 

Before that, the mapping of IP:port to service must be improved. We need to add IP:port to instance mapping.

Also after this, we could add OAL and manual relationship about service instance relationship and metrics of them.

The last step is providing GraphQL query for instance dependency. The idea right now is querying the instance relationship by given services of source and dest, we are not providing instance-level topology by default.

@kezhenxu94 Do you have interests to do this? Or any other @apache/skywalking-committers wants to do? Please let me know.

After the above is done, UI need to be upgraded too. When clicking the service relationship, the user should have a chance(button?) to dig deeper to instance relationships, and binding with metrics(similar to the service relationship and metrics in the topology). @Fine0830 @x22x22 @TinyAllen Which one of you could take this part in next release, please let me know.

arugal:
So cool. I want to try it, but I'm not sure I can handle it.

wu-sheng:
> So cool. I want to try it, but I'm not sure I can handle it.

You need to understand how current topology generated from trace raw data. This is not an easy task. This enhancement could be a separated release from my understanding. I would say 6.6.0 or 6.7.0 should for this, before the end of this year, as a closure and super powerful feature of SkyWalking 2019.

arugal:
> > So cool. I want to try it, but I'm not sure I can handle it.
> 
> You need to understand how current topology generated from trace raw data. This is not an easy task. This enhancement could be a separated release from my understanding. I would say 6.6.0 or 6.7.0 should for this, before the end of this year, as a closure and super powerful feature of SkyWalking 2019.

I know about the collection of the raw data, but I don't understand topology generation. maybe I can do the odd jobs?

wu-sheng:
You should read
1. MultiScopesSpanListener
1. Service Topology Query

Then, you may get the idea.

dmsolr:
I would like to do it with you, @arugal . But I am not familiar at OAL.

wu-sheng:
> I would like to do it with you, @arugal . But I am not familiar at OAL.

OAL is a super easy part :) The mapping and new query are really works.

arugal:
Give us two weeks to learn. If no one else accepts the assignment, just assign the task to us. How about this idea?

wu-sheng:
> Give us two weeks to learn. If no one else accepts the assignment, just assign the task to us. How about this idea?

[...]

PR #3950, opened 2019-11-27
[Feature] Service instance dependency

arugal:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#3829
___

+ Storage ```ServiceInstanceRelationServerSideMetrics``` and ```ServiceInstanceRelationClientSideMetrics```
+ Add and update  ```ServiceInstanceInventory#mappingServiceInstanceId```
+ Add ```getServiceInstanceTopology(clientServiceId: ID!, serverServiceId: ID!, duration: Duration!): Topology```  to ```TopologyQuery```


wu-sheng:
> getServiceInstanceTopology(serviceIds: [ID!]!, duration: Duration!): Topology to TopologyQuery

This is not right. For service instance relationship, it should not be defined as topology. The backend should allow query instance relationships when given two specific service IDs. Don't consider it as same as the old topology for service. They are different.

arugal:
When both ```serviceA -> serviceB``` and ```serviceB -> serviceA``` calls exist in the topology,  how do I handle it?

arugal:
works like this: 
service

instance

wu-sheng:
I will review this today.

arugal:
I verified es6 and es7 and found no problem.

wu-sheng:
Do you really have two instances of projectB?

arugal:
> Do you really have two instances of projectB?

Yes.

wu-sheng:
First, codes look good enough, but I also prefer others could review too. @kezhenxu94 Do you have time to do this?

@arugal For you, I think you need to update a document about storage extension. https://github.com/apache/skywalking/blob/master/docs/en/guides/storage-extention.md. I think you add new APIs, could you run and check whether we miss other APIs?

arugal:
> @arugal For you, I think you need to update a document about storage extension. https://github.com/apache/skywalking/blob/master/docs/en/guides/storage-extention.md. I think you add new APIs, could you run and check whether we miss other APIs?

I will check.

arugal:
@wu-sheng @kezhenxu94  Need to update the e2e test?

wu-sheng:
> Need to update the e2e test?

Yes. We should. But maybe after @kezhenxu94 basic code review?

PR #3973, opened 2019-11-29
skip unnecessary action

dmsolr:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues

___

To skip Github Action when it is unnecessary.


wu-sheng:
Skip successfully.

wu-sheng:
> Why not add skip condition to the other checks

Because I required the `required check`. In GitHub Action, if skip happens, it never shows as passed :)

PR #3997, opened 2019-12-03
[test/plugin]immigrate pulsar scenario

codelipenghui:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#3583


wu-sheng:
Your codes may miss one Apache 2.0 header. Rat check fails.

.github/workflows/plugins-test.yaml
```diff
@@ -492,3 +492,28 @@ jobs:
         run: ./mvnw -f test/plugin/pom.xml clean package -DskipTests docker:build -DBUILD_NO=local >/dev/null
       - name: Run mysql 5.1.2-8.0.15 (53)
         run: bash test/plugin/run.sh mysql-scenario
+
+  Pulsar:
```
wu-sheng: Pulsar test is quick enough, please add it into `Feign_Canal` group, and rename it to `Feign_Canal_ Pulsar`.

codelipenghui: Sure

codelipenghui: sure

test/plugin/scenarios/pulsar-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,84 @@
+registryItems:
```
dmsolr: Missing License Header here.

codelipenghui: Thanks for reminding

codelipenghui:
The Agent Plugin tests is still failed and then i check the test result, i got 404 there. How can i get some information about the failed tests.

wu-sheng:
@codelipenghui Ignore that. We are moving to GitHub action. Jenkins jobs are removed, but because of some INFRA team issues, they are keeping showing up.

PR #4006, opened 2019-12-05
[test/plugin] immigrate test oracle scenario.

aderm:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#3583 
___
### Bug fix
- Bug description.

- How to fix?

___

### DIFF
--- a/.github/workflows/plugins-test.yaml
+++ b/.github/workflows/plugins-test.yaml
@@ -465,16 +465,13 @@ jobs:
           java-version: 8
       - name: Build SkyWalking Agent
         run: ./mvnw clean package -DskipTests -Pagent >/dev/null
-      - name: Download apm-oracle-10.x-plugin-1.0.1.jar to agent plugins dir.
-        run: curl -L -o ./skywalking-agent/plugins/apm-oracle-10.x-plugin-1.0.1.jar https://github.com/SkyAPM/java-plugin-extensions/releases/download/1.0.1/apm-oracle-10.x-plugin-1.0.1.jar
-      - name: Download ojdbc14.jar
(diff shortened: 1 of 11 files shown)


.github/workflows/plugins-test.yaml
```diff
@@ -494,3 +494,34 @@ jobs:
         run: ./mvnw -f test/plugin/pom.xml clean package -DskipTests docker:build -DBUILD_NO=local >/dev/null
       - name: Run mysql 5.1.2-8.0.15 (53)
         run: bash test/plugin/run.sh mysql-scenario
+
+  Oracle:
```
wu-sheng: Oracle just have one version to test, so you could add this to a faster group, rather than add a new one.

aderm: ok

.github/workflows/plugins-test.yaml
```diff
@@ -465,8 +465,16 @@ jobs:
           java-version: 8
       - name: Build SkyWalking Agent
         run: ./mvnw clean package -DskipTests -Pagent >/dev/null
+      - name: Download apm-oracle-10.x-plugin-1.0.1.jar to agent plugins dir.
```
dmsolr: we can configure multi-command with `run:.`
I suggest,
``` 
- name: Run oracle 10.2.0.4.0 (1)
   run: |
       curl -L -o ...
       curl -L -o ...
       ./mvnw ...
       ./mvnw ...
       bash ....
```

WDYT?

aderm: ACK

PR #4007, opened 2019-12-05
spring @Bean annotation was annotation to the method.

ZhuoSiChen:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
spring @Bean annotation was annotation to the method.
- How to fix?

___

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
@@ -40,7 +40,7 @@ public class SnifferConfigInitializer {
     private static String DEFAULT_CONFIG_FILE_NAME = "/config/agent.config";
     private static String ENV_KEY_PREFIX = "skywalking.";
     private static boolean IS_INIT_COMPLETED = false;
-
+    private static Properties properties;
     /**
      * If the specified agent config path is set, the agent will try to locate the specified agent config. If the
      * specified agent config path is not set , the agent will try to locate `agent.config`, which should be in the
@@ -57,7 +57,7 @@ public class SnifferConfigInitializer {
 
         try {
             configFileStream = loadConfig();
-            Properties properties = new Properties();
+            properties = new Properties();
             properties.load(configFileStream);
             for (String key : properties.stringPropertyNames()) {
                 String value = (String)properties.get(key);


apm-sniffer/optional-plugins/optional-spring-plugins/spring-annotation-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/annotations/bean/SpringBeanInstrumentation.java
```diff
@@ -31,6 +31,6 @@
     public static final String ENHANCE_ANNOTATION = "org.springframework.context.annotation.Bean";
 
     @Override protected ClassMatch enhanceClass() {
-        return byClassAnnotationMatch(new String[] {ENHANCE_ANNOTATION});
+        return byMethodAnnotationMatch(new String[] {ENHANCE_ANNOTATION});
```
wu-sheng: Method annotation? What is this usage? `@Bean` on the method?

ZhuoSiChen: I had see the org.springframework.context.annotation.Bean
was this ```java
@Target({ElementType.METHOD, ElementType.ANNOTATION_TYPE})
@Retention(RetentionPolicy.RUNTIME)
@Documented
public @interface Bean {

wu-sheng: It means could be `method` or `class`. So you should not change this, you should add a new one. 

Please check here, https://stackoverflow.com/questions/10604298/spring-component-versus-bean. `@bean` on method is another case, if you want that, it is just simply changing the existing one, you need to add a new one for the new case :)

ZhuoSiChen: I want write a test for  that,but I don't how to do.

wu-sheng: First this change is not right, you should instrument classes and methods both. 

The method annotation is different. You should learn that first, and submit a new instrument, Then we talk about the tests which is easy part.

ZhuoSiChen: The java compiler was not allow for @Bean in a class. So why I edit this and not add a new one.

wu-sheng: The way you changed is, making the interceptor to intercept this method. But actually, in your mind, you should instrument the return instance. 

Even this works, it is not right. That is my point. The old and yours are both wrong. That is why I asked you to work on the right one. We should dig deep about how to make it right.

wu-sheng: I think you need to work on instrumenting Spring core about how this annotation works, and work there.

ZhuoSiChen: Ok! You are right。

ZhuoSiChen: Based on the above discussion, what I need to enhance is the result returned by the method which be annotation `@Bean`.And I was found the Spring core the `@Bean` Annotation was used by `org.springframework.context.annotation.ConfigurationClassEnhancer.BeanMethodInterceptor#intercept`. The method will return two types of beans.
one is  `org.springframework.beans.factory.FactoryBean` ,the other was normal `bean`。
And I still unknow how to enhance the method be annotationed `@Bean`,
Because the enhance Object was in return value.

wu-sheng: Yes, I am thinking about this too. Look like a little different from agent perspective.

wu-sheng: I think either using [custom enhance](https://github.com/apache/skywalking/blob/master/docs/en/setup/service-agent/java-agent/Customize-enhance-trace.md), or you need to do AOP by using Spring APIs for the return object. 
The agent plugin mechanism is only could give the opportunity to have the return object, anything else should be done through Spring APIs.

ZhuoSiChen: 1. I still unknown how to do with  custom enhance. Can you explain how to do it? 
Because I see the implement was still base on Class. And not base on method.
2. I don't know how to used AOP to return object in skywalking.( I think this will had a problem，if we call this method manually)

Can we do that when we load the Class has method annotated by `@Bean`
,we can modify the method return ClassType,and this ClassType was extends the orginal ClassType.

wu-sheng:
@geektcp What are you doing? This PR breaks all CI processes and has conflict with the master. Why is this good?

Issue #4012, opened 2019-12-06
Hope to add async span to modify operation name method 


cyejing:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [x] Feature or performance improvement
___
### Requirement or improvement
- Hope to add async span to modify ``OperationName`` method 

as https://github.com/apache/skywalking/pull/3945#issuecomment-562410420

PR #4015, opened 2019-12-06
Support set operation name in async mode span

cyejing:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/4012

___
add async span to modify OperationName method


cyejing:
I am thinking how to avoid using ``setOperationName`` through API , Because there are two ways to set operation name , Make user choose

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/AbstractTracingSpan.java
```diff
@@ -349,4 +349,10 @@ public AbstractSpan start(long startTime) {
         isAsyncStopped = true;
         return this;
     }
+
+    @Override
+    public AbstractSpan setOperationNameInAsync(String operationName) {
+        return setOperationName(operationName);
```
wu-sheng: You should check the span is in async mode. Meanwhile, you should make the `setOperationName` disable when in async mode. Right?

wu-sheng: Also, new API should be introduced in `Java-Plugin-Development-Guide.md` doc

wu-sheng:
@cyejing Any update?

cyejing:
Now updated, please check

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/EntrySpan.java
```diff
@@ -96,6 +96,9 @@ public AbstractTracingSpan setComponent(String componentName) {
 
     @Override
     public AbstractTracingSpan setOperationName(String operationName) {
+        if (isInAsyncMode) {
+            throw new RuntimeException("Span is in async mode, Should use setOperationNameInAsync method");
```
wu-sheng: This method is called by `#setOperationNameInAsync`. I am pretty sure this will be triggered :)

wu-sheng: Don't throw an exception, it caused too much.

cyejing: Isn't it necessary to add ``setOperationNameInAsync``,  judging from this place is async mode, and modify ``operationName``,Isn't this better, only one api

wu-sheng: Make sense. Let's keep API unchanged. And revert all other changes.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/EntrySpan.java
```diff
@@ -96,7 +96,7 @@ public AbstractTracingSpan setComponent(String componentName) {
 
     @Override
     public AbstractTracingSpan setOperationName(String operationName) {
-        if (stackDepth == currentMaxDepth) {
+        if (stackDepth == currentMaxDepth || isInAsyncMode) {
```
wu-sheng: I think you should add this to ExitSpan?

cyejing: ok, If exitspan also supports this scenario

wu-sheng: I think both of them are similar cases. Just for different plugins.

PR #3978, opened 2019-11-29
[Feature] Service instance dependency

arugal:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#3829
___
### New feature or improvement

+ Storage ```ServiceInstanceRelationServerSideMetrics``` and ```ServiceInstanceRelationClientSideMetrics```
+ Add and update  ```ServiceInstanceInventory#mappingServiceInstanceId```
+ Add ```getServiceInstanceTopology(clientServiceId: ID!, serverServiceId: ID!, duration: Duration!): ServiceInstanceTopology```  to ```TopologyQuery```

___
Sorry, because I deleted the fork repository branch by mistake, I can only open a new PR. 

Ref #3950

PR #4026, opened 2019-12-08
sync query-protocol repo

x22x22:

sync query-protocol repo


wu-sheng:
Please don’t sync this. This related PR #3978 is still under reviewing.

PR #4039, opened 2019-12-10
Fix NPE when worker not found, mostly new OAL metrics added

mrproliu:

- [-] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
If the worker name not found, There will get NullPointerException.

- How to fix?
modify the check field is handleWorker.
___


wu-sheng:
Add a scenario to trigger this NPE, new OAL script added, and OAP nodes update one by one. During this process, other old nodes will generate NPE by this.

Issue #3920, opened 2019-11-22
Document how to build and run docker/docker-compose located at main repo


hanahmily:
As indicated by title.

hanahmily:
`docker/README.md`

PR #4043, opened 2019-12-11
Add es7 support to docker

hanahmily:
1. ~~Add dockerfile for es7~~
1. ~~Update entrypoint bash~~
1. ~~Add docker compose file~~
1. ~~Update github action task~~
1. ~~document~~


wu-sheng:
Does this plan to fix #3920?

stalary:
How's it going? I need a Dockerfile for es7.

.github/workflows/docker-ci.yaml
```diff
@@ -24,6 +24,9 @@ env:
 jobs:
   build:
     runs-on: ubuntu-16.04
+    strategy:
+      matrix:
+        es: [6, 7]
```
wu-sheng: This way makes the CI name no clear. Please try to make to `docker-ci / build(es6)` at least.

hanahmily:
> Does this plan to fix #3920?

yep. it will fix #3920

docs/en/guides/How-to-build.md
```diff
@@ -51,18 +51,7 @@ or
...
-- Build oap server docker image
-> make docker.oap
-
-- Build ui docker image
-> make docker.ui
-
-`HUB` and `TAG` variables ares used to setup `REPOSITORY` and `TAG` of a docker image. To get
-a oap image with name `bar/oap:foo`, run the following command
-> HUB=bar TAG=foo make docker.oap
+Refer to [Build docker image](https://github.com/apache/skywalking/tree/master/docker) for more details.
```
wu-sheng: I think we should add refer to `makefile` doc?

hanahmily: The usage of `makefile` is written in `docker/README.md`.  When the user navigates to this folder, they can view documents and scripts at the same time, I think it's not a bad experience for them.

wu-sheng: I know the doc is there. But at here, the doc said, 

> We can build docker images of `backend` and `ui` with `Makefile` located in root folder.

There is no tip about reading the doc there.

PR #4045, opened 2019-12-12
Set up project builds on Windows and MacOS.

wu-sheng:
Make sure the project builds on Windows and MacOS.

### DIFF
--- a/.github/workflows/ci-it.yaml
+++ b/.github/workflows/ci-it.yaml
@@ -51,7 +51,7 @@ jobs:
         run: tools/dependencies/check-LICENSE.sh
 
   CI-on-Windows:
-    runs-on: indows-latest
+    runs-on: Windows-latest
     timeout-minutes: 180
     strategy:
       fail-fast: true
--- a/.github/workflows/ci-it.yaml
+++ b/.github/workflows/ci-it.yaml
@@ -44,8 +44,8 @@ jobs:
           java-version: 8
       - name: 'Install & Test'
         run: |
-          export MAVEN_OPTS='-Dmaven.repo.local=.m2/repository -XX:+TieredCompilation -XX:TieredStopAtLevel=1 -XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC -XX:-UseGCOverheadLimit -Xmx3g'
-          ./mvnw -P"agent,backend,ui,dist,CI-with-IT" clean cobertura:cobertura verify install javadoc:javadoc
+          export MAVEN_OPTS='-XX:+TieredCompilation -XX:TieredStopAtLevel=1 -XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC -XX:-UseGCOverheadLimit -Xmx3g'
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-it.yaml
```diff
@@ -49,3 +49,49 @@ jobs:
           CODECOV_TOKEN="d2065307-8f01-4637-9715-2781ef096db7" bash <(curl -s https://codecov.io/bash)
       - name: 'Check Dependencies Licenses'
         run: tools/dependencies/check-LICENSE.sh
+
+  CI-on-Windows:
+    runs-on: indows-latest
```
kezhenxu94: What an odd name, is it a typo or really what GitHub provides?

.github/workflows/ci-it.yaml
```diff
@@ -49,3 +49,49 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          submodules: true
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
```
kezhenxu94: Actually we don't need cache, because we use `export MAVEN_OPTS='-Dmaven.repo.local=.m2/repository` when running `mvn`, so it caches nothing, or we should remove `-Dmaven.repo.local=.m2/repository` below

.github/workflows/ci-it.yaml
```diff
@@ -49,3 +49,49 @@ jobs:
...
```
kezhenxu94: same here

.github/workflows/ci-it.yaml
kezhenxu94: seems a typo

.github/workflows/ci-it.yaml
wu-sheng: Do we remove all caches? They exist on the old CI linux section too.

wu-sheng:
> And I think appveyor can be removed now

It is harmless, I will ignore it for now :) Jenkins is the pain, today.

.github/workflows/ci-it.yaml
wu-sheng: Fixed.

.github/workflows/ci-it.yaml
kezhenxu94: > Do we remove all caches? They exist on the old CI linux section too.

All the caches in CI are unnecessary because we specify another local repositories directory in `MAVEN_OPTS='-Dmaven.repo.local=.m2/repository`, and I suggest to remove caches because we want to rebuild the packages (especially for SkyWalking modules) every time in every PR, in case we bring the packages from cache (in previous build), which may be wrong

wu-sheng: I think we don't need `-Dmaven.repo.local=.m2/repository` as we are running on the GitHub action, right?

kezhenxu94: > I think we don't need `-Dmaven.repo.local=.m2/repository` as we are running on the GitHub action, right?

Yes for sure, it should be removed when migrating into GitHub Action

PR #4059, opened 2019-12-14
Remove the local/exit span register mechanism in Java agent scenario

### LINKED ISSUE #4056: Remove local span and exit span OP name register permanently
Today, there are endpoint register happens for entry, exit and local span, but we don’t need exit and local span OP names in 99% cases. We should remove it.

The only two scenarios is 
1. Segment query including endpoint id for endpoint dashboard page, but that page only shows entry span as endpoint, so it is safe to keep the column empty
2. Endpoint dependency in ref header dependency analysis in the backend. Let’s ignore the scenario of end parent endpoint is local or entry. Treat it as none, and don’t generate it as endpoint relation in the OAP side.

@kezhenxu94 Would you want to do this? Or I should run this removal.

wu-sheng:
Resolves #4056

**Key**, reduce the load of register and amount of operation id. Also, reduce the agent memory cost.
At the same time, the local span and exit span plugins will not concern about `parameter in the op name` causing `too many registers` issue.

@kezhenxu94 Could you review and run a test locally? Especially for the plugin you are working on, `Runnable` case I mean. If inside `runnable`, an RPC is called, the endpoint dependency should show that endpoint still depends on the parent endpoint(RPC in the parent thread of the runnable). If no parent thread exist, as this thread is a local task(timer), then the endpoint should depend on the user. Could you verify these two cases? And confirm no local span and exit span register happen, check through
1. No local and exit operation name in the endpoint inventory
1. No warning log, like this `Unexpected endpoint register, endpoint isn't detected from server side. ` ,happens.

### DIFF
--- a/oap-server/server-receiver-plugin/skywalking-register-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/register/provider/handler/v6/grpc/RegisterServiceHandler.java
+++ b/oap-server/server-receiver-plugin/skywalking-register-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/register/provider/handler/v6/grpc/RegisterServiceHandler.java
@@ -163,14 +163,19 @@ public class RegisterServiceHandler extends RegisterGrpc.RegisterImplBase implem
             int serviceId = endpoint.getServiceId();
             String endpointName = endpoint.getEndpointName();
 
-            int endpointId = inventoryService.getOrCreate(serviceId, endpointName, DetectPoint.fromNetworkProtocolDetectPoint(endpoint.getFrom()));
+            DetectPoint detectPoint = DetectPoint.fromNetworkProtocolDetectPoint(endpoint.getFrom());
+            if (DetectPoint.SERVER.equals(detectPoint)) {
+                int endpointId = inventoryService.getOrCreate(serviceId, endpointName, detectPoint);
 
-            if (endpointId != Const.NONE) {
-                builder.addElements(EndpointMappingElement.newBuilder()
-                    .setServiceId(serviceId)
-                    .setEndpointName(endpointName)
-                    .setEndpointId(endpointId)
-                    .setFrom(endpoint.getFrom()));
+                if (endpointId != Const.NONE) {
+                    builder.addElements(EndpointMappingElement.newBuilder()
+                        .setServiceId(serviceId)
+                        .setEndpointName(endpointName)
+                        .setEndpointId(endpointId)
+                        .setFrom(endpoint.getFrom()));
+                }
+            } else {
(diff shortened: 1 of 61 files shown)


wu-sheng:
We include exit span in the plugin test cases, I have to remove them one by one, sadly....

wu-sheng:
@dmsolr @wayilau @arugal I already updated the plugin test document and existing plugin test cases. Please review this PR too. If this PR could pass @kezhenxu94 and @ascrutae 's local tests, I would like to merge this ASAP, due to it affects all potential and existing plugin test PRs.

wu-sheng:
FYI @cyejing

oap-server/server-receiver-plugin/skywalking-register-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/register/provider/handler/v6/grpc/RegisterServiceHandler.java
```diff
@@ -163,14 +163,19 @@ public RegisterServiceHandler(ModuleManager moduleManager) {
...
+
+                if (endpointId != Const.NONE) {
+                    builder.addElements(EndpointMappingElement.newBuilder()
+                        .setServiceId(serviceId)
+                        .setEndpointName(endpointName)
+                        .setEndpointId(endpointId)
+                        .setFrom(endpoint.getFrom()));
+                }
+            } else {
+                logger.warn("Unexpected endpoint register, endpoint isn't detected from server side. {}", request);
```
wu-sheng: To all reviewers, this needs to be added into changelog and FAQ, as people may face a lot of these logs when agents have not upgraded. What do you think?

wu-sheng:
@kezhenxu94 @ascrutae All tests passed, but I think you should run a local test to re-verify.

oap-server/server-receiver-plugin/skywalking-register-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/register/provider/handler/v6/grpc/RegisterServiceHandler.java
wu-sheng: I just submit a commit to do this. please recheck.

arugal:
@wu-sheng Hi, `go2sky` needs to be modified?

wu-sheng:
> Hi, go2sky needs to be modified?

You could recheck. In my mind, all other agents and SDK are not using register at the SDK side.

arugal:
> > Hi, go2sky needs to be modified?
> 
> You could recheck. In my mind, all other agents and SDK are not using register at the SDK side.

Checked,  `go2sky` does not register endpoint.

PR #4067, opened 2019-12-15
[Feature] Add Java agent plugin for Jdk Threading classes

kezhenxu94:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

### DIFF
--- a/apm-sniffer/config/agent.config
+++ b/apm-sniffer/config/agent.config
@@ -28,14 +28,14 @@ agent.service_name=${SW_AGENT_NAME:Your_ApplicationName}
 # agent.authentication = ${SW_AGENT_AUTHENTICATION:xxxx}
 
 # The max amount of spans in a single segment.
-# Through this config item, skywalking keep your application memory cost estimated.
+# Through this config item, SkyWalking keep your application memory cost estimated.
 # agent.span_limit_per_segment=${SW_AGENT_SPAN_LIMIT:300}
 
 # Ignore the segments if their operation names end with these suffix.
 # agent.ignore_suffix=${SW_AGENT_IGNORE_SUFFIX:.jpg,.jpeg,.js,.css,.png,.bmp,.gif,.ico,.mp3,.mp4,.html,.svg}
 
-# If true, skywalking agent will save all instrumented classes files in `/debugging` folder.
-# Skywalking team may ask for these files in order to resolve compatible problem.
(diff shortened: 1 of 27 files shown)


apm-sniffer/config/agent.config
```diff
@@ -61,4 +61,7 @@ logging.level=${SW_LOGGING_LEVEL:DEBUG}
 # logging.max_history_files=${SW_LOGGING_MAX_HISTORY_FILES:-1}
 
 # mysql plugin configuration
-# plugin.mysql.trace_sql_parameters=${SW_MYSQL_TRACE_SQL_PARAMETERS:false}
\ No newline at end of file
+# plugin.mysql.trace_sql_parameters=${SW_MYSQL_TRACE_SQL_PARAMETERS:false}
+
+# Jdk Threading configuration
+plugin.jdkthreading.threading_class_prefixes=${SW_JDK_THREADING_CLASS_PREFIXES:""}
```
wu-sheng: I think as an optional plugin, this is not recommended to add here.

wu-sheng:
I think we should write a blog(en) about this feature on the official website, with your demo codes and screenshot. I think people will love this. Including this will work with many thread pool, right?

apm-sniffer/config/agent.config
kezhenxu94: > I think as an optional plugin, this is not recommended to add here.

I was thinking that since it’s not activated by default, it does no harm to put the configuration here, I can comment it out if needed

kezhenxu94:
> I think we should write a blog(en) about this feature on the official website, with your demo codes and screenshot. I think people will love this. Including this will work with many thread pool, right?

Sure, I’ll try if I can get some time

apm-sniffer/config/agent.config
wu-sheng: But we have dozens of configs, if we don't this here, why don't we put others. I think currently, the principle is putting minimal config here, to ease the first when they are new to SW.

kezhenxu94: > But we have dozens of configs if we put this here, why don't we put others. I think currently, the principle is putting minimal config here, to ease the first when they are new to SW.

Make sense to me

wu-sheng:
@kezhenxu94 You got an error in your test case, could you recheck?

kezhenxu94:
> @kezhenxu94 You got an error in your test case, could you recheck?

because of the removal of config item in `agent.config`, checking

PR #4068, opened 2019-12-15
Add more users and most are very famous

wu-sheng:
This full list came from the Survey about the `End the supports of JDK 1.6`. End users submit the company info in the Survey.

### DIFF
--- a/docs/powered-by.md
+++ b/docs/powered-by.md
@@ -23,7 +23,7 @@ or providing commercial products including Apache SkyWalking.
 1. Chehejia Automotive. 北京车和家信息技术有限责任公司. https://www.chehejia.com/
 1. Chinadaas. 北京中数智汇科技股份有限公司. https://www.chinadaas.com/
 1. China Merchants Bank. 中国招商银行. http://www.cmbchina.com/
-1. China Mobile CMSS. 中国移动苏州研发中心.
+1. China Mobile 中国移动
 1. China Mobile 中国联通
 1. China Tower 中国铁塔
 1. Chtwm.com. 恒天财富投资管理股份有限公司. https://www.chtwm.com/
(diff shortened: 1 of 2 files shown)


docs/powered-by.md
```diff
@@ -7,32 +7,47 @@ or providing commercial products including Apache SkyWalking.
...
 1. Chehejia Automotive. 北京车和家信息技术有限责任公司. https://www.chehejia.com/
+1. China Eastern Airlines 中国东方航空 http://www.ceair.com/
+1. China Express Airlines 华夏航空 http://www.chinaexpressair.com/
 1. Chinadaas. 北京中数智汇科技股份有限公司. https://www.chinadaas.com/
+1. Chinasoft International 中软国际
 1. China Merchants Bank. 中国招商银行. http://www.cmbchina.com/
-1. China Mobile. CMSS. 中国移动苏州研发中心.
+1. China National Software 中软
+1. China Mobile 中国移动
+1. China Mobile 中国联通
```
kezhenxu94: typo

wu-sheng: Good catch. Fixed.

docs/powered-by.md
```diff
@@ -7,32 +7,47 @@ or providing commercial products including Apache SkyWalking.
...
-1. China Eastern Airlines 中国东方航空. http://www.ceair.com/
 1. Chehejia Automotive. 北京车和家信息技术有限责任公司. https://www.chehejia.com/
+1. China Eastern Airlines 中国东方航空 http://www.ceair.com/
+1. China Express Airlines 华夏航空 http://www.chinaexpressair.com/
 1. Chinadaas. 北京中数智汇科技股份有限公司. https://www.chinadaas.com/
+1. Chinasoft International 中软国际
 1. China Merchants Bank. 中国招商银行. http://www.cmbchina.com/
-1. China Mobile. CMSS. 中国移动苏州研发中心.
+1. China National Software 中软
+1. China Mobile 中国移动
```
wayilau: We are CMsoft. But belongs to ChinaMobile. so...

wu-sheng: > We are CMsoft. But belongs to ChinaMobile. so

It is very hard to maintain a department level user list, if basically, it is used, I hope this is good enough.

wu-sheng: Like Ping An, Minsheng Bank, all tech or platform departments are using. Because this is not a commercial deal :) 
So all entities here, we list as the company level. it is hard and impossible to share which system, right? And also, I think even we can, we shouldn't open that detail unless the owner of the system wants to, such as at big conference or blog.

wayilau: Yes, It's difficult to  Subdivision. I think it's good enough.

Issue #3752, opened 2019-10-30
Add gRPCHook to alarm-settings.yml


wu-sheng:
Add gRPCHooks to the alarm-settings.yml, which will automatically work for the dynamic update.

The part of setting filw should look like
```yaml
webhooks:
#  - http://127.0.0.1/notify/
#  - http://127.0.0.1/go-wechat/

gRPCHook: ip:port
```

As widely used gRPC protocol, supporting gRPC is more friendly to the 3rd party integration.

wu-sheng:
@JaredTan95 Do you have time to add this in these two days? As @x22x22 should submit UI update in this weekend, we should have 4-5 days to land this feature. If you are too busy, we could move this to 6.6.

Please confirm.

wu-sheng:
I am thinking about `webhooks` should keep one only, but for compatible, we could keep it unchanged. @kezhenxu94 Only idea about this?

kezhenxu94:
> I am thinking about `webhooks` should keep one only, but for compatible, we could keep it unchanged. @kezhenxu94 Only idea about this?

Agree to keep it only one, but it's trivial and incompatible, it should be better to let it be

JaredTan95:
> @JaredTan95 Do you have time to add this in these two days? As @x22x22 should submit UI update in this weekend, we should have 4-5 days to land this feature. If you are too busy, we could move this to 6.6.
> 
> Please confirm.

Move to 6.6.0 much better for me.

JaredTan95:
low priority in my view. Move to the next milestone. :-P

JaredTan95:
provided.

PR #4072, opened 2019-12-16
support min function.

JaredTan95:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
support Minimum aggregation and fix some unit test case.
___
### Bug fix
- Bug description.

- How to fix?

___

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/MinLongMetrics.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/MinLongMetrics.java
@@ -34,14 +34,10 @@ public abstract class MinLongMetrics extends Metrics implements LongValueHolder 
 
     protected static final String VALUE = "value";
 
-    @Getter @Setter @Column(columnName = VALUE, isValue = true) private long value;
+    @Getter @Setter @Column(columnName = VALUE, isValue = true) private long value = Long.MAX_VALUE;
 
     @Entrance
     public final void combine(@SourceFrom long count) {
-        if (this.value == 0) {
-            this.value = count;
-        }
-
         if (count < this.value) {
             this.value = count;
         }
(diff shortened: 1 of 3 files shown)


wu-sheng:
First, could you share why adding this func?

JaredTan95:
> First, could you share why adding this func?

simplely want to aggregation some metrics as follows:

```shell
service_latency_min = from(Service.latency).min();
endpoint_latency_min = from(Endpoint.latency).min();
```

wu-sheng:
Interesting, but make sense in some cases.

JaredTan95:
Yeah, and we also got `endpoint_latency_max` metrics by `Max` function, but found there is no `Min` function.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/MinLongMetrics.java
```diff
@@ -0,0 +1,57 @@
...
+
+    protected static final String VALUE = "value";
+
+    @Getter @Setter @Column(columnName = VALUE, isValue = true) private long value;
+
+    @Entrance
+    public final void combine(@SourceFrom long count) {
+        if (this.value == 0) {
+            this.value = count;
+        }
```
kezhenxu94: An obvious bug here, try the test codes:

```java
        MinLongMetricsImpl impl = new MinLongMetricsImpl();
        impl.combine(10);
        impl.combine(0);
        impl.combine(10000);
```

I expect `impl.getValue()` to be `0`, but it's `10000`, please recheck @JaredTan95

kezhenxu94: Making `value` default to `Long.MAX_VALUE` is my first intuition

kezhenxu94: > An obvious bug here, try the test codes:
> 
> ```java
>         MinLongMetricsImpl impl = new MinLongMetricsImpl();
>         impl.combine(10);
>         impl.combine(0);
>         impl.combine(10000);
> ```
> 
> I expect `impl.getValue()` to be `0`, but it's `10000`, please recheck @JaredTan95

I suggest putting this into unit test as well

JaredTan95: Thanks to your experienced guidance, I will update it.

wu-sheng:
@JaredTan95 Please take care of #3752

PR #4074, opened 2019-12-16
[CI] Add IT/E2E check statuses to index page

kezhenxu94:
I happened to find that several builds on master has failed and just re-run them, this patch exposes the check statues in the main page to make them more easily to be noticed, and recommend committers/PMCs to enable the notifications for the checks in case the master branch is broken by any chances and can be discovered more quickly, hope to find a GitHub Action to send email when builds failed (not found yet)

image


wu-sheng:
I could receive some mails when builds fail.

PR #4075, opened 2019-12-17
Fix service instance relationship ID build bug.

wu-sheng:
The existing and unused logic of service instance relationship ID includes the component ID, which is not suitable for today's topology logic. But as this is never used before now, it isn't been found.

This causes all metrics related to `service instance relationship`, includes the component id in the entity ID too. Then UI can't use the source+target IDs to get the metrics result.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/ServiceInstanceRelation.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/ServiceInstanceRelation.java
@@ -38,11 +38,11 @@ public class ServiceInstanceRelation extends Source {
         return String.valueOf(sourceServiceInstanceId) + Const.ID_SPLIT + String.valueOf(destServiceInstanceId);
     }
 
-    @Getter @Setter private int sourceServiceInstanceId;
+    @Getter @Setter @ScopeDefaultColumn.DefinedByField(columnName = "source_service_instance_id") private int sourceServiceInstanceId;
     @Getter @Setter @ScopeDefaultColumn.DefinedByField(columnName = "source_service_id") private int sourceServiceId;
     @Getter @Setter private String sourceServiceName;
     @Getter @Setter private String sourceServiceInstanceName;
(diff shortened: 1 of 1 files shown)


wu-sheng:
FYI @kezhenxu94 @JaredTan95

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/ServiceInstanceRelation.java
```diff
@@ -35,14 +35,14 @@
     }
 
     @Override public String getEntityId() {
-        return String.valueOf(sourceServiceInstanceId) + Const.ID_SPLIT + String.valueOf(destServiceInstanceId) + Const.ID_SPLIT + String.valueOf(componentId);
+        return String.valueOf(sourceServiceInstanceId) + Const.ID_SPLIT + String.valueOf(destServiceInstanceId);
```
wu-sheng: @arugal Remove the component id.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/ServiceInstanceRelation.java
```diff
@@ -35,14 +35,14 @@
...
     }
 
-    @Getter @Setter private int sourceServiceInstanceId;
+    @Getter @Setter @ScopeDefaultColumn.DefinedByField(columnName = "source_service_instance_id") private int sourceServiceInstanceId;
     @Getter @Setter @ScopeDefaultColumn.DefinedByField(columnName = "source_service_id") private int sourceServiceId;
     @Getter @Setter private String sourceServiceName;
     @Getter @Setter private String sourceServiceInstanceName;
-    @Getter @Setter private int destServiceInstanceId;
+    @Getter @Setter @ScopeDefaultColumn.DefinedByField(columnName = "dest_service_instance_id") private int destServiceInstanceId;
```
wu-sheng: Add these two columns to the metrics entity. Right now, you only could see service(source and dest) IDs in there.

PR #4093, opened 2019-12-20
[test/plugin] immigrate test shardingsphere 3.x scenario. 

tuohai666:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/3583
___
### Bug fix
- Bug description.

- How to fix?

___


.github/workflows/plugins-test.yaml
```diff
@@ -504,3 +504,28 @@ jobs:
...
+            ${{ runner.os }}-maven-
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 8
+      - name: Build SkyWalking Agent
+        run: ./mvnw clean package -DskipTests -Pagent >/dev/null
+      - name: Build the Docker image
+        run: ./mvnw -f test/plugin/pom.xml clean package -DskipTests docker:build -DBUILD_NO=local >/dev/null
+      - name: Run shardingsphere-3.x-scenario 3.0.0 (1)
+        run: bash test/plugin/run.sh shardingsphere-3.x-scenario
```
dmsolr: I think we can add it to the fastest group. Because it has one version only.

test/plugin/scenarios/shardingsphere-3.x-scenario/src/main/java/org/apache/skywalking/apm/testcase/shardingsphere/service/api/service/CommonServiceImpl.java
```diff
@@ -0,0 +1,120 @@
...
+    public void cleanEnvironment() {
+        getOrderRepository().dropTable();
+        getOrderItemRepository().dropTable();
+    }
+    
+    @Override
+    public void processSuccess(final boolean isRangeSharding) {
+        System.out.println("-------------- Process Success Begin ---------------");
+        printData(isRangeSharding);
+        System.out.println("-------------- Process Success Finish --------------");
```
dmsolr: We don't need those. Test-framework does not get `SystemOut` log, so those are unuseful.

.github/workflows/plugins-test.yaml
kezhenxu94: @tuohai666 more specifically, `Oracle_Kafka_JdkHttp_JdkThreading` is the fastest group now, moving to this group should be fine

test/plugin/scenarios/shardingsphere-3.x-scenario/src/main/java/org/apache/skywalking/apm/testcase/shardingsphere/service/api/service/CommonServiceImpl.java
tuohai666: Git it.

.github/workflows/plugins-test.yaml
tuohai666: Thanks for your  comments. Can you tell me what is the fastest gropu? I can find this key word in doc, but not quite understand:

Please check the prev agent related PRs, and add your case to the fastest group, in order to make the whole test finished as soon as possible.

wu-sheng: Simply, check this

image

Or more details at this page, https://github.com/apache/skywalking/pull/4093/checks?check_run_id=358026747

kezhenxu94: > Thanks for your comments. Can you tell me what is the fastest gropu? I can find this key word in doc, but not quite understand:
> 
> Please check the prev agent related PRs, and add your case to the fastest group, in order to make the whole test finished as soon as possible.

You can find the time costs for each group in the check statuses in this page, see attached picture:

image

tuohai666: Thanks for your quick response.

test/plugin/scenarios/shardingsphere-3.x-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,218 @@
...
+            endTime: not null
+            componentId: 1
+            componentName: ''
+            isError: false
+            spanType: Entry
+            peer: ''
+            peerId: 0
+            tags:
+              - {key: url, value: 'http://localhost:8080/shardingsphere-3.x-scenario/case/healthCheck'}
+              - {key: http.method, value: HEAD}
```
kezhenxu94: `healthCheck` is not needed (in both `operationNames` and `segments`

.github/workflows/plugins-test.yaml
```diff
@@ -318,6 +318,8 @@ jobs:
         run: ./mvnw -f test/plugin/pom.xml clean package -DskipTests docker:build -DBUILD_NO=local >/dev/null
       - name: Run redisson 3.x (37)
         run: bash test/plugin/run.sh redisson-scenario
+      - name: Run shardingsphere-3.x-scenario 3.0.0 (1)
```
wu-sheng: You miss the change of group name. `Redisson` -> `Redisson_ShardingSphere`

tuohai666: Get it.

Issue #4087, opened 2019-12-19
when i use skywalking, i founded kafkaProcuctor will degrade performance very much


adamni135:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?
when i use kafka producer.send(new ProducerRecord<String, byte[]>(topic, msg), callback);
callback must not be null

we found lots of threads waiting at org.apache.kafka.clients.producer.internals.BufferPool#allocate line134
codes: waitingTimeElapsed = !moreMemory.await(remainingTimeToBlockNs, TimeUnit.NANOSECONDS);

when we delete org.apache.skywalking.apm.plugin.kafka.CallbackInterceptor#beforeMethod
line43-48 then our springboot application performance will be as usual
___
### Bug
- Which version of SkyWalking, OS and JRE?
kafka maven:
org.apache.kafka kafka_2.12 2.2.0
skywalking plugin:
apm-kafka-plugin-6.5.0.jar

- Which company or project?

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.
when using kafka sender with callback will degrade performance very much
___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

wu-sheng:
> when we delete org.apache.skywalking.apm.plugin.kafka.CallbackInterceptor#beforeMethod
line43-48 then our springboot application performance will be as usual

These lines are making sure your trace wouldn't break when callback executed in another thread. I t looks normal to me.

 I am not following, why is the following block related to those codes? Could you provide more detail?
>codes: waitingTimeElapsed = !moreMemory.await(remainingTimeToBlockNs, TimeUnit.NANOSECONDS);

ping @ascrutae @stalary

adamni135:
our team is finding the reason too, we check the JVM safepoint log, we found that safe point reduce many, that means is not the safe point cause
if we start the application without that kafka plugin the qps is about 10000，when we add the plugin, the qps is 800
now I dump the theads info from vm when it work with kafka plugin
90% theads is waiting at that lines

"XNIO-1 task-214" #314 prio=5 os_prio=0 tid=0x00007f0e4402d050 nid=0x376b waiting on condition [0x00007f0c5c3c1000]
   java.lang.Thread.State: TIMED_WAITING (parking)
	at sun.misc.Unsafe.park(Native Method)
[...]

PR #4097, opened 2019-12-21
fix file writer block application thread

adamni135:

- [ ] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/4087
___
### Bug fix
- Bug description.
when the log queue is full, File writer will block the business thread 2 millis, this will degrade performance very much

- How to fix?
using offer without waiting and print error log
___


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/FileWriter.java
```diff
@@ -72,6 +73,10 @@ private FileWriter() {
                         for (String log : outputLogs) {
                             writeToFile(log + Constants.LINE_SEPARATOR);
                         }
+                    	if(isQueueFull) { // system error print when the queue is full
+                    		System.err.println("file log queue is full.");
```
wu-sheng: Take a look at what is system.err. There is a sync lock there.

PR #4099, opened 2019-12-21
write file unwait

adamni135:

- [x] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/4087
___
### Bug fix
- Bug description.
when the log queue is full, File writer will block the business thread 2 millis, this will degrade performance very much
- How to fix?
using offer without waiting
___


adamni135:
fix file writer block application thread

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/FileWriter.java
```diff
@@ -212,15 +212,11 @@ private boolean prepareWriteStream() {
...
      *
      * @param message to log
      */
     @Override public void write(String message) {
-        try {
-            logBuffer.offer(message, 2, TimeUnit.MILLISECONDS);
-        } catch (InterruptedException e) {
-            e.printStackTrace();
-        }
+        logBuffer.offer(message);
```
wu-sheng: After investing the JDK source codes, this will cause more lock time than original. Because this will never release lock until `offer` successfully.

dmsolr: `queue.offer(element)` will insert when the queue has free space. If not free space, the element will be discarded and release the lock immediately.
As you description, I think you say the operator is `queue.put(element)`.

wu-sheng: I just rechecked, it seems yes. I misread the codes last time. Then this PR should be good enough.

But as 6.6.0 codes have been frozen, this will only be merged after 6.6.0 release.

PR #4111, opened 2019-12-23
[Feature] Provide Java Agent for Apache HttpClient 3.x

kezhenxu94:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance


.github/workflows/plugins-test.yaml
```diff
@@ -264,6 +264,8 @@ jobs:
         run: bash test/plugin/run.sh spring-4.1.x-scenario
       - name: Run solrj 7.x (12)
         run: bash test/plugin/run.sh solrj-7.x-scenario
+      - name: Run httpclient 2.0-3.1 (5)
+        run: bash test/plugin/run.sh httpclient-3.x-scenario
```
wu-sheng: You miss changing the Group name

kezhenxu94: > You miss changing the Group name

The group name seems not to be related to versions (`Spring41x_SolrJ-Http`), what are you expecting me to change

wu-sheng: Oh, this is just version related. Ignore this.

PR #4116, opened 2019-12-24
keep the properties in SnifferConfigInitializer

ZhuoSiChen:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?
When I develop my plugin I find skywalking no way to get my properties
___


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
```diff
@@ -40,7 +40,7 @@
     private static String DEFAULT_CONFIG_FILE_NAME = "/config/agent.config";
     private static String ENV_KEY_PREFIX = "skywalking.";
     private static boolean IS_INIT_COMPLETED = false;
-
+    private static Properties properties;
```
wu-sheng: Why change this? Who will use this properties? If not, this uses memory and gc load.

PR #4117, opened 2019-12-24
Keep properties

ZhuoSiChen:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
```diff
@@ -192,4 +192,11 @@ private static InputStreamReader loadConfig() throws AgentPackageNotFoundExcepti
         }
         throw new ConfigNotFoundException("Failed to load agent.config.");
     }
+
+    /**
+     * return the properties read from DEFAULT_CONFIG_FILE_NAME
+     */
+    public static Properties getProperties(){
```
wu-sheng: Why open this? This is not an API.

ZhuoSiChen: Because I find no way to get my plugin propeties.

ZhuoSiChen: when I want to do some properties in my plugin. I find it hard to do that. So I open the this。

wu-sheng: No, this is a wrong way. If you want to support properties to inject values, there should be an initial mechanism for plugin properties intialization. 

Such as, having a service in the core, and plugin could initial its config object by a callback register. But, notice, this should be called once.

We should have deep discussion about this.

ZhuoSiChen: I just want something like the way skywalking doing in the  `org.apache.skywalking.apm.agent.core.conf.Config` and used it in any where.If I can get the properties. it was easy to me.

wu-sheng: No. This is not right for an open source project. It is reasonable to support the config object in the plugin, But the way you wrote is not right.

wu-sheng:
Any update? Do you know how to do this?

ZhuoSiChen:
I  still unknow how to do take this PR. 
My solution was init the configuration in My auto-instrumentation Constructor
But the configuration still can't used in my inteceptor,
Because the configuration and the interceptor  load by the different ClassLoader.

wu-sheng:
This initialization should be triggered by the interceptor initialization stage.

ZhuoSiChen:
You mean this code
 `SnifferConfigInitializer.initialize(agentArgs);`,
but this is hard code Can't extend？

wu-sheng:
I am answering the question about you said the two class loaders. 

`SnifferConfigInitializer.initialize(agentArgs)`. This could be changed, my point is, you shouldn't like this PR, change it for the plugin developer.

PR #4119, opened 2019-12-24
Update JDK compiling level to 8, and remove zip dist.

wu-sheng:
This change based on the PMC and community vote. All agents in 7.x will only consider JDK8 compatible. 

And I removed the zip dist from 7.x, as the release tarballs are too many and too large without benefit.

@dmsolr @arugal  Do any of you have time to move optional plugins caused by JDK8 requirements into the default plugins, and change the plugin test scripts? I think inducing
1. Lettuce
1. Play Framework
1. Spring Webflux 5.x


wu-sheng:
I am fixing the more deps in jar. They need to be shaded.

image

oap-server/server-cluster-plugin/cluster-kubernetes-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/kubernetes/KubernetesCoordinator.java
```diff
@@ -60,27 +60,29 @@
     }
 
     public void start() {
-        submitTask(MoreExecutors.listeningDecorator(Executors.newSingleThreadExecutor(new ThreadFactoryBuilder()
-            .setDaemon(true).setNameFormat("Kubernetes-ApiServer-%s").build())));
+        ExecutorService executorService = Executors.newSingleThreadExecutor(new ThreadFactoryBuilder()
+            .setDaemon(true).setNameFormat("Kubernetes-ApiServer-%s").build());
+        submitTask(MoreExecutors.listeningDecorator(executorService), executorService);
```
wu-sheng: @hanahmily I changed this due to Guava lib upgrade. Could you recheck this?

Issue #4121, opened 2019-12-25
Move Lettuce/Play Framework/Spring Webflux 5.x/Armeria plugins into default list


wu-sheng:
By following #4119, JDK8 is default JDK level, so there three plugins are safe to be in the default plugin list.

wu-sheng:
Armeria should be moved too

kezhenxu94:
moved

PR #4130, opened 2019-12-25
Move Armeria plugin to default plugin list

kezhenxu94:


kezhenxu94:
> Require change on supported list doc.

Supported list doc already contains all the versions

wu-sheng:
Oh, it missed the `Optional 2` in the first place. All optional should have that.

wu-sheng:
Related to #4121

PR #4132, opened 2019-12-26
Add E2E test case to verify compatibility between 6.x agents and 7.x …

kezhenxu94:
…backend

We don't have release 6.6.0 now, so I use agents in release 6.5.0


.github/workflows/e2e.yaml
```diff
@@ -50,6 +50,8 @@ jobs:
         run: export E2E_VERSION=jdk12-1.3 && bash -x test/e2e/run.sh e2e-single-service
       - name: Agent Reboot Tests(JDK8)
         run: export E2E_VERSION=jdk8-1.3 && bash -x test/e2e/run.sh e2e-agent-reboot
+      - name: Compatibility (6.x Agents & 7.x Backend)
```
wu-sheng: Let's create a new pipeline for this.

test/e2e/e2e-6.x-agent-7.x-oap-compatibility/src/docker/rc.d/rc0-prepare.sh
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+AGENT_VERSION=6.5.0
+
+ls /sw && echo 'Remove SkyWalking agent directory' && rm -rf /sw/agent && ls /sw
+
+echo 'Download SkyWalking 6.x...' \
+  && curl -o /tmp/sw.tar.gz https://dist.apache.org/repos/dist/release/skywalking/${AGENT_VERSION}/apache-skywalking-apm-${AGENT_VERSION}.tar.gz \
```
wu-sheng: Use dist from here, https://archive.apache.org/dist/skywalking/. 

`dist.apache.org` is a mirror site, the release could be removed in some day.

Issue #3777, opened 2019-11-04
Provide plugin for ShardingSphere 4.0.0-RC3


tuohai666:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [x] Requirement
- [ ] Feature or performance improvement

Provide plugin for ShardingSphere 4.0.0-RC3 of which the API is different from RC1 or RC2.

wu-sheng:
You should have witness classes for different version libs.

Also, shardingsphere test case should be moved. @tristaZero are you going to do this?

tuohai666:
I'll do this after 4.0.0-RC3 released.

wu-sheng:
Thanks @tuohai666 to provide this. Is this plugin working for 4.0.0 stable release too?

tuohai666:
Hi @wu-sheng, this plugin working only for 4.0.0-RC3. There're package name changes in 4.0.0 stable release.

wu-sheng:
@tuohai666 :( Hope the ShardingSphere could be more stable. It keeps changing in every release.

PR #4140, opened 2019-12-27
[Agent] Provide plugin for ShardingSphere 4.0.0-RC3

tuohai666:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#3777

___
### Bug fix
- Bug description.

- How to fix?

___


tuohai666:
I submit the plugin and its test case together, is it right?
It seems like the plugin didn't take effect when PluninsTest running.

wu-sheng:
What do you mean not effect?

tuohai666:
I've modified the interface parseSQL according to ShardingSphere 4.0.0-RC3. But in the PluginsTest there's no parseSQL segment.

tuohai666:
I follow this procedure locally and the test passed:
1. ./mvnw clean package -Pagent,dist
2. ./test/plugin/run.sh -f shardingsphere-4.x-RC3-scenario

wu-sheng:
The comparison logs are available here, https://github.com/apache/skywalking/pull/4140/checks?check_run_id=364942493. Do you check it?

tuohai666:
Yes, I checked this at the time when the test failed.

wu-sheng:
Clearly, some DB execution happens first, check this. 

image

tuohai666:
I can't see the picture. I think it doesn't matter about the order.

wu-sheng:
Definitely it matters. Span order is a part of validation. In the logs, you can see, the `close` happens first.

tuohai666:
Your findings are not in the expectedData.yaml.

wu-sheng:
Yes, so the validation fails. It exists in the actual data.

wu-sheng:
Even I re-run the jobs, it seems still fail. Please recheck.

tuohai666:
From the log of ci, I found PluginsTest / Redisson_ShardingSphere_Zookeeper (9:10) happened before CI AND IT / CI (9:36).
If the plugin for 4.0.0-RC3 haven't been packaged before test, the test case for 4.0.0-RC3 will fail like what logs show.

wu-sheng:
No, they are separated tasked. The agent tests have include the package, you could see that in the GitHub action YAML file

tuohai666:
I prefer check the log or the artifact rather than YAML file. If the test includes the new plugin, where is it?
image

wu-sheng:
What do you mean where is it? It is using the codes of your branch. This is a CI environment, it is not a VM.

dmsolr:
The log showed the reason for the failure. 

> 2019-12-30T01:10:46.3205524Z assert failed.
> 2019-12-30T01:10:46.3206119Z SegmentNotFoundException:
> 2019-12-30T01:10:46.3206501Z expected: 
> 2019-12-30T01:10:46.3206842Z   Segment:
> 2019-12-30T01:10:46.3207648Z   - span[0, 1] /ShardingSphere/parseSQL/
> 2019-12-30T01:10:46.3208434Z   - span[3, 4] H2/JDBI/PreparedStatement/executeQuery
> 2019-12-30T01:10:46.3209294Z   - span[2, 3] /ShardingSphere/executeSQL/
> 2019-12-30T01:10:46.3210410Z   - span[0, 2] /ShardingSphere/JDBCRootInvoke/
> 2019-12-30T01:10:46.3210802Z   - span[-1, 0] /shardingsphere-4.x-RC3-scenario/case/execute
> 2019-12-30T01:10:46.3210874Z 
> 2019-12-30T01:10:46.3235451Z actual: 
> 2019-12-30T01:10:46.3235774Z

apm-sniffer/apm-sdk-plugin/sharding-sphere-4.x-RC3-plugin/src/main/java/org/apache/skywalking/apm/plugin/shardingsphere/v4/JDBCRootInvokeInterceptor.java
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @author zhangyonglun
+ */
+public class JDBCRootInvokeInterceptor implements InstanceMethodsAroundInterceptor {
+    
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, MethodInterceptResult result) {
+        ContextManager.createLocalSpan("/ShardingSphere/JDBCRootInvoke/").setComponent(ComponentsDefine.SHARDING_SPHERE);
+        ShardingExecuteDataMap.getDataMap().put(Constant.CONTEXT_SNAPSHOT, ContextManager.capture());
+    }
```
dmsolr: I think we have to do `across thread propagation`.

tuohai666:
@dmsolr That is not the root cause. Since you are here, please show me where's the sharding-sphere-4.x-RC3-plugin.jar.
The new version of plugin built after the test finish. Sheng said the jar will be separately built in test phase that I didn't find.

apm-sniffer/apm-sdk-plugin/sharding-sphere-4.x-RC3-plugin/src/main/java/org/apache/skywalking/apm/plugin/shardingsphere/v4/ParseInterceptor.java
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @author zhangyonglun
+ */
+public class ParseInterceptor implements InstanceMethodsAroundInterceptor {
+    
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, MethodInterceptResult result) {
+        AbstractSpan span = ContextManager.createLocalSpan("/ShardingSphere/parseSQL/").setComponent(ComponentsDefine.SHARDING_SPHERE);
+        Tags.DB_STATEMENT.set(span, (String) allArguments[0]);
+    }
```
dmsolr: It is the same. And other LocalSpan(s).

dmsolr:
> That is not the root cause. Since you are here, please show me where's the sharding-sphere-4.x-RC3-plugin.jar.

`skywalking/skywalking-agent/plugins/`

> The new version of plugin built after the test finish. Sheng said the jar will be separately built in test phase that I didn't find.

Sorry, I don't see what you mean.

wu-sheng:
@dmsolr Yes. I read those too, that is why I am confused about the @tuohai666 's doubts.

tuohai666:
@wu-sheng sharding-sphere-4.x-RC3-plugin.jar happened after the test finish. To make the test success, the sharding-sphere-4.x-RC3-plugin.jar need to be packaged before test. How PluginsTest package plugins? I didn't find it in log.

kezhenxu94:
> @wu-sheng sharding-sphere-4.x-RC3-plugin.jar happened after the test finish. To make the test success, the sharding-sphere-4.x-RC3-plugin.jar need to be packaged before test. How PluginsTest package plugins? I didn't find it in log.

@tuohai666 every plugin test compiles and builds the plugin, check https://github.com/apache/skywalking/blob/6af14a36bcaa94741cd1a2709ba3193d5f764b9c/.github/workflows/plugins-test.yaml#L43 so don’t worry about the plugin package, it must exist absolutely, otherwise all plugin should fail from time to time

dmsolr:
> @wu-sheng sharding-sphere-4.x-RC3-plugin.jar happened after the test finish. To make the test success, the sharding-sphere-4.x-RC3-plugin.jar need to be packaged before test. How PluginsTest package plugins? I didn't find it in log.

I think you have a mistake. The first, we find it in the log, but it is a bit difficult because of too long.
And I don't think the test-framework has the issue which you said.
And then, the log shows so clearly, the plugin has worked. We can it is generated `actualData.yml` which means the plugin working.

kezhenxu94:
@tuohai666 The full qualified class names of the new plugin instrumentation classes (and interceptors) are exactly the same as the old ones, I’m not sure whether it’s related to this issue and whether it’s stably reproducible in local machine(at least I can’t reproduce), but it definitely matters in some degree, I don’t think JVM can load two same classes using the same class loader (AgentClassloader), I suspect that some of them are “hidden” by the old ones, could you recheck?

tuohai666:
@kezhenxu94 Thanks for your patience. I'll think about that.

tuohai666:
@kezhenxu94 At the same time, can you do me a favor to test my repo in your local? I cant pass the test in my local, now it's not enough.

kezhenxu94:
> @kezhenxu94 At the same time, can you do me a favor to test my repo in your local? I cant pass the test in my local, now it's not enough.

😂 @tuohai666  if you can not pass the test locally, then you should debug and check locally first, I assumed (mistakenly) that it passed on your local machine

tuohai666:
Typo, I mean I can... 
I found something, let me have a test first.

kezhenxu94:
> I suspect that some of them are “hidden” by the old ones, could you recheck?

@tuohai666 I just rechecked the codes, the classes with same names are mostly identical, so this is not the cause, but maybe we need to refactor the ShardingSphere plugins to extract some codes to be reused, maybe another PR after this gets merged. WDYT @wu-sheng

tuohai666:
@kezhenxu94 Yes, you are right, that will be better.

I checked my codes again and the test suceed again in local. Can you help to test in my repo? Maybe we can find something.

kezhenxu94:
> @kezhenxu94 Yes, you are right, that will be better.
> 
> I checked my codes again and the test suceed again in local. Can you help to test in my repo? Maybe we can find something.

I've just run it locally, and it passed too

wu-sheng:
Two things, 1. Maybe relate to thread, different JVM has different thread model. 2. JDBC connection. It may close before the next execution.

As far as I known, ss includes its own thread pool. So, be careful about this.

dmsolr:
Agree with you , @wu-sheng 

I run it on centos7. It is failed. And Github action runs on Ubuntu-18. I think both of you run on macOS.

image

tuohai666:
@dmsolr Thanks for your help. I will firstly change to openjdk and then find a Linux machine to find out the root cause.

tuohai666:
I have reproduced this problem in CentOS 7.6.
@kezhenxu94 I think your judgement is right. Classes are "hidden" in platforms other than MacOS, I'll change the package name.

tuohai666:
@kezhenxu94 You are so smart. Thanks.

PR #4144, opened 2019-12-28
Fix os maven plugin issues for Eclipse m2e

hanahmily:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

When building skywalking in Eclipse or Vscode with Java plugin, we will encounter some issues. 

- How to fix?

Refer to https://github.com/trustin/os-maven-plugin#issues-with-eclipse-m2e-or-other-ides, to add this plugin to build lifecycle instead of using it as an extension.

### DIFF
--- a/apm-protocol/apm-network/pom.xml
+++ b/apm-protocol/apm-network/pom.xml
@@ -58,6 +58,19 @@
 
         <plugins>
             <plugin>
+                <groupId>kr.motd.maven</groupId>
+                <artifactId>os-maven-plugin</artifactId>
+                <version>${os-maven-plugin.version}</version>
+                <executions>
+                    <execution>
+                        <phase>initialize</phase>
+                        <goals>
+                            <goal>detect</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
                 <groupId>org.xolstice.maven.plugins</groupId>
                 <artifactId>protobuf-maven-plugin</artifactId>
                 <version>${protobuf-maven-plugin.version}</version>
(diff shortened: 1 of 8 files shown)


wu-sheng:
Why do we pay attention on Eclipse now? Even no one maintain the compiling doc for eclipse.

wu-sheng:
Why do we pay attention on Eclipse now? Even no one maintain the compiling doc for eclipse.

wu-sheng:
And you should know, the current way is the way grpc-java provided officially.

wu-sheng:
After reading the doc you referred, clearly there is a way rather than changing the project files. I highly prefer don't do this, as no committer or PMC will work on keeping things checked, and no way to verify this by CI.

hanahmily:
> Why do we pay attention on Eclipse now? Even no one maintain the compiling doc for eclipse.

I'm switching my workspace from idea local development to vscode remote development(performance concerns and network issue), vscode uses m2e(eclipse plugin) to build java, so I should fix some compiling issues about it.

> After reading the doc you referred, clearly there is a way rather than changing the project files. I highly prefer don't do this, as no committer or PMC will work on keeping things checked, and no way to verify this by CI.

Yes, there's another way to do it. but if we merge this PR, other developers using Vscode and Eclipse will get benefits. And I will take care of those changes since I'm working with vscode.

As you have reminded, the old way is from grpc-java official example, but my solution is also from mave-os-plugin official suggestion. I think if these changes passed CI test, we don't have any concerns about production release, especially our most committers are using idea as their ide, so if there's something wrong for the development environment, we will find it soon.

Issue #4104, opened 2019-12-22
[FEATURE] Performance profile at method level in production environment


mrproliu:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [x] Feature or performance improvement

___
I wrote about the thread monitoring proposal in the mail list.
The doc is now complete to introduce this feature: https://docs.google.com/document/d/1rxMf1WN3PaFaZp7r8JmtwfdkmjLTcFW_ETAZv5FIU-s/edit

Because the amount of development is relatively large, I will divide the following steps into the development work:

Step 1: Save thread monitoring tasks to storage and corresponding graphQL
1. Define thread monitoring DAO, graphql, mainly including adding task interface
2. Implementing JDBC storage
3. Implementing Elasticsearch storage
4. Implementing Elastic-search 7 storage

Step 2: Downstream task to sniffer
1. Define the cache interface (CacheDAO), which can be used to query the cached data. Implementing JDBC storage
2. Complete queries in Elasticsearch and Elasticsearch 7
3. Create a CacheService and start a scheduled task to write data to the cache
4. Define the protocol file (query command) for OAP to communicate with the sniffer, and implement the gRPC server handler

Step 3: sniffer executes commands to monitor traces
1. Start timing tasks for command reading, and define the gRPC protocol (for snapshot upload)
2. Monitor trace generation and report data
3. Write Testcase (JUnit) to ensure that snapshots can be created and reported after receiving the command

Step 4: OAP saves thread snapshot
1. Define the receive module, define the interface for adding snapshots in the DAO, and receive the data save request passed by the sniffer
2. Implementing JDBC storage
3. Implementing Elasticsearch storage
4. Implementing Elastic-search 7 storage
5. Write Testcase, save operation based on received data

Step 5: Query data and display graphql
1. Define query interface and graphql for querying task list, related trace list, and snapshot
2. Implementing JDBC storage
3. Implementing Elasticsearch storage
4. Implementing Elastic-search 7 storage
5. Complete test code (e2e-cluster-test-runner)

wu-sheng:
Branch `profile` will be created after 6.6.0 release. PRs of all steps should send to there, after all features tested and e2e test setup, I will submit a `un-squash` merge PR to the master.

This is a very important feature, @apache/skywalking-committers, welcome and look forward all of you could take part in and help on review. Design doc should be ready.

wu-sheng:
[...]

PR #4145, opened 2019-12-29
Provide thread monitor create task feature

mrproliu:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#4104 
___
- Follow step one, create and query thread monitor task according to graphql


dmsolr:
This is for step 1?

wu-sheng:
The e2e tests are keeping failure, we need to figure out why first.

wu-sheng:
> This is for step 1?

Yes, this is for task creation and query.

arugal:
I have a suggestion for `e2e-profile `. I think we should add `ThreadMonitorTask` to`e2e-cluster e2e-mysql e2e-single-service` instead of adding a new e2e.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/profile/IThreadMonitorTaskQueryDAO.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * search task list in appoint time bucket
+     * @param serviceId monitor service id, maybe null
+     * @param endpointName endpoint name, maybe empty
+     * @param startTimeBucket time bucket bigger than or equals
+     * @param endTimeBucket time bucket small than or equals
+     * @return
+     */
+    List<ThreadMonitorTask> getTaskList(final Integer serviceId, final String endpointName, final long startTimeBucket, final long endTimeBucket) throws IOException;
```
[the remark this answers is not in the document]
wu-sheng: They have very similar parameter, but different logic? It is hard to understand.

mrproliu:
> I have a suggestion for `e2e-profile `. I think we should add `ThreadMonitorTask` to`e2e-cluster e2e-mysql e2e-single-service` instead of adding a new e2e.

At present, it's only graphql in OAP to add query test, but the interaction in sniffer will be integrated later. I think it's better to use a new E2E

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/profile/IThreadMonitorTaskQueryDAO.java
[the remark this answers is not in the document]
mrproliu: they are using different logic.
`getTaskListSearchOnStartTime`: It will use on creating a new task to check limit, Each service can monitor up to 1 endpoints during the execution of tasks. search time range on task start time.
`getTaskList`: It will use on the UI level GraphQL searching. search time range on time bucket.
At present, I don't have any good way to optimize these two methods.

wu-sheng:
@kezhenxu94 I can't find anything related to the e2e failure in the log, even no error logs from OAP. Could you check why? Or do we miss some logs?

.github/workflows/e2e.yaml
```diff
@@ -80,6 +80,8 @@ jobs:
         run: export E2E_VERSION=jdk8-1.3 && bash -x test/e2e/run.sh e2e-ttl/e2e-ttl-es
       - name: TTL ES7 Tests(JDK8)
         run: export E2E_VERSION=jdk8-1.3 DIST_PACKAGE=apache-skywalking-apm-bin-es7.tar.gz ES_VERSION=7.4.2 && bash -x test/e2e/run.sh e2e-ttl/e2e-ttl-es
+      - name: Profile Tests(JDK8)
+        run: export E2E_VERSION=jdk8-1.3 && bash -x test/e2e/run.sh e2e-profile/e2e-profile-test-runner
```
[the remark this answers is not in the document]
mrproliu: Ok. I will change it after the e2e task failed issue resolved.

kezhenxu94:
> @kezhenxu94 I can't find anything related to the e2e failure in the log, even no error logs from OAP. Could you check why? Or do we miss some logs?

The failures of the E2E tests are due to the missing update of the submodule (protocol), thus missing the newly-added file `query-protocol/profile.graphqls`, @mrproliu please update the submodule in this PR

kezhenxu94:
The failure should be discoverable and reproducible in your local machine, if you make sure that your codes are the same with the remote repo @mrproliu

wu-sheng:
@kezhenxu94 He sent me this directly, that is why I asked.
image

mrproliu:
> The failure should be discoverable and reproducible in your local machine, if you make sure that your codes are the same with the remote repo @mrproliu

OK, I have submitted it.

wu-sheng:
Oh, you didn't update the submodule, strange, why log can't show this?

kezhenxu94:
> Oh, you didn't update the submodule, strange, why log can't show this?

The failure is due to timeout and the e2e-container is killed by the plugin, so we have no chance to `cat` the logs

wu-sheng:
> The failure is due to timeout and the e2e-container is killed by the plugin, so we have no chance to cat the logs

So, if we face the OAP bootstrap issue, then this error shows up without the logs. Could I make this as a conclusion of a possible case? Could we make this a part of e2e test faq?

arugal:
> > I have a suggestion for `e2e-profile `. I think we should add `ThreadMonitorTask` to`e2e-cluster e2e-mysql e2e-single-service` instead of adding a new e2e.
> 
> At present, it's only graphql in OAP to add query test, but the interaction in sniffer will be integrated later. I think it's better to use a new E2E

But `es` and `mysql` is not covered,  need to be added? @wu-sheng @kezhenxu94

kezhenxu94:
> So, if we face the OAP bootstrap issue, then this error shows up without the logs. Could I make this as a conclusion of a possible case? 

Yes I agree.

Because bootstrap failure should be always stable to reproduce, and should be discovered when establishing a new E2E case, my assumption is that the contributor should verify locally, and guaranteed the codes are all pushed to remote repo, unlike this case :)

> Could we make this a part of e2e test faq?

Sure, or maybe as an important note to contribute new E2E cases

wu-sheng:
> Sure, or maybe as an important note to contribute new E2E cases

Could you submit a separated PR about this?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/profile/IThreadMonitorTaskQueryDAO.java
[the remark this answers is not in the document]
wu-sheng: These two names show no difference. The storage implementor must be confused.

wu-sheng:
Before the further change, please make sure we have consistent name. Let's remove thread monitor in all names, only keep profile in name, including Graphql protocol. They are used randomly.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/profile/IThreadMonitorTaskQueryDAO.java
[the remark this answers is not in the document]
wu-sheng: The point is storage should consider data query only. You could query the list, then check. I still don’t see the difference. You are talking using in different places. But this is wrong as two separated methods.

You should provide a method a query existing task list based on given conditions, that is all. This method is not matching the UI query or create task. It is a storage interface.

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringUtil.java
```diff
@@ -23,6 +23,10 @@ public static boolean isEmpty(String str) {
         return str == null || str.length() == 0;
     }
 
+    public static boolean isBlank(String str) {
```
wu-sheng: There is already `StringUtil#isEmpty`. This is duplicated.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/ProfileMutation.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+        return profileTaskService;
+    }
+
+    public ProfileTaskCreationResult createProfileTask(ProfileTaskCreationRequest creationRequest) throws IOException {
+        return getProfileTaskService().createTask(
+                creationRequest.getServiceId(),
+                creationRequest.getEndpointName(),
+                creationRequest.getStartTime() == null ? -1 : creationRequest.getStartTime(),
+                Math.toIntExact(DurationUtils.INSTANCE.toSecond(creationRequest.getDurationUnit(), creationRequest.getDuration())),
```
wu-sheng: Reading this makes me feel, we should set the duration in minute level only, and remove the step in the protocol. @kezhenxu94 @arugal  What do you think?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/ProfileTaskQueryEsDAO.java
```diff
@@ -0,0 +1,83 @@
...
+            boolQueryBuilder.must().add(QueryBuilders.termQuery(ProfileTaskNoneStream.SERVICE_ID, serviceId));
+        }
+
+        if (!StringUtil.isBlank(endpointName)) {
+            boolQueryBuilder.must().add(QueryBuilders.termQuery(ProfileTaskNoneStream.ENDPOINT_NAME, endpointName));
+        }
+
+        boolQueryBuilder.must().add(QueryBuilders.rangeQuery(ProfileTaskNoneStream.TIME_BUCKET).gte(startTimeBucket).lte(endTimeBucket));
+
+        final SearchResponse response = getClient().search(ProfileTaskNoneStream.INDEX_NAME, sourceBuilder);
```
wu-sheng: I think we need a `sourceBuilder.size(limit);` here, right? @kezhenxu94 Otherwise, ES query will only get top 10?

dmsolr: Yes, we get 10 docuemnts only by default.

.github/workflows/e2e.yaml
```diff
@@ -95,4 +95,20 @@ jobs:
...
+          submodules: true
+      - name: Set environment
+        run: export MAVEN_OPTS='-Dmaven.repo.local=~/.m2/repository -XX:+TieredCompilation -XX:TieredStopAtLevel=1 -XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC -XX:-UseGCOverheadLimit -Xmx3g'
+      - name: Compile & Install Test Codes
+        run: |
+          ./mvnw checkstyle:check apache-rat:check
+          ./mvnw -Dcheckstyle.skip -Drat.skip -T2 -Dmaven.compile.fork -Dmaven.compiler.maxmem=3072 -DskipTests clean install
+          ./mvnw -f test/e2e/pom.xml -pl e2e-base clean install
+      - name: Profile Tests(JDK8)
+        run: export E2E_VERSION=jdk8-1.3 && bash -x test/e2e/run.sh e2e-profile/e2e-profile-test-runner
```
wu-sheng: Profile should be tested in every storage. H2, MySQL, es6, es7.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/ProfileMutation.java
kezhenxu94: > Reading this makes me feel, we should set the duration in minute level only, and remove the step in the protocol. @kezhenxu94 @arugal What do you think?

Make sense to me, but `minute` level seems too large in time span, what about making it `second` level?

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringUtil.java
kezhenxu94: > There is already `StringUtil#isEmpty`. This is duplicated.

They're different (trivial though) IMO, `StringUtil.isEmpty("  \t") == false` while `StringUtil.isBlank("  \t") == true`

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/ProfileTaskQueryEsDAO.java
mrproliu: That's true. I think to get 10 doc it's enough. Should I add a size limit?

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringUtil.java
wu-sheng: I highly doubt this is the purpose of creating this API..

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/ProfileTaskQueryEsDAO.java
wu-sheng: Why 10 is enough? I don't remember we limit the task amount.

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringUtil.java
mrproliu: Maybe I should do String#trim on GrapthQL resolver?

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/ProfileMutation.java
wu-sheng: Yes, agree. Let's change this is `ms` level, but to be clear, this is not for span, it is for segment

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringUtil.java
wu-sheng: Why do you need a trim? Who will send an empty string?

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/ProfileMutation.java
wu-sheng: According to https://github.com/apache/skywalking-query-protocol/blob/master/profile.graphqls#L25-L26, this duration is for the task, I think minute makes more sense.

wu-sheng: This threshold for segment(nothing related to span) is minDurationThreshold, refer to https://github.com/apache/skywalking-query-protocol/blob/master/profile.graphqls#L30

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/ProfileTaskQueryEsDAO.java
mrproliu: Oh, I just think is using on the task creation limit check, forget it will use on GraphQL. Should I add a limit and from param in the GraphQL?

wu-sheng: You could have a limit in DAO interface level, but for UI, if we want to set the limit, we should set paging in the GraphQL parameter

wu-sheng: Because if we set a limit in UI without paging, people would not see the full list forever. I still think the full list is enough for UI as the profiling task should not a long list.

wu-sheng: One more thing, as we have TTL for profile task, we could remove the duration here, https://github.com/apache/skywalking-query-protocol/blob/master/profile.graphqls#L68. We should always show the full list. Any suggestion? @arugal @kezhenxu94

dmsolr: Do we need to support order by `time`?

wu-sheng: Yes, should order by start time `DESC`

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringUtil.java
kezhenxu94: > Why do you need a trim? Who will send an empty string?

@wu-sheng As long as the string is typed by the users, it's totally possible that it will include blank characters, or even unprintable/invisible characters, we're suffering from this because the users copy-and-pastes from else where (especially Windows users, `\r` is painful), another solution is to trim and validate the input  in the frontend, but we may have other user interface, CLI I mean

wu-sheng: If that is the concern, `trim` the data in the GraphQL service level, this kind of blank issue clearly is related to humans, not to codes.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/DurationUtils.java
```diff
@@ -187,6 +189,36 @@ public int secondsBetween(Downsampling downsampling, DateTime dateTime) {
         return durations;
     }
 
+    /**
+     * convert duration to second unit
+     * @param step
+     * @param duration
+     * @return
+     */
+    public long toSecond(Step step, long duration) {
```
wu-sheng: I think this method is not necessary as the duration unit has been removed from the protocol

oap-server/server-storage-plugin/storage-elasticsearch7-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch7/query/ProfileTaskQueryEs7DAO.java
```diff
@@ -0,0 +1,101 @@
...
+import org.elasticsearch.search.sort.SortOrder;
+
+import java.io.IOException;
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * @author MrPro
+ */
+public class ProfileTaskQueryEs7DAO extends EsDAO implements IProfileTaskQueryDAO {
```
wu-sheng: Does this DAO have any different with ES6's?

mrproliu: They are the same. I deleted, use `ProfileTaskQueryEsDAO` on the es7 module provides.

PR #4152, opened 2019-12-30
[Feature] Add tag annotation to allow tagging span with annotation

kezhenxu94:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

This PR provides another annotation `Tag` (similar to `Trace`) to allow users to apply a tag to the current active span, it's more convenient to tag a trace in companion with `@Trace`:

```java
@Trace
@Tag(key = "tag", value = "arg[0]")
public void method(String param) {
}
```

or multiple tags

```java
@Trace
@Tags({
    @Tag(key = "tag1", value = "arg[0]"),
    @Tag(key = "tag2", value = "arg[1]"),
})
public void method(String param1, String param2) {
}
```

or

```java
@Trace
@Tag(key = "tag1", value = "arg[0]")
@Tag(key = "tag2", value = "arg[1]")
public void method(String param1, String param2) {
}
```

### DIFF
--- a/apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/Tag.java
+++ b/apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/Tag.java
@@ -28,6 +28,7 @@ import java.lang.annotation.Target;
  * if there is no active span, this annotation takes no effect.
  *
  * @author kezhenxu94
+ * @see Tags
  */
 @Target(ElementType.METHOD)
 @Retention(RetentionPolicy.RUNTIME)
--- a/apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/Tag.java
+++ b/apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/Tag.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.toolkit.trace;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Tag the current active span with key {@link #key()} and value {@link #value()},
+ * if there is no active span, this annotation takes no effect.
+ *
+ * @author kezhenxu94
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Tag {
+    /**
+     * @return the key of the tag to be injected into the current active span
+     */
+    String key();
+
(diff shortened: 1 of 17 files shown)


wu-sheng:
CI keeps failure. Please recheck.

wayilau:
Does this PR support we can using anntation write our agent plugins ?

wu-sheng:
> Does this PR support we can using anntation write our agent plugins ?

In the test, it can't, because they are not in the release :) So they are copied into the test project too.

wayilau:
> > Does this PR support we can using anntation write our agent plugins ?
> 
> In the test, it can't, because they are not in the release :) So they are copied into the test project too.

Great feature!

kezhenxu94:
> Does this PR support we can using anntation write our agent plugins ?

Do you mean using `@Tag` in the agent plugin methods? No you can't, classes of agent plugins are ignored:

https://github.com/apache/skywalking/blob/1ee14fbbc8251a1faa5f3851d4f7d92f9ec1ab15/apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java#L90-L98

so although you can annotate the plugin method with `@Tag`, it takes no effect

apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/Tag.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Tag the current active span with key {@link #key()} and value {@link #value()},
+ * if there is no active span, this annotation takes no effect.
+ *
+ * @author kezhenxu94
+ * @see Tags
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Tag {
```
dmsolr: Miss @Repeatable?

wayilau:
> > Does this PR support we can using anntation write our agent plugins ?
> 
> Do you mean using `@Tag` in the agent plugin methods? No you can't, classes of agent plugins are ignored:
> 
> https://github.com/apache/skywalking/blob/1ee14fbbc8251a1faa5f3851d4f7d92f9ec1ab15/apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java#L90-L98
> 
> so although you can annotate the plugin method with `@Tag`, it takes no effect

Yes, you got what i thought.  I think it's meaningful & If we do, make our developer write plugins  simplfy. @kezhenxu94

apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/Tag.java
wayilau: why it's still ref to the user even i add the quotes.

kezhenxu94: > why it's still ref to the user even i add the quotes.

Use backtick instead of quotes

kezhenxu94: > Miss "@repeatable"?

May be optional ? Tests passed

dmsolr:
> Yes, you got what i thought. I think it's meaningful & If we do, make our developer write plugins simplfy. @kezhenxu94

It supports LocalSpan only now! :)

wu-sheng:
Making no-local span is more about exact and inject ContextCarrier :) Even we add that, I think it will be closer to OpenTracing APIs, even it is not widely used.

kezhenxu94:
> > Yes, you got what i thought. I think it's meaningful & If we do, make our developer write plugins simplfy. @kezhenxu94
> 
> It supports LocalSpan only now! :)

@dmsolr nope, it supports any kinds of span (it is not related to specific kind of span), except for those annotated by `RequestMapping` and similar annotations that are intercepted by other agent plugins

dmsolr:
I got it. We work with Spring-Annotations plugins.

kezhenxu94:
> Doc of https://github.com/apache/skywalking/blob/master/docs/en/setup/service-agent/java-agent/Application-toolkit-trace.md should be updated.

@wu-sheng did you find this? https://github.com/apache/skywalking/pull/4152/files#diff-26046b3a31cc7df0b065c9cdc6b03c7d

wu-sheng:
Got it now. Maybe browser cache, can't see it when I refreshed the page last time.

PR #4156, opened 2019-12-31
Log harmless exception in DEBUG level

### LINKED ISSUE #4155: Decrease log level when byte buddy failed to match type
https://github.com/apache/skywalking/blob/8a97aad021b16eb72591323691e902b06bd54087/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/ProtectiveShieldMatcher.java#L45-L51

I'd like to raise a discussion on the log level in the above code snippet.

For some reasons, in some of our applications, we have many classes that are failed to be resolved by byte buddy, and cases are that the applications all work well including the SkyWalking agent, but the logs are somewhat overwhelming, I've found an old issue https://github.com/raphw/byte-buddy/issues/373 , and seems this exception is mostly harmless, so I'm wondering whether we could just log this kind of logs in `DEBUG` level, what's your opinion, @wu-sheng @ascrutae @JaredTan95 @dmsolr @arugal etc.

kezhenxu94:
Yet users can still troubleshoot by looking into the log keywords `ProtectiveShieldMatcher : Byte-buddy occurs exception when match type: Cannot resolve type description for ...`

closes #4155

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/api/ILog.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/api/ILog.java
@@ -51,10 +51,7 @@ public interface ILog {
 
     void debug(String format, Object... arguments);
 
-    default void debug(Throwable t, String format, Object... arguments) {
-        // Just ignore the throwable by default, to make the compiler happy
-        debug(format, arguments);
-    }
+    void debug(Throwable t, String format, Object... arguments);
 
     void error(String format);
 }
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/api/ILog.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/api/ILog.java
@@ -51,5 +51,10 @@ public interface ILog {
 
     void debug(String format, Object... arguments);
 
+    default void debug(Throwable t, String format, Object... arguments) {
+        // Just ignore the throwable by default, to make the compiler happy
+        debug(format, arguments);
+    }
+
     void error(String format);
 }
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/ProtectiveShieldMatcher.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/ProtectiveShieldMatcher.java
@@ -46,7 +46,9 @@ public class ProtectiveShieldMatcher<T> extends ElementMatcher.Junction.Abstract
         try {
             return this.matcher.matches(target);
         } catch (Throwable t) {
-            logger.debug("Byte-buddy occurs exception when match type: {}", t.getMessage());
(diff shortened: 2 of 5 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/ProtectiveShieldMatcher.java
```diff
@@ -46,7 +46,7 @@ public boolean matches(T target) {
         try {
             return this.matcher.matches(target);
         } catch (Throwable t) {
-            logger.warn(t, "Byte-buddy occurs exception when match type.");
+            logger.debug("Byte-buddy occurs exception when match type: {}", t.getMessage());
```
wu-sheng: Maybe it is better to use `if logger.debugenable` then `logger.warn`? The debug log like this will miss the error stack.

kezhenxu94: > Maybe it is better to use `if logger.debugenable` then `logger.warn`? The debug log like this will miss the error stack.

What I meant is to eliminate the stack traces, they look overwhelming and make users think that the agent is not working well

And the log level is `DEBUG` by default, `if (logger.isDebugEnable) then logger.warn` behaves the same as before

wu-sheng: I just have concerns about no stack when we really need this.

kezhenxu94: > I just have concerns about no stack when we really need this.

I had looked carefully into the stack traces when I wanted to find out why it really failed, and I'm sure the stack traces don't help at all.

So what about overloading a `debug` method with `Throwable`, and making the log level default to `INFO`?

kezhenxu94: I mean:

```java
if (logger.isDebugEnabled()) {
    logger.debug(throwable, "xxx")
}
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/api/ILog.java
```diff
@@ -51,5 +51,10 @@
 
     void debug(String format, Object... arguments);
 
+    default void debug(Throwable t, String format, Object... arguments) {
```
wu-sheng: this `default` for?

kezhenxu94: > this `default` for?

There’re many anonymous inner classes implementing this interface in unit test, I just don’t want to implement this method there over and over, it’s useless

wu-sheng: This is not a good start for the codes. Others may refer to this in further PRs.

PR #4157, opened 2019-12-31
Add service type from register and support browser as service type

wu-sheng:
As this service could be registered from the browser soon, we need to separate the service type and query for different services. For 7.0.0, the first step should be, register happens from app agents and browsers.

Now, the new protocol should support this new way and by using gRPC compatibility, this should not break the old agents. Waiting for CI to verification.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/register/ServiceInventory.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/register/ServiceInventory.java
@@ -58,7 +58,10 @@ public class ServiceInventory extends RegisterSource {
     @Setter @Getter @Column(columnName = NAME, matchQuery = true) private String name = Const.EMPTY_STRING;
     @Setter @Getter @Column(columnName = IS_ADDRESS) private int isAddress;
     @Setter @Getter @Column(columnName = ADDRESS_ID) private int addressId;
-    @Setter(AccessLevel.PRIVATE) @Getter(AccessLevel.PRIVATE) @Column(columnName = NODE_TYPE) private int nodeType;
+    /**
+     * Node type of the service, default {@link NodeType#Normal}
+     */
+    @Setter(AccessLevel.PRIVATE) @Getter(AccessLevel.PRIVATE) @Column(columnName = NODE_TYPE) private int nodeType = 0;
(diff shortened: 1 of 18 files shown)


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
```diff
@@ -118,6 +119,25 @@ public int numOfConjectural(int nodeTypeValue) throws IOException {
         boolQueryBuilder.must().add(timeRangeQueryBuild(startTimestamp, endTimestamp));
 
         boolQueryBuilder.must().add(QueryBuilders.termQuery(ServiceInventory.IS_ADDRESS, BooleanUtils.FALSE));
+        boolQueryBuilder.must().add(QueryBuilders.termQuery(ServiceInventory.NODE_TYPE, NodeType.Normal.value()));
```
kezhenxu94: Node type should be “normal” **OR** “null”? If the user want to migrate from 6.x and keep the existed metadata in storage? Same for other storage types below

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/register/ServiceInventory.java
```diff
@@ -58,7 +58,10 @@
     @Setter @Getter @Column(columnName = NAME, matchQuery = true) private String name = Const.EMPTY_STRING;
     @Setter @Getter @Column(columnName = IS_ADDRESS) private int isAddress;
     @Setter @Getter @Column(columnName = ADDRESS_ID) private int addressId;
-    @Setter(AccessLevel.PRIVATE) @Getter(AccessLevel.PRIVATE) @Column(columnName = NODE_TYPE) private int nodeType;
+    /**
+     * Node type of the service, default {@link NodeType#Normal}
+     */
+    @Setter(AccessLevel.PRIVATE) @Getter(AccessLevel.PRIVATE) @Column(columnName = NODE_TYPE) private int nodeType = 0;
```
wu-sheng: @kezhenxu94 As this type is int already, I think the value currently should be 0 already for all existing versions. I just change it for more clear and match the new type usage. That is all.

PR #4158, opened 2020-01-01
Happy new year! Update NOTICE year

wu-sheng:
2019->2020, happy new year :)


wu-sheng:
@kezhenxu94 @innerpeacez Please update the year of CLI and k8s subprojects too.

PR #4162, opened 2020-01-02
Support collecting http parameters

### LINKED ISSUE #324: Add option to start/stop collecting http get/post parameter
cc @ascrutae @pengys5 

I want to add this feature, after we have config downstream mechanism.

kezhenxu94:
Plugins of Tomcat, Spring MVC, Armeria are supported for now

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

This pull request provides another config item to enable/disable collecting http parameters of a request.

It uses a general name `plugin.http.collect_http_params` but doesn't guarantee that all http plugins should be under its control, unless the plugin itself takes this config item into consideration, such as Tomcat, Spring MVC, and Armeria plugin.

This pull request also partially close #324 but I would suggest open another issue to track the downstreaming config feature, and I believe we need more discussion on that, @wu-sheng

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/CollectionUtil.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/CollectionUtil.java
@@ -33,7 +33,7 @@ public final class CollectionUtil {
     public static String toString(final Map<String, String[]> map) {
         return map
             .entrySet()
-            .parallelStream()
+            .stream()
             .map(entry -> entry.getKey() + "=" + Arrays.toString(entry.getValue()))
             .collect(Collectors.joining("\n"));
     }
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/CollectionUtil.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/CollectionUtil.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.agent.core.util;
+
+import java.util.Arrays;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * Some utility methods for collections.
+ * Reinvent the wheels because importing third-party libs just for some methods is not worthwhile in agent side
+ *
+ * @author kezhenxu94
+ * @since 7.0.0
+ */
+public final class CollectionUtil {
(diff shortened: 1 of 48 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/CollectionUtil.java
```diff
@@ -0,0 +1,40 @@
...
+ * Reinvent the wheels because importing third-party libs just for some methods is not worthwhile in agent side
+ *
+ * @author kezhenxu94
+ * @since 7.0.0
+ */
+public final class CollectionUtil {
+    public static String toString(final Map<String, String[]> map) {
+        return map
+            .entrySet()
+            .parallelStream()
```
dmsolr: As far as I know, `parallelStream()` could cause high latency on high requests.

kezhenxu94:
> Why change Tomcat plugin rather than changing SpringMVC plugin? It should be more adaptive?

Is it possible to assume that all http services are based on SpringMVC? Armeria is an apparent exception, and that's why we have Tomcat plugin and Spring MVC plugin now

wu-sheng:
I am not sure whether Tomcat plugin will face the parameter encoding issue.

kezhenxu94:
> I am not sure whether Tomcat plugin will face the parameter encoding issue.

Yes sadly, I've suffered from that, and that's why the codes to get parameter map in Tomcat plugin is more complicated than those in Spring MVC's, at least I've tested and the plugin test codes show that.

kezhenxu94:
And is running on my test environment now

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -118,6 +118,8 @@ property key | Description | Default |
 `plugin.opgroup.*`|Support operation name customize group rules in different plugins. Read [Group rule supported plugins](op_name_group_rule.md)|Not set|
 `plugin.springtransaction.simplify_transaction_definition_name`|If true, the transaction definition name will be simplified.|false|
 `plugin.jdkthreading.threading_class_prefixes` | Threading classes (`java.lang.Runnable` and `java.util.concurrent.Callable`) and their subclasses, including anonymous inner classes whose name match any one of the `THREADING_CLASS_PREFIXES` (splitted by `,`) will be instrumented, make sure to only specify as narrow prefixes as what you're expecting to instrument, (`java.` and `javax.` will be ignored due to safety issues) | Not set |
+`plugin.http.collect_http_params`| This config item controls that whether the plugins related to HTTP should collect the parameters of the request. The name `plugin.http.collect_http_params` is rather general, but it doesn't guarantee that all http plugins should be under its control, unless the plugin itself takes this config item into consideration, such as Tomcat, Spring MVC, and Armeria plugin. | `false` |
```
wu-sheng: Do we need to indicate this should work for UTF-8 encoding project? I think some vendors are using GBK in telecom, bank, etic

kezhenxu94: > Do we need to indicate this should work for UTF-8 encoding project? I think some vendors are using GBK in telecom, bank, etic

I've modified the case to specify the encoding as GBK, and it works, still

test/plugin/scenarios/spring-3.1.x-scenario/src/main/webapp/WEB-INF/web.xml
```diff
@@ -23,6 +23,20 @@
...
+            <param-name>encoding</param-name>
+            <param-value>GBK</param-value>
+        </init-param>
+    </filter>
+
+    <filter-mapping>
+        <filter-name>setCharacterEncodingFilter</filter-name>
+        <url-pattern>/*</url-pattern>
+    </filter-mapping>
+
```
kezhenxu94: @wu-sheng Here I meant

wu-sheng: Got it. Do we need to revert back? I don't want the GBK is a good example :)

kezhenxu94: > Got it. Do we need to revert back? I don't want the GBK is a good example :)

I’m ok with both situations, will revert it

PR #4169, opened 2020-01-03
FAQ for backend log4j2.xml

seifeHu:
- [ ] Improve performance
Too many  larger log files can causes the hard disk to become full    ,add   .gz and TimeBasedTriggeringPolicy  . add the two changes ,the skywalking oap logs will be zipped in .gz file . 
This changes can make the oap server run  longer times .

### DIFF
--- a/dist-material/log4j2.xml
+++ b/dist-material/log4j2.xml
@@ -23,11 +23,12 @@
     </Properties>
     <Appenders>
         <RollingFile name="RollingFile" fileName="${log-path}/skywalking-oap-server.log"
-                     filePattern="${log-path}/skywalking-oap-server-%d{yyyy-MM-dd}-%i.log" >
+                     filePattern="${log-path}/skywalking-oap-server-%d{yyyy-MM-dd}-%i.log.gz" >
             <PatternLayout>
                 <pattern>%d - %c - %L [%t] %-5p %x - %m%n</pattern>
             </PatternLayout>
             <Policies>
+              	<TimeBasedTriggeringPolicy />
                 <SizeBasedTriggeringPolicy size="102400KB" />
             </Policies>
             <DefaultRolloverStrategy max="30"/>


wu-sheng:
My question is for gz. This is only a requirement for all users, from my understanding. If you have the use case, please consider to add a FAQ doc, rather than changing this.

seifeHu:
> My question is for gz. This is only a requirement for all users, from my understanding. If you have the use case, please consider to add a FAQ doc, rather than changing this.

In my case , about running 15 days . My oap logs got more than 60G . so I think default log config do zip . Can lead to  oap server to running longer times.

wu-sheng:
I mean, users could have different solution. Such as using error log only, close log for some packages, or like you using zip mode. My point, yours is just one option. It fits FAQ more than changing the distance tar permanently.

wu-sheng:
@kezhenxu94 @arugal @JaredTan95 Any suggestion about this?

kezhenxu94:
I'm totally agree to keep it unchanged because the changes here are just configurations from third-party framework (log4j), users should modify the configurations in their own environments instead of in SkyWalking repo, we cannot meet every user's requirement if future users don't like the archived logs

seifeHu:
ok. I will make this content into FAQ

kezhenxu94:
@seifeHu any update?

JaredTan95:
We also used `.gz` feature to compress log file size in our scenario, but It doesn't mean suitable for all users. So, Let's keep it and add FAQ docs.

wu-sheng:
@seifeHu Could you update this PR?

wu-sheng:
@kezhenxu94 @JaredTan95 Should we close this? As no update in 10 days.

kezhenxu94:
Prefer to close it and reopen or open another PR if @seifehu is ready

PR #4172, opened 2020-01-04
Provide profile task downstream to sniffer

mrproliu:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#4104 

___
- Provide profile task downstream to sniffer, follow the step 2.


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/profile/ProfileTaskExecutionService.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Profile task executor, use {@link #addProfileTask(ProfileTask)} to add a new profile task.
+ *
+ * @author MrPro
+ */
+public class ProfileTaskExecutionService implements BootService {
+
+    private static volatile ScheduledExecutorService PROFILE_TASK_READY_SCHEDULE = Executors.newScheduledThreadPool(15, new DefaultNamedThreadFactory("PROFILE-TASK-READY-SCHEDULE"));
```
arugal: Hi, I suggest will `ProfileTaskExecutionService` disabled by default, enabled by the user through the displayed configuration.

mrproliu: Sorry, I'm not sure what you mean? Why need to disable it? I need to use it to process the profile task. Do you mean to make a switch, user decides sniffer can enable profile?

arugal: > Do you mean to make a switch, user decides sniffer can enable profile?

Yes.

wu-sheng: @arugal Why make it disabled in default? Could you explain a little?

mrproliu: please check.
I add a new agent switch config `SW_AGENT_ACTIVE_PROFILE`.
I will be working on the query profile tasks. If disable the query, the sniffer will not have any profile task need to be executed.

mrproliu: please check. I have added a new agent switch config `agent.active_profile`.
It will be working on the query profile tasks. If disable the query, the sniffer will not have any profile task need to be executed.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -118,6 +118,11 @@
          * Limit the length of the operationName to prevent errors when inserting elasticsearch
          **/
         public static int OPERATION_NAME_THRESHOLD = 500;
+
+        /**
+         * If true, skywalking agent will enable profile when user create a new profile task
+         */
+        public static boolean ACTIVE_PROFILE = false;
```
wu-sheng: As a key feature of 7.0.0, this flag should be `true` as default. @arugal Could you share your concern?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/profile/ProfileTaskExecutionService.java
```diff
@@ -0,0 +1,92 @@
...
+     * @param task
+     */
+    public void addProfileTask(ProfileTask task) {
+        // update last command create time
+        if (task.getStartTime() > lastCommandCreateTime) {
+            lastCommandCreateTime = task.getStartTime();
+        }
+
+        long timeFromStartMills = task.getStartTime() - System.currentTimeMillis();
+        if (timeFromStartMills <= 0) {
```
wu-sheng: `<` should be enough.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/profile/ProfileTaskExecutionService.java
kezhenxu94: Agent config? So if the user wants to enable profiling, he must restart the agent? Unacceptable to me

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/profile/ProfileTaskQueryService.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static org.apache.skywalking.apm.agent.core.conf.Config.Collector.GRPC_UPSTREAM_TIMEOUT;
+
+/**
+ * sniffer will check has new profile task list every {@link Config.Collector#GET_PROFILE_TASK_INTERVAL} second.
+ *
+ * @author MrPro
+ */
+public class ProfileTaskQueryService implements BootService, Runnable, GRPCChannelListener {
```
wu-sheng: Use `@DefaultImplementor` in all agent services.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
arugal: > As a key feature of 7.0.0, this flag should be `true` as default. @arugal Could you share your concern?

1.  I think thread monitoring is an auxiliary function
1.  Thread snapshot collection has a great impact on application performance

So I think it should be enabled by the user display

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/profile/ProfileTaskExecutionService.java
```diff
@@ -0,0 +1,92 @@
...
+    private static volatile ScheduledExecutorService PROFILE_TASK_READY_SCHEDULE = Executors.newScheduledThreadPool(15, new DefaultNamedThreadFactory("PROFILE-TASK-READY-SCHEDULE"));
+
+    // last command create time, use to next query task list
+    private volatile long lastCommandCreateTime = -1;
+
+    /**
+     * get profile task from OAP
+     * @param task
+     */
+    public void addProfileTask(ProfileTask task) {
```
wu-sheng: Tasklist in the queue list should be cached and do time window validation again. Even we did in the UI, but there could be bug or storage changed through DB directly. It is better to recheck here.

Same rule as backend did. There should be only one task in a certain time.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/profile/ProfileTaskExecutionService.java
kezhenxu94: Make it “final” and remove “volatile”

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
wu-sheng: That is why we set the acceptable task with the following limitations
1. Only one endpoint at a certain time.
1. 15 mins as max duration of the task
1. dumpPeriod >= 10ms
1. minDurationThreshold for collecting thread dump in meaningful case.
1. maxSamplingCount should less than 10. FYI @mrproliu I think `maxSamplingCount` is missing from existing codes, backend and protocol. right? Please fix this.

We should not set that in the false, because once the undetected performance issue happens, you have no change to reboot agent already.

wu-sheng: You could recheck these in the design doc, except the (5), we miss that in the doc too.

arugal: > You could recheck these in the design doc, except the (5), we miss that in the doc too.

I'll re-read the design document

wu-sheng: @mrproliu Set up a config class for profile first.

```java
public static class Profile {
           public static boolean ACTIVE = true;
}
```

apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/constants/ProfileConstants.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ */
+package org.apache.skywalking.apm.network.constants;
+
+/**
+ * profile task limit constants
+ *
+ * @author MrPro
+ */
+public class ProfileConstants {
```
wu-sheng: All these should be used in `ProfileTaskCommandExecutor#execute`. Once check fails, warning logs should output and the task wouldn't add into the scheduler.

mrproliu: Add the check info ProfileTaskExecutionService#addProfileTask, there saved task list and do the recheck.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
mrproliu: please check. `maxSamplingCount` have added into design document and codes.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/profile/ProfileTaskExecutionService.java
```diff
@@ -0,0 +1,207 @@
...
+ * Profile task executor, use {@link #addProfileTask(ProfileTask)} to add a new profile task.
+ *
+ * @author MrPro
+ */
+@DefaultImplementor
+public class ProfileTaskExecutionService implements BootService {
+
+    private static final ILog logger = LogManager.getLogger(ProfileTaskExecutionService.class);
+
+    private final static ScheduledExecutorService PROFILE_TASK_READY_SCHEDULE = Executors.newScheduledThreadPool(15, new DefaultNamedThreadFactory("PROFILE-TASK-READY-SCHEDULE"));
```
wu-sheng: You are creating a big thread pool here, could you share why? Is 15 really necessary?

wu-sheng: All tasks should be no conflicts, then it should be a single thread task scheduler.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/profile/ProfileTaskExecutionService.java
```diff
@@ -0,0 +1,207 @@
...
+    private final static ScheduledExecutorService PROFILE_TASK_FINISH_SCHEDULE = Executors.newScheduledThreadPool(2, new DefaultNamedThreadFactory("PROFILE-TASK-FINISH-SHCEDULE"));
```
wu-sheng: Same question for this scheduler. The task should only start after the prev one finished, right? If so, why two schedulers?

docs/en/setup/backend/backend-receivers.md
```diff
@@ -42,6 +42,8 @@ receiver_zipkin:
     host: 0.0.0.0
     port: 9411
     contextPath: /
+receiver-profile:
```
wu-sheng: This receiver has no document description, please add it.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/cache/CacheUpdateTimer.java
```diff
@@ -115,4 +121,25 @@ private void updateNetAddressInventory(ModuleDefineHolder moduleDefineHolder) {
             }
         });
     }
+
+    /**
+     * update all profile task list for each service
+     * @param moduleDefineHolder
+     */
+    private void updateProfileTaskNoneStream(ModuleDefineHolder moduleDefineHolder) {
```
wu-sheng: Why say `NoneStream`? There is no stream related thing here.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleConfig.java
```diff
@@ -67,6 +67,11 @@
     private long maxSizeOfEndpointInventory = 1_000_000L;
     private long maxSizeOfNetworkInventory = 1_000_000L;
 
+    /**
+     * Following are cache setting for none stream(s)
+     */
+    private long maxSizeOfProfileTaskNoneStream = 10_000L;
```
wu-sheng: Why related to NoneStream?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/cache/ProfileTaskCache.java
```diff
@@ -0,0 +1,124 @@
...
+
+        long initialSize = moduleConfig.getMaxSizeOfProfileTaskNoneStream() / 10L;
+        int initialCapacitySize = (int)(initialSize > Integer.MAX_VALUE ? Integer.MAX_VALUE : initialSize);
+
+        profileTaskCache = CacheBuilder.newBuilder().initialCapacity(initialCapacitySize).maximumSize(moduleConfig.getMaxSizeOfProfileTaskNoneStream())
+                // remove old profile task data
+                .expireAfterWrite(Duration.ofMinutes(1)).build();
+    }
+
+    public IProfileTaskQueryDAO getProfileTaskQueryDAO() {
```
wu-sheng: If this is only used internally, it should not be a public method.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/cache/ProfileTaskCache.java
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * query executable profile task
+     * @param serviceId
+     * @return
+     */
+    public List<ProfileTask> getProfileTaskList(int serviceId) {
+        List<ProfileTask> profileTaskList = profileTaskCache.getIfPresent(serviceId);
+
+        if (Objects.isNull(profileTaskList)) {
```
wu-sheng: Your cache already has auto-refresh mechanism, there is no need to read again.

wu-sheng: If this section is not required, `#getProfileTaskQueryDAO` method is not required too.

mrproliu:
@wu-sheng all comment issue has been resolved, please re-check.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/profile/ProfileTaskExecutionService.java
```diff
@@ -0,0 +1,207 @@
...
+            logger.warn("check command error, cannot process this profile task. reason: {}", dataError);
+            return;
+        }
+
+        // add task to list
+        profileTaskList.add(task);
+
+        // check task start now or make a schedule
+        long timeFromStartMills = task.getStartTime() - System.currentTimeMillis();
+        if (timeFromStartMills < 0) {
```
wu-sheng: I think we don't need this part. All tasks should be in the `PROFILE_TASK_SCHEDULE`. Could you share why we need this?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/profile/ProfileTaskExecutionService.java
```diff
@@ -0,0 +1,207 @@
...
+                @Override
+                public void run() {
+                    processProfileTask(task);
+                }
+            }, timeFromStartMills, TimeUnit.MILLISECONDS);
+        }
+    }
+
+    /**
+     * real process a new profile task
```
wu-sheng: This comment is not right. Should re-word to 
> Active the selected profile task to execution task, and start a removal task for it.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/profile/ProfileTaskExecutionService.java
mrproliu: If sniffer restart, It will also get the task from OAP, maybe some task before the current time. I used to think that the schedule could not pass in negative numbers, and I found that there would be no errors. resolved.

wu-sheng:
@mrproliu @kezhenxu94 @dmsolr @arugal I am merging this first to make the work continue. If any of you find the issue, please submit an issue to track.

PR #4175, opened 2020-01-05
Update README.md

arugal:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___


README.md
```diff
@@ -44,7 +44,7 @@ including
 
 
 # Document
-- [6.x Documents](docs/README.md).
+- [7.x Documents](docs/README.md).
```
wu-sheng: Let's add document ref for 6.6 too.

README.md
```diff
@@ -44,7 +44,8 @@ including
 
 
 # Document
-- [6.x Documents](docs/README.md).
+- [6.x Documents](https://github.com/apache/skywalking/blob/6.x/docs/README.md).
```
wu-sheng: Move this as the second one. 

The first line of this page includes 6 too, https://github.com/apache/skywalking/blob/master/docs/README.md

arugal: updated

docs/README.md
```diff
@@ -94,11 +94,11 @@ If you are already familiar with SkyWalking, you could use this catalog to find
...
 
-SkyWalking 5.x is still supported by the community. For the users planning to upgrade from 5.x to 6.x, you should know some definitions of concepts changed.
+SkyWalking 5.x is still supported by the community. For the users planning to upgrade from 5.x to 7.x, you should know some definitions of concepts changed.
 
 The most important two changed concepts are:
 
-1. Application (in 5.x) is changed to **Service** (in 6.x), also Application Instance is changed to **Service Instance**.
-2. Service (in 5.x) is changed to **Endpoint** (in 6.x).
+1. Application (in 5.x) is changed to **Service** (in 7.x), also Application Instance is changed to **Service Instance**.
+2. Service (in 5.x) is changed to **Endpoint** (in 7.x).
```
kezhenxu94: I don't think these version numbers should be changed from `6.x` to `7.x`, because `6.x` is the nearest version (from 5.x) that has changes in concepts, the changes in this PR makes future users misunderstand that migrating from 5.x to 6.x doesn't have these issues.

@wu-sheng  If the 5.x is not supported now, I'd suggest simply remove this section, or are we still maintaining 5,.x?

wu-sheng: Let's remove the whole 5.x section.  

BTW, 5.x protocol should be removed too, including codes in agent, receivers, query(s). Anyone will do this? @kezhenxu94 @dmsolr @arugal

kezhenxu94: > BTW, 5.x protocol should be removed too, including codes in agent, receivers, query(s). Anyone will do this? @kezhenxu94 @dmsolr @arugal

I'll remove them

wu-sheng: @kezhenxu94 I create #4176 to track this, please notice including protocol documents.

arugal: removed

arugal:
Screenshot also update?

PR #4177, opened 2020-01-05
Enhance gRPC plugin

devkanro:
# Skywalking with enhanced gRPC
This PR enhances the original gRPC plugin of Skywalking.  
I created PR early for the request for comments. @GuoDuanLZ and I will accept opinion and improve this PR.

## Checklist
This PR still in WIP state, you can merge this PR after all check is done.
- [x] Implementation
- [x] Test
- [x] Documents
- [x] CI build pass
- [ ] Review

## Description
The original gRPC plugin just provided very basically tracing function, the exception of business code will not be traced.

We refactored this plugin for those functions:
01. Server and Client tracing has the same operations.
02. Provide the internal or external gRPC server tracing mode.
03. Streamline span timeline.
04. The error caused by business code will make the "/Response/onClose" fail.
05. Reduce interface point and optimize the code.

## Operations
The operation name of that this plugin created will be like
```
       service  traceSide             event
        ----┘      -----┘         --------┘
foo.bar.MyApi.echo/server/Request/onMessage
------┐       ---┐        ------┐
package      method        side

```
There are two sides of tracing for gRPC: `client-side` and `server-side`.   
Trace side provides a view for gRPC request, it means what the gRPC request looks like in the client or the server.  

There are five combinations for sides and events.
```
/Request/onMessage
# Request message received on the server or sent on the client.

/Request/onComplete
# Client has completed request sending, there is no more incoming request.

/Request/onCancel
# Client has canceled the call.

/Response/onMessage
# Response message received on client or sent on server.

/Response/onClose
# Call closed with status and trailers.
``` 

For simplifying tracing span, this plugin just creates `onMessage` spans for streaming calls.  
- For unary calls, no `onMessage` span created.  
- For client-streaming calls, just `/Request/onMessage` span created.  
- For server-streaming calls, just `/Response/onMessage` span created.  
- For bi-streaming calls, all `/onMessage` span created.  

For example:
```
foo.bar.MyApi.echo/server/Request/onMessage
# means it is a server-side tracing for server receive a request message from the client.

foo.bar.MyApi.echo/client/Request/onMessage
# means it is a client-side tracing for client send a request message to server.

foo.bar.MyApi.echo/server/Response/onMessage
# means server send a response message to client.

foo.bar.MyApi.echo/server/Response/onClose
# means server completed all handling for a request.
```

### Internal and external gRPC server tracing
If you have two gRPC servers called `Server1` and `Server2`, `Server1` will call the `Server2.test` method in your business code, we just need the `server-side` tracing because of the `Server2` is an internal server, `Server2` will provide `server-side` view, `server-side` view will be more useful for tracing a call than `client-side` view.

If `Server2` is an external gRPC server which you can't get `server-side` view by it. You will need the `Server1` `client-side` view for the call.

## Configure plugin
This plugin provides three configs for configuring internal and external gRPC servers.

```Java
/**
 * If this config is false, only client spans of peer which configured in INCLUDED_CLIENT_TRACING_PEERS
 * will be collected. In this mode EXCLUDED_CLIENT_TRACING_PEERS will override INCLUDED_CLIENT_TRACING_PEERS.
 * <p>
 * If this config is true, only client spans of peer which configured in EXCLUDED_CLIENT_TRACING_PEERS
 * will not be collected. In this mode INCLUDED_CLIENT_TRACING_PEERS will override EXCLUDED_CLIENT_TRACING_PEERS.
 */
public static boolean DEFAULT_CLIENT_TRACING_ENABLE = false;

/**
 * Included client tracing peers. gRPC plugin will collect client spans of configured config.
 */
public static List<String> INCLUDED_CLIENT_TRACING_PEERS = new LinkedList<>();

/**
 * Excluded client tracing peers. gRPC plugin will not collect client spans of configured config.
 */
public static List<String> EXCLUDED_CLIENT_TRACING_PEERS = new LinkedList<>();
```

Default config is no external gRPC server, no `client-side` spans created. If you need the `client-side` view for three-part APIs, add the host and port like `services.googleapis.com:11800` into `INCLUDED_CLIENT_TRACING_PEERS`.

## Screenshots
WIP


devkanro:
> If you enhance the plugin, i think you should adjust Grpc tests cases, then use ci to check first.

Yes, it already in our checklist, and @GuoDuanLZ will help us in tests.

dmsolr:
This is a very good pull request template for the plugin contribution. 👍

wu-sheng:
> This is a very good pull request template for the plugin contribution. 

Anyone want to discuss about new issue and pull request templates?

wu-sheng:
> Provide the internal or external gRPC server tracing mode.

What is this? Could you explain a little more?

devkanro:
> What is this? Could you explain a little more?

I will update the doc of PR later, I am too busy to update PR in the working day.

devkanro:
@wu-sheng @wayilau @kezhenxu94 
This PR is ready for reviews.

wu-sheng:
> Internal and external gRPC server tracing

It is still unclear for me about this section. Client/server side tracing are both required in nearly every RPC plugin. What makes the gRPC different? Manually setting this is very painful for the end user.

cc @kezhenxu94 Could you get the point of this?

devkanro:
@wu-sheng 
If you trace the both of client and server-side, you would get the tracing like blow of bi-streaming call:
```
+
|
+-+(Server1) Exit: foo.bar.MyApi.echo
| |
| +-+(Server2) Entry: foo.bar.MyApi.echo
|   |
|   +-+(Server2) Local: foo.bar.MyApi.echo/server/Request/onMessage
|   | |
|   | +-+(Server2) Local: foo.bar.MyApi.echo/server/Response/onMessage
|   |
|   +-+(Server2) Local: foo.bar.MyApi.echo/server/Request/onMessage
|   | |
|   | +-+(Server2) Local: foo.bar.MyApi.echo/server/Response/onMessage
|   |
|   +-+(Server2) Local: foo.bar.MyApi.echo/server/Request/onComplete
|     |
|     +-+(Server2) Local: foo.bar.MyApi.echo/server/Response/onClose
|
+-+(Server1) Local: foo.bar.MyApi.echo/client/Request/onMessage
|
+-+(Server1) Local: foo.bar.MyApi.echo/client/Resoponse/onMessage
|
+-+(Server1) Local: foo.bar.MyApi.echo/client/Request/onMessage
|
+-+(Server1) Local: foo.bar.MyApi.echo/client/Resoponse/onMessage
|
+-+(Server1) Local: foo.bar.MyApi.echo/client/Request/onComplete
|
+-+(Server1) Local: foo.bar.MyApi.echo/client/Response/onClose
```

A call will be traced both in server1(client-side), and server2(server-side). They are redundant, and the server-side is more accurate than the client-side because of the gRPC framework. You can distinguish the server active message(parent is entry span) and passive  message(parent is request message span).

In the internal server case, we don't need the client-side tracing, but in the external server case, client-side tracing is the only way to tracing gRPC call.

devkanro:
In the default config, all gRPC calls are internal calls. No client-side tracing will be created. I think the default config is enough for most cases (most of external APIs will not be gRPC). But if you need the client-side tracing for separate peers, just add it to `INCLUDED_CLIENT_TRACING_PEERS`.

devkanro:
The original plugin doesn't provide the client-side tracing, I add it for my case and optional powerful tracing.

wu-sheng:
If no client-side span, there is no client-side metrics of topology. Then you would detect the error of unreachable or network perf unstable issues from trace and response time page.  The thing you posted is exactly an APM should collect.

Back to my point, this is a basic design of SkyWalking. We should not argue about this in a single one plugin. If you want to discuss that, it is more than this. You need to change the design and protocol of the project.

If you want to change this, I prefer you keep that in private, and only push the both sides tracing in the upstream.

devkanro:
I agree with you, the error of unreachable or network perf unstable issues are also important for service.

I can simplify client-side tracing for this, like no `onMessage` event, only `Complete/Close` event for internal client-side tracing.

wu-sheng:
> I can simplify client-side tracing for this, like no onMessage event, only Complete/Close event for internal client-side tracing.

I think in most cases, both of them should exist, including `onMessage/Complete/Close`. User could have further OP, such as DB or cache access in the onMessage like any other callback.

devkanro:
You convinced me, the client-side tracing is also important to trace.

wu-sheng:
I will do the review after you remove the configurations and make the tracing working at client/server sides with callback. Thanks for your contributions.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -388,5 +390,26 @@
...
+
+            /**
+             * Included client tracing peers. gRPC plugin will collect client spans of configured config.
+             */
+            public static List<String> INCLUDED_CLIENT_TRACING_PEERS = new LinkedList<>();
+
+            /**
+             * Excluded client tracing peers. gRPC plugin will not collect client spans of configured config.
+             */
+            public static List<String> EXCLUDED_CLIENT_TRACING_PEERS = new LinkedList<>();
```
wu-sheng: Do you forget to remove all these?

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/client/BlockingCallInterceptor.java
```diff
@@ -16,43 +16,48 @@
...
+    @Override
+    public void beforeMethod(Class clazz, Method method, Object[] allArguments, Class<?>[] parameterTypes,
+                             MethodInterceptResult result) {
+        Channel channel = (Channel) allArguments[0];
+        MethodDescriptor<?, ?> methodDescriptor = (MethodDescriptor<?, ?>) allArguments[1];
         final AbstractSpan span = ContextManager.createExitSpan(formatOperationName(methodDescriptor), channel.authority());
         span.setComponent(ComponentsDefine.GRPC);
-        SpanLayer.asRPCFramework(span);
+        span.setLayer(SpanLayer.RPC_FRAMEWORK);
+        ContextManager.getRuntimeContext().put(BLOCKING_CALL_EXIT_SPAN, span);
```
wu-sheng: When will remove this from the RuntimeContext?

wu-sheng: If not, whether the multiple RPCs will trigger dirty read?

devkanro: Yes, got it, I will fix it later

devkanro:
@GuoDuanLZ 
The test case need to update for client tracing

wu-sheng:
Got it. Waiting for you update.

devkanro:
It seems test case passed with client tracing

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCall.java
```diff
@@ -0,0 +1,111 @@
...
+                    if (status.getCause() == null) {
+                        span.errorOccurred().log(status.asRuntimeException());
+                    } else {
+                        span.errorOccurred().log(status.getCause());
+                    }
+                    break;
+                // Other status code means some predictable error occurred in server.
+                // Like PERMISSION_DENIED or UNAUTHENTICATED somethings.
+                // Similar to 4XX in HTTP status.
+                default:
```
wu-sheng: `case UNKNOWN`, `case INTERNAL` and `default` seem to have the same logic. What is the difference?

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCall.java
```diff
@@ -0,0 +1,111 @@
...
+                    // But if the status still has cause exception, we will log it too.
+                    if (status.getCause() != null) {
+                        span.errorOccurred().log(status.getCause());
+                    }
+                    break;
+            }
+            Tags.STATUS_CODE.set(span, status.getCode().name());
+            super.close(status, trailers);
```
wu-sheng: You should add a new `try/final` in this code block, and keep `super#close` in the final.

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCallListener.java
```diff
@@ -0,0 +1,102 @@
...
+    @Override
+    public void onMessage(REQUEST message) {
+        // We just create the request on message span for client stream calls.
+        if (!methodType.clientSendsOneMessage()) {
+            try {
+                final AbstractSpan span = ContextManager.createLocalSpan(operationPrefix + REQUEST_ON_MESSAGE_OPERATION_NAME);
+                span.setComponent(ComponentsDefine.GRPC);
+                span.setLayer(SpanLayer.RPC_FRAMEWORK);
+                ContextManager.continued(contextSnapshot);
+                super.onMessage(message);
```
wu-sheng: A similar suggestion, `try/final` should be added here.

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCallListener.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override
+    public void onCancel() {
+        try {
+            final AbstractSpan span = ContextManager.createLocalSpan(operationPrefix + REQUEST_ON_CANCEL_OPERATION_NAME);
+            span.setComponent(ComponentsDefine.GRPC);
+            span.setLayer(SpanLayer.RPC_FRAMEWORK);
+            ContextManager.continued(contextSnapshot);
+            super.onCancel();
```
wu-sheng: A similar suggestion, `try/final` should be added here.

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCallListener.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override
+    public void onHalfClose() {
+        try {
+            final AbstractSpan span = ContextManager.createLocalSpan(operationPrefix + REQUEST_ON_COMPLETE_OPERATION_NAME);
+            span.setComponent(ComponentsDefine.GRPC);
+            span.setLayer(SpanLayer.RPC_FRAMEWORK);
+            ContextManager.continued(contextSnapshot);
+            super.onHalfClose();
```
wu-sheng: A similar suggestion, `try/final` should be added here.

test/plugin/scenarios/grpc-scenario/config/expectedData.yaml
```diff
@@ -15,239 +15,542 @@
...
   applications:
-    - {grpc-scenario: 2}
+  - { grpc-scenario: 2 }
   instances:
-    - {grpc-scenario: 1}
+  - { grpc-scenario: 1 }
   operationNames:
-    - grpc-scenario: [Greeter.sayHello, GreeterBlocking.sayHello,
-                      /grpc-scenario/case/grpc-scenario]
+  - grpc-scenario: [/grpc-scenario/case/healthCheck, Greeter.sayHello, GreeterBlocking.sayHello, GreeterBlockingError.sayHello, /grpc-scenario/case/grpc-scenario]
```
wu-sheng: Health check should not be included.

test/plugin/scenarios/grpc-scenario/config/expectedData.yaml
```diff
@@ -15,239 +15,542 @@
...
-                 entryEndpointId: 0, refType: CrossThread, parentSpanId: 1, parentTraceSegmentId: not null,
-                 parentServiceInstanceId: 1, networkAddress: '', entryEndpoint: /grpc-scenario/case/grpc-scenario,
-                 entryServiceInstanceId: 1}
-
+- applicationCode: grpc-scenario
+  segmentSize: gt 10
+  segments:
+  - segmentId: not null
+    spans:
+    - operationName: /grpc-scenario/case/healthCheck
```
wu-sheng: Health check should not be included.

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCall.java
devkanro: `UNKNOWN` and `INTERNAL` is similar to the `5xx` errors in HTTP, it must case span error even status has no cause exception.

`default`(others) is similar to the `4xx` errors in HTTP, we just case error when status has cause exception for user debugging.

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCall.java
```diff
@@ -0,0 +1,111 @@
...
+                // Similar to 4XX in HTTP status.
+                default:
+                    // But if the status still has cause exception, we will log it too.
+                    if (status.getCause() != null) {
+                        span.errorOccurred().log(status.getCause());
+                    }
+                    break;
+            }
+            Tags.STATUS_CODE.set(span, status.getCode().name());
+        } catch (Throwable t) {
```
wu-sheng: What is this catch for? I can't see any code happens except the plugin logic. We should not log an error caused by our own codes, rather than users.

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCallListener.java
```diff
@@ -0,0 +1,102 @@
...
+    public void onMessage(REQUEST message) {
+        // We just create the request on message span for client stream calls.
+        if (!methodType.clientSendsOneMessage()) {
+            try {
+                final AbstractSpan span = ContextManager.createLocalSpan(operationPrefix + REQUEST_ON_MESSAGE_OPERATION_NAME);
+                span.setComponent(ComponentsDefine.GRPC);
+                span.setLayer(SpanLayer.RPC_FRAMEWORK);
+                ContextManager.continued(contextSnapshot);
+            } catch (Throwable t) {
+                ContextManager.activeSpan().errorOccurred().log(t);
```
wu-sheng: As I last comment, please just `try/final`, but not `catch`. I just want to guarantee our plugins would not break the original codes even our plugins have bugs.

Please recheck the codes based on this rule.

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCall.java
wu-sheng: Please follow the next comment.

devkanro:
Some discuss not about this PR.

I found there is bad throughput about OAP server collect tracing.
I have tried to use skywalking in my production env, but there are too many segments to collect.
And I scale the OAP server to 8c16gx8 instances, there are also too many errors about collecting(gRPC client canceled.)
The CPU looks fine about OAP server and ES server.

Maybe separate collectors from OAP server is a good idea?
Agents are not calling OAP server to collect spans directly, use the MessageQueue or log collector to do it, and make collectors consume those messages or log, one collector maybe 1c2g/2c4g, we can have many collectors with low cost.

wu-sheng:
You could use another issue or mail list to discuss this. But to be honest, I can't see much difference. Many discussions have been done about this.

test/plugin/scenarios/grpc-scenario/config/expectedData.yaml
```diff
@@ -15,239 +15,524 @@
...
+      componentName: ''
+      isError: false
+      spanType: Exit
+      peer: '127.0.0.1:18080'
+      peerId: 0
+    - {operationName: GreeterBlocking.sayHello/client/Request/onComplete, operationId: 0,
+       parentSpanId: 2, spanId: 3, spanLayer: RPCFramework, startTime: nq 0,
+       endTime: nq 0, componentId: 23, componentName: '', isError: false,
+       spanType: Local, peer: '', peerId: 0}
+    - {operationName: GreeterBlocking.sayHello/client/Response/onClose, operationId: 0,
```
wu-sheng: Why do these two use different formats? I know they may be right but it seems strange.

test/plugin/scenarios/grpc-scenario/config/expectedData.yaml
```diff
@@ -15,239 +15,524 @@
...
+      spanLayer: RPCFramework
+      startTime: nq 0
+      endTime: nq 0
+      componentId: 23
+      componentName: ''
+      isError: false
+      spanType: Exit
+      peer: '127.0.0.1:18080'
+      peerId: 0
+    - {operationName: GreeterBlockingError.sayHello/client/Request/onComplete, operationId: 0,
```
wu-sheng: This one too.

test/plugin/scenarios/grpc-scenario/config/expectedData.yaml
GuoDuanLZ: When the span is shorter, it is a simplified format.

GuoDuanLZ: > Why do these two use different formats? I know they may be right but it seems strange.

When the span is shorter, it is a simplified format.

test/plugin/scenarios/grpc-scenario/config/expectedData.yaml
GuoDuanLZ: > This one too.

This same as above

test/plugin/scenarios/grpc-scenario/config/expectedData.yaml
wu-sheng: Could you follow the same pattern? I prefer we keep this consistent same in all test cases.

GuoDuanLZ: > Could you follow the same pattern? I prefer we keep this consistent same in all test cases.

Yes, of course.

wu-sheng:
@kezhenxu94 Could you recheck this recently?

PR #4187, opened 2020-01-06
make a suggestion about DUBBO plugin

tzy1316106836:

- [ ] Bug fix

Dubbo plugin will have the following problems when retrying:
If the current link is triggered by other components that are not monitored, and it does not have EntrySpan, the stackDepth of the link is 0 when the dubbo is called internally. The traceId generated by the call is different. however, if the dubbo call timeout and retry, the traceId is the same, because the sw3 parameter of the last call is reused. In this scenario, The effect of multiple dubbo calls and dubbo retry should be the same. I think the problem needs to be fixed, that is, update the value of sw3 when retrying.


kezhenxu94:
@tzy1316106836 fix failed CI first, your code style doesn't meet the code style rules, add a space before `{`

tzy1316106836:
> @tzy1316106836 fix failed CI first, your code style doesn't meet the code style rules, add a space before `{`

done

apm-sniffer/apm-sdk-plugin/dubbo-plugin/src/main/java/org/apache/skywalking/apm/plugin/dubbo/DubboInterceptor.java
```diff
@@ -56,6 +57,9 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
         Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
         Invoker invoker = (Invoker)allArguments[0];
         Invocation invocation = (Invocation)allArguments[1];
+        if (invocation.getAttachments().containsKey(SW3CarrierItem.HEADER_NAME)) {
+            invocation.getAttachments().remove(SW3CarrierItem.HEADER_NAME);
+        }
```
wu-sheng: Remove should be something similar than this

```
            CarrierItem next = contextCarrier.items();
            while (next.hasNext()) {
                next = next.next();
                rpcContext.getAttachments().put(next.getHeadKey(), next.getHeadValue());
            }
```

Key is defined in there.

tzy1316106836: When dubbo retry, the parameter sw3 will not be updated.

tzy1316106836: Therefore, we should clear sw3 in the invocation first,  so that the latest sw3 will be placed according to the RpcContext

kezhenxu94: > When dubbo retry, the parameter sw3 will not be updated.

@tzy1316106836 what @wu-sheng meant is that you should not simply use `SW3CarrierItem.HEADER_NAME` as key to remove the context, you should use something like：

```java
            CarrierItem next = contextCarrier.items();
            while (next.hasNext()) {
                next = next.next();
                rpcContext.getAttachments().remove(next.getHeadKey());
            }
```

because there are chances that the users are using SkyWalking 6.x, which uses protocol v2 and the key is `SW6CarrierItem.HEADER_NAME` instead of `SW3CarrierItem.HEADER_NAME`, what's more, `SW3CarrierItem` will be removed soon

tzy1316106836: done

Issue #4182, opened 2020-01-06
Agent Authentication


lurui04:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [√ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?
the version of skywalking is 6.5.0.
skywalking agent has the config with 'agent.authentication';
but the application of backend setting, do not have the config of authentication

wu-sheng:
Yes. Because the backend doesn't include these codes. It used to be included in 5.x, but in 6.x and 7.x, no one submits pull requests to move those codes. If you have interests, please raise a pull request to do so.

lurui04:
OK.thanks for your helping

wu-sheng:
Provided.

PR #4197, opened 2020-01-08
add token authentication between agent and oap receiver.

JaredTan95:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/4182
___
Agent auth provided in Skywalking 5.x version, and I take it to 7.x version.


docs/en/setup/backend/backend-token-auth.md
```diff
@@ -0,0 +1,44 @@
+# Token Authentication
```
wu-sheng: I think this doc should be linked from main doc. At least, from following places
1. https://github.com/apache/skywalking/blob/master/docs/en/setup/backend/backend-setup.md#advanced-feature-document-link-list
1. Quick link from https://github.com/apache/skywalking/blob/master/docs/README.md

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/server/auth/AuthenticationHandler.java
```diff
@@ -0,0 +1,69 @@
...
+                public <REQ, RESP> ServerCall.Listener<REQ> interceptCall(ServerCall<REQ, RESP> serverCall,
+                    Metadata metadata,
+                    ServerCallHandler<REQ, RESP> next) {
+                    String token = metadata.get(AUTH_HEAD_HEADER_NAME);
+                    if (expectedToken.equals(token)) {
+                        return next.startCall(serverCall, metadata);
+                    } else {
+                        serverCall.close(Status.PERMISSION_DENIED, new Metadata());
+                        return new ServerCall.Listener() {
+                        };
```
kezhenxu94: Extract this listener to a field member instead of creating one for every call

docs/en/setup/backend/backend-token-auth.md
```diff
@@ -0,0 +1,44 @@
...
+## Supported version
+7.0.0+
+
+## How need token authentication after we have TLS?
```
kezhenxu94:
```suggestion
## Why need token authentication after we have TLS?
```

docs/en/setup/backend/backend-token-auth.md
```diff
@@ -0,0 +1,44 @@
...
+```yaml
+······
+receiver-sharing-server:
+  default:
+    authentication: ${SW_AUTHENTICATION:""}
+······
+```
+
+## Authentication fails
+The Skywalking OAP verifies every request from agent, allowed only the token match.
```
kezhenxu94:
```suggestion
The Skywalking OAP verifies every request from agent, only allows requests whose token matches the one configured in `application.yml`.
```

wu-sheng:
> AuthenticationHandler is actived through the sharing server, but after that, it could be working in core config. But if there is ip:port configuration in sharing server module, you will active the auth in core level, then server-internal communication will be blocked, the cluster will fail. Please fix this.

@kezhenxu94 Do you recheck this?

JaredTan95:
> 1. server-internal communication will be blocked, the cluster will fail. Please fix this.

I'm not sure this can avoid it:

```
if (!(targetService instanceof RemoteServiceHandler) && StringUtil.isNotEmpty(expectedToken)) {}
```

wu-sheng:
You should support add extra handler out of the core, then sharing server module could add the auth by itself.

JaredTan95:
I tested in my local, and i did not found `org.apache.skywalking.oap.server.core.remote.RemoteServiceHandler` instance step into `org.apache.skywalking.oap.server.core.server.auth.AuthenticationHandler`, do i miss something?

wu-sheng:
> I tested in my local, and i did not found org.apache.skywalking.oap.server.core.remote.RemoteServiceHandler instance step into org.apache.skywalking.oap.server.core.server.auth.AuthenticationHandler, do i miss something?

@JaredTan95 Looking at the codes, it should be. Another thing is, the auth should be a filter, rather than a handler. Otherwise, how are you sure the auth handler would execute before the other receiver handler?

kezhenxu94:
> > AuthenticationHandler is actived through the sharing server, but after that, it could be working in core config. But if there is ip:port configuration in sharing server module, you will active the auth in core level, then server-internal communication will be blocked, the cluster will fail. Please fix this.
> 
> @kezhenxu94 Do you recheck this?

will check later

oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -98,33 +98,33 @@ storage:
...
+#    concurrentRequests: ${SW_STORAGE_ES_CONCURRENT_REQUESTS:2} # the number of concurrent requests
+#    resultWindowMaxSize: ${SW_STORAGE_ES_QUERY_MAX_WINDOW_SIZE:10000}
+#    metadataQueryMaxSize: ${SW_STORAGE_ES_QUERY_MAX_SIZE:5000}
+#    segmentQueryMaxSize: ${SW_STORAGE_ES_QUERY_SEGMENT_SIZE:200}
+#    advanced: ${SW_STORAGE_ES_ADVANCED:""}
+  h2:
+    driver: ${SW_STORAGE_H2_DRIVER:org.h2.jdbcx.JdbcDataSource}
+    url: ${SW_STORAGE_H2_URL:jdbc:h2:mem:skywalking-oap-db}
+    user: ${SW_STORAGE_H2_USER:sa}
+    metadataQueryMaxSize: ${SW_STORAGE_H2_QUERY_MAX_SIZE:5000}
```
wu-sheng: You submit it by mistake.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/server/GRPCHandlerRegisterImpl.java
```diff
@@ -33,7 +35,7 @@ public GRPCHandlerRegisterImpl(GRPCServer server) {
     }
 
     @Override public void addHandler(BindableService handler) {
-        server.addHandler(handler);
+        AuthenticationFilter.INSTANCE.build(server, handler);
```
wu-sheng: I think you should add a method called 

> @Override public void addFilter()

Then in provider, you should add a filter, add(AuthenticationFilter), first. Then calling addHanlder will make those filters active.

The current implementation is working, but the codes look strange.

wu-sheng:
Please fix the CI, and add auth to one e2e case.

apm-webapp/pom.xml
```diff
@@ -133,7 +133,7 @@
                             <goal>npm</goal>
                         </goals>
                         <configuration>
-                            <arguments>install --registry=https://registry.npmjs.org/</arguments>
+                            <arguments>install --registry=https://registry.npm.taobao.org</arguments>
```
wu-sheng: This should not be submitted.

wu-sheng:
Look like CI still failing.

oap-server/server-receiver-plugin/skywalking-sharing-server-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/sharing/server/ReceiverGRPCHandlerRegister.java
```diff
@@ -18,22 +18,39 @@
...
 
     @Override public void addHandler(ServerServiceDefinition definition) {
         grpcHandlerRegister.addHandler(definition);
     }
+
+    @Override public void addFilter(ServerInterceptor interceptor) {
+        this.interceptor = interceptor;
+    }
+
+    public ServerServiceDefinition handlerInterceptorBind(BindableService handler, ServerInterceptor interceptor) {
```
wu-sheng: This should be private.

oap-server/server-receiver-plugin/skywalking-sharing-server-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/sharing/server/ReceiverGRPCHandlerRegister.java
```diff
@@ -18,22 +18,39 @@
...
+        } else {
+            grpcHandlerRegister.addHandler(handlerInterceptorBind(handler, interceptor));
+        }
     }
 
     @Override public void addHandler(ServerServiceDefinition definition) {
         grpcHandlerRegister.addHandler(definition);
     }
+
+    @Override public void addFilter(ServerInterceptor interceptor) {
```
wu-sheng: Based on your codes, the interceptor could only be one, so this method should be set, rather than add.
And, this method must be called before `addHandler`, should add this to the comment.

oap-server/server-receiver-plugin/skywalking-sharing-server-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/sharing/server/SharingServerModuleProvider.java
```diff
@@ -81,11 +89,17 @@ public SharingServerModuleProvider() {
             if (config.getGRPCThreadPoolSize() > 0) {
                 grpcServer.setThreadPoolSize(config.getGRPCThreadPoolSize());
             }
+
             grpcServer.initialize();
 
-            this.registerServiceImplementation(GRPCHandlerRegister.class, new GRPCHandlerRegisterImpl(grpcServer));
+            GRPCHandlerRegister grpcHandlerRegister = new GRPCHandlerRegisterImpl(grpcServer);
+
+            this.registerServiceImplementation(GRPCHandlerRegister.class, grpcHandlerRegister);
```
wu-sheng: Why change this? It seems the same.

oap-server/server-receiver-plugin/skywalking-sharing-server-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/sharing/server/ReceiverGRPCHandlerRegister.java
```diff
@@ -18,22 +18,39 @@
...
     @Setter private GRPCHandlerRegister grpcHandlerRegister;
+    private ServerInterceptor interceptor;
 
     @Override public void addHandler(BindableService handler) {
-        grpcHandlerRegister.addHandler(handler);
+        if (Objects.isNull(interceptor)) {
+            grpcHandlerRegister.addHandler(handler);
+        } else {
+            grpcHandlerRegister.addHandler(handlerInterceptorBind(handler, interceptor));
+        }
```
kezhenxu94: So you have an implicit prerequisite here: the `addFilter` method must be called before adding all handlers (`addHandler`), otherwise the filter doesn't fully take effect, right? If so, this prerequisite must be pointed out in the `interface`, but I don't think this is a good design, I'd suggest moving the interceptor to:

https://github.com/apache/skywalking/blob/6f90545a537d3830d0fcc1b82f51b444f848b5d4/oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/grpc/GRPCServer.java#L109

what do you think @wu-sheng

JaredTan95: Token authentication filter should only be activated in sharding server module. Moving to `GRPCServer` may not very suitable, in my thought.

kezhenxu94: > Token authentication filter should only be activated in sharding server module. Moving to `GRPCServer` may not very suitable, in my thought.

Well I missed that

PR #4213, opened 2020-01-10
Stop profiling task when backend doesn't support it

kezhenxu94:
When the backend doesn't support profiling task, the agent keep logging exception, this patch just cancel the scheduled task when the backend doesn't support this


kezhenxu94:
FYI @mrproliu

PR #4214, opened 2020-01-10
Support multiple linear values and merging p50/75/90/95/99 into percentile

### LINKED ISSUE #4190: Recommendations for reducing the number of indexes

- [ ] Question or discussion
- [ ] Bug
- [x] Requirement

___
### Question
- What do you want to know?
WX20200107-110218@2x

Under the default parameters of SW, the number of ES indexes has reached more than 500, and the number of shards has exceeded the default maximum index number of ES shards of 1000. Under certain conditions, the larger the number of shards will affect query performance, It will also increase the burden of cluster management and increase the instability of the cluster.
___
### Bug

- Which company or project?

- What happen?

___
Is it possible to combine hour, day, and month indexes and distinguish them by a type field inside the index, which is expected to reduce the number of indexes by 3/4. Query performance may be partially improved due to the use of fewer shards. But this has not been actually tested.

wu-sheng:
## New query protocol
```graphql
    # Query the type of metrics including multiple values, and format them as multiple linears.
    # The seq of these multiple lines base on the calculation func in OAL
    # Such as, should us this to query the result of func percentile(50,75,90,95,99) in OAL,
    # then five lines will be responsed, p50 is the first element of return value.
    getMultipleLinearIntValues(metric: MetricCondition!, numOfLinear: Int!, duration: Duration!): [IntValues!]!
```
- numOfLinear, represents how many lines expected. This related to the metrics name.

This will replace all existing Pxx metrics, which makes indexes merged and the number of index decreased. `percentile` is a new func in OAL, which calculate p50/75/90/95/99 all, and clearly reduce the performance of Pxx calculation.

### One side effect
1. We will remove the old PXX metrics in default, will affect dashboard, topology, and metrics comparison pages. cc @Fine0830 
1. One storage interface changed, a new method added. @dmsolr 

### TODO
Add the e2e tests for new query protocol.

Resolves #4190


wu-sheng:
FYI @aderm

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
```diff
@@ -343,6 +345,31 @@ private boolean isMatch() {
...
                         break;
+                    case MULTI_INTS:
+                        int[] ivalueArray = ((MultiIntValuesHolder)metrics).getValues();
+                        int[] iaexpected = RunningRule.this.threshold.getIntValuesThreshold();
+                        MULTI_VALUE_CHECK: for (int i = 0; i < ivalueArray.length; i++) {
+                            ivalue = ivalueArray[i];
+                            iexpected = 0;
+                            if (iaexpected.length > i) {
+                                iexpected = iaexpected[i];
+                            }
```
kezhenxu94: Is it reasonable to make it `0` by default? My intuition of the following configs:

```yml
threshold: 1000,1000
```

may be just monitor and alarm the `p99` and `p95` only, ignoring others percentile metrics, and your implementation seems to always send alarm? (because the other metrics may be almostly always > 0).

And are there any possibility when the users want to monitor `p99` and `p75`, how to config the threshold?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/PercentileMetrics.java
```diff
@@ -0,0 +1,120 @@
...
+    private static final int[] RANKS = {50, 75, 90, 95, 99};
+
+    @Getter @Setter @Column(columnName = VALUE, isValue = true) private IntKeyLongValueHashMap percentileValues;
+    @Getter @Setter @Column(columnName = PRECISION) private int precision;
+    @Getter @Setter @Column(columnName = DATASET) private IntKeyLongValueHashMap dataset;
+
+    private boolean isCalculated;
+
+    public PercentileMetrics() {
+        percentileValues = new IntKeyLongValueHashMap(5);
```
kezhenxu94:
```suggestion
        percentileValues = new IntKeyLongValueHashMap(RANKS.length);
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/PercentileMetrics.java
```diff
@@ -0,0 +1,120 @@
...
+    @Override
+    public final void calculate() {
+
+        if (!isCalculated) {
+            int total = dataset.values().stream().mapToInt(element -> (int)element.getValue()).sum();
+
+            int index = 0;
+            int[] roofs = new int[RANKS.length];
+            for (int i = 0; i < RANKS.length; i++) {
+                roofs[i] = Math.round(total * RANKS[i] * 1.0f / 100);
```
kezhenxu94: may reduce one multiply operation, I'm not sure though?

```suggestion
                roofs[i] = Math.round(total * RANKS[i] / 100.0);
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/PercentileMetrics.java
```diff
@@ -0,0 +1,120 @@
...
+            }
+
+            int count = 0;
+            IntKeyLongValue[] sortedData = dataset.values().stream().sorted(new Comparator<IntKeyLongValue>() {
+                @Override public int compare(IntKeyLongValue o1, IntKeyLongValue o2) {
+                    return o1.getKey() - o2.getKey();
+                }
+            }).toArray(IntKeyLongValue[]::new);
```
kezhenxu94: Do we need `.toArray`?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/PercentileMetrics.java
```diff
@@ -0,0 +1,120 @@
...
+import lombok.Getter;
+import lombok.Setter;
+import org.apache.skywalking.oap.server.core.analysis.metrics.annotation.Arg;
+import org.apache.skywalking.oap.server.core.analysis.metrics.annotation.Entrance;
+import org.apache.skywalking.oap.server.core.analysis.metrics.annotation.MetricsFunction;
+import org.apache.skywalking.oap.server.core.analysis.metrics.annotation.SourceFrom;
+import org.apache.skywalking.oap.server.core.storage.annotation.Column;
+
+/**
+ * Percentile is a better implementation than {@link PxxMetrics}. It is introduced since 7.0.0, it could calculate the
```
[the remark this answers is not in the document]
wu-sheng: This is using the func name.

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
wu-sheng: This is only a fail-safe mechanism, in case user doesn't set enough values.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/PercentileMetrics.java
wu-sheng: This should be optimized by the JDK, from my understanding.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/PercentileMetrics.java
wu-sheng: What do you suggest?

aderm:
can add default？ `getMultipleLinearIntValues(metric: MetricCondition!, duration: Duration!): [IntValues!]!` ，most are all.

wu-sheng:
> can add default？

What do you mean?

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
kezhenxu94: So the users are expected to set all thresholds for all percentile metrics?

wu-sheng: Yes. Or set to 0, then percentile alarm wouldn't be triggered by it. Because you will use `>` as OP.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/PercentileMetrics.java
```diff
@@ -0,0 +1,120 @@
...
+                @Override public int compare(IntKeyLongValue o1, IntKeyLongValue o2) {
+                    return o1.getKey() - o2.getKey();
+                }
+            }).toArray(IntKeyLongValue[]::new);
+            for (IntKeyLongValue element : sortedData) {
+                count += element.getValue();
+                for (int i = index; i < roofs.length; i++) {
+                    int roof = roofs[i];
+
+                    if (count >= roof) {
```
aderm: may be if condition `count >= roofs[i] && count<roofs[i+1]`? also change for i;

wu-sheng: Why? I am not following. Please provide test case to describe the concern.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/PercentileMetrics.java
kezhenxu94: Simply iterate over the “stream” without creating new object

wu-sheng: I am not familiar with the `stream` API. Do you have a proposal about how to change?

wu-sheng:
To all reviewers
Currently, all tests are not including the new `percentile` func, and the old `Pxx` OAL scripts have been removed from the OAL scripts. 

I and @kezhenxu94 will find a time today to fill the gap of e2e test.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/PercentileMetrics.java
kezhenxu94: Did not open in IDE just now, and the comparator can be simplified too, trivial though:

```java
dataset.values().stream().sorted(Comparator.comparingInt(IntKeyLongValue::getKey)).forEach(element -> {
                count += element.getValue();
                for (int i = index; i < roofs.length; i++) {
                    int roof = roofs[i];

                    if (count >= roof) {
                        percentileValues.put(index, new IntKeyLongValue(index, element.getKey() * precision));
                        index++;
                    } else {
                        break;
                    }
                }
            });
```

> Simply iterate over the “stream” without creating new object
 
Since you're using several variables that are non-final, it may make it uncomfortable to iterate over the `stream` directly, just ignore my previous comment, but the comparator can be simplified indeed

oap-server/server-bootstrap/src/main/resources/official_analysis.oal
```diff
@@ -17,22 +17,14 @@
  */
 
 // All scope metrics
-all_p99 = from(All.latency).p99(10);
-all_p95 = from(All.latency).p95(10);
-all_p90 = from(All.latency).p90(10);
-all_p75 = from(All.latency).p75(10);
-all_p50 = from(All.latency).p50(10);
```
aderm: Regarding pxx, we now use the percentile query at one time. From getMultipleLinearIntValues, we look at one index of the query, but now the back-end index is still divided into five Pxx. This is the next step of transformation? Or is it a merged index that has already been queried?

aderm:
> > can add default？
> 
> What do you mean?

@wu-sheng Wrongly written, can be ignored

oap-server/server-bootstrap/src/main/resources/official_analysis.oal
wu-sheng: > This is the next step of transformation? Or is it a merged index that has already been queried?

UI will change after this merged. @Fine0830 will do that part.

wu-sheng:
This is the query I used for testing
```
{
  getMultipleLinearIntValues(metric : {
    name: "all_percentile"
  }, duration: {step:MINUTE, start: "2020-01-10 2300", end: "2020-01-10 2330"},
  numOfLinear: 5) {
    values {
      id,
      value
    } 
  }
}
```

Then, I got the response
```
{
  "data": {
    "getMultipleLinearIntValues": [
      {
        "values": [
          {
            "id": "202001102300",
            "value": 0
          },
          {
            "id": "202001102301",
            "value": 0
          },
          {
            "id": "202001102302",
            "value": 0
          },
          {
            "id": "202001102303",
            "value": 0
          },
          {
            "id": "202001102304",
            "value": 0
          },
          {
            "id": "202001102305",
            "value": 0
          },
          {
            "id": "202001102306",
            "value": 0
          },
          {
            "id": "202001102307",
            "value": 0
          },
          {
            "id": "202001102308",
            "value": 0
          },
          {
            "id": "202001102309",
            "value": 0
          },
          {
            "id": "202001102310",
            "value": 0
          },
          {
            "id": "202001102311",
            "value": 0
          },
          {
            "id": "202001102312",
            "value": 0
          },
          {
            "id": "202001102313",
            "value": 0
          },
          {
            "id": "202001102314",
            "value": 0
          },
          {
            "id": "202001102315",
            "value": 0
          },
          {
            "id": "202001102316",
            "value": 0
          },
          {
            "id": "202001102317",
            "value": 0
          },
          {
            "id": "202001102318",
            "value": 0
          },
          {
            "id": "202001102319",
            "value": 0
          },
          {
            "id": "202001102320",
            "value": 4500
          },
          {
            "id": "202001102321",
            "value": 0
          },
          {
            "id": "202001102322",
            "value": 0
          },
          {
            "id": "202001102323",
            "value": 0
          },
          {
            "id": "202001102324",
            "value": 0
          },
          {
            "id": "202001102325",
            "value": 0
          },
          {
            "id": "202001102326",
            "value": 0
          },
          {
            "id": "202001102327",
            "value": 0
          },
          {
            "id": "202001102328",
            "value": 0
          },
          {
            "id": "202001102329",
            "value": 0
          },
          {
            "id": "202001102330",
            "value": 0
          }
        ]
      },
      .... Another 4 elements of array ignored
    ]
  }
}
```

wu-sheng:
@kezhenxu94 TTL still fails, I have changed the metrics from p99 to cpm, But still fail. What do I miss?

wu-sheng:
Maybe need to change avg response time? Due to cpm is less than 1?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/PercentileMetrics.java
wu-sheng: I copied most of the codes from Pxx, so, I prefer don't change it in this PR at least. If you want to change, let's move this to another new PR about code refactor, OK?

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
wu-sheng: Sorry, my mistake. All thresholds must be set at current implementation.

wu-sheng: I haven't found a better way to support setting part of it, maybe set `-` to represent skipping?

kezhenxu94: > I haven't found a better way to support setting part of it, maybe set `-` to represent skipping?

Sounds good to me

wu-sheng: Done, please review.

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
```diff
@@ -343,6 +345,38 @@ private boolean isMatch() {
...
+                                if (iNullableExpected == null) {
+                                    continue;
+                                }
+                            }
+                            switch (op) {
+                                case LESS:
+                                    if (ivalue < iNullableExpected) {
+                                        matchCount++;
+                                        break MULTI_VALUE_CHECK;
+                                    }
```
kezhenxu94: Missing `break` after `if` statement

```suggestion
                                    }
                                    break;
```

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
```diff
@@ -343,6 +345,38 @@ private boolean isMatch() {
...
+                                case GREATER:
+                                    if (ivalue > iNullableExpected) {
+                                        matchCount++;
+                                        break MULTI_VALUE_CHECK;
+                                    }
+                                case EQUAL:
+                                    if (ivalue == iNullableExpected) {
+                                        matchCount++;
+                                        break MULTI_VALUE_CHECK;
+                                    }
```
kezhenxu94: Same here

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetricsQueryEsDAO.java
```diff
@@ -121,7 +134,43 @@ protected void functionAggregation(Function function, TermsAggregationBuilder pa
...
+            for (int i = 0; i < intValuesArray.length; i++) {
+                KVInt kvInt = new KVInt();
+                kvInt.setId(id);
+                kvInt.setValue(0);
+                intValuesArray[i].addKVInt(kvInt);
+            }
+
+            if (idMap.containsKey(id)) {
+                Map<String, Object> source = idMap.get(id);
+                IntKeyLongValueHashMap multipleValues = new IntKeyLongValueHashMap(5);
```
kezhenxu94: I'm not very comfortable with the magic number here

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/h2/dao/H2MetricsQueryDAO.java
```diff
@@ -129,6 +143,48 @@ public IntValues getValues(String indName, Downsampling downsampling, long start
...
+        for (int i = 0; i < intValuesArray.length; i++) {
+            intValuesArray[i] = new IntValues();
+        }
+
+        try (Connection connection = h2Client.getConnection()) {
+            try (ResultSet resultSet = h2Client.executeQuery(connection, "select id, " + valueCName + " from " + tableName + " where id in (" + idValues.toString() + ")")) {
+                while (resultSet.next()) {
+                    String id = resultSet.getString("id");
+
+                    IntKeyLongValueHashMap multipleValues = new IntKeyLongValueHashMap(5);
```
kezhenxu94: Same here

docs/en/setup/backend/backend-alarm.md
```diff
@@ -13,7 +13,11 @@ Alarm rule is constituted by following keys
 endpoint name.
 - **Exclude names**. The following entity names are excluded in this rule. Such as Service name,
   endpoint name.
-- **Threshold**. The target value.
+- **Threshold**. The target value. 
+For multiple values metrics, such as **percentile**, the threshold is an array. Described like  `value1, value2, value3, value4, value5`.
+Each value could the threshold for each value of the metrics. Set the value to `-` if don't want to trigger alarm by this or some of the values.  
+Such as in **percentile**, `value1` is threshold of P50, and `-, -, value3, value4, value5` means, there is no threshold for P50 and P75 in percentile alarm rule.
+Could be an array, such as `value1, value2, value3` if the metrics are multiple values metrics. Element of the array could be NULL, then
```
kezhenxu94: Is this sentence completed? forget to update?

docs/en/setup/backend/backend-alarm.md
```diff
@@ -13,7 +13,11 @@ Alarm rule is constituted by following keys
 endpoint name.
 - **Exclude names**. The following entity names are excluded in this rule. Such as Service name,
   endpoint name.
-- **Threshold**. The target value.
+- **Threshold**. The target value. 
+For multiple values metrics, such as **percentile**, the threshold is an array. Described like  `value1, value2, value3, value4, value5`.
+Each value could the threshold for each value of the metrics. Set the value to `-` if don't want to trigger alarm by this or some of the values.  
+Such as in **percentile**, `value1` is threshold of P50, and `-, -, value3, value4, value5` means, there is no threshold for P50 and P75 in percentile alarm rule.
```
kezhenxu94: [1] (for reference)

docs/en/setup/backend/backend-alarm.md
```diff
@@ -51,13 +54,22 @@ rules:
     op: <
     period: 10
     count: 4
+  service_resp_time_percentile_rule:
+    # Metrics value need to be long, double or int
+    metrics-name: service_percentile
+    op: ">"
+    threshold: 1000,1000,1000,1000,1000
```
kezhenxu94: I don't know how other people think of this, I'm confused with the order of the thresholds at first glance, `p99` -> `p50` or `p50` -> `p99`, and seems only [1] tells the order, vaguely though

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetricsQueryEsDAO.java
wu-sheng: What do you expect? 5 should be suitable for 5-10 elements.

docs/en/setup/backend/backend-alarm.md
wu-sheng: Do you want me to put comments here? Or put doc at some places?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetricsQueryEsDAO.java
kezhenxu94: > What do you expect? 5 should be suitable for 5-10 elements.

oops, didn't notice it is a map, I thought it was array mistakenly, ignore me 😄

docs/en/setup/backend/backend-alarm.md
kezhenxu94: > Do you want me to put comments here? Or put doc at some places?

I prefer to point out the order here, users don't need to refer to other doc when they're uncertain with the order 😄

wu-sheng:
@kezhenxu94 I add more docs to ease your concern, please recheck.

wu-sheng:
@Fine0830 I will merge this tomorrow, then ask @ascrutae to deploy on the demo UI. Which will break part of the demo UI, but should be good for your UI update.

kezhenxu94:
I'll open a follow-up PR to do some refactoring work

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/PercentileMetrics.java
aderm: Should be correct, logical thinking is a bit problematic. above roofs should all hit

PR #4215, opened 2020-01-11
remove duplicate maven dependency.

JaredTan95:
remove duplicate  grpc-testing in server-core module maven dependency.


wu-sheng:
@JaredTan95 `duplicate` label means the issue/pr is duplicated with another one. :) This is not the case.

JaredTan95:
> @JaredTan95 `duplicate` label means the issue/pr is duplicated with another one. :) This is not the case.

got it.

PR #4217, opened 2020-01-12
Add pom exclude license file dir.

aderm:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___


aderm:
However, when executing this command `./mvnw install -Pagent -DskipTests`, the following error will occur. Personally, I think that a reasonable situation, the execution of the above command should pass normally.
WX20200112-184601@2x

wu-sheng:
I think you run the plugin test in the workspace, then this could trigger rat failure.

aderm:
> workspace

Indeed yes

pom.xml
```diff
@@ -450,6 +450,12 @@
                         <exclude>tools/dependencies/known-oap-backend-dependencies.txt</exclude>
                         <exclude>apm-checkstyle/CHECKSTYLE_HEAD</exclude>
 
+                        <!-- generated file from oap or webapp logs files -->
+                        <exclude>**/apm-dist/logs/**</exclude>
```
wu-sheng: Why are logs here?

aderm: If there is an error message during operation, it will be recorded here.
image

wu-sheng: I am saying why these logs exist. As nothing should be run there.

aderm: WX20200113-114517@2x

wu-sheng: But it should not start there. That is my point.

wu-sheng:
Do yoh continue work on this?

pom.xml
aderm: I ca n’t find it now. I looked at the time when the log was generated. It ’s been a long time ago. Indeed, from the current point of view, there is no entry to generate a log.
WX20200129-095628@2x

aderm:
so i close it?

wu-sheng:
> so i close it?

My point is `<exclude>**/apm-dist/logs/**</exclude>` shoule be removed. The workspace one makes sense to me.

aderm:
Ok, let me update

Issue #4194, opened 2020-01-07
Skywalking queries become slower as the amount of data becomes larger


Liu-XinYuan:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [x] Feature or performance improvement

___
### Question
- What do you want to know?

The skywalking page query becomes slower as the total amount of data increases, but we generally only query the data of the day, and when using elasticsearch as the back-end storage, all indexes are traversed according to the alias query, in fact we only need to query The index of the day, which caused the total query time to slow

- Which company or project?

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.

___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.
We only need to locate the index based on the time condition of the query and the timestamp information carried in the traceId. When the index is specified during the query, theoretically the number of indexes for the query will be reduced.

wu-sheng:
Thanks for submitting this, Please provide the new codes and benchmark.

Liu-XinYuan:
ok

wu-sheng:
Have been fixed in 8.1.0

PR #4231, opened 2020-01-14
Optimize the speed of UI query interface based on elasticsearch storage

Liu-XinYuan:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues
#4194 
___
### Bug fix
- Bug description.

- How to fix?

___


Liu-XinYuan:
These two parameters are passed from the front end. The backend determines which indexes the query data is based on these two times

wu-sheng:
You have the time-related parameter already. You should not break all.

Liu-XinYuan:
- startSecondTB and endSecondTB are not real time. 
- If I calculate the real time through these two parameters, I must get the parameter Step according to org.apache.skywalking.oap.server.core.query.DurationUtils and I must put this conversion in ElasticSearchClient. This will form a circular dependency between the library-client module and server-core

wu-sheng:
> startSecondTB and endSecondTB are not real time.

What do you mean not real? It is the exact time the client side query. And `Downsampling` is there too. What do you need? That is all. This is just an optimization inside the ES implementation, breaking all is not a good idea.

Liu-XinYuan:
It makes sense that I can convert to real time by downsampling and bucket.

PR #4239, opened 2020-01-15
Provide influxdb as a new storage plugin

dmsolr:

- [ ] Bug fix
- [X] New feature provided
- [ ] Improve performance
___

This PR provides a new storage plugin for OAP which uses two-level storage. 

The first level is a relationship database stored something likes `inventory`, which has transactions and updates. Stored in MySQL.(Now, I make H2 as relationship database. Because it is easy for test.)

The second level is a time-series database, stored `metrics`, `logs`, and `traces`. Those data have strong time-lined and no-transaction, no-updated, no-deleted. Stored in InfluxDB.

InfluxDB is a very powerful time-series database. It is good at manipulating time-series data. So, it is suitable for our trace records that have strong time-lined, no-updated, and no-deleted.

Now, it is an experimental feature. We need to continuously improve it with you. And it is a meaningful experience for others NoSQL/TSDB likes Apache Druid, OpenTSDB.

------------------------------------
## TODO list
- [x] feature
- [x] e2e with standone
- [x] e2e with cluster
- [x] document (backend-storage.md)
- [x] e2e-ttl
- [x] update license
- [x] comments (provided without interface implemented)
- [x] docker/docker-entrypoint.sh
- [x] refactor e2e-ttl and e2e-influxdb


dmsolr:
I don't deal with TTL yet. That can be implemented by RetentationPolicy, a InfluxDB built-in feature.

dist-material/application.yml
```diff
@@ -120,11 +120,11 @@ storage:
...
-  h2:
-    driver: ${SW_STORAGE_H2_DRIVER:org.h2.jdbcx.JdbcDataSource}
-    url: ${SW_STORAGE_H2_URL:jdbc:h2:mem:skywalking-oap-db}
-    user: ${SW_STORAGE_H2_USER:sa}
-    metadataQueryMaxSize: ${SW_STORAGE_H2_QUERY_MAX_SIZE:5000}
+#  h2:
+#    driver: ${SW_STORAGE_H2_DRIVER:org.h2.jdbcx.JdbcDataSource}
+#    url: ${SW_STORAGE_H2_URL:jdbc:h2:mem:skywalking-oap-db}
+#    user: ${SW_STORAGE_H2_USER:sa}
+#    metadataQueryMaxSize: ${SW_STORAGE_H2_QUERY_MAX_SIZE:5000}
```
wu-sheng: This should not be changed in PR.

dist-material/application.yml
```diff
@@ -135,6 +135,16 @@ storage:
 #      dataSource.prepStmtCacheSqlLimit: ${SW_DATA_SOURCE_PREP_STMT_CACHE_SQL_LIMIT:2048}
 #      dataSource.useServerPrepStmts: ${SW_DATA_SOURCE_USE_SERVER_PREP_STMTS:true}
 #    metadataQueryMaxSize: ${SW_STORAGE_MYSQL_QUERY_MAX_SIZE:5000}
+  influx:
+    myDriver: ${SW_STORAGE_MY_DRIVER:org.h2.jdbcx.JdbcDataSource}
```
wu-sheng: You should support to set metadata DB type. Otherwise, you have to link to single one DB implementation.

JaredTan95:
Powerful storage plugin～

wu-sheng:
Also, as we should market this feature, before 7.0.0 release, @dmsolr you should do some performance tests comparing this with ES storage in the same case. I am assuming this one has advantages :)

JaredTan95:
> The first level is a relationship database stored something likes inventory, which has transactions and updates. Stored in MySQL.(Now, I make H2 as relationship database. Because it is easy for test.)

BTW, also supported `TiDB`. :)

oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -98,28 +98,37 @@ storage:
...
-    # Batch process setting, refer to https://www.elastic.co/guide/en/elasticsearch/client/java-api/5.5/java-docs-bulk-processor.html
-    bulkActions: ${SW_STORAGE_ES_BULK_ACTIONS:1000} # Execute the bulk every 1000 requests
-    flushInterval: ${SW_STORAGE_ES_FLUSH_INTERVAL:10} # flush the bulk every 10 seconds whatever the number of requests
-    concurrentRequests: ${SW_STORAGE_ES_CONCURRENT_REQUESTS:2} # the number of concurrent requests
-    resultWindowMaxSize: ${SW_STORAGE_ES_QUERY_MAX_WINDOW_SIZE:10000}
-    metadataQueryMaxSize: ${SW_STORAGE_ES_QUERY_MAX_SIZE:5000}
-    segmentQueryMaxSize: ${SW_STORAGE_ES_QUERY_SEGMENT_SIZE:200}
-    advanced: ${SW_STORAGE_ES_ADVANCED:""}
+  influx:
+    myUrl: ${SW_STORAGE_MY_URL:org.h2.jdbcx.JdbcDataSource}
```
JaredTan95:
```
    myUrl: ${SW_STORAGE_MY_URL:org.h2.jdbcx.JdbcDataSource}
    myDriver: ${SW_STORAGE_MY_DRIVER:jdbc:h2:mem:skywalking-oap-db}
```
---> 

```
    myUrl: ${SW_STORAGE_MY_URL:jdbc:h2:mem:skywalking-oap-db}
    myDriver: ${SW_STORAGE_MY_DRIVER:org.h2.jdbcx.JdbcDataSource}
```

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/InfluxStorageDAO.java
```diff
@@ -53,6 +53,6 @@ public IRecordDAO newRecordDao(StorageBuilder<Record> storageBuilder) {
 
     @Override
     public INoneStreamDAO newNoneStreamDao(StorageBuilder<NoneStream> storageBuilder) {
-        return new NoneStreamDAO();
+        return new NoneStreamDAO(influxClient, storageBuilder);
```
wu-sheng: `INoneStreamDAO` is used for web interactive job, such as @mrproliu 's profiling task. Basically, this should have little data. Is the influxDB suitable for this?

dmsolr: That sounds the relationship database better.

wu-sheng: There would be much data, just easy to query should be enough. Anyway, based on your decision.

dmsolr:
How can I confirm that the submodule is the newest? If it is not the newest, what should I do?
I found my e2e always failed in my devel branch.

@wu-sheng , @kezhenxu94

kezhenxu94:
> How can I confirm that the submodule is the newest? If it is not the newest, what should I do?
> I found my e2e always failed in my devel branch.
> 
> @wu-sheng , @kezhenxu94

I don't think it's a matter of the sub-modules, the `interface IMetricsQueryDAO` is recently modified and you did not modify your `MetricsQuery` corespondingly

dmsolr:
I don't mean github action.  I am trying on my local branch which I have implemented.

wu-sheng:
> How can I confirm that the submodule is the newest? If it is not the newest, what should I do?

Check the commit IDs of submodules across checking with the other repos' commit ID.

wu-sheng:
Dependency changed, you need to update the LICENSE/NOTICE and dependency check files.

wu-sheng:
@dmsolr Besides the license issue, is this PR ready to be reviewed?

wu-sheng:
Please be noticed

> This branch is out-of-date with the base branch

Do you merge the master branch with this branch correctly? Please fix them before we can do the review.

dist-material/release-docs/NOTICE
```diff
@@ -905,4 +905,31 @@ be obtained at:
   * HOMEPAGE:
     * https://github.com/catapult-project/catapult
 
+------
+
+===========================================================================
```
wu-sheng: As MIT license you don't need to update NOTICE. NOTICE is for Apache 2.0 only. Please remove this part.

wu-sheng:
Is this PR ready to review?

dmsolr:
> Is this PR ready to review?

Yes.
Besides, I will adjust e2e-MySQL compatible with other storage-plugin(InfluxDB) in a new PR.

wu-sheng:
> I will adjust e2e-MySQL compatible

What is the meaning of this?

oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -146,6 +146,11 @@ storage:
 #    password: ${SW_STORAGE_PASSWORD:}
 #    database: ${SW_STORAGE_DATABASE:skywalking}
 #    metadataQueryMaxSize: ${SW_STORAGE_H2_QUERY_MAX_SIZE:5000}
+#    recordDataTTL: ${SW_STORAGE_RECORD_DATA_TTL:7} # keeps 7 days
+#    minuteMetricsDataTTL: ${SW_STORAGE_MINUTE_METRICS_DATA_TTL:64800} # Unit is minute, keeps 45 days
+#    hourMetricsDataTTL: ${SW_STORAGE_HOUR_METRICS_DATA_TTL:1080} # Unit is hour, keeps 45 days
+#    dayMetricsDataTTL: ${SW_STORAGE_DAY_METRICS_DATA_TTL:45} # Unit is day, keeps 45 days
+#    monthMetricsDataTTL: ${SW_STORAGE_MONTH_METRICS_DATA_TTL:18} # Unit is month, keeps 18 months
```
wu-sheng: If these are using the as same unit as the core module, you shouldn't add them. You could use the core's.

wu-sheng:
Recommend to add e2e for this new storage in this PR. At least it could recheck this new feature is basically ready.

test/e2e/e2e-influxdb/src/docker/rc.d/rc0-prepare.sh
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+echo "InfluxDB with H2 database is storage provider..."
+
+# Modify application.yml to set MySQL as storage provider.
```
kezhenxu94:
```suggestion
# Modify application.yml to set InfluxDB as storage provider.
```

test/e2e/e2e-influxdb/src/main/java/org/apache/skywalking/e2e/sample/client/SampleClientApplication.java
```diff
@@ -0,0 +1,34 @@
...
+
+package org.apache.skywalking.e2e.sample.client;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
+
+/**
+ * @author kezhenxu94
+ */
```
kezhenxu94: remove the `@author`s please

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/Model.java
```diff
@@ -44,5 +48,11 @@ public Model(String name, List<ModelColumn> columns, boolean capableOfTimeSeries
         this.scopeId = scopeId;
         this.name = ModelName.build(downsampling, name);
         this.record = record;
+        this.storageColumns = Maps.newTreeMap();
+        columns.forEach(column -> { storageColumns.put(column.getColumnName().getStorageName(), column); });
+    }
+
+    public ModelColumn getColumnByStorageCName(String storageCName) {
```
kezhenxu94: why add unused method

oap-server/server-storage-plugin/storage-influxdb-plugin/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+        <dependency>
+            <groupId>org.apache.skywalking</groupId>
+            <artifactId>storage-jdbc-hikaricp-plugin</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.influxdb</groupId>
+            <artifactId>influxdb-java</artifactId>
+            <version>2.15</version>
```
kezhenxu94: make it a property

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        try {
+            QueryResult result = getInflux().query(query);
+            if (result.hasError()) {
+                throw new IOException(result.getError());
+            }
+            return result.getResults();
+        } catch (Exception e) {
+            throw new IOException(e.getMessage() + "\r\nSQL Statement: " + query.getCommand(), e);
```
kezhenxu94: use `System.lineSeparator()` instead of hardcoded `\r\n`

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/MetricsDAO.java
```diff
@@ -0,0 +1,95 @@
...
+    private final InfluxClient client;
+
+    public MetricsDAO(InfluxClient client, StorageBuilder<Metrics> storageBuilder) {
+        this.client = client;
+        this.storageBuilder = storageBuilder;
+    }
+
+    @Override
+    public List<Metrics> multiGet(Model model, List<String> ids) throws IOException {
+        WhereQueryImpl query = select("*::field")
```
kezhenxu94: Raw use of parameterized class `WhereQueryImpl`

kezhenxu94:
The corresponding `yml` in `docker-entrypoint.sh` should be also updated

dmsolr:
Thanks for your review and reminder, @wu-sheng and @kezhenxu94 
I will fix them later.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java
```diff
@@ -0,0 +1,131 @@
...
+import org.influxdb.dto.Query;
+import org.influxdb.dto.QueryResult;
+import org.influxdb.querybuilder.time.TimeInterval;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.influxdb.querybuilder.BuiltQuery.QueryBuilder.ti;
+
+public class InfluxClient implements Client {
+    private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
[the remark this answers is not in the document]
wu-sheng: Agree. `@slf4j` good enough for me.

oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -98,33 +98,33 @@ storage:
...
+#    concurrentRequests: ${SW_STORAGE_ES_CONCURRENT_REQUESTS:2} # the number of concurrent requests
+#    resultWindowMaxSize: ${SW_STORAGE_ES_QUERY_MAX_WINDOW_SIZE:10000}
+#    metadataQueryMaxSize: ${SW_STORAGE_ES_QUERY_MAX_SIZE:5000}
+#    segmentQueryMaxSize: ${SW_STORAGE_ES_QUERY_SEGMENT_SIZE:200}
+#    advanced: ${SW_STORAGE_ES_ADVANCED:""}
+  h2:
+    driver: ${SW_STORAGE_H2_DRIVER:org.h2.jdbcx.JdbcDataSource}
+    url: ${SW_STORAGE_H2_URL:jdbc:h2:mem:skywalking-oap-db}
+    user: ${SW_STORAGE_H2_USER:sa}
+    metadataQueryMaxSize: ${SW_STORAGE_H2_QUERY_MAX_SIZE:5000}
```
wu-sheng: This shouldn't be changed. If start in IDE, we keep in ES

tools/dependencies/known-oap-backend-dependencies-es7.txt
```diff
@@ -159,3 +159,8 @@ sundr-core-0.9.2.jar
 swagger-annotations-1.5.12.jar
 t-digest-3.2.jar
 zookeeper-3.4.10.jar
+converter-moshi-2.5.0.jar
+influxdb-java-2.15.jar
+logging-interceptor-3.13.1.jar
+moshi-1.5.0.jar
+msgpack-core-0.8.16.jar
```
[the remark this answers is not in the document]
dmsolr: I am confusing about this. So, I need to take some time to know it.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxTTLCalculatorProvider.java
```diff
@@ -0,0 +1,102 @@
...
+ *
+ */
+package org.apache.skywalking.oap.server.storage.plugin.influxdb;
+
+import org.apache.skywalking.oap.server.core.DataTTLConfig;
+import org.apache.skywalking.oap.server.core.analysis.Downsampling;
+import org.apache.skywalking.oap.server.library.module.ModuleDefineHolder;
+import org.joda.time.DateTime;
+
+public class InfluxTTLCalculatorProvider {
```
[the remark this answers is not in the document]
dmsolr: got it.

tools/dependencies/known-oap-backend-dependencies-es7.txt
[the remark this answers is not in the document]
wu-sheng: Confused what parts? All new dependency should be added into the LICENSE, like you do for `influxdb-java`. You need to find the licenses for all other jars.

dmsolr:
> The corresponding `yml` in `docker-entrypoint.sh` should be also updated

Sorry, I don't understand. We need to update which file? @kezhenxu94

wu-sheng:
`docker-entrypoint.sh` files

oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -75,29 +75,29 @@ core:
...
+  #    monthMetricsDataTTL: ${SW_STORAGE_ES_MONTH_METRIC_DATA_TTL:18} # Unit is month
+  #    # Batch process setting, refer to https://www.elastic.co/guide/en/elasticsearch/client/java-api/5.5/java-docs-bulk-processor.html
+  #    bulkActions: ${SW_STORAGE_ES_BULK_ACTIONS:1000} # Execute the bulk every 1000 requests
+  #    flushInterval: ${SW_STORAGE_ES_FLUSH_INTERVAL:10} # flush the bulk every 10 seconds whatever the number of requests
+  #    concurrentRequests: ${SW_STORAGE_ES_CONCURRENT_REQUESTS:2} # the number of concurrent requests
+  #    resultWindowMaxSize: ${SW_STORAGE_ES_QUERY_MAX_WINDOW_SIZE:10000}
+  #    metadataQueryMaxSize: ${SW_STORAGE_ES_QUERY_MAX_SIZE:5000}
+  #    segmentQueryMaxSize: ${SW_STORAGE_ES_QUERY_SEGMENT_SIZE:200}
+  #    profileTaskQueryMaxSize: ${SW_STORAGE_ES_QUERY_PROFILE_TASK_SIZE:200}
+  #    advanced: ${SW_STORAGE_ES_ADVANCED:""}
```
wu-sheng: These formats are still changed.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java
```diff
@@ -0,0 +1,194 @@
...
+import org.influxdb.dto.Query;
+import org.influxdb.dto.QueryResult;
+import org.influxdb.querybuilder.time.TimeInterval;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.influxdb.querybuilder.BuiltQuery.QueryBuilder.ti;
+
+/**
+ *
```
wu-sheng: Empty comment?

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java
```diff
@@ -0,0 +1,194 @@
...
+
+    /**
+     * Request with one statement to InfluxDB and return a set of {@link QueryResult.Series}s.
+     *
+     * @param query
+     * @return a set of Series
+     * @throws IOException
+     */
+    public List<QueryResult.Series> queryForSeries(Query query) throws IOException {
+        return query(query).get(0).getSeries();
```
wu-sheng: What does this `#get(0)` represent?

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java
```diff
@@ -0,0 +1,194 @@
...
+        }
+    }
+
+    /**
+     * Data management, to delete data by a statement. If an exception isn't thrown, it means execution success.
+     *
+     * @param statement
+     * @throws IOException
+     */
+    public void queryForDelete(String statement) throws IOException {
```
wu-sheng: Don't see anyone uses this. Please confirm. And `queryForDelete` is not a good idea. For time-series data, it is better to delete by time bucket only.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxStorageConfig.java
```diff
@@ -0,0 +1,41 @@
...
+
+package org.apache.skywalking.oap.server.storage.plugin.influxdb;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.apache.skywalking.oap.server.library.module.ModuleConfig;
+
+@Setter
+@Getter
+public class InfluxStorageConfig extends ModuleConfig {
```
wu-sheng: Comments are required.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/BatchDAO.java
```diff
@@ -0,0 +1,105 @@
...
+        if (log.isDebugEnabled()) {
+            log.debug("batch sql statements execute, data size: {}", prepareRequests.size());
+        }
+
+        final BatchPoints.Builder builder = BatchPoints.builder();
+        prepareRequests.forEach(e -> {
+            builder.point(((InfluxInsertRequest)e).getPoint());
+        });
+
+        client.write(builder.build());
```
wu-sheng: Is this a blocking write? Meaning, the data is queryable when this method finished.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/BatchDAO.java
```diff
@@ -0,0 +1,105 @@
...
+            throw new UnexpectedException(e.getMessage(), e);
+        }
+
+        this.dataCarrier = new DataCarrier(1, 10000);
+        this.dataCarrier.consume(ConsumerPoolFactory.INSTANCE.get(name), new InfluxBatchConsumer(this));
+    }
+
+    @Override
+    public void asynchronous(InsertRequest insertRequest) {
+        dataCarrier.produce(insertRequest);
```
wu-sheng: Doesn't InfluxDB have an async write mode? I remember time serious DB is good at writing.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java
```diff
@@ -0,0 +1,194 @@
...
+    /**
+     * Data management, to drop a time-series by measurement and time-series name specified. If an exception isn't
+     * thrown, it means execution success.
+     *
+     * @param measurement
+     * @param timeBucket
+     * @throws IOException
+     */
+    public void dropSeries(String measurement, long timeBucket) throws IOException {
+        Query query = new Query("DROP SERIES FROM " + measurement + " WHERE time_bucket='" + timeBucket + "'");
```
wu-sheng: `=`?

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/MetricsDAO.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        final List<Metrics> metrics = Lists.newArrayList();
+        List<String> columns = series.get(0).getColumns();
+        Map<String, String> storageAndColumnNames = Maps.newHashMap();
+        for (ModelColumn column : model.getColumns()) {
+            storageAndColumnNames.put(column.getColumnName().getName(), column.getColumnName().getStorageName());
+        }
+
+        series.get(0).getValues().forEach(values -> {
```
wu-sheng: Another `#get(0)`. What does this mean?

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java
dmsolr: InfluxDB support user to send a request with multi statements. This is for a single statement in a request(query).
So, `#get(0)` means get the first result of the query.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java
dmsolr: Yes, I will remove

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java
wu-sheng: Do we use multiple statements? I think `#get(0)` should be included `#query`(return series). If we have multiple stat case, we should have a different API.

dmsolr: Yes.
In `ITraceQueryDAO#queryBasicTraces(...)`, we have 2 statements in a request. The one gets the number of traces, and another recalls all trace data.

wu-sheng: Could we add a `#singleStatementQuery` method for most cases?

dmsolr: Sorry, I make a mistake. This method works for `single statement`. `InfluxClient#query()` is for `multi-statements`

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/BatchDAO.java
dmsolr: Yes, it does. I will fix.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/BatchDAO.java
dmsolr: The data is non-blocking write. They are queryable after the method executed.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/BatchDAO.java
wu-sheng: Then you should not require datacarrier? Right?

docs/en/setup/backend/backend-storage.md
```diff
@@ -226,6 +227,27 @@ storage:
 All connection related settings including link url, username and password are in `application.yml`. 
 These settings can refer to the configuration of *MySQL* above.
 
+## InfluxDB
+InfluxDB as storage since SkyWalking 7.0. It depends on `H2/MySQL` storage-plugin to store `metadata` likes `Inventory` and `ProfileTask`. So, when we set `InfluxDB` as storage provider, we need to configure `InfluxDB`'s properties and `H2/MySQL`.
```
wu-sheng: `likes` -> `like`.
`we need to configure InfluxDB's properties and H2/MySQL` -> `We need to configure properties of InfluxDB and Metabase.

docs/en/setup/backend/backend-storage.md
```diff
@@ -226,6 +227,27 @@ storage:
...
+
+```yaml
+storage
+  influx:
+    # Metadata storage provider configuration
+    metabaseType: ${SW_STORAGE_METABASE_TYPE:H2}
```
wu-sheng: Please provide type options.

wu-sheng:
@kezhenxu94 I don't want the e2e use submodule actually. Setting static proto files is good for compatible check and easy to check. I think we talked about this before, right?

For test reactor, it is not a block. It is test optimization.

wu-sheng:
I don't want this big PR blocked too long, it is last a long time, we could create an issue to track your test change proposal.

test/e2e/e2e-ttl/e2e-ttl-influxdb/src/main/proto/common/CLR.proto
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+syntax = "proto3";
```
wu-sheng: All these protos should be removed, please follow this, https://github.com/apache/skywalking/pull/4329#issue-372713174

dmsolr:
> All `proto` files under `test/e2e` should be removed, instead, add the protocol repository as submodule, just like what `e2e-ttl-es` does, otherwise, they may be outdated
> 
> And I hope you could refactor the `e2e-ttl-es` to something like `e2e-ttl` that can be reused for multiple storage types, like `e2e-profile`, we can use the same codes to test multiple storage types

I am planning to do it, but it is a low priority in my schedule.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java
```diff
@@ -0,0 +1,203 @@
...
+        influx.enableBatch(config.getActions(), config.getDuration(), TimeUnit.MILLISECONDS);
+        influx.setDatabase(database);
+    }
+
+    /**
+     * To get a connection of InfluxDB.
+     *
+     * @return InfluxDB's connection
+     */
+    public InfluxDB getInflux() {
```
wu-sheng: I think this should be `private` to avoid misuse

wu-sheng:
Profile e2e is failing, please recheck.

wu-sheng:
#4335 is going to be merged today. There would be the last two new storage methods introduced.

wu-sheng:
The strict format rule has been merged, please update this PR.

.github/workflows/e2e.yaml
```diff
@@ -39,19 +39,21 @@ jobs:
...
+        run: export E2E_VERSION=jdk11-1.4 && bash -x test/e2e/run.sh e2e-single-service
       - name: Single Node Tests(JDK12)
-        run: export E2E_VERSION=jdk12-1.3 && bash -x test/e2e/run.sh e2e-single-service
+        run: export E2E_VERSION=jdk12-1.4 && bash -x test/e2e/run.sh e2e-single-service
       - name: Agent Reboot Tests(JDK8)
-        run: export E2E_VERSION=jdk8-1.3 && bash -x test/e2e/run.sh e2e-agent-reboot
+        run: export E2E_VERSION=jdk8-1.4 && bash -x test/e2e/run.sh e2e-agent-reboot
 
-  Cluster:
+  Cluster_N_TTL:
```
wu-sheng: What do you mean N_TTL?

kezhenxu94: > What do you mean N_TTL?

Possibly `N` == `And`, we don't need to save 2 letters to cause confusion, no?

wu-sheng: Three things
1. N represent AND is not a common case. `+` and `&` are better
1. This name should not be changed randomly, `Cluster` is a required test, without this check(named as `cluster`), we can't merge PR unless we request the INFRA change the settings.
1. If you want the test group names more clear, moving the TTL out of the cluster test group makes sense to me. But notice, it will take more time as it requires to compile the project again.

.github/workflows/e2e.yaml
```diff
@@ -39,19 +39,21 @@ jobs:
           ./mvnw --batch-mode -Dcheckstyle.skip -Drat.skip -T2 -Dmaven.compile.fork -Dmaven.compiler.maxmem=3072 -DskipTests clean install
           ./mvnw --batch-mode -f test/e2e/pom.xml -pl e2e-base clean install
       - name: Single Node Tests(JDK8)
-        run: export E2E_VERSION=jdk8-1.3 && bash -x test/e2e/run.sh e2e-single-service
+        run: export E2E_VERSION=jdk8-1.4 && bash -x test/e2e/run.sh e2e-single-service
```
wu-sheng: Upgrade to 1.5 please, based on, https://github.com/apache/skywalking/issues/4343#issuecomment-584990302

.github/workflows/e2e.yaml
dmsolr: Yes. I will revert it.

.github/workflows/e2e.yaml
dmsolr: Got it.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java
```diff
@@ -0,0 +1,210 @@
...
+
+    public void deleteByQuery(String measurement, long timestamp) throws IOException {
+        this.query(new Query("delete from " + measurement + " where time < " + timestamp + "ms"));
+    }
+
+    /**
+     * Write a {@link Point} into InfluxDB. Note that, the {@link Point} is written into buffer of InfluxDB Client and
+     * wait for buffer flushing.
+     *
+     * @param point Point
```
wu-sheng: From here, all comments about @param and @return are either empty and meaningless. Could you fix this?

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/PointBuilder.java
```diff
@@ -0,0 +1,138 @@
...
+import static org.apache.skywalking.oap.server.core.source.DefaultScopeDefine.DATABASE_SLOW_STATEMENT;
+import static org.apache.skywalking.oap.server.core.source.DefaultScopeDefine.HTTP_ACCESS_LOG;
+import static org.apache.skywalking.oap.server.core.source.DefaultScopeDefine.PROFILE_TASK_LOG;
+import static org.apache.skywalking.oap.server.core.source.DefaultScopeDefine.PROFILE_TASK_SEGMENT_SNAPSHOT;
+import static org.apache.skywalking.oap.server.core.source.DefaultScopeDefine.SEGMENT;
+
+/**
+ * A helper help to build a InfluxDB Point from StorageData.
+ */
+public class PointBuilder {
```
wu-sheng: Could you explain this class a little more? I am a little confused. All existing entities have the shared `prepareBatchInsert` and `prepareBatchUpdate` logic without explicit entity name required. Using the literal string name here is highly unstable. These names could be changed in any PR, and new entity could be added.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java
kezhenxu94: @dmsolr you should put the **description** after the parameter name and return name, **NOT** the type of the parameter and return object, they're meaningless, and will cheat the JavaDoc compiler, if they are already self-documented, simply remove the JavaDoc.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/LogQuery.java
```diff
@@ -0,0 +1,154 @@
...
+            }
+        }
+        if (!Strings.isNullOrEmpty(stateCode)) {
+            recallQuery.and(eq(STATUS_CODE, stateCode));
+        }
+        recallQuery.and(gte(AbstractLogRecord.TIME_BUCKET, startTB))
+                   .and(lte(AbstractLogRecord.TIME_BUCKET, endTB));
+
+        if (from > Const.NONE) {
+            limit += from;
```
kezhenxu94: why do you need this? seems the `limit` doesn't include the `offset`

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/TraceQuery.java
```diff
@@ -0,0 +1,217 @@
...
+            basicTrace.getTraceIds().add((String) values.get(7));
+
+            traceBrief.getTraces().add(basicTrace);
+        });
+        return traceBrief;
+    }
+
+    @Override
+    public List<SegmentRecord> queryByTraceId(String traceId) throws IOException {
+        WhereQueryImpl query = select().column(SegmentRecord.SEGMENT_ID)
```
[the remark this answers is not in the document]
dmsolr: Could you explain more? `WhereQueryImpl` is harmless and easy to append clauses.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/MetricsDAO.java
```diff
@@ -0,0 +1,106 @@
...
+
+    public MetricsDAO(InfluxClient client, StorageBuilder<Metrics> storageBuilder) {
+        this.client = client;
+        this.storageBuilder = storageBuilder;
+    }
+
+    @Override
+    public List<Metrics> multiGet(Model model, List<String> ids) throws IOException {
+        WhereQueryImpl<SelectQueryImpl> query = select()
+            .regex("*::field")
```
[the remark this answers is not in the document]
dmsolr: They are different in name only. 
If you think `raw` is more appropiate, I will fix them.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/TraceQuery.java
```diff
@@ -0,0 +1,217 @@
...
+
+            basicTrace.setSegmentId((String) values.get(2));
+            basicTrace.setStart(String.valueOf((long) values.get(3)));
+            basicTrace.getEndpointNames().add((String) values.get(4));
+            basicTrace.setDuration((int) values.get(5));
+            basicTrace.setError(BooleanUtils.valueToBoolean((int) values.get(6)));
+            basicTrace.getTraceIds().add((String) values.get(7));
+
+            traceBrief.getTraces().add(basicTrace);
+        });
```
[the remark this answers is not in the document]
dmsolr: It took me a lot of time to get it clearly. It is a bad story.

1. top/bottom cannot work with limit.
2. I think top/bottom collects data by PriorityQueue. And it does not re-sort them in the result set. So we get the dataset unorder. Anyway, we always get the un-ordered result set in reality.

I mean, the function `top` works after `limit`, if limit enabled.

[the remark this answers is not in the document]
kezhenxu94: 1. Resolved
2. The question is more about why you sort them in a different order without taking care of the given `orderBy`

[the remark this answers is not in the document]
kezhenxu94: 2. And there may be performance issue when paging deeply

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/TraceQuery.java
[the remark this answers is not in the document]
kezhenxu94: > Could you explain more? `WhereQueryImpl` is harmless and easy to append clauses.

- Program to `interface` not `Impl`
- `WhereQueryImpl` is generic-typed, but you simply ignored them, causing many warnings

Anyway, just ignore if you don’t care the warnings

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/TraceQuery.java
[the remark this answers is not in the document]
dmsolr: I got it. InfluxDB only ORDER BY time supported at this time. As far as I know, most of TSDB have the same issue.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/ProfileThreadSnapshotQuery.java
```diff
@@ -0,0 +1,165 @@
...
+        client.queryForSingleSeries(query).getValues().forEach(values -> {
+            ProfileThreadSnapshotRecord record = new ProfileThreadSnapshotRecord();
+
+            record.setTaskId((String) values.get(1));
+            record.setSegmentId((String) values.get(2));
+            record.setDumpTime(((Number) values.get(3)).longValue());
+            record.setSequence((int) values.get(4));
+            String dataBinaryBase64 = String.valueOf(values.get(5));
+            if (StringUtil.isNotEmpty(dataBinaryBase64)) {
+                record.setStackBinary(Base64.getDecoder().decode(dataBinaryBase64));
```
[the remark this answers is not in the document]
dmsolr: This class is included in skywaking-agent-core module. We don't import it.

.github/workflows/e2e.yaml
```diff
@@ -71,15 +73,19 @@ jobs:
           ./mvnw --batch-mode -Dcheckstyle.skip -Drat.skip -T2 -Dmaven.compile.fork -Dmaven.compiler.maxmem=3072 -DskipTests clean install
           ./mvnw --batch-mode -f test/e2e/pom.xml -pl e2e-base clean install
       - name: Cluster Tests (ES6/ZK/JDK8)
-        run: export E2E_VERSION=jdk8-1.3 && bash -x test/e2e/run.sh e2e-cluster/e2e-cluster-test-runner
+        run: export E2E_VERSION=jdk8-1.5 && bash -x test/e2e/run.sh e2e-cluster/e2e-cluster-test-runner --storage=elasticsearch
+      - name: Cluster Tests (InfluxDB/ZK/JDK8)
+        run: export E2E_VERSION=jdk8-1.5 && bash -x test/e2e/run.sh e2e-cluster/e2e-cluster-test-runner --storage=influxdb
       - name: Cluster With Gateway Tests (ES6/ZK/JDK8)
-        run: export E2E_VERSION=jdk8-1.3 && bash -x test/e2e/run.sh e2e-cluster-with-gateway/e2e-cluster-with-gateway-test-runner
+        run: export E2E_VERSION=jdk8-1.5 && bash -x test/e2e/run.sh e2e-cluster-with-gateway/e2e-cluster-with-gateway-test-runner
```
wu-sheng: Why this doesn't have `--storage=elasticsearch`? What is the default? Or what happens if no `--storage=`?

I expect a fail when missing this parameter.

kezhenxu94: > Why this doesn't have `--storage=elasticsearch`? What is the default? Or what happens if no `--storage=`?

Not every e2e case is aware of the parameter `--storage`, `e2e-cluster-with-gateway` is one of them, it's intruduced recently and only `e2e-cluster` and `e2e-profile` are refactored then, IIRC

dmsolr: Not add e2e-cluster-with-gateway test yet. Do we need it?

By the way, you are right, @kezhenxu94 .
(I am trying to add more testcases into e2e-ttl-influxdb, so I have not refactored it)

wu-sheng: I am OK with keeping the current status.

wu-sheng:
The ElasticSearch fails twice, I am rerunning it.

wu-sheng:
@dmsolr Please recheck. The ElasticSearch test fails agian. I don't find any change related to you. Please take a look, what is the issue.

dmsolr:
> @dmsolr Please recheck. The ElasticSearch test fails agian. I don't find any change related to you. Please take a look, what is the issue.

I had checked it yesterday. The error message is "No space leave on device". And this exception threw on other PR.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/installer/MetaTableDefine.java
```diff
@@ -0,0 +1,55 @@
...
+     * @param model Model
+     * @return true if the {@link Model} is stored in H2/MySQL
+     */
+    public static boolean contains(Model model) {
+        switch (model.getScopeId()) {
+            case SERVICE_INVENTORY:
+            case SERVICE_INSTANCE_INVENTORY:
+            case NETWORK_ADDRESS:
+            case ENDPOINT_INVENTORY:
+            case PROFILE_TASK:
```
wu-sheng: There is another missing here, `PROFILE_TASK_LOG`. Could you check why this is missed but still tests passed? Does InfluxDB have the profile e2e?

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/AggregationQuery.java
```diff
@@ -0,0 +1,152 @@
...
+
+        List<List<Object>> dataset = series.get(0).getValues();
+        List<TopNEntity> entities = Lists.newArrayListWithCapacity(dataset.size());
+        dataset.forEach(values -> {
+            final TopNEntity entity = new TopNEntity();
+            entity.setId((String) values.get(2));
+            entity.setValue(((Double) values.get(1)).longValue());
+            entities.add(entity);
+        });
+        Collections.sort(entities, comparator);
```
wu-sheng: As you mentioned last time, the result is unordered. Could you add the comments some places to ease the readers?

test/e2e/e2e-ttl/e2e-ttl-influxdb/src/test/java/org/apache/skywalking/e2e/StorageTTLITCase.java
```diff
@@ -0,0 +1,285 @@
...
+import java.time.ZoneOffset;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+
+import static org.apache.skywalking.e2e.metrics.MetricsQuery.SERVICE_RESP_TIME;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author kezhenxu94
+ */
```
kezhenxu94: remove this

wu-sheng: Why isn't this detected by CI process?

kezhenxu94: > Why isn't this detected by CI process?

CI belongs to the root module and its submodules, e2e and plugin tests are neither of them, maybe set up the same Checkstyle plugin in e2e and plugin tests?

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/installer/MetaTableDefine.java
dmsolr: profile_task_log is not metadata. It is stored on InfluxDB.

test/e2e/e2e-ttl/e2e-ttl-influxdb/src/test/java/org/apache/skywalking/e2e/StorageTTLITCase.java
dmsolr: We don't check `./test` directory.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/installer/MetaTableDefine.java
wu-sheng: What is the principle? Could you add some comments and guidelines about this?

test/e2e/e2e-ttl/e2e-ttl-influxdb/src/test/java/org/apache/skywalking/e2e/StorageTTLITCase.java
wu-sheng: I think we should do that. Creating an issue to track this?

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/installer/MetaTableDefine.java
wu-sheng: Use `model#capableOfTimeSeries` == true check. If YES, then it is a metadata table.

wu-sheng: And `PROFILE_TASK` is not a metadata entity

wu-sheng:
e2e fail because the profile task DAO is still accessing the H2/MySQL. It seems you miss a change. Please fix it.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java
```diff
@@ -0,0 +1,191 @@
...
+        influx.setDatabase(database);
+    }
+
+    /**
+     * To get a connection of InfluxDB.
+     *
+     * @return InfluxDB's connection
+     */
+    private InfluxDB getInflux() {
+        return influx;
```
aderm: recommend add a null value to judge here，subsequence queries are based on the available influx

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java
```diff
@@ -0,0 +1,191 @@
...
+    public void dropSeries(String measurement, long timeBucket) throws IOException {
+        Query query = new Query("DROP SERIES FROM " + measurement + " WHERE time_bucket='" + timeBucket + "'");
+        QueryResult result = getInflux().query(query);
+
+        if (result.hasError()) {
+            throw new IOException("Statement: " + query.getCommand() + ", ErrorMsg: " + result.getError());
+        }
+    }
+
+    public void deleteByQuery(String measurement, long timestamp) throws IOException {
```
aderm: change method name `deleteBefore` or else better？

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/HistoryDeleteDAO.java
```diff
@@ -0,0 +1,71 @@
...
+            TTLCalculator ttlCalculator;
+            if (model.isRecord()) {
+                ttlCalculator = storageTTL.recordCalculator();
+            } else {
+                ttlCalculator = storageTTL.metricsCalculator(model.getDownsampling());
+            }
+
+            client.deleteByQuery(
+                model.getName(),
+                TimeBucket.getTimestamp(ttlCalculator.timeBefore(DateTime.now(), configService.getDataTTLConfig()) + 1)
```
aderm: +1 can be deleted？ delete is <

dmsolr: The condition, we delete the history data, is `less than and equal`.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java
dmsolr: I don't think so. OAP crashes and exits, when Influx initializes failed.

wu-sheng:
We are waiting for nearly 24 hours, we will merge soon. Anyone has suggestions feels free to submit a separated PR

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/NoneStreamDAO.java
```diff
@@ -0,0 +1,57 @@
...
+        this.storageBuilder = storageBuilder;
+    }
+
+    @Override
+    public void insert(final Model model, final NoneStream noneStream) throws IOException {
+        final long timestamp = TimeBucket.getTimestamp(
+            noneStream.getTimeBucket(), model.getDownsampling()) * PADDING_SIZE + SUFFIX.getAndIncrement();
+
+        Point point = new InfluxInsertRequest(model, noneStream, storageBuilder)
+            .time(timestamp, TimeUnit.NANOSECONDS)
```
aderm: why this have Nanosecond? or millisecond?

wu-sheng: All data in influxdb is nanosecond.

PR #4244, opened 2020-01-16
Drop support of protocol v1 (sw3)

### LINKED ISSUE #4176: Remove all 5.x protocol and sw3 header from documents and codes
According to 

### Deprecated services
Deprecated service(s) are the gRPC services SkyWalking used before. In SkyWalking v6, in order to match the common concepts in CloudNative world, these services are deprecated. Although there services are still supported at this moment, but it will keep in supported at least before the end of 2019.

___
Now, it is 2020, we should remove all these permanently. If no objection, I will try to do this.

kezhenxu94:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance
- [x] Chore

- Related issues

closes #4176 

protocol removal will be done in the protocol repository


wu-sheng:
> data-collect-protocol also dropped or updated to latest?

I think should be before :)

JaredTan95:
> protocol removal will be done in the protocol repository

Sorry, I missed this.

wu-sheng:
@kezhenxu94 MySQL e2e test seems failure, maybe due to some 3rd party issue?

kezhenxu94:
> @kezhenxu94 MySQL e2e test seems failure, maybe due to some 3rd party issue?

E2E MySQL test seems to fail in all PRs, I'm checking

wu-sheng:
> E2E MySQL test seems to fail in all PRs, I'm checking

Yes. Thanks.

wu-sheng:
We should clean https://github.com/apache/skywalking-data-collect-protocol of v1. But should we do after #4220?

wu-sheng:
Let's back to this
> I think you miss TraceQueryService? There is a version=1 check there.

kezhenxu94:
> Let's back to this
> 
> > I think you miss TraceQueryService? There is a version=1 check there.

Are you still checking the cached page of GitHub? I’m sure I had already removed it

wu-sheng:
I thought you haven't worked on that. I will recheck today

apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/context/ContextCarrierV2HeaderTest.java
```diff
@@ -107,37 +58,19 @@ public void testSerializeV2Header() {
...
-                    Assert.fail("unexpected key");
-                }
-            }
-
-        } finally {
-            Config.Agent.ACTIVE_V1_HEADER = false;
+        next = contextCarrier.items();
+        while (next.hasNext()) {
+            next = next.next();
+            //TODO, no BASE64
```
wu-sheng: What is this TODO for?

wu-sheng:
> protocol removal will be done in the protocol repository
> Sorry, I missed this.

@kezhenxu94, what do you mean about `will be`? Are you going to provide another pull request to update the submodule? I prefer it should be done on this one only, to make sure the removal is completed. FYI@JaredTan95

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/parser/listener/segment/SegmentSpanListener.java
```diff
@@ -81,10 +82,7 @@ public void parseFirst(SpanDecorator spanDecorator, SegmentCoreInfo segmentCoreI
         segment.setIsError(BooleanUtils.booleanToValue(segmentCoreInfo.isError()));
         segment.setTimeBucket(timeBucket);
         segment.setDataBinary(segmentCoreInfo.getDataBinary());
-        /**
-         * Only consider v1, v2 compatible for now.
-         */
-        segment.setVersion(segmentCoreInfo.isV2() ? 2 : 1);
+        segment.setVersion(segmentCoreInfo.getVersion().number);
```
wu-sheng: `segmentCoreInfo.getVersion().number` Why write in this way? `.number`?

Issue #4230, opened 2020-01-14
potential thread-unsafe problem in server-alarm-plugin of oap-server module


liuhaoXD:
https://github.com/apache/skywalking/blob/f9cf17be24304bdf4afef423cbb0d5e4925f668f/oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java#L232

* `this.moveTo()`at line 234  is thread-unsafe.
* `minutesBetween` operation should be protected by lock()
* `if` statement at  line 233 is duplicate with line 240

wu-sheng:
Could you submit the issue in English or send a pull request to fix the issue?

PR #4247, opened 2020-01-16
fix thread unsafe problem in server-alarm-plugin (#4230)

liuhaoXD:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
#4230
___
### Bug fix
- Bug description.
fix thread-unsafe operation

- How to fix?

___


wu-sheng:
Two suggestions.
1. `#add` should not include `#moveTo`, as in `#in`, we move the time window first, then call add. 
1. This method should only be called by a single thread. So basically, no concurrency access. This lock is for fail-safe.

Could you follow these to update this PR?

wu-sheng:
Any update?

wu-sheng:
CI fails, please recheck.

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
```diff
@@ -187,7 +182,7 @@ public void moveTo(LocalDateTime targetTime) {
         private int silenceCountdown;
 
         private LinkedList<Metrics> values;
-        private ReentrantLock lock = new ReentrantLock();
+        private ReentrantLock reentrantLock = new ReentrantLock();
```
wu-sheng: Why change this name?

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
```diff
@@ -86,7 +83,7 @@ public RunningRule(AlarmRule alarmRule) {
      * Receive metrics result from persistence, after it is saved into storage. In alarm, only minute dimensionality
      * metrics are expected to process.
      *
-     * @param meta of input metrics
+     * @param meta    of input metrics
```
wu-sheng: Why change this?

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
```diff
@@ -290,28 +278,28 @@ private boolean isMatch() {
...
-                        double dvalue = ((DoubleValueHolder)metrics).getValue();
+                        double dvalue = ((DoubleValueHolder) metrics).getValue();
                         double dexpected = RunningRule.this.threshold.getDoubleThreshold();
                         if (op.test(dexpected, dvalue)) {
                             matchCount++;
                         }
                         break;
                     case MULTI_INTS:
-                        int[] ivalueArray = ((MultiIntValuesHolder)metrics).getValues();
+                        int[] ivalueArray = ((MultiIntValuesHolder) metrics).getValues();
```
wu-sheng: What change the format of all these cases?

liuhaoXD:
> Any update?

Actually, I think that `#moveTo` when doing `#in` is redundant, which can be covered by `#add` operation, and `#moveTo` in `#add` is an essential mechanism to make `public void add()` more safe at `window`'s initial stage, another PR will be created today :).

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
liuhaoXD: alibaba coding guideline plugin did these coding format automatically..

wu-sheng: But we are not using it .SkyWalking has its own format, please revert those.

wu-sheng:
> Actually, I think that #moveTo when doing #in is redundant

I think you have removed it.

wu-sheng:
The PR could be accepted when you make all CI passed. Please fix them.

wu-sheng:
@liuhaoXD Could you follow another two format change requests?

wu-sheng:
If this merge in time, you will be our 200th contributor on this main repo.

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
```diff
@@ -227,18 +221,17 @@ public void moveTo(LocalDateTime current) {
...
-
-            }
+            LocalDateTime timeBucket = TIME_BUCKET_FORMATTER.parseLocalDateTime(bucket + "");
 
-            lock.lock();
+            this.lock.lock();
             try {
+                if (this.endTime == null) {
+                    init();
+                    this.endTime = timeBucket;
```
wu-sheng: Why add this? End time has been initialized in the constructor. End time should never be null. Right?

kezhenxu94:
> LGTM. Format issues related to #4272. I will ignore this for now.
> 
> @kezhenxu94 Recheck?

Will check later

PR #4250, opened 2020-01-16
Issue about load balance of gRPC

bennyparlo:
Fixed issue about loading balance of gRPC on kubenetes via this line of code. If the clients of AGENT deploy on kubernetes and call the OAP service by default cluster IP mode without any technology of service mesh like istio etc. it will occur the issue about traffic load. So for we have resolved this issue via the headless service difination without the clust IP setting and registered multiple A records of DNS for the OAP service on that. Anyway it needs to enhance the code of this class for the defualt behavior of load balance and almost has not any impact if it is out of kubernetes environment. Seems to work smoothly

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___


wu-sheng:
Seems you commit wrong things.

bennyparlo:
Closed

PR #4251, opened 2020-01-16
Issue about load balance of gRPC

bennyparlo:
Fixed issue about loading balance of gRPC on kubenetes via this line of code. If the clients of AGENT deploy on kubernetes and call the OAP service by default cluster IP mode without any technology of service mesh like istio etc. it will occur the issue about traffic load. So for we have resolved this issue via the headless service difination without the clust IP setting and registered multiple A records of DNS for the OAP service on that. Anyway it needs to enhance the code of this class for the defualt behavior of load balance and almost has not any impact if it is out of kubernetes environment. Seems to work smoothly.

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/StandardChannelBuilder.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/StandardChannelBuilder.java
@@ -20,6 +20,7 @@ package org.apache.skywalking.apm.agent.core.remote;
 
 import io.grpc.ManagedChannelBuilder;
 import io.grpc.internal.DnsNameResolverProvider;
+import io.grpc.util.RoundRobinLoadBalancerFactory;
 
 /**
  * @author zhang xin
@@ -30,6 +31,10 @@ public class StandardChannelBuilder implements ChannelBuilder {
 
     @Override public ManagedChannelBuilder build(ManagedChannelBuilder managedChannelBuilder) throws Exception {
         return managedChannelBuilder.nameResolverFactory(new DnsNameResolverProvider())
+            /**
+             * It seems the code above can only work for low gRPC version.
+             */
+            .loadBalancerFactory(RoundRobinLoadBalancerFactory.getInstance())
             .maxInboundMessageSize(MAX_INBOUND_MESSAGE_SIZE)
             .usePlaintext(USE_PLAIN_TEXT);
     }
--- a/.github/workflows/ci-it.yaml
+++ b/.github/workflows/ci-it.yaml
@@ -21,6 +21,7 @@ on:
   push:
     branches: 
       - master
+      - 6.x
     tags:
       - 'v*'
     
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -21,6 +21,7 @@ on:
   push:
     branches:
       - master
+      - 6.x
     tags:
       - 'v*'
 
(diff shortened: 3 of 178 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/StandardChannelBuilder.java
```diff
@@ -30,6 +31,10 @@
 
     @Override public ManagedChannelBuilder build(ManagedChannelBuilder managedChannelBuilder) throws Exception {
         return managedChannelBuilder.nameResolverFactory(new DnsNameResolverProvider())
+            /**
+             * It seems the code above can only work for low gRPC version.
```
wu-sheng: What does this comment mean?

wu-sheng: We don't need to describe the history.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/StandardChannelBuilder.java
```diff
@@ -30,6 +31,10 @@
...
+             */
+            .loadBalancerFactory(RoundRobinLoadBalancerFactory.getInstance())
```
wu-sheng: According to the header of RoundRobinLoadBalancerFactory, this is experimental. And also, this could use more connections of OAP backend, you should create a Config, at `Config#Collector#OPEN_ROUND_ROBIN_LOAD_BALANCE` default false, to support this feature. 
```
/**
 * A {@link LoadBalancer} that provides round-robin load balancing mechanism over the
 * addresses from the {@link NameResolver}.  The sub-lists received from the name resolver
 * are considered to be an {@link EquivalentAddressGroup} and each of these sub-lists is
 * what is then balanced across.
 */
@ExperimentalApi("https://github.com/grpc/grpc-java/issues/1771")
public class RoundRobinLoadBalancerFactory extends LoadBalancer.Factory {
```

wu-sheng: Also, this doc should be updated as new config added. https://github.com/apache/skywalking/blob/master/docs/en/setup/service-agent/java-agent/README.md#table-of-agent-configuration-properties

bennyparlo: OK. Will add these details ASAP

kezhenxu94:
@wu-sheng I guess there're something wrong with the GitHub settings, the required checks are failed and no approvals from committers, the merge button still shows, can you inquire the Infra team why it's the case, please?

image

wu-sheng:
@kezhenxu84 I feel strange too, just created this, https://issues.apache.org/jira/browse/INFRA-19734

wu-sheng:
@kezhenxu94 Strange, https://github.com/apache/skywalking/pull/4220 shows checks are existing.

wu-sheng:
This PR is targeting the wrong branch, so this happens. @bennyparlo You should send a pull request to the master branch, as this is a feature change. 6.x is in maintenance mode only.

PR #4264, opened 2020-01-17
Refactor SnifferConfigInitializer and related componets with new features in JDK8+

JohnNiang:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___

Refactor SnifferConfigInitializer and related componets with new features in JDK8+


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
```diff
@@ -145,24 +165,21 @@ public static boolean isInitCompleted() {
     }
 
     /**
-     * Override the config by system properties. The property key must start with `skywalking`, the result should be as same
-     * as in `agent.config`
+     * Override the config by system properties. The property key must start with `skywalking`, the result should be as
+     * same as in `agent.config`
```
wu-sheng: The format changed again.

JohnNiang:
@wu-sheng Sorry about all format changes. I'll fix them later.

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/ConfigInitializer.java
```diff
@@ -105,30 +114,18 @@ private static List convert2List(String value) {
 }
 
 class ConfigDesc {
-    private LinkedList<String> descs = new LinkedList<String>();
+    private Stack<String> descs = new Stack<>();
```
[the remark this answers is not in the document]
JohnNiang: Got it. LinkedList will be back later.

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/StringUtil.java
```diff
@@ -27,6 +27,24 @@ public static boolean isNotEmpty(String str) {
...
+            }
+        }
+
+        return true;
+    }
+
+    public static boolean isNotBlank(CharSequence str) {
+        return !isBlank(str);
+    }
+
```
kezhenxu94: If we don't have to manipulate `CharSequence`, just reuse `org.apache.skywalking.apm.util.StringUtil#isNotEmpty`, only add codes when you really need it

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
```diff
@@ -172,24 +189,64 @@ private static void overrideConfigBySystemProp() throws IllegalAccessException {
...
         }
         throw new ConfigNotFoundException("Failed to load agent.config.");
     }
+
+    /**
+     * Option container.
+     *
+     * @author johnniang
+     */
+    private static class Option {
```
kezhenxu94: Use Lombok annotations to get rid of boilerplate codes (getters setters)

wu-sheng:
CI fails, please fix those.

wu-sheng:
> we may have to differentiate the meaning between Empty and Blank

What is the difference? And what is your case?

JohnNiang:
> > we may have to differentiate the meaning between Empty and Blank
> 
> What is the difference? And what is your case?

`Empty` means that the string contains nothing or is null, while `Blank` is not only `Empty` string, but also contains only `whitespace character(s)`.

test case: 

1. Empty: `null or ""` only;
2. Blank: `null, "",  " ", or any length of whitespace character` (Of course, the `\t`, `\n` , `\r` and `\f` are treated as whitespace character).

wu-sheng:
> Empty means that the string contains nothing or is null, while Blank is not only Empty string, but also contains only whitespace character(s).

I think @kezhenxu94 's point is why, I have the same concern too. Why do we need to separate these two? First, the `Blank` cost more resources, what is the benefit? Then, allowing these two existing will make the further reviews(of other PRs) confusing. It is hard to tell from all reviewers perspective, which one should be used in which cases.

If there isn't a case, required this, which I think there isn't as this PR didn't add a new feature, but refactor, I hope we don't add these new APIs.

JohnNiang:
###### 1: 

> the Blank cost more resources, what is the benefit?

In many cases, the mothod `org.apache.skywalking.apm.util.StringUtil#isBlank` won't cost more resources. If first character of the string isn't white space, this method returns `false` immediately.

Also, we need to check if the method `java.lang.Character#isWhitespace(int)` inside JDK takes time.

###### 2:

> allowing these two existing will make the further reviews(of other PRs) confusing. It is hard to tell from all reviewers perspective, which one should be used in which cases.

Maybe I should add java documents on two methods.

###### 3:

> If there isn't a case, required this, which I think there isn't as this PR didn't add a new feature, but refactor, I hope we don't add these new APIs.

Yeah, this PR doesn't contain any `new feature(for skywalking)`, just a refactor with the `new feature of JDK8+`. (BTW, I have never said this PR contains any new feature. : P

wu-sheng:
> Yeah, this PR doesn't contain any new feature(for skywalking), just a refactor with the new feature of JDK8+. (BTW, I have never said this PR contains any new feature. : P

I am not saying about the purpose of this PR. I am still talking about the `Blank` thing. I still think we don't need that. This util is also used in the backend, check the package, it doesn't belong to the agent.

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/PropertyPlaceholderHelper.java
```diff
@@ -83,17 +83,13 @@
      * Replaces all placeholders of format {@code ${name}} with the corresponding property from the supplied {@link
      * Properties}.
      *
-     * @param value the value containing the placeholders to be replaced
+     * @param value      the value containing the placeholders to be replaced
```
wu-sheng: Please revert this.

apm-sniffer/apm-agent-core/pom.xml
```diff
@@ -118,6 +119,10 @@
             <version>${ststem-rules.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+        </dependency>
```
wu-sheng: Don't add a dependency, we need to relocate all class package names in the agent. Unless it is necessary, we want the agent dependency lib as less as possible.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
```diff
@@ -18,6 +18,7 @@
 
 package org.apache.skywalking.apm.agent.core.conf;
 
+import lombok.Data;
```
wu-sheng: Don't use this.

wu-sheng: It is working before, please just refactor codes, rather than using new lib.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
```diff
@@ -72,9 +71,6 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
 
             pluginFinder = new PluginFinder(new PluginBootstrap().loadPlugins());
 
-        } catch (ConfigNotFoundException ce) {
```
wu-sheng: We need to catch this. In the `no config` scenario, the app should startup successfully.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
JohnNiang: I'm confused now. @kezhenxu94 mentioned that we should use `lombok` to avoid boilerplate codes(getters setters).

> @kezhenxu94 : [Use Lombok annotations to get rid of boilerplate codes (getters setters)](https://github.com/apache/skywalking/pull/4264#discussion_r367986163).

wu-sheng: @kezhenxu94 I think this should not use `Lombok`. Lombok is a good option for backend, but we don't need that at the agent side. 

Even we need to discuss whether we could accept this new lib, we should not ask in this pull request. @kezhenxu94 if you want this, please submit another pull request about this.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
JohnNiang: > We need to catch this. In the `no config` scenario, the app should startup successfully.

It will be caught by `[HERE](https://github.com/apache/skywalking/blob/4df7129ac15308582c4a77185f2df19cf619c074/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java#L68)`

```diff
    public static void initialize(String agentOptions) {

        try (InputStreamReader configFileStream = loadConfig()) {
            Properties properties = new Properties();
            properties.load(configFileStream);
            properties.forEach((key, value) -> {
                if (value instanceof String) {
                    properties.replace(key, PropertyPlaceholderHelper.INSTANCE.replacePlaceholders((String)value, properties));
                }
            });
            ConfigInitializer.initialize(properties, Config.class);
+        } catch (Exception e) {
            logger.error(e, "Failed to read the config file, skywalking is going to run in default config.");
        }
```

wu-sheng: Got it, make sense to me. Ignore this comment.

apm-sniffer/apm-agent-core/pom.xml
```diff
@@ -126,6 +131,12 @@
                 <artifactId>guava</artifactId>
                 <version>${guava.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.projectlombok</groupId>
+                <artifactId>lombok</artifactId>
+                <version>${lombok.version}</version>
+                <scope>provided</scope>
```
JohnNiang: @wu-sheng The scope of lombok dependency is provided. It won't be packaged into agent library. So please don't worry about the dependency will affect agent core.

wu-sheng: My points are two
1. Version should be managed at root pom, as Lombok is used at backend and agent.
1. With this feature added, a lot of codes could be removed at the agent side. It should be placed in another PR, with more codes changed. This will make the changelog clear when I do release several months later.

JohnNiang: Copy that. I will revert all changes related `lombok`.

wu-sheng: Submit an issue If Any of you want to add this later.

JohnNiang: Clear. 

BTY, I have reverted all changes related `lombok`, waiting for checking.

apm-commons/pom.xml
```diff
@@ -37,6 +37,6 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <compiler.version>1.6</compiler.version>
+        <compiler.version>1.8</compiler.version>
```
wu-sheng: This could be deleted. Right now, the whole project is compiled in 1.8 now.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/boot/AgentPackagePath.java
```diff
@@ -78,6 +78,7 @@ private static File findPath() throws AgentPackageNotFoundException {
             }
         }
 
+        // TODO No need to log this error message, the exception will be caught finally
```
wu-sheng: You could delete this log too. Don't leave TODO unless it is necessary.

kezhenxu94:
> Two nits, others LGTM. @kezhenxu94 @arugal Please help with rechecking.
> 
> @kezhenxu94 @JohnNiang Could you submit an issue for Lombok?

see #4272

JohnNiang:
May be I ought to start again util the rules come into effect.

Thank you for your reviews. @wu-sheng @kezhenxu94 .

wu-sheng:
@JohnNiang Why do you close? I am a little confused.

JohnNiang:
> @JohnNiang Why do you close? I am a little confused.

Every time I write some codes, and then I format code style with keyboard shortcut `Ctrl + Shift + F` carefully. 

I am afraid that something which shouldn't be changed has been changed.

At the same time, I imported `codeStyle.xml` into my IDEA code style config and disabled JavaDoc format, that felt very uncomfortable.

So I'm looking forward you put some stricter rules for formatting code style as mentioned in https://github.com/apache/skywalking/issues/4272.

I'll re-PR as soon as possible. I'm very sorry that this PR is taking up too much of your time!

wu-sheng:
There is already a `codeStyle.xml` in root path, once, you import it, you are good. @kezhenxu94 's point is making it better. 

The format I pointed out, is not about that part.

PR #4289, opened 2020-01-29
Refactor the storage to support `getSubsetOfMultipleLinearIntValues`

### LINKED ISSUE #4288: Support specific linear query for new `getMultipleLinearIntValues`
In the current query protocol, you have the only option to provide the total line number. As a result, such as for percentile, you have to query P50 if you want to other metrics, such as P50->P99 or P50-P90. 
```
    # Query the type of metrics including multiple values, and format them as multiple linears.
    # The seq of these multiple lines base on the calculation func in OAL
    # Such as, should us this to query the result of func percentile(50,75,90,95,99) in OAL,
    # then five lines will be responsed, p50 is the first element of return value.
    getMultipleLinearIntValues(metric: MetricCondition!, numOfLinear: Int!, duration: Duration!): [IntValues!]!
```

This new query could leverage the existing DAO feature, but filter the query result including specific linear(s) only. We could add `linearIndex` as a parameter.

```
    getSubsetOfMultipleLinearIntValues(metric: MetricCondition!, linearIndex: [Int!]!, duration: Duration!): [IntValues!]!
```

I will try to work on this. If any of @apache/skywalking-committers wants to lead, please let me know.

wu-sheng:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

-----
- Related issues
Resolves #4288

Support to query specific linear(s) from the storage directly. The existing `getMultipleLinearIntValues` has been refactored to adopt this new change.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/MetricQueryService.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/MetricQueryService.java
@@ -100,6 +100,19 @@ public class MetricQueryService implements Service {
         final Downsampling downsampling,
         final long startTB,
         final long endTB) throws IOException {
+        List<Integer> linearIndex = new ArrayList<>(numOfLinear);
+        for (int i = 0; i < numOfLinear; i++) {
+            linearIndex.add(i);
+        }
+
+        return getSubsetOfMultipleLinearIntValues(indName, id, linearIndex, downsampling, startTB, endTB);
+    }
+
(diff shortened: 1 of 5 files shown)


wu-sheng:
FYI @dmsolr This related to your storage PR. We could make it ready ASAP, it has been broken by other changes several times.

dmsolr:
> FYI @dmsolr This related to your storage PR. We could make it ready ASAP, it has been broken by other changes several times.

I got it. I will finish it ASAP.

PR #4303, opened 2020-01-31
provide profile analyze empty interfaces

mrproliu:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#4104 
https://github.com/apache/skywalking-query-protocol/pull/29
___
- provide profile GraphQL interfaces, make compiling and boot successfully.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/entity/ProfileStackElement.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/entity/ProfileStackElement.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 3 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/entity/ProfileStackElement.java
```diff
@@ -0,0 +1,48 @@
...
+@Setter
+public class ProfileStackElement {
+
+    // stack code signature
+    private String codeSignature;
+
+    // self include children duration(millisecond)
+    private int duration;
+
+    // self exclude children duration(millisecond)
```
wu-sheng: Wrong comment.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/entity/ProfileStackElement.java
```diff
@@ -0,0 +1,48 @@
...
+ * @author MrPro
+ */
+@Getter
+@Setter
+public class ProfileStackElement {
+
+    // stack code signature
+    private String codeSignature;
+
+    // self include children duration(millisecond)
```
wu-sheng: Wrong comment.

PR #4305, opened 2020-02-01
Performance tuning: replace AtomicInteger with AtomicIntegerFieldUpdater

kezhenxu94:
### Motivation:

Reduce unnecessary memory allocation in frequently used class.

### Modifications:

Replace the `AtomicInteger` with `AtomicIntegerFieldUpdater`

### Result:

We can save 4 bytes (for the reference to `AtomicInteger` itself) for each `TracingContext`, which saves credible memory in a high-concurrent agent service.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/TracingContext.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/TracingContext.java
@@ -20,7 +20,7 @@ package org.apache.skywalking.apm.agent.core.context;
 
 import java.util.LinkedList;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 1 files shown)


kezhenxu94:
If you're interested in the theory and experiment, please refer to http://normanmaurer.me/blog/2013/10/28/Lesser-known-concurrent-classes-Part-1/ and the counterpart Chinese version is here: https://kezhenxu94.github.io/2020/01/30/Concurrent-classes-AtomicFieldUpdater/

wu-sheng:
@kezhenxu94 I think you miss an important factor, in the `AtomicIntegerFieldUpdaterImpl`, it is using the refactor, which causes more CPU. This is a memory-CPU tradeoff, I am not sure this is the performance tuning. In my mind, in today's production env, CPU is more valuable than the memory.

kezhenxu94:
> @kezhenxu94 I think you miss an important factor, in the `AtomicIntegerFieldUpdaterImpl`, it is using the refactor, which causes more CPU. This is a memory-CPU tradeoff, I am not sure this is the performance tuning. In my mind, in today's production env, CPU is more valuable than the memory.
> 
> Especially, CPU is affecting latency, response time and throughput more.

@wu-sheng as you can see, the updater is `static` and is only instantiated once, and reflection is only use in the instantiation to find the field offset, so the CPU cost is an one-time cost

PR #4307, opened 2020-02-02
Clean up legacy v1 header logic, use built-in Base64 since JDK8

kezhenxu94:
### Motivation:

Codes clean up

### Modifications:

- Remove v1 header, follow up #4244

- Use built-in Base64 class (since JDK8)

### Result:

- No more legacy v1 headers

- No unnecessary codes concerns by using built-in ability

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextCarrier.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextCarrier.java
@@ -24,7 +24,6 @@ import org.apache.skywalking.apm.agent.core.base64.Base64;
 import org.apache.skywalking.apm.agent.core.context.ids.DistributedTraceId;
 import org.apache.skywalking.apm.agent.core.context.ids.ID;
 import org.apache.skywalking.apm.agent.core.context.ids.PropagatedTraceId;
-import org.apache.skywalking.apm.agent.core.context.trace.TraceSegment;
 import org.apache.skywalking.apm.agent.core.dictionary.DictionaryUtil;
 import org.apache.skywalking.apm.util.StringUtil;
 
@@ -35,9 +34,6 @@ import org.apache.skywalking.apm.util.StringUtil;
  * Created by wusheng on 2017/2/17.
  */
 public class ContextCarrier implements Serializable {
-    /**
-     * {@link TraceSegment#traceSegmentId}
-     */
     private ID traceSegmentId;
 
     /**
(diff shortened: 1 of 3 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextCarrier.java
```diff
@@ -108,53 +103,36 @@ String serialize(HeaderVersion version) {
...
+                    this.primaryDistributedTraceId = new PropagatedTraceId(Base64.decode2UTFString(parts[1]));
+                    this.traceSegmentId = new ID(Base64.decode2UTFString(parts[2]));
+                    this.spanId = Integer.parseInt(parts[3]);
+                    this.parentServiceInstanceId = Integer.parseInt(parts[4]);
+                    this.entryServiceInstanceId = Integer.parseInt(parts[5]);
+                    this.peerHost = Base64.decode2UTFString(parts[6]);
+                    this.entryEndpointName = Base64.decode2UTFString(parts[7]);
+                    this.parentEndpointName = Base64.decode2UTFString(parts[8]);
+                } catch (NumberFormatException ignored) {
+
```
kezhenxu94: Quick return to reduce nested levels

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextCarrier.java
```diff
@@ -108,53 +103,36 @@ String serialize(HeaderVersion version) {
...
                 }
-            } else {
-                throw new IllegalArgumentException("Unimplemented header version." + version);
             }
         }
-        return this;
+        throw new IllegalArgumentException("Unimplemented header version." + version);
```
wu-sheng: Don't throw an exception in agent core API. Output a warning log.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextCarrier.java
```diff
@@ -163,27 +141,16 @@ public boolean isValid() {
...
                 && traceSegmentId.isValid()
                 && getSpanId() > -1
                 && parentServiceInstanceId != DictionaryUtil.nullValue()
                 && entryServiceInstanceId != DictionaryUtil.nullValue()
                 && !StringUtil.isEmpty(peerHost)
                 && primaryDistributedTraceId != null;
-        } else {
-            throw new IllegalArgumentException("Unimplemented header version." + version);
         }
+        throw new IllegalArgumentException("Unimplemented header version." + version);
```
wu-sheng: Same here.

wu-sheng:
Many CI processes fail. Please recheck.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextCarrier.java
kezhenxu94: Just double check the original codes, the exception statement should never be reached, I simply remove them

PR #4308, opened 2020-02-02
provide thread stack analyzes

mrproliu:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#4104 

- provide thread analyzes to the stack tree.


docs/en/setup/backend/backend-profile.md
```diff
@@ -0,0 +1,22 @@
+# Profile
```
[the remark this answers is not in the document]
mrproliu: has linked to code.

mrproliu:
> And we plan to remove all the `@author` tags since it's an ASF project, better to remove those in newly-added files

Sure, removed.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
```diff
@@ -0,0 +1,242 @@
...
+     * combine from other {@link ProfileStackNode}
+     * @param node
+     * @return
+     */
+    public ProfileStackNode combine(ProfileStackNode node) {
+        // combine this node
+        this.combineThisNode(node);
+
+        // merge tree using pre order
+        LinkedList<Pair<ProfileStackNode, List<ProfileStackNode>>> stack = new LinkedList<>();
```
wu-sheng: - Q1. Which algorithm are you using for tree merging?
- Q2. Are you using the stack to avoid `recursion`? The codes seem understandable, but not very clear.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
```diff
@@ -0,0 +1,242 @@
...
+                ProfileStackNode combinedNode = combineTo.children.stream().filter(n -> needCombineChildren.matches(n)).findFirst().orElse(null);
+
+                if (combinedNode == null) {
+                    combineTo.children.add(needCombineChildren);
+                    combinedNode = needCombineChildren;
+                } else if (needCombineChildren != combinedNode) {
+                    combinedNode.combineThisNode(needCombineChildren);
+                } else {
+                    // if same reference, dont need to combine anymore
+                    sameReference = true;
```
wu-sheng: Why same reference happens?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
```diff
@@ -0,0 +1,242 @@
...
+     * @param node
+     * @return
+     */
+    public ProfileStackNode combine(ProfileStackNode node) {
+        // combine this node
+        this.combineThisNode(node);
+
+        // merge tree using pre order
+        LinkedList<Pair<ProfileStackNode, List<ProfileStackNode>>> stack = new LinkedList<>();
+        stack.add(new Pair<>(this, node.children));
```
wu-sheng: Why don't you use the same depth of key and value? In the following codes, you always use `combineTo.children`.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
mrproliu: It's a bug, has fixed. If combine node from B -> A, A node not found, just need to add children node, don't need to keep merge child.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
```diff
@@ -0,0 +1,242 @@
...
+    private int duration;
+
+    /**
+     * create new empty, un-init node
+     * @return
+     */
+    public static ProfileStackNode createEmptyNode() {
+        ProfileStackNode emptyNode = new ProfileStackNode();
+        emptyNode.owners = new LinkedList<>();
+        emptyNode.children = new LinkedList<>();
```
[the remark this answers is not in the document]
mrproliu: change `children` field to ArrayList.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
mrproliu: Q1: Using LDR to traversal tree node, and merge children nodes to the parent node.
Q2: Yes. If the stack has large tree stack depth, It will throw StackOverflowError.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
```diff
@@ -0,0 +1,235 @@
...
+    /**
+     * accumulate {@link ProfileStack} to this tree, it will invoke on the tree root node
+     * @param stack
+     */
+    public void accumulateFrom(ProfileStack stack) {
+        List<String> stackList = stack.getStack();
+        if (codeSignature == null) {
+            codeSignature = stackList.get(0);
+        }
+        // increase self
```
wu-sheng: Fix comment too.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
```diff
@@ -0,0 +1,244 @@
...
+        return this;
+    }
+
+    /**
+     * merge all children nodes to appoint node
+     * @param mergeTo
+     * @param combineNode
+     * @param needKeepEach if keep traversal, need report
+     */
+    private void combineChildrenNodes(ProfileStackNode mergeTo, ProfileStackNode combineNode, Consumer<Pair<ProfileStackNode, ProfileStackNode>> needKeepEach) {
```
wu-sheng: `needKeepEach` -> `continueChildrenMerging`

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
```diff
@@ -0,0 +1,244 @@
...
+        for (ProfileStackNode childrenNode : combineNode.children) {
+            // find node from mergeTo children
+            ProfileStackNode combinedNode = mergeTo.children.stream().filter(n -> childrenNode.matches(n)).findFirst().orElse(null);
+
+            if (combinedNode == null) {
+                mergeTo.children.add(childrenNode);
+            } else {
+                combinedNode.combineDetectedStacks(childrenNode);
+                needKeepEach.accept(new Pair<>(combinedNode, childrenNode));
+            }
```
wu-sheng: You are iterating `combineNode` and process found or not, but
1. Once it is found, it has not been removed from `mergeTo`, then all elements all been process multiple times.
1. You seem missing node left in `mergeTo#children`, if some of them are not included in `combineNode`

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
```diff
@@ -0,0 +1,268 @@
...
+        // all nodes add to single-level list (such as flat), work for parallel calculating
+        LinkedList<Pair<ProfileStackElement, ProfileStackNode>> nodeMapping = new LinkedList<>();
+        ProfileStackElement root = buildElement();
+        nodeMapping.add(new Pair<>(root, this));
+
+        // same with combine logic
+        LinkedList<Pair<ProfileStackElement, ProfileStackNode>> stack = new LinkedList<>();
+        stack.add(new Pair<>(root, this));
+        while (!stack.isEmpty()) {
+            Pair<ProfileStackElement, ProfileStackNode> needCombineNode = stack.pop();
```
wu-sheng: `needCombineNode` -> `MergingPair`

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
```diff
@@ -0,0 +1,268 @@
...
+            ProfileStackElement combineTo = needCombineNode.key;
```
wu-sheng: `combineTo` -> `respElement`

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
```diff
@@ -0,0 +1,268 @@
...
+                Pair<ProfileStackElement, ProfileStackNode> pair = new Pair<>(element, c);
+                stack.add(pair);
+                nodeMapping.add(pair);
+
+                return element;
+            }).collect(Collectors.toList()));
+        }
+
+        // calculate durations
+        nodeMapping.parallelStream().forEach(t -> t.value.calculateDuration(t.key));
```
wu-sheng: Why require `ProfileStackNode`(t.value) to calculate duration? Using List<ProfileStackElement> should be enough

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
```diff
@@ -0,0 +1,268 @@
...
+        nodeMapping.parallelStream().forEach(t -> t.value.calculateDurationExcludeChild(t.key));
```
wu-sheng: Why run `parallelStream` twice? Calculating two kinds of duration should be faster.

oap-server/server-core/src/test/resources/thread-snapshot.yml
```diff
@@ -0,0 +1,175 @@
...
+
+# verify: read to analyze stack info
+#   limit: per stack snapshot dump limit
+#   stack: per data means one snapshot, stack elements split by "-"
+# expected: need verify data analyze result
+#   info: follow this pattern: codeSignature(duration:durationExcludeChild)
+#   children: all children nodes
+
+list:
+  # case 1
```
wu-sheng: This format is much better than the prev one. Let's add one more case
- Case 5
```
A-B-C
A-B
D-E
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
```diff
@@ -0,0 +1,256 @@
...
+    /**
+     * calculate duration to {@link ProfileStackElement#getDuration()}
+     */
+    private void calculateDuration(ProfileStackElement element) {
+        if (this.detectedStacks.size() <= 1) {
+            element.setDuration(0);
+            return;
+        }
+
+        // sort owners
```
kezhenxu94: `owners` -> `detectedStacks`, but I think the code itself is documentation already

docs/en/guides/backend-profile.md
```diff
@@ -0,0 +1,44 @@
...
+### Read data and transform
+Read data from the database and convert it to a data structure in gRPC.
+```
+st=>start: Start
+e=>end: End
+op1=>operation: Load data using paging
+op2=>operation: Transform data using parallel
+
+st(right)->op1(right)->op2
+op2(right)->e
```
kezhenxu94: I'd prefer to generate an image and put that image here, the source codes seems not that straightforward, same as others below

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
mrproliu: Ok, I understand. I will delete this comment

docs/en/guides/backend-profile.md
wu-sheng: Don't put the image in the source codes. It will increase the payload of source download and release. All documents keep in texts as much as possible. All images must be hosted at the website repo, and only links here. For these comments, we could make it better, but I still think, visualizing on 3rd party system is better.

wu-sheng: For this graph, @mrproliu , I think we should describe the logic of each step doing, rather than stream API concept.

wu-sheng:
FYI @kezhenxu94 The whole profile should work for CLI too. We should add an issue to track.

Issue #4284, opened 2020-01-29
Add lombok dependency into  root pom globally


JohnNiang:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [x] Feature or performance improvement

___

### Requirement or improvement

Follow the discuss: <https://github.com/apache/skywalking/pull/4264#discussion_r368279843>

JohnNiang:
WIP.

PR #4309, opened 2020-02-02
 Move lombok dependency into root dependency manager for easier version management

JohnNiang:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

https://github.com/apache/skywalking/issues/4284

___

Move lombok dependency into root dependency manager for easier version management.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -186,7 +186,7 @@
         <checkstyle.version>6.18</checkstyle.version>
         <junit.version>4.12</junit.version>
         <mockito-all.version>1.10.19</mockito-all.version>
-        <lombok.version>1.18.4</lombok.version>
+        <lombok.version>1.18.10</lombok.version>
 
         <!-- core lib dependency -->
         <grpc.version>1.26.0</grpc.version>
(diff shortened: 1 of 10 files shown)


JohnNiang:
Maybe I have to downgrade lombok version.

kezhenxu94:
> Maybe I have to downgrade lombok version.

@JohnNiang no, using the latest Lombok version should be ok

pom.xml
```diff
@@ -275,6 +276,12 @@
                 <version>${powermock.version}</version>
                 <scope>test</scope>
             </dependency>
+            <dependency>
```
JohnNiang: @kezhenxu94 Should I move the `lombok` into root `dependencies` other than `dependencyManagement`.

wu-sheng: That depends on how you use it. Logically, they are the same. But `provided` is pointless.

wu-sheng: And from the current `dependencies` in the root `pom.xml`, no lib is declared there.

wu-sheng:
You should test the compiling locally first.

JohnNiang:
> You should test the compiling locally first.

The checks of test folder failed due to redundant dependency.

pom.xml
kezhenxu94: It should be better, and remove all Lombok in all sub modules

PR #4318, opened 2020-02-04
Reduce footprint via prototype pattern when tagging spans

kezhenxu94:
### Motivation:

Reduce footprint when tagging spans with the deprecated API: `AbstractSpan#tag(String, String)`.

### Modifications:

- Adopt [prototype pattern](https://en.wikipedia.org/wiki/Prototype_pattern) to create `Tag`, prevent from creating too many `StringTag` instances and `GC`ed, because `AbstractSpan#tag(String, String)` simply `new` a `StringTag` every time and calls `AbstractSpan#tag(AbstractTag, String)`.

- Replace `AbstractSpan#tag(String, String)` with `AbstractSpan#tag(AbstractTag, String)`.

### Result:

- Footprint is reduced.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/tag/Tags.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/tag/Tags.java
@@ -19,6 +19,9 @@
 
 package org.apache.skywalking.apm.agent.core.context.tag;
 
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
 /**
  * The span tags are supported by sky-walking engine. As default, all tags will be stored, but these ones have
  * particular meanings.
@@ -26,6 +29,8 @@ package org.apache.skywalking.apm.agent.core.context.tag;
  * Created by wusheng on 2017/2/17.
  */
 public final class Tags {
+    private static final Map<String, StringTag> TAG_PROTOTYPES = new ConcurrentHashMap<>();
+
     private Tags() {
     }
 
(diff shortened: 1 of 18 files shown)


kezhenxu94:
> Removing `AbstractSpan#tag(String, String)` is an incompatible change. I don't know, but assuming many, it is widely used out of our open source repo. Such as https://github.com/SkyAPM/java-plugin-extensions or private plugins?

Well, if it’s marked as deprecated, I assume it is to be removed some day, and the major version 7.x is a good time IMO, yet I can rewrite the deprecated API to use the new prototype tag creation, for compatibility, which one do you prefer :)

wu-sheng:
I know as a major upgrade, we could remove some, but basically, it is not necessary. So I want to keep them. 

> yet I can rewrite the deprecated API to use the new prototype tag creation, for compatibility

Yes, I want to keep that way, as it seems like it doesn't harm performance much.

wu-sheng:
BTW, we should add a new check rule, `don't use Deprecated API`.

JohnNiang:
> BTW, we should add a new check rule, `don't use Deprecated API`.

I don't think this is a good decision. `Deprecated` annotation warns future users not to use that method again but new api instead.

wu-sheng:
@JohnNiang This is not a decision, suggest taking the discussion to the mail list.

kezhenxu94:
> > BTW, we should add a new check rule, `don't use Deprecated API`.
> 
> I don't think this is a good decision. `Deprecated` annotation warns future users not to use that method again but new api instead.

Adding a new check rule is to avoid "future users use that method again", so what's your suggestion?

JohnNiang:
> Adding a new check rule is to avoid "future users use that method again", so what's your suggestion?

We can't avoid using deprecated method, but it's our duty to tell future users that method is deprecated.

Just like recent rectifications of invoking rewrited method in deprecated method body, but keep `Deprecated` annotation on the deprecated method.

e.g.

```java
@Deprecated // Keep this annotation
public T deprecatedMethod() {
    return newMethod();
}

public T rewritedMethod() {
    // Latest codes here...
}
```

In a short, Using `Deprecated` api is my suggestion.

PR #4320, opened 2020-02-05
[BUG]Fix the bug of port identification failure when connecting Oracle with Sid mode (JDBC: Oracle: thin: @ < host >: < port > / < Sid >)

AirTrioa:

- [ ] Bug fix

- Related issues

___
### Bug fix
- Bug description.
When some older projects (using Oracle database) access skywalking, we find that the port and database name of Oracle are incorrectly identified. The port is recognized as 1521 by default and the database name is recognized as < port > / < dbname >.
We found that when we cut the Oracle URL, we only recognize the URL of one sid connection mode (jdbc: oracle: thin: @ < host >: < port >: < Sid >); so here I added the recognition of the URL of another sidsid connection mode (jdbc: oracle: thin: @ < host >: < port > / < Sid >).

- How to fix?
An if branch is added before setting the default port.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/connectionurl/parser/OracleURLParser.java
+++ b/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/connectionurl/parser/OracleURLParser.java
@@ -94,6 +94,10 @@ public class OracleURLParser extends AbstractURLParser {
         String[] hostSegment = splitDatabaseAddress(host);
         String databaseName = fetchDatabaseNameFromURL();
         if (hostSegment.length == 1) {
+            if (databaseName.contains("/") && databaseName.split("/").length == 2) {
+                String[] portAndDatabaseName = databaseName.split("/");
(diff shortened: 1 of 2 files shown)


apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/connectionurl/parser/OracleURLParser.java
```diff
@@ -94,6 +94,10 @@ private ConnectionInfo commonsURLParse() {
         String[] hostSegment = splitDatabaseAddress(host);
         String databaseName = fetchDatabaseNameFromURL();
         if (hostSegment.length == 1) {
+            if (databaseName.contains("/") && databaseName.split("/").length == 2) {
+                String[] portAndDatabaseName = databaseName.split("/");
+                return new ConnectionInfo(ComponentsDefine.OJDBC, DB_TYPE, host, Integer.valueOf(portAndDatabaseName[0]), portAndDatabaseName[1]);
+            }
```
wu-sheng: Where is the `else`?

ascrutae:
LGTM

PR #4321, opened 2020-02-05
Refactor ConfigInitializer and improve map type config

JohnNiang:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance
___

1. Refactor ConfigInitializer  and enhance setting which field type is Map.
2. Add some test cases into `SnifferConfigInitializerTest` for Map field.


wu-sheng:
Could you describe which part you enhanced in this PR? Add some examples and tests to demonstrate it.

JohnNiang:
> Could you describe which part you enhanced in this PR? Add some examples and tests to demonstrate it.

If `Config.Agent` contains field as follow:

```java
public class Config {

    public static class Agent {
        // The int, long, float, double, list and enum type were not interpreted before
        public static Map<String, Integer> map;
    }
}
```

And if `agent.conf` like this:

```conf
skywalking.agent.map[test1] = 1
skywalking.agent.map[test2] = 2
```

This PR refactor some codes  lets the above config work well.

PR #4332, opened 2020-02-09
Change profile stack element data structure

mrproliu:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
#4104 
___
### Bug fix
- GraphQL cannot support self reference, cannot build query request.

- How to fix?
Change profile stack element using single level list, use `id` and `parentId` to reference parent element.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
@@ -181,7 +181,11 @@ public class ProfileStackNode {
         nodeMapping.parallelStream().forEach(t -> t.value.calculateDurationExcludeChild(t.key));
 
         ProfileStackTree tree = new ProfileStackTree();
-        tree.setElements(nodeMapping.stream().map(p -> p.key).collect(Collectors.toList()));
+        ArrayList<ProfileStackElement> elements = new ArrayList<>(nodeMapping.size());
+        for (Pair<ProfileStackElement, ProfileStackNode> pair : nodeMapping) {
+            elements.add(pair.key);
+        }
+        tree.setElements(elements);
 
         return tree;
     }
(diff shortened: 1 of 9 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
```diff
@@ -163,21 +166,24 @@ public ProfileStackElement buildAnalyzeResult() {
...
+            }
         }
 
         // calculate durations
         nodeMapping.parallelStream().forEach(t -> t.value.calculateDuration(t.key));
         nodeMapping.parallelStream().forEach(t -> t.value.calculateDurationExcludeChild(t.key));
 
-        return root;
+        ProfileStackTree tree = new ProfileStackTree();
+        tree.setElements(nodeMapping.stream().map(p -> p.key).collect(Collectors.toList()));
```
wu-sheng: This could be replaced by simplier `foreach`.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileStackNode.java
```diff
@@ -163,21 +165,28 @@ public ProfileStackElement buildAnalyzeResult() {
...
         nodeMapping.parallelStream().forEach(t -> t.value.calculateDuration(t.key));
         nodeMapping.parallelStream().forEach(t -> t.value.calculateDurationExcludeChild(t.key));
 
-        return root;
+        ProfileStackTree tree = new ProfileStackTree();
+        ArrayList<ProfileStackElement> elements = new ArrayList<>(nodeMapping.size());
+        for (Pair<ProfileStackElement, ProfileStackNode> pair : nodeMapping) {
+            elements.add(pair.key);
+        }
+        tree.setElements(elements);
```
wu-sheng: Use codes like `nodeMapping.forEach(node-> tree.getElements().add(node.value));`

PR #4353, opened 2020-02-12
Optimizing performance reduces es index queries scope by timebucket

aderm:
Optimizing ES query performance reduces the scope of index queries through precise queries

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues
#4194
___
### Bug fix
- Bug description.

- How to fix?

___


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/DurationUtils.java
```diff
@@ -213,4 +213,36 @@ private DateTime parseToDateTime(Downsampling downsampling, long time) {
         }
         throw new UnexpectedException("Unexpected downsampling: " + downsampling.name());
     }
+
+    public DateTime startTimeBucket2DateTime(Downsampling downsampling, long startTB) {
```
wu-sheng: First, time covert should be inside `TimeBucket`, take a look at InfluxDB, https://github.com/apache/skywalking/pull/4239/files#diff-df3b7b3f55b25fb0e29e345e7e6cc078. Maybe good for you.
I think you don't need the `DateTime`? Directly convert to long as day/month makes more sense?

wu-sheng: Currently, you covert to DateTime, then cover to long like `202002`, but actually, you could use divide to get `202002` directly and more quickly, right?

aderm: Indeed, it maybe faster to convert to long. When the date while loop statement is added, it needs to be converted to the character format of the date.

wu-sheng: Why do you need to add? The current prefix should be enough. `2020021015`-`2020021315` to days is 20200210`-`20200213`

Do I miss anything?

aderm: When crossing months, you need to consider the big and small months. In addition, you need to consider some leap years. like `20200227-20200305`

wu-sheng: OK, got it, make sense.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ModelName.java
```diff
@@ -30,10 +43,73 @@ public static String build(Downsampling downsampling, String modelName) {
...
             case Hour:
                 return modelName + Const.ID_SPLIT + Downsampling.Hour.getName();
-            //            case Second:
-            //                return modelName + Const.ID_SPLIT + Downsampling.Second.getName();
             default:
                 return modelName;
         }
     }
+
+    public static String[] build(Downsampling downsampling, String modelName, long startTB, long endTB) {
```
wu-sheng: This build method is for ES storage only, please move inside the plugin.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ModelName.java
```diff
@@ -30,10 +43,73 @@ public static String build(Downsampling downsampling, String modelName) {
...
+        List<String> indexNameList = new ArrayList<>();
+
+        List<String> whiteIndexList = new ArrayList<String>() {
```
wu-sheng: This should be a static list, rather than create everytime.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -291,6 +292,30 @@ public SearchResponse search(String indexName, SearchSourceBuilder searchSourceB
         return client.search(searchRequest);
     }
 
+    public SearchResponse search(String[] indexNames, SearchSourceBuilder searchSourceBuilder) throws IOException {
```
wu-sheng: Add comments for these methods.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -291,6 +292,30 @@ public SearchResponse search(String indexName, SearchSourceBuilder searchSourceB
...
+    }
+
+    public String[] filterNotExistIndex(String[] fullIndexNames, String indName) throws IOException {
+        // if no wrap, it is impossible to remove elements
+        List<String> indexNameList = new ArrayList<>(Arrays.asList(fullIndexNames));
+        if (fullIndexNames.length > 0) {
+            List<String> existIndex = retrievalIndexByAliases(indName);
+            indexNameList.removeIf(indexName -> {
+                //only filter index name with xxxx-xxxx
+                if (indexName.contains("-")) {
```
wu-sheng: Which one doesn't contain `-`?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ModelName.java
```diff
@@ -30,10 +43,73 @@ public static String build(Downsampling downsampling, String modelName) {
...
+                        String indexName = build(downsampling, modelName) + "-" + YYYYMMDD.print(startDT);
+                        indexNameList.add(indexName);
+                        startDT = startDT.plusDays(1);
+                    }
+                    break;
+                default:
+                    indexNameList.add(modelName);
+            }
+        }
+        return indexNameList.toArray(new String[0]);
```
wu-sheng: You covert to array at the first place, but then you covert to List again in `#filterNotExistIndex`. Seems unnecessary?

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
wu-sheng: Also relate to `String[] indexNames`, should we keep all in List and covert to array at last minute?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ModelName.java
aderm: If select a date with a long span, which dates are not in the middle, or the system has been stopped for a few days, there may be no index data for several days. If you query date  contain those days directly, searching the index will appear `index not found exception`. so, Did a filter that the date index does not exist

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
aderm: for example `segment alarm_record`

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ModelName.java
aderm: ok

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ModelName.java
aderm: ok

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
aderm: good idea

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
wu-sheng: Use `Const#LINE`

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ModelName.java
wu-sheng: Did you answer the question in the wrong context? I am not saying `filterNotExistIndex` is useless.

But based on the things you talked, I think you should establish a cache to `filterNotExistIndex` and cache the result for at least a day, considering the index could only be deleted day by day, right? Set up a guava cache with `#expireAfterWrite`?

aderm: > You covert to array at the first place, but then you covert to List again in `#filterNotExistIndex`. Seems unnecessary?
Oh, indeed i will fix it. Convert at last minitue.

> But based on the things you talked, I think you should establish a cache to `filterNotExistIndex` and cache the result for at least a day, considering the index could only be deleted day by day, right? Set up a guava cache with `#expireAfterWrite`?
Good advice, i will try.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
aderm: ACK

aderm: server-core depends on server-client, so `Const#LINE` can not used in server-client module.

wu-sheng: Then you could provide a filter func as the parameter to do the filter works. This indicates this logic doesn't belong to this class.

aderm:
There is indeed a problem that needs to determine whether the index exists. The better way I thought at the time was if it was found that the query part of the index does not exist, if ES can query part of the index, and there is a configuration that can be set to not report an error.

Regarding the narrow gap time, this commit has optimization is that if it is not found in the cache, it will query ES again to ensure that the index name list is up to date. If you want to be more precise in fine-grained time, for example ms level, the index change after index query, it may be that the current query will report an error(currently this situation will only occur in the early morning). It will have correct result query again without affecting subsequent queries.

wu-sheng:
> It will have correct result query again without affecting subsequent queries.

This may be an issue in the production environment. The error log outputs will trigger the alarm system, which is monitoring SkyWalking. We need to avoid them permanently, I suppose.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/EsModelName.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+    };
+
+    public static List<String> build(Downsampling downsampling, String modelName, long startTB, long endTB) {
+        if (WHITE_INDEX_LIST.contains(modelName)) {
+            return new ArrayList<String>() {
+                {
+                    add(modelName);
+                }
+            };
```
kezhenxu94: Same here, please be careful when using this kind of initialization, it creates unnecessary inner class, if you always use this, there may be many many unnecessary inner classes, exploding the meta space of the JVM, and the problem is that we have much better method to  do this, like `Arrays.asList`, Guava list, etc.: 

```java
              new ArrayList<String>() {
                  {
                      add(modelName);
                  }
              };
```

aderm:
@wu-sheng Worried about the problem of error in log( `index_not_found` exception ), I checked the relevant documents. Found that ES has configuration parameters that can be set.
Refer this:[[https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-index.html](url)](url)

wu-sheng:
`allow_no_indices` and `ignore_unavailable` should be good choices. Could you try to use these? And check whether there is error log in OAP and ES server?

Here is the report from the original issue author, in Chinese, 
[skywalking查询优化基准测试.pdf](https://github.com/apache/skywalking/files/4203428/skywalking.pdf)

This only shows the necessity to support this in the segment query, also, it requires a change in the UI.
@kezhenxu94 @zhaoyuguang @JaredTan95 @wayilau @aderm @arugal Did any of you face this issue in the production environment? Should we continue on this PR? Or tag this as TBD later?

wu-sheng:
Before doing code level deep dive, please read my comments, we should set the target clear about this optimization.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/EsModelName.java
```diff
@@ -0,0 +1,107 @@
...
+                    }
+                    break;
+                case Second:
+                    //current second index is also suffix with YYYYMMDD
+                    while (endDT.isAfter(startDT)) {
+                        String indexName = build(downsampling, modelName) + "-" + YYYYMMDD.print(startDT);
+                        indexNameList.add(indexName);
+                        startDT = startDT.plusDays(1);
+                    }
+                    break;
```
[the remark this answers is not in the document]
aderm: ACK

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/EsModelName.java
aderm: ack

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/EsModelName.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final DateTimeFormatter YYYYMMDD = DateTimeFormat.forPattern("yyyyMMdd");
+
+    public static final List<String> WHITE_INDEX_LIST = new ArrayList<String>() {
+        {
+            add(EndpointInventory.INDEX_NAME);
+            add(NetworkAddressInventory.INDEX_NAME);
+            add(ServiceInventory.INDEX_NAME);
+            add(ServiceInstanceInventory.INDEX_NAME);
+        }
+    };
```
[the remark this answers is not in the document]
aderm: ack

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -403,4 +421,11 @@ public String formatIndexName(String indexName) {
         }
         return indexName;
     }
+
+    public List<String> formatIndexNames(List<String> indexNameList) {
+        if (StringUtil.isNotEmpty(namespace)) {
+            indexNameList.stream().map(indexName -> namespace + "_" + indexName);
```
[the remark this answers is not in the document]
aderm: oops, ack

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -291,6 +291,24 @@ public SearchResponse search(String indexName, SearchSourceBuilder searchSourceB
...
 
+    /**
+     * Search results from ES search engine according to various search conditions,
+     * Note the method is usered for the list of index names is optimized based on
+     * the scope of startTimeBucket and endTimeBucket
+     * @param indexNameList  full index names list base on timebucket scope.
+     * Except for endpoint_inventory network_address_inventory service_inventory service_instance_inventory
+     * @param searchSourceBuilder Various search query conditions
+     * @return ES search query results
+     * @throws IOException throw IOException
```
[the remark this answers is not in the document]
aderm: ack

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/DurationUtils.java
```diff
@@ -213,4 +213,36 @@ private DateTime parseToDateTime(Downsampling downsampling, long time) {
...
+            case Day:
+                return YYYYMMDD.parseDateTime(String.valueOf(endTB));
+            case Hour:
+                return YYYYMMDDHH.parseDateTime(String.valueOf(endTB));
+            case Minute:
+                return YYYYMMDDHHMM.parseDateTime(String.valueOf(endTB));
+            case Second:
+                return YYYYMMDDHHMMSS.parseDateTime(String.valueOf(endTB));
+        }
+        throw new UnexpectedException("Unexpected downsampling: " + downsampling.name());
```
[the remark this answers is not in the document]
aderm: ack, for the day and month, different circulation methods are used. Considering that the monthly increase calculation should be time-consuming, it is still the original method. WDYT?

aderm:
> `allow_no_indices` and `ignore_unavailable` should be good choices. Could you try to use these? And check whether there is error log in OAP and ES server?

Try to use without problems, in addition, the machine tested some queries that do not exist across the date index, the results are normal，no error in log.

wu-sheng:
Hi @aderm , I want to hold this PR a little longer. The index calculation will face a little more challenge.

In PR #4364, I provide a new feature to reduce the index number, and I am thinking to provide TTL step to reduce the index number again, especially for metrics indexes.
@apache/skywalking-committers WHDT? To @hanahmily and @kezhenxu94, above is one being considered optimization of index number.

wu-sheng:
As an opposite direction of my previous two PRs about reducing the index number, this runs into another way. We have two choices AFAIK
1. Close this for now, or leave it for now.
1. Make this chance only affecting the trace query. Because it seems the only optimization we need. And because of that, it will only related to Segment Query DAOs. But still remain the question, we don't require time conditions in traceId query.

Personally, I prefer to don't chance, even we have put much energy on this. @apache/skywalking-committers please feedback what do you think?

aderm:
@wu-sheng agree with your suggestion

wu-sheng:
Which one? There are two above

aderm:
Option one

wu-sheng:
Option 2 has the condition issue. I have no idea how to fix it.

aderm:
> Option 2 has the condition issue. I have no idea how to fix it.

condition can more detail？ From the front debug, the parameters such as `startTimeBucket endTimeBucket step` are passed, but the step param， current interface `queryBasicTraces ` is not used.

aderm:
@wu-sheng @kezhenxu94 `queryBasicTraces` query index of segment is located according to the date range, PTAL.

wu-sheng:
> condition issue can more detail？

When we do query based on trace_id, don't need time duration.

wu-sheng:
@aderm Do we still need to keep this PR continued? This PR still seems changing a lot, but the report shows only trace query maybe need optimization for very few users.

aderm:
attention to this, other modifications can be removed. There is a suggestion. For trace query, add a configuration parameter. It is not enabled by default and is enabled when needed. WDYT?

wu-sheng:
I would support removing other besides the trace. And I prefer to keep this open and not included in 7.0.0. We could be back to this if there is someone(s) pay attentions on this because of their production cases.

aderm:
agree

PR #4356, opened 2020-02-13
Fill some comments for core.

wu-sheng:
Spend 1 hour to fill more comments in the core, should be helpful for people to read the source codes. 

FYI @apache/skywalking-committers Feel free to add more in the parts you lead. Within the whole 6.x and 7.x process, the core is stable. I think it is reasonable to fill more comments.


JaredTan95:
seems CI is not stable~

wu-sheng:
Checking whether there is format issue now.

PR #4361, opened 2020-02-14
More comments for important classes

wu-sheng:
Hi @apache/skywalking-committers 

This is the 3rd PR related to code comments, through #4356 #4357 and this, most important classes and interfaces have good explanation comments. These comments should help the new contributors to understand the codes better. This should be my last one in the short term, unless any of you think there are some important classes I should fill too.

Like always, welcome any committer or contributor to send pull requests to fill more comments.


wu-sheng:
In there 3 PRs, I add comments for nearly 100 files :)

PR #4362, opened 2020-02-14
Adapt Armeria plugin to version 0.98.0

kezhenxu94:
### Motivation:

Support Armeria 0.98.0 in its plugin.

### Modifications:

Add an intercept point that is refactored in Armeria 0.98.

### Result:

Now Armeria plugin supports 0.98.0, and 0.98.1

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/armeria-0.85.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/armeria/define/Armeria086ClientInstrumentation.java
+++ b/apm-sniffer/apm-sdk-plugin/armeria-0.85.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/armeria/define/Armeria086ClientInstrumentation.java
@@ -29,6 +29,9 @@ import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
 import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
 import static org.apache.skywalking.apm.agent.core.plugin.match.NameMatch.byName;
 
+/**
+ * Instruments Armeria client 0.86.x ~ 0.97.x
+ */
 public class Armeria086ClientInstrumentation extends ClassInstanceMethodsEnhancePluginDefine {
     private static final String ENHANCE_CLASS = "com.linecorp.armeria.client.UserClient";
     private static final String INTERCEPTOR_CLASS = "org.apache.skywalking.apm.plugin.armeria.Armeria086ClientInterceptor";
--- a/apm-sniffer/apm-sdk-plugin/armeria-0.85.x-plugin/pom.xml
+++ b/apm-sniffer/apm-sdk-plugin/armeria-0.85.x-plugin/pom.xml
@@ -28,7 +28,7 @@
     <artifactId>apm-armeria-0.85.x-plugin</artifactId>
     <name>armeria-0.85.x-plugin</name>
     <packaging>jar</packaging>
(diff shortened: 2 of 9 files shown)


test/plugin/scenarios/armeria-0.96plus-scenario/support-version.list
```diff
@@ -14,5 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+0.98.0
```
kezhenxu94: @wu-sheng let me add this version to verify it works (although I verified locally), and revert this to merge #4328 :)

kezhenxu94:
here are the changes between 0.97.0 and 0.98.0 that breaks the plugin

https://github.com/line/armeria/blob/aba612a0e6b980334c5a40818047efe2aecc8e4e/core/src/main/java/com/linecorp/armeria/client/UserClient.java#L133-L135

https://github.com/line/armeria/blob/d5ba7e73cd6d4df72e6ed55fe9153b6f7a7f19e9/core/src/main/java/com/linecorp/armeria/client/UserClient.java#L145-L147

apm-sniffer/apm-sdk-plugin/armeria-0.85.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/armeria/define/Armeria086ClientInstrumentation.java
```diff
@@ -32,6 +32,7 @@
 public class Armeria086ClientInstrumentation extends ClassInstanceMethodsEnhancePluginDefine {
```
wu-sheng: This class name is not suitable anymore. Please consider to refactor?

test/plugin/scenarios/armeria-0.96plus-scenario/support-version.list
wu-sheng: Make sense to me.

apm-sniffer/apm-sdk-plugin/armeria-0.85.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/armeria/define/Armeria086ClientInstrumentation.java
kezhenxu94: Sure, what about renaming to the lowest version that the instrumentation supports from? For example Armeria096ClientInstrumentation indicates that it supports 0.96+, and if it supports all future versions that would be best, if it doesn’t support from, for example 1.2.3, we will have another Armeria123...? 


But I’m recently consider to rewrite the plugin from 1.0.0 with a more recommended tracing solution in Armeria, decorator, but that’s another story.

wu-sheng: > For example Armeria096ClientInstrumentation indicates that it supports 0.96+

That totally works for me. 

> But I’m recently consider to rewrite the plugin from 1.0.0 with a more recommended tracing solution in Armeria, decorator, but that’s another story.

If that could replace some of the existing plugins, we could delete them later. It is not an issue.

wu-sheng:
Tests seem passed, please continue.

kezhenxu94:
The failure is due to network issue and I had re-run it several times, I'll merge this instead of re-running over and over

wu-sheng:
Go for that. It is fine.

PR #4363, opened 2020-02-15
Fix non-atomic operations on volatile variables

kezhenxu94:
### Motivation:

Fix non-atomic operations on `volatile` variables.

### Modifications:

Replace `int`, `long` with `AtomicInteger` and `AtomicLong`.

### Result:

No potential threading visibility issues caused by non-atomic operations on `volatile` variables.

### DIFF
--- a/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/MultipleChannelsConsumer.java
+++ b/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/MultipleChannelsConsumer.java
@@ -20,6 +20,7 @@ package org.apache.skywalking.apm.commons.datacarrier.consumer;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicLong;
 import org.apache.skywalking.apm.commons.datacarrier.buffer.Channels;
 import org.apache.skywalking.apm.commons.datacarrier.buffer.QueueBuffer;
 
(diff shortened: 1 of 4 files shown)


apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/MultipleChannelsConsumer.java
```diff
@@ -30,12 +31,13 @@
 public class MultipleChannelsConsumer extends Thread {
     private volatile boolean running;
     private volatile ArrayList<Group> consumeTargets;
-    private volatile long size;
+    private final AtomicLong size;
```
wu-sheng: This is just for changed value accessable from other consumer thread. No requirement for AtomicLong.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/GRPCChannelManager.java
```diff
@@ -48,7 +49,7 @@
     private final List<GRPCChannelListener> listeners = Collections.synchronizedList(new LinkedList<>());
     private volatile List<String> grpcServers;
     private volatile int selectedIdx = -1;
-    private volatile int reconnectCount = 0;
+    private AtomicInteger reconnectCount = new AtomicInteger(0);
```
wu-sheng: Why AtomicInteger?

apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/partition/SimpleRollingPartitioner.java
```diff
@@ -18,15 +18,17 @@
...
 package org.apache.skywalking.apm.commons.datacarrier.partition;
 
+import java.util.concurrent.atomic.AtomicInteger;
+
 /**
  * use normal int to rolling.
  */
 public class SimpleRollingPartitioner<T> implements IDataPartitioner<T> {
-    private volatile int i = 0;
+    private AtomicInteger i = new AtomicInteger(0);
```
wu-sheng: This will cause performance cost without much meaning. Rolling is for switching different channels, there is no need to be strict rule.

PR #4364, opened 2020-02-15
Support Downsampling Data Packing feature in ES storage implementation

wu-sheng:
### Downsampling Data Packing

Downsampling data packing(`storage/elasticsearch/enablePackedDownsampling`, default activated) is a new feature since 7.0.0.  
Metrics data has 4 different precisions, based on `core/default/downsampling` configurations. 
In previous(6.x), every precision of each metrics had one separated
index. After this is activated, metrics of day and hour precisions are merged into minute precision. The number of indexes
decreased, and cause less payload to the ElasticSearch server.

### The new data format preview
image
After enabling this feature, 3 precisions of metrics data are all existing in one index(minute precision).

### Why query could keep in the same.
From my understanding, once query could use the right name(merged index name), as they are using different timestamp format(because of precision, length is different), they will exclude other precision data automatically. So no change is required.

____
To all @apache/skywalking-committers 
This PR has a small(18 files only) change set, but influenced scope is large. And this doesn't have the side-effect, so I active it by default. Please review and give me some feedback.

### DIFF
--- a/oap-server/server-bootstrap/src/main/resources/application.yml
+++ b/oap-server/server-bootstrap/src/main/resources/application.yml
@@ -83,6 +83,7 @@ storage:
 #    #trustStorePass: ${SW_SW_STORAGE_ES_SSL_JKS_PASS:""}
 #    user: ${SW_ES_USER:""}
 #    password: ${SW_ES_PASSWORD:""}
+#    enablePackedDownsampling: ${SW_STORAGE_ENABLE_PACKED_DOWNSAMPLING:true} # Hour and Day metrics will be merged into minute index.
 #    indexShardsNumber: ${SW_STORAGE_ES_INDEX_SHARDS_NUMBER:2}
 #    indexReplicasNumber: ${SW_STORAGE_ES_INDEX_REPLICAS_NUMBER:0}
 #    # Those data TTL settings will override the same settings in core module.
@@ -104,6 +105,7 @@ storage:
     protocol: ${SW_STORAGE_ES_HTTP_PROTOCOL:"http"}
     #trustStorePath: ${SW_SW_STORAGE_ES_SSL_JKS_PATH:"../es_keystore.jks"}
     #trustStorePass: ${SW_SW_STORAGE_ES_SSL_JKS_PASS:""}
+    enablePackedDownsampling: ${SW_STORAGE_ENABLE_PACKED_DOWNSAMPLING:true} # Hour and Day metrics will be merged into minute index.
     user: ${SW_ES_USER:""}
     password: ${SW_ES_PASSWORD:""}
     indexShardsNumber: ${SW_STORAGE_ES_INDEX_SHARDS_NUMBER:2}
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchProvider.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchProvider.java
@@ -203,22 +203,20 @@ public class StorageModuleElasticsearchProvider extends ModuleProvider {
 
         if (enablePackedDownsampling) {
             // Packed downsampling converter.
-            converters.add(new PackedDownsamplingConverter(enablePackedDownsampling));
-            converters.add(new NamespaceConverter(namespace));
+            converters.add(new PackedDownsamplingConverter());
         }
+        converters.add(new NamespaceConverter(namespace));
         return converters;
     }
 
     private static class PackedDownsamplingConverter implements IndexNameConverter {
-        private final boolean enablePackedDownsampling;
         private final String[] removableSuffixes = new String[] {
             Const.ID_SPLIT + Downsampling.Day.getName(),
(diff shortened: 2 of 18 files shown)


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchProvider.java
```diff
@@ -174,4 +197,63 @@ private void overrideCoreModuleTTLConfig() {
         configService.getDataTTLConfig().setDayMetricsDataTTL(config.getDayMetricsDataTTL());
         configService.getDataTTLConfig().setMonthMetricsDataTTL(config.getMonthMetricsDataTTL());
     }
+
+    public static List<IndexNameConverter> indexNameConverters(String namespace, boolean enablePackedDownsampling) {
+        List<IndexNameConverter> converters = new ArrayList<>();
+
+        if (enablePackedDownsampling) {
+            // Packed downsampling converter.
+            converters.add(new PackedDownsamplingConverter(enablePackedDownsampling));
```
kezhenxu94: nit: `enablePackedDownsampling` is always `true`, and it's never used in class `PackedDownsamplingConverter `

wu-sheng:
### Explanation of No Side Effect
The size of dataset is the root reason of no side effect conclusion. In previous 6.x, minute/hour/day precision data are all organized day by day. So, in one day, the sizes of dataset are 60 * 24(minute), 24(hour) and 1(day) for a single entity(service, instance, endpoint or relationship). We merge them into one, then the increasement of one index is (24 + 1)/(60*24) = 1.7%, but reduce 2 indexes for one metrics. This is not a tradeoff, it is a pure enhancement.

### Added bonus
Because of the performance enhancement, the e2e tests will be more stable due to less indexes.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchProvider.java
wu-sheng: Fixed.

kezhenxu94:
Does this discuss elsewhere(I have impressions)? What about the performance of writing to/querying from the merged index? Because the amount of the data in the index grows, and I’m wondering whether it’s “pure enhancement” or not

wu-sheng:
> Does this discuss elsewhere(I have impressions)? What about the performance of writing to/querying from the merged index? Because the amount of the data in the index grows, and I’m wondering whether it’s “pure enhancement” or not

No, there isn't. I got reports from others, there is performance concern about the index number. I did the math, look at the previous comment, https://github.com/apache/skywalking/pull/4364#issuecomment-586596886.

wu-sheng:
About `writing to/querying from the merged index`, we always query/write from metrics in minute precision, if there is no performance issue there, there is no issue for merged index. Right?

aderm:
> About `writing to/querying from the merged index`, we always query/write from metrics in minute precision, if there is no performance issue there, there is no issue for merged index. Right?

Regarding the query of the index, if 3 precisions of metrics are merged into the minute index, have two small problems:
1. When the ES server loads index memory data, because lucene index storage is similar to columnar storage, the merge will take up more memory. Will timebucket or other field queries consume more resources? evaluate the memory resources at the minute level? Although the result set of the search is the same size;
2. When the SW daily data volume is large, reaching a single index to the level of 10 million or 100 million, will the performance not change if the index conditions are not changed? Maybe more scenarios are minute-level queries.

wu-sheng:
Hi @aderm What are your questions indicating? I am a little confused about reading this.

aderm:
just now get flower mean, the data only increased by 1.7%, which can basically be ignored.So the question I worry about doesn't exist.

> The size of dataset is the root reason of no side effect conclusion. In previous 6.x, minute/hour/day precision data are all organized day by day. So, in one day, the sizes of dataset are 60 * 24(minute), 24(hour) and 1(day) for a single entity(service, instance, endpoint or relationship). We merge them into one, then the increasement of one index is (24 + 1)/(60*24) = 1.7%, but reduce 2 indexes for one metrics. This is not a tradeoff, it is a pure enhancement

oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -104,6 +105,7 @@ storage:
     protocol: ${SW_STORAGE_ES_HTTP_PROTOCOL:"http"}
     #trustStorePath: ${SW_SW_STORAGE_ES_SSL_JKS_PATH:"../es_keystore.jks"}
     #trustStorePass: ${SW_SW_STORAGE_ES_SSL_JKS_PASS:""}
+    enablePackedDownsampling: ${SW_STORAGE_ENABLE_PACKED_DOWNSAMPLING:true} # Hour and Day metrics will be merged into minute index.
     user: ${SW_ES_USER:""}
     password: ${SW_ES_PASSWORD:""}
     indexShardsNumber: ${SW_STORAGE_ES_INDEX_SHARDS_NUMBER:2}
```
hanahmily: Do we need to increase it when packed downsampling enabled?

PR #4368, opened 2020-02-16
Support Daily Index Step feature in the OAP

wu-sheng:
### Daily Index Step
Daily index step(`storage/elasticsearch/dayStep`, default 1) represents the index creation period. In this period, several days(dayStep value)' metrics are saved.

Mostly, users don't need to change the value manually. As SkyWalking is designed to observe large scale distributed system.
But in some specific cases, users want to set a long TTL value, such as more than 60 days, but their ElasticSearch cluster isn't powerful due to the low traffic in the production environment.
This value could be increased to 5(or more), if users could make sure single one index could support these days(5 in this case) metrics and traces.

Such as, if dayStep == 11, 
1. data in [2000-01-01, 2000-01-11] will be merged into the index-20000101.
1. data in [2000-01-12, 2000-01-22] will be merged into the index-20000112.

NOTICE, TTL deletion would be affected by these. You should set an extra more dayStep in your TTL. Such as you want to TTL == 30 days and dayStep == 10, you actually need to set TTL = 40;

### Data in the storage
The data in the dayStep period are merged into one index.
image

### DIFF
--- a/dist-material/application.yml
+++ b/dist-material/application.yml
@@ -76,52 +76,54 @@ core:
     enableDatabaseSession: ${SW_CORE_ENABLE_DATABASE_SESSION:true}
     topNReportPeriod: ${SW_CORE_TOPN_REPORT_PERIOD:10} # top_n record worker report cycle, unit is minute
 storage:
-#  elasticsearch:
-#    nameSpace: ${SW_NAMESPACE:""}
-#    clusterNodes: ${SW_STORAGE_ES_CLUSTER_NODES:localhost:9200}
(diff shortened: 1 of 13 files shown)


wu-sheng:
@aderm This is another challenge for the PR you are working on, also one of the reason I want to hold that PR(#4353). This will give the lower traffic user more friendly UX.

JaredTan95:
LGTM
After https://github.com/apache/skywalking/pull/4214 about es index number reduced. This is another optimization.

wu-sheng:
@JaredTan95 Yes, so if the user only have low traffic, they could have 70-80 indexes but hosting 1-2weeks later with a small ES cluster.

Issue #4370, opened 2020-02-16
Topology view page ES node type show Unknow.


aderm:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS and JRE?

- Which company or project?

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.

<img width="743" alt="WX20200216-200511@2x" src="https://user-images.githubusercontent.com/2892433/74604365-49e3d180-50f8-11ea-9ddd-3b7193854bf0.png">
<img width="1081" alt="WX20200216-200600@2x" src="https://user-images.githubusercontent.com/2892433/74604371-510adf80-50f8-11ea-957d-b90466830dc5.png">

___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

wu-sheng:
Should be replaced by #4386

PR #4371, opened 2020-02-16
fix topology node type unknown.

aderm:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
#4370 
___
### Bug fix
- Bug description.
Interceptor uses different components to intercept in different requests, httyasyncclient ( `componentId=26` ) and rest-highlevel-client ( `componentId=77` ), but the node type cannot be identified by httyasyncclient.

- How to fix?
Update with Known Node Type
___


wu-sheng:
> Interceptor uses different components to intercept in different requests, httyasyncclient ( componentId=26 ) and rest-highlevel-client ( componentId=77 ), but the node type cannot be identified by httyasyncclient.

I could understand your fix, but the above description seems that, unknown caused by `httyasyncclient` doesn't have server side component mapping. If that is set, you are facing two kinds of target components.

I am not objecting this fix, I just mean, the component id still could be wrong.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/TopologyBuilder.java
```diff
@@ -78,11 +79,20 @@ Topology build(List<Call.CallDetail> serviceRelationClientCalls, List<Call.CallD
                 nodes.put(source.getSequence(), buildNode(source));
             }
 
-            if (!nodes.containsKey(target.getSequence())) {
-                nodes.put(target.getSequence(), buildNode(target));
+            int sequence = target.getSequence();
+            if (nodes.containsKey(sequence)) {
+                //if node type is Unknown, but some targer know node type, so update the node type.
```
wu-sheng:
```suggestion
                //if node type is Unknown, but the target may know the node type.
```

aderm:
> > Interceptor uses different components to intercept in different requests, httyasyncclient ( componentId=26 ) and rest-highlevel-client ( componentId=77 ), but the node type cannot be identified by httyasyncclient.
> 
> I could understand your fix, but the above description seems that, unknown caused by `httyasyncclient` doesn't have server side component mapping. If that is set, you are facing two kinds of target components.
> 
> I am not objecting this fix, I just mean, the component id still could be wrong.

not httyasyncclient but `httyasyncclient`.
`httpasyncclient` belongs to a common network component. cannot really infer the node type through it, and its type cannot be set.

Theoretically, there should be only one node type. If there are two different types inferred from A and B, it can only be a configuration error.

wu-sheng:
> not httyasyncclient, is httpasyncclient.
httpasyncclient belongs to a common network component. cannot really infer the node type through it, and its type cannot be set.

My point is, someone could set that mapping, then you will be back to facing two component.

Again, I am not saying this chance is wrong, but this could not fix the issue in some cases.

BTW, you need to fix the tests.

aderm:
> My point is, someone could set that mapping, then you will be back to facing two component.

Do need to add another check mechanism to prevent duplicate configuration?

wu-sheng:
> Do need to add another check mechanism to prevent duplicate configuration?

We can't. And there is another question about ES plugin, FYI @dmsolr , why ES client creates two exit spans? If two spans created, the traffic(such as CPM) of ES server seems wrong(1 request == 2 exit span == 2 request for ES server)

wu-sheng:
We need to look deeper what the scenario you described. Inside `elasticsearch-7.x-scenario`, I can't find the span you described, is it missed?

aderm:
> We need to look deeper what the scenario you described. Inside `elasticsearch-7.x-scenario`, I can't find the span you described, is it missed?

client-side info httyasyncclient ( componentId=26 ) and rest-highlevel-client ( componentId=77 ), are generated by two different endpoint Requests. For httyasyncclient (componentId = 26) the plugin scenario test healthCheck  interface `
 ClusterHealthResponse response = client.cluster (). Health (request, RequestOptions.DEFAULT) ` generate exitspan; rest-highlevel-client (componentId = 77) the entry span generated by other requests that intercept ES queries through rest-highlevel-client.

WX20200218-165406@2x
WX20200218-165406@2x

wu-sheng:
> For httyasyncclient (componentId = 26) the plugin scenario test healthCheck interface ClusterHealthResponse response = client.cluster (). Health (request, RequestOptions.DEFAULT) generate exitspan

Then this is the root cause. We should fix this create a exist span in ES, wrapping the existing httyasyncclient span. Then this component=26 span would be overrided.

aderm:
> > For httyasyncclient (componentId = 26) the plugin scenario test healthCheck interface ClusterHealthResponse response = client.cluster (). Health (request, RequestOptions.DEFAULT) generate exitspan
> 
> Then this is the root cause. We should fix this create a exist span in ES, wrapping the existing httyasyncclient span. Then this component=26 span would be overrided.

the problem is that there are many other rest-highlevel-client interfaces similar to the above. need to all wrap in rest-highlevel-client interceptor? ES another rest client `Low-Level-Rest-Client`  which the underlying is httyasyncclient Implementation may also have the same problem.

wu-sheng:
Yes. They should be. All request starts from es client jar, should be wrapped inside the es exit span.

PR #4379, opened 2020-02-17
Fix ehcache:  missing interceptor of private constructor and setName method

lsyf:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

___
### Bug fix
- Bug description.
When the instance of `Cache`  is created by the `clone` method,  all the codes `EnhancedInstance # getSkyWalkingDynamicField ()`  will return null in the ehcache interceptor .
**This will cause a `NullPointerException`.**
Beacase the code `EnhancedInstance # setSkyWalkingDynamicField()` is only called in the constructor's interceptor. But the private constructor and `setName`  method can not intercepted by the interceptor.

- How to fix?
Add interceptors of  private constructor and `setName` method


wu-sheng:
`/test/plugin/scenarios/ehcache-2.x-scenario` needs to be updated too, read [plugin test doc](https://github.com/apache/skywalking/blob/master/docs/en/guides/Plugin-test.md) to get more details. This is an easy update.

wu-sheng:
I think this fix needs to be improved. The reason of `#clone` not working is not caused by it is a private method, it is because the original intercept rule is targeting construct with specific parameter.

Notice this
```java
    @Override
    public ConstructorInterceptPoint[] getConstructorsInterceptPoints() {
        return new ConstructorInterceptPoint[] {
            new ConstructorInterceptPoint() {
                @Override
                public ElementMatcher<MethodDescription> getConstructorMatcher() {
                    return takesArgument(0, named("net.sf.ehcache.config.CacheConfiguration"));
                }

                @Override
                public String getConstructorInterceptor() {
                    return CONSTRUCTOR_CLASS_INTERCEPT_CLASS;
                }
            }
        };
    }
```

So basically, the better fix solution is add another constructor interceptor point to target the private constructor, I assume this one `private Cache(Cache original) throws CloneNotSupportedException`?

@lsyf What do you think?

wu-sheng:
Also, your plugin test is fail, please test locally.

lsyf:
> I think this fix needs to be improved. The reason of `#clone` not working is not caused by it is a private method, it is because the original intercept rule is targeting construct with specific parameter.
> 
> Notice this
> 
> ```java
>     @Override
>     public ConstructorInterceptPoint[] getConstructorsInterceptPoints() {
>         return new ConstructorInterceptPoint[] {
>             new ConstructorInterceptPoint() {
>                 @Override
>                 public ElementMatcher<MethodDescription> getConstructorMatcher() {
>                     return takesArgument(0, named("net.sf.ehcache.config.CacheConfiguration"));
>                 }
> 
>                 @Override
>                 public String getConstructorInterceptor() {
>                     return CONSTRUCTOR_CLASS_INTERCEPT_CLASS;
>                 }
>             }
>         };
>     }
> ```
> 
> So basically, the better fix solution is add another constructor interceptor point to target the private constructor, I assume this one `private Cache(Cache original) throws CloneNotSupportedException`?
> 
> @lsyf What do you think?

yes. Beacause the purpose of the interceptor is to save cache name，`#setName`  also need to be intercepted

lsyf:
> Also, your plugin test is fail, please test locally.

sorry. My computer operating system is windows. I cannot install docker due to company security software restrictions.  So I can only check by myself to make sure there are no errors

lsyf:
The weird thing is that I haven't changed the test cases and they will go wrong.

apm-sniffer/apm-sdk-plugin/ehcache-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/ehcache/v2/EhcacheConstructorInterceptor.java
```diff
@@ -18,18 +18,28 @@
...
+            CacheConfiguration cacheConfiguration = (CacheConfiguration) allArguments[0];
 
-        // get cache name
-        if (cacheConfiguration != null) {
-            objInst.setSkyWalkingDynamicField(new EhcacheEnhanceInfo(cacheConfiguration.getName()));
+            // get cache name
+            if (cacheConfiguration != null) {
+                objInst.setSkyWalkingDynamicField(new EhcacheEnhanceInfo(cacheConfiguration.getName()));
+            }
+        } catch (ClassCastException e) {
```
wu-sheng: You don't need to use this `try/catch`. `getConstructorsInterceptPoints` returns an array, you should set up the interceptors separately. This one you didi has a performance issue.

lsyf: Can I use `instanseof` instead? 
```java
       if (argument instanceof CacheConfiguration) {
            CacheConfiguration cacheConfiguration = (CacheConfiguration) argument;

            // get cache name
            if (cacheConfiguration != null) {
                objInst.setSkyWalkingDynamicField(new EhcacheEnhanceInfo(cacheConfiguration.getName()));
            }

        } else if (argument instanceof Cache) {
            Cache cache = (Cache) argument;

            // get cache name
            if (cache != null && cache.getCacheConfiguration() != null) {
                objInst.setSkyWalkingDynamicField(new EhcacheEnhanceInfo(cache.getCacheConfiguration().getName()));
            }
        }
```

wu-sheng: Why? `instanceof` is still much much slower than instrumentation level match.

lsyf: ok. Thank you for your seriousness

apm-sniffer/apm-sdk-plugin/ehcache-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/ehcache/v2/define/EhcachePluginInstrumentation.java
```diff
@@ -82,13 +88,41 @@
...
+
+                @Override
+                public String getMethodsInterceptor() {
+                    return CACHE_NAME_INTERCEPTOR_CLASS;
+                }
+
+                @Override
+                public boolean isOverrideArgs() {
+                        return false;
+                    }
```
mrproliu: Is this a code format error?

lsyf: Yes. Do I need to format and check?

wu-sheng: Yes, you should. The format check will be more strict soon for the test case.

wu-sheng:
@mrproliu Please don't use `merge` directly. because of this, this contributor wouldn't be counted, and the commit log breaks with multiple details commits.

PR #4383, opened 2020-02-19
Update committer review and merge guide

kezhenxu94:
Since we have more and more committers, this patch updates some worth-noted guide when committers are merging pull requests, avoid too many useless commits in Git log, like #4379

@apache/skywalking-committers and @aderm , please take some time to go through the documentation :)

### DIFF
--- a/docs/en/guides/asf/committer.md
+++ b/docs/en/guides/asf/committer.md
@@ -121,11 +121,11 @@ Notice, if never got your GitHub invitation? Visit github.com/apache to see if y
 If you want others could see you are in the Apache GitHub org, you need to go to [Apache GitHub org people page](https://github.com/orgs/apache/people), 
 search for yourself, and choose `Organization visibility` to `Public`.
 
-### Committer right, duty and responsibility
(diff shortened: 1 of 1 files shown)


wu-sheng:
The last PR breaks this, do we need to close another merge option?

kezhenxu94:
> The last PR breaks this, do we need to close another merge option?

I prefer to close other merge options, it seems that we don't have any scenario to merge without squashing so far

wu-sheng:
We lost the 208th contributor metrics because of that merge.

wu-sheng:
@kezhenxu94 Please help on submitting the JIRA ticket to close `Allow merge commits ` and keep `Allow squash merging`  only. I am failing to open JIRA site.

kezhenxu94:
> @kezhenxu94 Please help on submitting the JIRA ticket to close `Allow merge commits ` and keep `Allow squash merging` only. I am failing to open JIRA site.

https://issues.apache.org/jira/browse/INFRA-19869

PR #4385, opened 2020-02-19
Fix NPE of decoding IntKeyLongValuesHashMap

hanahmily:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

If IntKeyLongValuesHashMap#toStorageDate generate an empty string,
that usually because there's no any traffic in one minute. When syncing
it to cache, IntKeyLongValuesHashMap#toObject will raise a NPE.

- How to fix?
if an empty string pass into toObject, just return

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/IntKeyLongValueHashMap.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/IntKeyLongValueHashMap.java
@@ -17,6 +17,7 @@
 
 package org.apache.skywalking.oap.server.core.analysis.metrics;
 
+import com.google.common.base.Strings;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/IntKeyLongValueHashMap.java
```diff
@@ -57,6 +58,9 @@ public String toStorageData() {
 
     @Override
     public void toObject(String data) {
+        if (Strings.isNullOrEmpty(data)) {
```
wu-sheng: Personally, I want to say, this is harmful change. This `data` should not generate if there is no value of this column. `Calculate` func should make sure the `value` column is generated rather than using this `null` check. Such as `PercentileMetrics#calculate`, it guarantees the 5 values generated permanently. If there is no value, put the default value based on this metrics func, otherwise, which value should be shown up in the final query?

wu-sheng: @hanahmily Any feedback?

Issue #4388, opened 2020-02-20
Question about trace data ttl


dengliming:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [x] Question or discussion
- [ ] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?
I found that only the last two days of trace data can be searched at webapp(UI). How can I extend the time? Please help me to understand. Thanks.

Here is my es storage setting of ttl:
```
recordDataTTL: ${SW_STORAGE_ES_RECORD_DATA_TTL:15} # Unit is day
otherMetricsDataTTL: ${SW_STORAGE_ES_OTHER_METRIC_DATA_TTL:30} # Unit is day
monthMetricsDataTTL: ${SW_STORAGE_ES_MONTH_METRIC_DATA_TTL:6} # Unit is month
```

wu-sheng:
`recordDataTTL` is the one. Which version are you using?

dengliming:
version:6.6.0, Why close so fast (×_×)

wu-sheng:
We keep issue opened for bug and new feature discussion.

From my understanding, there isn't a bug about it. Simply, check the demo env, there is still trace in several days ago. It uses this setting.

PR #4391, opened 2020-02-21
Fix otherMetricsDataTTL config not working

dengliming:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
There is no setter of otherMetricsDataTTL method in StorageModuleElasticsearchConfig.This can lead to dayMetrics only stored 2 days default.
See #4388
- How to fix?
Add setter annotation.
___

### DIFF
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java
@@ -95,6 +95,7 @@ public class StorageModuleElasticsearchConfig extends ModuleConfig {
     private int hourMetricsDataTTL = 2;
     @Setter
     private int dayMetricsDataTTL = 2;
+    @Setter
     private int otherMetricsDataTTL = 0;
     @Setter
     private int monthMetricsDataTTL = 18;


dengliming:
Why need? Please check the code. you will see.
org.apache.skywalking.oap.server.storage.plugin.elasticsearch.StorageModuleElasticsearchConfig#getDayMetricsDataTTL

```
public int getDayMetricsDataTTL() {
        if (otherMetricsDataTTL > 0) {
            return otherMetricsDataTTL;
        }
        return dayMetricsDataTTL;
    }
```
If otherMetricsDataTTL config not working, otherMetricsDataTTL is always be zero,Then there is no
meaning here.

wu-sheng:
I remember setter is not required, are you sure this will be an issue?

dengliming:
> I remember setter is not required, are you sure this will be an issue?

I think is the reason for this problem. [https://github.com/apache/skywalking/issues/4388](https://github.com/apache/skywalking/issues/4388),It is a bug(I think)

wu-sheng:
I will check later. Currently, I am not sure.

wu-sheng:
In that PR,  you talked about trace TTL, but this one has nothing related to that, no matter it is a big. Are you are the thing? I doubt it.

dengliming:
> In that PR, you talked about trace TTL, but this one has nothing related to that, no matter it is a big. Are you are the thing? I doubt it.

I don't figure out why my setting not working,It seems only keep two days trace data.So I checked the code.And I'm not boring!

wu-sheng:
My point is, the codes you posted are not related to trace TTL at all. That is my question.

wu-sheng:
getDayMetricsDataTTL methd is for metrics ttl at the day precision level.

dengliming:
Thanks for  your detailed explanation.It's my misunderstanding.Digression If i want to keep more days trace data.How to do it?

wu-sheng:
image

My local test confirmed, this value could be set without `@setter`. Also, according to code `ModuleDefine#copyProperties`, it is using refactor to access field, rather than set method.

So, this is not the issue. even this is not an issue generally, no matter related to your question or not

wu-sheng:
This is the config of the official DEMO env using. It works as expected.

image

shook2012:
At the version 6.3.0，I'm sure it's a bug。
No matter how we modify the configuration ：otherMetricsDataTTL under the core or storage， the metric data only can stored 2 days 。

shook2012:
@wu-sheng

shook2012:
The envs:centos 6  and jdk 1.8.131

wu-sheng:
The whole thing has been long gone. Recommend you to upgrade to the latest release.

shook2012:
okay,  Thanks~

PR #4395, opened 2020-02-21
[test/plugin] immigrate h2 & struts scenario

arugal:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#3583

### DIFF
--- a/.github/workflows/plugins-test.yaml
+++ b/.github/workflows/plugins-test.yaml
@@ -29,7 +29,7 @@ on:
 
 jobs:
   Spring_Dubbo:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     timeout-minutes: 90
     strategy:
       fail-fast: true
@@ -54,7 +54,7 @@ jobs:
         run: bash test/plugin/run.sh okhttp-scenario
 
   Spring31x_ES5_Gateway_Avro_Struts25:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     timeout-minutes: 90
     strategy:
       fail-fast: true
@@ -87,7 +87,7 @@ jobs:
         run: bash test/plugin/run.sh struts2.5-scenario
 
   Toolkit_Webflux_SocketIO_RabbitMQ_GSON:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     timeout-minutes: 90
     strategy:
       fail-fast: true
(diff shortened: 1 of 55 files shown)


.github/workflows/plugins-test.yaml
```diff
@@ -53,7 +53,7 @@ jobs:
       - name: Run okhttp 3.0.x-3.14.x (34)
         run: bash test/plugin/run.sh okhttp-scenario
 
-  Spring31x_ES5_Gateway_Avro:
+  Spring31x_ES5_Gateway_Avro_Struts2.5:
```
kezhenxu94: `.` is not valid as a yaml section name, unless it’s quoted

.github/workflows/plugins-test.yaml
```diff
@@ -333,7 +335,7 @@ jobs:
       - name: Run zookeeper 3.4.x (14)
         run: bash test/plugin/run.sh zookeeper-scenario
 
-  Spring43:
+  Spring43_Struts2.3:
```
kezhenxu94: Ditto

.github/workflows/plugins-test.yaml
arugal: done

.github/workflows/plugins-test.yaml
arugal: done

test/plugin/scenarios/struts2.5-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,49 @@
...
+            operationId: 0
+            parentSpanId: -1
+            spanId: 0
+            spanLayer: Http
+            startTime: gt 0
+            endTime: gt 0
+            componentId: 15
+            componentName: ''
+            isError: false
+            spanType: Entry
```
wu-sheng: I prefer to enhance this case, we should check the ref by adding a http call.

arugal: Done.

wu-sheng:
Strange. Look like checkout failure?

wu-sheng:
@arugal You could rerun from the `Checks` tab.

arugal:
> @arugal You could rerun from the `Checks` tab.

Rerun still failed.

kezhenxu94:
@arugal could you fix the component id in https://github.com/apache/skywalking/blob/master/test/plugin/scenarios/undertow-scenario/config/expectedData.yaml , the component id of Undertow is changed to 84 in #4398 and it will fail then

arugal:
> @arugal could you fix the component id in https://github.com/apache/skywalking/blob/master/test/plugin/scenarios/undertow-scenario/config/expectedData.yaml , the component id of Undertow is changed to 84 in #4398 and it will fail then

Done.

arugal:
@wu-sheng @dmsolr The test has passed, please review.

PR #4397, opened 2020-02-22
Add Nginx component ID to main repo

wu-sheng:
Add 84 as Nginx component ID

### DIFF
--- a/oap-server/server-bootstrap/src/main/resources/component-libraries.yml
+++ b/oap-server/server-bootstrap/src/main/resources/component-libraries.yml
@@ -275,10 +275,6 @@ AvroServer:
 AvroClient:
   id: 83
   languages: Java
-Nginx:
-  id: 84
-  languages: Lua
-
 
 # .NET/.NET Core components
 # [3000, 4000) for C#/.NET only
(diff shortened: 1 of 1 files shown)


kezhenxu94:
@wu-sheng do we want to start from a larger number of id (instead of 84) for the NEW language Lua?

wu-sheng:
> do we want to start from a larger number of id (instead of 84) for the NEW language Lua?

I don't know :P Currently, only one component for Nginx Lua now.

wu-sheng:
@kezhenxu94 If you think we will have more span in Nginx level, I could change that.

aderm:
LGTM

kezhenxu94:
> > do we want to start from a larger number of id (instead of 84) for the NEW language Lua?
> 
> I don't know :P Currently, only one component for Nginx Lua now.

Seems the id ranges are divided according to different languages, instead of frameworks (i.e. Nginx),  if there are any possibility that we may another components written by Lua(not necessarily in Nginx level) , I think it's better to start from another id number and give Lua a range; what's more, 84 makes the Java id range non-continuous (1-83 and 85 to xxx) 😅 odd, harmless though

kezhenxu94:
I'm good to merge this anyway, it does no harm, feel free to merge 😄

wu-sheng:
@kezhenxu94 Updated ID to 6000 as a separated group.

PR #3859, opened 2019-11-14
fix component-libraries.yml miss component:rest-high-level-client

zhaoyuguang:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
 component-libraries.yml miss component:rest-high-level-client
- How to fix?
Add it

PR #4398, opened 2020-02-22
Add missing test for component-libraries.yml

kezhenxu94:
### Motivation:

Add test to guarantee that the components defined in class ComponentsDefine are in sync with those in file component-libraries.yml.

### Modifications:

- Add missing component in component-libraries.yml.

- Add test to verify the class and the yml are in sync.

- Fix wrong component id of Undertow.

### Result:

Components definitions in class ComponentsDefine and yaml are in sync, and will be verified in every pull request.


kezhenxu94:
Related issues are: #3859 #3241 #2679 #2122 , so many

kezhenxu94:
Well, I just forgot to update the component id in plugin test 😢

Issue #4392, opened 2020-02-21
Support choosing multiple services in the topology page


wu-sheng:
As the first step to enhance the topology page, we could support `multiple services` scope topology. Currently, we have one or all. This should be an improvement.

After this, we could support service group in the UI, save the group-services mapping in the Browser cache, to let the user to switch. But this should be the first to go.

I intent to add a new API rather than removing the single service query, but will tag it as `Deprecated`.

wu-sheng:
@TinyAllen The protocol PR has been merged. https://github.com/apache/skywalking-query-protocol/blob/master/topology.graphqls#L91

And the backend PR is on the way #4408.

wu-sheng:
Sub-topology provided through the custom group definition.

PR #4408, opened 2020-02-24
Support topology with multiple selected services

wu-sheng:
This PR is following #4392 and adopt the latest query protocol. Include the following changes

1. Format codes
2. Add comments for storage interface of topology 
3. Narrow the scope of secondary query to get the component name. Only query the nodes as client out of the given service Ids.
4. Adopt the new query parameter easily.

@arugal @kezhenxu94 @dmsolr @aderm Could any of you send a pull request to this branch to add the topology query check by using one given service id and multiple IDs? I think we missed that(given service id) before. This should only be checked in `e2e-cluster-with-gateway` case? WDYT?

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/TopologyQueryService.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/TopologyQueryService.java
@@ -102,10 +102,7 @@ public class TopologyQueryService implements Service {
     }
 
     public Topology getServiceTopology(final Downsampling downsampling, final long startTB, final long endTB,
-                                       final int serviceId) throws IOException {
-        List<Integer> serviceIds = new ArrayList<>();
-        serviceIds.add(serviceId);
-
+                                       final List<Integer> serviceIds) throws IOException {
         List<Call.CallDetail> serviceRelationClientCalls = getTopologyQueryDAO().loadSpecifiedClientSideServiceRelations(
             downsampling, startTB, endTB, serviceIds);
         List<Call.CallDetail> serviceRelationServerCalls = getTopologyQueryDAO().loadSpecifiedServerSideServiceRelations(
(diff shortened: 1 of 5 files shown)


aderm:
i want to try

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/TopologyQueryService.java
```diff
@@ -90,30 +90,44 @@ private EndpointInventoryCache getEndpointInventoryCache() {
...
+         * There is a special case, there may be a node of the `serviceIds` call these services as and only as a client, so it is included in the topology,
+         * its component name could be missed as not being queried before. We add another query about this.
+         */
+        List<Integer> outScopeSourceServiceIds = new ArrayList<>();
+        serviceRelationClientCalls.forEach(call -> {
+            // Client side relationships exclude the given services(#serviceIds)
+            // The given services(#serviceIds)'s component names have been included inside `serviceRelationServerCalls`
+            if (!serviceIds.contains(call.getSource())) {
+                outScopeSourceServiceIds.add(call.getSource());
+            }
```
aderm: when `serviceIds.size() == 0`, all from client side source will query in serverside. same as before. but when `serviceIds.size() != 0`, this will exclude some serviceId which query from server side. There may be a node type, which requires server side to determine, and is related to the selected service.

wu-sheng: All query in the condition service list scope, have been done before, this is just for filling the ones out of scope as the direct linked nodes.

What is your question? I am confused.

aderm: It ’s not too clear why the old code needs to set node type. If this is the case, ones out of scope also does not need to set the type.

wu-sheng: You need to read codes more :) Try to do practice by yourself, it will be much helpful than I told everything. That is open source.

aderm: your are right

wu-sheng:
Hi everyone, this PR becomes a block to @mrproliu . I plan to merge this to adopt the new protocol.

PR #4418, opened 2020-02-26
refactor Metrics class.

aderm:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___
### New feature or improvement
- Describe the details and related test reports.

PR #4419, opened 2020-02-26
[WIP] Fix checks error:  reference is not a tree.

aderm:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?
Refer: https://github.com/actions/checkout/issues/23#issuecomment-572688577
___

### DIFF
--- a/.github/workflows/ci-it.yaml
+++ b/.github/workflows/ci-it.yaml
@@ -33,8 +33,14 @@ jobs:
       fail-fast: true
     steps:
       - uses: actions/checkout@v2
-        with:
-          submodules: true
+      - name: Checkout submodules
+        shell: bash
+        run: |
+          # If your submodules are configured to use SSH instead of HTTPS please uncomment the following line
+          # git config --global url."https://github.com/".insteadOf "git@github.com:"
+          auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+          git submodule sync --recursive
+          git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
       - uses: actions/setup-java@v1
         with:
           java-version: 8
--- a/.github/workflows/ci-it.yaml
+++ b/.github/workflows/ci-it.yaml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: true
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
         with:
           submodules: true
       - uses: actions/setup-java@v1
@@ -52,7 +52,7 @@ jobs:
     strategy:
       fail-fast: true
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
         with:
           submodules: true
       - uses: actions/setup-java@v1
@@ -70,7 +70,7 @@ jobs:
     strategy:
       fail-fast: true
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
         with:
           submodules: true
       - uses: actions/setup-java@v1
--- a/.github/workflows/docker-ci.yaml
+++ b/.github/workflows/docker-ci.yaml
@@ -28,7 +28,7 @@ jobs:
       matrix:
         es: [es6, es7]
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - uses: actions/cache@v1
         with:
           path: ~/.m2/repository
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -28,7 +28,7 @@ jobs:
   Single:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
         with:
           submodules: true
       - name: Set environment
@@ -56,7 +56,7 @@ jobs:
   Cluster:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
         with:
           submodules: true
       - uses: actions/cache@v1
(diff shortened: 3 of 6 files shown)


wu-sheng:
Agree with @kezhenxu94 , please fix this in the #4418, if this is the issue. Then the master commit log makes more sense.

aderm:
test checkout@v2 about submodules.

.github/workflows/ci-it.yaml
```diff
@@ -32,9 +32,15 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v1
-        with:
-          submodules: true
+      - uses: actions/checkout@v2
+      - name: Checkout submodules
+        shell: bash
+        run: |
+          # If your submodules are configured to use SSH instead of HTTPS please uncomment the following line
+          # git config --global url."https://github.com/".insteadOf "git@github.com:"
```
wu-sheng: What these line for? It is confusing me, as we are using https://github.com

.github/workflows/ci-it.yaml
```diff
@@ -32,9 +32,15 @@ jobs:
...
+          auth_header="$(git config --local --get http.https://github.com/.extraheader)"
```
wu-sheng: Why we need auth_header? As we are linking to open source only.

wu-sheng:
I think mostly this PR works in `actions/checkout@v2`, @kezhenxu94 Recheck? Also about my comments.

.github/workflows/ci-it.yaml
kezhenxu94: I think we only need `git submodule sync --recursive`

aderm: > `v2` is not compatible with `v1`, at least the parameters are not compatible, and I suggest you fix this in the pull request where you encounter the issue, which can verify the fix works

Sorry for not being clearly marked. It is true that the current V2 does not support the submodule function. Later, try to complete the verification in the local environment before submitting the PR.

aderm: > What these line for? It is confusing me, as we are using https://github.com

IMO, current is use https, use ssh mean agent by ssh command, such as ssh-agent.
Maybe my understanding is wrong, I did n’t study this deeply.

.github/workflows/ci-it.yaml
aderm: This is a screenshot image when the previous run failed. 
WX20200227-074637@2x
This should be set before. This parameter will be taken after running the command. There is no in-depth study. This is the way to refer to the official action documentation.[https://github.com/actions/checkout#Checkout-submodules](url)

wu-sheng: My point is, you should be safe to run `git submodule update --init --force --recursive --depth=1` directly. I have run this command at local, we don't need that `auth_header`.

aderm:
e2e seems influxdb test failed.

WX20200227-094507@2x

FYI @dmsolr

.github/workflows/ci-it.yaml
aderm: @wu-sheng  Actions execute should be a bit different from local. For this successful example , the `auth_header` parameter be set in the before checkout step `actions/checkout@v2`, So all of the following need to take this parameter.

WX20200227-115704@2x

.github/workflows/ci-it.yaml
aderm: > I think we only need `git submodule sync --recursive`

`git submodule update` updates the contents of the submodules, `git submodule sync` only update metainfo.  FYI  [https://stackoverflow.com/questions/45678862/git-submodule-update-vs-git-submodule-sync/45679230](url)

.github/workflows/ci-it.yaml
wu-sheng: Why should be different? I am not following. We are using the same parameter to download from GitHub too.

Could you explain your meaning of SHOULD?

wu-sheng: Don't copy the thing if that is not necessary.

.github/workflows/ci-it.yaml
wu-sheng: You misunderstand my point. These are comments, what are you trying to say? Is this IF making any sense to anyone? Why We keep this?

.github/workflows/ci-it.yaml
aderm: > Don't copy the thing if that is not necessary.

get it. 


> Why should be different? I am not following. We are using the same parameter to download from GitHub too.
> 
> Could you explain your meaning of SHOULD?

It is indeed that I understand it wrong, `auth_header` can be deleted for sw project, because english is not good, the description is not accurate, it is possible.

.github/workflows/ci-it.yaml
```diff
@@ -32,9 +32,13 @@ jobs:
     strategy:
       fail-fast: true
     steps:
-      - uses: actions/checkout@v1
-        with:
-          submodules: true
+      - uses: actions/checkout@v2
+      # for checkout@v2 is not compatible with submodule
```
wu-sheng:
```suggestion
      # In the checkout@v2, it doesn't support git submodule. Execute the commands manually.
```

wu-sheng: Including all other comments.

.github/workflows/ci-it.yaml
aderm: get it, There is indeed a problem with the comment

.github/workflows/ci-it.yaml
aderm: ok

aderm:
> Once you think this PR is ready, remove the `[WIP]` prefix and ping the reviewers.

get it

wu-sheng:
@kezhenxu94 Need you recheck.

PR #4422, opened 2020-02-27
Support ref endpoint name register

wu-sheng:
The ref exchange is changing since we apply the change in 6.6.0

* Since 6.6.0 The endpoint in the ref should be server endpoint only. The agent will/should use `-1`, when it can't find the endpoint of entry span in the current tracing context when build the ref.
* Since 5.0 Endpoint in ref could be local or exit span's operation name. Especially if it is local span operation name, * exchange may not happen at agent, such as Java agent, then put literal endpoint string in the header, Need to try * to get the id by assuming the endpoint name is detected at server, local or client. If agent does the exchange, then always use endpoint id.

### DIFF
--- a/oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/parser/standardization/ReferenceIdExchanger.java
+++ b/oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/parser/standardization/ReferenceIdExchanger.java
@@ -139,29 +139,19 @@ public class ReferenceIdExchanger implements IdExchanger<ReferenceDecorator> {
                 standardBuilder.setNetworkAddressId(networkAddressId);
                 standardBuilder.setNetworkAddress(Const.EMPTY_STRING);
             }
-        } else {
-            /**
-             * Since 6.6.0, endpoint id could be -1, as it is not an endpoint. Such as local span and exist span.
-             */
         }
         return exchanged;
     }
 
     /**
-     * Endpoint in ref could be local or exit span's operation name. Especially if it is local span operation name,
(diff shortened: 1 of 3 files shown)


oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/parser/standardization/ReferenceIdExchanger.java
```diff
@@ -139,29 +139,19 @@ public boolean exchange(ReferenceDecorator standardBuilder, int serviceId) {
...
     private int getEndpointId(int serviceId, String endpointName) {
-        int endpointId = endpointInventoryRegister.get(serviceId, endpointName, DetectPoint.SERVER.ordinal());
-        if (endpointId == Const.NONE) {
-            endpointId = endpointInventoryRegister.get(serviceId, endpointName, DetectPoint.CLIENT.ordinal());
-            if (endpointId == Const.NONE) {
-                endpointId = endpointInventoryRegister.get(serviceId, endpointName, DetectPoint.UNRECOGNIZED.ordinal());
-            }
-        }
-        return endpointId;
+        return endpointInventoryRegister.getOrCreate(serviceId, endpointName, DetectPoint.SERVER);
```
arugal: It should be `Client` in Ref, right?

wu-sheng: No, it isn't. The operation name is from the entry span, which means, detect from server-side plugin, such as SpringMVC, Tomcat.

arugal: Got it.

PR #4438, opened 2020-03-03
Support change detection mechanism in DCS

wu-sheng:
I add an extra field UUID in the DCS. By using this, if the server side could have the UUID to determine the config changes, it wouldn't required to response the config.

UUID should be unique with the content.

The new proto is like this
```proto
syntax = "proto3";

option java_multiple_files = true;
option java_package = "org.apache.skywalking.oap.server.configuration.service";

service ConfigurationService {
    rpc call (ConfigurationRequest) returns (ConfigurationResponse) {
    }
}

message ConfigurationRequest {
    // Logic name of this cluster,
    // in case the remote configuration center implementation support
    // configuration management for multiple clusters.
    string clusterName = 1;
    // The config UUID response from the config server side.
    string uuid = 2;
}

message ConfigurationResponse {
    // Include all config items.
    // All config name should be not empty,
    // the name is composed by "module name"."provider name"."item name".
    // Each watcher implementor provides this, and it will be notified when the value changed.
    //
    // If the config center wants to set the value to NULL or empty,
    // must set the name with empty value explicitly.
    repeated Config configTable = 1;
    // UUID is literal string represents the content of the config table.
    // If config table is unchanged, then could response the same uuid, and config table is not required.
    string uuid = 2;
}

message Config {
    string name = 1;
    string value = 2;
}
```

### DIFF
--- a/oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigWatcherRegister.java
+++ b/oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigWatcherRegister.java
@@ -20,6 +20,7 @@ package org.apache.skywalking.oap.server.configuration.api;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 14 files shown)


wu-sheng:
> LGTM, the mechanism should work in other implementations of DC, right?

I am not sure for every implementation. Because they may have watching mechanism already, which doesn't require the uuid. This one is for reducing the network load in DCS implementation.

PR #4439, opened 2020-03-04
Support name column(s) in the storage for 3rd party integration, like Kibana. Default OFF

wu-sheng:
SkyWalking provides browser UI, CLI and GraphQL ways to support extensions. But some users may have the idea to query data 
directly from the storage. Such as in ElasticSearch case, Kibana is a great tool to do this.

In default, due to save memory/network and storage space, SkyWalking saves id(s) only in the entity and metadata saved in the
*_inventory entities only. But these tools usually don't support nested query, or don't work conveniently. In this special case,
SkyWalking provide a config to add all necessary name column(s) into the final metrics entities with ID as a trade-off.

Take a look at `core/default/activeExtraModelColumns` config in the `application.yaml`, and set it as `true` to open this feature.

This feature wouldn't provide any new feature to the native SkyWalking scenarios, just for the 3rd party integration.

### DIFF
--- a/docs/en/setup/backend/backend-setup.md
+++ b/docs/en/setup/backend/backend-setup.md
@@ -115,3 +115,15 @@ which format process is timezone related.
   
 In default, SkyWalking OAP backend choose the OS default timezone.
 If you want to override it, please follow Java and OS documents to do so.
+
+#### How to query the storage directly from 3rd party tool?
+SkyWalking provides browser UI, CLI and GraphQL ways to support extensions. But some users may have the idea to query data 
+directly from the storage. Such as in ElasticSearch case, Kibana is a great tool to do this.
+
+In default, due to save memory/network and storage space, SkyWalking saves id(s) only in the entity and metadata saved in the
+*_inventory entities only. But these tools usually don't support nested query, or don't work conveniently. In this special case,
+SkyWalking provide a config to add all necessary name column(s) into the final metrics entities with ID as a trade-off.
+
+Take a look at `core/default/activeExtraModelColumns` config in the `application.yaml`, and set it as `true` to open this feature.
+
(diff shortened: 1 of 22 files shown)


wu-sheng:
Here are the examples of index
image


image

wu-sheng:
If we are facing instance relationship, we could have 2 level names too.
image

I assume these are good enough for the Kibana case at least.

docs/en/setup/backend/backend-setup.md
```diff
@@ -115,3 +115,15 @@ which format process is timezone related.
   
 In default, SkyWalking OAP backend choose the OS default timezone.
 If you want to override it, please follow Java and OS documents to do so.
+
+#### How to query the storage directly from 3rd party tool?
+SkyWalking provides browser UI, CLI and GraphQL ways to support extensions. But some users may have the idea to query data 
+directly from the storage. Such as in ElasticSearch case, Kibana is a great tool to do this.
+
+In default, due to reduce memory, network and storage space usages, SkyWalking saves id(s) only in the entity and metadata saved in the
+*_inventory entities only. But these tools usually don't support nested query, or don't work conveniently. In this special case,
```
kezhenxu94: The italic font style is not terminated?

wu-sheng: No, it isn't. How to force it back in the markdown?

kezhenxu94: `*` and `_` should appear in pair IMO

wu-sheng: Fixed.

hanahmily:
LGTM. This's a handy feature for debugging purposes. I used suffering from switching querying between inventory and metric.

wu-sheng:
> LGTM. This's a handy feature for debugging purposes. I used suffering from switching querying between inventory and metric.

It could work as a debug feature if you like :)

Issue #4433, opened 2020-03-02
Add finagle plugin


huangyoje:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [x] Feature or performance improvement

___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS and JRE?

- Which company or project?

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.

___
### Requirement or improvement
- Add finagle plugin, version from 6.28.0 to 20.1.0

wu-sheng:
Contribution is welcome

huangyoje:
Hi, I have completed the main code and the automatic tests have passed.  I am very interested in contributing it to the community. how should i start?

arugal:
Hi @huangyoje, welcome to contribute. Please refer to the [contribute plugins guide](https://github.com/apache/skywalking/blob/master/docs/en/guides/Java-Plugin-Development-Guide.md#contribute-plugins-into-apache-skywalking-repository ).

PR #4441, opened 2020-03-04
Add finagle plugin

huangyoje:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
[https://github.com/apache/skywalking/issues/4433](https://github.com/apache/skywalking/issues/4433)
___
### Bug fix
- Bug description.

- How to fix?

___
- Add finagle plugin, version from 6.25.0 to 20.1.0


huangyoje:
> Please follow these four first
> 
> 1. Make codes well formatted, including comments of the codes
> 2. Add unit tests to your plugin codes.
> 3. Make CI and plugin integration test work
> 4. supported doc update.

OK!  I will follow these steps to improve.

dmsolr:
> Please follow these four first
> 
> 1. Make codes well formatted, including comments of the codes
> 2. Add unit tests to your plugin codes.
> 3. Make CI and plugin integration test work
> 4. supported doc update.

Additional:
1. add it to component-libraries.yaml 
2. add ICON to SkyWalking-UI repo

huangyoje:
> 2. add ICON to SkyWalking-UI repo
Hi, I think i need some help.  After checking the skywalking-ui directory,  I could't find where the icon should be added?  Also are there any requirements for icons, such as size.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/AbstractInterceptor.java
```diff
@@ -0,0 +1,76 @@
...
+
+    protected Object afterMethodImpl(EnhancedInstance objInst, Method method, Object[] allArguments,
+                                     Class<?>[] argumentsTypes, Object ret) throws Throwable {
+        return ret;
+    }
+
+    protected void handleMethodExceptionImpl(EnhancedInstance objInst, Method method, Object[] allArguments,
+                                             Class<?>[] argumentsTypes, Throwable t) {
+
+    }
```
aderm: These methods are overridden by methods in subsequent integrations. Recommend to modify these methods to abstract methods, so that the logic may be clearer.  Or they can be split into two classes?

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/ClientDestTracingFilterInterceptor.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Override
+    public Object afterMethodImpl(EnhancedInstance enhancedInstance, Method method, Object[] objects, Class<?>[] classes, Object o) throws Throwable {
+        return o;
+    }
+
+    @Override
+    public void handleMethodExceptionImpl(EnhancedInstance enhancedInstance, Method method, Object[] objects, Class<?>[] classes, Throwable throwable) {
+
```
aderm: Exception information to be added like `ContextManager.activeSpan().errorOccurred().log(t)`

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/ClientTracingFilterInterceptor.java
```diff
@@ -0,0 +1,97 @@
...
+        ((Future<?>) ret).addEventListener(new FutureEventListener<Object>() {
+            @Override
+            public void onSuccess(Object value) {
+                finagleSpan.asyncFinish();
+            }
+
+            @Override
+            public void onFailure(Throwable cause) {
+                finagleSpan.errorOccurred();
+//                finagleSpan.log(cause);
```
aderm: why comment this?

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/AbstractInterceptor.java
huangyoje: I will make these methods abstract.  

Because `ClientDestTracingFilterInterceptor` need to implement both `onConstructImpl ` and `beforeMethodImpl ` methods and java did not support mutiple inheritance,   so they can not be split into two class.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/ClientDestTracingFilterInterceptor.java
huangyoje: Sure, I will add it.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/ClientTracingFilterInterceptor.java
huangyoje:
```java
public ExitSpan log(Throwable t) {
        if (stackDepth == 1) {
            super.log(t);
        }
        return this;
    }
```
For ExitSpan, the log will only works when stackDepth=1.  I don't understand why, Does it ok to invoke log method here?

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/AbstractInterceptor.java
wu-sheng: Actually, in java, once you have set `AbstractInterceptor` as abstract, all these empty methods could be deleted directly :)

huangyoje:
> 2. You need to add your test into the CI pass by changing this file,  https://github.com/apache/skywalking/blob/master/.github/workflows/plugins-test.yaml . Please use the fast group.

After check the git history of the file, I find the last merged plugin is avro, does the fast group is Spring31x_ES5_Gateway_Avro_Struts25?  Please tolerate me with this kind of questions, I did not use github actions before.

wu-sheng:
> After check the git history of the file, I find the last merged plugin is avro, does the fast group is Spring31x_ES5_Gateway_Avro_Struts25? Please tolerate me with this kind of questions, I did not use github actions before.

Open the actions tab, https://github.com/apache/skywalking/pull/4441/checks. Open the PluginsTest tab on the left side, then you could check every sub group execution time. Or check the list above
image

That is easy.

huangyoje:
> Open the actions tab, https://github.com/apache/skywalking/pull/4441/checks. Open the PluginsTest tab on the left side, then you could check every sub group execution time. Or check the list above

I see. thanks!

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/AbstractInterceptor.java
huangyoje: These methods we discussed above are`onConstructImpl`,`beforeMethodImpl `, they are not inherited from interface.  The purpose of  `AbstractInterceptor ` is a adapter for `InstanceConstructorInterceptor , InstanceMethodsAroundInterceptor `.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/ClientTracingFilterInterceptor.java
wu-sheng: > For ExitSpan, the log will only works when stackDepth=1. I don't understand why, Does it ok to invoke log method here?

Why your plugin works as depth > 1? And yes, this is designed in that way, meaning, if there is a client-side RPC works, the nested one wouldn't be captured.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/CompatibilityChecker.java
```diff
@@ -0,0 +1,58 @@
...
+
+package org.apache.skywalking.apm.plugin.finagle;
+
+import org.apache.skywalking.apm.agent.core.logging.api.ILog;
+import org.apache.skywalking.apm.agent.core.logging.api.LogManager;
+
+import static org.apache.skywalking.apm.plugin.finagle.ContextHolderFactory.getLocalContextHolder;
+import static org.apache.skywalking.apm.plugin.finagle.ContextHolderFactory.getMarshalledContextHolder;
+
+public class CompatibilityChecker {
```
wu-sheng: Version compatibility feature is provided by a thing called `witness` class. If you provide that in the instrumentation definition, your plugin is activated only those class exists.

wu-sheng: The ThreadLocal has low performance in the agent, and cost more CPU resources, comparing to other ways. Your codes use this in high frequency which is a potential performance issue. We only accept one way to use thread local, is there is no way to communicate between the interceptor. From your class name, this is not for that case.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/ContextHolder.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.plugin.finagle;
+
+import javax.annotation.Nullable;
+
+/**
+ * @see com.twitter.finagle.context.Context
```
wu-sheng: This is not allowed, if it is copied from somewhere else, you are facing source code level LICENSE update. Also, as I commented above, I prefer to remove this.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/ContextHolderFactory.java
```diff
@@ -0,0 +1,308 @@
...
+import javax.annotation.Nullable;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+
+public class ContextHolderFactory {
```
wu-sheng: Read my comments about witness class, and remove this.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/AbstractInterceptor.java
wu-sheng: Let's leave this here. My review has been challenging deeper things than this code style  discussion. Please follow those first, then we could do the review again.

huangyoje:
> The comments of your codes are required. All the review are based on my guess, please don't let me do that.
> Comments are important for other contributors and reviewers.

You are right, sorry for that. I will add comments this weekend.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/ContextHolder.java
huangyoje: I have removed this.

huangyoje:
The comments have been added.  The implementation of this plugin is described in the `AbstractAbstractceptor` class, and other classes have added their own comments.  I hope these comments can be described clearly, if there is any ambiguity, please point out, i will modify again.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/CompatibilityChecker.java
huangyoje: > Version compatibility feature is provided by a thing called `witness` class. If you provide that in the instrumentation definition, your plugin is activated only those class exists.

The implementation detail of this plugin depend on a private class of Finagle Framework, we can not ensure that class will still exists in future versions. The mechanism of witness class can ensure this plugin can work in existing versions, it can not ensure this plugin will still work in future versions. So the purpose of the class  is to check whether this plugin is compatible with future versions, if it does't, the plugin just do nothing, avoiding unexpected runtime exceptions.
This has been added in comments.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/ContextHolderFactory.java
huangyoje: Comments has been added, please check `ContextHolder, ContextHolderFactory`, the class is necessary for the current implementation.

wu-sheng:
Have you followed no threadlocal and using the witness class? That should be followed.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/CompatibilityChecker.java
huangyoje: > The ThreadLocal has low performance in the agent, and cost more CPU resources, comparing to other ways. Your codes use this in high frequency which is a potential performance issue. We only accept one way to use thread local, is there is no way to communicate between the interceptor. From your class name, this is not for that case.

`ContextHolderFactory`?  There are no `ThreadLocal` used directly, however, it use `com.twitter.util.Local` which backed by `ThreadLocal`, but this is necessary, please check the comments.

wu-sheng: First, That thing liked I said, having IP issue. And same as I said, you shouldn't implement in this way.

I think I have said clear. Any question?

huangyoje:
> Have you followed no threadlocal and using the witness class? That should be followed.

https://github.com/apache/skywalking/pull/4441#discussion_r389245836

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/CompatibilityChecker.java
wu-sheng: The future versions are never being sure. So it is meaningless. That is why we have explicit version test CI process, and indicate them in support list doc. Only real test could verify that.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/CompatibilityChecker.java
```diff
@@ -0,0 +1,65 @@
...
+    static ILog LOGGER = LogManager.getLogger(CompatibilityChecker.class);
+
+    private static boolean COMPATIBILITY = false;
+
+    static {
+        try {
+            if (FinagleCtxs.RPC != null
+                    && FinagleCtxs.SW_SPAN != null
+                    && checkContextHolder()) {
+                COMPATIBILITY = true;
```
wu-sheng: You are using a Thread local to check compatibility, why you think it makes sense?

.github/workflows/plugins-test.yaml
```diff
@@ -271,6 +271,12 @@ jobs:
         run: bash test/plugin/run.sh undertow-scenario
       - name: Run jedis 2.4.0-2.9.0 (18)
         run: bash test/plugin/run.sh jedis-scenario
+      - name: Run finagle 6.25.0-6.43.0
+        run: bash test/plugin/run.sh finagle-6.25.x-scenario
+      - name: Run finagle 6.44.0-7.1.0
+        run: bash test/plugin/run.sh finagle-6.44.x-scenario
+      - name: Run finagle 17.10.0-20.1.0
+        run: bash test/plugin/run.sh finagle-17.10.x-scenario
```
wu-sheng: Are these versions using different coding API for framework end users? Even from 6.33 to 6.34? Could you explain?

wu-sheng: I am feeling you are not using the witness mechanism properly.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/CompatibilityChecker.java
huangyoje: Sorry, I misunderstanding your point, I will remove this class.

.github/workflows/plugins-test.yaml
wu-sheng: I just look your test cases, there are 3 folders but the codes are sharing or same. You should merge them into one case with version from 6.25-20.1. Because the only reason we separate the case, because the library APIs changed from version to version. But for finagle, I think it isn't. Am I right?

huangyoje: > Are these versions using different coding API for framework end users? Even from 6.33 to 6.34? Could you explain?

Yes.  To use finagle, at least needs two jars, `finagle-core` and `finagle-{protocol}(eg: finagle-thrift, finagle-http)`,  I choose `finagle-thrift` in tests.  The plugin implementation is only based on `finagle-core`, but the difference of coding API comes from `finagle-thrift`.  

Specifically, to use `finagle-thrift`, `scrooge-maven-plugin` is also needed to generate code from thrift file,  and the versions of  `finagle-thrift` and `scrooge-maven-plugin` should be compatible.   In these tests, the generatd code of  `finagle-6.25.x-scenario` does not compatible with finagle-thrift 6.44.  and the generated code of `finagle-6.44.x-scenario` indeed compatible with finagle-thrift 6.43, but does not  compatible with all versions from 6.25 to 6.43.  I put these versions that both tests compatible into `finagle-6.25.x-scenario`.  And from version 17.10, these three jars(`finagle-core`,`finagle-thrift`,`scrooge-maven-plugin`) just has the same version.

wu-sheng: OK. Let's leave the cases like this for now. Please continue on other changes.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/ClientTracingFilterInterceptor.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public void beforeMethodImpl(EnhancedInstance enhancedInstance, Method method, Object[] objects, Class<?>[] classes,
+                                 MethodInterceptResult methodInterceptResult) throws Throwable {
+        ContextCarrier contextCarrier = new ContextCarrier();
+        /*
+         * At this time, we can't know the operation name and peer address, so we just use placeholders here, the
+         * operation name will be filled by {@link AnnotationInterceptor$Rpc} and the peer address will be filled by
+         * {@link ClientDestTracingFilterInterceptor} later.
+         */
+        AbstractSpan finagleSpan = ContextManager.createExitSpan("pending", contextCarrier, "");
```
wu-sheng: Why don't you create the exit span in the `ClientDestTracingFilterInterceptor` directly? ContextCarrier require the peer information, otherwise the topology will break.

huangyoje: Because `ClientDestTracingFilter ` may be executed in another thread, not the thread that initiated the rpc request.   It is hard to know when and where the thread change occurs.

huangyoje:
> I think you update the query-protocol submodule by a mistake.

This has been corrected.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/ClientTracingFilterInterceptor.java
huangyoje: The peer infomation is added to ContextCarrier in `ClientDestTracingFilterInterceptor` by `ContextCarrierHelper`.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/ClientTracingFilterInterceptor.java
```diff
@@ -0,0 +1,107 @@
...
+        finagleSpan.setComponent(FINAGLE);
+        SpanLayer.asRPCFramework(finagleSpan);
+
+        ContextHolder marshlledContextHolder = getMarshalledContextHolder();
+        marshlledContextHolder.let(SWContextCarrier$.MODULE$, SWContextCarrier.of(contextCarrier));
+
+        ContextHolder localContextHolder = getLocalContextHolder();
+        localContextHolder.let(SW_SPAN, finagleSpan);
+
+        enhancedInstance.setSkyWalkingDynamicField(new CacheObjects(marshlledContextHolder, localContextHolder));
```
wu-sheng: Is the `enhancedInstance` used in only one thread only? If there is a race condition, you can't use the dynamic field.

wu-sheng: I am confused here, if the `dynamic field` is safe enough, why you need the `ContextHolder`?

wu-sheng: `dynamic field` could set any customized class, from my reading, it should be used to replace the `ContextHolder`.

Anyway, I think I need your further explanation about this.

wu-sheng:
As #4458 merged, the expected YAML file has fields been renamed. Take a look the PR and updated plugin test doc.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/AnnotationInterceptor.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        @Override
+        protected Object afterMethodImpl(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) throws Throwable {
+            return ret;
+        }
+
+        @Override
+        protected void handleMethodExceptionImpl(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Throwable t) {
+            ContextManager.activeSpan().errorOccurred().log(t);
```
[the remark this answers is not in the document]
huangyoje: I will remove this.  The `AnnotationInterceptor` only needs to implement `InstanceConstructorInterceptor`.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/AnnotationInterceptor.java
```diff
@@ -0,0 +1,87 @@
...
+     * to exitspan.
+     */
+    public static class Rpc extends Abstract {
+
+        @Override
+        protected void onConstruct(EnhancedInstance enhancedInstance, Object[] objects, AbstractSpan span) {
+            if (objects != null && objects.length == 1) {
+                String rpc = (String) objects[0];
+                if (span == null) {
+                    // in case the exitspan is created later
```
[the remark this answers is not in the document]
huangyoje: Iniitally, here is used to set op to server entryspan(sorry for the wrong comments),  and finally, I passed op from client to server, so it is no longer needed here. The codes and comments has been modified as below:
```java
if (span != null) {
    /*
     * The Rpc Annotation is created both in client side and server side, in server side, this
     * annotation is created only in finagle versions below 17.12.0.
     *
     * If the span is not null, which means we are in the client side, we just set op to the exitspan.
     *
     * If the span is null, which means we are in the server side with finagle version below 17.12.0.
     * In server side, we don't need this annotation, because we can get op from Contexts.broadcast
     * which comes from client.
     */
    span.setOperationName(rpc);
}
```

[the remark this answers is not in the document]
wu-sheng: Why do you plan to propagate the op name in the wire? Entry span should focus on the server side logic only. The only thing will propagate in your RPC header, is the official ContextCarrier.

Like my other comments say, I have concern about you are hijacking the ContextCarrier. If we change it some day, even it is rarely to see, it is hard to remember the Finagle plugin has this operation.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/scala/org/apache/skywalking/apm/plugin/finagle/SWContextCarrier.scala
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package org.apache.skywalking.apm.plugin.finagle
+
+import com.twitter.finagle.context.Contexts
+import com.twitter.io.Buf
+import com.twitter.util.{Return, Try}
+import org.apache.skywalking.apm.agent.core.context.ContextCarrier
+
+class SWContextCarrier(val carrier: ContextCarrier) {
```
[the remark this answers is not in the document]
huangyoje: `SWContextCarrier `  implement `MarshalledContext.Key`, which is a scala inner class,  I don't find a right way to do this in java.
```java 
class Impl1 extends MarshalledContext.Key<String> {
        public Impl1() {
            Contexts.broadcast().super("impl1");
        }
}
```
javac throws a compile error for above codes. However, when i fix the compile error, there will be a runtime error.   Java and scala's inner classes are slightly different[https://stackoverflow.com/a/39692316/7645128](https://stackoverflow.com/a/39692316/7645128), may be the error here comes from difference of javac and scalac.

I am still looking for a right way to do this in java,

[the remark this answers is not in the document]
wu-sheng: OK, if this is required in the original codes, we could keep it. This is not the issue.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/ClientTracingFilterInterceptor.java
huangyoje: > Is the enhancedInstance used in only one thread only?
The  `enhancedInstance ` is used in mulit threads, I will remove usage of it.

> I am confused here, if the dynamic field is safe enough, why you need the ContextHolder?
`ContextHolder` is used to put `finagleSpan`,`SWContextCarrier` to Context, thus these two can be accessed by `ClientDestTracingFilterInterceptor`, even if `ClientDestTracingFilter` is executed in another thread.

huangyoje:
> As #4458 merged, the expected YAML file has fields been renamed. Take a look the PR and updated plugin test doc.

OK.

wu-sheng:
I noticed, this PR fails every time on AppVeyor. Strange.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/AnnotationInterceptor.java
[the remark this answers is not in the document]
huangyoje: > Why do you plan to propagate the op name in the wire? 

For finagle version 17.12.0 and above,  we can not get op name in the server side directly.  If we add `finagle-{protocol}(eg: finagle-thrift)` into this plugin, then we could deserialize the bytes of the request paramenter and get op name, but this will introduce more complexity becase we have to be compatible with various protocols.

> Like my other comments say, I have concern about you are hijacking the ContextCarrier. If we change it some day, even it is rarely to see, it is hard to remember the Finagle plugin has this operation.

So far, I can't think of a better way to do this.  When we create span and contextCarrier, we don't know the remote address and op name,  but it is difficult to  propagate trace information across finagle's inner threads and then create span untill we know remote address and op name.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/ClientTracingFilterInterceptor.java
huangyoje: So far it is necessary.
https://github.com/apache/skywalking/pull/4441#discussion_r389333225

huangyoje:
> I noticed, this PR fails every time on AppVeyor. Strange.

I will check this issue.

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/AnnotationInterceptor.java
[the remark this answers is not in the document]
wu-sheng: Take a look at #4462

apm-sniffer/apm-sdk-plugin/finagle-6.25.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/finagle/ContextCarrierHelper.java
```diff
@@ -0,0 +1,65 @@
...
+
+    static void tryInjectContext() {
+        String operationName = getOperationName();
+        if (operationName == null) {
+            return;
+        }
+        String peer = getPeerHost();
+        if (peer == null) {
+            return;
+    }
```
wu-sheng: This format seems wrong.

huangyoje: Fixed.

huangyoje:
> Most concerns have been eased. Just `CodecUtils` and `ContextHolderFactory` may leak comments.
> 
> To be honest, I can't get too much details of this plugin, as I am not familiar with the Finagle :). All these requirements are making another contributor or further bug fix one could understand your logic. Feel free to add a `.md` in your plugin rood folder to do better explanation.

I will take your suggesiton, and to summarize what i will do next:
1. Fix `ContextHolderFactory`
    For the current implentation of `ContextHolderFactory`,   if server calls another rpc, for example  A ->B(serverB) ->C(serverC),  there will have context pollution in serverB. I will test this and fix.
2. Fix the failure of appveyor
     Looks likedto it caused by scala codes.  First, I will try to replace these codes with java, if not, I will try to find the compatible version for related components.
3. Improve comments.
4. Add a `implementation.md` in root folder

Hope to finish above in next week!

wu-sheng:
Thanks for the update. If all these provided in next week, we could add this as a part of 7.0.0 release.

> Looks like it caused by scala codes. First, I will try to replace these codes with java, if not, I will try to find the compatible version for related components.

`appveyor.yml` is controlling the CI process in the `appveyor`. We used to use appveyor for Windows compiling, but this has been rechecked in GitHub action too, here is the control file, https://github.com/apache/skywalking/blob/master/.github/workflows/ci-it.yaml#L53-L72.
I don't know what is the difference.

wu-sheng:
@aderm @arugal @dmsolr @kezhenxu94 Any of you could recheck?

wu-sheng:
Merging. Thank you for your contributions.

huangyoje:
image

I'm doing additional tests for scenario:  A -> ServerB -> ServerC and meet above exception.  I haven't had time to find out why, it better to revert this pr first.

wu-sheng:
@huangyoje I think this is my fault. I was not expecting the tracing context could be finished in the main thread when we do lazy injection. I will submit a fix PR soon.

huangyoje:
@wu-sheng OK.

wu-sheng:
@huangyoje Ping you at the PR.

huangyoje:
> Merging. Thank you for your contributions.

🍺🍺

PR #4442, opened 2020-03-04
Forward compatibility for DCS

wu-sheng:
In the old implementation of DCS, the old one will not response UUID, which could make the update not work. 

This make the forward compatibility for DCS, enhance #4438


dmsolr:
Emma, what does DCS mean?

wu-sheng:
@dmsolr read #4438? DCS = Dynamic Configuration Service. A gRPC based integration way.

dmsolr:
> @dmsolr read #4438? DCS = Dynamic Configuration Service. A gRPC based integration way.

Got it.

PR #4451, opened 2020-03-06
Fix a misguide of release doc.

wu-sheng:
Use the official key file.


wu-sheng:
I am not sure which is better :) `download.*` looks like a link for outside, such as download page. I am quite sure about this.

Issue #4396, opened 2020-02-22
Support duplicated instance-network address mapping


wu-sheng:
This is a potential bug, when I begin to write Nginx Lua plugin. The service instance inventory holds the mapping to network address, but if the agent reports duplicated peer(such as don't have real peer address like lua plugin) for different instances, which could cause duplicated instance-network mapping.

@arugal please recheck the span analysis and query, if there is bug, I think it is, please fix it ASAP.

PR #4454, opened 2020-03-06
Ignore no upstream real address agent

arugal:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/4396
___
### Bug fix
- Bug description.

- How to fix?

___


wu-sheng:
You seems breaking the e2e? Please recheck.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/cache/ServiceInstanceInventoryCache.java
```diff
@@ -119,4 +127,22 @@ public int getServiceInstanceId(int serviceId, int addressId) {
...
+                    if (key.equals(ServiceInstanceInventory.PropertyUtil.LANGUAGE)) {
+                        language = properties.get(key).getAsString().toLowerCase();
+                        languageCache.put(serviceInstanceId, language);
+                        return language;
+                    }
+                }
+            }
+        }
+        return Const.EMPTY_STRING;
+    }
```
kezhenxu94: Consider using "fast return", instead of so many nested blocks

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/NoUpstreamRealAddressAgentConfig.java
```diff
@@ -0,0 +1,119 @@
...
+        try {
+            final Reader reader = ResourceUtils.read(file);
+            return new Yaml().loadAs(reader, Languages.class);
+        } catch (FileNotFoundException e) {
+            log.error("Cannot load languages from: {}", file, e);
+        }
+        return Languages.EMPTY;
+    }
+
+    private Languages parseGatewayFromYml(final String ymlContent) {
```
kezhenxu94: Gateway?

apm-dist/src/main/assembly/binary.xml
```diff
@@ -51,6 +51,7 @@
                 <include>official_analysis.oal</include>
                 <include>gateways.yml</include>
                 <include>service-apdex-threshold.yml</include>
+                <include>no_upstream_real_address_agent.yml</include>
```
wu-sheng: Don't use a new file. This is a very simple config, please add a field to tracr receiver module provider. 
Only If the content is complex and usually changed, we need a separated file. This fix is not that case.

This config is nearly static, nearly no one should change.

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/TraceServiceModuleConfig.java
```diff
@@ -54,6 +54,9 @@
     @Setter
     @Getter
     private UninstrumentedGatewaysConfig uninstrumentedGatewaysConfig;
+    @Setter
+    @Getter
+    private NoUpstreamRealAddressAgentConfig noUpstreamRealAddressAgentConfig;
```
wu-sheng: We support list here, take a look at downsampling config. And you should put LUA as the default element.

dist-material/application.yml
```diff
@@ -180,6 +180,8 @@ receiver-trace:
     bufferFileCleanWhenRestart: ${SW_RECEIVER_BUFFER_FILE_CLEAN_WHEN_RESTART:false}
     sampleRate: ${SW_TRACE_SAMPLE_RATE:10000} # The sample rate precision is 1/10000. 10000 means 100% sample in default.
     slowDBAccessThreshold: ${SW_SLOW_DB_THRESHOLD:default:200,mongodb:100} # The slow database access thresholds. Unit ms.
+    noUpstreamRealAddressAgents: # The no upstream real address agents, service instance mapping and service instance client side relation are ignored.
```
wu-sheng: I think we don't need to open this to the end user. WDYT?

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/TraceServiceModuleConfig.java
```diff
@@ -41,7 +44,13 @@
...
     @Getter
     private int sampleRate = 10000;
-
+    /**
+     * Some of the agent can not have the upstream real network address, such as https://github.com/apache/skywalking-nginx-lua.
+     * service instance mapping and service instance client side relation are ignored.
+     */
+    @Setter
+    @Getter
+    private List<String> noUpstreamRealAddressAgents = Collections.emptyList();
```
wu-sheng: This should be initialized to include LUA. As a native agent, we know, LUA agent can't support in this value.

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/parser/listener/endpoint/MultiScopesSpanListener.java
```diff
@@ -306,7 +309,17 @@ public void build() {
 
             exitSourceBuilder.setTimeBucket(minuteTimeBucket);
             sourceReceiver.receive(exitSourceBuilder.toServiceRelation());
-            sourceReceiver.receive(exitSourceBuilder.toServiceInstanceRelation());
+            String sourceLanguage = instanceInventoryCache.getServiceInstanceLanguage(exitSourceBuilder.getSourceServiceInstanceId());
+            if (config.getNoUpstreamRealAddressAgents().contains(sourceLanguage)) {
+                /*
+                 * Some of the agent can not have the upstream real network address, such as https://github.com/apache/skywalking-nginx-lua.
+                 */
+                if (log.isDebugEnabled()) {
```
wu-sheng: I think we don't need log for this.

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/parser/listener/service/ServiceInstanceMappingSpanListener.java
```diff
@@ -79,6 +81,17 @@ public void parseEntry(SpanDecorator spanDecorator, SegmentCoreInfo segmentCoreI
         if (!spanDecorator.getSpanLayer().equals(SpanLayer.MQ)) {
             if (spanDecorator.getRefsCount() > 0) {
                 for (int i = 0; i < spanDecorator.getRefsCount(); i++) {
+                    ReferenceDecorator referenceDecorator = spanDecorator.getRefs(i);
+                    String parentLanguage = serviceInstanceInventoryCache.getServiceInstanceLanguage(referenceDecorator.getParentServiceInstanceId());
+                    if (config.getNoUpstreamRealAddressAgents().contains(parentLanguage)) {
+                        /*
+                         * Some of the agent can not have the upstream real network address, such as https://github.com/apache/skywalking-nginx-lua.
+                         */
+                        if (log.isDebugEnabled()) {
```
wu-sheng: We don't need the log too.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/cache/ServiceInstanceInventoryCache.java
arugal: Done, thanks for your review :laughing:

dist-material/application.yml
arugal: Agree.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/cache/ServiceInstanceInventoryCache.java
```diff
@@ -119,4 +125,24 @@ public int getServiceInstanceId(int serviceId, int addressId) {
         }
         return serviceInstanceId;
     }
+
+    public String getServiceInstanceLanguage(int serviceInstanceId) {
+        String language = languageCache.getIfPresent(serviceInstanceId);
```
wu-sheng: I think comparing to this implementation, I would say, adding a `no-@column` field in the `ServiceInstanceInventory` to cache the status should be better in performance and understanding.

And you should avoid check the language repeatedly if this instance hasn't language properties. Such as in the new implementation, you should separate the value from `null` and `Const#UNKNOWN`.

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/TraceServiceModuleConfig.java
arugal: done

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/TraceServiceModuleConfig.java
arugal: done

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/cache/ServiceInstanceInventoryCache.java
arugal: done

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/entity/LanguageTrans.java
```diff
@@ -33,6 +33,10 @@ public Language value(String language) {
                 return Language.PYTHON;
             case "ruby":
                 return Language.RUBY;
+            case "go":
+                return Language.GO;
+            case "Lua":
```
wu-sheng: I didn't notice this should be lower case. How about to use `toLowerCase` in the switch?

arugal: done.

PR #4462, opened 2020-03-08
[Agent Core] Support lazy ContextCarrier injection and lazy peer id setting 

wu-sheng:
This PR's idea was beginning when I am reviewing @huangyoje 's PR #4441. He did hijacking the ContextCarrier, which is not supposed to happen. After the discussion, I think that plugin has to do lazy injection due to the framework limitation.

Here is the proposal, right now, we officially support
1. Create exit span with `null` as peer, if the user wouldn't do injection.
1. Exit span could do `span#inject` directly, by leverage the span's internal tracing context ref.

FYI @apache/skywalking-committers This would be the first API change(adding) after the async span feature. Please review.

@huangyoje Please cross check this PR change, and give me the feedback whether this PR could remove your hijacking requirements.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AbstractTracerContext.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AbstractTracerContext.java
@@ -32,6 +32,15 @@ public interface AbstractTracerContext {
     void inject(ContextCarrier carrier);
 
     /**
+     * Prepare for the cross-process propagation based on the given exit span. The given exit span should belong to the
+     * current context. How to initialize the carrier, depends on the implementation.
+     *
+     * @param carrier  to carry the context for crossing process.
+     * @param exitSpan to represent the scope of current injection.
+     */
+    void _inject(AbstractSpan exitSpan, ContextCarrier carrier);
+
+    /**
      * Build the reference between this segment and a cross-process segment. How to build, depends on the
      * implementation.
      *
@@ -82,7 +91,8 @@ public interface AbstractTracerContext {
      * Create an exit span
      *
      * @param operationName most likely a service name of remote
-     * @param remotePeer    the network id(ip:port, hostname:port or ip1:port1,ip2,port, etc.)
+     * @param remotePeer    the network id(ip:port, hostname:port or ip1:port1,ip2,port, etc.). Remote peer could be set
+     *                      later, but must be before injecting.
      * @return the span represent an exit point of this segment.
      */
     AbstractSpan createExitSpan(String operationName, String remotePeer);
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AbstractTracerContext.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AbstractTracerContext.java
@@ -19,7 +19,6 @@
 package org.apache.skywalking.apm.agent.core.context;
 
 import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
-import org.apache.skywalking.apm.agent.core.context.trace.ExitTypeSpan;
 
 /**
  * The <code>AbstractTracerContext</code> represents the tracer context manager.
@@ -31,18 +30,6 @@ public interface AbstractTracerContext {
      * @param carrier to carry the context for crossing process.
      */
     void inject(ContextCarrier carrier);
-
-    /**
-     * Prepare for the cross-process propagation based on the given exit span. The given exit span should belong to the
-     * current context.This method wouldn't be opened in {@link ContextManager} like {@link #inject(ContextCarrier)}, it
-     * is only supported to be called inside the {@link ExitTypeSpan#inject(ContextCarrier)}
-     *
-     * How to initialize the carrier, depends on the implementation.
-     *
-     * @param carrier  to carry the context for crossing process.
-     * @param exitSpan to represent the scope of current injection.
-     */
-    void inject(AbstractSpan exitSpan, ContextCarrier carrier);
 
     /**
      * Build the reference between this segment and a cross-process segment. How to build, depends on the
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AbstractTracerContext.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AbstractTracerContext.java
@@ -19,6 +19,7 @@
 package org.apache.skywalking.apm.agent.core.context;
 
 import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
+import org.apache.skywalking.apm.agent.core.context.trace.ExitTypeSpan;
 
 /**
  * The <code>AbstractTracerContext</code> represents the tracer context manager.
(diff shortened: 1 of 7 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AbstractTracerContext.java
```diff
@@ -31,6 +31,15 @@
...
     void inject(ContextCarrier carrier);
 
+    /**
+     * Prepare for the cross-process propagation based on the given exit span. The given exit span should belong to the
+     * current context. How to initialize the carrier, depends on the implementation.
+     *
+     * @param carrier  to carry the context for crossing process.
+     * @param exitSpan to represent the scope of current injection.
+     */
+    void _inject(AbstractSpan exitSpan, ContextCarrier carrier);
```
kezhenxu94: why add the leading `_`

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/ExitTypeSpan.java
```diff
@@ -18,8 +18,15 @@
...
+
+/**
+ * The exit span has some additional behaviours
+ */
+public interface ExitTypeSpan {
     int getPeerId();
 
     String getPeer();
+
+    void inject(ContextCarrier carrier);
```
kezhenxu94: Consider returning `this` for chaining

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AbstractTracerContext.java
wu-sheng: I was wanting people think this could be used as an open API. That is the only repo. Do you want to keep the same method name? `inject`?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/ExitTypeSpan.java
wu-sheng: After the tests pass, will do.

wu-sheng: From the API lever, I could only return `ExitTypeSpan`, is that useful?

kezhenxu94: > From the API level, I could only return `ExitTypeSpan`, is that useful?

Yes, from the interface you have to return `ExitTypeSpan`, but you can return `ExitSpan` in its implementations, and chaining with the more specific types

image

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AbstractTracerContext.java
kezhenxu94: > I was wanting people think this could be used as an open API. That is the only repo. Do you want to keep the same method name? `inject`?

All methods in an `interface` are **OPEN API** IMO, there's no conventions that methods start with `_` is private or open, on the contrary, many languages use the leading `_` to indicate that the method is `private`

wu-sheng: From my idea, the interface is just open at the codes level, in here, to balance the noop tracing context and real tracing context. Anyway, I could change it to `inject` to more clear comment.

wu-sheng: Done.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/ExitTypeSpan.java
wu-sheng: Done.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AbstractTracerContext.java
kezhenxu94: > I was wanting people think this could be used as an open API. 

Well, I think you just expressed the opposite meanings ("want people **NOT** to used as an open API"), which makes me more confused why add this method here, instead of in `TracingContext` (package private), the `ExitSpan` holds an instance of `TracingContext` instead of `AbstractTracerContext`, it should illuminate both our concerns

dmsolr:
Wow, this PR is friendly to write any plugin on the async scenario. Before this, we usually need to waste much time to find the right point to enhance.

wu-sheng:
> Wow, this PR is friendly to write any plugin on the async scenario. Before this, we usually need to waste much time to find the right point to enhance.

@dmsolr Yes, that was the idea. But I would say just using this only if necessary.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AbstractTracerContext.java
wu-sheng: Good point. Let me do that.

wu-sheng: Fixed, should be better now.

huangyoje:
I just tried with this PR, and in my scenario, I encountered a problem:
1. I created `ExitSpan` without valid op name and peer host
2. the op name and peer host will be set to `ExitSpan` later, but the ordering is uncertain
3. but when can I invoke the `ExitSpan.inject(ContextCarrier)`?
    
Does the below way is OK?
```java
void tryInject(ExitSpan exitSpan) {
      if (isValid(exitSpan)) {
           return null;   
      }
      ContextCarrier carrier = new ContextCarrier();
      exitSpan.inject(carrier);
      return carrier;
}

boolean isValid(ExitSpan exitSpan) {
       return exitSpan.getOperatioName() != null && exitSpan.getPeer() != null;
}
```

wu-sheng:
> the op name and peer host will be set to ExitSpan later, but the ordering is uncertain

This is fine. just be sure these are set before the `#inject`. If you need to do before injecting check, then you could. Just seems strange, but at least not hijacking the core APIs. But please keep the comments clear about why you are doing this.

huangyoje:
> This is fine. just be sure these are set before the `#inject`. If you need to do before injecting check, then you could. Just seems strange, but at least not hijacking the core APIs. But please keep the comments clear about why you are doing this.

OK, I am testing with this PR, I will feedback the result as soon as the test is over.

wu-sheng:
> return exitSpan.getOperatioName() != null && exitSpan.getPeer() != null;

I prefer you use the plugin internal variables rather than using `exitSpan#get*`. @huangyoje

huangyoje:
> I prefer you use the plugin internal variables rather than using `exitSpan#get*`. @huangyoje
> Because the peer name could be exchanged to peer id. So avoid of using that, use something you could control.

OK, I will try to avoid using `exitSpan#get*`.

wu-sheng:
Waiting for @huangyoje's local test confirmation.

huangyoje:
> Waiting for @huangyoje's local test confirmation.

My loacl tests have all passed!  And Thanks for this PR.

wu-sheng:
Merged. Update your PR please.

huangyoje:
> Merged. Update your PR please.

OK.

PR #4470, opened 2020-03-09
Enable OAP gRPC SSL transportation

hanahmily:
Porting to OpenSSL to enable SSL transportation. The server private
key is in the format of PCKS#8, certificates is x509 though.


docs/en/setup/backend/grpc-ssl.md
```diff
@@ -0,0 +1,33 @@
+#Support gRPC SSL transportation for OAP server
```
kezhenxu94:
```suggestion
# Support gRPC SSL transportation for OAP server
```

docs/en/setup/backend/grpc-ssl.md
```diff
@@ -0,0 +1,33 @@
...
+
+For OAP communication we are currently using gRPC, a multi-platform RPC framework that uses protocol buffers for
+message serialization. The nice part about gRPC is that it promotes the use of SSL/TLS to authenticate and encrypt
+exchanges. Now OAP support to enable SSL transportation for gRPC receivers.
```
kezhenxu94:
```suggestion
exchanges. Now OAP supports to enable SSL transportation for gRPC receivers.
```

docs/en/setup/backend/grpc-ssl.md
kezhenxu94: without space, the header style is not correctly rendered

wu-sheng:
Please add the ssl into auth test case, single group, e2e

wu-sheng:
@hanahmily This is the agent side TLS doc for you, https://github.com/apache/skywalking/blob/master/docs/en/setup/service-agent/java-agent/TLS.md

hanahmily:
> 1. Only `ReceiverGRPCHandlerRegister` should support only server-side receiver. You could read the codes, only it has the `#addFilter` for `AuthenticationInterceptor`.

Don't get your point.`ReceiverGRPCHandlerRegister` intends to add `interceptor` or `service` to `NettyServerBuilder` by invoking its `addService` method. But SSL should invoke `NettyServerBuilder#sslContext`. I don't find any link between them. Do I miss something? feel free to correct me.

> 2. If we talk about the core module, you have to implement the client side TLS inside `GRPCRemoteClient`, in oder to make sure it matches the core configuration. And this TLS should be used for internal only.

make sense. I should enable client ssl at the same time. There's also a scenario for `sharing-server`, if we want to enable external SSL and disable internal SSL, client ssl needs to read config from `sharing-server` instead of `core`, right?

wu-sheng:
SSL for outside only could active when sharing server activated. Am I right? 
I don't think we should make the internal SSL CA as same as OAP internal. WDYT?

docs/en/setup/backend/grpc-ssl.md
```diff
@@ -0,0 +1,32 @@
...
+## Config OAP server 
+
+You can enable gRPC SSL by add following lines to `application.yml/core/default`.
+```json
+gRPCSslEnabled: true
+gRPCSslKeyPath: /path/to/server-key.pem
+gRPCSslCertChainPath: /path/to/server.crt
+```
+
+If you port to java agent, refer to [TLS.md](../service-agent/java-agent/TLS.md) to config java agent to enable TLS.
```
wu-sheng: This linked file should be updated too, to keep consistently

hanahmily:
> SSL for outside only could active when sharing server activated. Am I right?

right

> I don't think we should make the internal SSL CA as same as OAP internal. WDYT?

No. SSL becomes more widely accepted. We should support it both internally and externally.

hanahmily:
@wu-sheng e2e test case is added and passed. now I think we can merge it.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/remote/client/RemoteClientManager.java
```diff
@@ -55,21 +59,29 @@
...
      * Initial the manager for all remote communication clients.
-     *
-     * @param moduleDefineHolder for looking up other modules
+     *  @param moduleDefineHolder for looking up other modules
      * @param remoteTimeout      for cluster internal communication, in second unit.
+     * @param grpcSslEnabled true: enable SSL between clusters, false: plain gRPC between the cluster.
+     * @param grpcSslTrustCAPath trust certificate authorities file path.
      */
-    public RemoteClientManager(ModuleDefineHolder moduleDefineHolder, int remoteTimeout) {
+    public RemoteClientManager(ModuleDefineHolder moduleDefineHolder,
```
wu-sheng: Could you make the core style consistently. `GRPCRemoteClient` client use two constructors for ssl ON/OFF, here using a boolean and empty/null string.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/remote/client/GRPCRemoteClient.java
```diff
@@ -99,7 +110,7 @@ GRPCClient getClient() {
         if (Objects.isNull(client)) {
             synchronized (GRPCRemoteClient.class) {
                 if (Objects.isNull(client)) {
-                    this.client = new GRPCClient(address.getHost(), address.getPort());
+                    this.client = new GRPCClient(address.getHost(), address.getPort(), sslContext);
```
wu-sheng: A similar code style inconsistent here, using `sslContext==null` represents no SSL.

oap-server/server-receiver-plugin/skywalking-sharing-server-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/sharing/server/SharingServerModuleProvider.java
```diff
@@ -80,6 +81,15 @@ public void prepare() {
         if (config.getGRPCPort() != 0) {
             grpcServer = new GRPCServer(Strings.isBlank(config.getGRPCHost()) ? "0.0.0.0" : config.getGRPCHost(), config
```
wu-sheng: I think you miss the deletion of this?

test/e2e/e2e-ssl/e2e-ssl.iml
```diff
@@ -0,0 +1,120 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
wu-sheng: I think this should not be uploaded? Why git ignore is not working for this?

test/e2e/e2e-ssl/src/docker/certs/server-key.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
wu-sheng: Are these cert files alway valid? No expired time?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/remote/client/RemoteClientManager.java
hanahmily: done. keep all parameter list the same way.

oap-server/server-receiver-plugin/skywalking-sharing-server-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/sharing/server/SharingServerModuleProvider.java
hanahmily: removed it.

test/e2e/e2e-ssl/e2e-ssl.iml
hanahmily: deleted it.

test/e2e/e2e-ssl/src/docker/certs/server-key.pem
hanahmily: 2 years. Is it enough?

wu-sheng: My point is, this is for test. Could we set it forever? Or let's say 100 years?

wu-sheng:
You need to fix CI

test/e2e/e2e-ssl/src/docker/certs/server-key.pem
hanahmily: Now we get the  100 years/ century certificates.

PR #4471, opened 2020-03-09
[BugFix] Docker so11y cannot be disabled

kezhenxu94:
### Motivation:

This patch fixes the bug that there's no chance to disable the so11y in docker container.

### Modifications:

- Remove the default so11y section, and only add it when the environment variable exists.

### Result:

- Users can disable so11y by leaving the `SW_TELEMETRY` empty or set it to `none`

### DIFF
--- a/docker/oap-es7/docker-entrypoint.sh
+++ b/docker/oap-es7/docker-entrypoint.sh
@@ -384,8 +384,6 @@ receiver-jvm:
   default:
 receiver-clr:
   default:
-receiver-so11y:
-  default:
 service-mesh:
   default:
     bufferPath: \${SW_SERVICE_MESH_BUFFER_PATH:../mesh-buffer/}  # Path to trace buffer files, suggest to use absolute path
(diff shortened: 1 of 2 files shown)


docker/oap-es7/docker-entrypoint.sh
```diff
@@ -384,8 +384,6 @@ receiver-jvm:
   default:
 receiver-clr:
   default:
-receiver-so11y:
```
hanahmily: should move it to func `generateTelemetrySo11y `

kezhenxu94: It's already there, isn't it?

https://github.com/apache/skywalking/blob/5eba81e1409ca11204812bd456d4f4af206b76d2/docker/oap/docker-entrypoint.sh#L450-L454

hanahmily: sorry, didn't notice it

PR #4475, opened 2020-03-10
Add file change detection mechanism

wu-sheng:
Due to some strict management rules inside the enterprise, some companies use tool such as Vault to manage username and password of the storage. These tools would change the content by injecting the new username and password into some certain file, through using some tech like sidecar or agent.

This PR makes the first change, it provides the basic file content(or we should say modified flag) changed detection mechanism. Once it is changed, the notification will be triggered.

The next step will be activating this mechanism for ES storage. Once username and password changed, the ES client should trigger reconnection.


oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/FileChangeMonitor.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * File change monitor is a disk file watcher. It keeps to watch the file `last modified timestamp`, after it changed,
```
hanahmily: We should consider another scenario if the content of file doesn't change the monitor should not notify the related component.
A possible approach is to compare the binary data of them.

wu-sheng: > We should consider another scenario if the content of file doesn't change the monitor should not notify the related component.

I expect this is the responsibility of notifier implementation. I just don't want to read the file twice.

hanahmily: I prefer to leave it to the monitor since it extracts the input-stream from the target file. And the comparison is a common requirement for all of the notifiers, Do you have some scenario needs to get notification of modification without changing the content? if that exists, it's sane for it.

wu-sheng: I am not sure whether there is that case, but I prefer to provide an input stream rather than byte[]. 
Let's say in this way, if this class compare the changes, then we need extra memory space for the file contents(whole binary), but if we put the content into implementation, there is no extra memory cause, as you already have username/password configuration in provider config.

I think your idea is only coming from the cert file, as you don't need to read it, right? For ES case, this is a double read or extra memory case.

wu-sheng: Anyway, I could provide a default `ContentChangedNotifier` implementation, I think that will fit your wants?

hanahmily: > I think your idea is only coming from the cert file, as you don't need to read it, right? For ES case, this is a double read or extra memory case.

No. For ES and cert file, double read or extra memory is necessary. Because file contents change leads to a total reconnection. For cert file, all incoming gRPC connections should close as ES does. SSL file notifier also loads the binary into heap to compare the content to avoid unnecessary reconnection.

And I think we should provide an abstract class to implement notifier interface, which can do binary comparison. If anyone needs this feature can extend it instead of implementing notifier.

oap-server/server-library/library-util/src/test/java/org/apache/skywalking/oap/server/library/util/FileChangeMonitorTest.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.Charset;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class FileChangeMonitorTest {
+    private static String FILE_NAME = "FileChangeMonitorTest.tmp";
```
kezhenxu94: better place it else where, in case the RAT plugin scans it occasionally

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/FileChangeMonitor.java
wu-sheng: > No. For ES and cert file, double read or extra memory is necessary.

My point is for ES, we don't need extra memory, the user name and password are there, `ContentChangedNotifier` implementation doesn't have to hold it again(one in the ES storage config, the other in the watcher). I know your point to be easier for the watcher, but, this has a cost for ES implementation.

hanahmily: All right. Finally, we reach an agreement to provide to implement `ContentChangedNotifier` to provide this feature. Thanks a lot.

oap-server/server-library/library-util/src/test/java/org/apache/skywalking/oap/server/library/util/FileChangeMonitorTest.java
wu-sheng: What is `FileSystems.getDefault().newWatchService()` for? That doesn't look like a watcher with callback to me. Do I miss anything?

wu-sheng: I think I clear the file in the end of test case. Where should I put? It is different in different OS systems, with different access control maybe.

wu-sheng: Another concern is noticing this
```
     * @throws  UnsupportedOperationException
     *          If this {@code FileSystem} does not support watching file system
     *          objects for changes and events. This exception is not thrown
     *          by {@code FileSystems} created by the default provider.
```
Seems like we need to count on the JRE implementation somehow. I don't know the black magic inside now :)

kezhenxu94: > What is `FileSystems.getDefault().newWatchService()` for? That doesn't look like a watcher with callback to me. Do I miss anything?

Hope the code snippets helps, my concern is that the built-in tool is well optimized and tested, and reduce our maintenance work:

```java
    public static void main(String[] args) throws IOException, InterruptedException {
        final WatchService watchService = FileSystems.getDefault().newWatchService();
        final Path directoryToWatch = Paths.get("/tmp/directory-to-watch");
        directoryToWatch.register(watchService, StandardWatchEventKinds.ENTRY_MODIFY);

        for (final WatchKey readyKey = watchService.take(); readyKey.isValid(); readyKey.reset()) {
            for (final WatchEvent<?> event : readyKey.pollEvents()) {
                assert event.kind().equals(StandardWatchEventKinds.ENTRY_MODIFY);
                final Path path = (Path) event.context();
                System.out.println("file is modified = " + directoryToWatch.resolve(path));
            }
        }
    }
```

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/FileChangeMonitor.java
```diff
@@ -0,0 +1,156 @@
...
+                    notifier.newFileContent(fileInputStream);
+                } catch (FileNotFoundException e) {
+                    log.error("The existed file turns to missing, watch file=" + filePath, e);
+                } catch (IOException e) {
+                    log.error("Read file failure, watch file=" + filePath, e);
+                } finally {
+                    lastModifiedTimestamp = lastModified;
+                }
+            }
+        }
```
kezhenxu94: At least, this part should be replaced, and you can provide a so called `Callback` to the caller

oap-server/server-library/library-util/src/test/java/org/apache/skywalking/oap/server/library/util/FileChangeMonitorTest.java
wu-sheng: My question is, when the exception happens, I never used this API before. Evaluating the real case.

wu-sheng: This exception is rarely to see in the JDK classes. They only say this when necessary. I don't what this simple feature crashes.

wu-sheng:
@hanahmily I add the default implementation by following your prefer, please recheck.

@kezhenxu94 Binding implementation added.

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/FileChangeMonitor.java
```diff
@@ -0,0 +1,188 @@
...
+     * The callback when file changed.
+     */
+    public interface FileChangedNotifier {
+        /**
+         * Notify the new content by providing the file input stream
+         *
+         * @param readableStream points to the new content
+         * @throws IOException if error happens during reading.
+         */
+        void fileChanged(InputStream readableStream) throws IOException;
```
kezhenxu94: I don't think it's a good design, unless you're perfectly sure that the caller will always use the file content as an `InputStream`, otherwise, you create the USELESS input stream too eagerly

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/FileChangeMonitor.java
```diff
@@ -0,0 +1,188 @@
...
+    private long lastCheckTimestamp = 0;
+    /**
+     * The last modify time of the {@link #filePath}
+     */
+    private long lastModifiedTimestamp = 0;
+
+    /**
+     * Start the file monitor for this instance.
+     */
+    public synchronized void start() {
```
kezhenxu94: the `synchronized` makes no sense, it's locked by the caller, and every caller is different from others

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/FileChangeMonitor.java
```diff
@@ -0,0 +1,188 @@
...
+                                                       FileChangeMonitor::run, 1, 1,
+                                                       TimeUnit.SECONDS
+                                                   );
+        }
+
+        this.checkAndNotify();
+        MONITOR_INSTANCES.add(this);
+    }
+
+    public synchronized void stop() {
```
kezhenxu94: same here

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/FileChangeMonitor.java
kezhenxu94: And you may want to prevent the monitor being started more than once

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/FileChangeMonitor.java
wu-sheng: What else could they use?

oap-server/server-library/library-util/src/test/java/org/apache/skywalking/oap/server/library/util/FileChangeMonitorTest.java
dmsolr: I agree with @kezhenxu94 . JDK builtin WatchSerivce is a better choice for the file modified notify service.

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/FileChangeMonitor.java
wu-sheng: Will fix.

wu-sheng:
According to the feedback, I will add file group monitoring as CA cert requires this too. Update later.

wu-sheng:
@hanahmily @kezhenxu94 I submit the new implementation. As in CA cert scenario, there are multiple files which should be treated as an entity, I put these together and provide file content changes directly without `inputstream`/`findNotFound` notifications.

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/MultipleFilesChangeMonitor.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /**
+     * Check file changed status, if so, send the notification.
+     */
+    private void checkAndNotify() {
+        if (System.currentTimeMillis() - lastCheckTimestamp < watchingPeriodInSec * 1000) {
+            // Don't reach the period threshold, ignore this check.
+            return;
+        }
```
dmsolr: Why do we need to double-check? To avoid the new added Monitor will re-check at once?
IMO, we can set the check period as `watchingPeriodInSec`, because this case is a minority case and only affects the first check.

wu-sheng: Because the case is, the timer period is certain and shorter than watcher period. We need to make sure the check period follows the user requirements.

dmsolr: I got it. Monitors own their private schedule, `watchingPeriodInSec`.

PR #4488, opened 2020-03-11
Fix NoSuchElement exception for lazy injection.

wu-sheng:
According to https://github.com/apache/skywalking/pull/4441#issuecomment-597453559, my previous implementation of lazy injection faces an issue. 
If the main thread tracing has been finished, the exit span could still be in async mode, and do `#inject`, this error will show up.

@huangyoje Please give this a try, should be able to fix your issue.


huangyoje:
@wu-sheng OK, I will try soon.

huangyoje:
@wu-sheng I have tested and this fixed my issue.

wu-sheng:
Thanks. I am waiting for another commit to approve this PR :)

PR #4489, opened 2020-03-11
Delete appveyor.yml as don't use there.

wu-sheng:


wu-sheng:
TICKET submitted, too. https://issues.apache.org/jira/browse/INFRA-19955

PR #4490, opened 2020-03-11
Fix GraphQL profile queried error when cannot matches any snapshot

mrproliu:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- GraphQL query profile analyzation without matches any snapshot. Use ArrayList to return trees. Not using `null`.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileAnalyzer.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileAnalyzer.java
@@ -88,7 +88,10 @@ public class ProfileAnalyzer {
         }).flatMap(Collection::stream).map(ProfileStack::deserialize).distinct().collect(Collectors.toList());
 
         // analyze
-        analyzation.getTrees().addAll(analyze(stacks));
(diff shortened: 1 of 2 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profile/analyze/ProfileAnalyzer.java
```diff
@@ -88,7 +88,7 @@ public ProfileAnalyzation analyze(String segmentId, List<ProfileAnalyzeTimeRange
         }).flatMap(Collection::stream).map(ProfileStack::deserialize).distinct().collect(Collectors.toList());
 
         // analyze
-        analyzation.setTrees(analyze(stacks));
+        analyzation.getTrees().addAll(analyze(stacks));
```
wu-sheng: What is the difference? `addAll(null)` will cause a NPE from my understanding.

mrproliu: My fault. I will add an `if` judge before `addAll`

PR #4493, opened 2020-03-11
Support Secrets Management File in the ElasticSearch 6/7 storage

wu-sheng:
### Secrets Management File Of ElasticSearch Username and Password 
The value of `secretsManagementFile` should point to the secrets management file absolute path. 
The file includes username and password of ElasticSearch server in the properties format.
```properties
user=xxx
password=yyy
```

The major difference between using `user/password` configs in the `application.yaml` and this file is, this file is being watched by the OAP server. 
Once it is changed manually or through 3rd party tool, such as [Vault](https://github.com/hashicorp/vault), 
the storage provider will use the new username and password to establish the connection and close the old one. If the information exist in the file,
the `user/password` will be overrided.


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchProvider.java
```diff
@@ -117,6 +120,31 @@ public void prepare() throws ServiceNotProvidedException {
...
+                    if (secretsFileContent == null) {
+                        return;
+                    }
+                    Properties userAndPass = new Properties();
+                    userAndPass.load(new ByteArrayInputStream(secretsFileContent));
+                    config.setUser(userAndPass.getProperty("user"));
+                    config.setPassword(userAndPass.getProperty("password"));
+
+                    if (elasticSearchClient == null) {
+                        //In the startup process, we just need to change the username/password
```
kezhenxu94: Make `elasticSearchClient` volatile?

wu-sheng: Was thinking about this. But not sure. Does across Thread cache will keep forever? Because we are not sensitive this changed immediately, several seconds later is fine.
What do you think about this?

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -119,6 +119,13 @@ public ElasticSearchClient(String clusterNodes,
     @Override
     public void connect() throws IOException, KeyStoreException, NoSuchAlgorithmException, KeyManagementException, CertificateException {
         List<HttpHost> hosts = parseClusterNodes(protocol, clusterNodes);
+        if (client != null) {
+            try {
+                client.close();
+            } catch (Throwable t) {
+                log.error("ElasticSearch client reconnection fails based on new config", t);
+            }
+        }
```
kezhenxu94: Same thread visibility problem for `client`, if the monitor thread failed to detect the `client` is initialized below(in line 129), the `client` is not closed and left open forever

wu-sheng: Good point.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchProvider.java
wu-sheng: Added.

wu-sheng:
@kezhenxu94 @JaredTan95 @hanahmily Could you help on testing on this? Change the JKS or ES username/password and trigger the secrets file changed, check whether the connection reestablishes successfully.

wu-sheng:
Features added. I add the JKS and pass into this secret management machenism.

### ElasticSearch 6 With Https SSL Encrypting communications.
- File at `trustStorePath` is being monitored, once it is changed, the ElasticSearch client will do reconnecting.
- `trustStorePass` could be changed on the runtime through **Secrets Management File Of ElasticSearch Authentication**

### Secrets Management File Of ElasticSearch Authentication
The value of `secretsManagementFile` should point to the secrets management file absolute path. 
The file includes username, password and JKS password of ElasticSearch server in the properties format.
```properties
user=xxx
password=yyy
trustStorePass=zzz
```

The major difference between using `user, password, trustStorePass` configs in the `application.yaml` and this file is, this file is being watched by the OAP server. 
Once it is changed manually or through 3rd party tool, such as [Vault](https://github.com/hashicorp/vault), 
the storage provider will use the new username, password and JKS password to establish the connection and close the old one. If the information exist in the file,
the `user/password` will be overrided.

docs/en/setup/backend/backend-storage.md
```diff
@@ -124,6 +126,20 @@ Such as, if dayStep == 11,
...
+### Secrets Management File Of ElasticSearch Authentication
+The value of `secretsManagementFile` should point to the secrets management file absolute path. 
+The file includes username, password and JKS password of ElasticSearch server in the properties format.
+```properties
+user=xxx
+password=yyy
+trustStorePass=zzz
+```
+
+The major difference between using `user, password, trustStorePass` configs in the `application.yaml` and this file is, this file is being watched by the OAP server.
```
JaredTan95: typo duplicated with `this file is`?

wu-sheng: Yes. And fixed.

JaredTan95:
tested `user/password` changed scenario, and LGTM:

```bash
······
	Caused by: org.elasticsearch.client.ResponseException: method [POST], host [http://localhost:9200], URI [/profile_task/type/_search?typed_keys=true&ignore_unavailable=false&expand_wildcards=open&allow_no_indices=true&search_type=query_then_fetch&batched_reduce_size=512], status line [HTTP/1.1 401 Unauthorized]
{"error":{"root_cause":[{"type":"security_exception","reason":"failed to authenticate user [elastic]","header":{"WWW-Authenticate":"Basic realm=\"security\" charset=\"UTF-8\""}}],"type":"security_exception","reason":"failed to authenticate user [elastic]","header":{"WWW-Authenticate":"Basic realm=\"security\" charset=\"UTF-8\""}},"status":401}
······
org.apache.skywalking.oap.server.library.client.elasticsearch.ElasticSearchClient - 176 [pool-2-thread-1] INFO  [] - elasticsearch cluster nodes: localhost:9200
```

PR #4507, opened 2020-03-13
RestTemplate async plugin ClassCastException

cngdkxw:
org.apache.skywalking.apm.plugin.spring.resttemplate.async.FutureGetInterceptor throw ClassCastException

before fix bug picture
image

after fix bug picture
image


wu-sheng:
Merged, welcome.

PR #4509, opened 2020-03-14
SkyWalking Cross Process Correlation Headers Protocol

mrproliu:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

___
- Provide the data transmission protocol document.


docs/en/protocols/Skywalking-Cross-Process-Correlation-Headers-Protocol-v1.md
```diff
@@ -0,0 +1,24 @@
...
+(string key)-(string value) table in `Base64` encoded JSON format.
+
+Original value raw data `{"userKey1","userValue1","userKey2":"userValue2"}` string, the `Base64` encoded value
+`eyJ1c2VyS2V5MSIsInVzZXJWYWx1ZTEiLCJ1c2VyS2V5MiI6InVzZXJWYWx1ZTIifQ==`.
+
+## Recommendations of language APIs
+Recommended implementation in different language API.
+
+1. `CorrelationContext#set` and `CorrelationContext#get` are recommended to write and read the correlation context, with key/value string.
+1. The key should be added if it is inexistence.
```
kezhenxu94:
```suggestion
1. The key should be added if it is absent.
```

docs/en/protocols/Skywalking-Cross-Process-Correlation-Headers-Protocol-v1.md
```diff
@@ -0,0 +1,24 @@
...
+1. The later the write should override the prev value.
```
kezhenxu94:
```suggestion
1. The later writes should override the previous value.
```

docs/en/protocols/Skywalking-Cross-Process-Correlation-Headers-Protocol-v1.md
```diff
@@ -0,0 +1,24 @@
...
+1. The number of all keys should less than 3, and the length of each value should be less than 128 bytes.
```
kezhenxu94:
```suggestion
1. The total number of all keys should be less than 3, and the length of each value should be less than 128 bytes.
```

docs/en/protocols/Skywalking-Cross-Process-Correlation-Headers-Protocol-v1.md
```diff
@@ -0,0 +1,24 @@
...
+1. The context should be propageted when across thread and across process like do tracing context propagation.
```
kezhenxu94:
```suggestion
1. The context should be propagated as well when tracing context is propagated across threads and processes..
```

wu-sheng:
> I'm wondering whether it's worthwhile to make it JSON, I remember @mrproliu your first suggestion is to encode the key and value respectively, and concatenate them with :, which IMO is good enough and efficient, JSON is more human readable but cost more bandwidth

@kezhenxu94 , the JSON format is not written by me. I think @mrproliu changed this because your question about the `:` in the key or value.

dmsolr:
> Some suggestions inline
> 
> As for the formats, I forgot to reply to the mailing list thread,
> 
> > Cross Process Correlation Header key is `sw7-correlation`. The value is (string key)-(string value) table in `Base64` encoded JSON format.
> 
> I'm wondering whether it's worthwhile to make it JSON, I remember @mrproliu your first suggestion is to encode the `key` and `value` respectively, and concatenate them with `:`, which IMO is good enough and efficient, JSON is more human readable but cost more bandwidth

I have a raw idea. Format: [key.length][key][value.length][key]<[key.length][key][value.length][key]....>

We can use a character to represnet it's length.

wu-sheng:
> I have a raw idea. Format: [key.length][key][value.length][key]<[key.length][key][value.length][key]....>

This is widely used in the binary format, because length could be represented in 2 bytes, but in string format, I am not sure whether we need this.

dmsolr:
> This is widely used in the binary format, because length could be represented in 2 bytes, but in string format, I am not sure whether we need this.

Yes. This format is more widely in binary format.
In fact, `:` and `,` are also represented in 2 bytes. So we spend 2 bytes more only.

mrproliu:
> > I'm wondering whether it's worthwhile to make it JSON, I remember @mrproliu your first suggestion is to encode the key and value respectively, and concatenate them with :, which IMO is good enough and efficient, JSON is more human readable but cost more bandwidth
> 
> @kezhenxu94 , the JSON format is not written by me. I think @mrproliu changed this because your question about the `:` in the key or value.

My first thought was to use internal encode as data encoding, but I later thought that there might be a problem. I just reread the base64 specification, but there is no risk of ':'. In fact, it can be used.
Can we use 'Base64 (userkey1): Base64 (uservalue2), Base64 (userkey2): Base64 (uservalue2) `. But if we are thinking about coding constantly, will it affect the efficiency? @kezhenxu94 @wu-sheng  Any Suggestion?

wu-sheng:
Add the original discuss thread, https://lists.apache.org/thread.html/r50ff0ee6cd927910cad610d258626a58b09982c9e4e8c55827d727e7%40%3Cdev.skywalking.apache.org%3E

kezhenxu94:
> > > I'm wondering whether it's worthwhile to make it JSON, I remember @mrproliu your first suggestion is to encode the key and value respectively, and concatenate them with :, which IMO is good enough and efficient, JSON is more human readable but cost more bandwidth
> > 
> > 
> > @kezhenxu94 , the JSON format is not written by me. I think @mrproliu changed this because your question about the `:` in the key or value.
> 
> My first thought was to use internal encode as data encoding, but I later thought that there might be a problem. I just reread the base64 specification, but there is no risk of ':'. In fact, it can be used.
> Can we use 'Base64 (userkey1): Base64 (uservalue2), Base64 (userkey2): Base64 (uservalue2) `. But if we are thinking about coding constantly, will it affect the efficiency? @kezhenxu94 @wu-sheng Any Suggestion?

Why make it so complicated, I did say `:` could appear in the key or value, but `:` could never appear in **ENCODED** string, my thought is just encode the `key` and `value` respectively, and concatenate them with `:`, to put it simple, the final data would be something like: `<encoded_key>:<encoded_val>`, where `<encoded_key>` and `<encoded_val>` is Base64 encoded, and doesn't contain `:` at all, so the only `:` is our delimiter, did I miss anything @mrproliu ?

mrproliu:
> > > > I'm wondering whether it's worthwhile to make it JSON, I remember @mrproliu your first suggestion is to encode the key and value respectively, and concatenate them with :, which IMO is good enough and efficient, JSON is more human readable but cost more bandwidth
> > > 
> > > 
> > > @kezhenxu94 , the JSON format is not written by me. I think @mrproliu changed this because your question about the `:` in the key or value.
> > 
> > 
> > My first thought was to use internal encode as data encoding, but I later thought that there might be a problem. I just reread the base64 specification, but there is no risk of ':'. In fact, it can be used.
> > Can we use 'Base64 (userkey1): Base64 (uservalue2), Base64 (userkey2): Base64 (uservalue2) `. But if we are thinking about coding constantly, will it affect the efficiency? @kezhenxu94 @wu-sheng Any Suggestion?
> 
> Why make it so complicated, I did say `:` could appear in the key or value, but `:` could never appear in **ENCODED** string, my thought is just encode the `key` and `value` respectively, and concatenate them with `:`, to put it simple, the final data would be something like: `<encoded_key>:<encoded_val>`, where `<encoded_key>` and `<encoded_val>` is Base64 encoded, and doesn't contain `:` at all, so the only `:` is our delimiter, did I miss anything @mrproliu ?

I think we're talking in the same way. I agree with this pattern:
`base64(key):base64(value),base64(key):base64(value)`

wu-sheng:
Even we are not really using W3C spec, but please take a refer from this, https://www.w3.org/TR/trace-context/#combined-header-value

base64(key)=base64(value),base64(key)=base64(value)

wu-sheng:
Notice `tracestate` is not our thing, I just say refer the format only.

mrproliu:
> Even we are not really using W3C spec, but please take a refer from this, https://www.w3.org/TR/trace-context/#combined-header-value
> 
> base64(key)=base64(value),base64(key)=base64(value)

I think that's just a character different, use this way is goods to me.

wu-sheng:
I think the key better point is, `:` has a similar Chinese character, but `=` doesn’t have this issue.

mrproliu:
> > This is widely used in the binary format, because length could be represented in 2 bytes, but in string format, I am not sure whether we need this.
> 
> Yes. This format is more widely in binary format.
> In fact, `:` and `,` are also represented in 2 bytes. So we spend 2 bytes more only.

I have tried your scenarios. That's good to me, but we must make sure the `key` or `value` size must less than `32768`, otherwise, the data will be lost or parsing error.

wu-sheng:
> I have tried your scenarios. That's good to me, but we must make sure the key or value size must less than 32768, otherwise, the data will be lost or parsing error.

I think this is too complex.

kezhenxu94:
> I think the key better point is, `:` has a similar Chinese character, but `=` doesn’t have this issue.

@mrproliu please be advised that `=` is a valid character in encoded base64, and it often appears in encoded text as padding character, the documentations @wu-sheng you referred to is clear text, not encoded

wu-sheng:
@kezhenxu94 Don’t use that(not encode). That is what I mean we are not following W3C spec. That is not popular and not realistic. I just want to recommend `=`.

mrproliu:
> > I think the key better point is, `:` has a similar Chinese character, but `=` doesn’t have this issue.
> 
> @mrproliu please be advised that `=` is a valid character in encoded base64, and it often appears in encoded text as padding character, the documentations @wu-sheng you referred to is clear text, not encoded

Ops, I forget this point. I think we need to keep the `:` to separate the `key/value`.

wu-sheng:
@kezhenxu94 @mrproliu @dmsolr I changed the document by using `=` and `,`.

docs/en/protocols/Skywalking-Cross-Process-Correlation-Headers-Protocol-v1.md
```diff
@@ -0,0 +1,16 @@
...
+Cross Process Correlation Header key is `sw7-correlation`. The value is the `encoded(key):encoded(value)` list with elements splitted by `,` such as `base64(string key):base64(string value),base64(string key2):base64(string value2)`.
+
+## Recommendations of language APIs
+Recommended implementation in different language API.
+
+1. `CorrelationContext#set` and `CorrelationContext#get` are recommended to write and read the correlation context, with key/value string.
+1. The key should be added if it is absent.
+1. The later writes should override the previous value.
+1. The total number of all keys should be less than 3, and the length of each value should be less than 128 bytes.
+1. The context should be propagated as well when tracing context is propagated across threads and processes..
```
arugal:
```suggestion
1. The context should be propagated as well when tracing context is propagated across threads and processes.
```

wu-sheng:
I am going to merge this soon. @mrproliu You could feel free to work on this implementation.

Issue #4472, opened 2020-03-09
[Refactor] E2E test framework


kezhenxu94:
According to the feedbacks from our developers, committers and PMCs, especially from @JaredTan95 , the E2E framework is rather unfriendly to our contributors, especially new contributors, therefore I'm refactoring the E2E test framework, the goals are:

- more debuggable, locally, in IDE

- more intuitive, adopt well known technologies to let people get started more quickly

- less verbose/duplicated codes

I've had a runnable version and may need some time to refactor the remains test cases, hope it would be ready soon, anyone who has more ideas about how the new E2E tests should work (friendly), please let me know

wu-sheng:
I think debug in IDE is the key.

PR #4483, opened 2020-03-10
[BugFix] Docker InfluxDB cannot be enabled

kezhenxu94:
### Motivation:

This patch fixes the bug that there's no chance to enable the InfluxDB storage in docker container.

### Modifications:

Generate InfluxDB configuration section to enable it.

### Result:

Users can enable InfluxDB by specifying SW_STORAGE=influxdb

PR #4510, opened 2020-03-14
Rewrite the E2E test framework and tests

kezhenxu94:
Motivation:

According the feedback from the community, the e2e tests are hard for
them to understand and get them started, so this pull request aims to
rewrite the e2e tests with some improvements:

- debuggable locally, now you can run every individual case directly
inside the IDE, just like running a plain unit test.

- same functionalities of multiple counterparts share the same codes,
cases like different storages and cluster coordinators share the same
codes, easy to add more similar cases.

- the "end" now extends to the Doccker images, meaning that some bugs
like #4483 and #4471 can be more easily and early to be discovered.

- codes style and license headers are checked

Modifications:

- adopt jupeter (a.k.a junit 5), testcontainers as the new test
framework, all containers are orchestrated by docker-compose.yml,
well-known technology that developers should be familiar with.

- rewrite all test cases and verify them.

- reorganize the workflows in GitHub Actions.

Result:

- new e2e test should be easy to add.

- closes https://github.com/apache/skywalking/issues/4472


kezhenxu94:
It's hard to review the diff, so just checkout the branch locally and review the final codes

wu-sheng:
You need to keep name same, especially all CI required tasks. Otherwise, this can't be merged.

.github/workflows/e2e.yaml
```diff
@@ -24,125 +24,97 @@ on:
...
-        run: export E2E_VERSION=jdk12-1.5 && bash -x test/e2e/run.sh e2e-single-service
-      - name: Agent Reboot Tests(JDK8)
-        run: export E2E_VERSION=jdk8-1.5 && bash -x test/e2e/run.sh e2e-agent-reboot
-      - name: Agent Auth Tests(JDK8)
-        run: export E2E_VERSION=jdk8-1.5 && bash -x test/e2e/run.sh e2e-agent-auth
+      - name: Checkstyle
+        run: ./mvnw --batch-mode -f test/e2e/pom.xml -DskipTests checkstyle:check
 
-  Cluster:
+  FeaturesGroup1:
```
wu-sheng: Once this name changed, we are facing the awkward situation. The PR can't be merged as we set that by the INFRA team. Could we consider to change it back?

test/e2e/e2e-data/src/main/resources/profiledSegment.gql
```diff
@@ -1,10 +1,9 @@
...
-# regarding copyright ownership.  The ASF licenses this file
-# to you under the Apache License, Version 2.0 (the
-# "License"); you may not use this file except in compliance
-# with the License.  You may obtain a copy of the License at
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
```
wu-sheng: You are changing the header format, why?

kezhenxu94: > You are changing the header format, why?

The license headers in such kind of files are checked by Checkstyle plugin, by matching regular expression, so, have to

wu-sheng:
What are these? Some mistakes?

image

.github/workflows/e2e.yaml
kezhenxu94: > Once this name changed, we are facing the awkward situation. The PR can't be merged as we set that by the INFRA team. Could we consider to change it back?

I begin to use matrix builds, so even I change back to `Single` or `Cluster`, the names are still not the same as before, besides, the original name is rather vague, and we could make it more clear, if you agree, I can open a ticket to ask the infra team for the help, to change the required statuses

wu-sheng: I am OK to change. Just we need
1. A consensus  about the task lists 
1. Your PR passed and ready to be merged immediately. 

We could request the INFRA. Because otherwise, we can’t merge any PR if INFRA changed that before this is ready.

wu-sheng:
#4470 merged, so you are facing conflicts, please add this case in here too.

kezhenxu94:
@wu-sheng this pull request is ready IMO, and I add it back to 7.0.0, please review and let me know if it's possible to included in 7.0.0

wu-sheng:
Once this PR merged, please send the notification mails to
1. Release vote as a verified result
1. Juan Pan about the update
1. Separated notification mail about the e2e framework changed totally.

PR #4514, opened 2020-03-15
Add selector property in application.yml

### LINKED ISSUE #4511: [DISCUSSION] Optimize configurations `application.yml`
### Background

During the E2E test refactoring work, I found it troublesome to change some of the configurations, the similarities are that they are all single-choice configuration, examples are storages, cluster coordinators, dynamic configuration implementations

### Current status

The current solution is to list them all, but comment out some of them, only leaving one item uncommented, the trouble is when we're changing the configuration in auto workflow (such as Docker container, e2e tests), we have to do many string manipulations, (awk, shell, etc.), which is error prone and unstable.

### Proposal

I'm proposing to organize the configurations as follows, taking the storage as an example:

<details>

<summary>click to expand the suggested format</summary>

```yaml
storage:
  selector: ${SW_STORAGE:elasticsearch7}
  elasticsearch:
    nameSpace: ${SW_NAMESPACE:""}
    clusterNodes: ${SW_STORAGE_ES_CLUSTER_NODES:localhost:9200}
    protocol: ${SW_STORAGE_ES_HTTP_PROTOCOL:"http"}
    trustStorePath: ${SW_SW_STORAGE_ES_SSL_JKS_PATH:"../es_keystore.jks"}
    trustStorePass: ${SW_SW_STORAGE_ES_SSL_JKS_PASS:""}
    enablePackedDownsampling: ${SW_STORAGE_ENABLE_PACKED_DOWNSAMPLING:true} # Hour and Day metrics will be merged into minute index.
    dayStep: ${SW_STORAGE_DAY_STEP:1} # Represent the number of days in the one minute/hour/day index.
    user: ${SW_ES_USER:""}
    password: ${SW_ES_PASSWORD:""}
    secretsManagementFile: ${SW_ES_SECRETS_MANAGEMENT_FILE:""} # Secrets management file in the properties format includes the username, password, which are managed by 3rd party tool.
    indexShardsNumber: ${SW_STORAGE_ES_INDEX_SHARDS_NUMBER:2}
    indexReplicasNumber: ${SW_STORAGE_ES_INDEX_REPLICAS_NUMBER:0}
    # Those data TTL settings will override the same settings in core module.
    recordDataTTL: ${SW_STORAGE_ES_RECORD_DATA_TTL:7} # Unit is day
    otherMetricsDataTTL: ${SW_STORAGE_ES_OTHER_METRIC_DATA_TTL:45} # Unit is day
    monthMetricsDataTTL: ${SW_STORAGE_ES_MONTH_METRIC_DATA_TTL:18} # Unit is month
    # Batch process setting, refer to https://www.elastic.co/guide/en/elasticsearch/client/java-api/5.5/java-docs-bulk-processor.html
    bulkActions: ${SW_STORAGE_ES_BULK_ACTIONS:1000} # Execute the bulk every 1000 requests
    flushInterval: ${SW_STORAGE_ES_FLUSH_INTERVAL:10} # flush the bulk every 10 seconds whatever the number of requests
    concurrentRequests: ${SW_STORAGE_ES_CONCURRENT_REQUESTS:2} # the number of concurrent requests
    resultWindowMaxSize: ${SW_STORAGE_ES_QUERY_MAX_WINDOW_SIZE:10000}
    metadataQueryMaxSize: ${SW_STORAGE_ES_QUERY_MAX_SIZE:5000}
    segmentQueryMaxSize: ${SW_STORAGE_ES_QUERY_SEGMENT_SIZE:200}
    profileTaskQueryMaxSize: ${SW_STORAGE_ES_QUERY_PROFILE_TASK_SIZE:200}
  elasticsearch7:
    nameSpace: ${SW_NAMESPACE:""}
    clusterNodes: ${SW_STORAGE_ES_CLUSTER_NODES:localhost:9200}
    protocol: ${SW_STORAGE_ES_HTTP_PROTOCOL:"http"}
    trustStorePath: ${SW_SW_STORAGE_ES_SSL_JKS_PATH:"../es_keystore.jks"}
    trustStorePass: ${SW_SW_STORAGE_ES_SSL_JKS_PASS:""}
    enablePackedDownsampling: ${SW_STORAGE_ENABLE_PACKED_DOWNSAMPLING:true} # Hour and Day metrics will be merged into minute index.
    dayStep: ${SW_STORAGE_DAY_STEP:1} # Represent the number of days in the one minute/hour/day index.
    user: ${SW_ES_USER:""}
    password: ${SW_ES_PASSWORD:""}
    secretsManagementFile: ${SW_ES_SECRETS_MANAGEMENT_FILE:""} # Secrets management file in the properties format includes the username, password, which are managed by 3rd party tool.
    indexShardsNumber: ${SW_STORAGE_ES_INDEX_SHARDS_NUMBER:2}
    indexReplicasNumber: ${SW_STORAGE_ES_INDEX_REPLICAS_NUMBER:0}
    # Those data TTL settings will override the same settings in core module.
    recordDataTTL: ${SW_STORAGE_ES_RECORD_DATA_TTL:7} # Unit is day
    otherMetricsDataTTL: ${SW_STORAGE_ES_OTHER_METRIC_DATA_TTL:45} # Unit is day
    monthMetricsDataTTL: ${SW_STORAGE_ES_MONTH_METRIC_DATA_TTL:18} # Unit is month
    # Batch process setting, refer to https://www.elastic.co/guide/en/elasticsearch/client/java-api/5.5/java-docs-bulk-processor.html
    bulkActions: ${SW_STORAGE_ES_BULK_ACTIONS:1000} # Execute the bulk every 1000 requests
    flushInterval: ${SW_STORAGE_ES_FLUSH_INTERVAL:10} # flush the bulk every 10 seconds whatever the number of requests
    concurrentRequests: ${SW_STORAGE_ES_CONCURRENT_REQUESTS:2} # the number of concurrent requests
    resultWindowMaxSize: ${SW_STORAGE_ES_QUERY_MAX_WINDOW_SIZE:10000}
    metadataQueryMaxSize: ${SW_STORAGE_ES_QUERY_MAX_SIZE:5000}
    segmentQueryMaxSize: ${SW_STORAGE_ES_QUERY_SEGMENT_SIZE:200}
    profileTaskQueryMaxSize: ${SW_STORAGE_ES_QUERY_PROFILE_TASK_SIZE:200}
  h2:
    driver: ${SW_STORAGE_H2_DRIVER:org.h2.jdbcx.JdbcDataSource}
    url: ${SW_STORAGE_H2_URL:jdbc:h2:mem:skywalking-oap-db}
    user: ${SW_STORAGE_H2_USER:sa}
    metadataQueryMaxSize: ${SW_STORAGE_H2_QUERY_MAX_SIZE:5000}
  mysql:
    properties:
      jdbcUrl: ${SW_JDBC_URL:"jdbc:mysql://localhost:3306/swtest"}
      dataSource.user: ${SW_DATA_SOURCE_USER:root}
      dataSource.password: ${SW_DATA_SOURCE_PASSWORD:root@1234}
      dataSource.cachePrepStmts: ${SW_DATA_SOURCE_CACHE_PREP_STMTS:true}
      dataSource.prepStmtCacheSize: ${SW_DATA_SOURCE_PREP_STMT_CACHE_SQL_SIZE:250}
      dataSource.prepStmtCacheSqlLimit: ${SW_DATA_SOURCE_PREP_STMT_CACHE_SQL_LIMIT:2048}
      dataSource.useServerPrepStmts: ${SW_DATA_SOURCE_USE_SERVER_PREP_STMTS:true}
    metadataQueryMaxSize: ${SW_STORAGE_MYSQL_QUERY_MAX_SIZE:5000}
  influx:
    # Metadata storage provider configuration
    metabaseType: ${SW_STORAGE_METABASE_TYPE:H2} # There are 2 options as Metabase provider, H2 or MySQL.
    h2Props:
      dataSourceClassName: ${SW_STORAGE_METABASE_DRIVER:org.h2.jdbcx.JdbcDataSource}
      dataSource.url: ${SW_STORAGE_METABASE_URL:jdbc:h2:mem:skywalking-oap-db}
      dataSource.user: ${SW_STORAGE_METABASE_USER:sa}
      dataSource.password: ${SW_STORAGE_METABASE_PASSWORD:}
    mysqlProps:
      jdbcUrl: ${SW_STORAGE_METABASE_URL:"jdbc:mysql://localhost:3306/swtest"}
      dataSource.user: ${SW_STORAGE_METABASE_USER:root}
      dataSource.password: ${SW_STORAGE_METABASE_PASSWORD:root@1234}
      dataSource.cachePrepStmts: ${SW_STORAGE_METABASE_CACHE_PREP_STMTS:true}
      dataSource.prepStmtCacheSize: ${SW_STORAGE_METABASE_PREP_STMT_CACHE_SQL_SIZE:250}
      dataSource.prepStmtCacheSqlLimit: ${SW_STORAGE_METABASE_PREP_STMT_CACHE_SQL_LIMIT:2048}
      dataSource.useServerPrepStmts: ${SW_STORAGE_METABASE_USE_SERVER_PREP_STMTS:true}
    metadataQueryMaxSize: ${SW_STORAGE_METABASE_QUERY_MAX_SIZE:5000}
    # InfluxDB configuration
    url: ${SW_STORAGE_INFLUXDB_URL:http://localhost:8086}
    user: ${SW_STORAGE_INFLUXDB_USER:root}
    password: ${SW_STORAGE_INFLUXDB_PASSWORD:}
    database: ${SW_STORAGE_INFLUXDB_DATABASE:skywalking}
    actions: ${SW_STORAGE_INFLUXDB_ACTIONS:1000} # the number of actions to collect
    duration: ${SW_STORAGE_INFLUXDB_DURATION:1000} # the time to wait at most (milliseconds)
    fetchTaskLogMaxSize: ${SW_STORAGE_INFLUXDB_FETCH_TASK_LOG_MAX_SIZE:5000} # the max number of fetch task log in a request
```
</details>

the key point is to leave all choices uncommented, and add a `selector` to indicate which will take effect actually, in this way, we can simply change this selector by environment variable or system property.

kezhenxu94:
### Motivation
    
Ease the efforts of modifying configurations in automatic workflow and scripts, like Docker and helm chart.
    
### Modification
    
Add a `selector` property into `application.yml` to enable switching configuration simply by environment variables or system properties, and filter the unselected options, leaving only selected one.
    
### Result
    
- We can switch a configuration by environment variables or system property
    
- Closes #4511


kezhenxu94:
I just left the `application.yml`s in E2E intact, as they will be totally replaced in #4510 , and we can verify the compatibility with old `applicationl.yml` by the way

kezhenxu94:
Better to enable this option in this PR when reviewing because there're some spaces changes

image

docker/oap-es7/docker-entrypoint.sh
```diff
@@ -294,202 +33,51 @@ validateVariables() {
...
-    zookeeper) generateClusterZookeeper;;
-    kubernetes) generateClusterK8s;;
-    consul) generateClusterConsul;;
-    etcd) generateClusterEtcd;;
-    nacos) generateClusterNacos;;
-    esac
+# validate
+[[ -z "$SW_CLUSTER" ]] || [[ -z "$SW_STORAGE" ]] || [[ -z "$SW_CONFIGURATION" ]] \
+    || [[ -z "$SW_TELEMETRY" ]] \
+    && { echo "Error: please specify \"SW_CLUSTER\" \"SW_STORAGE\" \"SW_CONFIGURATION\" \"SW_TELEMETRY\""; exit 1; }
```
kezhenxu94: @hanahmily please review these, the original logic operators are `&&`, but according to the error message, I think they should be `||`, if I'm wrong, please correct me

oap-server/server-bootstrap/src/main/java/org/apache/skywalking/oap/server/starter/config/ApplicationConfigLoader.java
```diff
@@ -105,6 +107,19 @@ private void overrideConfigBySystemEnv(ApplicationConfiguration configuration) {
...
+    private void selectConfig(final Map<String, Object> configuration) {
+        if (configuration.size() <= 1) {
+            return;
+        }
+        if (configuration.containsKey("selector")) {
+            final String selector = (String) configuration.get("selector");
+            final String resolvedSelector = PropertyPlaceholderHelper.INSTANCE.replacePlaceholders(
+                selector, System.getProperties()
+            );
+            configuration.entrySet().removeIf(e -> !resolvedSelector.equals(e.getKey()));
```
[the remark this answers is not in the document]
kezhenxu94: > I think you need to add a check and error for selected provider not found, when you give a value to selector, but no suitable provider definition here.

Yes, thanks for pointing out, just found that and fixed

wu-sheng:
Please notice k8s installation fails. Whether something wrong about the docker?

docker/oap-es7/docker-entrypoint.sh
hanahmily: I think we should remove this assertation, leave it to oap config value checker.

docker/oap-es7/docker-entrypoint.sh
```diff
@@ -294,202 +33,51 @@ validateVariables() {
...
-    minuteMetricsDataTTL: \${SW_CORE_MINUTE_METRIC_DATA_TTL:90} # Unit is minute
-    hourMetricsDataTTL: \${SW_CORE_HOUR_METRIC_DATA_TTL:36} # Unit is hour
-    dayMetricsDataTTL: \${SW_CORE_DAY_METRIC_DATA_TTL:45} # Unit is day
-    monthMetricsDataTTL: \${SW_CORE_MONTH_METRIC_DATA_TTL:18} # Unit is month
-    # Cache metric data for 1 minute to reduce database queries, and if the OAP cluster changes within that minute,
-    # the metrics may not be accurate within that minute.
-    enableDatabaseSession: \${SW_CORE_ENABLE_DATABASE_SESSION:true}
-    topNReportPeriod: \${SW_CORE_TOPN_REPORT_PERIOD:10}
-EOT
+validateVariables "SW_CLUSTER" "$SW_CLUSTER" "standalone zookeeper kubernetes consul etcd nacos"
```
hanahmily: remove it.

docker/oap-es7/docker-entrypoint.sh
```diff
@@ -294,202 +33,51 @@ validateVariables() {
...
 
-    # generate storage
-    case ${SW_STORAGE} in
-    elasticsearch) generateStorageElastisearch;;
-    h2) generateStorageH2;;
-    mysql) generateStorageMySQL;;
-    influxdb) generateStorageInfluxDB;;
-    esac
+validateVariables "SW_STORAGE" "$SW_STORAGE" "elasticsearch h2 mysql influxdb"
```
hanahmily: the same, remove it.

docker/oap-es7/docker-entrypoint.sh
```diff
@@ -294,202 +33,51 @@ validateVariables() {
...
-alarm:
-  default:
-EOT
-    # generate telemetry
-    case ${SW_TELEMETRY} in
-    none) generateTelemetryNone;;
-    prometheus) generateTelemetryPrometheus;;
-    so11y) generateTelemetrySo11y;;
-    esac
+validateVariables "SW_CONFIGURATION" "$SW_CONFIGURATION" "none apollo nacos zookeeper"
```
hanahmily: same as above, remove this line.

docker/oap-es7/docker-entrypoint.sh
```diff
@@ -294,202 +33,51 @@ validateVariables() {
...
-    # generate configuration
-    case ${SW_CONFIGURATION} in
-    none) generateConfigurationNone;;
-    apollo) generateConfigurationApollo;;
-    nacos) generateConfigurationNacos;;
-    zookeeper) generateConfigurationZookeeper;;
-    consul) generateConfigurationConsul;;
-    grpc) generateConfigurationGRPC;;
-    esac
+validateVariables "SW_TELEMETRY" "$SW_TELEMETRY" "none prometheus so11y"
```
hanahmily: remove it.

docker/oap-es7/docker-entrypoint.sh
```diff
@@ -294,202 +33,51 @@ validateVariables() {
...
-
 # Override configuration files
 cp -vfR ${EXT_CONFIG_DIR}/ config/
-if [[ -z "$SW_L0AD_CONFIG_FILE_FROM_VOLUME" ]] || [[ "$SW_L0AD_CONFIG_FILE_FROM_VOLUME" != "true" ]]; then
-    generateApplicationYaml
-    echo "Generated application.yml"
-    echo "-------------------------"
-    cat ${var_application_file}
-    echo "-------------------------"
-fi
```
hanahmily: Why removing this block? user can't override generated yaml withou it.

docker/oap-es7/docker-entrypoint.sh
```diff
@@ -294,202 +33,51 @@ validateVariables() {
...
-  default:
-    host: \${SW_RECEIVER_ZIPKIN_HOST:0.0.0.0}
-    port: \${SW_RECEIVER_ZIPKIN_PORT:9411}
-    contextPath: \${SW_RECEIVER_ZIPKIN_CONTEXT_PATH:/}
-EOT
-    fi
+if [[ "$SW_ENVOY_ALS_ENABLED" = "true" ]]; then
+    export SW_ENVOY_METRIC_ALS_HTTP_ANALYSIS=k8s-mesh
+    echo "Set SW_ENVOY_METRIC_ALS_HTTP_ANALYSIS to ${SW_ENVOY_METRIC_ALS_HTTP_ANALYSIS}"
+fi
```
hanahmily: I suggest removing these codes. I should provide some "backwards" compbility, just like "SW_STORAGE", but keeping it only add complex and confusing variables convertion.

docker/oap-es7/docker-entrypoint.sh
```diff
@@ -294,202 +33,51 @@ validateVariables() {
...
-receiver_jaeger:
-  default:
-    gRPCHost: \${SW_RECEIVER_JAEGER_HOST:0.0.0.0}
-    gRPCPort: \${SW_RECEIVER_JAEGER_PORT:14250}
-EOT
-    fi
+if [[ "$SW_RECEIVER_ZIPKIN_ENABLED" = "true" ]]; then
+    export SW_RECEIVER_ZIPKIN=default
+    echo "Set SW_RECEIVER_ZIPKIN to ${SW_RECEIVER_ZIPKIN}"
+fi
```
hanahmily: same as above.

docker/oap-es7/docker-entrypoint.sh
```diff
@@ -294,202 +33,51 @@ validateVariables() {
...
-    if [[ "$SW_TELEMETRY" = "so11y" ]]; then
-        cat <<EOT >> ${var_application_file}
-receiver-so11y:
-  default:
-EOT
-    fi
+if [[ "$SW_RECEIVER_JAEGER_ENABLED" = "true" ]]; then
+    export SW_RECEIVER_JAEGER=default
+    echo "Set SW_RECEIVER_JAEGER to ${SW_RECEIVER_JAEGER}"
+fi
```
hanahmily: same

docker/oap-es7/docker-entrypoint.sh
```diff
@@ -294,202 +33,51 @@ validateVariables() {
...
+if [[ "$SW_TELEMETRY" = "so11y" ]]; then
+    export SW_RECEIVER_SO11Y=default
+    echo "Set SW_RECEIVER_SO11Y to ${SW_RECEIVER_SO11Y}"
+fi
 
-echo "[Entrypoint] Apache SkyWalking Docker Image"
+if [[ "$SW_EXPORTER_ENABLED" = "true" ]]; then
+    export SW_EXPORTER=grpc
+    echo "Set SW_EXPORTER to ${SW_EXPORTER}"
+fi
```
hanahmily: same as above, remove it.

hanahmily:
@kezhenxu94 Good job! Except for some nits.

I prefer to remove unnecessary environment variable converting in entrypoint script, so let's just delete them. But keep `so11y` related is a good idea because of the link between receiver and telemetry. Thanks

docker/oap-es7/docker-entrypoint.sh
wu-sheng: Receiver enable seems a little different, @hanahmily. They are separated new module, not provider. We may need another feature to resolve this for docker.

kezhenxu94:
Thanks for reviewing, the review comments will be addressed after figuring out the reason why tests failed

dist-material/application.yml
```diff
@@ -83,180 +85,230 @@ core:
...
-#    alsHTTPAnalysis: ${SW_ENVOY_METRIC_ALS_HTTP_ANALYSIS:k8s-mesh}
-#receiver_zipkin:
-#  default:
-#    host: ${SW_RECEIVER_ZIPKIN_HOST:0.0.0.0}
-#    port: ${SW_RECEIVER_ZIPKIN_PORT:9411}
-#    contextPath: ${SW_RECEIVER_ZIPKIN_CONTEXT_PATH:/}
+    alsHTTPAnalysis: ${SW_ENVOY_METRIC_ALS_HTTP_ANALYSIS:none}
+
+receiver_zipkin:
+  selector: ${SW_RECEIVER_ZIPKIN:none}
```
wu-sheng: What does this `none` mean? I don't think we have `none` provider for this receiver.

According to your implementation, https://github.com/apache/skywalking/blob/dae57dc8fddba8916d43351cdde824a116e2d58c/oap-server/server-bootstrap/src/main/java/org/apache/skywalking/oap/server/starter/config/ApplicationConfigLoader.java#L62-L64,
this seems meaning no provider for module `receiver_zipkin`? If this module is required by other module, what happens?

I don't think providing a potential NPE in the runtime is a good idea.

wu-sheng: Also, based on this, I could put some tricky parameter value such as any not existing provider name. And we magically accept it and start up successfully.

wu-sheng: At the module core level, should be strict validation to avoid illegal selector name.

kezhenxu94: > If this module is required by other module, what happens?
I don't think providing a potential NPE in the runtime is a good idea.
> Also, based on this, I could put some tricky parameter value such as any not existing provider name. And we magically accept it and start up successfully.

Can I just think of that this problem is not related to this pull request directly and it should be fixed in another pull request? In the old way, if the user just comment out the required module, what would happen? Are they the same problem here?

wu-sheng: Read old `ModuleDefine#prepare`. Once the provider doesn't exist, `throw new ProviderNotFoundException(this.name() + " module no provider exists.");` this shows up.

But because we enable multiple providers in YAML and using the selector, this check wouldn't work, that is why I asked.

kezhenxu94: @wu-sheng hope I got your point correctly and I try to fix it at the phase when processing the configuration, to ensure that the processed configurations are the same as before, and verify like this, correct me if I made anything wrong still

image

wu-sheng: This fails because some module request cluster module in the bootstrap stage, if not, then this error wouldn't show up. My point is, this should be a check rule.

oap-server/server-bootstrap/src/main/java/org/apache/skywalking/oap/server/starter/config/ApplicationConfigLoader.java
```diff
@@ -105,6 +109,26 @@ private void overrideConfigBySystemEnv(ApplicationConfiguration configuration) {
...
+        final Set<String> modulesWithoutProvider = new HashSet<>();
+        for (final Map.Entry<String, Map<String, Object>> entry : moduleConfiguration.entrySet()) {
+            final String moduleName = entry.getKey();
+            final Map<String, Object> providerConfig = entry.getValue();
+            if (providerConfig.containsKey("selector")) {
+                final String selector = (String) providerConfig.get("selector");
+                final String resolvedSelector = PropertyPlaceholderHelper.INSTANCE.replacePlaceholders(
+                    selector, System.getProperties()
+                );
+                providerConfig.entrySet().removeIf(e -> !resolvedSelector.equals(e.getKey()));
```
wu-sheng: @kezhenxu94 To be more clear,  after this remove, it could be no provider available, we should check and report error immidiately.

wu-sheng:
`backend-setup.md` should be updated too, about new `selector` and the 6.x way still supported.

docker/oap-es7/docker-entrypoint.sh
```diff
@@ -18,478 +18,18 @@
...
-
-generateApplicationYaml() {
-    # validate
-    [[ -z "$SW_CLUSTER" ]] && [[ -z "$SW_STORAGE" ]] && [[ -z "$SW_CONFIGURATION" ]] \
-        && [[ -z "$SW_TELEMETRY" ]] \
-        && { echo "Error: please specify \"SW_CLUSTER\" \"SW_STORAGE\" \"SW_CONFIGURATION\" \"SW_TELEMETRY\""; exit 1; }
-
-    validateVariables "SW_CLUSTER" "$SW_CLUSTER" "standalone zookeeper kubernetes consul etcd nacos"
-
-    validateVariables "SW_STORAGE" "$SW_STORAGE" "elasticsearch h2 mysql influxdb"
```
kezhenxu94: FYI @wu-sheng and @hanahmily , here is another bug in the Docker image, with this, the docker image for es7 never works

wu-sheng:
@kezhenxu94 I am sure why, but GitHub request you update branch manually.

docs/en/setup/backend/backend-setup.md
```diff
@@ -14,26 +14,41 @@ End user can switch or assemble the collector features by their own requirements
...
+      dataSource.useServerPrepStmts: ${SW_DATA_SOURCE_USE_SERVER_PREP_STMTS:true}
+    metadataQueryMaxSize: ${SW_STORAGE_MYSQL_QUERY_MAX_SIZE:5000}
+  # other configurations
 ```
-1. **core** is the module.
-1. **default** is the default implementor of core module.
-1. `restHost`, `restPort`, ... `gRPCHost` are all setting items of the implementor.
+
+1. **`core`** is the module.
+1. **`selector`** selects one out of the all providers listed below, the unselected ones take no effect as if they were deleted.
```
wu-sheng: I think we also should indicate `selector` is optional if there is only one provider.

docs/en/setup/backend/backend-setup.md
```diff
@@ -14,26 +14,41 @@ End user can switch or assemble the collector features by their own requirements
...
+1. **`default`** is the default implementor of core module.
+1. `driver`, `url`, ... `metadataQueryMaxSize` are all setting items of the implementor.
 
 At the same time, modules includes required and optional, the required modules provide the skeleton of backend,
-even modularization supported pluggable, remove those modules are meaningless. We highly recommend you don't try to
-change APIs of those modules, unless you understand SkyWalking project and its codes very well.
+even modularization supported pluggable, removing those modules are meaningless, for optional modules, if there is provider
+called `none`, you can simply set it to the `selector` to disable it, otherwise, please set `-` to the selector.
```
wu-sheng: `if there is provider called `none`, you can simply set it to the `selector` to disable it, otherwise, please set `-` to the selector`
->

Some modules have a provider implementation called none, meaning it only provides a shell with no actual logic, typically such as telemetry.

Set `-` to the selector means this whole module will be excluded in the runtime.

oap-server/server-bootstrap/src/main/java/org/apache/skywalking/oap/server/starter/config/ApplicationConfigLoader.java
```diff
@@ -105,6 +112,48 @@ private void overrideConfigBySystemEnv(ApplicationConfiguration configuration) {
...
+                selector, System.getProperties()
+            );
+            providerConfig.entrySet().removeIf(e -> !resolvedSelector.equals(e.getKey()));
+
+            if (!providerConfig.isEmpty()) {
+                continue;
+            }
+
+            if (!DISABLE_SELECTOR.equals(resolvedSelector)) {
+                throw new RuntimeException(
```
wu-sheng: This should be `ProviderNotFoundException`

oap-server/server-bootstrap/src/main/java/org/apache/skywalking/oap/server/starter/config/ApplicationConfigLoader.java
```diff
@@ -105,6 +112,48 @@ private void overrideConfigBySystemEnv(ApplicationConfiguration configuration) {
...
+            // now the module can be safely removed
+            modulesWithoutProvider.add(moduleName);
+        }
+
+        moduleConfiguration.entrySet().removeIf(e -> {
+            final String module = e.getKey();
+            final boolean shouldBeRemoved = modulesWithoutProvider.contains(module);
+
+            if (shouldBeRemoved) {
+                logger.info("removing module {} without any provider", module);
```
wu-sheng:
```suggestion
                logger.info("Remove module {} without any provider", module);
```

docs/en/setup/backend/backend-setup.md
```diff
@@ -14,26 +14,41 @@ End user can switch or assemble the collector features by their own requirements
 
 So, in `application.yml`, there are three levels.
 1. **Level 1**, module name. Meaning this module is active in running mode.
-1. **Level 2**, provider name. Set the provider of the module.
+1. **Level 2**, provider name and provider selector. Available providers are listed here with a selector to indicate which one will actually take effect.
```
wu-sheng:
```suggestion
1. **Level 2**, provider option list and provider selector. Available providers are listed here with a selector to indicate which one will actually take effect.
```

PR #4528, opened 2020-03-17
Fix profile trace not match when change the entry operationId/Name

mrproliu:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

___
### Bug fix
- Profile trace does not match with dynamic change the name, such as `SpringMVC`. I put the recheck profile in the wrong place.


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/TracingContext.java
```diff
@@ -353,6 +353,12 @@ public AbstractSpan createEntrySpan(final String operationName) {
         final AbstractSpan parentSpan = peek();
         final int parentSpanId = parentSpan == null ? -1 : parentSpan.getSpanId();
         if (parentSpan != null && parentSpan.isEntry()) {
+            /**
+             * Only add the profiling recheck on change entry span operationId/Name.
+             * Because Change name only happen on entry span.
+             * Exit and local span will check need profiling on construct the context.
+             */
```
kezhenxu94:
```suggestion
            /*
             * Only add the profiling recheck when changing an entry span's operationId/Name.
             * Because changing name only happens on entry span.
             * Exit and local span will check when constructing the context.
             */
```

wu-sheng: Already reword this comment.

PR #4529, opened 2020-03-17
[hotfix] Fix AlarmQuery query condition issue

dmsolr:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

`regex(..)` need to append prefix and suffix, '/'. 

- How to fix?
Uses `contains(...)` to instead of it.

### DIFF
--- a/oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/AlarmQuery.java
+++ b/oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/AlarmQuery.java
@@ -64,7 +64,7 @@ public class AlarmQuery implements IAlarmQueryDAO {
                        .and(lte(InfluxClient.TIME, InfluxClient.timeInterval(endTB)));
         }
         if (!Strings.isNullOrEmpty(keyword)) {
-            recallQuery.and(contains(AlarmRecord.ALARM_MESSAGE, keyword.replaceAll("/", "\\/")));
+            recallQuery.and(contains(AlarmRecord.ALARM_MESSAGE, keyword.replaceAll("/", "\\\\/")));
         }
         if (Objects.nonNull(scopeId)) {
             recallQuery.and(eq(AlarmRecord.SCOPE, scopeId));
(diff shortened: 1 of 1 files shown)


oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/AlarmQuery.java
```diff
@@ -64,7 +64,7 @@ public Alarms getAlarm(Integer scopeId, String keyword, int limit, int from, lon
                        .and(lte(InfluxClient.TIME, InfluxClient.timeInterval(endTB)));
         }
         if (!Strings.isNullOrEmpty(keyword)) {
-            recallQuery.and(regex(AlarmRecord.ALARM_MESSAGE, keyword));
+            recallQuery.and(contains(AlarmRecord.ALARM_MESSAGE, keyword.replaceAll("/", "\\/")));
```
kezhenxu94:
```suggestion
            recallQuery.and(contains(AlarmRecord.ALARM_MESSAGE, keyword.replaceAll("/", "\\\\/")));
```

wu-sheng: Why four \? How many do we need?

@kezhenxu94 we may need an e2e query for alarm in the future.

kezhenxu94: > Why four ? How many do we need?
> 

`"\\"` => `\`, `\` in regexp pattern is for escaping, so `\\\\` in string literal => `"\\(escape in string literal)\\(escape in string literal)"` => `\\` (escape `\` in regexp)

kezhenxu94: > @kezhenxu94 we may need an e2e query for alarm in the future.

please create an issue to track in case we all forget

kezhenxu94: > > Why four ? How many do we need?
> 
> `"\\"` => `\`, `\` in regexp pattern is for escaping, so `\\\\` in string literal => `"\\(escape in string literal)\\(escape in string literal)"` => `\\` (escape `\` in regexp)

@wu-sheng @dmsolr if you're not sure, please refer to the JavaDoc of `java.util.regex.Matcher#replaceAll` or simply run the following test:

```java
        System.out.println("/abc/".replaceAll("/", "\\/"));
        // still produce /abc/ , not \/abc\/
        System.out.println("/abc/".replaceAll("/", "\\\\/"));
        // produce \/abc\/ , which is what we want,
    }
```

wu-sheng: Got it

PR #4541, opened 2020-03-19
Make `CommandService` daemon to avoid blocking target application shu…

### LINKED ISSUE #4496: Application will not quit after main() done

- [x] Question or discussion
- [ ] Bug
- [ ] Requirement

___
### Question
Here is the example, assume that we have a class `Hello` as defined here

```java
public class Hello {
        System.out.println("Hello, World!");
    }
}
```

after doing this

```bash
 java -javaagent:skywalking-agent.jar Hello
```
we should got `Hello, World!` at terminal stdout, and then the application should quit with status code `0` (which means process quit without any errors)

However, what I got is:
1. the stdout is ok
2. the application KEEPS RUNNING until I trigger `CTRL-C`.  

Maybe it's a normal circumstance when some non-daemon background threads are still running, but it's still quite a wired phenomenon where the application has done it's work and keeps running without any response for common developer.
___

### LINKED ISSUE #4538: Can't shutdown the Tomcat

- [x] Question or discussion
- [ ] Bug
- [ ] Requirement

___
### Question
- What do you want to know?

___
### Bug

- Which company or project?

- What happen?

___

kezhenxu94:
…tting down gracefully

closes #4538 and closes #4496

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/commands/CommandService.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/commands/CommandService.java
@@ -17,8 +17,13 @@
 
 package org.apache.skywalking.apm.agent.core.commands;
 
+import java.util.ArrayList;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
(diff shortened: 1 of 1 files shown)


kezhenxu94:
unfortunately, `CommandService` is the only one non-daemon thread, sorry about that

PR #4551, opened 2020-03-22
[Chore] Fix license headers and rename E2E job name

kezhenxu94:


kezhenxu94:
> Do e2e tests have been checked by rat?

If the rat is only for checking license headers, then the checkstyle plugin has already done this, if the rat has other functionalities, I could add it to e2e project too

https://github.com/apache/skywalking/blob/ace781ff490310de09313f19194281826f9064b0/test/e2e/pom.xml#L145
https://github.com/apache/skywalking/blob/master/apm-checkstyle/CHECKSTYLE_HEAD

wu-sheng:
I am thinking `rat` is not set right. You could run this test, change any file in the source folder(not e2e) the header to `Licensed to the acs under one or more` at the first line, run check. You could see this

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  5.188 s
[INFO] Finished at: 2020-03-22T22:28:53+08:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.0:check (validate) on project apm-datacarrier: You have 1 Checkstyle violation. -> [Help 1]
```

kezhenxu94:
> I am thinking `rat` is not set right. 

Yes it's not working, after fixing the plugin's configuration itself, the checks failed to pass, but I don't have much time to inspect how the plugin actually work(seems regexp matching), yet the license headers are now checked by the CheckStyle plugin, so we can safely believe that the license headers did not conflict with the ASF

wu-sheng:
It is mainly for header and dependency lib license chekck.

kezhenxu94:
Then I may need some time to find out why it failed

wu-sheng:
I don't see the failing and the what are the changes related to the rat. Do you push those?

PR #4555, opened 2020-03-23
Correlation protocol implement

mrproliu:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#4509 
___
- Implement Correlation protocol, java version.


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -394,4 +394,16 @@
             public static int HTTP_PARAMS_LENGTH_THRESHOLD = 1024;
         }
     }
+
+    public static class Correlation {
+        /**
+         * Max key count in the correlation context.
+         */
+        public static int KEY_COUNT = 3;
```
wu-sheng: Should be renamed to `ELEMENT_MAX_NUMBER`.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -394,4 +394,16 @@
...
+
+        /**
+         * Max value length in each key.
+         */
+        public static int VALUE_LENGTH = 128;
```
wu-sheng: Should be renamed to `VALUE_MAX_LENGTH `.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
```diff
@@ -0,0 +1,164 @@
...
+            String perDataKey = parts[0];
+            String perDataValue = parts.length > 1 ? parts[1] : "";
+            data.put(Base64.decode2UTFString(perDataKey), Base64.decode2UTFString(perDataValue));
+        }
+    }
+
+    /**
+     * Reset correlation context from other context
+     */
+    public void resetFrom(CorrelationContext context) {
```
wu-sheng: Why there is `reset` here? Why does a context require `reset`?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
```diff
@@ -0,0 +1,164 @@
...
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Correlation context, use to propagation user custom data.
+ * Working on the protocol and delegate set/get method.
+ */
+public class CorrelationContext {
+
+    private volatile Map<String, String> data;
```
wu-sheng: Why this could be changed? And further as a `volatile`? What are the cases?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
wu-sheng: Why isn't this `inject`/`extract` like the existing API design?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/IgnoredTracerContext.java
```diff
@@ -32,30 +32,33 @@
...
 
     public IgnoredTracerContext() {
         this.stackDepth = 0;
+        this.correlationContext = new CorrelationContext();
     }
 
     @Override
     public void inject(ContextCarrier carrier) {
-
+        carrier.getCorrelationContext().resetFrom(this.correlationContext);
```
wu-sheng: `inject` is for further propagation, I am confused the implementation here.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextSnapshot.java
```diff
@@ -49,12 +49,15 @@
...
-    ContextSnapshot(ID traceSegmentId, int spanId, List<DistributedTraceId> distributedTraceIds) {
+    private CorrelationContext correlationContext;
+
+    ContextSnapshot(ID traceSegmentId, int spanId, List<DistributedTraceId> distributedTraceIds, CorrelationContext correlationContext) {
         this.traceSegmentId = traceSegmentId;
         this.spanId = spanId;
         if (distributedTraceIds != null) {
             this.primaryDistributedTraceId = distributedTraceIds.get(0);
         }
+        this.correlationContext = correlationContext;
```
wu-sheng: As you are propagating the context, and this is called a snapshot, you should clone the existing context to new thread. Otherwise, this is not a `ContextSnapshot`. Think in this way, when you do the thread snapshot, what does it mean?

docs/en/setup/service-agent/java-agent/Application-toolkit-correlation.md
```diff
@@ -0,0 +1,33 @@
...
+## Introduce
+This plugin is help user to transport custom data in the tracing context. [Follow this to get protocol description.](../../../protocols/Skywalking-Cross-Process-Correlation-Headers-Protocol-v1.md)
+
+## How to use it
+* Dependency the toolkit, such as using maven or gradle
+```xml
+   <dependency>
+      <groupId>org.apache.skywalking</groupId>
+      <artifactId>apm-toolkit-trace</artifactId>
+      <version>${skywalking.version}</version>
```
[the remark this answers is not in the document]
mrproliu: I seem we have [Application-toolkit-trace.md](https://github.com/apache/skywalking/blob/259a147682e82c862f5d4af90237e099b6d38b13/docs/en/setup/service-agent/java-agent/Application-toolkit-trace.md) and [Application-toolkit-trace-cross-thread.md](https://github.com/apache/skywalking/blob/259a147682e82c862f5d4af90237e099b6d38b13/docs/en/setup/service-agent/java-agent/Application-toolkit-trace-cross-thread.md), I not sure where do I need to put this into?

wu-sheng:
NOTICE, this is an agent core level API change, we must be very carefully about all codes.

docs/en/setup/service-agent/java-agent/Application-toolkit-correlation.md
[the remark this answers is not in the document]
wu-sheng: I don't think this feature has anything related to across-thread toolkit. The across-thread thing only happens in the codes level, but for user, they don't care and can't feel it. These are just simple new magic APIs for them.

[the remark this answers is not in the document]
mrproliu: Sure, I put this into `Application-toolkit-trace.md`.

docs/en/setup/service-agent/java-agent/Application-toolkit-correlation.md
```diff
@@ -0,0 +1,33 @@
...
+   </dependency>
+```
+
+* Use `CorrelationContext.set()` API to setting custom data.
+```java
+CorrelationSettingResult settingResult = CorrelationContext.set("customKey", "customValue");
```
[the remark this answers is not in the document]
mrproliu: I think users can record this setting result, such as log this setting result,to confirm whether the setting is successful. I not pretty sure do we need that? any suggestion?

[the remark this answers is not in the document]
wu-sheng: The point is, I am not sure who(users) will set interactive codes with an observability API? And why do they do this? 
From my understanding, they just expect we wouldn't throw exception. An error seem very confusing. Same principle in `getTraceId()`. We never report the reason why `trace id` is N/A or empty string.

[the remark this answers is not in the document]
mrproliu: Changed that return the previous value only.

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -160,6 +162,7 @@ Now, we have the following known bootstrap plugins.
 [log4j2](Application-toolkit-log4j-2.x.md), [logback](Application-toolkit-logback-1.x.md)
     * If you want to use annotations or SkyWalking native APIs to read context, try [SkyWalking manual APIs](Application-toolkit-trace.md)
     * If you want to continue traces across thread manually, use [across thread solution APIs](Application-toolkit-trace-cross-thread.md).
+    * If you want to cross process correlation your custom data, use [Correlation solution APIs](Application-toolkit-correlation.md).
```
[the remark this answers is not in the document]
mrproliu: Delete this file.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
mrproliu: Change that to final, and use the `extract` and `inject` to operate the data.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/IgnoredTracerContext.java
mrproliu: Add the `inject` and `extract` into correlation context, and delete the `resetFrom` method.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextSnapshot.java
mrproliu: In this case, I neglected the interference between threads. Change that using `clone` mode.

apm-sniffer/apm-toolkit-activation/apm-toolkit-trace-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/trace/CorrelationContextGetInterceptor.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public Object afterMethod(Class clazz, Method method, Object[] allArguments, Class<?>[] parameterTypes, Object ret) {
+        return ret;
+    }
+
+    @Override
+    public void handleMethodException(Class clazz, Method method, Object[] allArguments, Class<?>[] parameterTypes, Throwable t) {
+        logger.error("Failed to get correlation value.", t);
```
wu-sheng: Remove this, we don't log unless this is a serious issue.  Can't see the case this will happen.

apm-sniffer/apm-toolkit-activation/apm-toolkit-trace-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/trace/CorrelationContextSetInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public Object afterMethod(Class clazz, Method method, Object[] allArguments, Class<?>[] parameterTypes, Object ret) {
+        return ret;
+    }
+
+    @Override
+    public void handleMethodException(Class clazz, Method method, Object[] allArguments, Class<?>[] parameterTypes, Throwable t) {
+        logger.error("Failed to setting correlation.", t);
```
wu-sheng: Same here.

docs/en/setup/service-agent/java-agent/Application-toolkit-trace.md
```diff
@@ -53,3 +53,13 @@ public User methodYouWantToTrace(String param1, String param2) {
 }
 ```
 
+* Use `CorrelationContext.set()` API to setting custom data in tracing context. Please follow [the agent configuration](README.md#table-of-agent-configuration-properties) to get more setting limit.
```
wu-sheng:
```suggestion
* Use `CorrelationContext.set()` API to setting custom data in tracing context. 
```

docs/en/setup/service-agent/java-agent/Application-toolkit-trace.md
```diff
@@ -53,3 +53,13 @@ public User methodYouWantToTrace(String param1, String param2) {
...
+```java
+optional<String> previous = CorrelationContext.set("customKey", "customValue");
+```
+
+* Use `CorrelationContext.get()` API to get custom data.
+```java
+optional<String> value = CorrelationContext.get("customKey");
+```
+
```
wu-sheng:
```suggestion
CorrelationContext configuration descriptions could be found in the [the agent configuration](README.md#table-of-agent-configuration-properties) documentation, with `correlation.` as the prefix.
```

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -125,6 +125,8 @@ property key | Description | Default |
 `plugin.tomcat.collect_http_params`| This config item controls that whether the Tomcat plugin should collect the parameters of the request. Also, activate implicitly in the profiled trace. | `false` |
 `plugin.springmvc.collect_http_params`| This config item controls that whether the SpringMVC plugin should collect the parameters of the request, when your Spring application is based on Tomcat, consider only setting either `plugin.tomcat.collect_http_params` or `plugin.springmvc.collect_http_params`. Also, activate implicitly in the profiled trace. | `false` |
 `plugin.http.http_params_length_threshold`| When `COLLECT_HTTP_PARAMS` is enabled, how many characters to keep and send to the OAP backend, use negative values to keep and send the complete parameters, NB. this config item is added for the sake of performance.  | `1024` |
+`correlation.element_max_number`|Correlation max element count in the correlation context.|`3`|
```
wu-sheng:
```suggestion
`correlation.element_max_number`| Max element count of the correlation context.|`3`|
```

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -125,6 +125,8 @@ property key | Description | Default |
...
+`correlation.value_max_length`|Correlation max value length in each key.|`128`|
```
wu-sheng:
```suggestion
`correlation.value_max_length`|Max value length of correlation context element.|`128`|
```

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -158,7 +160,7 @@ Now, we have the following known bootstrap plugins.
     * If you want to use OpenTracing Java APIs, try [SkyWalking OpenTracing compatible tracer](Opentracing.md). More details you could find at http://opentracing.io
     * If you want to print trace context(e.g. traceId) in your logs, choose the log frameworks, [log4j](Application-toolkit-log4j-1.x.md), 
 [log4j2](Application-toolkit-log4j-2.x.md), [logback](Application-toolkit-logback-1.x.md)
-    * If you want to use annotations or SkyWalking native APIs to read context, try [SkyWalking manual APIs](Application-toolkit-trace.md)
+    * If you want your codes interact with SkyWalking agent, including `getting trace id`, `set tags`, etc.. Try [SkyWalking manual APIs](Application-toolkit-trace.md).
```
wu-sheng:
```suggestion
    * If you want your codes to interact with SkyWalking agent, including `getting trace id`, `set tags`, `propagating custom data` etc.. Try [SkyWalking manual APIs](Application-toolkit-trace.md).
```

test/plugin/scenarios/apm-toolkit-trace-scenario/src/main/java/org/apache/skywalking/apm/toolkit/trace/CorrelationContext.java
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.skywalking.apm.toolkit.trace;
+
+import java.util.Optional;
+
+/**
+ * Propagate the custom data in the tracing context.
```
wu-sheng:
```suggestion
 * CorrelationContext is the interactive API for end user to put/set custom data. 
```

apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/CorrelationContext.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.skywalking.apm.toolkit.trace;
+
+import java.util.Optional;
+
+/**
+ * Propagate the custom data in the tracing context.
```
wu-sheng:
```suggestion
 * CorrelationContext is the interactive API for end user to put/set custom data. 
```

apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/CorrelationContext.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    /**
+     * Setting the custom key/value into trace context.
+     *
+     * @return previous value if it exists.
+     */
+    public static Optional<String> set(String key, String value) {
+        return Optional.empty();
+    }
```
kezhenxu94: Personally, I don't recommend to return `Optional<String>` in these two APIs, for `String`s, we(developers, our users are developers too) don't usually **JUST** care about the nullability, we also want to check whether it's empty/blank, with `Optional<String>`, we have to unwrap the `Optional`, and check `isEmpty`/`isBlank`:

```java
        Optional<String> s = CorrelationContext.get(CORRELATION_CONTEXT_KEY);
        if (s.isPresent() && Strings.isNotBlank(s.get())) {
            // ...
        }
```

OR

```java
        if (!Strings.isNullOrEmpty(CorrelationContext.get(CORRELATION_CONTEXT_KEY).orElse(""))) {
            
        }
```

My point is that in this specific situation, `Optional<String>` may not bring much convenience as it should have done, simply returning nullable `String` may do, in the contrast

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    public Optional<String> set(String key, String value) {
+        // key must not null
+        if (key == null) {
+            return Optional.empty();
+        }
+        if (value == null) {
+            value = "";
+        }
```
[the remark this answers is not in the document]
mrproliu: Do you mean delete the key when the value is null? Should I add a new method to implement the `remove` method?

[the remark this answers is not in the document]
kezhenxu94: > Do you mean delete the key when the value is null? Should I add a new method to implement the `remove` method?

Adding a new API `remove` looks good to me, but I prefer to remove the item when the `value == null`(or even when `value.isEmpty()`) because the places in the context are "precious", and propagating `null` or `""` is a waste of bandwidth, WDYT @wu-sheng

apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/CorrelationContext.java
mrproliu: You're saying to unwrap the `Optional<String>`, so we don't need to use `get`(or `orElse`)  and `Strings.isNullOrEmpty` to double-check. It's looking good to me. @wu-sheng any suggestion?

kezhenxu94: > You're saying to unwrap the `Optional<String>`, so we don't need to use `get`(or `orElse`) and `Strings.isNullOrEmpty` to double-check

Exactly

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
```diff
@@ -0,0 +1,150 @@
...
+            String perDataValue = parts.length > 1 ? parts[1] : "";
+            data.put(Base64.decode2UTFString(perDataKey), Base64.decode2UTFString(perDataValue));
+        }
+    }
+
+    /**
+     * Prepare for the cross-process propagation. Inject the {@link #data} into {@link ContextCarrier#getCorrelationContext()}
+     */
+    void inject(ContextCarrier carrier) {
+        carrier.getCorrelationContext().data.putAll(this.data);
```
wu-sheng: This is using `putAll`, I am feeling we have a memory leak point. If user to inject a header with more elements than our limitation, `inject` and `deserialize` both wouldn't reject or ignore part of them, then basically, the config of this agent is useless.

wu-sheng: We will propagate all context key/value(s) without any protection. Please fix this.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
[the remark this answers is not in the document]
wu-sheng: Agree, we should remove the element when value doesn't exist. And for `remove`, adding or not both works for me. But the logic of `null/empty` triggers `removing` should be documented clearly.

apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/CorrelationContext.java
wu-sheng: As `Optional`, the point is avoiding NPE explicitly for API user. We could guarantee there is no EMPTY_STRING as the value, as we discussed above. Then user doesn't have `doublt-check` case. We also don't ask user to do `.orElse("")` and then `Strings.isNullOrEmpty`. Your demo codes are written by yourself in this way.

I am a little confusing about your conclusion.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
mrproliu: That's true. If another agent approves to setting more than element count or value length. So we will delete their data?

wu-sheng: I think so. We should keep the memory safe as 1st priority.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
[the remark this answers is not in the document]
kezhenxu94: @mrproliu please perform a removal when `value == null || value.isEmpty()`, and document it, `remove` is unnecessary then (I tend to keep the core API as simple as possible)

apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/CorrelationContext.java
kezhenxu94: > We could guarantee there is no EMPTY_STRING as the value, as we discussed above. 

Since [we have reached a consensus](https://github.com/apache/skywalking/pull/4555#discussion_r396902387) that `null`/`""` values should trigger a removal, this concern is gone accordingly

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
wu-sheng: My mistake, this comment should target `extract` and `deserialize`, not `inject`.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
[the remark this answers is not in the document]
mrproliu: fixed. It will remove the item when the value is empty or null.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
mrproliu: fixed.

apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/CorrelationContext.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package org.apache.skywalking.apm.toolkit.trace;
+
+import java.util.Optional;
+
+/**
+ * CorrelationContext is the interactive API for end user to put/set custom data.
+ */
+public class CorrelationContext {
```
wu-sheng: There are two `CorrelationContext` in your codes. Seems strange. In the old APIs, we have `TraceContext`(for users) and `ContextManager`/`TracingContext` for internal. I suggest we fix the duplicated naming issue.

Such as for users, we could have `TraceContext#putCorrelation`. @kezhenxu94 @mrproliu WDYT?

kezhenxu94: Looks good to me

mrproliu: Sure. I could change to `TraceContext#putCorrelation` and `TraceContext#getCorrelation`, replace the `CorrelationContext` toolkit. 
I think it's the same for users, they refer to the document and use them.

mrproliu:
> It was best to commit one pull request for one feature

This pull request is only used to provide the Java implementation of the correlation context. I don't understand what you mean?

kezhenxu94:
> It was best to commit one pull request for one feature

@geektcp  It's sure better, but never mind, we do a squash when merging

docs/en/setup/service-agent/java-agent/Application-toolkit-trace.md
```diff
@@ -53,3 +53,14 @@ public User methodYouWantToTrace(String param1, String param2) {
 }
 ```
 
+* Use `TraceContext.getCorrelation()` API to put custom data in tracing context. 
+```java
+Optional<String> previous = TraceContext.getCorrelation("customKey", "customValue");
```
wu-sheng: I think this is a typo of the documentation? get with 2 parameters?

docs/en/setup/service-agent/java-agent/Application-toolkit-trace.md
```diff
@@ -53,3 +53,14 @@ public User methodYouWantToTrace(String param1, String param2) {
...
+```
+CorrelationContext will remove the item when the value is `null` or empty.
+
+* Use `TraceContext.putCorrelation()` API to get custom data.
+```java
+Optional<String> value = TraceContext.putCorrelation("customKey");
```
wu-sheng: Another typo here.

docs/en/setup/service-agent/java-agent/Application-toolkit-trace.md
mrproliu: My mistake. Write in the wrong place.

PR #4566, opened 2020-03-24
[CI] Reorganize the Plugin Tests to speed up the PR checks

kezhenxu94:
This pull request should speed up the checks as it leverages the parallelism of runners, and makes it quicker to rerun in case the tests are unstable because it splits the tests into separated  groups. Also, it adds some filtering paths to avoid unnecessary builds


kezhenxu94:
Some of the test include too many versions, we may want to omit some minor versions to speed up the checks further

wu-sheng:
Yes. We should. But, We could do some random deletions.

PR #4570, opened 2020-03-26
[Inventory Design Change] Remove endpoint register and endpoint inventory

wu-sheng:
If you are familiar with the latest OAP core, you will know, we register id for the endpoint. But this design has the following issues
1. The register has low performance as it can't be run in parral even we have clustering OAP. Even worse, in the cluster mode, the register is slower
2. The ID is designed as an integer, it could be run out in some time. We don't have good idea to restore or extend it.
3. Endpoint including the parameter(such as, inside the URI) is always an issue, and could only be improved in the agent plugin, but never could be fixed.

The new changes are
1. Use `EndpointTraffic` to replace `EndpointInventory`. EndpointTraffic is a manual metrics, not a register inventory.
2. ID register would not happen, and service_id + base64(endpoint name) will be used as the entity id.
3. EndpointTraffic also keeps the time series feature like all other metrics, so it could be removed by TTL.
4. EndpointRelation follows the changes of EndpointInevntory removal. It keeps the source-service_id + base64(source endpoint name) and dest-service_id + base64(dest endpoint name).

- The Endpoint_Traffic looks like this
image

- The Endpoint* metrics, such as endpoint_sla. It is wider than before, because of entity_id format changed.
image


oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/mysql/MySQLTableInstaller.java
```diff
@@ -111,7 +111,7 @@ protected void createIndexes(JDBCHikariCPClient client, Model model) throws Stor
             case SERVICE_INVENTORY:
             case SERVICE_INSTANCE_INVENTORY:
             case NETWORK_ADDRESS:
-            case ENDPOINT_INVENTORY:
+            case ENDPOINT_TRAFFIC:
                 createInventoryIndexes(client, model);
```
kezhenxu94: `EndpointTraffic` is not `RegisterSource` type, but creating `InventoryIndexes` needs a `sequence` column, which is defined in `RegisterSource`

wu-sheng: Removed.

PR #4574, opened 2020-03-26
[Agent] Add JDK13, JDK14 to build matrix

kezhenxu94:
Motivation:

Guarantee that the agent supports JDK13 and JDK14

Modifications:

Add JDK13 and JDK14 to build matrix

Result:

Build with JDK13 and JDK14

### DIFF
--- a/docs/en/setup/service-agent/java-agent/README.md
+++ b/docs/en/setup/service-agent/java-agent/README.md
@@ -1,5 +1,5 @@
 # Setup java agent
-1. Agent is available for JDK 8 - 12 in 7.x releases. JDK 1.6 - JDK 12 are supported in all 6.x releases [NOTICE¹](#notice)
+1. Agent is available for JDK 8 - 14 in 7.x releases. JDK 1.6 - JDK 12 are supported in all 6.x releases [NOTICE¹](#notice)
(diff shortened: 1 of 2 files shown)


wu-sheng:
Does this make the OAP running on the JDK 13/14 too?

kezhenxu94:
> Does this make the OAP running on the JDK 13/14 too?

Nope, the OAP is running in the docker built by the `skywalking/docker/oap/Dockerfile.oap`, which is only based on `openjdk:8u181-jdk-stretch`

wu-sheng:
BTW, I think you need to update the agent document about this?

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -1,5 +1,5 @@
 # Setup java agent
-1. Agent is available for JDK 8 - 12 in 7.x releases. JDK 1.6 - JDK 12 are supported in all 6.x releases [NOTICE¹](#notice)
+1. Agent is available for JDK 8 - 14 in 7.x releases. JDK 1.6 - JDK 14 are supported in all 6.x releases [NOTICE¹](#notice)
```
wu-sheng: We don't test 6.x for JDK 13/14. We should not indicate those.

PR #4587, opened 2020-03-29
OAP Core polish, especially storage level

wu-sheng:
There are a lot of `if/else` codes due to the differences of the storages, but I plan to merge them in a more elegance and consistent way.

1. Add length definition with reasonable default value.
1. `content` of `@Column` has been renamed to `storageOnly` . I add this to many fields as they are not being query in any case.
1. Merge H2 and MySQL columntype mapping back to consistent.
1. Remove `@IDColumn`.
1. Support `@QueryUnifiedIndex`.
1. Refactor the MySQL and H2 installers to use `@Column` and `@QueryUnifiedIndex` definitions to create indices automatically. But the index naming rule has been changed to `entity`_seq_`IDX`. `seq` is the Increment Interger for every entity. 
1. Support @MetricsExtension and insertOnly in the MetricsPersistentWorker worker.
1. Optimize MetricsStreamProcessor


wu-sheng:
@hanahmily @dmsolr @mrproliu @JaredTan95 @kezhenxu94 This PR is ready to review. Please feedback.

If you have more optimization idea, I am happy to know.

FYI @apache/skywalking-committers , this PR provides more elegance way for SQL storage and ES storage. But for InfluxDB, I have nothing to do. According to @dmsolr, it seems having different logic. Also, I have asked him to summarize those and we could discuss in another thread. InfluxDB is not a very production ready implementation today. I am still expecting the performance test.

wu-sheng:
Another update, if you has read my previous #4570, you may have noticed there is duplicated record in the endpoint_traffic index. In this PR, I also update the id()/hashcode()/equal() logic for it, therefore, every endpoint has an unique record, once it was deleted by TTL, it will regenerate when new request of this endpoint comes.

Notice the **_id**, no time bucket included anymore. Time bucket is only for time-related insert and TTL mechanism.

image

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/OALRuntime.java
```diff
@@ -377,10 +376,19 @@ private Class generateDispatcherClass(String scopeName,
...
+                                                                                                                      .getCanonicalName(),
+                                                                                                                  new SignatureAttribute.TypeArgument[] {
+                                                                                                                      new SignatureAttribute.TypeArgument(
+                                                                                                                          new SignatureAttribute.ClassType(
+                                                                                                                              sourceClassName))
+                                                                                                                  }
+                                                                                                              )
+                                                                                                          }
```
hanahmily: Why blank lines here?

wu-sheng: These are not blank :) Scroll to the right.

wu-sheng: I could reformat it manually a little.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/MetricsExtension.java
```diff
@@ -0,0 +1,42 @@
...
+public @interface MetricsExtension {
+    /**
+     * @return true if this metrics stream support down sampling.
+     */
+    boolean supportDownSampling();
+
+    /**
+     * @return true if this metrics doesn't support update.
+     */
+    boolean insertOnly();
```
hanahmily: We should name it to `supportUpdate` or change the comment description. They are not identical.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
```diff
@@ -55,5 +55,12 @@
     /**
      * The column is just saved, never used in query.
      */
-    boolean content() default false;
+    boolean storageOnly() default false;
```
mrproliu: I suggest the storageOnly default is `true` in the Column annotation? Because I think most of the columns are just for storage.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/MetricsExtension.java
wu-sheng: Rename `insertOnly` to `supportUpdate`. EndpointTraffic has wrong definition before.

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/OALRuntime.java
wu-sheng: Updated, should be better now.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ExtraQueryIndex.java
```diff
@@ -0,0 +1,61 @@
...
+    public ExtraQueryIndex(String mainColumn, final String[] withColumns) {
+        if (CollectionUtils.isNotEmpty(withColumns)) {
+            this.columns = new String[withColumns.length + 1];
+            for (int i = 0; i < columns.length; i++) {
+                if (i == 0) {
+                    columns[i] = mainColumn;
+                } else {
+                    columns[i] = withColumns[i - 1];
+                }
+            }
```
hanahmily: We can simplify them with `System.arraycopy`

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
wu-sheng: No, you can't. There are countless place in the codes, including the OAL generated codes. If you use this, we are facing more changes.

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/OALRuntime.java
hanahmily: LTGM

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/MetricsExtension.java
hanahmily: 👍

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ExtraQueryIndex.java
wu-sheng: Yes, adopted this method.

wu-sheng:
@hanahmily Recheck please.

PR #4589, opened 2020-03-30
[BugFix]Escape '/' on InfluxDB query

dmsolr:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

'/' is a specific symbol on InfluxDB query, so need to escape by '\/'.

- How to fix?

___


kezhenxu94:
Reminds me of the `%` in MySQL and H2, `%` is a similar in MySQL/H2 to `/` in InfluxDB, and is valid as part of the URL, right @wu-sheng ?

wu-sheng:
What is the context? I am guessing you are talking about mismatch?
1. `/` could be a part of the URI/endpoint for sure and common, so, when we are using the influxdb, it could be an issue. 
2. `%` could be a part of the URI, but not common, as we are not using encodedURI. So, the impact of MySQL would be less, but still there, yes.

This should be discussed on the mail list, I think. Far away from this PR.

PR #4593, opened 2020-03-30
[BugFix] Fix wrong H2 column type

kezhenxu94:
`MEDIUMTEXT` in H2 is CLOB in JDBC type, and casting it to `String` causes `ClassCastException`

### DIFF
--- a/oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/h2/dao/H2TableInstaller.java
+++ b/oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/h2/dao/H2TableInstaller.java
@@ -92,7 +92,7 @@ public class H2TableInstaller extends ModelInstaller {
         } else if (String.class.equals(type)) {
             return "VARCHAR(" + column.getLength() + ")";
         } else if (IntKeyLongValueHashMap.class.equals(type)) {
-            return "MEDIUMTEXT";
+            return "VARCHAR(20000)";
         } else if (byte[].class.equals(type)) {
             return "MEDIUMTEXT";
         } else {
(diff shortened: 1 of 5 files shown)


kezhenxu94:
And some trivial polishes, following up #4587

kezhenxu94:
http://www.h2database.com/html/datatypes.html


image

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/h2/dao/H2TableInstaller.java
```diff
@@ -92,7 +92,7 @@ protected String getColumnType(ModelColumn column) {
         } else if (String.class.equals(type)) {
             return "VARCHAR(" + column.getLength() + ")";
         } else if (IntKeyLongValueHashMap.class.equals(type)) {
-            return "MEDIUMTEXT";
+            return "VARCHAR(20000)";
```
wu-sheng: Why change this?

kezhenxu94: > Why change this?

image

wu-sheng: But e2e still passed? 

Then you can't change this only. You need to override this method in the MySQL installer. Otherwise, the size of row could be over-length easily.

kezhenxu94: > But e2e still passed?

https://github.com/apache/skywalking/issues/4594

wu-sheng:
Fix CI please.

kezhenxu94:
> Fix CI please.

Fixed already

PR #4595, opened 2020-03-30
Set the endpoint name length to 150, to keep safe and avoid over-length.

wu-sheng:
This time, we set the very strict rule at the javaagent and OAP backend. 
Especially, the OAP backend protects the storage.

`core/endpointNameMaxLength` = 150 as recommended. And comments as following
```yaml
core:
  selector: ${SW_CORE:default}
  default:
    ...
    # The max length of the endpoint name.
    # In the current practice, we don't recommend the length over 190.
    endpointNameMaxLength: ${SW_CORE_ENDPOINT_NAME_MAX_LENGTH:150}
```

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/EndpointRelation.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/EndpointRelation.java
@@ -20,6 +20,7 @@ package org.apache.skywalking.oap.server.core.source;
 
 import lombok.Getter;
 import lombok.Setter;
+import org.apache.skywalking.oap.server.core.CoreModule;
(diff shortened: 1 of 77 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/EndpointRelation.java
```diff
@@ -71,16 +75,19 @@ public String getEntityId() {
     @ScopeDefaultColumn.DefinedByField(columnName = "child_service_id")
     private int childServiceId;
     @Getter
-    @Setter
```
kezhenxu94: This should be reverted

wu-sheng: Yes. Added back.

PR #4598, opened 2020-03-31
fix Skywalking header protocol v3 docs.

JaredTan95:


wu-sheng:
All works happen on `8.x-core` branch. I am working on the changes.

PR #4599, opened 2020-03-31
Make 8.0.0 Core available. New protocol and register removed.

wu-sheng:
This would be a huge pull request, relating to **protocol**, agent and backend. 8.x would be an new core without back forward compatibility.
After the experiences of removing endpoint_inventory, I found out this
strategy is successful.
Especially, we totally get rid of register, so I want to do more.

*SkyWalking 8.0.0*
First, it is already unexpected for me, we have to move to 8.0.0 so
quickly, but after the discussion with @hanahmily , and thinking about this for several days, I think
we have to.

The key chances are following
1. Remove service, service instance, and network address register. The old
register protocols are totally going to be removed.
2. The agent doesn't need to do register anymore. Service name and Service
Instance name are generated by the agent itself, but the extra information,
such as IP, hostname, language, should report to backend separately.
3. Service Traffic should be added just like the endpoint traffic but keep
the time bucket as we need accurate service name in the given duration
4. Service Instance Traffic should be added too, with external information,
such as language, hostname.
5. Trace context propagation context should be changed to accept string in
service instance name, endpoint name and network address. This could ease
the agent logic, but also, requires changes in all language agent and
plugin test tool,
6. Trace report protocol requires to change too, in order to adopt the
string.
7. e2e tests have to ignore PHP and LUA at first, and remove the 6.x
compatibility test(doesn't support anymore).

The benefits we will get are
1. Don't worry about the inventory(s) that has been deleted randomly by end
users. (We received a lot of issue reports about this)
2. The upgrade could be easier erasing the whole storage and reboot the new
one. (Users don't feel comfortable about upgrade)
3. No hot-reboot case in the agent side
4. No cache of network address register information in the agent.
5. No service and service instance cache in the OAP
6. No register lock in the OAP
7. No file buffer mechanism in the OAP too, same as no register happens.

In my mind, I think this totally break upgrade is super meaningful and will
be good change. Even we break many things, they are easy to follow.
@mrproliu  I think by following this, we need
to change the collaboration header to `sw8` :) As no 7.1.0 release will
happen.

Link to mail list, https://lists.apache.org/thread.html/rda36fa8d191fc5750fc793993c69e14a917138bf97f9c2a461b72811%40%3Cdev.skywalking.apache.org%3E


wu-sheng:
The codes level key changes are

1. Set up traffic for service and service instance. Service Inventory and Instance Inventory are removed.
2. Add ServiceInstanceProperties as a new source. 
3. Remove Month as downsamping and query step.
...

wu-sheng:
@kezhenxu94 @ascrutae @arugal @heyanlong @ElderJames 

New protocol
- https://github.com/apache/skywalking/blob/8.x-core/docs/en/protocols/Skywalking-Cross-Process-Propagation-Headers-Protocol-v3.md
- https://github.com/apache/skywalking/blob/8.x-core/docs/en/protocols/Trace-Data-Protocol-v3.md

wu-sheng:
@apache/skywalking-committers , I just made the project package passed. Now we could move on the core tests.

1. I will continue working on local testing including H2, MySQL and ElasticSearch 6.
1. @dmsolr Please help on agent plugin tests and InfluxDB storage implementation.
1. @arugal Please help on go2sky project
1. @heyanlong Please help on PHP project
1. @elderjames Please help on .net project
1. @ascrutae Please help on nodejs project.
1. @kezhenxu94 Please help on testing and new e2e setup
1. @mrproliu Please help on backend testing, especially profiling and LUA agent test.
1. @JaredTan95 Please help on backend testing.
1. @hanahmily Please run tests on mesh case after the above are done.
1. @TinyAllen @Fine0830 Please remove the `month` step support when UI time selector choose more than one month. In this case, still set to days, and Max days should be 60 days. The `month` step has been removed from the GraphQL.

wu-sheng:
@BFergerson Please notice this PR, as you want to move the vert.x case in the main repo too.

We are changing the protocol, to make the backend has better performance and agent implementation easier. Read this #4595 as step one and this PR as step two, you will have more context.

kezhenxu94:
Mark it as "Ready for review" as the test cases should be fixed now, and move to next step if anyone will review the code

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/TracingContext.java
```diff
@@ -324,14 +324,7 @@ public AbstractSpan createExitSpan(final String operationName, final String remo
             exitSpan = parentSpan;
         } else {
             final int parentSpanId = parentSpan == null ? -1 : parentSpan.getSpanId();
-            if (StringUtil.isEmpty(remotePeer)) {
```
wu-sheng, line 363: This deletion actually is not related to this PR. The lazy peer initialization is another feature :)

wu-sheng:
# Summary
This pull request is open for 10 days, and I worked on this 2 weeks ago. Thanks everyone to help making this PR and tests work now.

Here are the list of changes we made in this pull request. Top 3 are the principle changes, others are following these principles.
1. New agent and mesh report protocol. 
1. New agent header protocol. 
1. Service register, instance register and network address register have been removed permanently.
1. Service traffic, instance traffic and network alias metrics are added to replace the service, instance and network address inventory.
1. Register process has been removed.
1. Metrics stream process supports insert only mode, especially for traffic entities.
1. Metrics stream process supports no-downsampling mode for traffic entities and network alias.
1. Remove all register mechanism and cache in the java agent.
1. Remove MONTH step in GraphQL query.
1. Update UI to remove MONTH step query, the max query range is 60 days now.
1. Simplify the TTL to metrics and record. And the unit has been formatted in Day unit. No specific TTL for ElasticSearch storage.
1. Buffer mechanism of trace receiver and mesh receiver has been removed due to no register.
1. New service id, instance id and endpoint id rules, including service relation, instance relation and endpoint relation id rules.
1. Java agent support `keep tracing` mode, meaning, agent generating tracing context even the backend is unconnected/unavailable.
1. Plugin test tool up to date, in order to support new protocol.
1. Plugin tests expected data files updated.
1. E2E tests updated.
1. [TBD] InfluxDB storage implementation is not available, need @dmsolr to fix later, in order to reduce the master change block by this PR.

wu-sheng:
@kezhenxu94 Are you going to do code level review?

kezhenxu94:
> @kezhenxu94 Are you going to do code level review?

I can do another round of review recently, but that’s not a blocker to merge this, it looks good to me generally, just merge it if necessary

wu-sheng:
@kezhenxu94 I am adjusting some documentations, after that, I will merge this.

PR #4636, opened 2020-04-11
Count the coverage in E2E tests and Plugin tests

kezhenxu94:


wu-sheng:
A reminder, what happens if plugin tests don't run? I assume it will affect the coverage rate.

wu-sheng:
Do we need to run them always to keep the coverage reasonable? Once this is done, I think you should close #2438 due to this should provide a very high coverage rate.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/entity/ProfileTaskCreationResult.java
```diff
@@ -16,7 +16,7 @@
  *
  */
 
-package org.apache.skywalking.oap.server.core.profile.entity;
+package org.apache.skywalking.oap.server.core.query.entity;
```
wu-sheng, line 19: I think this is mischange?

kezhenxu94: > I think this is mischange?

The GraphQL resolver will generate some classes for the query entities, which breaks the Jacoco agent, we have to exclude these classes in Jacoco, though I can have another rules to exclude this particular class (`ProfileTaskCreationResult`), but I think it belongs package `org.apache.skywalking.oap.server.core.query.entity`, as other similar query entities

wu-sheng: Good to me.

PR #4640, opened 2020-04-12
Fix bug that the agent heart beats too frequently

kezhenxu94:


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -119,7 +119,7 @@
         /**
          * service and endpoint registry check interval
          */
-        public static long APP_AND_SERVICE_REGISTER_CHECK_INTERVAL = 3;
+        public static long HEARTBEAT_PERIOD = 3;
```
wu-sheng, line 122: Please fix comment.

wu-sheng: We could set 40s, as no real register.

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -85,7 +85,7 @@ property key | Description | Default |
 `agent.operation_name_threshold `|The operationName max length, setting this value > 190 is not recommended.|`150`|
 `agent.keep_tracing`|Keep tracing even the backend is not available if this value is `true`.|`false`|
 `collector.grpc_channel_check_interval`|grpc channel status check interval.|`30`|
-`collector.app_and_service_register_check_interval`|application and service registry check interval.|`3`|
+`collector.heartbeat_period`|application heartbeat period.|`3`|
```
wu-sheng, line 88: Wrong default doc 😂

kezhenxu94: fixed

PR #4646, opened 2020-04-13
Nginx correlation e2e test

mrproliu:
- After [Nginx correlation implement](https://github.com/apache/skywalking-nginx-lua/pull/35), this PR is adding the Nginx correlation e2e test.
- If we got an illegal correlation section, then we should abandon that.


wu-sheng:
Could you explain a little about this PR? `correlation` can't be tested in the e2e from my perspective. I noticed you added the correlation header in the nginx, which should/could be read in the java. But there is no way to test this in e2e.

I think this header should only be tests as a part of plugin test. What do you think? 
FYI @kezhenxu94 , WDYT?

test/e2e/e2e-service-provider/src/main/java/org/apache/skywalking/e2e/lua/LuaController.java
```diff
@@ -36,8 +37,15 @@
...
+        TraceContext.putCorrelation("entry", "entry_value");
         final ResponseEntity<String> response = restTemplate.postForEntity(url.toURI(), null, String.class);
         return response.getBody();
     }
 
+    @PostMapping("/nginx/end/info")
+    private String nginxEnd() throws MalformedURLException, URISyntaxException {
+        return TraceContext.getCorrelation("entry").orElse("")
+            + "_" + TraceContext.getCorrelation("nginx").orElse("");
+    }
```
kezhenxu94, line 49: > correlation can't be tested in the e2e from my perspective. 

@wu-sheng maybe here is how @mrproliu he made it possible to test correlation in the E2E, get the correlation context in a service and expose them in an endpoint that can be verified in the test codes?

test/e2e/e2e-test/src/test/java/org/apache/skywalking/e2e/LuaE2E.java
```diff
@@ -92,14 +95,21 @@
...
+        trafficController(entryProvider, "/nginx/entry/info");
     }
 
     @AfterAll
     public void tearDown() {
         trafficController.stop();
     }
 
+    @RetryableTest
+    public void correlation() throws Exception {
```
kezhenxu94, line 107: Let's remove the `public` modifier, Junit 5 doesn't require this

mrproliu:
> Could you explain a little about this PR? `correlation` can't be tested in the e2e from my perspective. I noticed you added the correlation header in the nginx, which should/could be read in the java. But there is no way to test this in e2e.
> 
> I think this header should only be tests as a part of plugin test. What do you think?
> FYI @kezhenxu94 , WDYT?

I think the plugin test only tests with the java plugins, I think it cannot cross-language agent test? Do I miss something?

wu-sheng:
> I think the plugin test only tests with the java plugins, I think it cannot cross-language agent test? Do I miss something?

Currently, yes. But my point is only, this can't be tested in the e2e too, right?

test/e2e/e2e-service-provider/pom.xml
```diff
@@ -44,6 +44,11 @@
             <artifactId>h2</artifactId>
             <version>${h2.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.skywalking</groupId>
+            <artifactId>apm-toolkit-trace</artifactId>
+            <version>8.0.0-SNAPSHOT</version>
```
kezhenxu94, line 50: This will be hard to maintain, though I didn't figure out a better way

test/e2e/e2e-test/src/test/java/org/apache/skywalking/e2e/LuaE2E.java
```diff
@@ -92,14 +95,21 @@
     public void setUp() throws Exception {
         queryClient(swWebappHostPort);
 
-        trafficController(entryProvider, "/nginx/entry/info?backend=" + nginxHostPort.host() + ":" + nginxHostPort.port());
```
kezhenxu94, line 95: Remove the declaration of `nginxHostPort` (line#89) if you don't need it anymore

test/e2e/e2e-service-provider/pom.xml
kezhenxu94: Well, it also breaks the CI because we don't have this version when building this service

test/e2e/e2e-service-provider/src/main/java/org/apache/skywalking/e2e/lua/LuaController.java
mrproliu: Here my changed, `java1` -> `Nginx` -> `java2`, `java1` and `Nginx` to put the correlation data, then I get the all correlation data in `java2` and return it. So, I make a JUnit to verify the `java2` result.

wu-sheng: `/nginx/end/info` is the validated endpoint name, but the response entity wouldn't be tests. Please correct me if I am wrong.

test/e2e/e2e-test/src/test/java/org/apache/skywalking/e2e/LuaE2E.java
```diff
@@ -92,14 +95,21 @@
...
     @AfterAll
     public void tearDown() {
         trafficController.stop();
     }
 
+    @RetryableTest
+    public void correlation() throws Exception {
+        final URL url = new URL("http", entryProvider.host(), entryProvider.port(), "/nginx/entry/info");
+        final ResponseEntity<String> response = restTemplate.postForEntity(url.toURI(), trafficData, String.class);
+        Assert.assertEquals(response.getBody(), "entry_value_nginx_value");
```
wu-sheng, line 110: @mrproliu Do you mean this check?

mrproliu: Yes, there will check all the correlation data from `java2` response.

test/e2e/e2e-service-provider/pom.xml
kezhenxu94: So @mrproliu you have to copy class `org.apache.skywalking.apm.toolkit.trace.TraceContext` into e2e folder, and remove this dependency,

mrproliu: Yes, I have tried on this. But When I copy on this, the IDE will give me an alert, because it also a source directory. Should I need to add the e2e project to a new IDE window?
image

kezhenxu94: > Yes, I have tried on this. But When I copy on this, the IDE will give me an alert, because it also a source directory. Should I need to add the e2e project to a new IDE window?
> image

Copy the file out of the IDEA, i.e. in the `Finder` or `Terminal`, or open the `e2e` in a separate IDEA window and copy between the two windows (`apm` and `e2e`)

test/e2e/e2e-test/src/test/java/org/apache/skywalking/e2e/LuaE2E.java
wu-sheng: So, once the propagation fails, this e2e will fail directly. Got it.

PR #4647, opened 2020-04-13
Add `java` -> `go2sky` -> `java` e2e test case, and adapt v3 protocol

arugal:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/4618


wu-sheng:
You just add our 100th check.

test/e2e/e2e-test/docker/go/go2sky-e2e/go.sum
```diff
@@ -0,0 +1,79 @@
+cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
wu-sheng, line 1: Do we really need this file? I think you have the go2sky commitid, should have these.

arugal: > Do we really need this file?

`go.sum` should be kept.

https://golang.org/cmd/go/#hdr-Module_authentication_using_go_sum

> I think you have the go2sky commitid, should have these already, if we run go vendor or something in the CI process.

I depend on `go2sky` with the [pseudo-version](https://golang.org/cmd/go/#hdr-Pseudo_versions), which is actually the same as the `commit-id` effect.

wu-sheng: My point is, you have the go2sky commit I'd already, you should be able to regenerate all these. Otherwise, how users use our repo?

wu-sheng: I know this file is needed, but not hosted

arugal: Users can use it by adding it to [go.mod](https://github.com/arugal/skywalking/blob/5470877696289f77b40d1debb6f454417b003d27/test/e2e/e2e-test/docker/go/go2sky-e2e/go.mod#L5).

wu-sheng: I think you haven't followed my point. Let's talk in this way, one day, user came to the go2sky repo, he didn't need anything rather than standard Go env. Right? He could do git clone, And more commands to make things ready. Now, we have commit ID file, we should not maintain this file in main repo.
Is this clear now?

test/e2e/e2e-test/docker/go/Dockerfile.go
```diff
@@ -13,16 +13,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.12
+FROM golang:1.12 AS builder
+
+ARG COMMIT_HASH=fdb185d66faddad1651c18150d63bae32610f3ac
+ARG GO2SKY_REPO=${COMMIT_HASH}.tar.gz
```
kezhenxu94, line 19: `${COMMIT_HASH}.tar.gz` is not `GO2SKY_REPO`, `https://github.com/SkyAPM/go2sky/archive/${GO2SKY_REPO}` is

test/e2e/e2e-test/docker/go/Dockerfile.go
```diff
@@ -0,0 +1,37 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM golang:1.12 AS builder
+
+ARG COMMIT_HASH=fdb185d66faddad1651c18150d63bae32610f3ac
```
wu-sheng, line 18: We need to update this once go2sky updated, right?

arugal: Yes

PR #4656, opened 2020-04-15
Remove issue from contact channel, and add bilibili site

wu-sheng:
Issue should be kept as bug report and feature discussion, rather than contact channel, so I remove it.
Bilibili site should be highlighted for Chinese users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,10 +85,11 @@ Host in Beijing. Go to [demo](http://122.112.182.72:8080).
 Follow this [document](docs/en/guides/How-to-build.md).
 
 # Contact Us
-* Submit an [issue](https://github.com/apache/skywalking/issues)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -85,10 +85,11 @@ Host in Beijing. Go to [demo](http://122.112.182.72:8080).
 Follow this [document](docs/en/guides/How-to-build.md).
 
 # Contact Us
-* Submit an [issue](https://github.com/apache/skywalking/issues)
```
dmsolr, line 88: Why we need to remove this?

wu-sheng: -> Issue should be kept as bug report and feature discussion, rather than contact channel, so I remove it.

dmsolr: Got it

PR #4667, opened 2020-04-18
fix hystrix-plugin error

Indifer:
**fix hystrix- error**

This exception affects the running of the business code

CE4DDFFE-E683-4f99-ACA2-9C5C739681A0

image

ContextManager.capture(), It is possible to throw an exception


apm-sniffer/apm-sdk-plugin/hystrix-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hystrix/v1/SWExecutionHookWrapper.java
```diff
@@ -44,10 +49,15 @@ public SWExecutionHookWrapper(HystrixCommandExecutionHook actual) {
 
         EnhancedInstance enhancedInstance = (EnhancedInstance) commandInstance;
         EnhanceRequireObjectCache enhanceRequireObjectCache = (EnhanceRequireObjectCache) enhancedInstance.getSkyWalkingDynamicField();
-        if (ContextManager.isActive()) {
-            enhanceRequireObjectCache.setContextSnapshot(ContextManager.capture());
+        try {
+            if (ContextManager.isActive()) {
```
wu-sheng, line 53: @BFergerson isActive should have checked the TracingContext is there,  why NPE in the `#capture`?

@Indifer Your fix is not right, because this should not happen at all. We need to find out why. The reason is above.

Indifer: ContextManager.isActive() is judge AbstractTracerContext is not null。
image

but ContextManager.capture() may be throw ex
image
image

image

wu-sheng: With no span in the context, shouldn't have the TracingContext. That is my point.

BFergerson: @wu-sheng, I'm not familiar with Hystrix through I agree this change isn't the solution. My first thought is maybe the version of Hystrix @Indifer is using isn't supported yet. I think we would need a reproducible example to move forward.

wu-sheng: Agree. @Indifer I hope you could debug more about why this happens.

Indifer: The version of Hystrix is 1.5.11, I'm not familiar with Hystrix too, I'll debug it sometime.

zifeihan:
@Indifer do you get this error in 6x version?

Indifer:
> @Indifer do you get this error in 6x version?
yes

zifeihan:
@Indifer @wu-sheng  i also has this problem, if this version  in LTS, i will send PR for it.

wu-sheng:
> i also has this problem, if this version in LTS, i will send PR for it.

The plugin is nearly 100% compatible since the SkyWalking 5.0. The version number doesn't affect this. Send the pull request at any time when you are ready.

Indifer:
> @Indifer @wu-sheng i also has this problem, if this version in LTS, i will send PR for it.

I haven't had time to deal with it lately, and you know why

zifeihan:
hello @Indifer  , one of this problem is redisson bug, ref:https://github.com/apache/skywalking/pull/4737, my skywalking version is 6.4.0. take a  look, when "org.apache.skywalking.apm.agent.core.context.ContextManager#createExitSpan(java.lang.String,
java.lang.String)" throw exception, but the AbstractTracerContext is active with thread. next call "ContextManager.capture()" , there has exception to throw.

PR #4671, opened 2020-04-19
[Test] Resolve the problem when hard disk reaches a high watermark in ElasticSearch

kezhenxu94:


wu-sheng:
No File change detected?

kezhenxu94:
> No File change detected?

The first commit is to confirm that the failure is stably reproduced, and the second one fixes it.

FYI @dmsolr , I've removed all the images in E2E in this pull request, please remember to remove all the images in plugin tests, and any other resources that are possible to occupy disk spaces in the plugin tests, we'd rather spend some (short) time to pull images instead of suffering this kind of problems in the future, painful enough

kezhenxu94:
The disk space seems back to normal, 8.3GB/10GB available now,

PR #4672, opened 2020-04-19
Fix the issue which component Name is Unknown

ascrutae:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/config/ComponentLibraryCatalogService.java
```diff
@@ -65,7 +65,7 @@ public String getComponentName(int componentId) {
     @Override
     public String getServerNameBasedOnComponent(int componentId) {
         Integer serverComponentId = componentId2ServerId.get(componentId);
-        return serverComponentId == null ? Const.UNKNOWN : getComponentName(serverComponentId);
+        return serverComponentId == null ? getComponentName(serverComponentId) : getComponentName(serverComponentId);
```
kezhenxu94, line 68: What does it fix? If `serverComponentId == null` then `getComponentName(serverComponentId)` always throws NPE

PR #4674, opened 2020-04-19
Add Java versions 9~14 to E2E tests

kezhenxu94:


.github/workflows/e2e.jdk-versions.yaml
```diff
@@ -47,6 +47,8 @@ jobs:
           git submodule sync --recursive
           git -c protocol.version=2 submodule update --init --force --recursive --depth=1
       - name: Build Docker Image
+        env:
+          JAVA_VERSION: ${{ matrix.jdk }}
```
wu-sheng, line 51: So, agent and backend tests will run in the same JDK versions, right?

wu-sheng:
@hanahmily I think you will like this :)

.github/workflows/e2e.jdk-versions.yaml
kezhenxu94: > So, agent and backend tests will run in the same JDK versions, right?

Yes, the agent and backend should be independent in JDK versions, so just to make it simple

wu-sheng:
A lot of Tests fail. Please confirm.

wu-sheng:
@hanahmily This PR changes `bash` to `sh`, do you have any concern about this?

kezhenxu94:
> @hanahmily This PR changes `bash` to `sh`, do you have any concern about this?

@hanahmily the real concern is that I change the base docker image to a slim one

pom.xml
```diff
@@ -255,6 +255,12 @@
             <version>${lombok.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>javax.annotation</groupId>
+            <artifactId>javax.annotation-api</artifactId>
+            <version>1.3.2</version>
```
hanahmily, line 261: Add a property?

wu-sheng: Add a core level dependency for?

hanahmily:
@kezhenxu94 Greate job! except for some nits, these changes look good to me.

pom.xml
hanahmily: jdk11 has removed it from rt.jar, that's he add this dependency here.

kezhenxu94: > Add a core level dependency for?

Just to make it compile in JDK 9+, the scope is `provided` so nothing changed actually

PR #4641, opened 2020-04-12
Upgrade the InfluxDB storage-plugin to protocol V3

dmsolr:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
#4599
#4615

PR #4677, opened 2020-04-20
[test/plugin] Enhance PluginTest

dmsolr:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance
- [x] Enhancement

- Related issues

___
1. Re-balance the workload of plugintests
2. Restore testcases of MySQL, remove on  #4641 
3. Support user to runs plugin-test on debug mode. It will not remove the log files and actualData.yaml 
4. remove paramater 'parallel_run_size' and disable parallel test.


kezhenxu94:
> Re-balance the workload of plugintests

I don't think we need this and you actually cannot "rebalance" them, as we're using matrix build, all the plugin tests of each plugin are run parallelly , no mater which workflow they belong to

.github/workflows/plugins-test.0.yaml
```diff
@@ -48,9 +48,9 @@ jobs:
           - { name: 'elasticsearch-6.x-scenario', title: 'ElasticSearch 6.7.1-6.8.4 (7)' }
           - { name: 'elasticsearch-7.x-scenario', title: 'ElasticSearch 7.0.0-7.5.2 (14)' }
           - { name: 'feign-scenario', title: 'Feign 9.0.0-9.5.1 (8)' }
-          - { name: 'finagle-17.10.x-scenario', title: 'Finagle 17.10.0-20.1.0' }
-          - { name: 'finagle-6.25.x-scenario', title: 'Finagle 6.25.0-6.43.0' }
-          - { name: 'finagle-6.44.x-scenario', title: 'Finagle 6.44.0-7.1.0' }
+          - { name: 'finagle-17.10.x-scenario', title: 'Finagle 17.10.0-20.1.0(19)' }
+          - { name: 'finagle-6.25.x-scenario', title: 'Finagle 6.25.0-6.43.0 (18)' }
+          - { name: 'finagle-6.44.x-scenario', title: 'Finagle 6.44.0-7.1.0 (4)' }
```
kezhenxu94, line 53: I personally don't like these numbers `(19), (18), (4)`, somewhat pointless yet increase maintenance work, WDYT? @wu-sheng

kezhenxu94:
> > Re-balance the workload of plugintests
> 
> I don't think we need this and you actually cannot "rebalance" them, as we're using matrix build, all the plugin tests of each plugin are run parallelly , no mater which workflow they belong to

The time it costs only depends on the version numbers of the plugin test and the test itself

.github/workflows/plugins-test.0.yaml
wu-sheng: I am OK in either way. No prefer. Maintaining this is good for rebalancing only.

kezhenxu94: > I am OK in either way. No prefer. Maintaining this is good for rebalancing only.

We don't and can't rebalance them now, all plugin tests run in parallel, saying "rebalance" makes no sense at all

wu-sheng: I don't feel much objective about this word `rebalance`. From my understanding, rebalancing could target time or number of group elements. For time, yes, they are parallel, but if one fails, the number of group elements matters when we re-run.

Basically you could consider my point as `no harm at least`.

kezhenxu94: For rerunning failed tests, make sense to me now, didn’t thought about this

wu-sheng: OK. Since the concerns have be resolved, merging.

dmsolr: 1. I agree with it is hard to maintain the number of versions. And now, I still cannot sure they are correct in all.
2. Make every group have the same number of versions as far as possible. It is necessary. If not, we don't need to split them into multi groups. At least, the adjustment is harmless.

Issue #4669, opened 2020-04-19
CacheTimer complaints when disable profile in OAL


hanahmily:
Once profile relevant metrics are disabled in OAL, just like below:

```
disable(profile_task);
disable(profile_task_log);
disable(profile_task_segment_snapshot);
```
 CacheTimer will raise exceptions about the missing of index:

```
Caused by: org.elasticsearch.client.ResponseException: method [POST], host [http://elasticsearch:9200], URI [/profile_task/type/_search?typed_keys=true&ignore_unavailable=false&expand_wildcards=open&allow_no_indices=true&search_type=query_then_fetch&batched_reduce_size=512], status line [HTTP/1.1 404 Not Found]
{"error":{"root_cause":[{"type":"index_not_found_exception","reason":"no such index","resource.type":"index_or_alias","resource.id":"profile_task","index_uuid":"_na_","index":"profile_task"}],"type":"index_not_found_exception","reason":"no such index","resource.type":"index_or_alias","resource.id":"profile_task","index_uuid":"_na_","index":"profile_task"},"status":404}
                at org.elasticsearch.client.RestClient$1.completed(RestClient.java:377) ~[elasticsearch-rest-client-6.3.2.jar:6.3.2]
                at org.elasticsearch.client.RestClient$1.completed(RestClient.java:366) ~[elasticsearch-rest-client-6.3.2.jar:6.3.2]
                at org.apache.http.concurrent.BasicFuture.completed(BasicFuture.java:119) ~[httpcore-4.4.5.jar:4.4.5]
                at org.apache.http.impl.nio.client.DefaultClientExchangeHandlerImpl.responseCompleted(DefaultClientExchangeHandlerImpl.java:177) ~[httpasyncclient-4.1.2.jar:4.1.2]
                at org.apache.http.nio.protocol.HttpAsyncRequestExecutor.processResponse(HttpAsyncRequestExecutor.java:436) ~[httpcore-nio-4.4.5.jar:4.4.5]
                at org.apache.http.nio.protocol.HttpAsyncRequestExecutor.inputReady(HttpAsyncRequestExecutor.java:326) ~[httpcore-nio-4.4.5.jar:4.4.5]
                at org.apache.http.impl.nio.DefaultNHttpClientConnection.consumeInput(DefaultNHttpClientConnection.java:265) ~[httpcore-nio-4.4.5.jar:4.4.5]
[...]

PR #4682, opened 2020-04-21
fix cache update timer exception when disable profile in the backend

mrproliu:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
#4669 
___
### Bug fix
- Use `DisableRegister.INSTANCE.include` to check the profile task is enabled.


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/cache/CacheUpdateTimer.java
```diff
@@ -55,7 +57,9 @@ public void start(ModuleDefineHolder moduleDefineHolder, int ttl) {
 
     private void update(ModuleDefineHolder moduleDefineHolder) {
         updateNetAddressAliasCache(moduleDefineHolder);
-        updateProfileTask(moduleDefineHolder);
+        if (!DisableRegister.INSTANCE.include(ProfileTaskRecord.INDEX_NAME)) {
```
wu-sheng, line 60:
```suggestion
        // Profile could be disabled by the OAL script. Only load the task when it is activated.
        if (!DisableRegister.INSTANCE.include(ProfileTaskRecord.INDEX_NAME)) {
```

wu-sheng:
Fix the conflicts, please.

Issue #4377, opened 2020-02-17
Checkstyle plugin checks test codes too


kezhenxu94:
Currently, the Checkstyle plugin in `pom.xml` of the root project only checks codes in the root project and its submodules, we also need to check those codes in `test/e2e` and `test/plugin`

- reusethe same checkstyle configurations

- add Checkstyle plugin in the `test/plugin/pom.xml` and `test/e2e/pom.xml`, and add the Checkstyle goal to `.github/workflow/plugins-test.yml` and `.github/workflow/e2e.yml`

kezhenxu94:
@huangbowen521 are you going to do this :)

huangbowen521:
I will work on it

wu-sheng:
Move this to help wanted, as PR isn't updated for a long time.

vcjmhg:
@wu-sheng , is there anyone else currently working on this issue? If not I can have a try.

wu-sheng:
@vcjmhg I am not sure what is the status. You could try first.

vcjmhg:
ok

PR #4690, opened 2020-04-22
Fix profile recheck error

mrproliu:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

___
### Bug fix
- Fix cannot profile when using restful URI.
- Simplify the e2e thread dump check.
- Make the e2e provider using `test.apache.skywalking.e2e`, then `SpringMVC` plugin could be enhanced.
- Make the `http.params` tag can be overwritten.


test/e2e/e2e-service-provider/src/main/java/test/apache/skywalking/e2e/User.java
```diff
@@ -16,7 +16,7 @@
  *
  */
 
-package org.apache.skywalking.e2e;
+package test.apache.skywalking.e2e;
```
kezhenxu94, line 19: What does the renaming fix? If it really fixes something, please add the directory to

https://github.com/apache/skywalking/blob/8e6a527f78108c7a466ba2dd87eaf5c47d1765cc/tools/coverage/report.sh#L32-L34

i.e. `sudo rm -rf "${JACOCO_HOME}"/classes/"${exec_data}"/test || true `

kezhenxu94: So the package name rules need to be updated too, is the renaming a must?

mrproliu: Use the `org.apache.skywalking` package prefix, then the `SpringMVC` plugin cannot be enhanced, because that package is ignored. So avoid using this package.

mrproliu: @kezhenxu94  I seem the e2e also has the style check, and It's must match the package name with regex, `^org\.apache\.skywalking(\.[a-zA-Z][a-zA-Z0-9]*)+$`, so how can I fix that?

kezhenxu94: > Use the `org.apache.skywalking` package prefix, then the `SpringMVC` plugin cannot be enhanced, because that package is ignored. So avoid using this package.

Then you need to:

- update the package name rules:

https://github.com/apache/skywalking/blob/dcd66ee6e940bdc9a7ecf90415929bfd48a3a284/apm-checkstyle/checkStyle.xml#L99

- add the package to exclusion list

https://github.com/apache/skywalking/blob/8e6a527f78108c7a466ba2dd87eaf5c47d1765cc/tools/coverage/report.sh#L32-L34

i.e. `sudo rm -rf "${JACOCO_HOME}"/classes/"${exec_data}"/test || true `

wu-sheng: I think test.apache.skywalking package is used in the spring plugin test too.

mrproliu: > I think test.apache.skywalking package is used in the spring plugin test too.

Yes.

mrproliu: > > Use the `org.apache.skywalking` package prefix, then the `SpringMVC` plugin cannot be enhanced, because that package is ignored. So avoid using this package.
> 
> Then you need to:
> 
> * update the package name rules:
> 
> https://github.com/apache/skywalking/blob/dcd66ee6e940bdc9a7ecf90415929bfd48a3a284/apm-checkstyle/checkStyle.xml#L99
> 
> * add the package to exclusion list
> 
> https://github.com/apache/skywalking/blob/8e6a527f78108c7a466ba2dd87eaf5c47d1765cc/tools/coverage/report.sh#L32-L34
> 
> i.e. `sudo rm -rf "${JACOCO_HOME}"/classes/"${exec_data}"/test || true `

I have changed, please take a look.

apm-checkstyle/checkStyle.xml
```diff
@@ -96,7 +96,7 @@
             <property name="format" value="(^[A-Z][A-Z0-9]*(_[A-Z0-9]+)*$)"/>
         </module>
         <module name="PackageName">
-            <property name="format" value="^org\.apache\.skywalking(\.[a-zA-Z][a-zA-Z0-9]*)+$"/>
+            <property name="format" value="^(org|test)\.apache\.skywalking(\.[a-zA-Z][a-zA-Z0-9]*)+$"/>
```
wu-sheng, line 99: Question, in spring 4.1.x test case, we have this package name, so why we change this now? Does only e2e follow this style file, but plugin tests not?

wu-sheng: @kezhenxu94 @dmsolr

kezhenxu94: > Question, in spring 4.1.x test case, we have this package name, so why we change this now? Does only e2e follow this style file, but plugin tests not?

Yes obviously, there is checkstyle plugin in E2E `pom.xml`, but not in the plugin tests,

wu-sheng: @dmsolr Please take care of plugin test codes when you have time.

dmsolr: PluginTests don't enable checkstyle yet. ref #4377

mrproliu:
> Well, forgot that if you change the package name, all the existing expected data will be broke, can you just move the class you need into the new package, and leave the others intact?

mrproliu:
@wu-sheng @kezhenxu94 I resolved the issue, please take a look. Thanks.

PR #4694, opened 2020-04-23
Provide Meter(metrics) system 

wu-sheng:
Hi dev

This pull request introduces the meter system into the SkyWalking core. Meter system is compatible with any metrics system in theory. It shares the same streaming calculation core and still pluggable to different protocols and different metrics read modes(pull/push, grpc/http/mq).

Here are the basic and simple usages of the MeterSystem APIs.

## Meter Creation
New meter could be created based on metrics name, function, scope and data type. 
1. Metrics name is still the storage entity name.
1. Functions are every similar with the OAL function, just accepting different input.
```java
final MeterSystem meterSystem = MeterSystem.meterSystem(getManager());
meterSystem.create("test_long_metrics", "avg", ScopeType.SERVICE, Long.class);
```
NOTICE, this creation should only be called in the `module#prepare` stage, otherwise, `Can't create new metrics anymore` exception will be raised after the **CORE** module `start` stage finished. You may find out, there is a chance `create` could be executed successfully in your own `module#start` stage, but it it just because of the sequence of provider loaded by the class loader, **no guarantee, so please don't do that**.

## Runtime Calculation
`AcceptableValue` is the object created at the runtime to accept new metrics value.
```java
         final MeterSystem service = getManager().find(CoreModule.NAME).provider().getService(MeterSystem.class);
        Executors.newSingleThreadScheduledExecutor().scheduleAtFixedRate(new Runnable() {
            @Override
            public void run() {
                final AcceptableValue<Long> value = service.buildMetrics("test_long_metrics", Long.class);
                value.accept(MeterEntity.newService("abc"), 5L);
                value.setTimeBucket(TimeBucket.getMinuteTimeBucket(System.currentTimeMillis()));
                service.doStreamingCalculation(value);
            }
        }, 2, 2, TimeUnit.SECONDS);
```

## Meter Functions
Right now, only `avg` function has been implemented. I submit this PR as soon as possible to get your feedback. I will add more functions.


wu-sheng:
@mrproliu @JaredTan95 @arugal Does any of you what to add a meter/metrics toolkit package for metrics collection? Also take part in this PR about the backend feature.

oap-server/server-fetcher-plugin/prometheus-fetcher-plugin/src/main/java/org/apache/skywalking/oap/server/fetcher/prometheus/provider/PrometheusFetcherProvider.java
```diff
@@ -0,0 +1,87 @@
...
+        Executors.newSingleThreadScheduledExecutor().scheduleAtFixedRate(new Runnable() {
+            @Override
+            public void run() {
+                final AcceptableValue<Long> value = service.buildMetrics("test_long_metrics", Long.class);
+                value.accept(MeterEntity.newService("abc"), 5L);
+                value.setTimeBucket(TimeBucket.getMinuteTimeBucket(System.currentTimeMillis()));
+                service.doStreamingCalculation(value);
+            }
+        }, 2, 2, TimeUnit.SECONDS);
+    }
```
dmsolr, line 81: Is FetchProvider a demonstration?
The Metrics are reported by Prometheus client or SkyWalking Agent.

wu-sheng: For demo, yes. Prometheus doesn't have client, OAP fetchs data from target instrumented applications.

wu-sheng:
> need comment!

@geektcp What do you mean?

docs/en/setup/backend/backend-fetcher.md
```diff
@@ -0,0 +1,13 @@
+# Open Fetcher
+Fetcher is a concept in SkyWalking backend. It uses pulling mode rather than [receiver](backend-receivers.md), which
+read the data from the target systems. This mode is typically in some metrics SDKs, such as Prometheus.
+
+## Prometheus Fetcher
+```yaml
+prometheus-fetcher:
+  selector: ${SW_PROMETHEUS_FETCHER:default}
```
JaredTan95, line 8: BTW, I'm a little confused about the "selectors", which scenariodos do we need to provide our own Prometheus fetcher provider? 🤔️

wu-sheng: What do you mean? I think in any case, we need our own fecthcer, otherwise, who will fetch the data?

wu-sheng:
1. Make slow trace query available in the sampled record. Logically, they are the same thing. The UI doesn't need to concern about the trace as a special case.
1. Endpoint dependency will be removed from the dashboard. Because no query available for it. We will provide a new page for that in 8.1, according to @Fine0830 
1. Comparison page will be removed due to dashboard is powerful enough to replace it.

PR #4706, opened 2020-04-24
Modification Plugin-test.md

arugal:
https://github.com/apache/skywalking/pull/4624#issuecomment-618963765

PR #4708, opened 2020-04-24
Update Plugin-test.md

BFergerson:
Made it a bit more obvious entryService is GET and healthCheck is HEAD.


wu-sheng:
@BFergerson How about review #4706, I think more changes happen there.

PR #4719, opened 2020-04-26
Support histogram function in Meter system.

wu-sheng:
## Histogram Function
Histogram includes data range buckets and the amount matched/grouped in the buckets.
This is for original histogram graph visualization

## API Usage
```java
final MeterSystem service = getManager().find(CoreModule.NAME).provider().getService(MeterSystem.class);
// Histogram Example
final AcceptableValue<BucketedValues> histogramMetrics = service.buildMetrics(
                        "test_histogram_metrics", BucketedValues.class);
 value.setTimeBucket(TimeBucket.getMinuteTimeBucket(System.currentTimeMillis()));
histogramMetrics.accept(servEntity, new BucketedValues(
                        new long[] {
                            0,
                            50,
                            100,
                            250
                        },
                        new long[] {
                            0,
                            4,
                            10,
                            10
                        }
                    ));
service.doStreamingCalculation(histogramMetrics);
```

## Constraint Condition
The time bucket should be set completely at the first place and immutable. Once it is changed, 
1. If in the `AcceptableValue` level, `IllegalArgumentException` raised for blocking further calculation.
1. If in the internal `combine` level, such as memory-db combine, warning logs show up, and ignore the combine operation. 

Don't change the bucket in any case. If the bucket is really changed, the data will be saved in next time bucket, such as next minute/hour/day based on the precision.

___
NOTICE, Percentile/Quantiles function will be provided in a similar idea of the `histogram` function,  buckets and ranks should not be changed once they are set.


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/BucketedValues.java
```diff
@@ -0,0 +1,61 @@
...
+ * BucketedValues represents a value set, which elements are grouped by time bucket.
+ */
+@ToString
+@Getter
+public class BucketedValues {
+    /**
+     * The element in the buckets represent the minimal value of this bucket, the max is defined by the next element.
+     * Such as 0, 10, 50, 100 means buckets are [0, 10), [10, 50), [50, 100), [100, infinite+).
+     */
+    private long[] buckets;
```
hanahmily, line 36: How to set up infinite lower bound? for example, [infinite-, 0)

wu-sheng: I don't support this in this function. Do we need it now? I expect another function for that.

hanahmily: What I need it is like 
```
[-infinite, -50), [-50,-10),[-10,0),[0,10),[10,+infinite]
```
If you intend to implement it in another function, this one seems not necessary, right?

wu-sheng:
@hanahmily I added the `infinite-` supported. FYI @Fine0830 

## Set Infinite negative as the bucket
The Integer#MIN_VALUE could be the first bucket element to indicate there is no minimum value.

## Usage
```java
                    // Histogram Example
                    final AcceptableValue<BucketedValues> histogramMetrics = service.buildMetrics(
                        "test_histogram_metrics", BucketedValues.class);
                    value.setTimeBucket(TimeBucket.getMinuteTimeBucket(System.currentTimeMillis()));
                    histogramMetrics.accept(servEntity, new BucketedValues(
                        new int[] {
                            Integer.MIN_VALUE,
                            0,
                            50,
                            100,
                            250
                        },
                        new long[] {
                            3,
                            1,
                            4,
                            10,
                            10
                        }
                    ));
```

## Demonstration
```graphql
{
  readHeatMap(condition : {
    name: "test_histogram_metrics",
    entity: {
      scope: Service,
      serviceName: "mock_service",
      normal: true,
    }
  }, duration: {step:MINUTE, start: "2020-04-27 1130", end: "2020-04-27 1135"}
  ) {
    values {
      id,
      values
    } 
    buckets {
      min,
      max
    }
  }
}
```

```graphql
{
  "data": {
    "readHeatMap": {
      "values": [
        {
          "id": "202004271130_bW9ja19zZXJ2aWNl.1",
          "values": [
            0,
            0,
            0,
            0,
            0
          ]
        },
        {
          "id": "202004271131_bW9ja19zZXJ2aWNl.1",
          "values": [
            0,
            0,
            0,
            0,
            0
          ]
        },
        {
          "id": "202004271132_bW9ja19zZXJ2aWNl.1",
          "values": [
            366,
            122,
            488,
            1220,
            1220
          ]
        },
        {
          "id": "202004271133_bW9ja19zZXJ2aWNl.1",
          "values": [
            0,
            0,
            0,
            0,
            0
          ]
        },
        {
          "id": "202004271134_bW9ja19zZXJ2aWNl.1",
          "values": [
            0,
            0,
            0,
            0,
            0
          ]
        },
        {
          "id": "202004271135_bW9ja19zZXJ2aWNl.1",
          "values": [
            0,
            0,
            0,
            0,
            0
          ]
        }
      ],
      "buckets": [
        {
          "min": "infinite-",
          "max": "0"
        },
        {
          "min": "0",
          "max": "50"
        },
        {
          "min": "50",
          "max": "100"
        },
        {
          "min": "100",
          "max": "250"
        },
        {
          "min": "250",
          "max": "infinite+"
        }
      ]
    }
  }
}
```

wu-sheng:
Additional this is how histogram storage in the elasticsearch

image

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/BucketedValues.java
wu-sheng: Added.

PR #4735, opened 2020-04-29
Optimize L1 and L2 aggregation.

wu-sheng:
- L1 Aggregation. Use **MergableBufferedData** to replace **ReadWriteSafeCache**. This could avoid the lock for every time aggregation of L1.
- L2 Aggregation. Support **batch mode** in the **MergableBufferedData**, also reduce the lock and lock race for the aggregation of L2.
- Remove the ** endOfBatch** from the ** StreamData**. With the batch mode supported in the workers, we don't need that. We take advantage of DataCarrier batch consuming feature.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/data/ReadWriteSafeCache.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/data/ReadWriteSafeCache.java
@@ -50,10 +50,29 @@ public class ReadWriteSafeCache<T> {
         lock = new ReentrantLock();
     }
 
+    /**
+     * Write the into the {@link #writeBufferPointer} buffer.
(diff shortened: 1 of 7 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/data/ReadWriteSafeCache.java
```diff
@@ -59,6 +64,20 @@ public void write(T data) {
         }
     }
 
+    /**
+     * Write the collection of data into the {@link #writeBufferPointer} buffer.
+     *
+     * @param data to enqueue.
+     */
+    public void write(List<T> data) {
```
dmsolr, line 72: Is it ambiguous of method name? Perhaps `writeTo` is better.

PR #4624, opened 2020-04-08
Vertx plugin test re-implementation

BFergerson:
Having trouble with the tests failing the health check even though I can access them when I bash into the container, like so:
```sh
$ docker exec -it vertx-web-3.x-scenario-3.7.0-local bash
root@8ceaf8677392:/usr/local/skywalking/scenario# curl http://localhost:8080/vertx-web-3-scenario/case/healthCheck
Success 
```

Hoping someone can take a look

Issue #4609, opened 2020-04-05
vert.x plugin can not boot 


zohar-soul:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ x ] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS and JRE?
skywalking 6.6, OS debian 9 , jre openjdk version "11.0.6" 
- Which company or project?

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.
with vert.x 3.8.5, vertx httpclient and HttpServerResponse can't start normally.

`ERROR 2020-04-05 15:15:09:096 main ConstructorInter : ConstructorInter failure. 
java.lang.ClassCastException: class io.vertx.core.net.impl.SocketAddressImpl cannot be cast to class java.lang.String (io.vertx.core.net.impl.SocketAddressImpl is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap')
        at org.apache.skywalking.apm.plugin.vertx3.HttpClientRequestImplEndInterceptor.onConstruct(HttpClientRequestImplEndInterceptor.java:46)
        at org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ConstructorInter.intercept(ConstructorInter.java:65)
        at io.vertx.core.http.impl.HttpClientRequestImpl.<init>(HttpClientRequestImpl.java)
        at io.vertx.core.http.impl.HttpClientImpl.createRequest(HttpClientImpl.java:1108)
        at io.vertx.core.http.impl.HttpClientImpl.createRequest(HttpClientImpl.java:1074)
        at io.vertx.core.http.impl.HttpClientImpl.request(HttpClientImpl.java:606)
        at io.vertx.ext.web.client.impl.HttpContext.handlePrepareRequest(HttpContext.java:324)
        at io.vertx.ext.web.client.impl.HttpContext.execute(HttpContext.java:283)
        at io.vertx.ext.web.client.impl.HttpContext.next(HttpContext.java:270)
        at io.vertx.ext.web.client.impl.predicate.PredicateInterceptor.handle(PredicateInterceptor.java:69)
        at io.vertx.ext.web.client.impl.predicate.PredicateInterceptor.handle(PredicateInterceptor.java:32)
        at io.vertx.ext.web.client.impl.HttpContext.next(HttpContext.java:267)
[...]

PR #4736, opened 2020-04-29
Increase Vert.x Compatability

BFergerson:
- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues
#4609 

___
#4624 just added the Vert.x tests back to the main repository. Those test weren't entirely correct as they were missing segments but I just wanted to get them back in the repo and fix in another PR. This is that PR. This PR also increases the compatibility for Vert.x Eventbus/Web to 3.9.0.


apm-sniffer/apm-sdk-plugin/vertx-plugins/vertx-core-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/vertx3/HttpClientRequestImplEndInterceptor.java
```diff
@@ -40,11 +40,17 @@ public void onConstruct(EnhancedInstance objInst, Object[] allArguments) {
         String host;
         int port;
         if (allArguments[3] instanceof Integer) {
+            //ver. 3.0.x - 3.3.x
```
wu-sheng, line 43: This is an HTTP interceptor, I think `instanceof` is acceptable. Only remind you, using witness class to different versions is better. Because it has no payload at the runtime, even in the JDK8 `instanceof` is faster, but at least nothing is faster :)

BFergerson: Wouldn't that mean creating three `HttpClientRequestImplEndInterceptor` classes? One for each version range

dmsolr: Yeap. If they have a right witness class in each version range.

wu-sheng: You could put three sub interceptors, as the difference is onlu about onConstruct. WDYT?

BFergerson:
@wu-sheng, the test that's failing is from another plugin. Is there anything I need to do to resolve this?

wu-sheng:
I am rerunning them, should be nothing related to this PR. Just unstable tests.

wu-sheng:
@dmsolr Please reorganize the commit log as this PR has a long commit logs, not good for squash merge.

@BFergerson I think you missed the supported list doc update? You added the new versions, I haven't checked whether should update that doc.

BFergerson:
@wu-sheng, supported list doc already mentions support for 3.x. Now it's even more true

dmsolr:
Sorry. I got

PR #4738, opened 2020-05-03
Fix nested query and type conversion in MySQL, remove unused codes

kezhenxu94:


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/type/TopNEntity.java
```diff
@@ -38,6 +38,6 @@ public TopNEntity() {
     public TopNEntity(SelectedRecord record) {
         this.name = record.getName();
         this.id = record.getId();
-        this.value = Long.parseLong(record.getValue());
+        this.value = Double.valueOf(record.getValue()).longValue();
```
kezhenxu94, line 41: The codes below returns something like "4010.00" even if the value is a `bigint` in MySQL, causing `Long.parseLong` to throw exception

https://github.com/apache/skywalking/blob/0e4e4b888f0f1edac5ab6ae5bef5e860aa14f20b/oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/h2/dao/H2AggregationQueryDAO.java#L82

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/h2/dao/H2AggregationQueryDAO.java
```diff
@@ -68,7 +68,7 @@ public H2AggregationQueryDAO(JDBCHikariCPClient h2Client) {
             });
         }
         sql.append(" group by ").append(Metrics.ENTITY_ID);
-        sql.append(") order by value ")
+        sql.append(")  as T order by value ")
```
kezhenxu94, line 71: MySQL needs an alias in nested query

wu-sheng: Is alias Working in H2?

kezhenxu94: > Is alias Working in H2?

yes

wu-sheng:
Your branch can't be updated automatically.

apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumerPoolFactory.java
```diff
@@ -48,43 +47,4 @@ public ConsumerPool get(String poolName) {
         return pools.get(poolName);
     }
 
-    /**
-     * Default pool provides the same capabilities as DataCarrier#consume(IConsumer, 1), which alloc one thread for one
-     * DataCarrier.
-     */
-    public static final ConsumerPool DEFAULT_POOL = new ConsumerPool() {
```
wu-sheng, line 55: Doesn't agent use this API, right?

kezhenxu94: > Doesn't agent use this API, right?

Yes, never used

PR #4739, opened 2020-05-03
Fix the cache unclear bug.

wu-sheng:
The cache should be read-once. I forgot to support this in the new implementation, the aggregation will be inaccurate and keep growing. This would burn the CPU and storage.

This is a very serious bug.


wu-sheng:
> not yet ready

Please recheck the data. The e2e shows the data is aggregated. and from your DM screenshot, the precision is hour level, which means your CPM could really be 0, or at least you need 1 * 60 for the one hour slot.

Issue #3907, opened 2019-11-20
MariaDB doesn't appear in the topology view and is completely not recognized by SkyWalking


TigerIcecold:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [x] Question or discussion
- [x] Bug
- [ ] Requirement
- [x] Feature or performance improvement

___
### Question
- What do you want to know?
I'd like to monitor my Liferay Portal CE 7.2 GA1 instance with Apache SkyWalking 6.5.0. Liferay is a Java CMS. On that DEV environment following software is installed: Ubuntu 18.04.3 LTS, OpenJDK 11.0.4, Apache Tomcat 9.0.17 and MariaDB 10.1.41 as Liferay storage.

In than configuration the MariaDB database is not recognized by SkyWalking completely.

On a different VM with Ubuntu 19.10, Azul Zulu JDK 11.0.5 and MySQL 5.7.28 Skywalking recognizes the DB just fine. On that environment the DB is a different server, on the first one everything is on one machine.

Is that a bug or SkyWalking just doesn't work with MariaDB and it's a missing Feature? Or coud it be a configuration issue? (Config is on both environments is similar).

Thank you.

___
### Bug
- Which version of SkyWalking, OS and JRE?
SkyWalking 6.5.0, Ubuntu 18.04.3 LTS, OpenJDK 11.0.4, Apache Tomcat 9.0.17 and MariaDB 10.1.41
- Which company or project?

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.
SkyWalking doesn't "see" the MariaDB database of the monitored application.
___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

wu-sheng:
Do you read the supported-list doc? https://github.com/apache/skywalking/blob/master/docs/en/setup/service-agent/java-agent/README.md#supported-middleware-framework-and-library. SkyWalking's auto-detection is totally based on variable plugins. I am assuming the MariaDB has its own driver, right?

Could you try to follow MySQL driver plugin and contribute one to us?

TigerIcecold:
I have migrated the my DB from MariaDB to MySQL. Now tracing works.

wu-sheng:
I see. I still hope someone could contribute MariaDB driver plugin

wu-sheng:
@TigerIcecold The plugin has been provided.

PR #4740, opened 2020-05-03
Mariadb JDBC plugin

efekaptan:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#3907 

___
- A JDBC driver plugin implemented for Mariadb. As Mariadb is based on the Mysql project, there are similarities between two drivers. 
- Mariadb uses the Mysql url format so I re-used the `MysqlURLParser` with a small modification.
- PreparedStatement parameter string building logic is exactly the same between Mysql, PostgreSql and Mariadb. I created a parameter builder class to re-use between these plugins. That's why I touched a little bit to Postgresql and Mysql plugins and I checked the unit tests and scenarios to be sure to not break anything.
- The latest minor versions of the the driver(2.x) are included in the test scenario.

image

### DIFF
--- a/apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
+++ b/apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
@@ -156,4 +156,6 @@ public class ComponentsDefine {
     public static final OfficialComponent UNDERTOW = new OfficialComponent(84, "Undertow");
 
     public static final OfficialComponent FINAGLE = new OfficialComponent(85, "Finagle");
+
+    public static final OfficialComponent MARIADB = new OfficialComponent(86, "Mariadb");
 }
--- a/apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
+++ b/apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
@@ -157,5 +157,5 @@ public class ComponentsDefine {
 
     public static final OfficialComponent FINAGLE = new OfficialComponent(85, "Finagle");
 
-    public static final OfficialComponent MARIADB = new OfficialComponent(86, "Mariadb");
+    public static final OfficialComponent MARIADB_JDBC = new OfficialComponent(87, "mariadb-jdbc");
 }
--- a/docs/en/setup/service-agent/java-agent/README.md
+++ b/docs/en/setup/service-agent/java-agent/README.md
@@ -125,6 +125,8 @@ property key | Description | Default |
 `plugin.http.http_params_length_threshold`| When `COLLECT_HTTP_PARAMS` is enabled, how many characters to keep and send to the OAP backend, use negative values to keep and send the complete parameters, NB. this config item is added for the sake of performance.  | `1024` |
 `correlation.element_max_number`|Max element count of the correlation context.|`3`|
 `correlation.value_max_length`|Max value length of correlation context element.|`128`|
+`plugin.mariadb.trace_sql_parameters`|If set to true, the parameters of the sql (typically `java.sql.PreparedStatement`) would be collected.|`false`|
+`plugin.mariadb.sql_parameters_max_length`|If set to positive number, the `db.sql.parameters` would be truncated to this length, otherwise it would be completely saved, which may cause performance problem.|`512`|
 
 ## Optional Plugins
 Java agent plugins are all pluggable. Optional plugins could be provided in `optional-plugins` folder under agent or 3rd party repositories.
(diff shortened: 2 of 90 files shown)


oap-server/server-bootstrap/src/main/resources/component-libraries.yml
```diff
@@ -287,6 +287,9 @@ Undertow:
 Finagle:
   id: 85
   languages: Java,Scala
+Mariadb:
```
wu-sheng, line 290: I think you need a logo PR at the rocketbot UI repo.

One question, does `Mariadb` or `Mariadb-JDBC` make more sense? if `Mariadb-JDBC`, then should add mapping to `Component-Server-Mappings:`

apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
```diff
@@ -156,4 +156,6 @@
     public static final OfficialComponent UNDERTOW = new OfficialComponent(84, "Undertow");
 
     public static final OfficialComponent FINAGLE = new OfficialComponent(85, "Finagle");
+
+    public static final OfficialComponent MARIADB = new OfficialComponent(86, "Mariadb");
```
wu-sheng, line 160: I think in Java, this should be `Mariadb-jdbc` or something indicating we instrumenting the client driver only.

arugal:
test case good to me

oap-server/server-bootstrap/src/main/resources/component-libraries.yml
efekaptan: Ok, I agree. Other JDBC drivers are also using similar naming convention, will refactor it.

efekaptan:
Hi, I couldn't succeed to receive the `TopNRecords` from the server. I couldn't receive for Mysql driver either. During my investigation I noticed that below line doesn't return any value:

https://github.com/apache/skywalking/blob/master/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/PersistenceWorker.java#L73

TopNWorker can't save any record. Am i missing some point? Can I get your feedback about this situation?

wu-sheng:
Submit #4741 to fix the thing you reported, thanks.

efekaptan:
@wu-sheng I tried again after your fix but unfortunately `TOP_N_DATABASE_STATEMENT` table remains empty and can't get any topN result from the server. I have also decreased the `topNReportPeriod` parameter to 1 minute interval.

I will compare with the demo application, would like to see if the problem is related to my setup or not.

arugal:
> @wu-sheng I tried again after your fix but unfortunately `TOP_N_DATABASE_STATEMENT` table remains empty and can't get any topN result from the server. I have also decreased the `topNReportPeriod` parameter to 1 minute interval.
> 
> I will compare with the demo application, would like to see if the problem is related to my setup or not.

@efekaptan I have updated the master code for you, please pull the code and try again.

wu-sheng:
@efekaptan The easiest way, use the `AgentDataMock` in the source codes, it could generate the slow SQL. Then you could see it on the UI. This is how I tested in the morning. :)

wu-sheng:
BTW, this is how the slow SQL generated, https://github.com/apache/skywalking/blob/44cae97c259c06f95d1fc4b78ee41268670b6542/oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/parser/listener/MultiScopesAnalysisListener.java#L188-L222.

efekaptan:
Thanks for the information, let me try to debug and understand the reason behind it.

efekaptan:
After decreasing `slowDBAccessThreshold`, I was able to receive slow sqls from the server, thanks for the hint. 

- I refactored component name as `mariadb-jdbc`.
- I am attaching a sample database dashboard screenshot.

Screenshot 2020-05-06 at 19 49 24

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -125,6 +125,8 @@ property key | Description | Default |
 `plugin.http.http_params_length_threshold`| When `COLLECT_HTTP_PARAMS` is enabled, how many characters to keep and send to the OAP backend, use negative values to keep and send the complete parameters, NB. this config item is added for the sake of performance.  | `1024` |
 `correlation.element_max_number`|Max element count of the correlation context.|`3`|
 `correlation.value_max_length`|Max value length of correlation context element.|`128`|
+`plugin.mariadb.trace_sql_parameters`|If set to true, the parameters of the sql (typically `java.sql.PreparedStatement`) would be collected.|`false`|
+`plugin.mariadb.sql_parameters_max_length`|If set to positive number, the `db.sql.parameters` would be truncated to this length, otherwise it would be completely saved, which may cause performance problem.|`512`|
```
wu-sheng, line 129: Please move these two just below the existing database config. Let's keep the database config in a same place

test/plugin/scenarios/mariadb-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,136 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+segmentItems:
+- serviceName: mariadb-scenario
+  segmentSize: ge 2
+  segments:
+  - segmentId: not null
+    spans:
+    - operationName: /mariadb-scenario/case/healthCheck
+      operationId: 0
```
wu-sheng, line 23: OperationId isn't required, this is a legacy thing, please remove all.

wu-sheng:
The logo is for topology, you should see a question mark to represent the database node currently, yes?

docs/en/setup/service-agent/java-agent/README.md
efekaptan: Done.

test/plugin/scenarios/mariadb-scenario/config/expectedData.yaml
efekaptan: Done

PR #4741, opened 2020-05-05
Fix two serious bugs in new persistent

wu-sheng:
I found these two issues based on @efekaptan 's https://github.com/apache/skywalking/pull/4740#issuecomment-623763022 reports.

1. TopN Query is not working since the 8.0 storage re-implementation. FYI @kezhenxu94, another thing we don't cover in the e2e.
1. Metrics are not calculated for the first time, we only can get the values in update process. Easy fix but hard to find :P


wu-sheng:
@ascrutae please update the demo, let's see whether everything is back to normal.

PR #4745, opened 2020-05-05
Support multiple ways to analyze segments in rest

mrproliu:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

___
- Support multiple ways to analyze segments in rest:
  - "/v3/segment": single segment json
  - "/v3/segments": segment json array


oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/handler/v8/rest/TraceSegmentReportListJsonServletHandler.java
```diff
@@ -0,0 +1,75 @@
...
+            SegmentObject.Builder upstreamSegmentBuilder = SegmentObject.newBuilder();
+            ProtoBufJsonUtils.fromJSON(element.toString(), upstreamSegmentBuilder);
+            segments.add(upstreamSegmentBuilder.build());
+        }
+        return segments;
+    }
+
+    @Override
+    public String pathSpec() {
+        return "/v3/segmentList";
```
kezhenxu94, line 73:
```suggestion
        return "/v3/segments.json";
```

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/handler/v8/rest/TraceSegmentReportMultipleLineServletHandler.java
```diff
@@ -0,0 +1,60 @@
...
+            ProtoBufJsonUtils.fromJSON(line, upstreamSegmentBuilder);
+            segments.add(upstreamSegmentBuilder.build());
+        }
+
+        return segments;
+    }
+
+    @Override
+    public String pathSpec() {
+        return "/v3/segments";
```
kezhenxu94, line 58:
```suggestion
        return "/v3/segments.jsonl";
```

wu-sheng:
Are we going to break the endpoint? If so, lua agent has to follow the change, and test tool has to be changed too. You need to be careful about this kind of change.

wu-sheng:
Personally, I think this is an enhancement, rather than a break. Let's discuss this first, and decide the right move.

wu-sheng:
About the `/segments` endpoint, take a look at this, https://github.com/apache/skywalking-nginx-lua/blob/master/lib/skywalking/client.lua#L150

@dmsolr the main repo lua agent e2e hasn't been restored, could you add it back?

wu-sheng:
@mrproliu Could you add the HTTP document back? https://github.com/apache/skywalking/blob/master/docs/en/protocols/HTTP-API-Protocol.md

I think I deleted them when I was refactor codes.

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/handler/v8/rest/TraceSegmentReportMultipleLineServletHandler.java
wu-sheng: @kezhenxu94 This will break the e2e, LUA agent, and test tool.

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/test/java/org/apache/skywalking/oap/server/receiver/trace/provider/handler/v8/rest/TraceSegmentReportServletHandlerTest.java
```diff
@@ -0,0 +1,127 @@
...
+            new TraceSegmentReportSingleServletHandler(moduleManager, null, null);
+
+        when(request.getReader()).thenReturn(new BufferedReader(new StringReader(singleJson)));
+        final List<SegmentObject> segmentObjects = singleServletHandler.parseSegments(request);
+        Assert.assertEquals(segmentObjects.size(), 1);
+    }
+
+    @Test
+    public void testMultipleLine() throws IOException {
+        String multipleLineJson = "{" +
```
wu-sheng, line 85: I don't think this is a good example. This is not an official format in JSON.

wu-sheng: And isn't better than the official JSON list way.

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/handler/v8/rest/TraceSegmentReportMultipleLineServletHandler.java
wu-sheng: If we really hope to break the endpoint, consider using `/segment` and `/segments`, which could represent better. But before the change, need to update the test tool and LUA agent, then back to this PR.

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/test/java/org/apache/skywalking/oap/server/receiver/trace/provider/handler/v8/rest/TraceSegmentReportServletHandlerTest.java
mrproliu: It adapts to the current LUA agent. Should I need to fix the LUA first? let the LUA agent using the JSON object or JSON list.

mrproliu:
> Personally, I think this is an enhancement, rather than a break. Let's discuss this first, and decide the right move.

I think we just need to use '/segment' as a JSON object, '/segments' as the JSON list.

wu-sheng:
> I think we just need to use '/segment' as a JSON object, '/segments' as the JSON list.

Then you need to update the test tool to support both of these, https://github.com/apache/skywalking-agent-test-tool

mrproliu:
> > I think we just need to use '/segment' as a JSON object, '/segments' as the JSON list.
> 
> Then you need to update the test tool to support both of these, https://github.com/apache/skywalking-agent-test-tool

Sure, I will change the endpoint, then update the test tool and LUA agent.

docs/en/protocols/HTTP-API-Protocol.md
```diff
@@ -1,6 +1,186 @@
...
 
 HTTP API Protocol defines the API data format, including api request and response data format.
+They use the HTTP1.1 wrapper of the official [SkyWalking Trace Data Protocol v3](Trace-Data-Protocol-v3.md). Read it for more details.
 
-HTTP APIs are the HTTP1.1 wrapper of the official [SkyWalking Trace Data Protocol v3](Trace-Data-Protocol-v3.md). Read it for more details.
+## Instance Management
 
+Detail information about data format can be found in [Instance Management](https://github.com/apache/skywalking-data-collect-protocol/blob/master/management/Management.proto).
+
+- Report service instance
```
wu-sheng, line 10:
```suggestion
- Report service instance properties
```

docs/en/protocols/HTTP-API-Protocol.md
```diff
@@ -1,6 +1,186 @@
...
+OutPut:
+
+```json
+{}
+```
+
+## Trace Report
+
+Detail information about data format can be found in [Instance Management](https://github.com/apache/skywalking-data-collect-protocol/blob/master/language-agent/Tracing.proto).
+There has two way to report segment data.
```
wu-sheng, line 54:
```suggestion
There are two ways to report segment data, one segment per request or segment array in the bulk mode.
```

mrproliu:
> Two suggestions, wait for your other updates. e2e should be broken as endpoint changed.

The test tools and LUA agent has been finished, and all check has been passed. please re-check.

wu-sheng:
Please update the PR description, they are out of date.

mrproliu:
> Please update the PR description, they are out of date.

done

wu-sheng:
This PR is required manually update.

PR #4748, opened 2020-05-06
Support define multiple OAL configuration

arugal:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/oal/rt/OALDefine.java
```diff
@@ -0,0 +1,52 @@
...
+
+import lombok.Getter;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Define multiple OAL configuration
+ */
+@Getter
+public enum OALDefine {
```
wu-sheng, line 28: Enum is now good for extension, because you have to change the source codes. Please consider this as an abstract type.

arugal: Is that right ?
```java
@Getter
public abstract class OALDefine {

    OALDefine(final String configFile,
              final String sourcePackage,
              final String dynamicMetricsClassPackage,
              final String dynamicMetricsBuilderClassPackage, final String dynamicDispatcherClassPackage) {
        this.configFile = requireNonNull(configFile);
        this.sourcePackage = requireNonNull(sourcePackage);
        this.dynamicMetricsClassPackage = requireNonNull(dynamicMetricsClassPackage);
        this.dynamicMetricsBuilderClassPackage = requireNonNull(dynamicMetricsBuilderClassPackage);
        this.dynamicDispatcherClassPackage = requireNonNull(dynamicDispatcherClassPackage);
    }

    private final String configFile;
    private final String sourcePackage;
    private final String dynamicMetricsClassPackage;
    private final String dynamicMetricsBuilderClassPackage;
    private final String dynamicDispatcherClassPackage;
}
```
```java
public class OfficialOALDefine extends OALDefine {

    public static final OfficialOALDefine INSTANCE = new OfficialOALDefine();

    private OfficialOALDefine() {
        super(
            "official_analysis.oal",
            "org.apache.skywalking.oap.server.core.source.",
            "org.apache.skywalking.oal.rt.official.metrics.",
            "org.apache.skywalking.oal.rt.official.metrics.builder.",
            "org.apache.skywalking.oal.rt.official.dispatcher."
        );
    }
}
```

wu-sheng: Yes. Like this.

arugal: Done.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/oal/rt/OALEngineLoaderService.java
```diff
@@ -0,0 +1,78 @@
...
+@RequiredArgsConstructor
+public class OALEngineLoaderService implements Service {
+
+    private final Set<OALDefine> oalDefineSet = new HashSet<>();
+    private final ModuleManager moduleManager;
+
+    /**
+     * Normally it is invoked in the {@link ModuleProvider#start()} of the receiver-plugin module.
+     */
+    public void loader(OALDefine define) throws ModuleStartException {
```
wu-sheng, line 46: `loader` -> `load`

wu-sheng: Should not have `ModuleStartException`, this is not a module

arugal: `OALEngine` throws `ModuleStartException` and `OALCompileException`. Meanwhile `OALEngineLoaderService#load()` is called in `ModuleProvider#start()`.
https://github.com/apache/skywalking/blob/d713076f011af2b97f61969e96fb2b56dbccbb06/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/oal/rt/OALEngine.java#L28-L36

arugal: > `loader` -> `load`

Done,

wu-sheng:
I can't update your branch, seems you close it. You have to update the branch by yourself.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/oal/rt/OALEngineLoaderService.java
```diff
@@ -0,0 +1,78 @@
...
+public class OALEngineLoaderService implements Service {
+
+    private final Set<OALDefine> oalDefineSet = new HashSet<>();
+    private final ModuleManager moduleManager;
+
+    /**
+     * Normally it is invoked in the {@link ModuleProvider#start()} of the receiver-plugin module.
+     */
+    public void load(OALDefine define) throws ModuleStartException {
+        if (oalDefineSet.contains(define)) {
```
kezhenxu94, line 47: Seems you did not override the `hashCode` and `equals` methods in `OALDefine` and its subclasses

wu-sheng: Default, hash* uses the object ID, so, this only works for the same object, rather than the same OAL.

This should be fixed.

BTW, I am considering to enhance this later, merging the content rather than the file only. But this is totally another story.

kezhenxu94:
@arugal can you elaborate more on how to define more than one OAL scripts, or give us a typical case when a user want to add a customized OAL script? I didn't see any related configurations

arugal:
> @arugal can you elaborate more on how to define more than one OAL scripts, or give us a typical case when a user want to add a customized OAL script? I didn't see any related configurations or docs

This is not part of the user action, just like `official_analysis.oal` and [browser_analysis.oal](https://github.com/apache/skywalking/pull/4228/files#diff-d60fcc6e93a2874d1d70aef648be04f8). It can also be used for meter-system.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/oal/rt/OALEngineLoaderService.java
arugal: Done.

wu-sheng:
> can you elaborate more on how to define more than one OAL scripts, or give us a typical case when a user want to add a customized OAL script? I didn't see any related configurations or docs

One is he mentioned above, the other is, I am expecting to separate the OAL into every receiver. Such as if CLR receiver is not active, then CLR metrics process/index are not there automatically. The same story is in the mesh solution, Zipkin receiver, and Jaeger too. Make sense?

wu-sheng:
Take a look at the storage provider, such as `StorageModuleElasticsearchProvider#start`.  Storage installation happens there
```java
StorageEsInstaller installer = new StorageEsInstaller(getManager(), config);
installer.install(elasticSearchClient);
```

The OAL engine is required to process before that, but currently, there is no guarantee, right?

arugal:
> The OAL engine is required to process before that

Yes. 

> but currently, there is no guarantee, right?

I've moved `Installer#install` to `notifyAfterCompleted `, `AnnotationScan#scan` as well.

wu-sheng:
Take a look at my PR, don't change in that way. Moving it there could make no index in ttl timer and cache timer, Because the same startup sequence issue too. 

You could keep the codes in previous way, I have made the whole thing works.

wu-sheng:
Please fix the conflict including follow the latest installation logic.

PR #4756, opened 2020-05-07
Update the path of  tls_key_generate.sh script

songzhendong:
Update the tls_key_generate.sh script to the new path(The old path cannot be browsed)


wu-sheng:
I think you should create a new branch based on the main repo master, rather than keeping adding new commit.

PR #4758, opened 2020-05-08
Format index name with namespace

hanahmily:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

### Bug fix
- Bug description.
When setting up a namespace for elasticsearch storage, TTL modal would try to create an existed index.

- How to fix?
Generate correct latest index name by formatIndexName method.

### DIFF
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/HistoryDeleteEsDAO.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/HistoryDeleteEsDAO.java
@@ -66,7 +66,7 @@ public class HistoryDeleteEsDAO extends EsDAO implements IHistoryDeleteDAO {
         for (String prepareDeleteIndex : prepareDeleteIndexes) {
             client.deleteByIndexName(prepareDeleteIndex);
         }
-        String latestIndex = TimeSeriesUtils.latestWriteIndexName(model);
+        String latestIndex = client.formatIndexName(TimeSeriesUtils.latestWriteIndexName(model));
         if (!leftIndices.contains(latestIndex)) {
             client.createIndex(latestIndex);
         }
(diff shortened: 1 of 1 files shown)


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/HistoryDeleteEsDAO.java
```diff
@@ -66,7 +66,7 @@ public void deleteHistory(Model model, String timeBucketColumnName, int ttl) thr
         for (String prepareDeleteIndex : prepareDeleteIndexes) {
             client.deleteByIndexName(prepareDeleteIndex);
         }
-        String latestIndex = TimeSeriesUtils.latestWriteIndexName(model);
+        String latestIndex = client.formatIndexName(TimeSeriesUtils.latestWriteIndexName(model));
         if (!leftIndices.contains(latestIndex)) {
             client.createIndex(latestIndex);
         }
```
kezhenxu94, line 72: `client.createIndex` already calls `formatIndexName` internally, I think you may need to `formatIndexName` before checking `if (!leftIndices.contains(latestIndex))` and use the original one in `client.createIndex`, am I right?

kezhenxu94: > `client.createIndex` already calls `formatIndexName` internally, I think you may need to `formatIndexName` before checking `if (!leftIndices.contains(latestIndex))` and use the original one in `client.createIndex`, am I right?

Otherwise, the namespace with be prepended twice, perhaps

wu-sheng: `latestIndex` variable should be used in `contains` check only, `createIndex` should use the unformatted name like the old codes.

hanahmily: done

PR #4770, opened 2020-05-10
Test: add E2E test for Python Agent

kezhenxu94:


test/e2e/e2e-test/docker/python/docker-compose.yml
```diff
@@ -0,0 +1,78 @@
...
+
+  ui:
+    extends:
+      file: ../base-compose.yml
+      service: ui
+    depends_on:
+      oap:
+        condition: service_healthy
+
+  provider:
```
wu-sheng, line 32: Typically, I recommend you set the e2e in `python`->`java`->`python`, because it could verify the header propagation more accurately. WDYT?

PR #4775, opened 2020-05-11
Support regex to include or exclude the names in alarm module

mrproliu:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

___
- Follow the dev mail list discussion, I provide the regular to include or exclude the names in alarm module.

### DIFF
--- a/dist-material/alarm-settings.yml
+++ b/dist-material/alarm-settings.yml
@@ -50,7 +50,6 @@ rules:
     metrics-name: service_instance_resp_time
     op: ">"
     threshold: 1000
-    include-names-regex: instance\_\d+
     period: 10
     count: 2
     silence-period: 5
--- a/dist-material/alarm-settings.yml
+++ b/dist-material/alarm-settings.yml
@@ -50,6 +50,7 @@ rules:
     metrics-name: service_instance_resp_time
     op: ">"
     threshold: 1000
+    include-names-regex: instance\_\d+
     period: 10
     count: 2
     silence-period: 5
--- a/docs/en/setup/backend/backend-alarm.md
+++ b/docs/en/setup/backend/backend-alarm.md
@@ -22,8 +22,8 @@ Alarm rule is constituted by following keys
 [List of all potential metrics name](#list-of-all-potential-metrics-name).
 - **Include names**. The following entity names are included in this rule. Please follow [Entity name define](#entity-name).
 - **Exclude names**. The following entity names are excluded in this rule. Please follow [Entity name define](#entity-name).
-- **Include names regex**. Provide a regex to include the entity names.
-- **Exclude names regex**. Provide a regex to exclude the exclude names.
+- **Include names regex**. Provide a regex to include the entity names. If both setting the include name list and include name regex, both rules will take effect.
+- **Exclude names regex**. Provide a regex to exclude the exclude names. If both setting the exclude name list and exclude name regex, both rules will take effect.
 - **Threshold**. The target value. 
 For multiple values metrics, such as **percentile**, the threshold is an array. Described like  `value1, value2, value3, value4, value5`.
 Each value could the threshold for each value of the metrics. Set the value to `-` if don't want to trigger alarm by this or some of the values.  
--- a/docs/en/setup/backend/backend-alarm.md
+++ b/docs/en/setup/backend/backend-alarm.md
@@ -22,6 +22,8 @@ Alarm rule is constituted by following keys
 [List of all potential metrics name](#list-of-all-potential-metrics-name).
 - **Include names**. The following entity names are included in this rule. Please follow [Entity name define](#entity-name).
 - **Exclude names**. The following entity names are excluded in this rule. Please follow [Entity name define](#entity-name).
+- **Include names regex**. Provide a regex to include the entity names.
+- **Exclude names regex**. Provide a regex to exclude the exclude names.
 - **Threshold**. The target value. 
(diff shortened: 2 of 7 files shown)


oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
```diff
@@ -115,6 +121,24 @@ public void in(MetaInAlarm meta, Metrics metrics) {
...
+                    log.trace("{} isn't matches include regex {}", meta.getName(), includeNamesRegex);
+                }
+                return;
+            }
+        }
+
+        if (StringUtil.isNotEmpty(excludeNamesRegex)) {
+            if (meta.getName().matches(excludeNamesRegex)) {
+                if (log.isTraceEnabled()) {
+                    log.trace("{} is matches exclude regex {}", meta.getName(), excludeNamesRegex);
```
kezhenxu94, line 136:
```suggestion
                    log.trace("{} matches the exclude regex {}", meta.getName(), excludeNamesRegex);
```

dist-material/alarm-settings.yml
```diff
@@ -50,6 +50,7 @@ rules:
     metrics-name: service_instance_resp_time
     op: ">"
     threshold: 1000
+    include-names-regex: instance\_\d+
```
kezhenxu94, line 53: This default value may break existed versions, right?

mrproliu: I write on the wrong file, I will change it into the demo file tomorrow.

docs/en/setup/backend/backend-alarm.md
```diff
@@ -22,6 +22,8 @@ Alarm rule is constituted by following keys
 [List of all potential metrics name](#list-of-all-potential-metrics-name).
 - **Include names**. The following entity names are included in this rule. Please follow [Entity name define](#entity-name).
 - **Exclude names**. The following entity names are excluded in this rule. Please follow [Entity name define](#entity-name).
+- **Include names regex**. Provide a regex to include the entity names.
+- **Exclude names regex**. Provide a regex to exclude the exclude names.
```
wu-sheng, line 26: What happens if `include names` and `include names regex` exist both? I think we should indicate in the document.

dist-material/alarm-settings.yml
mrproliu: Change it into the `alarm-settings-sample.yml` file.

docs/en/setup/backend/backend-alarm.md
mrproliu: I think both will take effect.

wu-sheng:
Could you post the mail list link here?

mrproliu:
> Could you post the mail list link here?

Sure, here is link. 
https://lists.apache.org/x/thread.html/rb6d91093776cf61880425449d3d25f9d560cd32363c0075de9a85d13@%3Cdev.skywalking.apache.org%3E

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
```diff
@@ -115,6 +121,24 @@ public void in(MetaInAlarm meta, Metrics metrics) {
...
+            if (!meta.getName().matches(includeNamesRegex)) {
+                if (log.isTraceEnabled()) {
+                    log.trace("{} doesn't match the include regex {}", meta.getName(), includeNamesRegex);
+                }
+                return;
+            }
+        }
+
+        if (StringUtil.isNotEmpty(excludeNamesRegex)) {
+            if (meta.getName().matches(excludeNamesRegex)) {
```
wu-sheng, line 134: This string#matches includes `Pattern p = Pattern.compile(regex)`, due to we have the regex expression before all checked, I think we should compile them first when loading the config or config changed.

Could you run a benchmark to show the performance difference between this and the pre-compiled match?

mrproliu: I do a little benchmark, using regex `Service\_\d+` to match` Service_1`, make a loop with 10 million times. 

String#matches: 7198 millisecond
Pattern#matcher()#matches(): 977 millisecond

I will change to using the pre-build mode, thanks.

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
```diff
@@ -97,19 +106,38 @@ public void in(MetaInAlarm meta, Metrics metrics) {
...
+        if (includeNamesRegex != null) {
+            if (!includeNamesRegex.matcher(metaName).matches()) {
+                if (log.isTraceEnabled()) {
+                    log.trace("{} doesn't match the include regex {}", metaName, includeNamesRegex);
+                }
+                return;
+            }
+        }
+
+        if (excludeNamesRegex != null) {
```
wu-sheng, line 137: Just a reminder, as we are using JDK8+, Optional could replace all these `!=null`. But both of these styles exist today. So, don't require change, anyway.

mrproliu: Thanks, I will change to this way since next PR.

PR #4791, opened 2020-05-15
fix: fix double-checked locking bug

RobotZhong:
fix issue about instruction reordering.
more detail in https://rules.sonarsource.com/java/tag/multi-threading/RSPEC-2168

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?
use volatile static field.
___
more detail in https://rules.sonarsource.com/java/tag/multi-threading/RSPEC-2168

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/FileWriter.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/FileWriter.java
@@ -42,7 +42,7 @@ import org.apache.skywalking.apm.util.RunnableWithExceptionProtection;
  * The <code>FileWriter</code> support async file output, by using a queue as buffer.
  */
 public class FileWriter implements IWriter {
-    private static FileWriter INSTANCE;
+    private volatile static FileWriter INSTANCE;
     private static final Object CREATE_LOCK = new Object();
     private FileOutputStream fileOutputStream;
     private ArrayBlockingQueue logBuffer;


wu-sheng:
First, this class in practice, actually wouldn't be initialized concurrently, because of the pre-main.

Then, if you want to fix from the code review perspective, I prefer you to choose `With inner static holder` solution. 

What would you say?

wu-sheng:
@RobotZhong Any feedback?

RobotZhong:
It is not necessary, i will close this pr.
Thank you for your reply.

wu-sheng:
@RobotZhong This code misguided you, so it ma confuse others. I think making a change makes sense to me, I prefer not to add volatile is for runtime performance consideration. That is why I asked you to change.

kezhenxu94:
> It is not necessary, i will close this pr.
> Thank you for your reply.

@RobotZhong technically speaking it's indeed a bug, although it seldom happens in production, fixing this is a good first step for you to go through the contribution process in this project and for further contributions, if you will.

Let's go back to the problem itself, there're multiple fixes in my mind, but initiating it eagerly seems to fit this case most, `FileWriter` is mostly always used in production, `volatile` makes it always loaded from the memory in every call, and thus does more harm than good. @wu-sheng's suggestion "with inner static holder" also makes sense to me

RobotZhong:
Still have confusion with org.apache.skywalking.apm.agent.core.logging.core.FileWriter#FileWriter

    private FileWriter() {
        logBuffer = new ArrayBlockingQueue(1024);
        final ArrayList<String> outputLogs = new ArrayList<String>(200);
        Executors
            .newSingleThreadScheduledExecutor(new DefaultNamedThreadFactory("LogFileWriter"))
            .scheduleAtFixedRate(new RunnableWithExceptionProtection(new Runnable() {
                @Override public void run() {
                    try {
                        logBuffer.drainTo(outputLogs);
                        for (String log : outputLogs) {
                            writeToFile(log + Constants.LINE_SEPARATOR);
                        }
                        try {
                            fileOutputStream.flush();
                        } catch (IOException e) {
                            e.printStackTrace();
                        }
                    } finally {
                        outputLogs.clear();
                    }
                }
            }, new RunnableWithExceptionProtection.CallbackWhenException() {
                @Override public void handle(Throwable t) {
                }
            }
            ), 0, 1, TimeUnit.SECONDS);
    }


Is it possible in some cases, outputLogs is empty and cause a NullPointerException?
Because fileOutputStream is only constructed in #prepareWriteStream, which is only called by #writeToFile

wu-sheng:
`outputLogs` could be empty, but not NULL. How could an assigned and initialized variable back to null? Especially it is `final` already. A little confused about your question.

RobotZhong:
Sorry for my poor English.
I mean `fileOutputStream ` is null .

wu-sheng:
I think that is a fail-safe coding style only. Nothing special.

RobotZhong:
Why not call #prepareWriteStream before iterate `outputLog`?

                        if (prepareWriteStream()) {
                            for (String log : outputLogs) {
                                writeToFile(log + Constants.LINE_SEPARATOR);
                            }
                            try {
                                fileOutputStream.flush();
                            } catch (IOException e) {
                                e.printStackTrace();
                            }
                        }

wu-sheng:
I don't know. These are not wroten by me, and these are not related to this context. If you want to shape this, I would recommend you to talk this in the dev mail list or another PR about your proposal about this part. 

SkyWalking is a super complex and hundreds of contributors project. I think even for me, I can't explain everything, or can't explain many things. The codes you saw are there, could be right and could be wrong. Take part in, change them if you could approve they are wrong or not good enough. We could continue the discussion there.

PR #4753, opened 2020-05-07
[CVE] Update Nacos 1.0.0 to Nacos 1.2.0(Solve the security problem of the old version of fastjson)

songzhendong:
___
### Bug fix
- Bug description.
  Fastjson1.4.7 referenced  by Nacos1.0.0 has a security vulnerability

- How to fix?
  Update Nacos 1.0.0 to Nacos 1.2.0（Fastjson1.4.7 upgrade to fastjson1.5.8）

PR #4804, opened 2020-05-20
Fastjson has a serious security problem in 1.2.58

wcc526:
Fastjson has a serious security problem in 1.2.58,which will cause RCE
https://www.anquanke.com/post/id/199527
https://www.cnblogs.com/tr1ple/p/12348886.html
https://github.com/jas502n/fastjson-1.2.58-rce

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___

### DIFF
--- a/tools/dependencies/known-oap-backend-dependencies-es7.txt
+++ b/tools/dependencies/known-oap-backend-dependencies-es7.txt
@@ -38,7 +38,7 @@ elasticsearch-x-content-7.0.0.jar
 error_prone_annotations-2.3.2.jar
 etcd4j-2.17.0.jar
 failureaccess-1.0.1.jar
-fastjson-1.2.58.jar
+fastjson-1.2.68.jar
 freemarker-2.3.28.jar
 graphql-java-8.0.jar
 graphql-java-tools-5.2.3.jar


wu-sheng:
What is the meaning od this pull request? You changed the check file which will cause CI failure. Fastjson in introduced by the Nacos, I think. You may need to update that.

wu-sheng:
Take #4753  as a good example.

wu-sheng:
Fastjson is really breaking too many time, all @apache/skywalking-commitskywa, we really need to keep eyes on this lib. Consider to remove it and Nacos with them.

Unstable and unsecurity lib is a big concern. I will send the official mail to the mail list.

wu-sheng:
Submitted the request to remove(fastjson) dependency on the Nacos, https://github.com/alibaba/nacos/issues/2842. This has been accepted and tagged as 1.3.0

KomachiSion:
@wu-sheng, It seems that Nacos can replace fastjson with Jackson this week, so can you pending this PR for several days?

wu-sheng:
Like I said, this PR is not right, You didn't update anything.

About Nacos, if they could release on time, we could upgrade the version and remove the fastjson.

wu-sheng:
Fastjson has been removed from the Nacos. We are safe for mow

PR #4805, opened 2020-05-20
Check http protocol in Python agent as well

kezhenxu94:


wu-sheng:
I think you should set up 2 test cases? One for HTTP, the other for gRPC.

kezhenxu94:
> I think you should set up 2 test cases? One for HTTP, the other for gRPC.

There're two services, provider and consumer, they use grpc and http respectively, I think it's enough, no?

wu-sheng:
> There're two services, provider and consumer, they use grpc and http respectively, I think it's enough, no?

I see, it is enough.

wu-sheng:
Seems python e2e fails.

wu-sheng:
What happens here?

kezhenxu94:
Should be good to go

Issue #4707, opened 2020-04-24
Agent send a lot of IPV4s,so it causes the following error:org.h2.jdbc.JdbcSQLException: Value too long for column "PROPERTIES VARCHAR(2000)


welcome6688:
When agent is in VM,  it only sends **a IPV4 record**.  When agent is in docker, it sends many IPV4 records. so I want to modify  docker's agent

_Originally posted by @welcome6688 in https://github.com/apache/skywalking/issues/2511#issuecomment-618983747_

wu-sheng:
Contributions are welcome, this is easy for the first time contributor.

welcome6688:
I have a question about agent. Agent sends OSUtil information to server,result in table field error of server.Because there are too many IPv4 addresses obtained on the agent machine, the table fields of the server cannot be placed. Agent is in K8s. Server is oap-es-7.
![image](https://user-images.githubusercontent.com/8703738/80212414-05316500-866a-11ea-9eee-ea95bd1dbf75.png)
I want the agent not to send IPv4 information. I don't know how to change it?

welcome6688:
So I don't understand why I only get one record in the virtual machine, and I get many records in the docker.
Can there be any place to modify the agent so that the agent does not send IPv4s.

welcome6688:
![image](https://user-images.githubusercontent.com/8703738/80213409-b684ca80-866b-11ea-9fa0-ba7fc1f6e895.png)
The picture above shows a lot of IPV4s records

wu-sheng:
That is based on your docker virtual network. It is designed by docker and k8s, etc. 

We should set the limitation of the list. OSUtil is already the place to change.

welcome6688:
![image](https://user-images.githubusercontent.com/8703738/80215490-2052a380-866f-11ea-8c65-9d12d3612849.png)
I modified this method to return only empty arrays, but this is still the problem in the final run. Is there any other place to change

wu-sheng:
No, there isn't, this is the only place.

wu-sheng:
@welcome6688 I think @JaredTan95 provided this for you.

PR #4817, opened 2020-05-25
Fix too many mang ipv4 records reported by the agent when deployed in container.

JaredTan95:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/4707
___
### Bug fix
- Bug description.
too much ipv4 record causes too long when inserting into DB.
- How to fix?
add a limitation of the ipv4 list size.
___

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/os/OSUtil.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/os/OSUtil.java
@@ -57,7 +57,7 @@ public class OSUtil {
 
     public static List<String> getAllIPV4() {
         if (IPV4_LIST == null) {
-            IPV4_LIST = new LinkedList<String>();
+            IPV4_LIST = new LinkedList<>();
             try {
                 Enumeration<NetworkInterface> interfs = NetworkInterface.getNetworkInterfaces();
                 while (interfs.hasMoreElements()) {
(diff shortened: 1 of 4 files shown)


docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -84,6 +84,7 @@ property key | Description | Default |
 `agent.force_reconnection_period `|Force reconnection period of grpc, based on grpc_channel_check_interval.|`1`|
 `agent.operation_name_threshold `|The operationName max length, setting this value > 190 is not recommended.|`150`|
 `agent.keep_tracing`|Keep tracing even the backend is not available if this value is `true`.|`false`|
+`osinfo.ipv4_list_size`| Limit the length of the ipv4 list size. |`5`|
```
wu-sheng, line 87: Seems document doesn't match the codes, in the Config, it is 10 as default :)

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/os/OSUtil.java
```diff
@@ -116,6 +116,9 @@ public static int getProcessNo() {
         }
         List<String> allIPV4 = getAllIPV4();
         if (allIPV4.size() > 0) {
+            if (allIPV4.size() >= ipv4Size) {
```
kezhenxu94, line 119: if `allIPV4.size() == ipv4Size`, there is no need to `subList`, it must be itself 

```suggestion
            if (allIPV4.size() > ipv4Size) {
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/os/OSUtil.java
```diff
@@ -116,6 +116,9 @@ public static int getProcessNo() {
...
+                allIPV4.subList(0, ipv4Size);
```
wu-sheng, line 120: From my reading of the `ArrayList#subList`, the input parameter would not be changed, the new list is the return list. Could you check locally?

PR #4833, opened 2020-05-27
Correct the service node type.

wu-sheng:
All `service type` in the segment should be `NodeType.Normal`, I think I wasn't coding right. :)


kezhenxu94:
@arugal although the changes may not be related to the plugins, we'd be better to wait for all the tests to pass before merging, and, please delete the branch after merging

Issue #3338, opened 2019-08-27
vertx plugin bug


wentjiang:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [X] plugin
- [X] Bug
- [X] vertx
 
___
### Bug
- Which version of SkyWalking, OS and JRE?
V6.2.0 jdk8

- What happen?
my project use vertx v3.4 ,the agent log show 
java.lang.NullPointerException
	at org.apache.skywalking.apm.plugin.vertx3.HttpServerResponseImplEndInterceptor.beforeMethod(HttpServerResponseImplEndInterceptor.java:39)
I review code and debug, the context is null, so I find there are not use class io.vertx.ext.web.impl.RouterImpl in my project 
finally,I solve it by I create io.vertx.core.http.impl.ServerConnection enhance, it works healthy

kezhenxu94:
Thanks for reporting, @wentjiang since you have already found the solution, will you send a pull request to fix the bug?

wu-sheng:
FYI @BFergerson

PR #4840, opened 2020-05-29
Increased Vert.x support

BFergerson:
- [x] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues

___
### Bug fix

Fixes https://github.com/apache/skywalking/issues/3338. @wentjiang had the right idea for the fix but the method chosen to enhance is not present ver. 3.5+. I've chosen another method that also works and is present in all versions. 

___

I added support for ver. 3.9.1 as well for more complex route handling. The tests I've been using were too simple so I made them more realisitic. The tests I'm using now look like:
 - User sends request to `/web-case`
- `/web-case` sends a request to `/healthcheck` (ver. 3.5- = sync / ver. 3.6+ = async)
- Once response is recieved another request is sent to `/withBodyHandler` (ver. 3.5- = sync / ver. 3.6+ = async)
- Before `/withBodyHandler` is called this is processed by Vert.x's internal `BodyHandler`
- `BodyHandler` passes request off to `/withBodyHandler`
- Once the response is received the original request to `/web-case` is responded to

This results in the segment trace stack:
```
ENTER: /case/web-case
     ASYNC SPAN STARTED [1]
     LOCAL: VertxWebController$$Lambda$.handle(RoutingContext)
          EXIT: /case/healthCheck
               ASYNC SPAN STARTED [2]
          CLOSE EXIT
     CLOSE LOCAL
CLOSE ENTER

ENTER: /case/healthCheck
     ASYNC SPAN STARTED [3]
     LOCAL: VertxWebController$$Lambda$.handle(RoutingContext)
           ASYNC SPAN FINISHED [3]
     CLOSE LOCAL
CLOSE ENTER

ASYNC SPAN FINISHED [2]
LOCAL: #/case/healthCheck
     EXIT: /case/web-case/withBodyHandler
          ASYNC SPAN STARTED [4]
     CLOSE EXIT
CLOSE LOCAL

ENTER: /case/web-case/withBodyHandler
     ASYNC SPAN STARTED [5]
     LOCAL: BodyHandlerImpl.handle(RoutingContext)
     EXIT LOCAL
CLOSE ENTER

LOCAL: VertxWebController$$Lambda$.handle(RoutingContext)
     ASYNC SPAN FINISHED [5]
CLOSE LOCAL

ASYNC SPAN FINISHED [4]
LOCAL: #/vertx-web-3_54minus-scenario/case/web-case/withBodyHandler
     ASYNC SPAN FINISHED [1]
CLOSE LOCAL
```

If you look at the above the `ENTER: /case/web-case/withBodyHandler` is called on a separate context than `LOCAL: VertxWebController$$Lambda$.handle(RoutingContext)`. This is because of the added `BodyHandlerImpl.handle(RoutingContext)`. This would occur with any additional handlers before the user-code handler. With the added solution you can add additional handlers (e.g. `CorsHandler`) and it'll still result in the spans being created correctly:
trace-list

Apologies for the constant changes. I believe this PR to offer the most thorough support and I wanted to make sure to get better Vert.x support in SkyWalking 8.0.0. When I saw the previous PR couldn't support 3.9.1, I rushed it through because I saw there were bigger changes that I needed to make and I wasn't sure how quickly I could make them.


wu-sheng:
> Apologies for the constant changes. I believe this PR to offer the most thorough support and I wanted to make sure to get better Vert.x support in SkyWalking 8.0.0. When I saw the previous PR couldn't support 3.9.1, I rushed it through because I saw there were bigger changes that I needed to make and I wasn't sure how quickly I could make them.

It should be fine to change the plugin, basically, that is why we design the agent-based plugin system. Users wouldn't feel these changes happened, because they never really use the APIs, and even our changes have some issues, they could remove or rollback easily by themselves.

BTW, I noticed, there is no UT for the new interception, but if those have been covered by the plugin integration test, it should be fine. Let's what coverage result says.

BFergerson:
@wu-sheng, what is UT?

wu-sheng:
> what is UT?

Unit test. The e2e about InfluxDB seems not stable somehow.

Issue #4850, opened 2020-06-02
hystrix plugin cause biz error


zifeihan:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS and JRE?
skywalking 6.4.0, hystrix 1.4.3

- Which company or project?

- What happen?
when i use skywalking 6.4.0, hystrix 1.4.3, there has error message about hystrix plugin. like this :
"HystrixRequestContext.initializeContext() must be called at the beginning of each request before RequestVariable functionality can be used.", when remove skywalking agent, everything be ok.
___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

zifeihan:
this problem is, origin object HystrixConcurrencyStrategyDefault is wrapped to HystrixConcurrencyStrategy. under hystrix 1.4.3 com.netflix.hystrix.HystrixRequestLog#getCurrentRequest(com.netflix.hystrix.strategy.concurrency.HystrixConcurrencyStrategy) this method doesn't check isCurrentThreadInitialized, but  checked in other high version
![image](https://user-images.githubusercontent.com/16839929/83721643-12c90a00-a66e-11ea-8fd0-a4454fc3e979.png)

PR #4857, opened 2020-06-04
fix hystrix plugin cause biz error 

zifeihan:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/4850
___
### Bug fix
- Bug description.
this problem is, origin object HystrixConcurrencyStrategyDefault is wrapped to HystrixConcurrencyStrategy. under hystrix 1.4.3 com.netflix.hystrix.HystrixRequestLog#getCurrentRequest(com.netflix.hystrix.strategy.concurrency.HystrixConcurrencyStrategy) this method doesn't check isCurrentThreadInitialized, but checked in other high version
- How to fix?
like hystrix high version, Return null if the {@link HystrixRequestContext} has not been initialized for the current thread.
___


apm-sniffer/apm-sdk-plugin/hystrix-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hystrix/v1/SWHystrixLifecycleForwardingRequestVariable.java
```diff
@@ -0,0 +1,84 @@
...
+     * <p>
+     * If {@link HystrixRequestContext} has been initialized then call method in superclass:
+     * {@link HystrixRequestVariableDefault#get()}
+     */
+    @Override
+    public T get() {
+        if (!HystrixRequestContext.isCurrentThreadInitialized()) {
+            return null;
+        }
+        return super.get();
```
wu-sheng, line 81: So, this could be NPE, right? In some old versions.

zifeihan: this is super.get() method
```
public T get() {
        if (HystrixRequestContext.getContextForCurrentThread() == null) {
            throw new IllegalStateException(HystrixRequestContext.class.getSimpleName() + ".initializeContext() must be called at the beginning of each request before RequestVariable functionality can be used.");
        } else {
            ConcurrentHashMap<HystrixRequestVariableDefault<?>, HystrixRequestVariableDefault.LazyInitializer<?>> variableMap = HystrixRequestContext.getContextForCurrentThread().state;
            HystrixRequestVariableDefault.LazyInitializer<?> v = (HystrixRequestVariableDefault.LazyInitializer)variableMap.get(this);
            if (v != null) {
                return v.get();
            } else {
                HystrixRequestVariableDefault.LazyInitializer<T> l = new HystrixRequestVariableDefault.LazyInitializer(this);
                HystrixRequestVariableDefault.LazyInitializer<?> existing = (HystrixRequestVariableDefault.LazyInitializer)variableMap.putIfAbsent(this, l);
                return existing == null ? l.get() : existing.get();
            }
        }
    }
```
super.get() will trigger IllegalStateException, in some old versions

wu-sheng: Then next question is, why this method is being called? Is it called by SkyWalking plugin codes?

zifeihan: it ic called by hystrix AbstractCommand constructor.

wu-sheng: My question is, the old version caller, AbstractCommand, should not call this, right? Otherwise, it triggers Illegal issue itself. Could you be more clear what context is missing

zifeihan: concurrencyStrategy is HystrixConcurrencyStrategyDefault instance in no skywalking agent, but skywalking plugin wrapped it to HystrixConcurrencyStrategy instance. not HystrixConcurrencyStrategyDefault. thus skip "concurrencyStrategy instanceof HystrixConcurrencyStrategyDefault" condition. i don't know where context be initialize, but when "concurrencyStrategy instanceof HystrixConcurrencyStrategyDefault" condition, it checkd  this "HystrixRequestContext.isCurrentThreadInitialized()".
image

wu-sheng: Then, why don't we change the SkyWalking wrapper logic? It seems more reasonable, right?

wu-sheng: With offline communication, right now, this seems the easiest solution w/o harm. I will keep this for now. Thanks.

PR #4863, opened 2020-06-05
Optimize TraceQuery

dominicqi:

- [x] Improve performance

- Related issues
___
1. Use FilterContext replace QueryContext.(reduce scoring cost and use filter cache)
2. reduce query index scope
   - range query use `multiple indices` `ignore_unavailable` params ignore missing or closed index
   - traceId query get the trace start time (got in #4353). add one more(next) index to query.


dominicqi:
> I am sorry to say, today, we don't require timestamp in thr trace id anymore.
> Many agents don't implement in this way, including our native lua agent.

So sad , many ES cluster was deploy as a `hot-warm architecture` , put the old data  in warm datanode. This will cause search very slow use the alias.

dominicqi:
@wu-sheng  Another way, can with save segment routing by `trace_id`. Then the `TraceIdQuery` can use the traceId routing to the right shard . I think i will improve the query very much.
Here is the [routing doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-routing-field.html)

dominicqi:
@wu-sheng  Another way, can with save segment routing by `trace_id`. Then the `TraceIdQuery` can use the traceId routing to the right shard . I think i will improve the query very much.
Here is the [routing doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-routing-field.html)

wu-sheng:
I know direct index query is faster. several seconds vs 500ms.

dominicqi:
> I know direct index query is faster. several seconds vs 500ms.

Use routing  then we could query with alias  , it is fast too. No need to found the direct index. Can direct to the right shard.

dominicqi:
image

image

routing vs no routing

wu-sheng:
Custom routing and sharding rule seems reasonable. But we need an explicit sharding key annotation, or add attribute into Column annotation. Let's think more clear about this. Make this possible in 8.1 release. OK?

dominicqi:
> Column

Ok

wu-sheng:
@dominicqi As the 8.0.0 is tagged and on voting, we could continue on this, if you have time.

wu-sheng:
#5132 is closing to merge, it provides similar features.

PR #4871, opened 2020-06-08
add php component

heyanlong:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- add php component


oap-server/server-bootstrap/src/main/resources/component-libraries.yml
```diff
@@ -421,3 +441,7 @@ Component-Server-Mappings:
   pulsar-consumer: Pulsar
   rest-high-level-client: Elasticsearch
   mariadb-jdbc: Mariadb
+  cURL: HttpClient
```
wu-sheng, line 444: This mapping should be targeting server, `HttpClient` isn't the case, I think. Why do you need this?

heyanlong: Because this is an `HttpClient` toolkit

wu-sheng: It doesn't represent the client type, it is for the server side. Such as, once you have a client exit span, which type of the server node should be shown in the topology page.

Issue #4865, opened 2020-06-05
kafka consumer start error


amwyyyy:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ x ] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Bug
version:
SkyWalking7.0.0
spring-kafka:2.2.8.RELEASE
kafka-clients:2.0.1

When kafka consumer uses topicPattern, the agent will report an error.
```
@KafkaListener(topicPattern = KAFKA_TOPIC_NAME)
public void onDemoMessage(ConsumerRecord<String, String> record) {
// ...
}
```
skywalking-api.log
```
ERROR 2020-06-05 17:57:30:137 restartedMain InstMethodsInter : class[class org.apache.kafka.clients.consumer.KafkaConsumer] before method[subscribe] intercept failure 
java.lang.ClassCastException: java.util.regex.Pattern cannot be cast to java.util.Collection
	at org.apache.skywalking.apm.plugin.kafka.SubscribeMethodInterceptor.beforeMethod(SubscribeMethodInterceptor.java:33)
	at org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstMethodsInter.intercept(InstMethodsInter.java:76)
	at org.apache.kafka.clients.consumer.KafkaConsumer.subscribe(KafkaConsumer.java)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.subscribeOrAssignTopics(KafkaMessageListenerContainer.java:584)

```

wu-sheng:
This seems a plugin incompatible case. Most plugins are contributed by the users, so, please try to learn the plugin development( we provided doc), and add supports to this case. We are welcome your contribution.

PR #4873, opened 2020-06-08
fix Kafka consumer subscribe topics from pattern

amwyyyy:
fix Kafka consumer subscribe topics from pattern 
[#4865](https://github.com/apache/skywalking/issues/4865)


wu-sheng:
Tests are failing, please check the logs and fix.

wu-sheng:
Once all CIs passed, please add the plugin tests to cover this case.

wu-sheng:
@amwyyyy Any update? Or are you clear what you should do?

amwyyyy:
@wu-sheng I don’t know what to do, I’m groping

wu-sheng:
When you are changing the plugin source codes, we would request you to add plugin tests. 
https://github.com/apache/skywalking/blob/master/docs/en/guides/Plugin-test.md

The point is this, we can't run your case locally, so we request you to add in the process of CI testing. Please read the document.

test/plugin/scenarios/kafka-scenario/src/main/java/test/org/apache/skywalking/apm/testcase/kafka/controller/CaseController.java
```diff
@@ -200,5 +208,44 @@ public void run() {
             consumer.close();
         }
     }
+
+    public class ConsumerThread2 extends Thread {
```
wu-sheng, line 212: After you change this, there is an `expectedData.yaml` file, you should add the expected segment data in it.

wu-sheng:
It seems tests fail. Could you try to run the tests locally?

wu-sheng:
Merged, welcome and congrats.

PR #4876, opened 2020-06-08
Add default templates for dashboard and topology.

wu-sheng:
With the UI closing to the final version, I could finish all configurations and add the default templates.

image

image

image


wu-sheng:
@dmsolr Seems the template initialization makes all InfluxDB related e2e cases fail. Please help.

PR #4902, opened 2020-06-11
Update prometheus fetcher for so11y

hanahmily:
1. Document how to set up so11y for cluster
1. Add label filter to aggregate metrics from dedicated labels.
1. Update self.yaml
1. Add UI template for so11y.


hanahmily:
image

PR #4916, opened 2020-06-12
Set version to 8.1.0 for next iteration.

wu-sheng:
8.0.0->8.1.0


wu-sheng:
@yu199195 ShardingSphere test cases show unstable(data don't match) twice. I am retrying. Please be advised.

wu-sheng:
@yu199195 It seems to keep failing? Could you take a look?

yu199195:
@wu-sheng  hi， sheng，my PR provider  is shardingsphere-4.x-scenario， but I haven't seen the test results of this project， why？，  shardingsphere-4.x-RC1-RC2-scenario  is an error. It was provided before，I'll keep an eye on it。

wu-sheng:
@yu199195 Because I rerun it. And yes, it comes from the old case. The logs just say expected data don't match. I don't know what causes that. 
I want to merge this PR, so just retry it. But the unstable seems to keep happening.

wu-sheng:
@yu199195 The error is here, https://github.com/apache/skywalking/pull/4916/checks?check_run_id=764945306

yu199195:
i down logs， and check in

yu199195:
I down logs， and check in

yu199195:
I compared the logs carefully with  expectedData.yaml in shardingSphere-4.x-RC3, no mismatches found..

wu-sheng:
@dmsolr ShardingSphere tests fail 3+ times. 

@yu199195 I noticed, there are large logs generated. Could you check locally?

PR #4936, opened 2020-06-16
Reduce the length of metric name

### LINKED ISSUE #4932: Self Observability can't start up when using MySQL storage
The table name is too long, reported by https://lists.apache.org/thread.html/r68b6f44f953f1bd9f2718a82f4bd1a8fea2a6c6e7d5b70e3cc9e8276%40%3Cdev.skywalking.apache.org%3E

Please adjust the name, and dashboard configuration. Also, need to add MySQL storage into the e2e, @kezhenxu94 we don't have that yet, right?

And Self Observability should be tested in the e2e.

hanahmily:

Reduce the length of metric name in so11y, and add so11y to e2e.

fixed #4932


test/e2e/e2e-test/docker/base-compose.yml
```diff
@@ -34,6 +34,8 @@ services:
       SW_STORAGE_INFLUXDB_URL: http://influxdb:8086
       JAVA_OPTS: >-
         -javaagent:/jacoco/jacocoagent.jar=classdumpdir=/jacoco/classes/oap,destfile=/jacoco/oap.exec,includes=org.apache.skywalking.*,excludes=org.apache.skywalking.oap.query.*:org.apache.skywalking.oap.server.core.query.*
+      SW_PROMETHEUS_FETCHER_ACTIVE: true
+      SW_TELEMETRY: prometheus
```
wu-sheng, line 38: This should be able to open self-observation. But I think we should add GraphQL query to verify the metrics really have values. NOTICE, usually, metrics value should not be 0, if there are some cases of that, we could skip those metrics, such as GC count/time.

hanahmily: done

test/e2e/e2e-test/src/test/java/org/apache/skywalking/e2e/simple/SimpleE2E.java
```diff
@@ -126,11 +132,12 @@ void setUp() throws Exception {
...
         trafficController.stop();
     }
-
+     
     @RetryableTest
     void services() throws Exception {
-        final List<Service> services = graphql.services(new ServicesQuery().start(startTime).end(now()));
+        List<Service> services = graphql.services(new ServicesQuery().start(startTime).end(now()));
 
+        services = services.stream().filter(s -> !s.getLabel().equals("oap-server")).collect(Collectors.toList());
```
wu-sheng, line 140: What is this for?

test/e2e/e2e-test/src/test/java/org/apache/skywalking/e2e/storage/StorageE2E.java
```diff
@@ -112,8 +118,9 @@ public void tearDown() {
 
     @RetryableTest
     void services() throws Exception {
-        final List<Service> services = graphql.services(new ServicesQuery().start(startTime).end(now()));
+        List<Service> services = graphql.services(new ServicesQuery().start(startTime).end(now()));
 
+        services = services.stream().filter(s -> !s.getLabel().equals("oap-server")).collect(Collectors.toList());
```
wu-sheng, line 123: Same question here. You have separated so11y verification.

wu-sheng:
@kezhenxu94 Please review

test/e2e/e2e-test/src/test/java/org/apache/skywalking/e2e/simple/SimpleE2E.java
wu-sheng: Oh, got it. I need to exclude the oap nodes.

Issue #4941, opened 2020-06-17
Multi-data source (same host:port) gets db.instance data error


HackerRookie:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [* ] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS and JRE?
1 .mysql 5.x plug: Multi-data source (same host:port) gets db.instance data error
2. ConnectionCreate5xInterceptor
   @Override
    public Object afterMethod(Class clazz, Method method, Object[] allArguments, Class<?>[] parameterTypes,
        Object ret) {
        if (ret instanceof EnhancedInstance) {
            ConnectionInfo connectionInfo = ConnectionCache.get(allArguments[0].toString(), allArguments[1].toString());
            if (connectionInfo == null) {
                connectionInfo = URLParser.parser(allArguments[4].toString());
            }
            ((EnhancedInstance) ret).setSkyWalkingDynamicField(connectionInfo);
        }
        return ret;
    }
- Which company or project?

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.

___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

PR #4942, opened 2020-06-17
Multi-data source (same host:port) gets db.instance data error #4941

HackerRookie:

- [ *] Bug fix
- [ *] New feature provided
- [ ] Improve performance

- Related issues
#4941 
___
### Bug fix
- Bug description.

- How to fix?

___


wu-sheng:
Your codes are not passing the CI. Please fix.

HackerRookie:
> I am not sure your fix. Why remove old cache? This could be performance issue.

1. The ConnectionCache here uses host:port to make the key, in the case of shard tables or multiple data sources (if a database instance has many database, hos and port will be the same) get the db.instance (intercepted from url) is not accurate database name .
2. Here's just parsing url. Performance is not responsive to ignore

wu-sheng:
I still can't get your point. An ip:port is a unique address in the network, where there are many instances? Where are they?

wu-sheng:
More importantly, why you could have a different result from URL parsing? Could you share more details about what you are doing?

HackerRookie:
> I still can't get your point. An ip:port is a unique address in the network, where there are many instances? Where are they?

image
sns01-sns0 host:port same(sns01 is database name)
sns05 - sns10 ........
sns11 - sns 16

Three sets of database services

wu-sheng:
We don't separate the database name, that is another thing. The topology is focusing on the distributed system. You are talking about the multiple databases in one DB process, right?

HackerRookie:
> We don't separate the database name, that is another thing. The topology is focusing on the distributed system. You are talking about the multiple databases in one DB process, right?

Right

wu-sheng:
Right now, we don't have the intent to separate the database.  I can't tell whether we should or shouldn't, but either of them is not the reason, you break the cache.

If you want to add the database entity as a part of topology and the analysis perspective, you should change the cache to include `host:port:db-name`, and also change the peer of the span, which could make the database metrics in the DB dashboard separated too.

HackerRookie:
> Right now, we don't have the intent to separate the database. I can't tell whether we should or shouldn't, but either of them is not the reason, you break the cache.
> 
> If you want to add the database entity as a part of topology and the analysis perspective, you should change the cache to include `host:port:db-name`, and also change the peer of the span, which could make the database metrics in the DB dashboard separated too.

image
In the case of multiple data sources, tags db.instance is currently wrong to take the value of the code.

wu-sheng:
> In the case of multiple data sources, tags db.instance is currently wrong to take the value of the code.

What do you mean wrong? What should it be?

HackerRookie:
> > In the case of multiple data sources, tags db.instance is currently wrong to take the value of the code.
> 
> What do you mean wrong? What should it be?

because of sns01 sns02 sns03 sns04 host:port same. So there's only one database name in the collection cache. Should be displayed separately

wu-sheng:
I think that is caused by the usage of `Tags#DB_INSTANCE`. It saved the database name for now. We haven't given a very specific definition, but it seems like they are treated as the database name in many plugins, I mean for now. 

Like this,
image

Issue #4950, opened 2020-06-20
Support Quasar fiber tracing


zhaoyuguang:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [x] Requirement
- [ ] Feature or performance improvement

___
### Requirement or improvement

Support Quasar fiber tracing

PR #4951, opened 2020-06-20
Support Quasar fiber tracing

zhaoyuguang:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/4950

### DIFF
--- a/test/plugin/scenarios/quasar-scenario/config/expectedData.yaml
+++ b/test/plugin/scenarios/quasar-scenario/config/expectedData.yaml
@@ -1,0 +1,70 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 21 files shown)


test/plugin/scenarios/quasar-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,70 @@
...
+      endTime: nq 0
+      componentId: 12
+      isError: false
+      spanType: Exit
+      peer: localhost:8080
+      skipAnalysis: false
+      tags:
+      - {key: http.method, value: GET}
+      - {key: url, value: 'http://localhost:8080/quasar-scenario/case/ping'}
+    - operationName: QUASAR/Fiber
```
wu-sheng, line 54: Guasar Fiber is not an async bridge?

zhaoyuguang: Yes, main thread is Http entry, async is Fiber and http exit

Issue #4879, opened 2020-06-08
[plugin/lettuce] sometimes trace doesn‘t show component type


dagmom:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Bug
- Which version of SkyWalking, OS and JRE?
8, macos, jre8

___
![image](https://user-images.githubusercontent.com/43675197/84052238-df2a0f00-a9e2-11ea-9ad9-1ab2c66305e5.png)

wu-sheng:
Check the interceptor codes. Maybe they miss some set(s). We don't have any special process about this in the agent core and backend.

jackiepon31:
yep, i've checked the inteceptors under the luttuce plugin , they really did miss some sets on the local spans. Maybe we can add some on that and update the plugin。

wu-sheng:
Please send a pull request to fix them.

PR #4954, opened 2020-06-20
fix lettuce plugin sometimes trace doesn‘t show span layer

J-Cod3r:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
fixed [#4879](https://github.com/apache/skywalking/issues/4879)


apm-sniffer/apm-sdk-plugin/lettuce-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/lettuce/v5/AsyncCommandMethodInterceptor.java
```diff
@@ -40,6 +42,8 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
         String operationName = "Lettuce/" + asyncCommand.getType().name();
         AbstractSpan span = ContextManager.createLocalSpan(operationName + "/onComplete");
         span.setComponent(ComponentsDefine.LETTUCE);
+        Tags.DB_TYPE.set(span, "Redis");
```
wu-sheng, line 45: I think you should create `Tags#CACHE_TYPE`. Redis is not DB from my understanding.

wu-sheng: Ignore me, I noticed the old codes showing this as a DB.

PR #4958, opened 2020-06-22
add brpc-java plugin

IanCao:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___


wu-sheng:
What is a surprise, you are doing this.

oap-server/server-bootstrap/src/main/resources/component-libraries.yml
```diff
@@ -302,6 +302,9 @@ InfluxDB:
 influxdb-java:
   id: 90
   languages: Java
+brpc-java:
```
wu-sheng, line 305: Is brpc having a logo?

oap-server/server-core/src/test/resources/component-libraries.yml
```diff
@@ -266,6 +266,9 @@ InfluxDB:
 influxdb-java:
   id: 90
   languages: Java
+brpc-java:
```
wu-sheng, line 269: I think test file actually doesn't need to change every time.

apm-sniffer/apm-sdk-plugin/brpc-java-plugin/src/main/java/org.apache.skywalking.apm.plugin.brpc/define/ClientInstrumentation.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ClassInstanceMethodsEnhancePluginDefine;
+import org.apache.skywalking.apm.agent.core.plugin.match.ClassMatch;
+import org.apache.skywalking.apm.agent.core.plugin.match.NameMatch;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+public class ClientInstrumentation extends ClassInstanceMethodsEnhancePluginDefine {
+
+    public static final String INTERCEPTOR_CLASS = "org.apache.skywalking.apm.plugin.brpc.ClientInterceptor";
+    public static final String ENHANCE_CLASS = "com.baidu.brpc.interceptor.LoadBalanceInterceptor";
```
wu-sheng, line 34: As this is still in the Baidu package name, please rename the plugin to `baidu-brpc-plugin`. Because brpc is in the ASF incubator, but it seems don't very active. And `java` is not needed, this is a Java repo

wu-sheng:
Also, I noticed, there is brpc 3.x, do you plan to support it too?

oap-server/server-bootstrap/src/main/resources/component-libraries.yml
IanCao: up to now, i did not find it

oap-server/server-core/src/test/resources/component-libraries.yml
IanCao: done

apm-sniffer/apm-sdk-plugin/brpc-java-plugin/src/main/java/org.apache.skywalking.apm.plugin.brpc/define/ClientInstrumentation.java
IanCao: done

IanCao:
> Also, I noticed, there is brpc 3.x, do you plan to support it too?

ok, let me investigate

IanCao:
> > Also, I noticed, there is brpc 3.x, do you plan to support it too?
> 
> ok, let me investigate

This seems a bit difficult 
image

loadBalance and netty `writeAndFlush` are in a same method

wu-sheng:
Let's skip that version for now. We could copy/paste that method and change codes, but as this plugin is not that popular, let's wait.

IanCao:
116 testcase have passed,one fail
but it's no valid error msg  @wu-sheng

PR #4985, opened 2020-06-29
Support `logic-endpoint` series tag(s).

### LINKED ISSUE #4976: Introduce the `logic-endpoint` series tag(s) and analysis in the backend.
Right now, endpoint analysis is based on the operation name of the entry span. I found there are some special cases, we need more than one endpoint for a single request analysis, typically [GraphQL](https://graphql.org/)

The `/graphql` endpoint should be detected, which is reasonable, and it matches the real traffic. So, service, instance, and endpoint are matched. But, at the same time, GraphQL could be like this,
```graphql
{
  r1:readMetricsValues(condition : {
    name: "service_cpm",
    entity: {
      scope: Service,
      serviceName: "projectD.business-zone",
      normal: true,
    }
  }, 
  duration: {step:MINUTE, start: "2020-05-28 1040", end: "2020-05-28 1050"}
  ) {
    label,
    values {
      values {
        id,
        value
      }
    }
  },
  
  r2:readMetricsValues(condition : {
    name: "service_cpm",
    entity: {
      scope: Service,
      serviceName: "projectA.business-zone",
      normal: true,
    }
  }, 
  duration: {step:MINUTE, start: "2020-05-28 1040", end: "2020-05-28 1050"}
  ) {
    label,
    values {
      values {
        id,
        value
      }
    }
  }
}
```

The user would expect the `readMetricsValues` endpoint detected, with 2 calls sharding different(or same) latency.

Base on the above scenario, I want to add `x-le`(extension logic endpoint) series tag(s)
- Tag key `x-le`, value is JSON format. `name:xxx, latency:yyy`.

Same as other tags in SkyWalking, we could have multiple tags with the same key. Such as in the above case, there should be two tags with values like this
```
     * Tag, x-le(extension logic endpoint) series tag. Value is JSON format.
     * <pre>
     * {
     *   "name": "GraphQL-service",
     *   "latency": 100,
     *   "status": true
     * }
     * </pre>
     *
     * Also, could use value to indicate this local span is representing a logic endpoint.
     * <pre>
     * {
     *   "logic-span": true
     * }
     * </pre>
```

NOTICE, this endpoint traffic would not be added into the instance or service level, to keep those two traffic data as same as the real network call. And there is no endpoint dependency for logic endpoints.

wu-sheng:
Resolves #4976

At the same time, FYI @apache/skywalking-committers. In this PR, I added `MultiScopesAnalysisListenerTest`, which is a good way to study how OAP analysis trace segments.

### DIFF
--- a/docs/en/guides/Java-Plugin-Development-Guide.md
+++ b/docs/en/guides/Java-Plugin-Development-Guide.md
@@ -170,7 +170,7 @@ The value of `x-le` should be JSON format, with two options.
 ```json
 {
   "name": "GraphQL-service",
-  latency": 100,
+  "latency": 100,
   "status": true
 }
 ```
--- a/docs/en/guides/Java-Plugin-Development-Guide.md
+++ b/docs/en/guides/Java-Plugin-Development-Guide.md
@@ -138,7 +138,7 @@ Create ExitSpan by operation name(e.g. service name, uri) and new **ContextCarri
      */
     AbstractSpan setOperationName(String endpointName);
 ```
-Besides set operation name, tags and logs, two attributes shoule be set, which are component and layer, 
+Besides setting operation name, tags and logs, two attributes should be set, which are component and layer, 
 especially for EntrySpan and ExitSpan
 
 SpanLayer is the catalog of span. Here are 5 values:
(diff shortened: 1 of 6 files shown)


wu-sheng:
Plugin development doc has been updated, as they need to know which tags are special, and how should they use.

PR #5002, opened 2020-07-01
fix shardingsphere plugin span to DB type

swift1911:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
shardingsphere plugin dosn't set the span type to DB.So it show that the unknown span in the web.

image

- How to fix?

___


swift1911:
> As a local span, it should not be as a DB layer.

Displaying unknown on the page is bad for the user experience. It will make user confuse about the Span.Do you think we should display the type of span is Local rather than Unknown?

wu-sheng:
This thing is called `span layer`. There is no value for the local span.

```Java
public enum SpanLayer {
    DB(1), RPC_FRAMEWORK(2), HTTP(3), MQ(4), CACHE(5);
}
```

swift1911:
> This thing is called `span layer`. There is no value for the local span.
> 
> ```java
> public enum SpanLayer {
>     DB(1), RPC_FRAMEWORK(2), HTTP(3), MQ(4), CACHE(5);
> }
> ```

Yes.There is no value for the local span.Do you think we should add a span type for local span because there are many local span just like SpringAsync,Shardingsphere shows unknown.

wu-sheng:
> Do you think we should add a span type for local span because there are many local span

To say what? Local?

PR #5021, opened 2020-07-03
Fix bug caused by setting SW_OAL_ENGINE_DEBUG=Y.

Ax1an:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

  Only envoy related classes are generated after setting SW_OAL_ENGINE_DEBUG=Y at system env.

- How to fix?

  Create a oal-rt folder only when the first oal script is loaded.

___

### DIFF
--- a/oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/OALRuntime.java
+++ b/oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/OALRuntime.java
@@ -99,6 +99,7 @@ public class OALRuntime implements OALEngine {
         "data2Map",
         "map2Data"
     };
+    private static boolean IS_RT_TEMP_FOLDER_INIT_COMPLETED = false;
 
     private final OALDefine oalDefine;
     private final ClassPool classPool;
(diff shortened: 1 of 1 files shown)


oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/OALRuntime.java
```diff
@@ -135,7 +136,10 @@ public void setDispatcherListener(DispatcherDetectorListener listener) throws Mo
 
     @Override
     public void start(ClassLoader currentClassLoader) throws ModuleStartException, OALCompileException {
-        prepareRTTempFolder();
+        if (!IS_RT_TEMP_FOLDER_INIT_COMPLETED) {
```
wu-sheng, line 139: What do you mean? This is always true.

Ax1an: Create a oal-rt folder only when the first oal script is loaded.

Before fix:
image

After fix:
image

PR #5029, opened 2020-07-04
Add Opencensus receiver

hanahmily:
Changes of opencensus receiver:

 * Add oc-receiver module and document how to setup 
 * Create the prometheus->meter-system converter to support oc-receiver and prometheus-fethcer at the same time.

### DIFF
--- a/docs/en/setup/backend/backend-receivers.md
+++ b/docs/en/setup/backend/backend-receivers.md
@@ -13,6 +13,7 @@ We have following receivers, and `default` implementors are provided in our Apac
 1. **receiver-profile**. gRPC services accept profile task status and snapshot reporter. 
 1. **receiver_zipkin**. See [details](#zipkin-receiver).
 1. **receiver_jaeger**. See [details](#jaeger-receiver).
+1. **receiver-oc**. See [details](#oc-receiver).
 
 The sample settings of these receivers should be already in default `application.yml`, and also list here
 ```yaml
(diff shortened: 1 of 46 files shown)


docs/en/setup/backend/backend-receivers.md
```diff
@@ -109,4 +110,21 @@ receiver_jaeger:
     gRPCPort: ${SW_RECEIVER_JAEGER_PORT:14250}
 ``` 
 
-NOTICE, Jaeger receiver is only provided in `apache-skywalking-apm-x.y.z.tar.gz` tar.
\ No newline at end of file
+NOTICE, Jaeger receiver is only provided in `apache-skywalking-apm-x.y.z.tar.gz` tar.
+
+## Opencensus receiver
+
+Opencensus receiver supports to ingest agent metrics by meter-system. OAP can load the configuration at bootstrap.
```
wu-sheng, line 117: I think you should mention the OpenTelemetry. For a wider known concept.

docs/en/setup/backend/backend-receivers.md
```diff
@@ -109,4 +110,21 @@ receiver_jaeger:
...
 ``` 
 
-NOTICE, Jaeger receiver is only provided in `apache-skywalking-apm-x.y.z.tar.gz` tar.
\ No newline at end of file
+NOTICE, Jaeger receiver is only provided in `apache-skywalking-apm-x.y.z.tar.gz` tar.
+
+## Opencensus receiver
+
+Opencensus receiver supports to ingest agent metrics by meter-system. OAP can load the configuration at bootstrap. 
+If the new configuration is not well-formed, OAP fails to start up. The files are located at `$CLASSPATH/oc-rules`.
```
wu-sheng, line 118: With the possibility of these files added continuously, should we consider to use `application.yml` to activate the files? A similar question about the fetcher too. Because these files are located in the image together.

oap-server/server-bootstrap/src/main/resources/oc-rules/oap.yaml
```diff
@@ -0,0 +1,296 @@
...
+#    "PT20.345S" -- parses as "20.345 seconds"
+#    "PT15M"     -- parses as "15 minutes" (where a minute is 60 seconds)
+#    "PT10H"     -- parses as "10 hours" (where an hour is 3600 seconds)
+#    "P2D"       -- parses as "2 days" (where a day is 24 hours or 86400 seconds)
+#    "P2DT3H4M"  -- parses as "2 days, 3 hours and 4 minutes"
+#    "P-6H3M"    -- parses as "-6 hours and +3 minutes"
+#    "-P6H3M"    -- parses as "-6 hours and -3 minutes"
+#    "-P-6H+3M"  -- parses as "+6 hours and -3 minutes"
+# </pre>
+metricsRules:
```
wu-sheng, line 31: How to separate the config for different services? You have `staticConfig` in the fetcher case

wu-sheng: I think should consider filtering through service name from the oc format?

docs/en/setup/backend/backend-receivers.md
wu-sheng: How about adding a name for which rule file? And activate them through the system env, like ALS analyzer.

hanahmily: Make sense to me. Let me implement it in another PR. I touched too many files in this one. 

and I also want to introduce the `configuration` model to support dynamic rules update.

PR #5030, opened 2020-07-04
support Graphql Logic endpoint

IanCao:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___


apm-sniffer/apm-sdk-plugin/graphql-plugin/graphql-12.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/graphql/v12/GraphqlInterceptor.java
```diff
@@ -0,0 +1,85 @@
...
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) throws Throwable {
+        ExecutionStrategyParameters parameters = (ExecutionStrategyParameters) allArguments[1];
+        if (parameters == null || parameters.getParent().getPath() != ExecutionPath.rootPath()) {
+            return ret;
+        }
+        String name = parameters.getField().getSingleField().getName();
+        long latency = System.currentTimeMillis() - (long) objInst.getSkyWalkingDynamicField();
+        String info = buildLogicEndpointTagInfo(name, latency, null);
+        AbstractSpan span = ContextManager.firstSpan();
+        if (span == null || !span.isEntry()) {
```
wu-sheng, line 55: How a GraphQL span could not be the first span? Or not an entry span? A little confused.

wu-sheng:
Plugin test and CI fail, please recheck.

kezhenxu94:
`@author` tag and Chinese comments should be removed

IanCao:
> There're `@author` tags and Chinese comments

I've removed them all

apm-sniffer/apm-sdk-plugin/graphql-plugin/graphql-12.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/graphql/v12/GraphqlInterceptor.java
IanCao: i've used a localSpan to represent a Grqphql subquery

test/plugin/scenarios/graphql-12.x-scenario/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+    <pluginRepositories>
+        <pluginRepository>
+            <id>spring-snapshots</id>
```
wu-sheng, line 118: Why need this?

test/plugin/scenarios/graphql-8.x-scenario/pom.xml
```diff
@@ -0,0 +1,125 @@
...
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+    <pluginRepositories>
+        <pluginRepository>
+            <id>spring-snapshots</id>
```
wu-sheng, line 117: Same here.

test/plugin/scenarios/graphql-9.x-scenario/pom.xml
```diff
@@ -0,0 +1,125 @@
...
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+    <pluginRepositories>
+        <pluginRepository>
+            <id>spring-snapshots</id>
```
wu-sheng, line 117: Same here.

test/plugin/scenarios/graphql-9.x-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,92 @@
...
+      spanLayer: Unknown
+      startTime: nq 0
+      endTime: nq 0
+      componentId: 92
+      isError: false
+      spanType: Local
+      peer: ''
+      skipAnalysis: false
+      tags:
+      - {key: x-le, value: '{logic-span=true}'}
```
kezhenxu94, line 35: Such kind of tags are not processable by the backend, according to [the doc](https://github.com/apache/skywalking/blob/310b8857b2266768ccd9e1338da13c86c313ed3f/docs/en/guides/Java-Plugin-Development-Guide.md#extension-logic-endpoint-tag-key-x-le)

> The value of x-le should be JSON format

@IanCao  did you check this plugin with the backend in a whole? FYI @wu-sheng

apm-sniffer/apm-sdk-plugin/graphql-plugin/graphql-9.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/graphql/v9/GraphqlInterceptor.java
```diff
@@ -0,0 +1,78 @@
...
+    private void dealException(Throwable throwable) {
+        AbstractSpan span = ContextManager.activeSpan();
+        span.errorOccurred();
+        span.log(throwable);
+    }
+
+    private String buildLogicEndpointSpan() {
+        Map<String, Object> logicEndpointSpan = new HashMap<>();
+        logicEndpointSpan.put("logic-span", true);
+        return logicEndpointSpan.toString();
```
kezhenxu94, line 76: The value of `x-le` should be JSON format, this is obviously not the case by simply calling `toString` on a `Map`

test/plugin/scenarios/graphql-9.x-scenario/config/expectedData.yaml
wu-sheng: Why not? I added this new feature :) It is here

https://github.com/apache/skywalking/blob/310b8857b2266768ccd9e1338da13c86c313ed3f/docs/en/guides/Java-Plugin-Development-Guide.md#extension-logic-endpoint-tag-key-x-le

kezhenxu94: I mean `{logic-span=true}` is not JSON format, did you process this kind of format @wu-sheng ? I mean the `=`, it's not JSON format

kezhenxu94: @wu-sheng do you mean this, I didn't see any clue that the map format `{logic-span=true}` is also supported

image

wu-sheng: Oh, I see, I didn't check that part.

PR #5032, opened 2020-07-05
add python components

alonelaval:

- [ ] Bug fix
- [* ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___

[add Flask plugin](https://github.com/apache/skywalking-python/pull/31)


alonelaval:
> Does this require server side mapping?

yes.

wu-sheng:
Then you need to update the config file and add the mapping. You only change the component definition.

kezhenxu94:
> > Does this require server side mapping?
> 
> yes.

@alonelaval you misunderstood the meaning, we don't need server mapping for the two components actually

alonelaval:
> > > Does this require server side mapping?
> > 
> > 
> > yes.
> 
> @alonelaval you misunderstood the meaning, we don't need server mapping for the two components actually

OK

Issue #5036, opened 2020-07-06
@Tag returnedObject bug


zoumingzm:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Bug  
___
### Bug
- Which version of SkyWalking, OS and JRE?
7.0.0

- Which company or project?
powersi.com    

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.
Only nomal object is supported.  
List, Array, Map is not supported.  
___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.  
Preparation returnContext and usage returnContext are inconsistent needs to be fixed.
--
example:
return List or Array :  returnedObject.['key'] or returnedObject.key. or returnedObject.key.fieldName
return Map: returnedObject.['key'] or returnedObject.key. 
--
before fix:
```
    public static Map<String, Object> evaluationReturnContext(Object ret)  {
        Map<String, Object> context = new HashMap<>();
        Field[] fields = ret.getClass().getDeclaredFields();
        for (Field field : fields) {
            field.setAccessible(true);
            try {
                context.put(field.getName(), field.get(ret));
            } catch (Exception e) {
                logger.debug("evaluationReturnContext error, ret is {}, exception is {}", ret, e.getMessage());
            }
        }
        return context;
    }
```
--
after fix:
```
    public static Map<String, Object> evaluationReturnContext(Object ret)  {
        Map<String, Object> context = new HashMap<>();
        if (ret != null){
            if (ret instanceof List) {
                List retList = (List) ret;
                int retLength = retList.size();
                for (int i = 0; i < retLength; i++) {
                    context.put(String.valueOf(i), retList.get(i));
                }
            } else if (ret.getClass().isArray()) {
                int length = Array.getLength(ret);
                for (int i = 0; i < length; i++) {
                    context.put(String.valueOf(i), Array.get(ret, i));
                }
            } else if (ret instanceof Map) {
                context.putAll((Map) ret);
            } else {
                Field[] fields = ret.getClass().getDeclaredFields();
                for (Field field : fields) {
                    field.setAccessible(true);
                    try {
                        context.put(field.getName(), field.get(ret));
                    } catch (Exception e) {
                        logger.debug("evaluationReturnContext error, ret is {}, exception is {}", ret, e.getMessage());
                    }
                }
            }
        }
        return context;
    }
```

- Test
return value:
Map: {"username":"张三", "age":18, "info": "username=张三&age=18"}
[...]

PR #5045, opened 2020-07-07
fix @Tag returnedObject bug #5036

zoumingzm:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
#5036 

___
### Bug fix
- Bug description.
Tag returnedObject cannot support List, Array, Map
- How to fix?
modify method: CustomizeExpression.evaluationReturnContext(ret)


apm-sniffer/apm-toolkit-activation/apm-toolkit-trace-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/util/TagUtil.java
```diff
@@ -38,6 +38,6 @@ public static void tagReturnSpanSpan(final AbstractSpan span, final Map<String,
 
     public static Boolean isReturnTag(String expression) {
         String[] es = expression.split("\\.");
-        return es.length == 2 && "returnedObj".equals(es[0]);
+        return "returnedObj".equals(es[0]);
```
kezhenxu94, line 41: Are there possibilities that the method returns a `String` and the user wants to use that as a tag? In such case, `es.length == 1`

zoumingzm: > Are there possibilities that the method returns a `String` and the user wants to use that as a tag? In such case, `es.length == 1`

Uh, there is a problem here. When returning objects such as numbers, strings, Boolean values, characters, etc., they need to be processed, otherwise the tag value will be null. Whether to handle only common types such as numbers, strings, Boolean values, and characters.

kezhenxu94: > > Are there possibilities that the method returns a `String` and the user wants to use that as a tag? In such case, `es.length == 1`
> 
> Uh, there is a problem here. When returning objects such as numbers, strings, Boolean values, characters, etc., they need to be processed, otherwise the tag value will be null. Whether to handle only common types such as numbers, strings, Boolean values, and characters.

What about simply call `toString` on the `returnedObj`

wu-sheng:
Any update about this PR?

apm-sniffer/apm-toolkit-activation/apm-toolkit-trace-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/util/TagUtil.java
zoumingzm: > > > Are there possibilities that the method returns a `String` and the user wants to use that as a tag? In such case, `es.length == 1`
> > 
> > 
> > Uh, there is a problem here. When returning objects such as numbers, strings, Boolean values, characters, etc., they need to be processed, otherwise the tag value will be null. Whether to handle only common types such as numbers, strings, Boolean values, and characters.
> 
> What about simply call `toString` on the `returnedObj`

That's a good idea.

zoumingzm:
> Any update about this PR?

Yes, there are still some modifications

apm-sniffer/apm-toolkit-activation/apm-toolkit-trace-activation/src/test/java/org/apache/skywalking/apm/toolkit/activation/trace/TagAnnotationTest.java
```diff
@@ -138,6 +144,145 @@ public void testTraceWithTags() throws Throwable {
 
     }
 
+    @Test
+    public void testTraceWithList() throws Throwable {
```
wu-sheng, line 148:
```suggestion
    public void testTraceWithReturnList() throws Throwable {
```
I think the test method names are not accurate? Please confirm. Sam as other methods.

PR #5064, opened 2020-07-08
Better Vert.x endpoint names

BFergerson:
- [ ] Bug fix
- [ ] New feature provided
- [X] Improve performance

___

PR contains two changes. The first is to include the HTTP request method in the endpoint name. A GET endpoint will now have an endpoint like `{GET}/orders`, instead of `/orders`. This follows the same style used in the Spring MVC plugin.

The second change is to implement endpoint name grouping so endpoints like `/order/1` and `/order/2` will have an endpoint name `/order/:id` instead of creating a new endpoint for each unique URL.

Combine those two changes and you get `{POST}/order/:id`, which I believe correctly reflects what an endpoint is expected to be.


test/plugin/scenarios/vertx-web-3.54minus-scenario/config/expectedData.yaml
```diff
@@ -97,10 +113,37 @@ segmentItems:
...
+              - {key: url, value: '/vertx-web-3_54minus-scenario/case/healthCheck'}
+              - {key: status_code, value: '200'}
+            refs:
+              - {parentEndpoint: '{GET}/vertx-web-3_54minus-scenario/case/web-case', networkAddress: 'localhost:8080',
+                 refType: CrossProcess, parentSpanId: 3, parentTraceSegmentId: not null,
+                 parentServiceInstance: not null, parentService: vertx-web-3.54minus-scenario,
+                 traceId: not null}
+      - segmentId: not null
+        spans:
+          - {operationName: org.apache.skywalking.apm.testcase.vertxweb.controller.VertxWebController$$Lambda$.handle(RoutingContext),
```
wu-sheng, line 125: This span expected data seems in a different format.

test/plugin/scenarios/vertx-web-3.6plus-scenario/config/expectedData.yaml
```diff
@@ -97,10 +113,37 @@ segmentItems:
...
+              - {key: url, value: '/vertx-web-3_6plus-scenario/case/healthCheck'}
+              - {key: status_code, value: '200'}
+            refs:
+              - {parentEndpoint: '{GET}/vertx-web-3_6plus-scenario/case/web-case', networkAddress: 'localhost:8080',
+                 refType: CrossProcess, parentSpanId: 3, parentTraceSegmentId: not null,
+                 parentServiceInstance: not null, parentService: vertx-web-3.6plus-scenario,
+                 traceId: not null}
+      - segmentId: not null
+        spans:
+          - {operationName: org.apache.skywalking.apm.testcase.vertxweb.controller.VertxWebController$$Lambda$.handle(RoutingContext),
```
wu-sheng, line 125: Same format issue here.

test/plugin/scenarios/vertx-web-3.54minus-scenario/config/expectedData.yaml
BFergerson: These operation names correspond to the lambdas which handle the response message. Like here: https://github.com/apache/skywalking/blob/6035f4869550009918b416df7811661c882f3de1/test/plugin/scenarios/vertx-web-3.6plus-scenario/src/main/java/org/apache/skywalking/apm/testcase/vertxweb/controller/VertxWebController.java#L47

What else would you call it? If a class and method name can be determined you get: https://github.com/apache/skywalking/blob/6035f4869550009918b416df7811661c882f3de1/test/plugin/scenarios/vertx-web-3.6plus-scenario/config/expectedData.yaml#L152. I can't get a method name for a lambda.

BFergerson: Actually I think I'm wrong here. I would expect an operation name with `handleWebCase` in it somewhere. I'm looking into it but FYI it's not an issue introduced with this PR.

wu-sheng: It is good to know you find another issue, should I wait for you to fix this? My thought was just about the format, which doesn't separate the lines for each attribute.

PR #5084, opened 2020-07-11
#4907 - Duplicate Spring MVC endpoint fix

### LINKED ISSUE #4907: one url has two endpoint ?

- [ ] Question or discussion

___
### Question
- like #3675,  [https://github.com/apache/skywalking/issues/3675](url)
  skywalking7.0 , springboot1.5.8 ,  app.war is put in external tomcat deploy。

  My URL is http://127.0.0.1:8082/my/v2/ad  。application context is "**my**", and endpoint is "**/v2/ad**" and "**{POST}/v2/ad**"

 my java code like this
`@RequestMapping(path = "/v2/ad", method = {RequestMethod.POST})`

捕获
捕获2
捕获3

BFergerson:
Why submit this pull request?
- [X] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
fixes #4907

___
### Bug fix
- Bug description.

The `requestURL` is updated after `addPathMapping`. This means it will use `{POST}/url` the first time the endpoint is hit and then fall back to `/url` every time it's hit afterward.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-commons/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/commons/interceptor/AbstractMethodInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-commons/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/commons/interceptor/AbstractMethodInterceptor.java
@@ -82,11 +82,9 @@ public abstract class AbstractMethodInterceptor implements InstanceMethodsAround
             EnhanceRequireObjectCache pathMappingCache = (EnhanceRequireObjectCache) objInst.getSkyWalkingDynamicField();
             String requestURL = pathMappingCache.findPathMapping(method);
             if (requestURL == null) {
-                requestURL = getRequestURL(method);
-                pathMappingCache.addPathMapping(method, requestURL);
-                requestURL = getAcceptedMethodTypes(method) + pathMappingCache.findPathMapping(method);
+                pathMappingCache.addPathMapping(method, getRequestURL(method));
             }
-            operationName = requestURL;
+            operationName = getAcceptedMethodTypes(method) + pathMappingCache.findPathMapping(method);
         }
 
         HttpServletRequest request = (HttpServletRequest) ContextManager.getRuntimeContext()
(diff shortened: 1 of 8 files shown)


apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-commons/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/commons/interceptor/AbstractMethodInterceptor.java
```diff
@@ -82,11 +82,9 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
             EnhanceRequireObjectCache pathMappingCache = (EnhanceRequireObjectCache) objInst.getSkyWalkingDynamicField();
             String requestURL = pathMappingCache.findPathMapping(method);
             if (requestURL == null) {
-                requestURL = getRequestURL(method);
-                pathMappingCache.addPathMapping(method, requestURL);
-                requestURL = getAcceptedMethodTypes(method) + pathMappingCache.findPathMapping(method);
+                pathMappingCache.addPathMapping(method, getRequestURL(method));
             }
-            operationName = requestURL;
+            operationName = getAcceptedMethodTypes(method) + pathMappingCache.findPathMapping(method);
```
wu-sheng, line 87: > The requestURL is updated after addPathMapping. This means it will use {POST}/url the first time the endpoint is hit and then fall back to /url every time it's hit afterward.

I think you just need move original L86 after L87. The way you changing right now, will make the cache invalid. RIght?

BFergerson: I've taken a look at this and moving L87 to L86 is not ideal because of this line:
https://github.com/apache/skywalking/blob/c5972c2b7da417ca6fe2e0e76d1eb5dac77f669c/apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-commons/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/commons/PathMappingCache.java#L52

This appends whatever URL was cached with a method to `classPath`. Meaning a method annotated with `@RequestMapping("/test", method = RequestMethod.GET)` which falls under a class annotated with `@RequestMapping("/case")` will result in the URL `/case/{GET}/test` being cached to the method.

wu-sheng: My point is, there is a logic error above. execute `pathMappingCache.findPathMapping(method)` twice if cache found.

PR #5086, opened 2020-07-12
Add support for spring webclient

dengliming:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___

Hi~ Now we usually use WebClient instead of AsyncRestteamplate, And It't one part of spring-webflux's features, but we only support the server side now. So I add the client side support. Please see if that makes sense.


test/plugin/scenarios/webflux-scenario/config/expectedData.yaml
```diff
@@ -193,6 +193,66 @@ segmentItems:
...
+      tags:
+      - {key: url, value: 'http://localhost:18080/testcase/webclient/test1'}
+      - {key: http.method, value: GET}
+      - {key: status_code, value: '200'}
+      refs:
+      - {parentEndpoint: /projectA/testcase, networkAddress: 'localhost:18080', refType: CrossProcess,
+        parentSpanId: 9, parentTraceSegmentId: not null, parentServiceInstance: not null, parentService: not null, traceId: not null}
+  - segmentId: not null
+    spans:
+    - operationName: /testcase/annotation/healthCheck
```
wu-sheng, line 219: Why health check in the expected span?

dengliming: Thanks for quick reply. I use it for webclient testcase. The following code：
```
@RequestMapping("/testcase/webclient/test1")
    public String test1() {
        Mono<String> resp = WebClient.builder().exchangeStrategies(ExchangeStrategies.builder()
                .build())
                .build()
                .get().uri("http://" + hostBAddress + "/testcase/annotation/healthCheck")
                .retrieve().bodyToMono(String.class);
        resp.subscribe();
        return "ok";
    }
```
Would you mind sharing your tips here?

wu-sheng: Don't access health check by codes. It is designed for the test framework. If you need a uri to access, add a new one.

dengliming: Thanks. fixed at last commits.

apm-sniffer/apm-sdk-plugin/spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/DefaultExchangeFunctionInterceptor.java
```diff
@@ -0,0 +1,105 @@
...
+            ((EnhancedInstance) clientRequest).setSkyWalkingDynamicField(contextCarrier);
+        }
+        span.prepareForAsync();
+        ContextManager.stopSpan(span);
+        objInst.setSkyWalkingDynamicField(span);
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                              Object ret) throws Throwable {
```
wu-sheng, line 69:
```suggestion
    public void beforeMethod(EnhancedInstance objInst, 
                                                 Method method, Object[] allArguments, 
                                                 Class<?>[] argumentsTypes,
                                                 MethodInterceptResult result) throws Throwable {
```

apm-sniffer/apm-sdk-plugin/spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/DefaultExchangeFunctionInterceptor.java
```diff
@@ -0,0 +1,105 @@
...
+        span.setComponent(ComponentsDefine.SPRING_WEBCLIENT);
+        Tags.URL.set(span, requestURL.getScheme() + "://" + requestURL.getHost() + ":" + port + formatURIPath);
+        Tags.HTTP.METHOD.set(span, httpMethod.toString());
+        SpanLayer.asHttp(span);
+
+        if (clientRequest instanceof EnhancedInstance) {
+            ((EnhancedInstance) clientRequest).setSkyWalkingDynamicField(contextCarrier);
+        }
+        span.prepareForAsync();
+        ContextManager.stopSpan(span);
```
arugal, line 63: @dengliming  It seems to have called the `stopSpan` method too early, right?

apm-sniffer/apm-sdk-plugin/spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/DefaultExchangeFunctionInterceptor.java
```diff
@@ -0,0 +1,105 @@
...
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                             MethodInterceptResult result) throws Throwable {
+        final ClientRequest clientRequest = (ClientRequest) allArguments[0];
+        final URI requestURL = clientRequest.url();
+        final HttpMethod httpMethod = clientRequest.method();
+        final ContextCarrier contextCarrier = new ContextCarrier();
+
+        int port = getPort(requestURL);
+        String remotePeer = requestURL.getHost() + ":" + port;
+        String formatURIPath = getURIPath(requestURL);
```
arugal, line 51: Recommend `methodName` + `Path`

test/plugin/scenarios/webflux-scenario/config/expectedData.yaml
```diff
@@ -193,6 +193,66 @@ segmentItems:
...
+      tags:
+      - {key: url, value: 'http://localhost:18080/testcase/webclient/ping'}
+      - {key: http.method, value: GET}
+      - {key: status_code, value: '200'}
+      refs:
+      - {parentEndpoint: /testcase/webclient/ping, networkAddress: 'localhost:18080', refType: CrossProcess,
+        parentSpanId: 0, parentTraceSegmentId: not null, parentServiceInstance: not null, parentService: not null, traceId: not null}
+  - segmentId: not null
+    spans:
+    - operationName: /testcase/webclient/ping
```
arugal, line 240: It should be and `Entry/testcase/webclient/test1` belong to the same segment, right?

wu-sheng: This is an exit, the entry is the above one. Isn't it?

dengliming:
> @dengliming I think you haven't updated `Supported-list.md`. Could you update it?

Only test 2.1.x, 2.2.x and 2.3 not test yet.

test/plugin/scenarios/webflux-scenario/config/expectedData.yaml
wu-sheng: @dengliming I think the issue here, this exit span isn't in the main segment.

wu-sheng: Segment should be 
- /testcase/webclient/test1
   -  /testcase/webclient/ping

test/plugin/scenarios/webflux-scenario/webflux-projectA-scenario/src/main/java/org/apache/skywalking/apm/testcase/sc/webflux/projectA/controller/TestController.java
```diff
@@ -43,6 +43,7 @@ public String testcase() throws IOException {
         visit("http://" + hostBAddress + "/testcase/route/error");
         visit("http://" + hostBAddress + "/notFound");
         visit("http://" + hostBAddress + "/testcase/annotation/mono/hello");
+        visit("http://" + hostBAddress + "/testcase/webclient/test1");
```
wu-sheng, line 46: Where is the exit span of this?

test/plugin/scenarios/webflux-scenario/config/expectedData.yaml
wu-sheng: Wait a moment, @arugal , this seems a bug of `refType` and test tool? @dengliming Are these two exit spans both in across-thread ref?

wu-sheng: No, it isn't. @dengliming please help us confirming the case. From my reading of your codes. 

- /testcase/webclient/test1
  - /testcase/webclient/ping

This is required, your expected data doesn't show parent span or ref, there is something missing.

dengliming: Let me check it.

dengliming: I think these two span not in same thread 
```
/testcase/webclient/test1
/testcase/webclient/ping
```
any idea to transfer the parent context?

dengliming: here 
org.springframework.web.reactive.function.client.DefaultWebClient.DefaultRequestBodyUriSpec#exchange
```
@Override
		public Mono<ClientResponse> exchange() {
			ClientRequest request = (this.inserter != null ?
					initRequestBuilder().body(this.inserter).build() :
					initRequestBuilder().build());
			return Mono.defer(() -> exchangeFunction.exchange(request)
					.checkpoint("Request to " + this.httpMethod.name() + " " + this.uri + " [DefaultWebClient]")
					.switchIfEmpty(NO_HTTP_CLIENT_RESPONSE_ERROR));
		}
```
I think missing context reason is here.

dengliming: A bit tricky. seems the same as httpclient.
The right trace should be:
```
-/projectA/testcase(webflux-projectA-scenario)
	-/testcase/annotation/success (http-client)(webflux-projectA-scenario)
```
but “-/testcase/annotation/success (http-client)(webflux-projectA-scenario)” span don‘t show parent.’ don't know if I misunderstood.

arugal: Looks like `tarceId` hasn't been passed on, `BodyInserterRequestInterceptor` is in a new thread to run?

wu-sheng: @dengliming If it is across thread, the parent span is not there, check the ref(type==acorss thread).

dengliming: I think it's been a problem before (without webclient). Maybe the problem was the loss of traceId after springWebFlux processed the request. I'm not sure yet. Do you guys have any ideas？

test/plugin/scenarios/webflux-scenario/webflux-projectA-scenario/src/main/java/org/apache/skywalking/apm/testcase/sc/webflux/projectA/controller/TestController.java
wu-sheng, line 46: If we lost the context before, this is the place broke. But it doesn't seem the case. Webflux plugin still guarantee the spring mvc has correct context, isn't it?

dengliming: en It‘s ’weird. Are you sure the other trace segment is right？(exclude the test I added)

wu-sheng: I am not. But webflux server plugin works in Gateway, nd springmvc in the real world. Others, I am not sure.

dengliming: I mean the trace segment in expectedData.

test/plugin/scenarios/webflux-scenario/config/expectedData.yaml
dengliming: > Looks like `tarceId` hasn't been passed on, `BodyInserterRequestInterceptor` is in a new thread to run?

From my local test. It's not. It seems the webflux server stop span early and hasn't passed traceId to webclient side.

test/plugin/scenarios/webflux-scenario/webflux-projectB-scenario/src/main/java/org/apache/skywalking/apm/testcase/sc/webflux/projectB/controller/TestWebclientController.java
```diff
@@ -0,0 +1,47 @@
...
+    public String ping() {
+        return "ok";
+    }
+
+    @RequestMapping("/testcase/webclient/test1")
+    public String test1() {
+        Mono<String> resp = WebClient.builder().exchangeStrategies(ExchangeStrategies.builder()
+                .build())
+                .build()
+                .get().uri("http://" + hostBAddress + "/testcase/webclient/ping")
```
wu-sheng, line 42: This is the place I have doubts. You have `webclient/test1` entry span, and linked to the previous, so, `visit` method works. And in here, you call `webclient/ping`, 
You can see here
https://github.com/apache/skywalking/blob/4a08b24c6ad2630dc51af0c1fc10bee6e3de74f2/test/plugin/scenarios/webflux-scenario/config/expectedData.yaml#L238-L255
This is a separated segment and span, no parent span and no ref. Meaning, the way you instrumented can not access the context here, which indicates there is another thread executing the codes. 

Is this more clear for you?

dengliming: Thanks. It's very clear. I replace test with this: 
```
@RequestMapping("/testcase/webclient/test1")
    public String test1() {
        CloseableHttpClient httpclient = HttpClients.createDefault();
        try {
            HttpGet httpget = new HttpGet("http://" + hostBAddress + "/testcase/webclient/ping");
            ResponseHandler<String> responseHandler = new ResponseHandler<String>() {
                public String handleResponse(
                        org.apache.http.HttpResponse response) throws ClientProtocolException, IOException {
                    HttpEntity entity = response.getEntity();
                    return entity != null ? EntityUtils.toString(entity) : null;
                }

            };
            return httpclient.execute(httpget, responseHandler);
        } finally {
            httpclient.close();
        }
        return "ok";
    }
```
It's still not link. It's weird. Would you mind testing it locally?

arugal: I'm testing it locally

arugal: https://github.com/apache/skywalking/blob/e356da4b0e264a7617b3356febe6d6e25361f20a/apm-sniffer/apm-sdk-plugin/spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/DispatcherHandlerHandleMethodInterceptor.java#L70
https://github.com/apache/skywalking/blob/e356da4b0e264a7617b3356febe6d6e25361f20a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManager.java#L179-L184

Hi @wu-sheng @dengliming `AbstractTracerContext` in `DefaultExchangeFunctionInterceptor# beforeMethod` was removed

ascrutae: It's still not link because Controller and WebClient calling run in different threads and the WebClient plugin doesn't link the trace context. I suggest that you should save the `ContextSnapshot` when create WebClient Object and `continued` when WebClient send request.

I also found another problem that the trace miss span about WebClient plugin,  In my case, the trace will miss the database span.
```java
 @RequestMapping("/testcase/webclient/test1")
    public String test1() {
        Mono<String> resp = WebClient.builder().exchangeStrategies(ExchangeStrategies.builder()
                .build())
                .build()
                .get().uri("http://" + hostBAddress + "/testcase/webclient/ping")
                .retrieve().bodyToMono(String.class);
        resp.subscribe(new Consumer<String>() {
            @Override
            public void accept(final String s) {
                // read data from database 
            }
        });
        return "ok";
    }
```

dengliming: Hi~ @ascrutae Thanks for reply. I agree with your second point, Now I haven't found a good way to pass the context in Reactor Mono Flow, This is a tricky one.
But the first point, I have also tested WebFlux locally, and the link will also be missing, which has nothing to do with the use of WebClient. This should be an existing bug

ascrutae: That Controller and WebClient calling running in different threads are completely random and this is the characteristic of Reactor Mono Flow.

wu-sheng: @dengliming You have the context in the `#visit`, as the ref exists. I am not sure what do you mean about the bug. One way or another, that is totally different thing. Let's focus on the client issue now.

wu-sheng: @dengliming Any update?

dengliming: Not yet. Sorry for the late reply. I tried ways to enhance WebFilter and WebClient ExchangeFilterFunction, But the client span still not link. That's probably because I'm missing something or we should also add reactor-core plugin passing the context on between Mono subscriber. I would really appreciate it if you have some ideas about how to do that or you can just modify it on this pr
Any way, We can just keep this PR open for now and see if anyone shares some of their solutions or close it for now if that make no sense. Thanks.

wu-sheng: @ascrutae Could you take a deep look at this?

ascrutae: ok, Got it

Issue #5039, opened 2020-07-06
Duplicate service name


amwyyyy:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Bug
Version: 8.0.0

I found a duplicate service name in the drop-down box
![image](https://user-images.githubusercontent.com/8274512/86578011-b7ac6f00-bfad-11ea-9f4b-48330d2fce52.png)

View elasticsearch, there are two data.
```json
{
  "_index": "sw8_service_traffic-20200704",
  "_type": "type",
  "_id": "bWJwLWJvbnVzLWNhbGN1bGF0ZQ==.1",
  "_version": 1,
  "_score": 2,
  "_source": {
    "node_type": 0,
    "name": "mbp-bonus-calculate"
  }
}
```
```json
{
  "_index": "sw8_service_traffic-20200706",
  "_type": "type",
  "_id": "bWJwLWJvbnVzLWNhbGN1bGF0ZQ==.1",
  "_version": 1,
  "_score": 2,
  "_source": {
    "node_type": 0,
    "name": "mbp-bonus-calculate"
  }
}
```

wu-sheng:
From the data, we can see the service exists in 0704 and 0706 index both. So whether there is sometimes, the 0704 is not readable? Once it isn't, the data will be written into latest index again. This is how SkyWalking metadata works today.

amwyyyy:
@wu-sheng 
I found the error log, which should be related to this. but this situation is sometimes unavoidable, I hope to be compatible with this problem.
![image](https://user-images.githubusercontent.com/8274512/86892261-99bc4700-c132-11ea-919b-cb8faa74e10e.png)

wu-sheng:
We can't, because once it is available, we will write a new one. There is no point to deal with it in the query. Because
1. This will restore with time, TTL will delete them.
1. Any service you chose, will get the same result.

wu-sheng:
ES is not available is not a small issu.e

amwyyyy:
There will throw an exception,many features will not be available.
![image](https://user-images.githubusercontent.com/8274512/86912691-07c33700-c150-11ea-8c3c-0696b93faad1.png)

![image](https://user-images.githubusercontent.com/8274512/86913664-8cfb1b80-c151-11ea-9d42-d6005e368396.png)
Change line 48 can make the code robust
```
[...]

PR #5096, opened 2020-07-14
optimization MetricsEsDAO multiGet

amwyyyy:
See this issues [#5039](https://github.com/apache/skywalking/issues/5039)

The problem is because the es query times out
image
Cause duplicate service_traffic
```
{
  "_index": "sw8_service_traffic-20200704",
  "_type": "type",
  "_id": "bWJwLWJvbnVzLWNhbGN1bGF0ZQ==.1",
  "_version": 1,
  "_score": 2,
  "_source": {
    "node_type": 0,
    "name": "mbp-bonus-calculate"
  }
}

{
  "_index": "sw8_service_traffic-20200706",
  "_type": "type",
  "_id": "bWJwLWJvbnVzLWNhbGN1bGF0ZQ==.1",
  "_version": 1,
  "_score": 2,
  "_source": {
    "node_type": 0,
    "name": "mbp-bonus-calculate"
  }
}
```
Then the server starts throw exception
image

View server source code, under normal circumstances method "ids" the return should be less than or equal to the length of String[] ids. However, because service_traffic is repeated, one id can search two results, resulting in totalHits greater than the actual. So throw ArrayIndexOutOfBoundsException.
image
image

***We should guarantee the stability of es, but there will always be unexpected situations, hoping that skywalking is fault-tolerant. Because in the previous version, at most only some trace data was lost.***

### DIFF
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/MetricsEsDAO.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/MetricsEsDAO.java
@@ -44,8 +44,8 @@ public class MetricsEsDAO extends EsDAO implements IMetricsDAO {
     public List<Metrics> multiGet(Model model, List<String> ids) throws IOException {
         SearchResponse response = getClient().ids(model.getName(), ids.toArray(new String[0]));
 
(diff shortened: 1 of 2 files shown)


dmsolr:
In actually, `gitHits().length` is more robustness. In general, `gitHits().Totalhits` works well unless the query timeout. 

`gitHits().totalHits`  is as many documents matching the query. (hits)
`gitHits().length` means the ResultSet size of query. (hits and recalls)

wu-sheng:
Notice, this still could cause duplicate raws in different index. This should be clear.

wu-sheng:
Because timeout still could happen, then duplicate records generated.

PR #5097, opened 2020-07-14
Fix apdex threshold config bug when dynamic configuration is enabled.

Ax1an:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

   Fix apdex threshold config bug when dynamic configuration is enabled.
   This bug will result in a large number of warn logs, even if the service-apdex-threshold.yml file exist.
image

- How to fix?

   Optimize processing logic.
   If there is a core.default.apdexThreshold in the dynamic configuration, use it.
   If there is no core.default.apdexThreshold in the dynamic configuration, the service-apdex-threshold.yml file is read. And it is read only once when the configuration is switched. For example, delete core.default.apdexThreshold in the dynamic configuration.
   Only print the warn log if the service-apdex-threshold.yml file does not exist.

___


wu-sheng:
Behavior should be consistent acrosee all config items. I think you should fix the dynamic configuration center logic, rather than fixing like this.

PR #5099, opened 2020-07-14
Storage elasticsearch health check

hanahmily:
These changes intend to add health checks to elasticsearch storage plugin. 

 * ~~Update elasticsearch client 6.x and 7.x to the latest version, which tries to fix some underlying connection issues~~.
> Upgrading client is hard, let's leave them alone.
 * Add HealthChecker to combinate between provider and storage service client. The HealthChecker opts to "sync" strategy which means it doesn't set up an async scheduler to check backend server periodically. That is because our persistence timer will touch the storage server every 3 seconds(the default value), so we can leverage it as a health checking trigger. 
 * Some users report issues like https://github.com/elastic/elasticsearch/issues/39946, in this PR, I fix this issue by recreating a new RestClient once catching the IllegalStateExecption indicate the collection pool is closed under the hood. 

I only update the elasticsearch 6 in the current commit. After getting enough feedbacks to polish the implementation, I will update elasticsearch 7 accordingly.


oap-server/pom.xml
```diff
@@ -64,7 +64,7 @@
         <h2.version>1.4.196</h2.version>
         <commons-dbcp.version>1.4</commons-dbcp.version>
         <commons-io.version>2.6</commons-io.version>
-        <elasticsearch.version>6.3.2</elasticsearch.version>
+        <elasticsearch.version>6.8.10</elasticsearch.version>
```
wu-sheng, line 67: Could we don't do this? Once you did, I remember, you can't access 6.3-6.7 server any more.

hanahmily: I notice the same thing, so just revert that commit.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchProvider.java
```diff
@@ -197,7 +208,7 @@ public void notifyAfterCompleted() {
 
     @Override
     public String[] requiredModules() {
-        return new String[]{CoreModule.NAME};
+        return new String[]{CoreModule.NAME, TelemetryModule.NAME};
```
wu-sheng, line 211: CoreModule has depended on TelemetryModule, I think.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/healthcheck/HealthChecker.java
```diff
@@ -0,0 +1,49 @@
...
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * HealthChecker could provide health status to the listener.
+ */
+@Slf4j
+@RequiredArgsConstructor
+public class HealthChecker {
```
wu-sheng, line 29: Why need HealthChecker as a kind of util? And keep in static final(DEFAULT_CHECKER)?

According to #5046, Health check is a module/provider. So, the thing about the checker should be a service, which you could get a reference from ServiceManager in any module.

wu-sheng: Also, based on the way you use this, it is likely, you should have this module up and running always.
I think we need a little more discussion about health checker model behavior.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchProvider.java
hanahmily: done

hanahmily:
@wu-sheng I enclose the gauge value into a new metric class, HealthCheckMetrics, so other developers don't have to set the value directly. 

But when I try to move the initialization of the client to the start phase, some other module, for instance, CoreModule, complains they can't find the service provided by Storage modules. It seems we must keep the registration of storage service in the `prepare` phase. So I add a new DeletegatedHealthChecker as a final field of client classes, it can receive the HealthChecker created by Telemetry module, that's available in the start phase.

oap-server/server-storage-plugin/storage-elasticsearch7-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch7/client/ElasticSearch7Client.java
```diff
@@ -256,9 +292,9 @@ public void synchronousBulk(BulkRequest request) {
             int size = request.requests().size();
             BulkResponse responses = client.bulk(request, RequestOptions.DEFAULT);
             log.info("Synchronous bulk took time: {} millis, size: {}", responses.getTook().getMillis(), size);
-        } catch (IOException e) {
-            log.error(e.getMessage(), e);
-        }
+            healthChecker.health();
+        } catch (Throwable t) {
+            healthChecker.unHealth(t);        }
```
wu-sheng, line 297: This seems a format issue.

hanahmily: done

Issue #4965, opened 2020-06-24
InstMethodsInter : class[class com.mysql.cj.jdbc.ClientPreparedStatement] before method[executeQuery] intercept java.lang.NullPointerException  failure


whoisjohnlee:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [x] Bug

___
### Question
- What do you want to know?
How to fix this issue

___
### Bug
- Which version of SkyWalking, OS and JRE?
SkyWalking 8.0.1
OS:Linux
mysql-connector-java version:8.0.11

- Which company or project?
project

- What happen?
When start my project,I found this error message.
[before] and [after] method both happened

ERROR 2020-06-23 14:45:48:469 http-bio-8080-exec-33 InstMethodsInter : class[class com.mysql.cj.jdbc.ClientPreparedStatement] before method[executeQuery] intercept failure 
java.lang.NullPointerException
	at org.apache.skywalking.apm.plugin.jdbc.mysql.PreparedStatementExecuteMethodsInterceptor.beforeMethod(PreparedStatementExecuteMethodsInterceptor.java:43)
	at org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstMethodsInter.intercept(InstMethodsInter.java:76)
	at com.mysql.cj.jdbc.ClientPreparedStatement.executeQuery(ClientPreparedStatement.java)
	at com.mysql.cj.jdbc.DatabaseMetaData$8.forEach(DatabaseMetaData.java:3426)
	at com.mysql.cj.jdbc.DatabaseMetaData$8.forEach(DatabaseMetaData.java:3379)
	at com.mysql.cj.jdbc.IterateBlock.doForAll(IterateBlock.java:56)
	at com.mysql.cj.jdbc.DatabaseMetaData.getProceduresAndOrFunctions(DatabaseMetaData.java:3494)
	at com.mysql.cj.jdbc.DatabaseMetaData.getProcedureOrFunctionColumns(DatabaseMetaData.java:3270)
	at com.mysql.cj.jdbc.DatabaseMetaData.getProcedureColumns(DatabaseMetaData.java:3207)
	at com.mysql.cj.jdbc.CallableStatement.determineParameterTypes(CallableStatement.java:757)
	at com.mysql.cj.jdbc.CallableStatement.<init>(CallableStatement.java:558)
	at com.mysql.cj.jdbc.CallableStatement.getInstance(CallableStatement.java:467)
[...]

PR #5109, opened 2020-07-16
Fix issue 4965

Patrick0308:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
#4965 
___
### Bug fix
- Bug description.
When invoking prepareCall, it will determine procedures's parameter types by execute sql. `StatementEnhanceInfos` is set after prepareCall, so `StatementEnhanceInfos` is null in `PreparedStatementExecuteMethodsInterceptor.beforeMethod`, the NPE was happen.
- How to fix?
I suggest to avoid NPE. We don't need to record sql which determine procedures's parameter.
___

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementExecuteMethodsInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementExecuteMethodsInterceptor.java
@@ -43,7 +43,8 @@ public class PreparedStatementExecuteMethodsInterceptor implements InstanceMetho
         /**
          * For avoid NPE. In this particular case, Execute sql inside the {@link com.mysql.jdbc.ConnectionImpl} constructor,
          * before the interceptor sets the connectionInfo.
-         * Fixed https://github.com/apache/skywalking/issues/4965. When invoking prepareCall, cacheObject is null. Because it will determine procedures's parameter types by querying the table of proc in mysql.
+         * When invoking prepareCall, cacheObject is null. Because it will determine procedures's parameter types by executing sql in mysql 
+         * before the interceptor sets the statementEnhanceInfos.
          * @see JDBCDriverInterceptor#afterMethod(EnhancedInstance, Method, Object[], Class[], Object)
          */
         if (cacheObject != null && cacheObject.getConnectionInfo() != null) {
--- a/apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementExecuteMethodsInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementExecuteMethodsInterceptor.java
@@ -72,7 +72,7 @@ public class PreparedStatementExecuteMethodsInterceptor implements InstanceMetho
     public final Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments,
                                     Class<?>[] argumentsTypes, Object ret) {
         StatementEnhanceInfos cacheObject = (StatementEnhanceInfos) objInst.getSkyWalkingDynamicField();
-        if (cacheObject.getConnectionInfo() != null) {
+        if (cacheObject != null && cacheObject.getConnectionInfo() != null) {
             ContextManager.stopSpan();
         }
         return ret;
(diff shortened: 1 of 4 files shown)


apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementExecuteMethodsInterceptor.java
```diff
@@ -40,15 +40,14 @@
     public final void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments,
                                    Class<?>[] argumentsTypes, MethodInterceptResult result) {
         StatementEnhanceInfos cacheObject = (StatementEnhanceInfos) objInst.getSkyWalkingDynamicField();
-        ConnectionInfo connectInfo = cacheObject.getConnectionInfo();
         /**
          * For avoid NPE. In this particular case, Execute sql inside the {@link com.mysql.jdbc.ConnectionImpl} constructor,
          * before the interceptor sets the connectionInfo.
-         *
+         * Fixed https://github.com/apache/skywalking/issues/4965. When invoking prepareCall, cacheObject is null. Because it will determine procedures's parameter types by querying the table of proc in mysql.
```
wu-sheng, line 46: Don't add the issue link, unless it is a very special thing.

wu-sheng: About `#prepareCall`, why `cacheObject` should be null? I prefer you should fix this NPE by having the correct ConnectionInfo rather than skipping it.

wu-sheng: You should forward the ConnectionInfo from Connection when `#prepareCall` called.

Patrick0308: When invoking prepareCall, it will determine procedures's parameter types by execute sql.` StatementEnhanceInfos` is set after prepareCall, so `cacheObject is` null in `PreparedStatementExecuteMethodsInterceptor.beforeMethod`. 
If we don't need to record the sql which determine procedures's parameter types when invoking prepareCall, we can skip it.

apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementExecuteMethodsInterceptor.java
```diff
@@ -40,15 +40,14 @@
...
          * @see JDBCDriverInterceptor#afterMethod(EnhancedInstance, Method, Object[], Class[], Object)
          */
-        if (connectInfo != null) {
-
+        if (cacheObject != null && cacheObject.getConnectionInfo() != null) {
```
wu-sheng, line 49: According to the codes here, if `cacheObject.getConnectionInfo() == null)`, no span will be generated. My previous question is about this, why?

Also, on the existing codes, the check condition is different, https://github.com/apache/skywalking/blob/e7e72d937b769f40e50ec644a210bc497f2204f9/apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementExecuteMethodsInterceptor.java#L75.

I want to make sure what is the issue of that. Because all changes in the expectedData.yml seems not related to some parts of the change.

Patrick0308: Intercepter set `connectionInfo` after `Driver.connect`. However, in the constructor of `ConnectionImpl`, executing built-in SQL will don't have `connectionInfo`. 
https://github.com/apache/skywalking/blob/e7e72d937b769f40e50ec644a210bc497f2204f9/apm-sniffer/apm-sdk-plugin/mysql-common/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/PreparedStatementExecuteMethodsInterceptor.java#L44

Both the condition `cacheObject.getConnectionInfo() != null` and the condition `cacheObject != null` are for not log error message and skip the jdbc's built-in SQL. 
And the NPE not affect recording user's SQL.

Issue #5063, opened 2020-07-08
build connection between agent client and oap server with service discover


EvanLjp:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
In some cases,BACKEND_SERVICE is config as dns:port;
some dns production is not support Load balancing,for example core dns in cncf.

so i want to add a dns lookup to support the case

wu-sheng:
Just confirm, you want to do `DNS lookup` manually by codes, then select the backend from candidates?

EvanLjp:
yes，maybe can support more ways，for example 
1. dns
2. Kubernetes service
Maybe kubernetes service load balancer mode is a better way,but the way depends on cloud (aws,Ali),it would be not work in a privite kubernetes cluster 
So.i want to add the feature, like kubernetes dubbo 
1.Use kubernetes service self support 
1.if have time,also supports dns nslookup

wu-sheng:
Do you have more material about DNS level load balancer can't work automatically? From your asking, it seems, we need codes to do a DNS lookup to get the candidate list. Am I right?

wu-sheng:
If you want to add new dependency on the agent core, that wouldn't be recommended, especially for this case. The way would be adding a plugin for feature extension. If there is no extension point currently, we could create one.
Anymore, I need more details about what you are trying to do. Because I don't know what dubbo-k8s did, actually, that is not have to be our reference. Agent is very different from a dev framework.

EvanLjp:
yes，i should change a new title
build the connection between  agent client and oap server with service discover
for example :
Kubernetes
dns
etcd
zk
.....

create a new extension point like oap

wu-sheng:
OK, that will not be an easy job. NOTICE, receiver role OAP role is never registered through the coordinator, but your service discovery is for looking up receiver role nodes.

wu-sheng:
And service discovery of OAP node is not recommended in many companies, because the OAP nodes are deployed in the system/operation platform zone, rather than the business zone. Service discovery wouldn't work in this case.

EvanLjp:
OK，thx

EvanLjp:
i found the purpose of boot service  int erface for single purpose.
if we want to support  the  polymorphism in service discovery.
2 ways: 
1. modify the boot service to support polymorphism
2. add a new interface.

Do you have any Suggestions

PR #4847, opened 2020-06-01
Provide kafka as collector/reporter

dmsolr:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### New feature or improvement
- Describe the details and related test reports.

1. Introduce Kafka as Agent Reporter and OAP Receiver, which are optional plugins.
2. Agent still needs GRPC when it is an interactive case. (Agent get the Profile Task List. But `ProfileThreadSnapshot` is reported by MQ)
3. Currently, it is missing cluster(it is coming soon) and e2e test.

PR #5120, opened 2020-07-17
Agent service discorvery

EvanLjp:

- [ ] New feature provided

- Related issues
https://github.com/apache/skywalking/issues/5063

___


EvanLjp:
I will add unit tests later, thanks for review first

EvanLjp:
fixing unit test,please wait for a moment

wu-sheng:
Don't change the backend serive url config, it will break many deployed system. Please don't do that.

EvanLjp:
> I think I mentioned before, neither service discovery tech should not be bounded in the agemt core level. If you want to contribute this feature, you need to put it as an optional plugin only. Take a look on Kafka transport plugin and zipkin branch.
@wu-sheng 
it is designed as a plugin mode:
image

EvanLjp:
so i cannot extension with service mode , i would be add a new extension interface ?

wu-sheng:
The concern is, the default config changed. And new backend service path added. This would be very confused. How to write plugin doc would be an issue.

EvanLjp:
> The concern is, the default config changed. And new backend service path added. This would be very confused. How to write plugin doc would be an issue.

how about this mode:
origin backend url is as default.the extension discovery is in Service Discovery 
image

wu-sheng:
By thinking the PR and Kafka transport PR, I prefer you work on a new PR.
Introduce a new SPI service, called ConfigInitService. All plugins and core could provide their services. Then you could separate the current Config in the core and different plugins.
Notice, all config keys should be kept as much as possible.

What do you think? @dmsolr @kezhenxu94 @arugal @EvanLjp

Let's make that decision first, then back to this feature.

EvanLjp:
@wu-sheng  the PR is for agent discovery

wu-sheng:
@EvanLjp Yes, this one and #4847 show, the config of different plugins have been keeping added into the core Config. I want to remove this first. That is the reason I talked about ConfigInitService(not a BootService).

EvanLjp:
i open a new issue  first

wu-sheng:
Once we have this feature, this plugin will be more elegant, because there will be no code changed in the agent core, purely plugin. Same as the Kafka transport plugin.

EvanLjp:
I wonder if the plug-in configuration needs to be refactored,for example mysql

wu-sheng:
> I wonder if the plug-in configuration needs to be refactored,for example mysql

At the code level, yes. But you still could keep the config name unchanged. You just create more Config classes in different plugins.

EvanLjp:
no porblem , i would like be to open a new issue first

wu-sheng:
I think you could improve this PR now. I have set up the new Config initialization core.

EvanLjp:
ok

wu-sheng:
Do we still plan to do this? I think we discussed, cliemt load balance is not a very good way. It is hard to be balaned.

EvanLjp:
i read the oap code and agent code. oap transform data from one oap server to another oap server is get client form a grpc client pool. so oap server can load balance.
why not we still maintain the mode in agent.
we create a grpc clients manager,when transform data ,we random get a client

EvanLjp:
> Do we still plan to do this? I think we discussed, cliemt load balance is not a very good way. It is hard to be balaned.

i don't think keep alive one by one is a good way.Don't you think this is skywalking's current problem?Why don't we try to optimize?

wu-sheng:
Rechecking, whether this feature should be or is planned to continue? It has been no update for a month.

EvanLjp:
may be close is a good choose, thx

Issue #5023, opened 2020-07-03
trace query performance


EvanLjp:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
the trace quey is a range query with  timeBucket.
is the trace segment data keeps 7 days. 
And The amount of data we get every day is huge.
the Query performance will be very poor

the reason:
the skywalking trace query only consider  common coding styles ranther than the query performance in es.

how to improve:
a query form front end ,and the time bucket is between 20200627000238 and  20200627000438

so if the index spilt by day step

the data must contains in 20200627 index. so we only need to query data in this index ranther than all index

wu-sheng:
Yes, logically, it is. But the change wouldn't be that easy. #4863 stopped, someone tried to do so too. Basically, you need to consider when you do trace_id query, there is no timestamp.

EvanLjp:
okay, i would like to read the issue. and try to solve it later

EvanLjp:
_Basically, you need to consider when you do trace_id query, there is no timestamp._

why not to add a time key for backend.

i read the[PR](https://github.com/apache/skywalking/pull/4863) ,i don't parse time from traceId is a good idea.for example the traceId in nginx lua agent.
so my advice: 
1. normal query.the index time is format between  [queryStartTime -1h,queryEndTime +1h]
2. add a time key when traceID query
3. if the advice2 is not implement, Short term plan: if have 7 segment index, from 7.1 to 7.7 ,we can search every index from 7.7 to 7.1 .if hits some data,we also to search the prefix one (when cross day) .

wu-sheng:
> i read thePR ,i don't think parse time from traceId is a good idea.for example the traceId in nginx lua agent.

No, that is not a good idea.

> Basically, you need to consider when you do trace_id query, there is no timestamp.

Because finding the timestamp may be an issue. When we inject the trace id into the logs or even in codes/database, there may be no timestamp there. And the trace query page supports `?traceId=xxx` to get the result directly.

> if the advice2 is not implement, Short term plan: if have 7 segment index, from 7.1 to 7.7 ,we can search every index from 7.7 to 7.1 .if hits some data,we also to search the prefix one (when cross day) .

This is as same as using Alias query, which is better because of parallel query.

-------
Basically, in `trace_id` query case, I think we still need the current query mode. If you want to optimize the query when timestamps exist, I think it is possible.

EvanLjp:
[...]

PR #5132, opened 2020-07-20
Improve Es trace basic  query performance

EvanLjp:

- [ ] Improve performance

- Related issues
https://github.com/apache/skywalking/issues/5023
___

improve trace query performance


EvanLjp:
@wu-sheng thx for review

EvanLjp:
@wu-sheng  I have a new idea to solve es query speed on traceId query
1: shard_num = (hash(_routing) + hash(_id) % routing_partition_size) % num_primary_shards 
when writing: we can compute the shards number by some rules
2. when reading ,we can compute the data in which shards .and only get data from this shards

how do u think about this idea

oap-server/server-storage-plugin/storage-elasticsearch7-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch7/query/TraceQueryEs7DAO.java
```diff
@@ -115,7 +116,8 @@ public TraceBrief queryBasicTraces(long startSecondTB,
         sourceBuilder.size(limit);
         sourceBuilder.from(from);
 
-        SearchResponse response = getClient().search(SegmentRecord.INDEX_NAME, sourceBuilder);
+        String[] indices = TimeSeriesUtils.traceQueryIndices(SegmentRecord.INDEX_NAME, startSecondTB, endSecondTB);
```
wu-sheng, line 119: Still, the issue is, currently, we don't ask for time range because of trace id query.

EvanLjp:
@wu-sheng  thx for review ,the codes is also for super size dataset

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/TimeSeriesUtils.java
```diff
@@ -60,6 +61,31 @@ public static String latestWriteIndexName(Model model) {
         }
     }
 
+    /**
+     * @return Concrete index name for super dataset index
+     */
+    public static String[] querySuperDatasetIndices(String indexName, long startTimeBucket, long endTimeBucket) {
+        if (startTimeBucket == 0 || endTimeBucket == 0) {
+            return new String[] {indexName};
```
wu-sheng, line 69: So, does this means, once the time buckets don't exist, still use the alias name, right?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/SuperDatasetRangeElasticSearchClient.java
```diff
@@ -0,0 +1,81 @@
...
+import org.elasticsearch.index.query.QueryBuilder;
+import org.elasticsearch.index.query.RangeQueryBuilder;
+import org.elasticsearch.search.builder.SearchSourceBuilder;
+
+import static org.apache.skywalking.oap.server.storage.plugin.elasticsearch.base.TimeSeriesUtils.querySuperDatasetIndices;
+
+public class SuperDatasetRangeElasticSearchClient {
+
+    public static SearchResponse search(ElasticSearchClient client,
+                                        String rangeTimeName,
```
wu-sheng, line 36: Is `rangeTimeName` actually `timeBucketColName`?

EvanLjp: yes

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/TimeSeriesUtils.java
EvanLjp: yes,the is plan B

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/SuperDatasetRangeElasticSearchClient.java
```diff
@@ -0,0 +1,81 @@
...
+import org.elasticsearch.index.query.BoolQueryBuilder;
+import org.elasticsearch.index.query.QueryBuilder;
+import org.elasticsearch.index.query.RangeQueryBuilder;
+import org.elasticsearch.search.builder.SearchSourceBuilder;
+
+import static org.apache.skywalking.oap.server.storage.plugin.elasticsearch.base.TimeSeriesUtils.querySuperDatasetIndices;
+
+public class SuperDatasetRangeElasticSearchClient {
+
+    public static SearchResponse search(ElasticSearchClient client,
```
wu-sheng, line 35: Based on the implementation, is this method should be more specific about it is for `searchOnSuperDataSet`? In the L40, you call `#querySuperDatasetIndices` directly.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/SuperDatasetRangeElasticSearchClient.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.skywalking.oap.server.library.client.elasticsearch.ElasticSearchClient;
+import org.elasticsearch.action.search.SearchResponse;
+import org.elasticsearch.index.query.BoolQueryBuilder;
+import org.elasticsearch.index.query.QueryBuilder;
+import org.elasticsearch.index.query.RangeQueryBuilder;
+import org.elasticsearch.search.builder.SearchSourceBuilder;
+
+import static org.apache.skywalking.oap.server.storage.plugin.elasticsearch.base.TimeSeriesUtils.querySuperDatasetIndices;
+
+public class SuperDatasetRangeElasticSearchClient {
```
wu-sheng, line 33: Don't add a client with static methods, that break the code style.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/SuperDatasetRangeElasticSearchClient.java
```diff
@@ -0,0 +1,81 @@
...
+     * @return the time range query builder
+     */
+    private static Optional<RangeQueryBuilder> findTimeRangeQueryBuilder(final SearchSourceBuilder searchSourceBuilder,
+                                                                         final String rangeTimeName) {
+        QueryBuilder query = searchSourceBuilder.query();
+        if (query instanceof RangeQueryBuilder) {
+            RangeQueryBuilder rangeQueryBuilder = (RangeQueryBuilder) query;
+            return rangeQueryBuilder.fieldName().equals(rangeTimeName) ? Optional.of(rangeQueryBuilder) : Optional.empty();
+        } else if (query instanceof BoolQueryBuilder) {
+            return findTimeRangeQueryBuilder(rangeTimeName, ((BoolQueryBuilder) query).must(), ((BoolQueryBuilder) query).filter());
```
wu-sheng, line 59: There is recursion, why?

EvanLjp:
> Basically, I it seems timeRangQuery should be explicit in the API level with a new timeRange API to generate the condition. Currently, the codes seem are not this way.
> 
> I need a little more update from you to follow the code changes.

no problem.I can modify it if possible

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/SuperDatasetRangeElasticSearchClient.java
EvanLjp: image
it is just a overload method

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/SuperDatasetRangeElasticSearchClient.java
EvanLjp: 1. in my first option, the range query client is working on metrics or record data
2. but when i submit superdataset day step pr, i thought the query cient would not works well in mertics or normal record data.
3. so i renamed the client as superdataset range query client
4. the client is designed to extract range query condition from sourcebuilder rather than input again.

wu-sheng:
Ideally, I think we just need to add a range query method to `ElasticSearchClient`, and provide a generator for index names including all ranges. The codes should be easy to maintain.

EvanLjp:
> Ideally, I think we just need to add a range query method to `ElasticSearchClient`, and provide a generator for index names including all ranges. The codes should be easy to maintain.

there is a problem. day steps info is not contains in client library.so cannot compute index name in client library module

wu-sheng:
> there is a problem. day steps info is not contains in client library.so cannot compute index name in client library module
that is why i write the codes in storage module

This calculation should be a part of TimeSeriesUtils, that is right.

EvanLjp:
there is 2 ways :
1. build a client wrapper for es search. 
2. TimeSeriesUtils build index names. pass the indices to es client library module.

in my option, first way is easy to maintain. in second way ,maybe forgot it at  some time.
Of course the first can be more elegant

wu-sheng:
> build a client wrapper for es search.

Wrapper is fine, my point is, the current codes are mixing the normal(99%) search and the range query(for trace only). I want to separate them, which is easier to read and make the reviewer easier for more queries in the future.

wu-sheng:
> TimeSeriesUtils build index names. pass the indices to es client library module.

Even for this, I don't think there is a forgotten thing, because if you just access one index, you can't get enough data. In SkyWalking, you either use the alias or the real index names.

wu-sheng:
You could make ES client accepting the current index name or a index name builder.

EvanLjp:
> > TimeSeriesUtils build index names. pass the indices to es client library module.
> 
> Even for this, I don't think there is a forgotten thing, because if you just access one index, you can't get enough data. In SkyWalking, you either use the alias or the real index names.

now. we talk about range query, this input range would be calculated to real index names ,so would not lost data

EvanLjp:
> You could make ES client accepting the current index name or a index name builder.

let me thinking this way

wu-sheng:
> now. we talk about range query, this input range would be calculated to real index names ,so would not lost data

I was talking about your point of misusing the API.

EvanLjp:
@wu-sheng how about this version
add a function interface called IndexNameMaker to supply a common way to pass how to build index name

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/TimeSeriesUtils.java
```diff
@@ -60,6 +61,31 @@ public static String latestWriteIndexName(Model model) {
         }
     }
 
+    /**
+     * @return Concrete index name for super dataset index
+     */
+    public static String[] querySuperDatasetIndices(String indexName, long startTimeBucket, long endTimeBucket) {
```
wu-sheng, line 67: I think this should be renamed to `superDatasetIndexName`, right?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/TimeSeriesUtils.java
```diff
@@ -60,6 +61,31 @@ public static String latestWriteIndexName(Model model) {
...
+        DateTime startDateTime = TIME_BUCKET_FORMATTER.parseDateTime(startDay + "");
+        DateTime endDateTime = TIME_BUCKET_FORMATTER.parseDateTime(endDay + "");
+
+        int steps = Math.max((Days.daysBetween(startDateTime, endDateTime).getDays()) / SUPER_DATASET_DAY_STEP, 0);
+        return IntStream.rangeClosed(0, steps)
+                        .mapToObj(step -> indexName + Const.LINE + startDateTime.plusDays(Math.toIntExact(SUPER_DATASET_DAY_STEP * step)).toString(TIME_BUCKET_FORMATTER))
+                        .toArray(String[]::new);
+    }
+
+    private static long convert2DayTimeBucket(long timeBucket) {
```
wu-sheng, line 82: I think basically, we don't need this method. Read Joda lib, such as  `LocalDateTime.now().truncatedTo(ChronoUnit.MINUTES)`

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/TimeSeriesUtils.java
```diff
@@ -60,6 +61,31 @@ public static String latestWriteIndexName(Model model) {
...
+    public static String[] querySuperDatasetIndices(String indexName, long startTimeBucket, long endTimeBucket) {
+        if (startTimeBucket == 0 || endTimeBucket == 0) {
+            return new String[] {indexName};
+        }
+        long startDay = compressTimeBucket(convert2DayTimeBucket(startTimeBucket), SUPER_DATASET_DAY_STEP);
+        long endDay = compressTimeBucket(convert2DayTimeBucket(endTimeBucket), SUPER_DATASET_DAY_STEP);
+        DateTime startDateTime = TIME_BUCKET_FORMATTER.parseDateTime(startDay + "");
+        DateTime endDateTime = TIME_BUCKET_FORMATTER.parseDateTime(endDay + "");
+
+        int steps = Math.max((Days.daysBetween(startDateTime, endDateTime).getDays()) / SUPER_DATASET_DAY_STEP, 0);
```
wu-sheng, line 76: Max with `0`? Do you avoid negative value? What happens when `the steps == 0`?

wu-sheng: I think this method should simply follow this logic,

1. List all days between `startTime` and `endTime`.
1. Use `compressTimeBucket` to get all indexes(could have duplicated records)
1. Distinct the list.

wu-sheng: The above processes may need a little refactor about the `compressTimeBucket` method, I should be able to accept `DateTime` as one parameter too.

wu-sheng:
Please check the e2e, look like it fails somehow.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -331,9 +333,25 @@ public boolean deleteTemplate(String indexName) throws IOException {
...
+
+    public SearchResponse search(String[] indexNames, SearchSourceBuilder searchSourceBuilder) throws IOException {
+        indexNames = Arrays.stream(indexNames).map(this::formatIndexName).toArray(String[]::new);
+        return doSearch(searchSourceBuilder, indexNames);
+    }
+
+    private SearchResponse doSearch(SearchSourceBuilder searchSourceBuilder,
+                                    String... indexNames) throws IOException {
+        SearchRequest searchRequest = new SearchRequest(indexNames);
+        searchRequest.indicesOptions(IndicesOptions.fromOptions(true, true, true, false));
```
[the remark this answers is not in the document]
EvanLjp: for example: 
if user select a time range : 2020.08.01==>2020.08.31
our index only has xxx-segment-2020.08.10

if pass so many unavailable index to before search,we will got a exception notice  xxx index is not exist

so the param is to ignore unexist index

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -331,9 +333,25 @@ public boolean deleteTemplate(String indexName) throws IOException {
...
+        return search(indexNames, searchSourceBuilder);
+    }
+
     public SearchResponse search(String indexName, SearchSourceBuilder searchSourceBuilder) throws IOException {
         indexName = formatIndexName(indexName);
-        SearchRequest searchRequest = new SearchRequest(indexName);
+        return doSearch(searchSourceBuilder, indexName);
+    }
+
+    public SearchResponse search(String[] indexNames, SearchSourceBuilder searchSourceBuilder) throws IOException {
```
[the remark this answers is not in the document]
EvanLjp: yes

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/TimeSeriesUtils.java
EvanLjp: ok

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/TimeSeriesUtils.java
EvanLjp: thx for your notice

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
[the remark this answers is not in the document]
wu-sheng: Are other values default?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/TimeSeriesUtils.java
EvanLjp: the method is to convert input time bucket to yyyyMMdd.
for example the input long maybe following:
1. yyyyMMdd
2. yyyyMMddHH
2.yyyyMMddHHmm
3.yyyyMMddHHmmss

wu-sheng: Really? From my check, all these are running for TimeRangeIndexNameMaker and Trace query, the input parameter is always `startSecondTB`? Why it could be different formats?

EvanLjp: sorry.i am wrong.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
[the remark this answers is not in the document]
EvanLjp: image
i read the source code, in my opion, this config nedd to be manually set

EvanLjp:
@wu-sheng thx for review again,all done

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -331,9 +333,25 @@ public boolean deleteTemplate(String indexName) throws IOException {
         return response.getStatusLine().getStatusCode() == HttpStatus.SC_OK;
     }
 
+    public SearchResponse search(IndexNameMaker indexNameMaker, SearchSourceBuilder searchSourceBuilder) throws IOException {
+        String[] indexNames = Arrays.stream(indexNameMaker.make()).map(this::formatIndexName).toArray(String[]::new);
+        return search(indexNames, searchSourceBuilder);
```
wu-sheng, line 338: Why need and call `search` again? And in both methods, you are only doing the `formatIndexName`. I think the namespace would be added twice.

wu-sheng: What happens if the `indexNames#size` == 0?

EvanLjp: 1. get real index name by String[] indexNames = Arrays.stream(indexNameMaker.make()).map(this::formatIndexName).toArray(String[]::new);
2. do search search(indexNames, searchSourceBuilder) input is real indexNames. 

in TimeSeriesUtils, the default name is alias name ,so the input would not be empty

EvanLjp:
> I need to recheck the e2e, they are failing. All tests should pass before the review.
may be because of 
image

i only add unit test for TimeSeriesUtils ,but i thought it is enough

wu-sheng:
> i only add unit test for TimeSeriesUtils ,but i thought it is enough

Not these ones, I am talking about several `e2e/*`, they timed out. They would not be real failed.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
wu-sheng: You are calling another `search`, not `doSearch`. That is what I am asking. Please recheck the codes.

EvanLjp: sorry

wu-sheng: Do we still need another `search(array)`?

EvanLjp:
i would like to sync the codes to our private repository and test again first

wu-sheng:
> i would like to sync the codes to our private repository and test again first

Sure, will wait for your feedback.

wu-sheng:
@EvanLjp Have you located the issue? About why the trace query is not working?

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -325,15 +327,24 @@ public boolean createTemplate(String indexName, Map<String, Object> settings,
...
 
+    public SearchResponse search(IndexNameMaker indexNameMaker, SearchSourceBuilder searchSourceBuilder) throws IOException {
+        String[] indexNames = Arrays.stream(indexNameMaker.make()).map(this::formatIndexName).toArray(String[]::new);
+        return doSearch(searchSourceBuilder, indexNames);
+    }
+
     public SearchResponse search(String indexName, SearchSourceBuilder searchSourceBuilder) throws IOException {
         indexName = formatIndexName(indexName);
-        SearchRequest searchRequest = new SearchRequest(indexName);
+        return doSearch(searchSourceBuilder, indexName);
```
wu-sheng, line 341: You refactor this method, but forgot the `indexName = formatIndexName(indexName);`, which is used to exist.

I think that is the reason, your PR is not working.

oap-server/server-storage-plugin/storage-elasticsearch7-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch7/query/TraceQueryEs7DAO.java
```diff
@@ -124,8 +125,7 @@ public TraceBrief queryBasicTraces(long startSecondTB,
         }
         sourceBuilder.size(limit);
         sourceBuilder.from(from);
-
-        SearchResponse response = getClient().search(SegmentRecord.INDEX_NAME, sourceBuilder);
+        SearchResponse response = getClient().search(new TimeRangeIndexNameMaker(SegmentRecord.INDEX_NAME, startSecondTB, endSecondTB), sourceBuilder);
```
wu-sheng, line 128: I think ES7 and ES6 should use different TimeRangeIndexNameMaker, due to different package name or something like that. Please recheck.

wu-sheng: image

You could find this in the e2e logs. And es7 e2e is going to fail because of the error.

wu-sheng: Notice in the current codes, `ElasticSearch7Client` and `ElasticSearchClient` have their own `#search` methods, your new `#doSearch` method has not been included in the ES 7 client.

EvanLjp: today i test the codes in our test env  and find the bug,already fix it ,and pass the tests

EvanLjp: waiting to  pass all ci proesses

wu-sheng: This is a very tricky bug due to Elastic changed the package name, but with the same class name.

EvanLjp: yes,the error is no such method exception when testiing,i realize may be jars conflict.but took so many time to  find it

wu-sheng: Finally, we get this method working, and it should be helpful to many users.

PR #5136, opened 2020-07-20
Separate config initialization in the plugin out of core level Config.

### LINKED ISSUE #5130: Plugin Config Loader

___

A spi interface to load plugin config

### implementation
1. define a interface called Plugin Config
2. define a spi interface to load Plugin Config(all config schema is defined by submitter)
3. define a container to contain the plugin config,for example Map<class extends config loader spi,Plugin Config>
4. the container is shared with all agent plugins

wu-sheng:
In the current agent core, the `Config` in the core is heavily binding with many plugins, because it is the only way to do configurations injection through agent.config, properties,  and system env. With the increment of plugins, especially the transport plugin(Kafka) and backend service discovery plugin(k8s), the Config class is being out of control.
In this PR, I support all plugins to provide its one or more Config Initialization Services and could initial the plugin configurations used by themselves only. This should be the better way to provide a plugin as an independent module as we mentioned in our design principle.
By leveraging our ClassLoader, once the `org.apache.skywalking.apm.agent.core.boot.PluginConfig` implementations loaded, it would be initialized inside the agent core.
At the same time, all plugin configurations should be kept unchanged. Users wouldn't feel this change happened.

Resolves #5130 FYI @EvanLjp


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/boot/ConfigInitializationService.java
```diff
@@ -0,0 +1,32 @@
...
+ * ConfigInitializationService provides the config class which should host all parameters originally from agent setup.
+ * {@link org.apache.skywalking.apm.agent.core.conf.Config} provides the core level config, all plugins could implement
+ * this interface to have the same capability about initializing config from agent.config, system properties and system
+ * environment variables.
+ */
+public interface ConfigInitializationService {
+    /**
+     * @return Config to host parameters, all static fields set based on the config variable name.
+     */
+    Class config();
```
kezhenxu94, line 31: I didn't see any particular usage of this method, all implementations of this interface return the class itself, just to get the implemented class, I think this method is no needed actually

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/boot/PluginConfigInitializer.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * PluginConfigInitializer loads Config(s) in all existing plugins, and initialize them through current agent settings.
+ */
+public class PluginConfigInitializer {
+    public void initConfigurationsOfAllPlugins() {
+        final ServiceLoader<ConfigInitializationService> configServiceLoader = ServiceLoader.load(
+            ConfigInitializationService.class, AgentClassLoader.getDefault());
+        configServiceLoader.forEach(configInitializationService -> {
+            SnifferConfigInitializer.initializeConfig(configInitializationService.config());
```
kezhenxu94, line 33: Here is what I mean

```suggestion
            SnifferConfigInitializer.initializeConfig(configInitializationService.getClass());
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/boot/ConfigInitializationService.java
wu-sheng: I just want to follow the SPI mode, because the developer actually could not return this class type.

kezhenxu94: > I just want to follow the SPI mode, because the developer actually could not return this class type.

This is nothing todo with SPI IMO, and the developer doesn't need to return the type, we just get the actual class in the service loader, see my another comment inline

wu-sheng: I have read that. I mean, this Config class is loaded through SPI.

wu-sheng:
I think this SPI solution is facing ClassLoader issue, so the tests can't pass. I have to implement it in another more difficult way.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/boot/ConfigInitializationService.java
wu-sheng: I think the new initialization mode fits your preference more. Anyway, the old way is wrong in the tech perspective.

apm-sniffer/optional-plugins/customize-enhance-plugin/src/main/java/org/apache/skywalking/apm/plugin/customize/conf/PluginConfigService.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ */
+
+package org.apache.skywalking.apm.plugin.customize.conf;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.skywalking.apm.agent.core.boot.PluginConfig;
+
+public class PluginConfigService implements PluginConfig {
```
kezhenxu94, line 25: Is it intentionally named with a `Service` suffix? Seems very different from others

apm-sniffer/apm-toolkit-activation/apm-toolkit-trace-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/ToolkitPluginConfigService.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.toolkit.activation;
+
+import org.apache.skywalking.apm.agent.core.boot.PluginConfig;
+
+public class ToolkitPluginConfigService implements PluginConfig {
```
kezhenxu94, line 23: And this one

wu-sheng:
`@PluginConfig(root=Class)` added. For keeping the configuration named unchanged, I keep the `Inner Static` class style. In the latest implementation, this actually is not required. This annotation could be used in any place, but keeping it in the simple Config Class should be better for the clean code. FYI @apache/skywalking-committers

PR #5137, opened 2020-07-20
Add Nacos back into SkyWalking as an option(nacos-1.3.1)

songzhendong:
Add Nacos back into SkyWalking as an option(nacos-1.3.1)，Local compilation passed


wu-sheng:
@KomachiSion I am going to merge this, once the CI passed.

PR #5139, opened 2020-07-20
Add swctl to docker image for health check

hanahmily:
As indicated by the title. 

`swctl` is built as the first stage of the docker build.


wu-sheng:
@kezhenxu94 We need to be careful about this kind of change. All codes in the docker should be a part of the official release vote already.

kezhenxu94:
@hanahmily the CLI 0.3.0 is released

docker/docker-compose.yml
```diff
@@ -42,6 +42,14 @@ services:
     environment:
       SW_STORAGE: elasticsearch
       SW_STORAGE_ES_CLUSTER_NODES: elasticsearch:9200
+      SW_HEALTH_CHECKER: default
+      SW_TELEMETRY: prometheus
+    healthcheck:
+      test: ["CMD", "./bin/swctl", "ch"]
```
wu-sheng, line 48: I didn't found this command doc on the CLI repo. Seems we missed it. Please update the doc. @hanahmily

Issue #4449, opened 2020-03-06
Support Jetty config in Core and SharingServer modules


wu-sheng:
As #4399 merged, Jetty container is providing trace/metrics collection capabilities. So we need to set a better Jetty settings including the defaults and open them to the end users in our YAML.

@JaredTan95 Please choose which are the default values, and which should be open as configurations in the YAML.

wu-sheng:
@JaredTan95 Any update about this?

PR #5147, opened 2020-07-22
support jetty server advanced configuration.

JaredTan95:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
fix https://github.com/apache/skywalking/issues/4449
___
### Bug fix
- Bug description.

- How to fix?

___

### DIFF
--- a/oap-server/server-bootstrap/src/main/resources/application.yml
+++ b/oap-server/server-bootstrap/src/main/resources/application.yml
@@ -57,6 +57,11 @@ core:
     restHost: ${SW_CORE_REST_HOST:0.0.0.0}
     restPort: ${SW_CORE_REST_PORT:12800}
     restContextPath: ${SW_CORE_REST_CONTEXT_PATH:/}
+    restMinThreads: ${SW_CORE_REST_JETTY_MIN_THREADS:1}
+    restMaxThreads: ${SW_CORE_REST_JETTY_MAX_THREADS:200}
+    restIdleTimeOut: ${SW_CORE_REST_JETTY_IDLE_TIMEOUT:30000}
+    restAcceptorPriorityDelta: ${SW_CORE_REST_JETTY_DELTA:0}
+    restAcceptQueueSize: ${SW_CORE_REST_JETTY_QUEUE_SIZE:0}
     gRPCHost: ${SW_CORE_GRPC_HOST:0.0.0.0}
     gRPCPort: ${SW_CORE_GRPC_PORT:11800}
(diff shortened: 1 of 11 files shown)


oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -156,7 +156,15 @@ storage:
 receiver-sharing-server:
   selector: ${SW_RECEIVER_SHARING_SERVER:default}
   default:
+    host: ${SW_RECEIVER_JETTY_HOST:0.0.0.0}
```
wu-sheng, line 159: Don't open this, because once you did, all query and trace in http need to be reported through this port, which breaks quick start and tests.

Your change should be in the doc mostly.

JaredTan95: We can also set port default to 0?

oap-server/server-receiver-plugin/skywalking-sharing-server-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/sharing/server/SharingServerModuleProvider.java
```diff
@@ -67,26 +68,38 @@ public ModuleConfig createConfigBeanIfAbsent() {
...
+                                                               .jettyAcceptorPriorityDelta(config.getJettyAcceptorPriorityDelta())
+                                                               .jettyIdleTimeOut(config.getJettyIdleTimeOut()).build();
+
+        if (config.getPort() != 0) {
+            jettyServerConfig.setHost(Strings.isBlank(config.getHost()) ? "0.0.0.0" : config.getHost());
+            jettyServerConfig.setPort(config.getPort());
+            jettyServerConfig.setContextPath(config.getContextPath());
         }
+        jettyServer = new JettyServer(jettyServerConfig);
+        jettyServer.initialize();
```
kezhenxu94, line 86: Can you explain more about this, the original logic is initializing `JettyServer` when `config.getRestPort() != 0`, but after your changes, `JettyServer` is always initialized, is it expected?

wu-sheng: The logic should not be changed. Once port is invalid, the jetty server should not be activated.

Issue #5124, opened 2020-07-18
Support alarm labels


dagmom:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [x] Question or discussion
- [ ] Bug
- [ ] Requirement
- [x] Feature or performance improvement

___
### Question
- Alarm support level

Many monitoring system alarms support levels, such as info, warn, critical.
I want to enhance the alarm rules with alert level.

wu-sheng:
What are the use cases of that? As the SkyWalking doesn't react from it, can't see the real value.

dagmom:
@wu-sheng In large data centers, there can be many alerts.
Some alarms must be dealt with immediately. And some may cause problems, we'll deal with that later.
For example successful rate of endpoint lower than 75% is warning, and lower than 50% is critical.

wu-sheng:
OK, If you are talking about large scale env, I would recommend bringing the `tag` system into an alarm message. For alarm rule, you could set multiple tags(name: value), if the alarm happens, these tags would be sent with message too. What do you think?

dagmom:
@wu-sheng 
Like prometheus alertmanager?
```yml
groups:
- name: PromRules
  rules:
  - alert: Node-Load
    annotations: {description: 'descTest',
      summary: 'summaryTest'}
    expr: node_load1 > 0.01
    for: 30s
    labels: {resType: 'Node',severity: 'warning'}
  - alert: Node-Load
    annotations: {description: 'descTest',
      summary: 'summaryTest'}
    expr: node_load1 > 0.02
    for: 30s
    labels: {resType: 'Node',severity: 'critical'}
```

wu-sheng:
About the `labels` part, yes.

wu-sheng:
No update.

liqiangz:
@wu-sheng I think this feature is useful when we have a lot of alarm rules to manage. Now I distinguish by keywords in messge. if there are labels it will be more convenient. I will solve this issue.

wu-sheng:
> @wu-sheng I think this feature is useful when we have a lot of alarm rules to manage. Now I distinguish by keywords in messge. if there are labels it will be more convenient. I will solve this issue.

Like the above comments, yes, labels make sense. This feature just ends w/o more processing. You are involving in many issues already, we only respond to the developers contributing to the ecosystem, as too many people asking.

Issue #4530, opened 2020-03-17
Add alarm tests in the e2e


wu-sheng:
As @dmsolr provide a hotfix for 7.0.0 release, we noticed, we miss one part of backend feature. Alarm test should be separated groups, we could use the mesh interface to report data which could guarantee the alarm occurs.

This could be tested in different storage too.

mrproliu:
@kezhenxu94 Hi, have you started this e2e test? I'm interested in this.

kezhenxu94:
> @kezhenxu94 Hi, have you started this e2e test? I'm interested in this.

Not yet, go ahead

PR #5156, opened 2020-07-22
Support alarm labels

dagmom:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

#5124


wu-sheng:
And your branch has not been sync from the master, and you don't allow us to sync.

dagmom:
> And your branch has not been sync from the master, and you don't allow us to sync.

@wu-sheng 
Invite you or an org to my fork repository ? or others way

wu-sheng:
> Invite you or an org to my fork repository ? or others way

I am not asking for that, I am just speaking a status. You could update the branch by yourself. The point is, unless. you update the branch with master, we can't merge your PR.

This is what I saw.
image

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/alarm/AlarmRecord.java
```diff
@@ -64,6 +66,8 @@ public String id() {
     private long startTime;
     @Column(columnName = ALARM_MESSAGE, matchQuery = true)
     private String alarmMessage;
+    @Column(columnName = LABELS, matchQuery = true)
+    private String labels;
```
wu-sheng, line 70: Adding this? Are you going to modify UI?

dagmom: I'm not going to modify UI, just store data.
I've just read the code about `@ Column` in other persistent scenario.

```java
@Column(columnName = LABELS, dataType = Column.ValueDataType.LABELED_VALUE)
private DataTable labels;
```

Is that more appropriate? But I have a question, how to search by the specified label(key=value)？

dagmom:
image

wu-sheng:
I know, you set label-values(s) in the entity. But my point is, this can't show up, unless the page changed, right?

dagmom:
> I know, you set label-values(s) in the entity. But my point is, this can't show up, unless the page changed, right?

Yes, do i need to finish UI in this PR? or next PR about UI and query.

wu-sheng:
Please make sure CI passed first, then you need a query protocol PR and UI related PR. Right?

FYI, we are closing the 8.1.0 release, so this PR will be blocked until we are ready to tagging the v8.1.0. So, make your time well planned, you have much time for query protocol and UI PRs.

wu-sheng:
For finishing this PR, you at least need to query-protocol PR merged and implement here. 
Also, considering you have enough time, please add an e2e check for the alarm case. You could take a reference from the e2e TTL, which is using the mesh telemetry to mock data, you could use the same way to mock the data which could trigger the alarm. Then you have everything you need to check. Related to #4530

dagmom:
> For finishing this PR, you at least need to query-protocol PR merged and implement here.
> Also, considering you have enough time, please add an e2e check for the alarm case. You could take a reference from the e2e TTL, which is using the mesh telemetry to mock data, you could use the same way to mock the data which could trigger the alarm. Then you have everything you need to check. Related to #4530

ok

wu-sheng:
No update for a long time.

PR #5169, opened 2020-07-24
Add Tomcat 9 install document for java agent (windows)

songzhendong:
Add Tomcat 9 install document for java agent (windows)


HarryFQ:
> Add Tomcat 9 install document for java agent (windows)

Hello, Mr. Song, you are also using skywalking in your production environment, right, do you use Kafka as a buffer?

PR #5174, opened 2020-07-26
Support regular expression in labelFilter

hanahmily:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

___
The labelFilter supports regular expression to filter labels. This feature provides a more flexible path for users to extract and aggregate metrics.

For instance,  `http_request_count` metric has a `status` label to indicate the status code. If users want to sum all of 4xx(400, 404 and etc) request, they can configure `4\d{2}` to match them.

I don't intend to pre-compile the regular expression to improve the performance due to the low frequency of the Prometheus fetcher/receiver interval. Before some performance concerns are raised, I prefer to use cleaner and simpler implement.


wu-sheng:
Why tagged as 8.2.0? 8.1.0 has not been in code frozen status.

hanahmily:
> Why tagged as 8.2.0? 8.1.0 has not been in code frozen status.

just afraid can't be merged before code frozen.

wu-sheng:
Is this still in WIP?

Issue #5186, opened 2020-07-27
Support dynamic labels in meter-system


hanahmily:
Some labels in Prometheus practice can't be predicted by users previously, in another word, they are not in a limit range or their all candidates can't be recognized. 
For example, a label indicates the `kind` of Kubernetes resource, not only there are a number of official resources, for instance, deployment, statefulset, service and etc, but users can also define any custom resource they want. 

Based on the above requirement, we should support ingesting dynamic labels into the meter-system. 

A potential list of changes might be:

 - Add a new function to store multiple labels in one single line. 
 - Update labeled query to support fetching all of the values.
 - Update the alarm submodule to support configuring rules based on label values. 
 
@Fine0830 Would you please confirm if users don't set up any labels to label select template, UI also can display a chart based on the labels queried from graphql?

Fine0830:
@hanahmily Now, UI is able to support the requirement without labels.

wu-sheng:
I have merged the pull request, the UI submodule has not been updated, please recheck the status.

Fine0830:
I have updated the UI submodule.

hanahmily:
Thanks to @Fine0830 's contribution and @wu-sheng 's review, we have finished all of subtasks.

PR #5187, opened 2020-07-27
Add AvgLabeledFunction to ingest multiple labels

hanahmily:
Related to #5186, to introduce a new meter function.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/AvgLabeledFunction.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/AvgLabeledFunction.java
@@ -20,9 +20,9 @@ package org.apache.skywalking.oap.server.core.analysis.meter.function;
 
 import java.util.Comparator;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.ToString;
@@ -78,7 +78,7 @@ public abstract class AvgLabeledFunction extends Metrics implements AcceptableVa
 
     @Override
     public final void calculate() {
-        List<String> keys = count.sortedKeys(Comparator.naturalOrder());
+        Set<String> keys = count.keys();
         for (String key : keys) {
             Long s = summation.get(key);
             if (Objects.isNull(s)) {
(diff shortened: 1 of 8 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/AvgLabeledFunction.java
```diff
@@ -0,0 +1,215 @@
...
+    @Override
+    public final void combine(Metrics metrics) {
+        AvgLabeledFunction longAvgMetrics = (AvgLabeledFunction) metrics;
+        summation.append(longAvgMetrics.summation);
+        count.append(longAvgMetrics.count);
+    }
+
+    @Override
+    public final void calculate() {
+        List<String> keys = count.sortedKeys(Comparator.naturalOrder());
```
wu-sheng, line 81: I think you should add `keys` instead of using `sortedKeys`? I noticed you used `Comparator#naturalOrder`, but base on the `Comparators#NaturalOrderComparator`, you still have to loop and reorder hashmap, it still does this.
```
        @Override
        public int compare(Comparable<Object> c1, Comparable<Object> c2) {
            return c1.compareTo(c2);
        }
```

hanahmily: yep, for a hashmap, sorting is unnecessary.

hanahmily: done

hanahmily:
> Question is this Labeled function requires the backend query change for adopting the all uncertain labels?

The changes of query will be in another PR.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/AvgLabeledFunction.java
```diff
@@ -0,0 +1,215 @@
...
+        return getTimeBucket() + Const.ID_CONNECTOR + entityId;
+    }
+
+    @Override
+    public void accept(final MeterEntity entity, final DataTable value) {
+        this.entityId = entity.id();
+        this.serviceId = entity.serviceId();
+        this.summation.append(value);
+        DataTable c = new DataTable();
+        value.sortedKeys(Comparator.naturalOrder()).forEach(key -> c.put(key, 1L));
```
wu-sheng, line 160: There is another sort.

hanahmily: done

PR #5202, opened 2020-07-29
Query all of labels once selected labels condition is absent.

hanahmily:
Related to #5186


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetricsQueryEsDAO.java
```diff
@@ -138,42 +139,12 @@ public MetricsValues readMetricsValues(final MetricsCondition condition,
...
-                        data = (long) defaultValue;
-                    }
-                    KVInt kv = new KVInt();
-                    kv.setId(id);
-                    kv.setValue(data);
-                    values.addKVInt(kv);
-                });
-            }
-
+        Map<String, DataTable> idMap = new LinkedHashMap<>();
```
wu-sheng, line 142: Others are `HashMap`, this is `LinkedHashMap`. Why?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetricsQueryDAO.java
```diff
@@ -76,5 +87,45 @@ public static IntValues sortValues(IntValues origin, List<String> expectedOrder,
...
+            final Map<String, DataTable> idMap) {
+            List<String> allLabels;
+            if (Objects.isNull(labels) || labels.size() < 1) {
+                allLabels = idMap.values().stream()
+                    .flatMap(dataTable -> dataTable.keys().stream())
+                    .distinct().collect(Collectors.toList());
+            } else {
+                allLabels = labels;
+            }
+            final int defaultValue = ValueColumnMetadata.INSTANCE.getDefaultValue(condition.getName());
```
wu-sheng, line 106: I feel like you should just copy the old codes rather than re-implement.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetricsQueryDAO.java
```diff
@@ -76,5 +87,45 @@ public static IntValues sortValues(IntValues origin, List<String> expectedOrder,
...
+            return allLabels.stream()
+                .flatMap(label -> ids.stream()
+                    .map(id -> Tuple.of(
+                        label,
+                        id,
+                        Optional.ofNullable(idMap.getOrDefault(id, new DataTable()).get(label)).orElse(0L))))
```
wu-sheng, line 112: Why `orElse(0L)`?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetricsQueryDAO.java
wu-sheng: Once you have the `allLabels`, copying is a safer way. This re-implementation is hard to be reviewed.

wu-sheng: Could you revert them? Is there any block there?

hanahmily: I used to follow copy and past strategy but found it would bring more mess more than what I expect. Unlike Es, H2 and Influxdb don't fetch all data before composing the final result object. That causes I have to tweak the current algorithm unavoidably. Future more, there are too many iterations that bring some performance concerns, I want to optimize it in the new implement.

hanahmily: I agree with that lamda is too hard to review. I'd like to improve it with other code styles.

hanahmily: And I also add a unit test case to help define the behavior of this function. Please review it to check if it cover the main scope.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetricsQueryDAO.java
hanahmily: I should be `defaultValue`.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetricsQueryEsDAO.java
hanahmily: by mistake.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetricsQueryDAO.java
hanahmily: done

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetricsQueryEsDAO.java
hanahmily: done

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetricsQueryDAO.java
hanahmily: Updated, open a new thread for review if necessary.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetricsQueryDAO.java
```diff
@@ -76,5 +87,54 @@ public static IntValues sortValues(IntValues origin, List<String> expectedOrder,
...
+            List<String> allLabels;
+            if (Objects.isNull(labels) || labels.size() < 1) {
+                allLabels = idMap.values().stream()
+                    .flatMap(dataTable -> dataTable.keys().stream())
+                    .distinct().collect(Collectors.toList());
+            } else {
+                allLabels = labels;
+            }
+            final int defaultValue = ValueColumnMetadata.INSTANCE.getDefaultValue(condition.getName());
+            List<LabeledValue> labeledValues = Sets.newTreeSet(allLabels).stream()
```
wu-sheng, line 107: According this the documents, this `#newTreeSet` is **deprecated**

```
   * <p><b>Note for Java 7 and later:</b> this method is now unnecessary and should be treated as
   * deprecated. Instead, use the {@code TreeSet} constructor directly, taking advantage of the new
   * <a href="http://goo.gl/iz2Wi">"diamond" syntax</a>.
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetricsQueryDAO.java
```diff
@@ -76,5 +87,54 @@ public static IntValues sortValues(IntValues origin, List<String> expectedOrder,
...
+                .flatMap(label -> ids.stream().map(id ->
+                    new LabeledValue(label, id, Optional.ofNullable(
+                        idMap.getOrDefault(id, new DataTable()).get(label)).orElse((long) defaultValue))))
```
wu-sheng, line 110: I think this nested expression should be better formated manually than IDE.
```suggestion
                    new LabeledValue(
                        label, 
                        id, 
                        Optional.ofNullable(idMap.getOrDefault(id, new DataTable()).get(label)).orElse((long) defaultValue)))
                    )
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetricsQueryDAO.java
```diff
@@ -76,5 +87,54 @@ public static IntValues sortValues(IntValues origin, List<String> expectedOrder,
...
+                .collect(toList());
+            MetricsValues current = new MetricsValues();
+            List<MetricsValues> result = new ArrayList<>();
+            for (LabeledValue each : labeledValues) {
+                if (!Strings.isNullOrEmpty(current.getLabel()) && each.label.equals(current.getLabel())) {
```
wu-sheng, line 115: You are using `each.label.equals`, then `!Strings.isNullOrEmpty(current.getLabel())` isn't required, right?
```suggestion
                if (each.label.equals(current.getLabel())) {
```
Or you need to check `each.label != null`, from my understanding.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetricsQueryDAO.java
hanahmily: done

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetricsQueryDAO.java
hanahmily: done.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetricsQueryDAO.java
hanahmily: `Objects.equals` is a better choice.

PR #5215, opened 2020-07-31
sql optimize : query alarm in mysql 

dagmom:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

___
- `where 1 = 1` will cause query index invalid
- use condition check instead of simple `1 = 1`


wu-sheng:
> where 1 = 1 will cause query index invalid

@tristaZero Could you confirm that? From my understanding, it shouldn't. I am away from the SQL details for a long time.

tristaZero:
@dagmom @wu-sheng My impression is that MySQL will take some optimization for this SQL, so `1 = 1` does not matter.
Here is a simple test,
``` sql
mysql> explain select * from business where id = 1;
+----+-------------+----------+------------+-------+---------------+---------+---------+-------+------+----------+-------+
| id | select_type | table    | partitions | type  | possible_keys | key     | key_len | ref   | rows | filtered | Extra |
+----+-------------+----------+------------+-------+---------------+---------+---------+-------+------+----------+-------+
|  1 | SIMPLE      | business | NULL       | const | PRIMARY       | PRIMARY | 4       | const |    1 |   100.00 | NULL  |
+----+-------------+----------+------------+-------+---------------+---------+---------+-------+------+----------+-------+
1 row in set, 1 warning (0.02 sec)

mysql> explain select * from business where 1=1 and id = 1;
+----+-------------+----------+------------+-------+---------------+---------+---------+-------+------+----------+-------+
| id | select_type | table    | partitions | type  | possible_keys | key     | key_len | ref   | rows | filtered | Extra |
+----+-------------+----------+------------+-------+---------------+---------+---------+-------+------+----------+-------+
|  1 | SIMPLE      | business | NULL       | const | PRIMARY       | PRIMARY | 4       | const |    1 |   100.00 | NULL  |
+----+-------------+----------+------------+-------+---------------+---------+---------+-------+------+----------+-------+
```

wu-sheng:
I am going to close this for now, if anything we missed, please let us know.

PR #5217, opened 2020-08-01
Fix error format of time in instance list query.

zifeihan:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
org.apache.skywalking.oap.server.storage.plugin.elasticsearch.query.MetadataQueryEsDAO#getServiceInstances,
this interface return incorrect data.

- How to fix?
Fix error format of time in org.apache.skywalking.oap.server.storage.plugin.elasticsearch.query.MetadataQueryEsDAO#getServiceInstances
___

### DIFF
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
@@ -26,7 +26,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import org.apache.skywalking.oap.server.core.analysis.NodeType;
-import org.apache.skywalking.oap.server.core.analysis.TimeBucket;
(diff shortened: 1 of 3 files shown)


wu-sheng:
@kezhenxu94 This PR should fix a bug of instance list but seems to cause the e2e broken. Could you take a look about this? Is there an existing e2e check bug too?

In the current version, the instance list will always be all until the TTL deadline, but that is caused by a wrong query condition.

kezhenxu94:
The `startTime` when verifying in the tests should be updated accordingly

https://github.com/apache/skywalking/blob/dcd66ee6e940bdc9a7ecf90415929bfd48a3a284/test/e2e/e2e-test/src/test/java/org/apache/skywalking/e2e/base/SkyWalkingTestAdapter.java#L32

```java
    protected final LocalDateTime startTime = LocalDateTime.now(ZoneOffset.UTC).minusMinutes(30);
```

zifeihan:
> I rechecked data from traffic instance data,
> 
> image
> 
> `last_ping` is time bucket format. Please recheck.

sorry, my fault. only this code,"org.apache.skywalking.oap.server.core.metric.promethues.PrometheusMetricConverter#generateTraffic",  use timestamp to update serviceTraffic. i will resend pr.
image

PR #5219, opened 2020-08-02
Fix wrong start time in e2e query

wu-sheng:
Follow https://github.com/apache/skywalking/pull/5217#issuecomment-667636827


wu-sheng:
The original PR seems having issue.

kezhenxu94:
> The original PR seems having issue.

Just recheck the `startTime`, it's correctly set to `now` before all the services in docker start, so the changes in this PR is unnecessary, it's the original PR that has issue

wu-sheng:
Closing.

Issue #5212, opened 2020-07-31
hbase-client plugin support


zhaoyuguang:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [x] Feature or performance improvement

___
### Requirement or improvement
According to the needs of our company, I will make a plug-in for HBase, which mainly tracks HBaseAdmin & HTable, and it will be implemented based on 1.3.6 as the base version. If you want me to care about other versions, please register below

PR #5225, opened 2020-08-04
Add HBase hbase-client plugin

zhaoyuguang:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues
#5212

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/hbase-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hbase/HTableInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/hbase-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hbase/HTableInterceptor.java
@@ -21,6 +21,9 @@ package org.apache.skywalking.apm.plugin.hbase;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.client.ClusterConnection;
 import org.apache.hadoop.hbase.client.HTable;
+import org.apache.hadoop.hbase.client.OperationWithAttributes;
+import org.apache.skywalking.apm.agent.core.context.CarrierItem;
+import org.apache.skywalking.apm.agent.core.context.ContextCarrier;
 import org.apache.skywalking.apm.agent.core.context.ContextManager;
 import org.apache.skywalking.apm.agent.core.context.tag.Tags;
 import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
@@ -31,10 +34,12 @@ import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedI
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+import org.apache.skywalking.apm.agent.core.util.CollectionUtil;
 import org.apache.skywalking.apm.network.trace.component.ComponentsDefine;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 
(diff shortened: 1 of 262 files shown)


apm-sniffer/apm-sdk-plugin/hbase-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hbase/define/HTableInstrumentation.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Override
+    public InstanceMethodsInterceptPoint[] getInstanceMethodsInterceptPoints() {
+        return new InstanceMethodsInterceptPoint[]{
+            new InstanceMethodsInterceptPoint() {
+                @Override
+                public ElementMatcher<MethodDescription> getMethodsMatcher() {
+                    return named("delete").or(named("put")).or(isPublic().and(named("get")))
+                            .or(named("getScanner").and(takesArguments(1))
+                                    .and(takesArgument(0, named("org.apache.hadoop.hbase.client.Scan"))));
```
wu-sheng, line 86: Does this method provide a header/metadata injection way? Or does HBase network protocol support that? If so, we should consider that, because as a Java based server, we will have a chance to tracing and monitoring the HBase server in the future.

zhaoyuguang: Let me do some research

zhaoyuguang:
image
I looked it up in the source code and the blog, and found all of the operation classes extend `OperationWithAttributes `, there's an `OperationWithAttributes attribute` in there, and some documentation says this is for writing the`coprocessor ` and the `log`, so I think it looks like we're looking for it, but i`m not sure.

wu-sheng:
If these `OperationWithAttributes` could be propagated to the HBase server side, and support the custom data w/o breaking the HBase server, I think we could use it.

apm-sniffer/apm-sdk-plugin/hbase-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hbase/HTableInterceptor.java
```diff
@@ -0,0 +1,129 @@
...
+                    operations = list;
+                    canTracingServer = true;
+                }
+            } else if (allArguments[0] instanceof OperationWithAttributes) {
+                operation = (OperationWithAttributes) allArguments[0];
+                canTracingServer = true;
+            }
+        }
+        AbstractSpan span;
+        if (canTracingServer) {
```
wu-sheng, line 72: What is the case `not instanceof OperationWithAttributes`?

zhaoyuguang: I don't know how all versions of Hbase-Client evolve, just in case

wu-sheng: OK, got it:)

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -70,6 +70,8 @@
     * [SolrJ](https://github.com/apache/lucene-solr/tree/master/solr/solrj) 7.x
   * [Cassandra](https://github.com/apache/cassandra) 3.x
     * [cassandra-java-driver](https://github.com/datastax/java-driver) 3.7.0-3.7.2
+  * Hbase
+    * [hbase-client](https://github.com/apache/hbase) HTable 1.x
```
wu-sheng, line 74: HBase client with HTable? What is the relationship between these?

oap-server/server-bootstrap/src/main/resources/component-libraries.yml
```diff
@@ -311,6 +311,9 @@ GraphQL:
 spring-annotation:
   id: 93
   languages: Java
+Hbase:
```
wu-sheng, line 314: I think all `Hbase` should be `HBase` according to the official document.

zhaoyuguang: I  fix  Htable--> HTable too, wait ci

docs/en/setup/service-agent/java-agent/Supported-list.md
zhaoyuguang: The HBase Client also has classes that interact with the remote server, such as `Admin `, which can modify tables, operate clusters, manage connections, etc. The PR only monitors the HTABLE

apm-sniffer/apm-sdk-plugin/hbase-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hbase/HTableInterceptor.java
```diff
@@ -0,0 +1,129 @@
...
+            Configuration connection = ((ClusterConnection) allArguments[1]).getConfiguration();
+            Field field = connection.getClass().getDeclaredField("overlay");
+            field.setAccessible(true);
+            Properties properties = (Properties) field.get(connection);
+            for (Map.Entry entry : properties.entrySet()) {
+                if ("hbase.zookeeper.quorum".equals(entry.getKey())) {
+                    objInst.setSkyWalkingDynamicField(entry.getValue().toString());
+                }
+            }
+        } catch (Exception e) {
```
wu-sheng, line 125: The agent core has `try-catch` already. We don't need this in the interceptor.

oap-server/server-bootstrap/src/main/resources/component-libraries.yml
```diff
@@ -493,4 +496,5 @@ Component-Server-Mappings:
   Mysqli: Mysql
   influxdb-java: InfluxDB
   Predis: Redis
-  PyMysql: Mysql
\ No newline at end of file
+  PyMysql: Mysql
+  Hase: HBase
```
wu-sheng, line 500: What is this? If they are same, you don't need mapping.

zhaoyuguang: fixed

apm-sniffer/apm-sdk-plugin/hbase-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hbase/HTableInterceptor.java
zhaoyuguang: just throw?

oap-server/server-bootstrap/src/main/resources/component-libraries.yml
wu-sheng: Please try https://github.com/apache/skywalking/pull/5232, you should not need this once the bug fixed.

oap-server/server-bootstrap/src/main/resources/component-libraries.yml
```diff
@@ -493,4 +496,5 @@ Component-Server-Mappings:
   Mysqli: Mysql
   influxdb-java: InfluxDB
   Predis: Redis
-  PyMysql: Mysql
\ No newline at end of file
+  PyMysql: Mysql
+  HBase: HBase
```
wu-sheng, line 500: Whether we don't need this? Have you checked locally?

zhaoyuguang: fixed

zhaoyuguang:
run CI AND IT

PR #5226, opened 2020-08-04
fix version forgot update

zhaoyuguang:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance


zhaoyuguang:
> If you want to change, use 8.1.0, there is no 8.2.0-snapshot. 8.1-snapshot is pushed to the apache mavem for purpose only. This is not a local build version.
> 
> Take a look on the pom.xml.

got it, and change 8.1.0

test/e2e/e2e-service-provider/pom.xml
```diff
@@ -51,7 +51,7 @@
         <dependency>
             <groupId>org.apache.skywalking</groupId>
             <artifactId>apm-toolkit-micrometer-registry</artifactId>
-            <version>8.1.0-SNAPSHOT</version>
+            <version>8.1.0</version>
```
wu-sheng, line 54: Remove the apache snapshot repo from the pom.xml, they are not necessary.

wu-sheng:
@zhaoyuguang Please do an update?

PR #5218, opened 2020-08-01
 add a commons config API for get config key/value by: Config.getProperty

qxo:
with this config API, the external plugin can get config without change the apm-agent-core Config.java.

then using as below:
````
  private static final int CUSTOM_CMP_ID = Integer.parseInt(Config.getProperty("ex_plugin_comp_id", "100"));
````

Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [*] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___
### New feature or improvement
- Describe the details and related test reports.
添加此通用配置API的目的：
* 为外部扩展插件提供一种统一的方式来存取配置项
* 原来的方式添加配置项要修改Config类的方式不适用于外部插件

PR #5249, opened 2020-08-06
Add Customization Config section for plugin development.

wu-sheng:
FYI @qxo 

As promised, the latest doc is there. It should be better for the plugin developers. 

Improve #5218

The protocol submodule update is just including many many new documents, required from the Envoy community side.

### DIFF
--- a/docs/en/guides/Java-Plugin-Development-Guide.md
+++ b/docs/en/guides/Java-Plugin-Development-Guide.md
@@ -375,6 +375,54 @@ public class URLInstrumentation extends ClassEnhancePluginDefine {
 **NOTICE**, doing bootstrap instrumentation should only happen in necessary, but mostly it effect the JRE core(rt.jar),
 and could make very unexpected result or side effect.
 
+### Provide Customization Config for the Plugin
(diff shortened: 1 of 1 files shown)


docs/en/guides/Java-Plugin-Development-Guide.md
```diff
@@ -375,6 +375,54 @@ public class URLInstrumentation extends ClassEnhancePluginDefine {
 **NOTICE**, doing bootstrap instrumentation should only happen in necessary, but mostly it effect the JRE core(rt.jar),
 and could make very unexpected result or side effect.
 
+### Provide Customization Config for the Plugin
+The config could provide different behaviours based on the configurations. SkyWalking plugin mechanism provides the configuration
+injection and initialization system in the agent core.
+
+Every plugin could declare more or more classes to represent the config by using `@PluginConfig` annotation. The agent core
```
kezhenxu94, line 382:
```suggestion
Every plugin could declare one or more classes to represent the config by using `@PluginConfig` annotation. The agent core
```

PR #5270, opened 2020-08-08
[IMPORTANT] Query traces with tags as condition

wu-sheng:
Since the beginning of SkyWalking, we have concerns of using tags as trace query conditions, but this requirement keeps showing up. In this pull request, I implement this feature in ElasticSearch 6/7 and H2/MySQL/TiDB storages. InfluxDB will be implemented by @dmsolr later.

The latest trace query GraphQL, https://github.com/apache/skywalking-query-protocol/blob/master/trace.graphqls#L50

UI and CLI could use multiple tags(`key=value`) to filter the traces with all other conditions. The example of new query looks like this
```graphql
{
  queryBasicTraces(condition : {
    traceState: ALL,
    queryOrder: BY_START_TIME,
    queryDuration : {
      step:MINUTE, start: "2020-08-08 2200", end: "2020-08-08 2300"
    },
    paging: {
      pageNum: 0,
      pageSize: 50,
      needTotal: false
    },
    # Duplicate keys are allowed, because one segment includes multiple spans, those spans could have different tag values
    tags: [{
      key: "status_code"
      value: "404"
    },{
      key: "status_code"
      value: "200"
    }]
  }
  ) {
    total,
    traces {
      segmentId,
      endpointNames,
      duration,
      traceIds
    }
  }
}
```


wu-sheng:
1. ElasticSearch is using the collection type value.
1. H2/MySQL is using the multiple physical columns to represent a logic tag column.
image

wu-sheng:
This PR is WIP, waiting for the InfluxDB implementation.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/InfluxInsertRequest.java
```diff
@@ -68,6 +71,15 @@ public InfluxInsertRequest addFieldAsTag(String fieldName, String tagName) {
         return this;
     }
 
+    public InfluxInsertRequest tags(List<SpanTag> tags) {
+        if (!Objects.isNull(tags) && !tags.isEmpty()) {
+            for (final SpanTag tag : tags) {
+                builder.addField(tag.getKey(), tag.getValue());
```
wu-sheng, line 77: Tag keys could be duplicated as it is in the different spans, what will happen here?

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/TraceQuery.java
```diff
@@ -121,6 +125,13 @@ public TraceBrief queryBasicTraces(long startSecondTB,
                 recallQuery.and(eq(SegmentRecord.IS_ERROR, BooleanUtils.FALSE));
                 break;
         }
+        if (Objects.nonNull(tags) && !tags.isEmpty()) {
```
wu-sheng, line 128: This should be replaced by `CollectionUtil#isNotEmpty`.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/TraceQuery.java
```diff
@@ -121,6 +125,13 @@ public TraceBrief queryBasicTraces(long startSecondTB,
...
+            WhereNested<WhereQueryImpl<SelectQueryImpl>> nested = recallQuery.andNested();
+            for (final SpanTag tag : tags) {
+                nested.or(contains(tag.getKey(), tag.getValue()));
```
wu-sheng, line 131: This should be `and`. All tags in the conditions should be matched.

wu-sheng: Where is the regex match you mentioned?

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/InfluxInsertRequest.java
```diff
@@ -68,6 +71,20 @@ public InfluxInsertRequest addFieldAsTag(String fieldName, String tagName) {
         return this;
     }
 
+    public InfluxInsertRequest tags(List<SpanTag> tags) {
```
wu-sheng, line 74: Local IDE shows, this method is useless.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/TraceQuery.java
wu-sheng: Oh, I see you just `contains` as existing, so could be a mismatch if there is one value includes another one's, such as `200` with `2000`.

PR #5274, opened 2020-08-08
make CPUMetricsAccessor SPI switchable and suppport other jvm

qxo:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___
add JmxCpuAccessor for support ORACLE/SUN jvm and IBM jvm by:
```
-Dskywalking.cpu_provider=org.apache.skywalking.apm.agent.core.jvm.cpu.JmxCpuAccessor
```

and should we switch default to JmxCpuAccessor ?


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/cpu/CPUProvider.java
```diff
@@ -30,10 +30,11 @@
     CPUProvider() {
         int processorNum = ProcessorUtil.getNumberOfProcessors();
         try {
-            this.cpuMetricsAccessor = (CPUMetricsAccessor) CPUProvider.class.getClassLoader()
-                                                                            .loadClass("org.apache.skywalking.apm.agent.core.jvm.cpu.SunCpuAccessor")
-                                                                            .getConstructor(int.class)
-                                                                            .newInstance(processorNum);
+            //-Dskywalking.cpu_provider=org.apache.skywalking.apm.agent.core.jvm.cpu.JmxCpuAccessor
+            final String cpuAccessorCls = System.getProperty("skywalking.cpu_provider",
```
wu-sheng, line 34: This is a wrong way to extend. 

1. System properties are supported for config injection in the Config.java.
1. Why use try catch and config at the same time?
1. Take a reference from memory gc access. The different accessors should be used based on JVM automatical detection.

qxo:
> You should make the codes detecting the JVM type automatically. And your codes include a lot try catch and reflection. Why do you write codes in this way?

Cause I don't want my code depends on oracle/sun jvm api (com.sun.management.OperatingSystemMXBean)  or ibm jvm api （com.ibm.lang.management.OperatingSystemMXBean）。
With reflection we support  oracle/sun and ibm jvm and maybe other jvm :)

wu-sheng:
> With reflection we support oracle/sun and ibm jvm and maybe other jvm

This is the thing which should be enhanced. You never need a config to activate this accessor. You could use the reflection too. If the target class(es)/method(s) found, then this could be activated. 
The current accessor should be the default. 

You should not use SPI(this word), there is no SPI. SPI is the JVM level mechanism.

qxo:
> > With reflection we support oracle/sun and ibm jvm and maybe other jvm
> 
> This is the thing which should be enhanced. You never need a config to activate this accessor. You could use the reflection too. If the target class(es)/method(s) found, then this could be activated.
> The current accessor should be the default.
> 
> You should not use SPI(this word), there is no SPI. SPI is the JVM level mechanism.

只所以让实现类可配置是为了其可扩展性：有切入点可以不用直接核实代码而在官方不支持情况提供适配置或兼容处理。

SPI只是技术术语，并不仅限于JVM吧， 而应是反过来的： JVM提供了其SPI方式的扩展机制。我们也可以有其它方式的SPI。

SPI means Service Provider Infrastructure。目的是为了让服务或实现可替换。

If it's not satisfied for you , you may change the code :)

我提交PR目的是为了回馈开源社区， 但没空细究于此了： 按你们喜欢的方式改吧。怎么同时兼容oralce/sun和ibm jvm我已经提供样例，具体如何修改只是技术实现细节问题。

wu-sheng:
My point is, there is no need for config. You could use `class#forName` to determine whether this is IBM/SUN JDK or others.

The open-source community is welcome and looking forward to the feedback, but also, at the same time, the high quality of open source project is based on the review mechanism, and making the codes as best as you can. 

I could understand your point of how to make this works, but I also hope you could understand, open-source work is a kind of the collaboration.

qxo:
> My point is, there is no need for config. You could use `class#forName` to determine whether this is IBM/SUN JDK or others.
> 
> The open-source community is welcome and looking forward to the feedback, but also, at the same time, the high quality of open source project is based on the review mechanism and making the codes as best as you can.
> 
> I could understand your point of how to make this works, but I also hope you could understand, open-source work is a kind of collaboration.

Sorry, I'm busy.

我之前特意采用反射方式就是想编译时不依赖特定jdk的特定API， 并用同一段代码逻辑就能适配不同的jdk.
不好意思，没空了， 把问题反馈上游分支目的到已达到，具体怎么改请社区酌情采用或修改吧：）

zifeihan:
@qxo @wu-sheng @kezhenxu94  hello, you can take a view `com.ibm.lang.management.OperatingSystemMXBean`, this class extends `com.sun.management.OperatingSystemMXBean`, so we use `SunCpuAccessor` to get the CPU usage is no problem.

PR #5275, opened 2020-08-08
add ignore *.log for hs_err_pid*.log/replay_pid*.log

qxo:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ OALLexer.tokens
 .vscode 
 .checkstyle
 .externalToolBuilders
+*.log
 /test/plugin/dist
 /test/plugin/workspace
 /test/jacoco/classes


wu-sheng:
This pr conflicts with master. And where are these logs from?

qxo:
> This pr conflicts with master. And where are these logs from?

hs_err_pid*.log/replay_pid*.log generated when jvm crash on maven build.

wu-sheng:
That is unexpected, why we need to ignore those? We don't have tests for JVM crashing. The ignore file is not for all caees, it just ignore for tool config or test generated files.

qxo:
> That is unexpected, why we need to ignore those? We don't have tests for JVM crashing. The ignore file is not for all caees, it just ignore for tool config or test generated files.

当构建时内存不足就会产生这些临时文件, we should ignore it just like target dir.

wu-sheng:
You could have many files, such as dump file. It is not reasonable to add ignore for all.

Issue #5062, opened 2020-07-08
Add env config control when loading agent jar


EvanLjp:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
in our department , a unified agent  jar package will be used when Helm Chart  package in all the services.
maybe .some service don't need so many detail when tracing. for example spring plugin.
however,other service need it.
nowadays.we cannot now control the loading of the unified Jars  by reading environment variables.
so i want to add the feature. loading by env config.

for example:
/plugins has 5 jar
Optional way:
1. inclues: a.jar will only load a.jar
2.excludes: a.jar will load other 4 jars.

wu-sheng:
How do you expert the option plugins?

EvanLjp:
some plans:
(1) don't care option plugins.This will break now model.
the includes or excludes only work for plugins dir.

(2) work on option plugins.
1. all load plugins position and option plugins position.
2. includes work on option plugins dir , default is excludes all
3. excludes work on plugins dir , default is includes all

wu-sheng:
I hope it could work in this way
1. `include-plugins`. When this is activated(not null), then only these plugins should be loaded. Path should be `/plugins/xxx.jar`, `optional-plugins/yyy.jar` or `activations/zzz.jar`
1. `exclude-plugins`. This should be based on the default only, `plugins/` and `activations`.

What do you think about these policies?

EvanLjp:
question:
/plugins/ has 20 jars 
/optional-plugins/ has 10 jars

when need 8 jars in plugins and 2 jars in optional-plugins ,how it works?
1. include-plugins config 10 jars.

when need 10 jars in plugins and 2 jars in optional-plugins ,how it works?
1.  include-plugins config 12 jars.

Is only when you don't use the option-plugins,  the excludes works ?


maybe this is a good idea, user only care the plugins which is selected. So excludes status should be weakening

wu-sheng:
> Is only when you don't use the option-plugins, the excludes works?

Yes and No. By reading your use case, you actually are looking for a solution for a unified base image. So, you could copy the all optional plugins into plugins, and use `excludes` to make things work.

Or, you should think again, whether `include` and `exclude` are both needed. I am feeling, `include` is required only. As that, I prefer this config called `activate-plugins`.

PR #5293, opened 2020-08-11
agent plugins selector

EvanLjp:

- [ ] New feature provided

- Related issues
resolves https://github.com/apache/skywalking/issues/5062

___


EvanLjp:
> By reading the context from #5062 , I'm not sure whether you and @wu-sheng have reached to the consensus by DMs or not, I agree with what @wu-sheng said
> 
> > By reading your use case, you actually are looking for a solution for a unified base image. So, you could copy the all optional plugins into plugins, and use excludes to make things work.
> 
> I agree that only providing an `exclude` option should be enough, users always want more plugins, unless the plugin breaks something, only under this kind of scenario, they need to `exclude` plugins, otherwise, all the plugins can be activated harmlessly

no problem

EvanLjp:
@kezhenxu94 thx for review again

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/PluginSelector.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.apache.skywalking.apm.agent.core.conf.Config.Plugin.EXCLUDE_PLUGINS;
+
+public class PluginSelector {
+
+    public List<PluginDefine> select(List<PluginDefine> pluginDefines) {
+        if (!EXCLUDE_PLUGINS.isEmpty()) {
+            List<String> includes = Arrays.asList(EXCLUDE_PLUGINS.toLowerCase().split(","));
```
kezhenxu94, line 31: You forgot to update the name `includes`, should be `excludes`

apm-sniffer/config/agent.config
```diff
@@ -90,3 +90,6 @@ logging.level=${SW_LOGGING_LEVEL:INFO}
 
 # Kafka producer configuration
 plugin.kafka.bootstrap_servers=${SW_KAFKA_BOOTSTRAP_SERVERS:localhost:9092}
+
+# elclude some plugins which define in plugins dir,
+plugin.exclude_plugins=${SW_EXCLUDE_PLUGINS:""}
```
kezhenxu94, line 95: This config item should go to [the doc](https://github.com/apache/skywalking/tree/master/docs/en/setup/service-agent/java-agent#table-of-agent-configuration-properties) as well

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/PluginSelector.java
EvanLjp: fix

apm-sniffer/config/agent.config
EvanLjp: add doc

EvanLjp:
> We need a new doc including a long list to have all plugin names, otherwise, most users would not understand. It should be linked from the `plugin.exclude_plugins` descriptions.

done

docs/en/setup/service-agent/java-agent/Plugin-list.md
```diff
@@ -0,0 +1,101 @@
...
+    - toolkit-histogram
+    - toolkit-log4j
+    - toolkit-log4j2
+    - toolkit-logback
+    - toolkit-opentracing
+    - toolkit-tag
+    - toolkit-trace
+    - undertow-2.x-plugin
+    - vertx-core-3.x
+    - zookeeper-3.4.x
```
wu-sheng, line 101: @kezhenxu94 Do you have any change to use some mechanisms to check this? Such as building a mock agent core main?

apm-sniffer/config/agent.config
```diff
@@ -90,3 +90,6 @@ logging.level=${SW_LOGGING_LEVEL:INFO}
 
 # Kafka producer configuration
 plugin.kafka.bootstrap_servers=${SW_KAFKA_BOOTSTRAP_SERVERS:localhost:9092}
+
+# exclude some plugins define in plugins dir
```
wu-sheng, line 94:
```suggestion
# Exclude activated plugins
```

EvanLjp: fix

apm-sniffer/config/agent.config
```diff
@@ -90,3 +90,6 @@ logging.level=${SW_LOGGING_LEVEL:INFO}
...
+plugin.exclude_plugins=${SW_EXCLUDE_PLUGINS:""}
```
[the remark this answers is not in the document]
EvanLjp: fix

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/PluginSelector.java
```diff
@@ -0,0 +1,38 @@
...
+
+package org.apache.skywalking.apm.agent.core.plugin;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.apache.skywalking.apm.agent.core.conf.Config.Plugin.EXCLUDE_PLUGINS;
+
+public class PluginSelector {
```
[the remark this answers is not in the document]
EvanLjp: fix

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -266,6 +266,11 @@
          * Control the length of the peer field.
          */
         public static int PEER_MAX_LENGTH = 200;
+
+        /**
+         * Excludes some plugins which define in plugins dir
```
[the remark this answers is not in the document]
EvanLjp: fix

docs/en/setup/service-agent/java-agent/Plugin-list.md
kezhenxu94: possible

EvanLjp: the list is got from my shell tools, already push

tools/plugin/scan-agent-plugins.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
wu-sheng, line 1: @kezhenxu94 Please review this script.

@EvanLjp Tool script should be used in the CI process. Like the license check script. Check the result with the markdown document, make sure the new contributors have updated the doc.

dmsolr: LGTM except for the code style

kezhenxu94: > @kezhenxu94 Please review this script.
> 
> 
> 
> @EvanLjp Tool script should be used in the CI process. Like the license check script. Check the result with the markdown document, make sure the new contributors have updated the doc.

Refer to the check-dependencies.sh script, your scrip is expected to be rewrote to check actual plugin list and the md doc, instead of simply generating a list, of course you can print the diff so that one knows what's wrong when running your script, another tip is to put this check at the very beginning of the CI stages, because it's more lightweight and doesn't depends on the final tar (while check-dependencies.sh does)

kezhenxu94: > @kezhenxu94 Please review this script.
> 
> 
> 
> @EvanLjp Tool script should be used in the CI process. Like the license check script. Check the result with the markdown document, make sure the new contributors have updated the doc.

Refer to the check-dependencies.sh script, your scrip is expected to be rewrote to check actual plugin list and the md doc, instead of simply generating a list, of course you can print the diff so that one knows what's wrong when running your script, another tip is to put this check at the very beginning of the CI stages, because it's more lightweight and doesn't depends on the final tar (while check-dependencies.sh does)

EvanLjp: push again, the script looks like check-LICENSE.sh

apm-sniffer/config/agent.config
```diff
@@ -90,3 +90,5 @@ logging.level=${SW_LOGGING_LEVEL:INFO}
 
 # Kafka producer configuration
 # plugin.kafka.bootstrap_servers=${SW_KAFKA_BOOTSTRAP_SERVERS:localhost:9092}
+#  Exclude activated plugins
+#plugin.exclude_plugins=${SW_EXCLUDE_PLUGINS:""}
```
wu-sheng, line 94: Need to pay attention on the details.
```suggestion

# Exclude activated plugins
# plugin.exclude_plugins=${SW_EXCLUDE_PLUGINS:""}
```

EvanLjp: thx

EvanLjp:
add a mock plugin in plugin list for test ci process.remove it later

wu-sheng:
CI passed, but I think which is wrong :)

tools/plugin/check-javaagent-plugin-list.sh
```diff
@@ -0,0 +1,51 @@
...
+    rm -rf temp.txt
+}
+
+function getMdJavaagentPluginList() {
+    md_javaagent_plugins_file=${WORK_DIR}/docs/en/setup/service-agent/java-agent/Plugin-list.md
+    cat  ${md_javaagent_plugins_file}|grep -v "#" |awk -F " " '{ print $2}'|sort >${MD_PLUGINS_LIST}
+}
+
+function checkDiff() {
+    diff -w -bB -U0 ${MD_PLUGINS_LIST} ${GENERNATE_PLUGINS_LIST}
```
kezhenxu94, line 42: should pass this exit code to the end of the script, the CI status depends on the final exit code

EvanLjp: yes,already fix,try again

.github/workflows/ci-it.yaml
```diff
@@ -45,6 +45,8 @@ jobs:
           bash <(curl -s https://codecov.io/bash)
       - name: 'Check Dependencies Licenses'
         run: tools/dependencies/check-LICENSE.sh
+      - name: 'Check Javaagent Plugin List'
+        run: tools/plugin/check-javaagent-plugin-list.sh
```
kezhenxu94, line 49: I suggest to move this to the beginning of the the steps (line 42) because this script is more lightweight that can fail quickly, and thus prevents unnecessary `Install & Test`

wu-sheng: Make sense, it could fail fast.

wu-sheng:
```
--- /home/runner/work/skywalking/skywalking/tools/plugin/md-javaagent-plugin-list.txt	2020-08-13 17:45:14.440143046 +0000
+++ /home/runner/work/skywalking/skywalking/tools/plugin/genernate-javaagent-plugin-list.txt	2020-08-13 17:45:14.436142998 +0000
@@ -41 +41,0 @@
-mock-for-test
@@ -92,0 +93 @@
+toolkit-kafka
```

Really detected one besides the mock. You need some fix on the doc this time. :)

EvanLjp:
already fix ,waiting to pass all ci processes

tools/plugin/check-javaagent-plugin-list.sh
```diff
@@ -41,10 +41,5 @@ function getMdJavaagentPluginList() {
 genernateJavaagentPluginList
 getMdJavaagentPluginList
 diff -w -bB -U0 ${MD_PLUGINS_LIST} ${GENERNATE_PLUGINS_LIST}
-status=$?
-echo ${status}
-rm -rf  ${MD_PLUGINS_LIST}
-rm -rf ${GENERNATE_PLUGINS_LIST}
-[[ ${status} -ne 0 ]] && exit ${status}
```
wu-sheng: What are you testing? I think we need the `exit code` to fail the CI process.

tools/plugin/check-javaagent-plugin-list.sh
```diff
@@ -0,0 +1,45 @@
...
+}
+
+function getMdJavaagentPluginList() {
+    md_javaagent_plugins_file=${WORK_DIR}/docs/en/setup/service-agent/java-agent/Plugin-list.md
+    cat  ${md_javaagent_plugins_file}|grep -v "#" |awk -F " " '{ print $2}'|grep -E '^[a-z].*'|sort|uniq|awk NF >${MD_PLUGINS_LIST}
+}
+
+genernateJavaagentPluginList
+getMdJavaagentPluginList
+diff -w -bB -U0 ${MD_PLUGINS_LIST} ${GENERNATE_PLUGINS_LIST}
```
wu-sheng, line 43: We need the thing you deleted.

kezhenxu94: The exit code of the last line of script will serve as the final exit code of the entire script, it's ok here

wu-sheng:
> LGTM, thanks for the patience

@kezhenxu94 The exit code has been removed from the shell, please recheck.

kezhenxu94:
> > LGTM, thanks for the patience
> 
> @kezhenxu94 The exit code has been removed from the shell, please recheck.

It's ok as I commented, 

> The exit code of the last line of script will serve as the final exit code of the entire script, it's ok here

And that's what we are doing in `check-LICENSE.sh`

https://github.com/apache/skywalking/blob/408a777d9d546003f75177e96152a587b26617cc/tools/dependencies/check-LICENSE.sh#L51

PR #5299, opened 2020-08-12
fix bug when extend the scope and recommend to start with 10000 or the scope greater than 1000

gnr163:

- [x ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.
  when extend the scope and recommend to start with 10000 or the scope greater than 1000   , freemarker  generate MetricsMetaInfo method will cause error such as MetricsMetaInfo("xxxx_cpm", 10,001, entityId)
- How to fix?
 
___


wu-sheng:
Could you explain a little more? I am not following.

gnr163:
> Could you explain a little more? I am not following.

At first I extended a new scope whose sourceScopeId was 10001.  And the placeholder"${sourceScopeId}" in the getMeta.ftl whitch was explained by freemarker was "10,001" , not "10001"  as expected.  Finaly the constructor was "new MetricsMetaInfo("metricsName",10,001,"id")" not  "new MetricsMetaInfo("metricsName",10001,"id")" as expected. 

reference [ https://blog.csdn.net/itchiang/article/details/7490064](url)

wu-sheng:
Got it, make sense. Thanks.

Issue #5295, opened 2020-08-11
Support elasticsearch in python agent


alonelaval:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [x] Feature or performance improvement

___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS and JRE?

- Which company or project?

- What happened?
If possible, provide a way to reproduce the error. e.g. demo application, component version.

___
### Requirement or improvement
- Please describe your requirements or improvement suggestions.

wu-sheng:
Please set the milstone and labels for your own.

alonelaval:
> Please set the milstone and labels for your own.

ok

PR #5303, opened 2020-08-12
Update elasticsearch Python Plugin Component

alonelaval:
apache/skywalking#5295


wu-sheng:
Don't use `resolve #5295` in the related PR, only link them. Because this word in `GitHub` means, close the issue after this PR merged, this is not what you expected.

alonelaval:
> Don't use `resolve #5295` in the related PR, only link them. Because this word in `GitHub` means, close the issue after this PR merged, this is not what you expected.

👌

PR #5319, opened 2020-08-15
Spring annotation Enhance

EvanLjp:

- [ ] New feature provided

- Related issues

___

Because spring annotation only match spring bean by annotation.So is would be facing so many question when using on production system.
so add a second condition regex expression to match spring bean


wu-sheng:
Please use a clean branch to do a new pull request. Thanks.

PR #5320, opened 2020-08-15
spring annotation enhance

EvanLjp:

- [ ] New feature provided

Because spring annotation only match spring bean by annotation.So is would be facing so many question when using on production system.
so add a second condition regex expression to match spring bean


apm-sniffer/optional-plugins/optional-spring-plugins/spring-annotation-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/annotations/SpringAnnotationConfig.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class SpringAnnotationConfig {
+
+    public static class Plugin {
+        @PluginConfig(root = SpringAnnotationConfig.class)
+        public static class SpringAnnotation {
+            /**
+             * regex expression to match spring bean
+             */
+            public static String REGEX_EXPRESSION = ".*Service.*,.*Dao.*,.*Repository.*";
```
kezhenxu94, line 31:
```suggestion
            public static String REGEX_EXPRESSION = ".*Service$,.*Dao$,.*DAO$,.*Repository$";
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/ClassAnnotationPackageRegexMatch.java
```diff
@@ -0,0 +1,97 @@
...
+import static net.bytebuddy.matcher.ElementMatchers.isAnnotatedWith;
+import static net.bytebuddy.matcher.ElementMatchers.isInterface;
+import static net.bytebuddy.matcher.ElementMatchers.nameMatches;
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.not;
+
+/**
+ * Match the class by the given annotations and regex expression matching package name.
+ */
+public class ClassAnnotationPackageRegexMatch implements IndirectMatch {
```
kezhenxu94, line 38: This class is very case-specific, please consider reusing `org.apache.skywalking.apm.agent.core.plugin.match.logical.LogicalMatchOperation` to perform the logical operations

apm-sniffer/optional-plugins/optional-spring-plugins/spring-annotation-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/annotations/SpringAnnotationConfig.java
wu-sheng: This should not You set this works in default, which could make users confused, because, once their classes are not matching these default rules, they will not be traced.

So, this should be default OFF.

apm-sniffer/config/agent.config
```diff
@@ -93,3 +93,7 @@ logging.level=${SW_LOGGING_LEVEL:INFO}
 
 # Exclude activated plugins
 # plugin.exclude_plugins=${SW_EXCLUDE_PLUGINS:""}
+
+# Optional Plugins config
```
wu-sheng, line 97:
```suggestion
```

Many plugins above are optional too. If you want to indicate some are optional, which requires activating first, then we could work on this in a separate PR.

apm-sniffer/optional-plugins/optional-spring-plugins/spring-annotation-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/annotations/SpringAnnotationConfig.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class SpringAnnotationConfig {
+
+    public static class Plugin {
+        @PluginConfig(root = SpringAnnotationConfig.class)
+        public static class SpringAnnotation {
+            /**
+             * regex expression to match spring bean
+             */
+            public static String PACKAGE_MATCH_REGEX_EXPRESSION = ".*";
```
wu-sheng, line 31: 1. I think this is the Class Name match? Not a package name?
1. The default value should be empty, which is more high-efficiency.

apm-sniffer/optional-plugins/optional-spring-plugins/spring-annotation-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/annotations/services/SpringServicesInstrumentation.java
```diff
@@ -19,16 +19,21 @@
...
+import static org.apache.skywalking.apm.agent.core.plugin.match.RegexMatch.byRegexMatch;
 
 public class SpringServicesInstrumentation extends AbstractSpringBeanInstrumentation {
 
     public static final String ENHANCE_ANNOTATION = "org.springframework.stereotype.Service";
 
     @Override
     protected ClassMatch enhanceClass() {
-        return byClassAnnotationMatch(new String[] {ENHANCE_ANNOTATION});
+        return LogicalMatchOperation.and(
```
wu-sheng, line 34: Should only activate `and byRegexMatch` if the  `#getRegexExpressions` return not null. Then, it is better for matching mechanism performance.

EvanLjp: resolved

apm-sniffer/optional-plugins/optional-spring-plugins/spring-annotation-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/annotations/SpringAnnotationConfig.java
EvanLjp: resolved

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -151,6 +150,8 @@ property key | Description | Default |
 `plugin.kafka.topic_segment` | Specify which Kafka topic name for traces data to report to. | `skywalking_segments` |
 `plugin.kafka.topic_profilings` | Specify which Kafka topic name for Thread Profiling snapshot to report to. | `skywalking_profilings` |
 `plugin.kafka.topic_management` | Specify which Kafka topic name for the register or heartbeat data of Service Instance to report to. | `skywalking_managements` |
+`plugin.springannotation.classname_match_regex_expression` | Match spring bean with regex expression for classname. | `the spring beans tagged with @Bean,@Service,@Dao or @Repository` |
```
wu-sheng, line 153:
```suggestion
`plugin.springannotation.classname_match_regex_expression` | Match spring beans with regex expression for the class name. Multiple expressions could be separated by a comma. This only works when `Spring annotation plugin` has been activated. | All the spring beans tagged with @Bean,@Service,@Dao, or @Repository. |
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/RegexMatch.java
```diff
@@ -0,0 +1,68 @@
...
+
+package org.apache.skywalking.apm.agent.core.plugin.match;
+
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+import static net.bytebuddy.matcher.ElementMatchers.nameMatches;
+
+/**
+ * Match the class by the given annotations and regex expression matching package name.
```
wu-sheng, line 27:
```suggestion
 * Match the class by the given annotations and regex expression matching class name.
```

EvanLjp: thx for your notice. Has been resolved. Would be careful next time.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/RegexMatch.java
```diff
@@ -0,0 +1,68 @@
...
+    @Override
+    public boolean isMatch(TypeDescription typeDescription) {
+        boolean isMatch = false;
+        for (String matchExpression : regexExpressions) {
+            isMatch = isMatch || typeDescription.getTypeName().matches(matchExpression);
+        }
+        return isMatch;
+    }
+
+    private ElementMatcher.Junction buildEachMatchExpression(String matchExpression) {
```
wu-sheng, line 61: Question, this is a `one-line` method, why need this?

EvanLjp: sorry,has been removed

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/RegexMatch.java
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+        return regexJunction;
+    }
+
+    @Override
+    public boolean isMatch(TypeDescription typeDescription) {
+        boolean isMatch = false;
+        for (String matchExpression : regexExpressions) {
+            isMatch = isMatch || typeDescription.getTypeName().matches(matchExpression);
```
wu-sheng, line 56: This is just a reminder, (I don't know whether existing matches have this issue), even a type has been matched, it still needs to iterate the whole expression array. There should be a break.

EvanLjp: thx for your notice.already add break

PR #5327, opened 2020-08-16
Focus to record 'status_code' tag

ascrutae:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance


wu-sheng:
I think we could keep `not record status code` if it is 200, because most people don't care in this case, and it could save network and storage(a little).

PR #5339, opened 2020-08-17
Add support for spring @Scheduled

hailin0:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___

Hi~ Now we use spring @Scheduled to execute scheduled tasks, but it seems to be not supported. 
So I add the client side support.  Please see if it makes sense.


.github/workflows/plugins-test.3.yaml
```diff
@@ -55,6 +55,7 @@ jobs:
           - { name: 'hbase-scenario', title: 'hbase-scenario (5)' }
           - { name: 'spring-kafka-2.2.x-scenario', title: 'Spring-Kafka 2.2.x (7)' }
           - { name: 'spring-kafka-2.3.x-scenario', title: 'Spring-Kafka 2.3.x (7)' }
+          - { name: 'spring-scheduled-scenario', title: 'Spring Scheduled 3.1.x-5.2.x (119)' }
```
wu-sheng, line 58: Don't use all 119 versions, it costs too many CI resources and time. Just choose the typical ones, such as choose the latest version in 5.0.x to represent all versions.

test/plugin/scenarios/spring-scheduled-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,48 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+segmentItems:
+  - serviceName: spring-scheduled-scenario
+    segmentSize: ge 2
+    segments:
```
wu-sheng, line 19: From my reading of your test case codes, there should be a segment starting with `SchedulingJob#work`, where it is? Or do I miss anything?

apm-sniffer/apm-sdk-plugin/spring-plugins/scheduled-annotation-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/scheduled/ScheduledMethodInterceptor.java
```diff
@@ -0,0 +1,79 @@
...
+public class ScheduledMethodInterceptor implements InstanceMethodsAroundInterceptor, InstanceConstructorInterceptor {
+
+    private static final String DEFAULT_OPERATION_NAME = "SpringScheduled";
+
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
+        String targetMethodName = (String) objInst.getSkyWalkingDynamicField();
+        String operationName = targetMethodName != null ? targetMethodName : DEFAULT_OPERATION_NAME;
+
+        AbstractSpan span = ContextManager.createEntrySpan(operationName, new ContextCarrier());
```
wu-sheng, line 41: This is just a local scheduler task, not an entry. Please read the plugin dev document. This should be a local span. Could you share why you put this as Entry Span?

test/plugin/scenarios/spring-scheduled-scenario/config/expectedData.yaml
hailin0: Yes, it's on [line 37](https://github.com/apache/skywalking/pull/5339/files/7f6bba6a27a0ba0ccb41c639f0002e5276b7de3a#diff-f1c217f8303ef2da8c91cd14aaa67300R37).

wu-sheng: You missed the segment from the peer. There should be entry span representing the server side of this exit span.

I know it is not the thing you added, but it should work, and the segment is expected.

apm-sniffer/apm-sdk-plugin/spring-plugins/scheduled-annotation-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/scheduled/ScheduledMethodInterceptor.java
hailin0: Thank you, I will read the plugin dev document.
I referenced [elasticjob plugin](https://github.com/apache/skywalking/blob/master/apm-sniffer/apm-sdk-plugin/elastic-job-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/elasticjob/ElasticJobExecutorInterceptor.java#L39)

wu-sheng: If it is an entry, then it is a bug. Could you confirm? If yes. please send a pull request to fix.

And if you want to analysis the performance of the job, consider to use logic endpoint tag, which is also included in the dev doc, too.

wu-sheng: @zhaoyuguang Please recheck the ElasticJob plugin too.

zhaoyuguang: > @zhaoyuguang Please recheck the ElasticJob plugin too.

Well, I'm using it the wrong way

zhaoyuguang:
When annotation plugin spring-annotation-plugin is introduced, What's going to happen? It doesn't seem very friendly to 'spring-annotation-plugin'

wu-sheng:
> When annotation plugin spring-annotation-plugin is introduced, What's going to happen? It doesn't seem very friendly to 'spring-annotation-plugin', What do you think

@zhaoyuguang What is the issue? Does any old plugin cover this case?

hailin0:
> When annotation plugin spring-annotation-plugin is introduced, What's going to happen? It doesn't seem very friendly to 'spring-annotation-plugin', What do you think

I tested them and everything was normal.

CD3D86CCE385769C58EB7BE7FCF8FD79

wu-sheng:
Old plugin is for Spring bean mostly, I think.

hailin0:
> Old plugin is for Spring bean mostly, I think.

yes！

*spring-annotation-plugin* support  @Service、@Repository、@Component、@Bean
*spring-async-annotation-plugin* support  @Async
*mvc-annotation-plugin*  rename operationName...

wu-sheng:
I think currently it is fine. Let's keep the legacy name for now. If we want to make it more accurate,  let's do that in a separate PR.

apm-sniffer/apm-sdk-plugin/spring-plugins/scheduled-annotation-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/scheduled/ScheduledMethodInterceptor.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+import org.apache.skywalking.apm.network.trace.component.ComponentsDefine;
+
+import java.lang.reflect.Method;
+
+public class ScheduledMethodInterceptor implements InstanceMethodsAroundInterceptor, InstanceConstructorInterceptor {
+
+    private static final String DEFAULT_OPERATION_NAME = "SpringScheduled";
+    private static final String DEFAULT_LOGIC_ENDPOINT_CONTENT = "{\"logic-span\":true}";
```
wu-sheng, line 35: Could you move this into `Tags` as more plugins are going to use this? And name this as `VAL_LOCAL_SPAN_AS_LOGIC_ENDPOINT·

apm-sniffer/apm-sdk-plugin/spring-plugins/scheduled-annotation-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/scheduled/ScheduledMethodInterceptor.java
```diff
@@ -0,0 +1,81 @@
...
+    @Override
+    public void onConstruct(EnhancedInstance objInst, Object[] allArguments) throws Throwable {
+        Object targetMethod = allArguments[1];
+        String targetMethodName = getTargetMethodName(targetMethod);
+
+        objInst.setSkyWalkingDynamicField(targetMethodName);
+    }
+
+    private String getTargetMethodName(Object targetMethod) {
+        if (targetMethod instanceof String) {
```
wu-sheng, line 67: `Instanceof` is not recommended in the interceptor. Usually, we set up two instrumentation for constructures with different parameters(if necessary)

wu-sheng (commented):
Basically, you need to put comments on your codes.

apm-sniffer/apm-sdk-plugin/spring-plugins/scheduled-annotation-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/scheduled/ScheduledMethodInterceptor.java
hailin0: Add logic endpoint into others plugin？I could try

wu-sheng: Move this value constant into the Tags(core class), and do a little refactor to GraphQL plugin, by sharing this new constant

hailin0: ok

test/plugin/scenarios/spring-scheduled-scenario/src/main/java/org/apache/skywalking/apm/testcase/spring/scheduled/job/SchedulingJob.java
```diff
@@ -0,0 +1,48 @@
...
+
+@Component
+@EnableScheduling
+public class SchedulingJob {
+
+    private static final Logger logger = LogManager.getLogger(SchedulingJob.class);
+    
+    private static final OkHttpClient client = new OkHttpClient.Builder().build();
+
+    @Scheduled(fixedDelay = 5000)
```
wu-sheng, line 40: @hailin0, I think I got @zhaoyuguang 's point as he DM me.

This bean has `@Component` already, so, why this method wouldn't work? I think Spring annotation plugin tracing all methods of the bean. Unless you only want to activate this, rather than other beans?

apm-sniffer/apm-sdk-plugin/spring-plugins/scheduled-annotation-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/scheduled/ScheduledMethodInterceptor.java
wu-sheng: This has been done in #5347

wu-sheng: I will merge that soon, please update the codes after that.

test/plugin/scenarios/spring-scheduled-scenario/src/main/java/org/apache/skywalking/apm/testcase/spring/scheduled/job/SchedulingJob.java
hailin0: I found some bugs...  @wu-sheng 

Expected to be normal when I use @Configuration
-------------------

9CFE35D62F59B825122F05C9D3555F27
A37D3CE0897B9D84C08F71E841E6AF46


Duplicate tracing occurs when I use @Component
-------------------
0C5D9D790774A98490A8E3513A7EFCE8
1A2F323E9BFF2FC94BD11ACCD690C27B

hailin0: Maybe the @Scheduled method should be filtered out in spring-annotation-plugi.

1. @Scheduled is not an optional plugin
2. spring-scheduler can be started through configuration files instead of just annotations

wu-sheng: The first one is easy to explain, we don't watch the configuration. The second one is not clear, why `Duplicate tracing occurs when I use @component`? Who will trace twice?

wu-sheng: Does this duplicated tracing cause by this new plugin with `@component`?

wu-sheng: Note, I know this plugin should be useful at some points, but I need to make it clear.

hailin0: > The first one is easy to explain, we don't watch the configuration. The second one is not clear, why `Duplicate tracing occurs when I use @component`? Who will trace twice?

spring-scheduled-annotation-plugin trace once
then spring-annotation-plugin(optional-plugin) trace it again

wu-sheng: OK, seems byte-buddy added some features. FYI @apache/skywalking-committers. It used to can't instrument a class twice, but this test proved it is not the case today.

wu-sheng: According to the official doc, https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Configuration.html

`@Configuration` should only be used for `@Bean`. Is there abuse? I am not working on Spring for a long time, could you explain more about the use case.

hailin0: > According to the official doc, https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Configuration.html
> 
> `@Configuration` should only be used for `@Bean`. Is there abuse? I am not working on Spring for a long time, could you explain more about the use case.

The document describes @Configuration and @Scheduled can be used together

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/annotation/EnableScheduling.html

wu-sheng: >  @configuration and @scheduled can be used together

Yes, but with `component` is the right usage?

hailin0: > > @configuration and @scheduled can be used together
> 
> Yes, but with `component` is the right usage?

The document indicates that `@Component` and `@Scheduled` can also be used together

hailin0: > > > @configuration and @scheduled can be used together
> > 
> > 
> > Yes, but with `component` is the right usage?
> 
> The document indicates that `@Component` and `@Scheduled` can also be used together

Duplicate tracing occurs when `@Component` and `@Scheduled` are used together

wu-sheng: This is a tricky point, I think maybe you should consider put this plugin as optional too. And add a FAQ doc to describe these cases, and provide recommended solutions.

hailin0: OK，I'll try

hailin0: I need to set it's an optional plugin? spring-annotation-plugin is already in optional list

hailin0: Do you think there is a problem with using @Component and @Async generate two spans?

It seems that @Component, @Scheduled and @Component, @Async are similar scenarios

hailin0: > Do you think there is a problem with using @component and @async generate two spans?
> 
> It seems that `@Component`, `@Scheduled` and `@Component`, `@Async` are similar scenarios

wu-sheng: > I need to set it's an optional plugin? 

I mean this scheduled plugin.

> Maybe we should keep the status quo and use SpringScheduled as a operationName.

You could have two, but only `SpringScheduled` as an OP name(also will be an endpoint) is not a good idea, you could use `SpringScheduled-` + `full class name` + `method name` as a better operation name. Because from the analysis perspective you need a meaningful name.

hailin0: > You could have two, but only `SpringScheduled` as an OP name(also will be an endpoint) is not a good idea, you could use `SpringScheduled-` + `full class name` + `method name` as a better operation name. Because from the analysis perspective you need a meaningful name.

Yes, I agree with you.

So, are these the final conclusions?
1. spring-scheduled-annotation-plugin is not an optional plugin
2. use `SpringScheduled- + full class name + method name` as a better operation name.

wu-sheng: Yes, agree.

Issue #5356, opened 2020-08-20
Support for JSON logger


lujiajing1126:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [x] Feature or performance improvement
___
### Motivation:
In our `Kubernetes` cluster, we output logs to `STDOUT` and then collect through `Kubernetes` log driver. And the format of logs we use is `JSON` which can be easily parsed by most collector agents, for example `logstash`, `logtail`, etc. 

The critical problem of the pattern-based log format is that it is very difficult to parse, search and view from a structural aspect. In some edge cases, we have to write complex Regex to parse this log.

So I think it is necessary to implement `JsonLogger` in skywalking and allow users to switch between `PatternLogger` and `JsonLogger`. 

### Initiative:

1. Add an implementation of `JsonLogger` in `agent-core`.
2. Allow to switch `JsonResolver` before `premain` entry is invoked.

I would like to submit a PR to implement this feature.

PR #5357, opened 2020-08-20
JsonLogger implementation (#5356)

lujiajing1126:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

#5356 

- [x] Implement `JsonLogger` and `JsonResolver`. `Json` serialization with `Gson`
- [x] Allow user to specify `Resolver` via SystemProperty, for example, `-Dskywalking.logger=JSON`
- [x] Additional tests


wu-sheng:
There is a pattern expression in the config, why don't you config it in a JSON format?

lujiajing1126:
> There is a pattern expression in the config, why don't you config it in a JSON format?

Then escape of those special character like `"` would be a problem.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
```diff
@@ -203,4 +213,25 @@ public void onError(final String typeName,
         public void onComplete(String typeName, ClassLoader classLoader, JavaModule module, boolean loaded) {
         }
     }
+
+    static void configureLogger() {
```
wu-sheng, line 217: Please read the existing source codes, all config should be controlled through `Config`, which has supported dynamic injection already.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -259,6 +260,43 @@
          * @see org.apache.skywalking.apm.agent.core.logging.core.PatternLogger#DEFAULT_CONVERTER_MAP
          */
         public static String PATTERN = "%level %timestamp %thread %class : %msg %throwable";
+
+        public static class JSON {
+            /**
+             * Key of the AgentName
+             */
+            public static String AGENT_NAME_KEY = "agent_name";
```
wu-sheng, line 268: I don't think these belong to Config.

wu-sheng (commented):
Your codes need to follow the existing config injection pattern.

wu-sheng:
> The critical problem of the pattern-based log format is that it is very difficult to parse, search and view from a structural aspect. In some edge cases, we have to write complex Regex to parse this log.

To be honest, I want to challenge this. The logs are very traditional things, I haven't seen which log framework put the JSON format output as a high priority thing. `Parsing log` in the logstash(or similar thing) is existing for a long time, and powerful enough.
Also, what is the point to analyze SkyWalking agent's log? It nearly outputs nothing.

FYI @kezhenxu94 @JaredTan95 and @apache/skywalking-committers , are any of you feel this is necessary?

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
lujiajing1126: The problem here is now the resolver is set to `PatternResolver` by default.

We have to setup the resolver before any `LogManager.getLogger(...)` is called. So this method is supposed to used in the static block of `SkywalkingAgent`.

wu-sheng: I am not saying this, take a closer look about how we initialize the Config's fields.

wu-sheng:
I am forwarding this discussion(whether accept this) to the mail list. 

> Mail list: dev@skywalking.apache.org. 

Mail to dev-subscribe@skywalking.apache.org, follow the reply to subscribe the mail list.

You could read the mail list from the web page, https://lists.apache.org/list.html?dev@skywalking.apache.org. But you have to subscribe the mail to interact with others.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
lujiajing1126: > I am not saying this, take a closer look about how we initialize the Config's fields.

Yes. I understand the dynamic injection is implemented in `SnifferConfigInitializer` with the help of `ConfigInitializer`. 

Now I take advantage of `ConfigInitializer`, in which the `logger` is declared but actually not used, to inject the system properties, though limitation is still there.

wu-sheng: The variable should not be injected through the way you did.

lujiajing1126: > The variable should not be injected through the way you did.

Well. I remove this static block and first init config as it previously did. Then I reconfigure the `Logger` after config initialization finishes.

The procedure now completely follows the rules except the first few log calls in `SnifferConfigInitializer` will still use the default logger.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
wu-sheng: This question is unresolved, who will set these keys? Are these configurable?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
```diff
@@ -40,7 +42,7 @@
  * The <code>SnifferConfigInitializer</code> initializes all configs in several way.
  */
 public class SnifferConfigInitializer {
-    private static final ILog logger = LogManager.getLogger(SnifferConfigInitializer.class);
+    private static ILog LOGGER = LogManager.getLogger(SnifferConfigInitializer.class);
```
wu-sheng, line 45: Why change this?

lujiajing1126: The name `logger` cannot pass style-check. Maybe something wrong with the style-check preference?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
```diff
@@ -207,4 +212,15 @@ private static InputStreamReader loadConfig() throws AgentPackageNotFoundExcepti
         }
         throw new ConfigNotFoundException("Failed to load agent.config.");
     }
+
+    static void configureLogger() {
```
wu-sheng, line 216: This is not a good way to do this switch. I am waiting for the PMC confirmation about whether JSON format is acceptable. If YES, we need to improve this.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
lujiajing1126: > This question is unresolved, who will set these keys? Are these configurable?

Yes. It can be configured. They are used as key of the generated json map

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -249,6 +251,11 @@
          */
         public static LogOutput OUTPUT = LogOutput.FILE;
 
+        /**
+         * The log resolver type. Default is PATTERN which will create PatternLogResolver later.
+         */
+        public static ResolverType LOGGER = ResolverType.PATTERN;
```
wu-sheng, line 257:
```suggestion
        public static ResolverType RESOLVER = ResolverType.PATTERN;
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
wu-sheng: Technically, yes, my question is more about, why and who will config this.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -259,6 +266,43 @@
...
+        public static class JSON {
+            /**
+             * Key of the AgentName
+             */
+            public static String AGENT_NAME_KEY = "agent_name";
+
+            /**
+             * Key of the timestamp
+             */
+            public static String TIMESTAMP_KEY = "@timestamp";
```
wu-sheng, line 279: Why starting with `@` in JSON?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
wu-sheng: It used to be able to pass, then it still can. All codes in the master branch have to pass the style check. Please recheck again.

In the rule, we set an exception for the logger due to the legacy issue.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/Converter.java
```diff
@@ -19,10 +19,13 @@
...
 
 /**
- * The Converter, It is used to convert the LogEvent to the String.
+ * The Converter, it is used to convert the LogEvent to the String.
  */
 public interface Converter {
 
     String convert(LogEvent logEvent);
 
+    default String getKey() {
```
wu-sheng, line 28: Once you add this, why need a default implementation?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/JsonLogger.java
```diff
@@ -0,0 +1,60 @@
...
+    public JsonLogger(String targetClass, Gson gson) {
+        super(targetClass);
+        this.gson = gson;
+        for (Map.Entry<String, Class<? extends Converter>> entry : DEFAULT_CONVERTER_MAP.entrySet()) {
+            try {
+                if (converters instanceof LiteralConverter) {
+                    continue;
+                }
+                converters.add(entry.getValue().newInstance());
+            } catch (IllegalAccessException | InstantiationException ignore) {
```
wu-sheng, line 43: Why you hide the expression? What happens once these happen?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/JsonLogger.java
```diff
@@ -0,0 +1,60 @@
...
+            }
+        }
+    }
+
+    protected void logger(LogLevel level, String message, Throwable e) {
```
wu-sheng, line 48: Why need to override this method? I think this method should be in `AbstractLogger` too, and the two implementations only have their own `format` method(an abstract method in the parent).

apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/logging/core/LoggingBenchmark.java
```diff
@@ -0,0 +1,64 @@
...
+import com.google.gson.Gson;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+
+import java.util.concurrent.TimeUnit;
+
+public class LoggingBenchmark {
```
wu-sheng, line 30: Traditionally, we put the benchmark result in the class comments. Then other developers could have a clear result at the first time, rather than running on its own laptop.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
```diff
@@ -65,63 +66,68 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
 
             pluginFinder = new PluginFinder(new PluginBootstrap().loadPlugins());
         } catch (AgentPackageNotFoundException ape) {
-            logger.error(ape, "Locate agent.jar failure. Shutting down.");
+            LOGGER = LogManager.getLogger(SkyWalkingAgent.class);
```
wu-sheng, line 69: You should make `SnifferConfigInitializer.initializeCoreConfig(agentArgs);` in a separate `try/catch/finally` to reset the logger, please do this so many times.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
kezhenxu94: @lujiajing1126 , what kind of style check do you mean, we don't check that, did you use a plugin in the IDE that's not relevant to SkyWalking's code style?

wu-sheng (commented):
1. LOGGER doesn't have to rename, please revert.
1. Your codes are required to add comments for all of them
1. Other comments inline.

wu-sheng:
The `logger` related code style is here, https://github.com/apache/skywalking/blob/4e821415ea8b930d3c089346d86c650788cbbf8a/apm-checkstyle/checkStyle.xml#L82.

It is not popular, but this is the legacy. We could change it all in another one PR, but don't mix it here.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
lujiajing1126: > @lujiajing1126 , if you have a `final`, it can be lowercase `logger`, why remove `final`?

Yes, exactly. It cannot pass without `final`. 

I remove `final` here because I want to re-create the `logger` instance with the updated logger resolver set up just after the config initialization.  On line 105 the logger will possibly be called again.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/Converter.java
lujiajing1126: > Once you add this, why need a default implementation?

We don't need to implement `getKey()` for `LiteralConverter`

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
lujiajing1126: > Technically, yes, my question is more about, why and who will config this.

The application developer can config this provided that they use a different json key set for their application. So unification of these keys may be necessary for them.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
lujiajing1126: Just a convention used in [`logstash-logback-encoder`](https://github.com/logstash/logstash-logback-encoder#providers-for-loggingevents)

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/JsonLogger.java
lujiajing1126: Good idea

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/JsonLogger.java
```diff
@@ -0,0 +1,60 @@
...
+        WriterFactory.getLogWriter().write(generateJson(level, message, e));
+    }
+
+    String generateJson(LogLevel level, String message, Throwable e) {
+        LogEvent logEvent = new LogEvent(level, message, e, this.targetClass);
+        Map<String, String> log = new HashMap<>();
+        for (Converter converter : this.converters) {
+            log.put(converter.getKey(), converter.convert(logEvent));
```
wu-sheng, line 56: But this could be a potential bug in the future, as `getKey` returns empty string in default, so they could override with each other.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
wu-sheng: I think know why they are necessary. I prefer don't provide an over-configurable feature. If someone wants it, they could add them by providing a real use case.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
wu-sheng: Then, I have a feeling, this PR should be after the all `logger` and the exception rule to be removed. I don't want the codes with 2 style logger.

apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/logging/core/LoggingBenchmark.java
lujiajing1126: Report added

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/JsonLogger.java
lujiajing1126: Implemented

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
lujiajing1126: Changed here

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/JsonLogger.java
lujiajing1126: I agree. Default behavior has been removed

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/JsonLogger.java
lujiajing1126: Normally this would not happen if and only if someone change the constructor of the `Converter`s to private or so.

It will just iterate over the next `Converter` without having side effect.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
lujiajing1126: Pls check again the current modification

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
lujiajing1126: > Then, I have a feeling, this PR should be after the all `logger` and the exception rule to be removed. I don't want the codes with 2 style logger.

Agree. I have reverted the styles, just wait for the polish of the style-check rules.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
wu-sheng: Why we need to adopt logstash?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/JsonLogger.java
wu-sheng: But eating the exception is not expected in the core.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
wu-sheng: Would you like to do that first? I may not have time to do this recently. You could remove the old exception rule, once style check passed, you have changed all of them.

lujiajing1126: > Would you like to do that first? I may not have time to do this recently. You could remove the old exception rule, once style check passed, you have changed all of them.

Sure. I can do this. Do you prefer a new PR with only rule changes?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
lujiajing1126: > Why we need to adopt logstash?

I suppose it would be better to comply to the conventions which are common used by others?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
lujiajing1126: > I think know why they are necessary. I prefer don't provide an over-configurable feature. If someone wants it, they could add them by providing a real use case.

I partially agree in the sense that some keys like `agent_name` are rarely used by real application, so it can be just a constant (I can remove it). 

But the others, for example the concepts of "@timestamp", "message", "class" are also adopted by other logging framework. In order to simplify the preference of collectors, I suppose it is necessary to just keep these configurations here. They are very similar to what "pattern" does here.

I think the most users can keep the default settings like we always do.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/JsonLogger.java
lujiajing1126: I see. Then re-thorwing another RuntimeException may be better here.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
wu-sheng: Rule change and code adoption, otherwise, the CI will fail.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
```diff
@@ -75,14 +84,14 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
...
         AgentBuilder agentBuilder = new AgentBuilder.Default(byteBuddy).ignore(
-            nameStartsWith("net.bytebuddy.").or(nameStartsWith("org.slf4j."))
-                                            .or(nameStartsWith("org.groovy."))
-                                            .or(nameContains("javassist"))
-                                            .or(nameContains(".asm."))
-                                            .or(nameContains(".reflectasm."))
-                                            .or(nameStartsWith("sun.reflect"))
-                                            .or(allSkyWalkingAgentExcludeToolkit())
-                                            .or(ElementMatchers.isSynthetic()));
+                nameStartsWith("net.bytebuddy.").or(nameStartsWith("org.slf4j."))
```
wu-sheng, line 87: This reformat seems very strange. Why do this? If you want to reformat, this is a better choice.
```
        AgentBuilder agentBuilder = new AgentBuilder.Default(byteBuddy).ignore(
            nameStartsWith("net.bytebuddy.")
                .or(nameStartsWith("org.slf4j."))
                .or(nameStartsWith("org.groovy."))
                .or(nameContains("javassist"))
                .or(nameContains(".asm."))
                .or(nameContains(".reflectasm."))
                .or(nameStartsWith("sun.reflect"))
                .or(allSkyWalkingAgentExcludeToolkit())
                .or(ElementMatchers.isSynthetic()));
```

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
```diff
@@ -109,10 +118,10 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
         }
 
         agentBuilder.type(pluginFinder.buildMatch())
-                    .transform(new Transformer(pluginFinder))
-                    .with(AgentBuilder.RedefinitionStrategy.RETRANSFORMATION)
-                    .with(new Listener())
-                    .installOn(instrumentation);
+                .transform(new Transformer(pluginFinder))
```
wu-sheng, line 121: Same strange format here.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
wu-sheng: > for example the concepts of "@timestamp", "message", "class" are also adopted by other logging framework.

Please provide the official documentation. If they are there, we could adopt it. But either way, still it is not necessary to make them configurable.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
wu-sheng: Discussion continues on https://github.com/apache/skywalking/pull/5357/files#r473956466

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
lujiajing1126: Oh. Just something strange with my editor, I will try to improve them.

Any standard format plugin or rules we can use for IDEA?

lujiajing1126:
> The `logger` related code style is here,
> 
> https://github.com/apache/skywalking/blob/4e821415ea8b930d3c089346d86c650788cbbf8a/apm-checkstyle/checkStyle.xml#L82
> 
> .
> It is not popular, but this is the legacy. We could change it all in another one PR, but don't mix it here.

I just want to make sure what you meannt, whether the "|(^logger)" option should be removed from this legacy format, i.e. every `logger` should be rename to `LOGGER`.

Or just declare `logger` as another exception of valid `StaticVariableName`s

https://github.com/apache/skywalking/blob/4e821415ea8b930d3c089346d86c650788cbbf8a/apm-checkstyle/checkStyle.xml#L103

wu-sheng:
I mean **every logger should be rename to LOGGER**, no exception. The exception should be removed from the check rules.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
lujiajing1126: Fixed

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
lujiajing1126: Fixed

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/JsonLogger.java
lujiajing1126: Fixed

lujiajing1126:
> I mean **every logger should be rename to LOGGER**, no exception. The exception should be removed from the check rules.

The recent PR #5363 has been accepted, so rebase this PR.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
lujiajing1126: The `@timestamp` is used in logstash by default, the source code can be find here, https://github.com/elastic/logstash/blob/0d82bc064ca8184722d159e3e8036ffe1931e666/logstash-core/src/main/java/org/logstash/Event.java#L54

The classic `logback` (which is the default logging framework in Spring Boot) uses the following json layout,

https://github.com/qos-ch/logback-contrib/blob/c68c2e5ddc0ed589b0f1e24bac7b9300a0ea5153/json/classic/src/main/java/ch/qos/logback/contrib/json/classic/JsonLayout.java#L105-L113

I suppose so far this is the only controversial part.

I agree in this PR, we can just keep the layout a constant with some reasonable choices of the default keys. And then wait for the feedback from the community.

I will remove them later.

lujiajing1126:
Necessary docs are added. Any other improvement is welcome.

kezhenxu94:
@lujiajing1126 so the problem now is that even you configure the format to JSON, some (at the beginning) of the logs are still pattern-based until the config is actually initialized (that the logger manger can determine which format to use), it's kind of circular dependency here

lujiajing1126:
> @lujiajing1126 so the problem now is that even you configure the format to JSON, some (at the beginning) of the logs are still pattern-based until the config is actually initialized (that the logger manger can determine which format to use), it's kind of circular dependency here

Yes, exactly. But I suppose the problem only exists in `SnifferConfigInitializer`, specifically this part,

https://github.com/apache/skywalking/blob/77d94b4d2a906cf160cfab28fd8a045ec2a2ce1d/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java#L62-L90

What about postponing log output in this class? For example, push the logs into a queue.

kezhenxu94:
> What about postponing log output in this class or globally with a so-called DelayedLogger? For example, push the logs into a queue.

That makes things too complicated unnecessarily, I was just wondering whether the partial pattern-based logs break the JSON log collector or not

lujiajing1126:
@kezhenxu94 I have tested with `logstash` collector. Plain text will not break the log processing. 

Only complains will be reported if invalid json has been detected. And this behavior can be changed if [`skin_on_invalid_json`](https://www.elastic.co/guide/en/logstash/current/plugins-filters-json.html#plugins-filters-json-skip_on_invalid_json) has been set to true.

wu-sheng:
Your codes style check are fail, are you still watching this PR?

wu-sheng:
To all reviewer, once this PR doesn't change, we have no point to cost CI resource to keep `update`

lujiajing1126:
> Your codes style check are fail, are you still watching this PR?

I suppose this is due to the newly merged commit?

lujiajing1126:
> Your codes style check are fail, are you still watching this PR?

https://github.com/apache/skywalking/pull/5357/checks?check_run_id=1040223513#step:5:3150

I don't think these errors are related to this PR

wu-sheng:
The master branch and other PRs are all checked. You are the one adding new things.

pom.xml
```diff
@@ -250,6 +250,14 @@
             <version>${jmh.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <version>${jmh.version}</version>
+            <scope>test</scope>
+        </dependency>
```
wu-sheng, line 259: You are adding this.

lujiajing1126: Oh, what's wrong with this dependency?

I didn't manage to run the micro benchmark without this module.

wu-sheng: Take a close look at the CI error log, you posted to me. You are the one taking care of new things.

I am not saying anything is wrong, I am pointing you out, your changes trigger errors.

lujiajing1126: > You are adding this.

I see what's wrong here. I am trying to resolve this.

lujiajing1126:
> The master branch and other PRs are all checked. You are the one adding new things.

The problem has been resovled.

pom.xml
lujiajing1126: Fixed

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
lujiajing1126: Fixed

pom.xml
```diff
@@ -581,6 +588,7 @@
                     </resourceExcludes>
                     <excludes>
                         **/target/generated-test-sources/**,
+                        org/apache/skywalking/apm/agent/core/logging/core/generated/LoggingBenchmark*.java,
```
wu-sheng, line 591: Could you share which part generates these classes? 

And please move this to the last one.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
```diff
@@ -62,7 +63,15 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
         final PluginFinder pluginFinder;
         try {
             SnifferConfigInitializer.initializeCoreConfig(agentArgs);
+        } catch (Exception e) {
+            LogManager.getLogger(SkyWalkingAgent.class)
+                    .error(e, "SkyWalking agent initialized failure. Shutting down.");
+            return;
+        } finally {
+            LOGGER = LogManager.getLogger(SkyWalkingAgent.class);
```
wu-sheng, line 71: Add comments for this.

pom.xml
lujiajing1126: Fixed.

These synthetic codes are generated by the "jmh-annotation-processor" according to the [document](https://openjdk.java.net/projects/code-tools/jmh/).

In the past, this dependency is not included even if there is another benchmark `LinkedArrayBenchmark`. I guess this was the trick.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
lujiajing1126: Added

pom.xml
wu-sheng: Could you highlight which parts of documents talking about this code generation?

wu-sheng: `LinkedArrayBenchmark` doesn' generate codes, as you can see from the `excludes` list, nothing about DataCarrier packages.

lujiajing1126: In the "Basic Considerations" section, it reads

> In all cases, the key to using JMH is enabling the annotation- or bytecode-processors to generate the synthetic benchmark code.

wu-sheng: > LinkedArrayBenchmark doesn' generate codes, as you can see from the excludes list, nothing about DataCarrier packages.

Could you check why this? Even now, it doesn't generate new codes. I prefer we don't need to generate anything.
Less dependency(even test scope only), less maintenance.

wu-sheng: Also, without adding this dependency, `LinkedArrayBenchmark` still is able to run in the IntelliJ, what is the difference or do I miss anything?

lujiajing1126: > Also, without adding this dependency, `LinkedArrayBenchmark` still is able to run in the IntelliJ, what is the difference or do I miss anything?

I've tried to run this benchmark, but I failed. An error is reported,

```
Exception in thread "main" java.lang.RuntimeException: ERROR: Unable to find the resource: /META-INF/BenchmarkList
	at org.openjdk.jmh.runner.AbstractResourceReader.getReaders(AbstractResourceReader.java:98)
	at org.openjdk.jmh.runner.BenchmarkList.find(BenchmarkList.java:122)
	at org.openjdk.jmh.runner.Runner.internalRun(Runner.java:263)
	at org.openjdk.jmh.runner.Runner.run(Runner.java:209)
	at org.apache.skywalking.apm.commons.datacarrier.LinkedArrayBenchmark.main(LinkedArrayBenchmark.java:176)
```

How can you run this benchmark in IDEA?

lujiajing1126: Probably you have some special setup with your IDE? Actually I found some issues related to the "jmh-generator-annprocess",

https://github.com/artyushov/idea-jmh-plugin/issues/13
https://github.com/undertow-io/undertow/pull/833

lujiajing1126: @kezhenxu94 Can you also kindly make a cross-check?

kezhenxu94: > @kezhenxu94 Can you also kindly make a cross-check?

The same error as yours

wu-sheng: But why that benchmark doesn't generate anything? That is why I am asking the question.

apm-sniffer/apm-agent-core/pom.xml
```diff
@@ -104,6 +104,12 @@
             <version>${ststem-rules.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <version>${jmh.version}</version>
+            <scope>test</scope>
+        </dependency>
```
wu-sheng, line 112: You are adding this for agent core only, so the datacarrier benchmark is still not runnable. I don't think this is the right way.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/JsonLogger.java
```diff
@@ -0,0 +1,80 @@
...
+import com.google.gson.Gson;
+import org.apache.skywalking.apm.agent.core.logging.core.converters.LiteralConverter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * An alternative logger for the SkyWalking agent. The default layout is
+ * {
+ *     "timestamp": "", // timestamp
```
wu-sheng, line 30: Could you recheck, whether we are going to use `timestamp` or `@timestamp`.

apm-sniffer/apm-agent-core/pom.xml
```diff
@@ -104,6 +104,12 @@
             <version>${ststem-rules.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <version>${jmh.version}</version>
```
wu-sheng, line 110: This is not required, right? As you have used the dependency management.

pom.xml
```diff
@@ -290,6 +291,12 @@
                 <version>${powermock.version}</version>
                 <scope>test</scope>
             </dependency>
+            <dependency>
+                <groupId>org.openjdk.jmh</groupId>
+                <artifactId>jmh-generator-annprocess</artifactId>
+                <version>${jmh.version}</version>
+                <scope>test</scope>
+            </dependency>
```
wu-sheng, line 299: Why I can't find the dependency on the DataCarrier pom.xml?

wu-sheng (commented):
You haven't updated the agent setup doc to describe how to use the new logger format.

pom.xml
lujiajing1126: > But why that benchmark doesn't generate anything? That is why I am asking the question.

I am not clear about what happened in the past. I guess it was a trick to avoid the style-check issue. But I am sure the dependency should have been declared there.

I am still curious why you can run it seamlessly?

apm-sniffer/apm-agent-core/pom.xml
lujiajing1126: Yes. Removed now

pom.xml
lujiajing1126: I am also curious

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/JsonLogger.java
lujiajing1126: A typo here. Should be `@timestamp`. Fixed

pom.xml
wu-sheng: > I am still curious why you can run it seamlessly?

I am not sure either. The dependency seems not an issue. I am just checking the inconsistent issue, two benchmarks having different exclusion configuration concerns me.

> As I understand, this LOC should help

Then why your benchmark isn't excluded?

lujiajing1126:
> You haven't updated the agent setup doc to describe how to use the new logger format.

Updated in `docs/en/setup/service-agent/java-agent/README.md`

pom.xml
wu-sheng: I mean you should, my local IDE may be set differently, but as you are adding this, we should make sure this is added in all necessary places.

lujiajing1126: > I mean you should, my local IDE may be set differently, but as you are adding this, we should make sure this is added in all necessary places.

I have added this dependency to all necessary modules to ensure those benchmarks can be run locally. 

Also two general exclude rules are introduced to exclude all jmh generated classes.

```
**/generated/*_jmhType*.java,
**/generated/*_jmhTest.java
```

wu-sheng:
```
**/generated/*_jmhType*.java,
**/generated/*_jmhTest.java
```

This is good.

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -96,7 +96,8 @@ property key | Description | Default |
 `logging.file_name`|Log file name.|`skywalking-api.log`|
 `logging.output`| Log output. Default is FILE. Use CONSOLE means output to stdout. |`FILE`|
 `logging.dir`|Log files directory. Default is blank string, means, use "{theSkywalkingAgentJarDir}/logs  " to output logs. {theSkywalkingAgentJarDir} is the directory where the skywalking agent jar file is located |`""`|
-`logging.pattern `|logging format. There are all conversion specifiers: <br>&nbsp;&nbsp;* `%level` means log level. <br>&nbsp;&nbsp;*  `%timestamp` means now of time with format `yyyy-MM-dd HH:mm:ss:SSS`.<br>&nbsp;&nbsp;*   `%thread` means name of current thread.<br>&nbsp;&nbsp;*   `%msg` means some message which user logged. <br>&nbsp;&nbsp;*  `%class` means SimpleName of TargetClass. <br>&nbsp;&nbsp;*  `%throwable` means a throwable which user called. <br>&nbsp;&nbsp;*  `%agent_name` means `agent.service_name`  |`%level %timestamp %thread %class : %msg %throwable`|
+`logging.resolver`|Logger resovler: `PATTERN` or `JSON`. Default is `PATTERN` with which `PatternLogger` will be used. `JSON`: `JsonLogger` is used after initialization. |`PATTERN`|
```
wu-sheng, line 99:
```suggestion
`logging.resolver`|Logger resolver: `PATTERN` or `JSON`. The default is `PATTERN`, which uses `logging.pattern` to print traditional text logs. `JSON` resolver prints logs in JSON format. |`PATTERN`|
```

wu-sheng (commented):
Basically LGTM. @kezhenxu94 Recheck?

Just rewrite the document, don't need to expose the codes to the end-user.

wu-sheng:
Another recommendation, could you enhance `PatternLoggerTest` and `JsonLoggerTest` to check the print texts, rather than times. Because since this PR merged, the format could have a bug in theory.

wu-sheng:
Seems your codes break the format check.

kezhenxu94 (approved):
LGTM, thanks for the patience

lujiajing1126:
> Another recommendation, could you enhance `PatternLoggerTest` and `JsonLoggerTest` to check the print texts, rather than times. Because since this PR merged, the format could have a bug in theory.

I have refactored the test codes to impose stricter tests. CI seems to be fine so far.

wu-sheng (approved):
LGTM. Thank you.

lujiajing1126:
Grazie a voi 😄

PR #5376, opened 2020-08-25
Hot reload gRPC certs of OAP.

### LINKED ISSUE #5110: gRPC SSL certs hot reloading
When using SSL to perform the AuthN between client/agent and OAP. To make things more secure, OAP should support hot reload certs.

After some investigation, I found there're some potential solutions existing in the gRPC community, for instance, https://github.com/grpc/grpc-java/issues/5335. 

I prefer to delegate SSLContext, anyone who has any thoughts, pls feel free to let me know, thanks.

hanahmily:
intends to fix #5110


wu-sheng (approved):

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/grpc/ssl/DynamicSslContext.java
```diff
@@ -0,0 +1,134 @@
...
+        updateContext(caFile);
+        monitor = new MultipleFilesChangeMonitor(
+            10,
+            readableContents -> updateContext(caFile),
+            caFile);
+    }
+
+    private void updateContext(String caFile) {
+        try {
+            GrpcSslContexts.forClient().trustManager(Paths.get(caFile).toFile()).build();
```
kezhenxu94, line 71: The returned value is just ignored? If that's intentionally, do we need a null check in the following uses of `cox`?

wu-sheng: I think this is a lambda based on `MultipleFilesChangeMonitor#FilesChangedNotifier#filesChanged`. It is triggered by the file change monitor, I think there is nothing you could do.

kezhenxu94: > I think this is a lambda based on `MultipleFilesChangeMonitor#FilesChangedNotifier#filesChanged`. It is triggered by the file change monitor, I think there is nothing you could do.

That's not what I meant, compare it with the other overloaded method `updateContext`

wu-sheng: Got it, you mean this

```suggestion
            ctx = GrpcSslContexts.forClient().trustManager(Paths.get(caFile).toFile()).build();
```

@hanahmily Please take a look.

wu-sheng (changes requested):

kezhenxu94 (approved):
LGTM

wu-sheng (approved):

PR #5389, opened 2020-08-26
Provide TraceSegmentReportService#collectInSync service as the alternative to `TraceSegmentReportService.collect`

### LINKED ISSUE #5358: Unary alternative to `TraceSegmentReportService.collect`
Why do you submit this issue?

- [x] Feature request

Can we have a unary version of https://github.com/apache/skywalking-data-collect-protocol/blob/b2f381e63702bc43216ef5576637195102302c6e/language-agent/Tracing.proto#L31? Reason is to provide a simpler client-side implementation. An additional supporting argument, see https://buf.build/docs/style-guide/#recommendations at the very end of that section. 

```proto
rpc ReportSegments(ReportSegmentsRequest) returns (ReportSegmentsResponse) {}

message ReportSegmentsRequest {
  repeated SegmentObjects segments = 1;
}

message ReportSegmentResponse {
  Commands commands = 1;   
}
```

cc. @ZackButcher

wu-sheng:
All discussion is tracked on #5358, this should resolve #5358. To all @apache/skywalking-committers, from my understanding, I still think all agents should keep using streaming as it is today.

FYI @kezhenxu94 No e2e for this, as I don't see the point of using this in native Java agent.

To @dio @lizan @ZackButcher, I am not sure whether Envoy would like to use this for real, but added this anyway.


kezhenxu94 (approved):
LG

Issue #5311, opened 2020-08-14
rocketmq plugin of agent throw NullPointException


naah69:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [x] Question or discussion
- [ ] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?
why it throwed exception and what i will do can make it work.
___
### Bug
- Which version of SkyWalking, OS and JRE?
agent 8.10 , openjdk 8 ,rocketmq4.3
- Which company or project?

- What happened?
If possible, provide a way to reproduce the error. e.g. demo application, component version.

i use sync send message in rocketmq,
it throw error and agent did't work ,don't send data to collector
```

ERROR 2020-08-14 14:47:48:750 AsyncSenderExecutor_1 InstMethodsInter : class[class org.apache.rocketmq.client.trace.AsyncTraceDispatcher$AsyncAppenderRequest$1] before method[onException] intercept failure 
java.lang.NullPointerException
        at org.apache.skywalking.apm.plugin.rocketMQ.v4.OnExceptionInterceptor.beforeMethod(OnExceptionInterceptor.java:42)
        at org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstMethodsInter.intercept(InstMethodsInter.java:76)
        at org.apache.rocketmq.client.trace.AsyncTraceDispatcher$AsyncAppenderRequest$1.onException(AsyncTraceDispatcher.java)
        at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl$4.run(DefaultMQProducerImpl.java:512)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

___
### Requirement or improvement
- Please describe your requirements or improvement suggestions.

wu-sheng:
I recommend you debug locally, and check what is going on. No plugin test about RocketMQ plugin yet. If could build one by following plugin development doc.

xbkaishui:
please assign me

xbkaishui:
[...]

PR #5412, opened 2020-08-29
#5311 Fix ActiveMQ NullPointerException

xbkaishui:

- [*] Bug fix

- Related issues
 #5311
___
### Bug fix
- Bug description.
when use new version client(has metric collect feature) to connect to server , the plugin throws null pointer exception
- How to fix?
check the enhanceInfo null condition


apm-sniffer/apm-sdk-plugin/rocketMQ-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rocketMQ/v4/OnExceptionInterceptor.java
```diff
@@ -34,15 +34,22 @@
...
     public static final String CALLBACK_OPERATION_NAME_PREFIX = "RocketMQ/";
+    private static final String DEFAULT_TOPIC = "no_topic";
 
     @Override
     public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
         MethodInterceptResult result) throws Throwable {
         SendCallBackEnhanceInfo enhanceInfo = (SendCallBackEnhanceInfo) objInst.getSkyWalkingDynamicField();
-        AbstractSpan activeSpan = ContextManager.createLocalSpan(CALLBACK_OPERATION_NAME_PREFIX + enhanceInfo.getTopicId() + "/Producer/Callback");
+        String topicId = DEFAULT_TOPIC;
+        if (enhanceInfo != null) {
```
wu-sheng, line 44:
```suggestion
        // The SendCallBackEnhanceInfo could be null when there is an internal exception in the client API,
        // such as MQClientException("no route info of this topic")
        if (enhanceInfo != null) {
```

xbkaishui: done

wu-sheng (approved):

nisiyong:
BTW, this PR is about RocketMQ rather than ActiveMQ, I think the title should be changed.

PR #5425, opened 2020-09-01
Introduce label system to histogram

### LINKED ISSUE #5340: Introduce label system to histogram and percentile of meter-system
We need to handle histogram with  multiple labels just like below:

```
# HELP rpc_duration_seconds A summary of the RPC duration in seconds.
# TYPE rpc_duration_seconds summary
rpc_duration_seconds{quantile="0.01" status="200"} 3102
rpc_duration_seconds{quantile="0.05" status="200"} 3272
rpc_duration_seconds{quantile="0.5" status="200"} 4773
rpc_duration_seconds{quantile="0.9" status="200"} 9001
rpc_duration_seconds{quantile="0.99" status="200"} 76656
rpc_duration_seconds_sum 1.7560473e+07
rpc_duration_seconds_count 2693
rpc_duration_seconds{quantile="0.01" status="400"} 0
rpc_duration_seconds{quantile="0.05" status="400"} 12
rpc_duration_seconds{quantile="0.5" status="400"} 12
rpc_duration_seconds{quantile="0.9" status="400"} 14
rpc_duration_seconds{quantile="0.99" status="400"} 14
rpc_duration_seconds_sum 15
rpc_duration_seconds_count 220
```

hanahmily:
Fix #5340 

As below image shows that there are three groups and each one contains two percentile, 95%, and 99%
image

In this PR, the bucket unit of the histogram can be set through the metric rule. Since Prometheus official documents confirm the unit of a bucket should be Seconds, but other libraries, for example, open census, can set a different unit.


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/AvgHistogramFunction.java
```diff
@@ -83,33 +84,39 @@ public void accept(final MeterEntity entity, final BucketedValues value) {
...
+            summation.valueAccumulation(key, values[i]);
+            count.valueAccumulation(key, 1L);
         }
     }
 
     @Override
     public void combine(final Metrics metrics) {
         AvgHistogramFunction histogram = (AvgHistogramFunction) metrics;
-
-        if (!summation.keysEqual(histogram.getSummation())) {
```
wu-sheng, line 98: Do you remove this for accepting different labels in 2 instances?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/AvgHistogramFunction.java
```diff
@@ -83,33 +84,39 @@ public void accept(final MeterEntity entity, final BucketedValues value) {
...
         this.count.append(histogram.count);
     }
 
     @Override
     public void calculate() {
-        final List<String> sortedKeys = summation.sortedKeys(Comparator.comparingInt(Integer::parseInt));
-        for (String key : sortedKeys) {
-            dataset.put(key, summation.get(key) / count.get(key));
+        final Set<String> keys = summation.keys();
+        for (String key : keys) {
```
wu-sheng, line 111:
```suggestion
        for (String key : summation.keys()) {
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/AvgHistogramFunction.java
```diff
@@ -157,6 +165,7 @@ public void deserialize(final RemoteData remoteData) {
 
         remoteBuilder.addDataObjectStrings(count.toStorageData());
         remoteBuilder.addDataObjectStrings(summation.toStorageData());
+        remoteBuilder.addDataObjectStrings(dataset.toStorageData());
```
wu-sheng, line 168: Do you forget to serialize this field?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/AvgHistogramPercentileFunction.java
```diff
@@ -125,11 +135,17 @@ public void accept(final MeterEntity entity, final AvgPercentileArgument value)
 
         this.entityId = entity.id();
 
+        String template = "%s";
+        if (!Strings.isNullOrEmpty(value.getBucketedValues().getGroup())) {
+            template   = value.getBucketedValues().getGroup() + ":%s";
```
wu-sheng, line 140:
```suggestion
            template  = value.getBucketedValues().getGroup() + ":%s";
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/AvgHistogramPercentileFunction.java
```diff
@@ -125,11 +135,17 @@ public void accept(final MeterEntity entity, final AvgPercentileArgument value)
...
+        }
         final long[] values = value.getBucketedValues().getValues();
         for (int i = 0; i < values.length; i++) {
-            String bucketName = String.valueOf(value.getBucketedValues().getBuckets()[i]);
-            summation.valueAccumulation(bucketName, values[i]);
-            count.valueAccumulation(bucketName, 1L);
+            long bucket = value.getBucketedValues().getBuckets()[i];
+            String bucketName = bucket == Long.MIN_VALUE ? Bucket.INFINITE_NEGATIVE : String.valueOf(bucket);
+            String key = String.format(template, bucketName);
+            summation.valueAccumulation(key, values[i]);
```
wu-sheng, line 147: Could you add comments about how you format this key?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/AvgHistogramPercentileFunction.java
```diff
@@ -168,38 +178,64 @@ public void combine(final Metrics metrics) {
...
+
+                        count += value;
+                        for (int rankIdx = loopIndex; rankIdx < roofs.length; rankIdx++) {
+                            int roof = roofs[rankIdx];
+
+                            if (count >= roof) {
+                                if (group.equals(DEFAULT_GROUP)) {
+                                    percentileValues.put(String.valueOf(ranks.get(rankIdx)), Long.parseLong(key));
+                                } else {
+                                    percentileValues.put(String.format("%s:%s", group, ranks.get(rankIdx)), Long.parseLong(key));
```
wu-sheng, line 230: So, according to this, the `name:99` is formated from here, right?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/AvgHistogramFunction.java
hanahmily: exactly.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/AvgHistogramFunction.java
hanahmily: yep. With the unit test's help, I found this issue.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/AvgHistogramPercentileFunction.java
hanahmily: It's "<group>:<percentile>" which is composed of some lines above.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/AvgHistogramPercentileFunction.java
hanahmily: Exactly. `:` is the splitter.

wu-sheng (approved):

wu-sheng (changes requested):
All e2e fail, so most likely the codes are not right.

hanahmily:
> All e2e fail, so most likely the codes are not right.

seems file changed monitor test failed:
```
2020-09-01T22:35:06.2945140Z [ERROR] Failures: 
2020-09-01T22:35:06.2945330Z [ERROR]   MultipleFilesChangeMonitorTest.test:81
```

I encountered similar issues several times. might we not create tmp files in e2e.

wu-sheng:
I don't remember we created tmp file in the e2e. You are just posting the CI tests.

kezhenxu94:
> 2020-09-01T22:35:06.2945140Z [ERROR] Failures: 
> 2020-09-01T22:35:06.2945330Z [ERROR]   MultipleFilesChangeMonitorTest.test:81
> I encountered similar issues several times. might we not create tmp files in e2e.

E2E tests never ever run the unit tests, thus won't cause the failure `MultipleFilesChangeMonitorTest.test:81`, and the E2E failed because of timeout (now)

wu-sheng (approved):

PR #5416, opened 2020-08-31
Support segment status based on entry/first span only

EvanLjp:
Please answer these questions before submitting a pull request

- Why submit this pull request?

- [ ] New feature provided

- Related issues
https://github.com/apache/skywalking/issues/4980

___
### New feature or improvement
- Describe the details and related test reports.
Support segment status based on entry/first span only

PR #5426, opened 2020-09-01
exception-ignore-plugin

EvanLjp:

- [ ] Bug fix
- [ ] New feature provided

- Related issues
https://github.com/apache/skywalking/pull/5416

___
now logic is only adjust status when throw ignore exception,
first we define 2 kinds of exception. IgnoreException and OtherException

if response code is 500 and don't throw exception ,the final status is error.
if response code is 500 and throw OtherException,the final status is errror.
if response code is 500 and throw IgnoreException,the final status is not error.
for example in spring mvc, maybe throw exception would be cause response code is 500, if the exception could be ignored,the span final staus would be not error


EvanLjp:
all manual test case is pass:
1. test exception-ignore-plugin
The plugin is designed to ignore business exception which definition is the exception could be ignore in the span, the demand is very common in our department.The feature would be ignore some unnecessary alarm.
ignore some exception in span
image
image

wu-sheng (commented):
Some basic guidelines for this feature.
___
The exception exclusion mechanism should be implemented on the exception-catch stage directly, rather than trying to put some black magic codes in the core. 
___
Such as `AbstractTracingSpan#log(Throwable t)` could trigger `errorOccurred` automatically first, or don't set the error status if the `StatusTriggerService`(new) determine the processing logic.

Which exceptions should be ignored, this should base on the class name configuration. NOTE, how to check is not easy.
1. The user exceptions could be a hierarchy relationship. 
1. The exception could be wrapped in another exception, in another exception, whether to check `cause` should be configured, the default should be NO for performance concern.
1. If you want to config the exception name, the class type can't be initialized in the bootstrap, it only could be known when the exception happened.

That is why this feature is still not provided, you need to consider many details to make this works :) But this is a good feature, and will help to understand more about the agent core.

apm-sniffer/optional-plugins/exception-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/exception/ignore/ExceptionIgnorerService.java
```diff
@@ -0,0 +1,92 @@
...
+    private final Set<String> exceptionClassNames = new HashSet<>();
+
+    @Override
+    public void prepare() throws Throwable {
+        if (StringUtil.isNotEmpty(IGNORED_EXCEPTION_CLASS_NAMES)) {
+            exceptionClassNames.addAll(Arrays.asList(IGNORED_EXCEPTION_CLASS_NAMES.split(",")));
+        }
+        if (StringUtil.isNotEmpty(IGNORED_EXCEPTION_SCAN_SCOPE)) {
+            ClassPath classpath = ClassPath.from(PluginBootstrap.class.getClassLoader());
+            ImmutableSet<ClassPath.ClassInfo> classes = classpath.getTopLevelClasses(IGNORED_EXCEPTION_SCAN_SCOPE);
```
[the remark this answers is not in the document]
EvanLjp: I've done this before in self apm system,there is some thinking about this feature.

1. the package classes scan is only for the exceptions tagged with @IgnoredException in  user code,so we don't need to scan all classes.PluginBootstrap.class.getClassLoader() have been meet the demand.
2. If some exception defined in other lib which out of user control, the exceptions cannot be tagged be @IgnoredException,so package scan scope is no sense to it. the kind of exception should be ignored by hard code config.

In my mind , user would be like to add a new  ignored exception by @IgnoredException annotation rather than change skywalking agent config frequency.

How do you think about it? @wu-sheng

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/AbstractTracingSpan.java
```diff
@@ -96,6 +99,8 @@
      */
     protected boolean skipAnalysis;
 
+    protected String exceptionKind = "";
```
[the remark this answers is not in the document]
EvanLjp: 1. in local span only may be 1 exception at most
2. many exceptions log may be only entry span or exist span. 
(entry1)-->(entry2)-->(local3)-->(exit4)-->(exit5)
the exceptions in entry2,entry3,entry4 should be thought as exception Kind.
the exceptions in entry1,exit5 should not be thought as exception kind.
how do u think about it?

apm-sniffer/optional-plugins/exception-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/exception/ignore/ExceptionIgnorerService.java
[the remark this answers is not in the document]
kezhenxu94: > In my mind , user would be like to add a new ignored exception by @IgnoredException annotation rather than change skywalking agent config frequency.

I don't agree, for those exceptions thrown by some frameworks, users cannot do anything to ignore them but only change the agent config, and I suppose that's a common case where users want to ignore exceptions from frameworks, if they have control to where the exceptions are thrown, they can choose not to throw the exceptions and only log some errors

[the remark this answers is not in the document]
EvanLjp: > I don't agree, for those exceptions thrown by some frameworks, users cannot do anything to ignore them but only change the agent config

yes, that is the 2 point what i thought. 
> 2\. If some exception defined in other lib which out of user control, the exceptions cannot be tagged be @IgnoredException,so package scan scope is no sense to it. the kind of exception should be ignored by hard code config.

my option is @IgnoredException is more friendly to use in user service code.





> 1. the package classes scan is only for the exceptions tagged with @IgnoredException in  user code,so we don't need to scan all classes.PluginBootstrap.class.getClassLoader() have been meet the demand.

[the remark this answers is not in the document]
wu-sheng: > the package classes scan is only for the exceptions tagged with @IgnoredException in user code,so we don't need to scan all classes.PluginBootstrap.class.getClassLoader() have been meet the demand.

That would be the case. Once you open `@IgnoredException` you can't tell where it will be used. `PluginBootstrap.class.getClassLoader()` wouldn't be able to access many classes when you use the standard J2EE container. Scan tech is not allowed in the agent core stack or plugin stack.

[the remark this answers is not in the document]
wu-sheng: `User code` is the wrong statement. All codes are user codes from the agent's perspective. And the user could have a class loader, the framework could have a class loader.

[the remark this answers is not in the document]
EvanLjp: this would be cause so many problems.

1.  PluginBootstrap.class.getClassLoader() is application classloader, cannot access custom classloder classes.
2. So we cannot to solve it but for to judge it at run time,for excample have following steps.
     2.1 find classes in every call by the hard config
     2.2 check the status
I don't know if this has a performance impact when find classes instance every time

@wu-sheng @kezhenxu94 
Do you have any ideas to implement this feature?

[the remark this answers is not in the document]
wu-sheng: That is why, I asked you don't do the scan for class match. You should determine the flow after exception happens and logs.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/AbstractTracingSpan.java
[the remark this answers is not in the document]
wu-sheng: Whether one or mutiple error logs are based on plugin, not core. They didn't say that API is allowed to be called max once per span in any document

apm-sniffer/optional-plugins/exception-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/exception/ignore/ExceptionIgnorerService.java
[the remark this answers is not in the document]
kezhenxu94: > 1. PluginBootstrap.class.getClassLoader() is application classloader, cannot access custom classloder classes.

If you check the exception only when it's thrown(in the `Span#log(Throwable e)`), you can simply use the same class loader to load the configured class by name

> 2. So we cannot to solve it but for to judge it at run time,for excample have following steps.
>    2.1 find classes in every call by the hard config
>    2.2 check the status
>    I don't know if this has a performance impact when find classes instance every time

Some simple caching is enough, I don't think the `Ignored Exception` list would be very long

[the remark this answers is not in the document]
EvanLjp: it's almost similar to a custom implement skywalking plugin in our company, i would push it laster,thx

> Some simple caching should be enough, I don't think the `Ignored Exception` list would be very long

EvanLjp:
@wu-sheng @kezhenxu94 thx for check this version, the version may be fix all question in past version.

If the idea works,
i would like to refactor some plugins
brefore:
after:
ContextManager.activeSpan().log(t);

if in some plugins only use errorOccurred method , this one will not be removed.for example in spring mvc plugin

wu-sheng:
> if in some plugins only use errorOccurred method , this one will not be removed.for example in spring mvc plugin

If error status based on other things(HTTP response code), then it should keep what it is today.

apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/IgnoredException.java
```diff
@@ -0,0 +1,34 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * the annotation is declared on business exception class
+ **/
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface IgnoredException {
```
wu-sheng, line 33: Why still use `annotation`?

apm-sniffer/apm-agent-core/pom.xml
```diff
@@ -53,6 +53,11 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>org.apache.skywalking</groupId>
+            <artifactId>apm-toolkit-trace</artifactId>
+            <version>8.2.0-SNAPSHOT</version>
+        </dependency>
```
wu-sheng, line 60: This depedency is not allowed, related to you are still using annotation.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/DefaultExceptionIgnorerService.java
```diff
@@ -0,0 +1,52 @@
...
+
+    }
+
+    @Override
+    public void shutdown() throws Throwable {
+
+    }
+
+    @Override
+    public AbstractTracingSpan ignore(final AbstractTracingSpan span) {
```
wu-sheng, line 49: Should this be removed?

EvanLjp:
yes，agree with u

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/AnnotationMatchExceptionCheckStrategy.java
```diff
@@ -0,0 +1,46 @@
...
+    public boolean isError(final Throwable e) {
+        if (ignoredExceptions.contains(e.getClass())) {
+            return false;
+        }
+        StatusCheckService statusTriggerService = ServiceManager.INSTANCE.findService(StatusCheckService.class);
+        Class<? extends Annotation> ignoredExceptionAnnotationClass = statusTriggerService.ignoredExceptionAnnotationClass;
+        if (Objects.isNull(ignoredExceptionAnnotationClass)) {
+            return true;
+        }
+        if (e.getClass().isAnnotationPresent(ignoredExceptionAnnotationClass)) {
```
wu-sheng, line 40: I still don't like the idea use the annotation. 
1. The toolkit classes are in the user class loader zone, you can't assume it is usable in the agent core.
1. Why simply use the exception name string match? This is much safer and easier than this.
1. Like (2), the exception class could be cached.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/AnnotationMatchExceptionCheckStrategy.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private final HashSet<Class<? extends Throwable>> ignoredExceptions = new HashSet<>(16);
+
+    @Override
+    public boolean isError(final Throwable e) {
+        if (ignoredExceptions.contains(e.getClass())) {
+            return false;
+        }
+        StatusCheckService statusTriggerService = ServiceManager.INSTANCE.findService(StatusCheckService.class);
+        Class<? extends Annotation> ignoredExceptionAnnotationClass = statusTriggerService.ignoredExceptionAnnotationClass;
```
wu-sheng, line 36: Accessing the field directly is not allowed.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/FastFailExceptionCheckStrategy.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.agent.core.context.status;
+
+public class FastFailExceptionCheckStrategy implements ExceptionCheckStrategy {
+    @Override
+    public boolean isError(final Throwable e) {
+        return !(e instanceof Exception);
```
wu-sheng, line 24: What is this? Why are you doing this?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/DefaultExceptionIgnorerService.java
EvanLjp: yes

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/FastFailExceptionCheckStrategy.java
EvanLjp: 1. in throwable, there are 2parts error or exception.

error occurs mostly occurs when system error
exception mostly occurs  when  application error

so we should only allow adjust exception status

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/AnnotationMatchExceptionCheckStrategy.java
EvanLjp: package access

wu-sheng: The same thing, all accesses should be through methods unless it is in the same class and children. But this is not the key, please check comments about the `annotation`, which is my primary concern.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/AnnotationMatchExceptionCheckStrategy.java
EvanLjp: > 1. The toolkit classes are in the user class loader zone, you can't assume it is usable in the agent core.
1. load the ignoredExceptionAnnotationClass by application classloader, it is already passed in the manual test.


> 2\. Why simply use the exception name string match? This is much safer and easier than this.


If i was a user,i would to add ignore exception by annotation,this would be easy to control rather than change env config frequency. in our department,skywalking is throught as a platform, env is a only way to custom config.But env config is out of user code config, user would not like to frequency config.Env should keep stable.






> 3\. Like (2), the exception class could be cached.

private final HashSet<Class<? extends Throwable>> ignoredExceptions = new HashSet<>(16);
already cache ignored exceptions.
we should add a new cache to store unignored exceptions.Is that what you're saying?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/DefaultExceptionIgnorerService.java
EvanLjp: i would remove the past version codes if the idea of this version is passed

apm-sniffer/apm-agent-core/pom.xml
EvanLjp: If i was a user,i would to add ignore exception by annotation,this would be easy to control rather than change env config frequency. in our department,skywalking is throught as a platform, env is a only way to custom config.But env config is out of user code config, user would not like to frequency config.Env should keep stable.

that is why i want to keep annotation . Could i extract a common light sdk for tookit and agent together?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/AnnotationMatchExceptionCheckStrategy.java
EvanLjp: > The same thing, all accesses should be through methods unless it is in the same class and children. But this is

already reply

kezhenxu94 (commented):
Seems we haven't reached to a consensus yet, @wu-sheng and I expect that there is no annotation at all, but that's not the case in this version. Here is what I've been thinking about why there is no need to have an annotation

> If i was a user,i would to add ignore exception by annotation,this would be easy to control rather than change env config frequency.

Things you can do with annotation can be done with the config all the same, for example you have a lot of `Exception`s (`FooException`, `BarException`) that are annotated with `IgnoredException`, you can create a root `Exception`, let's say `IgnoredException`(which is a class, not an annotation), and make your `{Foo,Bar}Exception` extend from `IgnoredException`, and only configuring the `IgnoredException` should be enough.

P.S. inheritance can be fragile sometimes, but for `Exception`, I think it's fine, and since we have mentioned that it should also check the wrapped exception (`cause`), users can also wrap any exception into the `IgnoredException` without inheritance)

EvanLjp:
> Things you can do with annotation can be done with the config all the same, for example you have a lot of `Exception`s (`FooException`, `BarException`) that are annotated with `IgnoredException`, you can create a root `Exception`, let's say `IgnoredException`(which is a class, not an annotation), and make your `{Foo,Bar}Exception` extend from `IgnoredException`, and only configuring the `IgnoredException` should be enough.

okay,i got what u thought.u are right,i will remove annotation later.
@wu-sheng @kezhenxu94

EvanLjp:
If the idea of v3 version works, would be add doc and refactor other plugin later.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/InheriteMatchExceptionCheckStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+package org.apache.skywalking.apm.agent.core.context.status;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import org.apache.skywalking.apm.agent.core.boot.ServiceManager;
+
+public class InheriteMatchExceptionCheckStrategy implements ExceptionCheckStrategy {
```
wu-sheng, line 26: `Inherite` -> `hierarchy`

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/StatusChecker.java
```diff
@@ -0,0 +1,91 @@
...
+     * ignoreExceptions, return normal status.<br/> 3. The exception superclass name is exist in ignoreExceptions,
+     * return normal status.<br/>
+     *
+     * @see Config.StatusCheck#IGNORED_EXCEPTIONS
+     */
+    INHERIT_NAMED_MATCH(
+        Collections.singletonList(new FastFailExceptionCheckStrategy()),
+        Arrays.asList(
+            new NamedMatchExceptionCheckStrategy(),
+            new InheriteMatchExceptionCheckStrategy()
```
wu-sheng, line 57: If I read the codes right, InheriteMatchExceptionCheckStrategy should have included the logic of `NamedMatchExceptionCheckStrategy`.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/StatusChecker.java
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * If all strategy check the status is error,the error status is true.
+     */
+    private final List<ExceptionCheckStrategy> logicAndStrategies;
+
+    StatusChecker(final List<ExceptionCheckStrategy> logicOrStrategies,
+                  final List<ExceptionCheckStrategy> logicAndStrategies) {
+        this.logicOrStrategies = logicOrStrategies;
+        this.logicAndStrategies = logicAndStrategies;
+    }
```
wu-sheng, line 74: Do we really this complex constructor? From the coding perspective, this should be an `And ExceptionCheckStrategy(s1, s2)`, but I doubt whether this is really necessary for now.

kezhenxu94 (commented):
The idea looks good generally, some review comments are inline

wu-sheng (commented):
Basically, this time design works for me. @kezhenxu94 WDYT?

But, as I commented, this PR would need more polish on details.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/InheriteMatchExceptionCheckStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+    List<String> getSuperclasses(Class<? extends Throwable> clazz) {
+        List<String> superClasses = new ArrayList<>();
+        superClasses.add(clazz.getName());
+        Class<?> superclass = clazz.getSuperclass();
+        while (superclass != null && !superclass.equals(Exception.class)) {
+            superClasses.add(superclass.getName());
+            superclass = superclass.getSuperclass();
+        }
+        return superClasses;
+    }
```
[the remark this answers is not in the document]
EvanLjp: i know ，but i want to ignore classloader affect. we don't know how to load config class by which classloader.so I only judged it by class name.because of only little class in loop,it would not cost much time.

apm-sniffer/apm-agent-core/pom.xml
wu-sheng: If you want to use annotation, consider using toolkit + activation mode. My point is, the core should not depend on the toolkit. It is not safe in the class loader manipulation.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/AnnotationMatchExceptionCheckStrategy.java
wu-sheng: > load the ignoredExceptionAnnotationClass by application classloader, it is already passed in the manual test.

`ignoredExceptionAnnotationClass by application classloader`, it could not be.

> private final HashSet<Class<? extends Throwable>> ignoredExceptions = new HashSet<>(16);
already cache ignored exceptions.
we should add a new cache to store unignored exceptions.Is that what you're saying?

We could use a Map directly.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/InheriteMatchExceptionCheckStrategy.java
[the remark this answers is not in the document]
wu-sheng: If you want this, use this way
1. c = e.getClass().getClassLoader()
1. If c.forName(exception Class) has classNotFound, then should not ignore
1. If c.forName(exception Class) has a class reference, you could do `isAssignableFrom`

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/AnnotationMatchExceptionCheckStrategy.java
wu-sheng: Also, the key is not right, you need to use `classLoaderInstance + Exception class` as the key. The same class could be loaded by different class loaders with the same name.

wu-sheng:
I replied inline a lot. Generally, please understand, the load loader is a very tricky thing. When we are going to use it, and the agent core is manipulating it too, we need to be very careful. We need to value all codes in theory, not just through a local test. Also, plugin tests would be required in the Spring app and Tomcat app mode to simulate widely used JRE class loader cases.

wu-sheng:
@candyleer @ascrutae I hope you two could join the review, this PR related to many class loader cases. We need to be very carefully.

EvanLjp:
@wu-sheng  @kezhenxu94  thx for review again.I have listened to wusheng's advice. Support annotation tag with a new way rather than judge by classloader on runtime.

EvanLjp:
If no problem,would refactor other plugins later.

wu-sheng:
> thx for review again.I have listened to wusheng's advice. Support annotation tag with a new way rather than judge by classloader on runtime.

I have left comments about how to use the class loader appropriately.

> If no problem,would refactor other plugins later.

Why do many plugin tests fail? 

___
My concerns are still there. First of all, annotation in the toolkit is acceptable, but you are using it in the wrong way. The toolkit is only a part of the plugin system, you can't depend on it in this core, no matter in any case. Because this will cause class not found when the core is running.

EvanLjp:
> My concerns are still there. First of all, annotation in the toolkit is acceptable, but you are using it in the wrong way. The toolkit is only a part of the plugin system, you can't depend on it in this core, no matter in any case. Because this will cause class not found when the core is running.

first,core is not depends on tookit class.
when agent runing, will inject some new method for the exception tagged with @IgnoredException.
so only check exception status by finding tag method ,i don;t know why it would cause class not find

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/HierarchyMatchExceptionCheckStrategy.java
```diff
@@ -0,0 +1,51 @@
...
+
+package org.apache.skywalking.apm.agent.core.context.status;
+
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+import org.apache.skywalking.apm.agent.core.boot.ServiceManager;
+
+public class HierarchyMatchExceptionCheckStrategy implements ExceptionCheckStrategy {
+
+    private final Set<Class<? extends Throwable>> ignoredExceptions = new CopyOnWriteArraySet<>();
```
wu-sheng, line 27: You don't cache the checked class, so basically, most exceptions have to be checked every time, which is a performance issue.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/AnnotationMatchExceptionCheckStrategy.java
```diff
@@ -0,0 +1,44 @@
...
+    private final Set<Class<? extends Throwable>> ignoredExceptions = new CopyOnWriteArraySet<>();
+
+    @Override
+    public boolean isError(final Throwable e) {
+        Class<? extends Throwable> clazz = e.getClass();
+        if (ignoredExceptions.contains(clazz)) {
+            return false;
+        }
+        try {
+            String value = (String) clazz.getMethod("getSkyWalkingDynamicField").invoke(e);
```
wu-sheng, line 35: This could do `instanceof EnhancedInstance`, should not use reflect.

wu-sheng (commented):
> so only check exception status by finding tag method ,i don;t know why it would cause class not find

Have rechecked your update, this would not be an issue. Comment more. Basically this time should be better.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -268,6 +267,19 @@
...
     }
 
+    public static class StatusCheck {
+        /**
+         * Ignored exception list, also affect their subclasses.
+         */
+        public static String IGNORED_EXCEPTIONS = "";
+
+        /**
+         * Agent would do hierarchy check for the exception unless the max_recursive_depth equals to 0.And -1 means no limit.
```
wu-sheng, line 277: Don't provide no limitation. That could cause OOM

wu-sheng: All negative value should be considered as 0.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/AnnotationMatchExceptionCheckStrategy.java
```diff
@@ -0,0 +1,48 @@
...
+package org.apache.skywalking.apm.agent.core.context.status;
+
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+
+public class AnnotationMatchExceptionCheckStrategy implements ExceptionCheckStrategy {
+
+    private final Set<Class<? extends Throwable>> ignoredExceptions = new CopyOnWriteArraySet<>();
+    private final Set<Class<? extends Throwable>> exceptions = new CopyOnWriteArraySet<>();
```
wu-sheng, line 28:
```suggestion
    private final Set<Class<? extends Throwable>> errorStatusExceptions = new CopyOnWriteArraySet<>();
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/AnnotationMatchExceptionCheckStrategy.java
```diff
@@ -0,0 +1,48 @@
...
+
+package org.apache.skywalking.apm.agent.core.context.status;
+
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+
+public class AnnotationMatchExceptionCheckStrategy implements ExceptionCheckStrategy {
+
+    private final Set<Class<? extends Throwable>> ignoredExceptions = new CopyOnWriteArraySet<>();
```
wu-sheng, line 27: @kezhenxu94 @EvanLjp Any of you know the performance difference between `CopyOnWriteArraySet` and `ConcurrentHashMap`? This check is highly performance sensitive.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/AnnotationMatchExceptionCheckStrategy.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public boolean isError(final Throwable e) {
+        Class<? extends Throwable> clazz = e.getClass();
+        if (ignoredExceptions.contains(clazz)) {
+            return false;
+        }
+        if (exceptions.contains(clazz)) {
+            return true;
+        }
+        if (e instanceof EnhancedInstance && ((EnhancedInstance) e).getSkyWalkingDynamicField().equals(TAG_NAME)) {
```
wu-sheng, line 40: Use `TAG_NAME.equals...` to avoid NPE.

wu-sheng (commented):
Many comments are missing, and integration plugin tests(use annotation or class name config).

@kezhenxu94 @candyleer @ascrutae What do you think about this solution?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
EvanLjp: ok

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/AnnotationMatchExceptionCheckStrategy.java
EvanLjp: Read some doc , and replace it  with the  set view of concurrentHashMap to store exception class.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -268,6 +267,21 @@
         public static String PATTERN = "%level %timestamp %thread %class : %msg %throwable";
     }
 
+    public static class StatusCheck {
+        /**
+         * Ignored exception list, also affect their subclasses.
```
wu-sheng, line 272:
```suggestion
         * Listed exceptions would not be treated as an error. Because in some codes, the exception is being used as a way of controlling business flow.
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -268,6 +267,21 @@
...
+         */
+        public static String IGNORED_EXCEPTIONS = "";
+
+        /**
+         * The max depth for recursive check exception status. Default 0 means all exceptions will be thought of as
+         * error status. If an exception is listed in org.apache.skywalking.apm.agent.core.conf.Config.StatusCheck#IGNORED_EXCEPTIONS
+         * or tagged with org.apache.skywalking.apm.toolkit.trace.IgnoredException, the exception will not be thought of
+         * as error status, also affects its subclasses.
```
wu-sheng, line 280:
```suggestion
         * The max recursive depth when checking the exception traced by the agent. Typically, we don't recommend setting this more than 10, which could cause a performance issue. Negative value and 0 would be ignored, which means all exceptions would make the span tagged in error status.
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -268,6 +267,21 @@
...
+         */
+        public static Integer MAX_RECURSIVE_DEPTH = 0;
```
wu-sheng, line 282:
```suggestion
        public static Integer MAX_RECURSIVE_DEPTH = 1;
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/AnnotationMatchExceptionCheckStrategy.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+package org.apache.skywalking.apm.agent.core.context.status;
+
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+
+/**
+ * The exception tagged with org.apache.skywalking.apm.toolkit.trace.IgnoredException will not be thought as an error,
```
wu-sheng, line 26:
```suggestion
 * The exception annotated with org.apache.skywalking.apm.toolkit.trace.IgnoredException would not set the span in the error status.
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/HierarchyMatchExceptionCheckStrategy.java
```diff
@@ -0,0 +1,58 @@
...
+
+package org.apache.skywalking.apm.agent.core.context.status;
+
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import org.apache.skywalking.apm.agent.core.boot.ServiceManager;
+
+/**
+ * The exception listed in org.apache.skywalking.apm.agent.core.conf.Config.StatusCheck#IGNORED_EXCEPTIONS will not be
+ * thought as an error, also affects its subclasses.
```
wu-sheng, line 27:
```suggestion
 * HierarchyMatchExceptionCheckStrategy does a hierarchy check for a traced exception. If it or its parent has been listed in org.apache.skywalking.apm.agent.core.conf.Config.StatusCheck#IGNORED_EXCEPTIONS, the error status of the span wouldn't be changed.
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/OffExceptionCheckStrategy.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.agent.core.context.status;
+
+/**
+ * All exceptions will be thought of as error status.
```
wu-sheng, line 22:
```suggestion
 * All exceptions would make the span tagged as the error status.
```

wu-sheng (commented):
Two key in the comments.
1. `will` should not be used. In the code processing, there is no such thing as `will`.
1. `thought` should not be used. This is code, no human.

Please refer the review comments and adjust other comments. I don't change all of them.

EvanLjp:
thx for your suggestion,  I'll pay more attention to comments next time

apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/IgnoredException.java
```diff
@@ -0,0 +1,36 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * When opening the exception status checker in the agent, a traced exception wouldn't be marked as error status which
+ * tagged with this annotation in a span. Besides the annotation supports integration, the subclasses would be also
+ * annotated with it.
```
wu-sheng, line 30:
```suggestion
 * After the exception status checker activated in the agent, the span wouldn't be marked as error status if the exception
 * has this annotation. 
```

> Besides the annotation supports integration, the subclasses would be also annotated with it.

I think subclass status is determined by the `MAX_RECURSIVE_DEPTH`, isn't it?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/status/StatusCheckService.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.skywalking.apm.agent.core.boot.BootService;
+import org.apache.skywalking.apm.agent.core.boot.DefaultImplementor;
+import org.apache.skywalking.apm.agent.core.conf.Config;
+import org.apache.skywalking.apm.util.StringUtil;
+
+import static org.apache.skywalking.apm.agent.core.context.status.StatusChecker.HIERARCHY_MATCH;
+import static org.apache.skywalking.apm.agent.core.context.status.StatusChecker.OFF;
+
+/**
+ * The <code>StatusCheckService</code> checks the status of a exception traced in a span.
```
wu-sheng, line 32:
```suggestion
 * The <code>StatusCheckService</code> determines whether the span should be tagged in error status if an exception captured in the scope.
```

wu-sheng:
@kezhenxu94 Please recheck, I thin basically this is good.

apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/IgnoredException.java
EvanLjp: Yes, 2 ways are all controlled by MAX_RECURSIVE_DEPTH . They are almost equivalent, and both support inheritance.

org.apache.skywalking.apm.toolkit.trace.IgnoredException
org.apache.skywalking.apm.agent.core.conf.Config.StatusCheck#IGNORED_EXCEPTIONS

wu-sheng:
Twitter about this feature, https://twitter.com/ASFSkyWalking/status/1301818654661595136

apm-application-toolkit/apm-toolkit-trace/src/main/java/org/apache/skywalking/apm/toolkit/trace/IgnoredException.java
wu-sheng: Then, please correct the comment.

wu-sheng:
Also, you should add plugin tests for this feature, one on Spring, the other on Tomcat, through 2 ways of configuration.

EvanLjp:
> Also, you should add plugin tests for this feature, one on Spring, the other on Tomcat, through 2 ways of configuration.

will do it on weekend

kezhenxu94 (commented):
Generally good to me, once the tests are provided, my approval

EvanLjp:
@wu-sheng When I add a spring plugin test, I generate a module named exception checker spring. And the default package name starts with  org.apache.skywalking.apm.testcase. But when I run the manual test case, no spring span will be created.
So I read the agent source code, found the following codes:
```
private static ElementMatcher.Junction<NamedElement> allSkyWalkingAgentExcludeToolkit() {
        return nameStartsWith("org.apache.skywalking.").and(not(nameStartsWith("org.apache.skywalking.apm.toolkit.")));
    }
```
Wheather to add a new package name named "org.apache.skywalking.apm.testcase" here , or has another way to solve it.

wu-sheng:
> Wheather to add a new package name named "org.apache.skywalking.apm.testcase" here , or has another way to solve it.

Take a look at this section, https://github.com/apache/skywalking/blob/master/docs/en/guides/Plugin-test.md#test-project-hierarchical-structure. We provide a recommended package name when you need to trace them.

EvanLjp:
@wu-sheng  how to trigger the entry point? 
the following config is defined in configuration.yml
```
type: jvm
entryService: http://localhost:8080/exception-checker-spring/case/exceptionchecker
healthCheck: http://localhost:8080/exception-checker-spring/case/healthCheck
startScript: ./bin/startup.sh
runningMode: with_optional
withPlugins: apm-spring-annotation-plugin-*.jar
```
And the configuration in expectedData.yaml is like following:
```
segmentItems:
  - serviceName: exception-checker-spring
    segmentSize: ge 2
    segments:
      - segmentId: not null
        spans:
          - operationName: /case/exceptionchecker
            operationId: 0
            parentSpanId: -1
            spanId: 0
            spanLayer: Http
            startTime: nq 0
            endTime: nq 0
            componentId: 1
            isError: false
            spanType: Entry
            peer: ''
            tags:
              - {key: url, value: 'http://localhost:8080/exception-checker-spring/case/exceptionchecker'}
              - {key: http.method, value: GET}
            skipAnalysis: 'false'
```
When trigger the plugin test, only find health check segment and not find entry point.
```
assert failed.
SegmentNotFoundException:
expected: 
  Segment:
  - span[-1, 0] /case/exceptionchecker

actual: 
  Segment[1722c440d2ce4213804937d65ea51945.35.15993703308650000] e
  expected:     Span[-1, 0] /case/exceptionchecker
  actual:       span[-1, 0] /case/healthCheck
  reason:       [operation name]: expected=>{/case/exceptionchecker}, actual=>{/case/healthCheck}

```

wu-sheng:
After the health check service returns the 200, the plugin test framework would begin to request the entryService

EvanLjp:
> the plugin test framework would begin to request the entryService

so I should use trace-ignore-plugin to ignore it ?

wu-sheng:
> so I should use trace-ignore-plugin to ignore it ?

Once you have the real trace, this will be ignored by the test tool. Currently, as this is the only trace, the tool outputs the error as trace segment not matched.

EvanLjp:
> > so I should use trace-ignore-plugin to ignore it ?
> 
> Once you have the real trace, this will be ignored by the test tool. Currently, as this is the only trace, the tool outputs the error as trace segment not matched.

thx for your help,I would like to check the configuration again.

EvanLjp:
@wu-sheng Have added the spring annotation plugin test. But the tomcat plugin is because of HTTP response code is always gt 400, so the error status is always true when an exception is thrown Spring MVC.

wu-sheng:
> Have added the spring annotation plugin test. But the tomcat plugin is because of HTTP response code is always gt 400, so the error status is always true when an exception is thrown Spring MVC.

You don't have to throw the exception to the controller. You could make the controller in no-exception status, but add a local span to test this. The point I am asking both Spring and Tomcat cases is for testing different class loader cases.

wu-sheng:
`exception-checker-spring-scenario` is failing, please recheck.

EvanLjp:
https://github.com/apache/skywalking-agent-test-tool/pull/23 @wu-sheng  a little bug found when testing. After merge it, would push later.

EvanLjp:
@wu-sheng there are some plugin tests fail because of the bug in the plugin test tool.I don't know whether it can be modified directly, please help check.

wu-sheng:
> there are some plugin tests fail because of the bug in the plugin test tool. I don't know whether it can be modified directly, please help check.

I think we didn't plan to test status=error. Maybe this is some copy/paste issue. Such as hytrix is checking status==false, I don't have an idea why is doing this.

I think you better make the plugin tool update and test case fix in another PR first, then we are back on this. This should be easier, right?

EvanLjp:
> > there are some plugin tests fail because of the bug in the plugin test tool. I don't know whether it can be modified directly, please help check.
> 
> I think we didn't plan to test status=error. Maybe this is some copy/paste issue. Such as hytrix is checking status==false, I don't have an idea why is doing this.
> 
> I think you better make the plugin tool update and test case fix in another PR first, then we are back on this. This should be easier, right?

https://github.com/apache/skywalking/pull/5456 @wu-sheng

EvanLjp:
Grpc plugin test also has some error.

Some codes in grpc agent plugin:
```
org.apache.skywalking.apm.plugin.grpc.v1.client.TracingClientCall#cancel

public void cancel(@Nullable String message, @Nullable Throwable cause) {
        final AbstractSpan span = ContextManager.createLocalSpan(operationPrefix + REQUEST_ON_CANCEL_OPERATION_NAME);
        span.setLayer(SpanLayer.RPC_FRAMEWORK);
        ContextManager.continued(snapshot);

        if (cause != null) {
            span.log(cause);
        }

        try {
            super.cancel(message, cause);
        } catch (Throwable t) {
            ContextManager.activeSpan().log(t);
            throw t;
        } finally {
            ContextManager.stopSpan();
        }
    }

```
The author creates a local span to store error status when canceling. Because only log the exception, the before status should always is a normal span. But the status will be error now. I think this situation can be ignored. After all, OAP can determine the status based on Entry Span only. What do u think?
@wu-sheng

wu-sheng:
> The author creates a local span to store error status when canceling. 

If there is an exception, then the status could be error=true, what is the issue of that?

EvanLjp:
> If there is an exception, then the status could be error=true, what is the issue of that?

For our thinking, errorOccurred and log should come in pairs. For example
``
``
But the author implement is like following:
```
@Override
    public void cancel(@Nullable String message, @Nullable Throwable cause) {
        final AbstractSpan span = ContextManager.createLocalSpan(operationPrefix + REQUEST_ON_CANCEL_OPERATION_NAME);
        span.setLayer(SpanLayer.RPC_FRAMEWORK);
        ContextManager.continued(snapshot);

        if (cause != null) {
            span.log(cause);
        }

        try {
            super.cancel(message, cause);
        } catch (Throwable t) {
            throw t;
        } finally {
            ContextManager.stopSpan();
        }
    }
```
The local span only triggers the log function rather than trigger the 2 functions together. So the local span should always be a normal span. It seems that the author did this on purpose.

wu-sheng:
From my understanding, in default, `ContextManager.activeSpan().errorOccurred().log(t);` == `ContextManager.activeSpan().log(t);`, isn't it?

I know the OAP could determine the exception, I just said, this seems not an issue. Could you be more clear?

EvanLjp:
> From my understanding, in default, `ContextManager.activeSpan().errorOccurred().log(t);` == `ContextManager.activeSpan().log(t);`, isn't it?
The code in grpc plugin is like following:
```
 AbstractSpan span = ContextManager.createLocalSpan(operationPrefix + REQUEST_ON_CANCEL_OPERATION_NAME);
        if (cause != null) {
            span.log(cause);
        }
```
Before : the status of the local span is always false.
Now:  the status of the local span is always true.

The question I want to ask is can we ignore this situation.

wu-sheng:
Got your point this time, I think it should be in error status, this is a previous bug. In default, we always treat exception as the same as error status.

EvanLjp:
ok, will change the grpc plugin test later

test/plugin/scenarios/exception-checker-spring-scenario/src/main/java/org/apache/skywalking/apm/toolkit/trace/IgnoredException.java
```diff
@@ -0,0 +1,35 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * After the exception status checker activated in the agent, the span wouldn't be marked as error status if the
+ * exception has this annotation.
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+public @interface IgnoredException {
```
wu-sheng, line 34: I think you missed the toolkit document update about how to use this annotation.

EvanLjp: thx for your notice will add it later

EvanLjp:
@wu-sheng @kezhenxu94  all plugins test have passed and already added the doc, thx check again.

wu-sheng (approved):
I updated the document a little. This PR is good to me.

PR #5431, opened 2020-09-03
Support SSL for prometheus telemetry and prometheus fetcher

hanahmily:
As docs indicate, the changes intend to introduce SSL to Prometheus telemetry and fetcher.


wu-sheng (commented):
I think you changed the dependency, license check fails.

tools/dependencies/known-oap-backend-dependencies-es7.txt
```diff
@@ -157,7 +157,6 @@ retrofit-2.3.0.jar
 simpleclient-0.6.0.jar
 simpleclient_common-0.6.0.jar
 simpleclient_hotspot-0.6.0.jar
-simpleclient_httpserver-0.6.0.jar
```
wu-sheng, line 160: The whole simpleclient dependencies are missed in the DIST LICENSE, please fix.

wu-sheng (approved):

PR #5454, opened 2020-09-08
Parse TLS mode from ALS

hanahmily:
Set up TLS mode property to Service and instance relationship source, which intends to parse TLS mode from envoy AccessLog. 

The type of TLS mode is "String"  instead of "Enum" due to the issues of variable receivers. In this PR only ALS relevant TLS mode is supported. Tracing or other analyzers might adopt to different terms to identify TLS mode.


hanahmily:
Waiting for https://github.com/apache/skywalking-data-collect-protocol/pull/32

oap-server/server-bootstrap/src/main/resources/oal/core.oal
```diff
@@ -69,6 +69,12 @@ database_access_sla = from(DatabaseAccess.*).percent(status == true);
 database_access_cpm = from(DatabaseAccess.*).cpm();
 database_access_percentile = from(DatabaseAccess.latency).percentile(10);
 
+// Example of TLS related metric. Detected point was not distinguished, that means mTLS requests from both sides will be
+// counted in.
+//////////
+// service_relation_mtls_cpm = from(ServiceRelation.*).filter(tlsMode == "mTLS").cpm();
+// service_instance_relation_mtls_cpm = from(ServiceInstanceRelation.*).filter(tlsMode == "mTLS").cpm();
```
wu-sheng, line 76: We don't visualize tls in the current UI. Please remove them.

wu-sheng (commented):
Please fix e2e. You are breaking the tests.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/ServiceInstanceRelation.java
```diff
@@ -103,6 +103,9 @@ public String getEntityId() {
     @Getter
     @Setter
     private DetectPoint detectPoint;
+    @Getter
+    @Setter
+    private String tlsMode;
```
wu-sheng, line 108: This and service relation source have been changed, `scope-definitions.md` doc should be updated. 

> service_relation_mtls_cpm = from(ServiceRelation.*).filter(tlsMode == "mTLS").cpm();

This example should belong to the doc I pointed to you.

oap-server/server-bootstrap/src/main/resources/oal/core.oal
hanahmily: done

hanahmily: done

wu-sheng (approved):

Issue #4581, opened 2020-03-27
i use es7  When start collector server,There an error : "storage module no provider exists"


MrZhouFor:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug

___
### Question
- What do you want to know?
i use es7  when start collector server, there an error : "storage module no provider exists".(Both versions 6.6 and 7.0 have this error.)
Hao can i start right collector? Thanks
___
### Bug
- Which version of SkyWalking, OS and JRE?
skywalking version : 6.6 or 7.0  
elasticsearch version: 7.6.1    
OS version :  Linux version 3.10.0-1062.12.1.el7.x86_64  (gcc version 4.8.5 20150623 (Red Hat 4.8.5-39) (GCC) ) 

- What happen?

**error log:**
```
2020-03-26 22:52:58,186 - org.apache.skywalking.oap.server.library.module.ModuleDefine - 89 [main] INFO  [] - Prepare the default provider in receiver-clr module.
2020-03-26 22:52:58,187 - org.apache.skywalking.oap.server.starter.OAPServerBootstrap - 59 [main] ERROR [] - storage module no provider exists.
org.apache.skywalking.oap.server.library.module.ProviderNotFoundException: storage module no provider exists.
	at org.apache.skywalking.oap.server.library.module.ModuleDefine.prepare(ModuleDefine.java:86) ~[library-module-6.6.0.jar:6.6.0]
	at org.apache.skywalking.oap.server.library.module.ModuleManager.init(ModuleManager.java:52) ~[library-module-6.6.0.jar:6.6.0]
	at org.apache.skywalking.oap.server.starter.OAPServerBootstrap.start(OAPServerBootstrap.java:47) [server-bootstrap-6.6.0.jar:6.6.0]
	at org.apache.skywalking.oap.server.starter.OAPServerStartUp.main(OAPServerStartUp.java:28) [server-starter-6.6.0.jar:6.6.0]
```

** V6.6 storage config:**
![image](https://user-images.githubusercontent.com/14541486/77727107-862e1a00-7034-11ea-91d7-6fc95b06e2cc.png)

wu-sheng:
Read the documentation, there is one es7 package there.

KiritoStudio:
I met the same issue. The resolution is download **Binary Distribution for ElasticSearch 7** instead of **Binary Distribution** from the [official download page](http://skywalking.apache.org/downloads/).

[...]

PR #5459, opened 2020-09-09
Update README.md

O-ll-O:
If you package with the default value es6, an error is storage module no provider exists will be reported in the cluster environment, add this description.

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -32,6 +32,7 @@ The tag of docker image. The default value is `latest`.
 ### ES_VERSION
 
 The elasticsearch version this image supports. The default value is `es6`, available values are `es6` and `es7`.
+If the image version is 7.+,please set the value to es7.
 
 
 For example, if we want to build images with a hub `foo.io` and a tag `bar`, and it supports elasticsearch 7 at the same time.


docker/README.md
```diff
@@ -32,6 +32,7 @@ The tag of docker image. The default value is `latest`.
 ### ES_VERSION
 
 The elasticsearch version this image supports. The default value is `es6`, available values are `es6` and `es7`.
+If the image version is 7.+,please set the value to es7.
```
wu-sheng, line 35: What does this mean?

O-ll-O: If the value is es6, the docker build reference is apache-skywalking-apm-bin.tar.gz, if the value of helm elastaticsearch.imageTag is 7.+, an error will be reported when deploying in k8s cluster. For details, see: [https://github.com/apache/skywalking/issues/4581]

wu-sheng: I mean, this sentence seems duplicated. The previous sentence says, there are two option values.

O-ll-O: This only gives the values of the two options, but does not indicate that the purpose of these two values is to select the files needed when build the image. This is just a kind reminder

wu-sheng: If you want to give the statement more clear, provide full message for both values. Maybe a table or list.

O-ll-O: The elasticsearch version this image supports. The default value is `es6`, available values are `es6` and `es7`.These two values means:
1. ‘es6’ is to build image through the apache-skywalking-apm-bin.tar.gz file;
2. ‘es7’ is to build image through apache-skywalking-apm-bin-es7.tar.gz file, the elasticsearch version used for storage is 7.+, use this value.

wu-sheng: What do you mean about above comments? I was suggesting you to update the document in this way.
And don't use 7+, ElasticSearch as 8.x releases already, which we never tested.

Issue #5507, opened 2020-09-17
Kafka ProducerConstructorMapInterceptor explicit type-cast problem


lujiajing1126:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Bug
- Which version of SkyWalking, OS and JRE?

8.1.0

- What happened?

According to the definition in the `Kafka Client`, the type of the `bootstrap.servers` property can be either list or comma-separated string (precisely, comma with white space). See [`ConfigDef.java`](https://github.com/apache/kafka/blob/aa5263fba903c85812c0c31443f7d49ee371e9db/clients/src/main/java/org/apache/kafka/common/config/ConfigDef.java#L699-L708)  and [`ProducerConfig.java`](https://github.com/apache/kafka/blob/aa5263fba903c85812c0c31443f7d49ee371e9db/clients/src/main/java/org/apache/kafka/clients/producer/ProducerConfig.java#L294) in the Kafka client library.

But in the [`ProducerConstructorMapInterceptor`](https://github.com/apache/skywalking/blob/2ebb5bdf3af7ae52e70a2145702baa6afeeac3bb/apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/ProducerConstructorMapInterceptor.java#L34), we explicitly cast `Object` to `String`, which in some cases will cause issues, for example #4652 .

Due to the https://github.com/apache/kafka/pull/5781 (from Kafka 2.1), the first parameter of the main constructor has been changed to `Map`, so we have to check the type before casting here. Also by default, [`Spring Boot`](https://github.com/spring-projects/spring-boot/blob/602e62998eb2254517d695fa67246399ba40b8bf/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java#L64) defines `spring.kafka.bootstrap-servers` as `ArrayList`, which may probably lead to errors.

I would like to submit a PR to fix this problem.

wu-sheng:
Thank you for the report, welcome to fix. And note, please update the plugin test cases to verify your changes.

wu-sheng:
> the first parameter of the main constructor has been changed to Map, so we have to check the type before casting here

[...]

Issue #4652, opened 2020-04-15
kafka plugin ERROR


heytong:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?
skywalking 7.0

ERROR 2020-04-15 10:26:09:123 Thread-16 ConstructorInter : ConstructorInter failure. 
java.lang.ClassCastException: java.util.ArrayList cannot be cast to java.lang.String
	at org.apache.skywalking.apm.plugin.kafka.ProducerConstructorMapInterceptor.onConstruct(ProducerConstructorMapInterceptor.java:34)
	at org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ConstructorInter.intercept(ConstructorInter.java:65)
	at org.apache.kafka.clients.producer.KafkaProducer.<init>(KafkaProducer.java)
	at org.apache.kafka.clients.producer.KafkaProducer.<init>(KafkaProducer.java:287)
	at org.apache.kafka.clients.producer.KafkaProducer.<init>(KafkaProducer.java)
	at org.springframework.kafka.core.DefaultKafkaProducerFactory.createKafkaProducer(DefaultKafkaProducerFactory.java:405)
	at org.springframework.kafka.core.DefaultKafkaProducerFactory.createProducer(DefaultKafkaProducerFactory.java:391)
	at org.springframework.kafka.core.KafkaTemplate.getTheProducer(KafkaTemplate.java:463)
	at org.springframework.kafka.core.KafkaTemplate.doSend(KafkaTemplate.java:401)
	at org.springframework.kafka.core.KafkaTemplate.send(KafkaTemplate.java:216)
	
___
### Bug
- Which version of SkyWalking, OS and JRE?
el7, openjdk version "1.8.0_242"
- Which company or project?

- What happen?
If possible, provide a way for reproducing the error. e.g. demo application, component version.

___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

wu-sheng:
Please check whether your Kafka client/server versions in our supported list.

heytong:
> Please check whether your Kafka client/server versions in our supported list.

where is the url for supported list of  kafka version, please.
thanks.

wu-sheng:
It is in the agent set up document. Please read the document.

heytong:
<dependency>
[...]

PR #5510, opened 2020-09-17
Check bootstrapServers type before casting for kafka-clients >= 2.1

lujiajing1126:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

#4652 
#5507 
___
### Bug fix

See issues #5507 for details.

Just check both List and String types.

- [x] unit test cases added for `ProducerConstructorMapInterceptor`
- [x] integration test cases: Setup producer with list-type bootstrap servers without error

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/ProducerConstructorMapInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/ProducerConstructorMapInterceptor.java
@@ -22,16 +22,29 @@ import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedI
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
 import org.apache.skywalking.apm.util.StringUtil;
 
+import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
+/**
+ * Target at kafka-clients 2.1+
+ */
 public class ProducerConstructorMapInterceptor implements InstanceConstructorInterceptor {
+    private static final Pattern COMMA_WITH_WHITESPACE = Pattern.compile("\\s*,\\s*");
 
     @Override
     public void onConstruct(EnhancedInstance objInst, Object[] allArguments) {
         Map<String, Object> config = (Map<String, Object>) allArguments[0];
         // prevent errors caused by secondary interception in kafkaTemplate
         if (objInst.getSkyWalkingDynamicField() == null) {
-            objInst.setSkyWalkingDynamicField(StringUtil.join(';', ((String) config.get("bootstrap.servers")).split(",")));
+            Object bootstrapServers = config.get("bootstrap.servers");
+            if (bootstrapServers instanceof List) {
+                objInst.setSkyWalkingDynamicField(String.join(";", ((List<String>) bootstrapServers)));
+            } else {
+                objInst.setSkyWalkingDynamicField(StringUtil.join(';', COMMA_WITH_WHITESPACE.split((String) bootstrapServers, -1)));
+            }
         }
     }
 }
--- a/apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/ProducerConstructorMapInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/ProducerConstructorMapInterceptor.java
@@ -26,9 +26,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
 
-/**
- * Target at kafka-clients >=2.1.0
- */
 public class ProducerConstructorMapInterceptor implements InstanceConstructorInterceptor {
     private static final Pattern COMMA_WITH_WHITESPACE = Pattern.compile("\\s*,\\s*");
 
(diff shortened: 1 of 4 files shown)


test/plugin/scenarios/spring-kafka-2.3.x-scenario/src/main/java/test/org/apache/skywalking/apm/testcase/spring/kafka/controller/CaseController.java
```diff
@@ -82,6 +85,21 @@ private void setUpProvider() {
...
+        props.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, StringSerializer.class);
+        kafkaTemplate2 = new KafkaTemplate<String, String>(new DefaultKafkaProducerFactory<>(props));
+        try {
+            kafkaTemplate2.send(topicName, "key", "ping").get();
+            kafkaTemplate2.flush();
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+    }
+
```
wu-sheng, line 102: Once you changed this, you should change the expected data yaml to map this.

zhaoyuguang:
I checked version 2.1.0 of kafka-client, this adapter code is necessary
So, LGTM

test/plugin/scenarios/spring-kafka-2.3.x-scenario/src/main/java/test/org/apache/skywalking/apm/testcase/spring/kafka/controller/CaseController.java
lujiajing1126: I have added more test cases for this scenario.  Pls have a check since I am not familiar with this part.

apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/ProducerConstructorMapInterceptor.java
```diff
@@ -22,16 +22,27 @@
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
 import org.apache.skywalking.apm.util.StringUtil;
 
+import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 
+/**
+ * Target at kafka-clients 2.1+
+ */
```
wu-sheng, line 31: Why add this comment? This interceptor is used for 2.1-, isn't it?

lujiajing1126: > Why add this comment? This interceptor is used for 2.1-, isn't it?

No. The `Map` argument is used from Kafka 2.1+

wu-sheng: But this interceptor still supports `String` as a parameter, I still can see the String casting.

wu-sheng: You didn't add any new instrumentation definition. So, this must be compatible with older releases.

lujiajing1126: > But this interceptor still supports `String` as a parameter, I still can see the String casting.

Exactly. Regarding `bootstrap.servers`, the type of the value in the map can be `String` or `List`. Both types are supported by kafka client lib.

wu-sheng: I am not talking about the codes. You added comment says this is for 2.1+, which is not accurate. It supports 2.1 also.

lujiajing1126: > I am not talking about the codes. You added comment says this is for 2.1+, which is not accurate. It supports 2.1 also.

So it should be ">=2.1.0", I guess?

lujiajing1126: Fixed

apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/ProducerConstructorMapInterceptor.java
```diff
@@ -22,16 +22,27 @@
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
 import org.apache.skywalking.apm.util.StringUtil;
 
+import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 
+/**
+ * Target at kafka-clients >=2.1.0
```
wu-sheng, line 30: Still confused what your new comments. Isn't 2.0 even 1.x client libraries using this interceptor?

lujiajing1126: > Still confused what your new comments. Isn't 2.0 even 1.x client libraries using this interceptor?

I see. You are right. It can be used even below 2.0, if users intend to use Map type constructors.

I have removed this comment.

wu-sheng (approved):

PR #5524, opened 2020-09-20
Better witness class for mongo3.8.x

harvies:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___

com.mongodb.operation.AggregateOperationImpl may be more suitable as a witness class for mongodb-3.8, because com.mongodb.client.ClientSession also exists in 4.x. There is a high probability that 3.x and 4.x coexist will lead to 4. The plug-in of x is invalid, I will submit pr to prove it.

com.mongodb.operation.AggregateOperationImpl detail info
image

### DIFF
--- a/test/plugin/scenarios/mongodb-3.x-scenario/bin/startup.sh
+++ b/test/plugin/scenarios/mongodb-3.x-scenario/bin/startup.sh
@@ -18,4 +18,4 @@
 
 home="$(cd "$(dirname $0)"; pwd)"
 
-java -jar ${agent_opts} ${home}/../libs/mongodb-3.x-scenario.jar &
+java -jar -Dskywalking.plugin.mongodb.trace_param=true ${agent_opts} ${home}/../libs/mongodb-3.x-scenario.jar &
--- a/apm-sniffer/apm-sdk-plugin/mongodb-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mongodb/v3/define/v38/MongoDBOperationExecutorInstrumentation.java
+++ b/apm-sniffer/apm-sdk-plugin/mongodb-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mongodb/v3/define/v38/MongoDBOperationExecutorInstrumentation.java
@@ -35,7 +35,7 @@ import org.apache.skywalking.apm.agent.core.plugin.match.NameMatch;
  */
 public class MongoDBOperationExecutorInstrumentation extends ClassInstanceMethodsEnhancePluginDefine {
 
-    private static final String WITNESS_CLASS = "com.mongodb.client.ClientSession";
+    private static final String WITNESS_CLASS = "com.mongodb.operation.AggregateOperationImpl";
 
     private static final String ENHANCE_CLASS = "com.mongodb.client.internal.MongoClientDelegate$DelegateOperationExecutor";
 
--- a/apm-sniffer/apm-sdk-plugin/mongodb-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mongodb/v3/support/MongoSpanHelper.java
+++ b/apm-sniffer/apm-sdk-plugin/mongodb-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mongodb/v3/support/MongoSpanHelper.java
@@ -39,7 +39,7 @@ public class MongoSpanHelper {
         SpanLayer.asDB(span);
 
         if (MongoPluginConfig.Plugin.MongoDB.TRACE_PARAM) {
-            Tags.DB_STATEMENT.set(span, executeMethod + " " + MongoOperationHelper.getTraceParam(operation));
+            Tags.DB_BIND_VARIABLES.set(span, MongoOperationHelper.getTraceParam(operation));
         }
     }
 }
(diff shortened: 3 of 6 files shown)


wu-sheng:
Would you be willing to share your tool as a new open source project? The screenshot seems to be generated by a tool.

test/plugin/scenarios/mongodb-3.x-scenario/bin/startup.sh
```diff
@@ -18,4 +18,4 @@
 
 home="$(cd "$(dirname $0)"; pwd)"
 
-java -jar ${agent_opts} ${home}/../libs/mongodb-3.x-scenario.jar &
\ No newline at end of file
+java -jar -Dskywalking.plugin.mongodb.trace_param=true ${agent_opts} ${home}/../libs/mongodb-3.x-scenario.jar &
```
wu-sheng, line 21: You activated this, but can't see the param in the tags.

harvies:
> Would you be willing to share your tool as a new open source project? The screenshot seems to be generated by a tool.

Consider contributing after it is perfected.

test/plugin/scenarios/mongodb-3.x-scenario/bin/startup.sh
harvies: param is db.statement,In terms of content judgment, I have encountered a problem, can you help me see it, for example, 
image
I can't see the difference between the estimated value and the actual value, but it is an error.

wu-sheng: What do you mean `error`? Why CI reports success?

wu-sheng: > param is db.statement,In terms of content judgment

Also, this place seems like a bug. The parameter should be another tag, like other plugins. Could you fix them too? And if no parameter exists, the tag should not be there too.

harvies: > What do you mean `error`? Why CI reports success?

This pr uses not null judgment, I think it is better to use string equality judgment.

wu-sheng: Let's adjust the tag and don't worry about this check for now, OK?

wu-sheng: Wait, is the statement including the entity name of the operation or not?

wu-sheng: From my reading from your screenshot, only operation is a part of the statement, I think the plugin code has some issue about assembling the statement and parameter. Statement should be an expression and parameters are variable. I am not sure whether mongodb is friendly to this?

wu-sheng: I just read the codes again, the statement tag should be removed, as this actually isn't a statement anymore. We should use `DB_BIND_VARIABLES` tag to replace `DB_STATEMENT`, because the statement is being used to check slow SQL. Also, please help to remove the `executeMethod` as the prefix, it has been the operation name of the span.

harvies: > From my reading from your screenshot, only operation is a part of the statement, I think the plugin code has some issue about assembling the statement and parameter. Statement should be an expression and parameters are variable. I am not sure whether mongodb is friendly to this?

Mongo does not seem to have a placeholder similar to sql.

harvies: > I just read the codes again, the statement tag should be removed, as this actually isn't a statement anymore. We should use `DB_BIND_VARIABLES` tag to replace `DB_STATEMENT`, because the statement is being used to check slow SQL. Also, please help to remove the `executeMethod` as the prefix, it has been the operation name of the span.

Okay, let me adjust.

harvies:
mongo-java-driver seems to be an old package that contains codes such as bson. After version 3.9.1, maintenance has been discontinued and replaced by mongodb-driver-core+bson+mongodb-driver-sync. apm-mongodb3.x-plugin relies on mongo-java-driver, but mongodb-3.x-scenario relies on mongodb-driver (mongodb-driver-core+bson). Does it need to be adjusted?

image

image


https://repo1.maven.org/maven2/org/mongodb/mongodb-driver-core/

harvies:
> If these changes happened, then we need witness classes to support both version. Two interceptors for 3.9.0 and 3.9.1+
Okay, I will test it.

harvies:
Need to adapt to all revisions, such as 3.12.0-3.12.6.

harvies:
I saw that the version of our project is mongo-java-driver 3.0. I think that 3.x has no bugs and will not change it. I only change db.statement to db.bind_vars and remove executeMethod. What do you think?

wu-sheng:
> I saw that the version of our project is mongo-java-driver 3.0. I think that 3.x has no bugs and will not change it. I only change db.statement to db.bind_vars and remove executeMethod. What do you think?

Once it is working, it is good to me.

wu-sheng:
I think you broke the UTs.

wu-sheng (approved):

PR #4816, opened 2020-05-24
Provide agent-side meter api

mrproliu:
Please answer these questions before submitting pull request

- Why submit this pull request?
- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
#4808 
___
### New feature or improvement
- Provide the agent-side meter API.

PR #5529, opened 2020-09-21
Add `MOUNT` config in the agent core.

wu-sheng:
We are preparing to accept the server-side plugins, starting with ShardingSphere. @mrproliu is providing the meter APIs for this kind of plugins in #4816. 

In the ShardingSphere server plugin, it would only work when the agent installed in the ShardingSphere proxy. In this case, I hope we could mount the whole metrics plugin like this, `java -javaagent:skywalking-agent.jar -jar proxy.jar -Dskywalking.plugin.mount=server-plugins/shardingsphere`. Then all default plugins would not work, but plugins in `server-plugins/shardingsphere` would be loaded.

Also, this could be required when we have more server side plugins, rather than plugins for client libraries in the business applications.

WDYT? @apache/skywalking-committers


dmsolr (approved):

kezhenxu94 (approved):
LGTM

PR #5557, opened 2020-09-25
Add a muse configuration file

TomMD:
Skywalking is going to be part of the ApacheCon bug bash.  As part of that, ASF Infra will be installing the github app "muse-dev" in this repository.  While there will certainly be a list of issues provided, scores can be kept track of automated via the bot.  Beyond ApacheCon, the muse app is much like other CI serves in that there are actions taking on PRs and those actions need some minor amount of configuration.

In short: This PR adds a muse configuration and file, sufficient to allow running Muse CI on Skywalking.

### DIFF
--- a/.muse/config.toml
+++ b/.muse/config.toml
@@ -1,0 +1,2 @@
+jdk11 = true
+setup = ".muse/setup.sh"
--- a/.muse/setup.sh
+++ b/.muse/setup.sh
@@ -1,0 +1,2 @@
+#!/usr/bin/env bash
+git submodule update --init --recursive
--- a/pom.xml
+++ b/pom.xml
@@ -451,6 +451,7 @@
                         <exclude>**/.git/**</exclude>
 
                         <!-- CI files -->
+                        <exclude>**/.muse/**</exclude>
                         <exclude>**/.travis.yml</exclude>
 
                         <!-- GitHub files -->


wu-sheng (commented):
Apache 2.0 LICENSE header should be added.

wu-sheng:
A question, where should we be able to see the scores, issues, and comments? And are those available after we merge this?

TomMD:
> A question, where should we be able to see the scores, issues, and comments? And are those available after we merge this?

The scores will be available when the bug bash starts.

Issues are available any time via the muse console.  This PR actually already has a muse result, but it is partial because some tools can't run without the configuration file in master.  A full run can be triggered via the console, I've done so here: https://console.muse.dev/result/TomMD/skywalking/01EK1C2GQM8YVF69MPYWTF2PMY?tab=results

Notice the full list of issues is verbose.  The typical work flow is to ignore that list, continue developing as you always have.  Muse will post a comment to the pull request discussion whenever an issue is detected that relates to the pull request.  If you want to work on quality metrics, such as at a bug bash, then obviously looking through the issues list list becomes central - we will be issuing some suggestions there and the search bar on the console can help.

kezhenxu94 (commented):
@TomMD either add license headers or exclude the two files from license check, and it looks good to me

TomMD:
@kezhenxu94 How do I exclude the files from the license check?  Looking around I don't see any examples of doing that, happy to declare these are public domain, unlicensed, or what have you.

kezhenxu94:
> @kezhenxu94 How do I exclude the files from the license check? Looking around I don't see any examples of doing that, happy to declare these are public domain, unlicensed, or what have you.

@TomMD it's here 

https://github.com/apache/skywalking/blob/f397c1bb49e5eff7aef3c936f316a4c5f62484b0/pom.xml#L422

TomMD:
Thank you @kezhenxu94, I've now updated the PR.

kezhenxu94 (approved):
LGTM

wu-sheng (approved):

Issue #5556, opened 2020-09-24
Try to set up cancel-workflow-runs plugin to our Action flows.


wu-sheng:
Following the talks with the Airflow team here, https://github.com/apache/airflow/issues/11114, Jarek built the very cool plugin to cancel unnecessary tasks.

https://github.com/marketplace/actions/cancel-workflow-runs

@kezhenxu94 Please try to set this up. If you face some issues, we could ping `potiuk`(GitHub id) for help.

kezhenxu94:
@wu-sheng I finally realize that this plugin may be less useful in SkyWalking than we thought, `cancel-workflow-runs` itself is a workflow **that needs available slots** to schedule it, and usually once a commit is pushed, 100+ jobs are fired , if new commit is pushed, we expect the `cancel-workflow-runs` to run and cancel the old builds, but NOW there is no available slots to run `cancel-workflow-runs` actually (due the previous outdated builds), and we have to wait the previous jobs to finish (or manually cancel them), which is the same as before, the only benefit is that if **some of** the old builds are too slow to finish, it can be cancelled when `cancel-workflow-runs` is scheduled, but this can be achieved by specifying a relatively small `timeout-minutes`, is it the case? @potiuk or did I miss anything?

wu-sheng:
@kezhenxu94 I think this definitely needs a slot to run at least. My question would be, do this plugin could detect, there is further commits after the current one? If so, could it cancel all other builds(if haven't run yet) of this commit?

My expectation for the plugin is mainly about the above point.

kezhenxu94:
> @kezhenxu94 I think this definitely needs a slot to run at least. My question would be, do this plugin could detect, there is further commits after the current one? If so, could it cancel all other builds(if haven't run yet) of this commit?

@wu-sheng I do some experiments and see the picture below

<img width="953" alt="image" src="https://user-images.githubusercontent.com/15965696/94253221-8eb9cb80-ff57-11ea-8fd2-035bf4254e57.png">

wu-sheng:
I could have your point, YES, the effect could be limited, but in special circumstances, it still helps, right?

kezhenxu94:
> I could have your point, YES, the effect could be limited, but in special circumstances, it still helps, right?

Sure, since the cancelling job is not time consuming, it could be better in some scenarios, I'll continue the PR

potiuk:
[...]

PR #5558, opened 2020-09-25
[CI] Cancel outdated jobs when new builds are fired

kezhenxu94:
Also exclude paths not to run plugin tests

### DIFF
--- a/.github/workflows/cancel.yaml
+++ b/.github/workflows/cancel.yaml
@@ -26,6 +26,7 @@ on:
       - requested
 
 jobs:
+  # Cancel outdated builds in the repo, including all action tasks of all commits.
   cancel-outdated-builds:
     runs-on: ubuntu-18.04
     timeout-minutes: 10
--- a/.github/workflows/cancel.yaml
+++ b/.github/workflows/cancel.yaml
@@ -1,0 +1,38 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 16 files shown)


.github/workflows/cancel.yaml
```diff
@@ -0,0 +1,38 @@
...
+  workflow_run:
+    workflows:
+      - CI AND IT
+      - E2E
+      - PluginsTest
+    types:
+      - requested
+
+jobs:
+  cancel-outdated-builds:
```
wu-sheng, line 29: Could you share what cancel tests could do now? And add those add comments here?

kezhenxu94: Compared to https://github.com/apache/skywalking/issues/5556#issuecomment-698836978 , now the improvement is

image

this is helpful whenever the `cancel-outdated-builds` job get a slot to run, it cancels **ALL** outdated builds, no matter which PR it is running for. For example, if PR 1 has 2 jobs in queue, where 1 jobs are outdated, and PR 2 has 100 jobs in queue, where 99 jobs are outdated, when PR1 triggers a `cancel-outdated-builds` to run, it will cancel outdated jobs in both PR1 and PR2, leaving only 2 latest jobs to run for PR1 and PR2 respectively.

wu-sheng: OK, it will scan all outdates jobs in all PRs. Then it only requires one slot, then the blocking queue could be clear as much as possible.

wu-sheng: Could you add some comments about this?

```suggestion
  # Cancel outdated builds in the repo, including all action tasks of all commits.
  cancel-outdated-builds:
```

wu-sheng (approved):

wu-sheng:
@potiuk We are going to land this on the master branch :) Thanks for providing this cool thing.

potiuk:
I've also added few more things today - including self-preservation mechanism (v4_4) after some edge cases we discovered in Airflow. I am looking at our experiences and in case I find any problems I am fixing it rather quickly - so let me know if you experience any problems. The way I am doing it now is rather "aggressive" and there might be some edge-cases that I missed (it is surprisingly complex with some missing APIs from GitHub - BTW - I requested those :))

potiuk:
Just a small update - the versions are updated fast :). Some edge cases and bug fixes and we are at 4_7 now (I chose this weekend to implement some refactoring that led to some problems. All of them should be fixed in 4.7 version: https://github.com/marketplace/actions/cancel-workflow-runs :D.

PR #5565, opened 2020-09-26
Polish comments of the meter at the agent side.

wu-sheng:
The previous comments have some grammar issues, and some of them are copied from the Prometheus. Polish all of them and add more.


mrproliu (approved):
Thanks for helping to polish comments.

PR #5571, opened 2020-09-28
inject span tags from correlation

EvanLjp:

- [ ] New feature provided

___
- supports full link stress testing to distinct the traffic source.
- supports tags query on UI

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/NodeType.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/NodeType.java
@@ -22,14 +22,13 @@ import org.apache.skywalking.apm.network.language.agent.v3.SpanLayer;
 import org.apache.skywalking.oap.server.core.UnexpectedException;
 
 /**
- * Node type describe which kind of node of Service or Network address represents to.
- * <p>
- * The value comes from 'org.apache.skywalking.apm.network.language.agent.SpanLayer' at first place, but most likely it
- * will extend and be used directly from different sources, such as Mesh.
+ * Node type describe which kind of node of Service or Network address represents to. The node with {@link #Normal} and
+ * {@link #Browser} type would be treated as an observed node.
  */
 public enum NodeType {
     /**
-     * <code>Unknown = 0;</code>
+     * <code>Normal = 0;</code>
+     * This node type would be treated as an observed node.
      */
     Normal(0),
     /**
@@ -54,6 +53,7 @@ public enum NodeType {
     Cache(5),
     /**
      * <code>Browser = 6;</code>
+     * This node type would be treated as an observed node.
      */
     Browser(6),
     /**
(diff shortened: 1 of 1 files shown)


EvanLjp:
@wu-sheng @kezhenxu94 Do you think the MD5 algorithm needs to be modified?

wu-sheng (changes requested):
This will make duplicated injection for every segment. I prefer we don't do this. You are making this similar to OpenTracing Baggage, which we didn't plan to support when we disigned the correlation context.

Also, every new feature should be discussed before changing im case to avoid argument after you have finished the prototype codes.

wu-sheng:
MD5 thing is not the thing we should discuss, please read above comments. That is the key.

EvanLjp:
> This will make duplicated injection for every segment. I prefer we don't do this. You are making this similar to OpenTracing Baggage, which we didn't plan to support when we disigned the correlation context.
> 
> Also, every new feature should be discussed before changing im case to avoid argument after you have finished the prototype codes.

The above code is only 

> This will make duplicated injection for every segment. I prefer we don't do this. You are making this similar to OpenTracing Baggage, which we didn't plan to support when we disigned the correlation context.
> 
> Also, every new feature should be discussed before changing im case to avoid argument after you have finished the prototype codes.

Do you have any other ideas? This PR comes from support for online flow pressure measurement.

wu-sheng:
The key issue is, injecting this has high-security and memory/network/backend overload concerns. We should carefully reconsider how to do this. Personally, I still prefer an API or a pure optional plugin. The key point is, asking users to activate this manually, then we could set up documentation with it to tell users, note the security issue.

EvanLjp:
> The key issue is, injecting this has high-security and memory/network/backend overload concerns. We should carefully reconsider how to do this. Personally, I still prefer an API or a pure optional plugin. The key point is, asking users to activate this manually, then we could set up documentation with it to tell users, note the security issue.

@wu-sheng  there are some ideas about supporting it in an optional way.
step 1:
refactor some  codes of the span behavior:
1. define a new span callback interface for service loader
2. define a span filter chain to process span behavior, such as extract for the carrier or inject to carrier
3. the order of the span filter chain triggered is ordered by the number of callback interfaces. 

step 2:
supports correlation behavior by filter chain:
1. the default correlation behavior is only propagate
2. the optional behavior will add an injection behavior, the code way is to add a filter rather than override to ensure the stability of the main process.

EvanLjp:
@wu-sheng thx for review this version. I have extracted the dangerous behavior to the optional plugin. If this behavior is tolerated, I will modify the doc and add the unit test later.

wu-sheng:
> Do you think the MD5 algorithm needs to be modified?

Back to this, MD5 could be replaced by BASE64, basically in the open-source.

wu-sheng (commented):
I am a little confused about this plugin. Could you consider simply providing `ContextManagerExtendService`'s extension? It could create an extension of `TracingContext`.

Then, could add new logic into `TracingContext#extract` w/o changing any core level code.

Make sense?

EvanLjp:
> I am a little confused about this plugin. Could you consider simply providing `ContextManagerExtendService`'s extension? It could create an extension of `TracingContext`.
> 
> Then, could add new logic into `TracingContext#extract` w/o changing any core level code.
> 
> Make sense?

no problem

PR #5579, opened 2020-09-28
Support JDK 11 compiling and upgrade dependencies

wu-sheng:
Hi all @apache/skywalking-committers 

According to OpenJDK website(https://adoptopenjdk.net/support.html), there are 2 TLS versions for now
1. Java 8 (LTS). At Least May 2026
1. Java 11 (LTS). At Least Oct 2024

We should make the JDK11 compiling passed. That is the primary agenda when I started this work. Also, with doing this, I planned to upgrade the gRPC too.
During the JDK11 compiling fix, the version of PowerMock was detected as the main block, so, I upgraded it to the 2.x directly.

So, you would note,
1. Some libraries' versions(such as protobuf, protoc) changed because of gRPC or PowerMock requirements.
1. Some tests are deleted because they are not valid in the latest PowerMock. Sorry, I don't have enough time to fix all, because some of them make no sense to me.
1. LICENSE updated

If you have any concerns, please reply.


oap-server/server-configuration/configuration-etcd/src/test/java/org/apache/skywalking/oap/server/configuration/etcd/EtcdConfigWatcherRegisterTest.java
```diff
@@ -1,134 +0,0 @@
...
-
-@RunWith(PowerMockRunner.class)
-@PrepareForTest({
-    EtcdKeysResponse.class,
-    EtcdUtils.class,
-    EtcdClient.class,
-    URI.class
-})
-@PowerMockIgnore({"javax.management.*"})
-public class EtcdConfigWatcherRegisterTest {
```
wu-sheng, line 58: @kezhenxu94 @dmsolr @JaredTan95 This test can't pass on the GitHub Action's JDK11, but passed on the local JDK 11.

wu-sheng: At the GitHub Linux JDK1 env, this causes

> jdk.internal.loader.ClassLoaderValue (in module java.base) because module java.base does not export jdk.internal.loader to unnamed module

NOTE, I changed the ignore to `@PowerMockIgnore({"com.sun.org.apache.xerces.*", "javax.xml.*", "org.xml.*", "javax.management.*", "org.w3c.*"})`

kezhenxu94 (approved):

PR #5583, opened 2020-09-29
Fix server-starter-es7 missing dependencies

Gallardot:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

Binary distribution for ElasticSearch 7 are missing dependent jars. If we enable `receiver_zipkin` or `receiver_jaeger`, we will get an exception of `Module not found`.
- How to fix?

Add the dependent jars.
___

### DIFF
--- a/oap-server/server-starter-es7/pom.xml
+++ b/oap-server/server-starter-es7/pom.xml
@@ -44,6 +44,21 @@
             <artifactId>storage-elasticsearch7-plugin</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.skywalking</groupId>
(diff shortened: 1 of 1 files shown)


dmsolr (approved):

kezhenxu94 (changes requested):
If I remember correctly, we don't support (yet) es7 with {Zipkin,Jaeger} at the moment, and they need some minor adaptation, simply adding these dependencies doesn't work as expected

wu-sheng (changes requested):
I don't think this is right. The codes must be rewiten because of APIs are not compatible at the ElasticSearch level.

PR #5585, opened 2020-09-29
fix timebucket converter issue

dmsolr:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/TimeBucket.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/TimeBucket.java
@@ -113,12 +113,15 @@ public class TimeBucket {
             case Second:
                 calendar.set(Calendar.SECOND, (int) (timeBucket % 100));
                 timeBucket /= 100;
+                // Fall through
             case Minute:
                 calendar.set(Calendar.MINUTE, (int) (timeBucket % 100));
                 timeBucket /= 100;
+                // Fall through
             case Hour:
                 calendar.set(Calendar.HOUR_OF_DAY, (int) (timeBucket % 100));
                 timeBucket /= 100;
+                // Fall through
             case Day:
                 calendar.set(Calendar.DAY_OF_MONTH, (int) (timeBucket % 100));
                 timeBucket /= 100;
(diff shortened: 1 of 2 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/TimeBucket.java
```diff
@@ -104,6 +104,10 @@ public static boolean isDayBucket(long timeBucket) {
     public static long getTimestamp(long timeBucket, DownSampling downsampling) {
         Calendar calendar = Calendar.getInstance();
         calendar.setTimeInMillis(0);
+        calendar.set(Calendar.MILLISECOND, 0);
+        calendar.set(Calendar.SECOND, 0);
+        calendar.set(Calendar.MINUTE, 0);
+        calendar.set(Calendar.HOUR_OF_DAY, 0);
```
kezhenxu94, line 110: Do you need these if you already have `calendar.setTimeInMillis(0);`?

oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/analysis/TimeBucketTest.java
```diff
@@ -18,63 +18,54 @@
...
+            }
+            case Hour: {
+                instance.set(Calendar.MINUTE, 0);
+            }
+            case Minute: {
+                instance.set(Calendar.SECOND, 0);
+            }
+            case Second: {
+                instance.set(Calendar.MILLISECOND, 0);
+            }
```
kezhenxu94, line 67: Suggest to add comment about the deliberate "fall through" of the `switch` statement (as well as the ordering), in case future developers consider it as a bug, which is common

kezhenxu94 (commented):
Looks generally good to me, some nits inline

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/TimeBucket.java
dmsolr: Yes, they are required. The time equals zone offset  when `calendar.setTimeInMillis(0);` was presented. And then, I found some time zone is not a times as hour. But they all is a times as minute. (L107/L108 is optional.)

wu-sheng (approved):

Issue #5584, opened 2020-09-29
skywalking maybe not supports some concepts in zipkin: rc rs ss sc


EvanLjp:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
Skywalking maybe not supports some concepts in Zipkin: cr cs ss sr

___
### Question
- What do you want to know?
![image](https://user-images.githubusercontent.com/31562192/94554428-a7462080-028c-11eb-8b85-5e72b7b7a6dc.png)

Skywalking supports compute the relation in service or endpoint scope. But the latency maybe not successful or not clear.

for computing the endpoint relation:
![image](https://user-images.githubusercontent.com/31562192/94554749-20457800-028d-11eb-9d48-63a53c60ba81.png)
![image](https://user-images.githubusercontent.com/31562192/94554887-52ef7080-028d-11eb-8b11-b8b5d0170596.png)
![image](https://user-images.githubusercontent.com/31562192/94554940-67cc0400-028d-11eb-9aa7-34a72c15fac7.png)

the latency is only the entry span cost (ss - sr) rather sr - cs in Zipkin.
So we cannot monitor the network latency in Skywalking, is it right? Is there any other way to solve it?.

wu-sheng:
You should understand what is endpoint relationship. It is not the client server relationship.

It is entrance uri to upstream entrance uri.

EvanLjp:
sorry, I mistake the relations. But monitor the cost (sr - cs) is also a common condition, especially in MQ condition.  Is there any other way to solve it by Skywalking?

wu-sheng:
For a trace view, it is easy, same in SkyWalking and Zipkin. 

But from my guessing, you are asking the statistics of the MQ consuming. That is not available in the Zipkin too. So, if you want to talk about that, the only high-performance way is to propagate the sending time(header injecting timestamp) in the extension header, https://github.com/apache/skywalking/blob/master/docs/en/protocols/Skywalking-Cross-Process-Propagation-Headers-Protocol-v3.md#extension-header-item.

If you are going to implement this, we should consider making it optional, such as
- `sw8-x` = `0-mq_timestamp` or `-mq_timestamp`
- `sw8-x` = `0`

FYI @arugal Would this be a break for the go2sky side if we enhance in this way?

arugal:
> For a trace view, it is easy, same in SkyWalking and Zipkin.
> 
[...]

PR #5590, opened 2020-09-29
Make the extension header really extendable

wu-sheng:
The old implementation doesn't work when we extend this header. The extension header should be used and enhanced when we do deeper analysis which required more context. Like this, https://github.com/apache/skywalking/issues/5584#issuecomment-700793580


wu-sheng:
UT will fail because of #5585.

arugal (approved):
LGTM

PR #5602, opened 2020-09-30
Promote Integer to Long before multiplying to avoid numeric overflow

kezhenxu94:


apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/buffer/Channels.java
```diff
@@ -36,12 +36,13 @@ public Channels(int channelSize, int bufferSize, IDataPartitioner<T> partitioner
...
             if (BufferStrategy.BLOCKING.equals(strategy)) {
-                bufferChannels[i] = new ArrayBlockingQueueBuffer<T>(bufferSize, strategy);
+                bufferChannels[i] = new ArrayBlockingQueueBuffer<>(bufferSize, strategy);
             } else {
-                bufferChannels[i] = new Buffer<T>(bufferSize, strategy);
+                bufferChannels[i] = new Buffer<>(bufferSize, strategy);
             }
         }
-        size = channelSize * bufferSize;
+        // noinspection PointlessArithmeticExpression
```
wu-sheng, line 44: I think this comment makes sense. 
The code fix is fine, but please correct the comments.

wu-sheng (approved):

PR #5605, opened 2020-09-30
Replace getClass() with instanceOf()

ascrutae:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

Here is the [report](https://console.muse.dev/result/TomMD/skywalking/01EHQHJ02AQZNKB6JN80X17XHA?search=EqualsGetClass&tab=results)

### DIFF
--- a/apm-application-toolkit/apm-toolkit-meter/src/main/java/org/apache/skywalking/apm/toolkit/meter/MeterId.java
+++ b/apm-application-toolkit/apm-toolkit-meter/src/main/java/org/apache/skywalking/apm/toolkit/meter/MeterId.java
@@ -64,7 +64,7 @@ public class MeterId {
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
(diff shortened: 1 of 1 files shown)


apm-application-toolkit/apm-toolkit-meter/src/main/java/org/apache/skywalking/apm/toolkit/meter/MeterId.java
```diff
@@ -64,7 +64,7 @@ public MeterId copyTo(String name, MeterType type) {
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
+        if (o == null || !(o instanceof MeterId)) return false;
```
wu-sheng, line 67: This is how IDE(IntelliJ) generated. I don't think this is an issue.

wu-sheng (changes requested):

PR #5607, opened 2020-09-30
Only get backing array of ByteBuffer when it has one

kezhenxu94:
ByteBuffer.array() shouldn't be called unless ByteBuffer.arrayOffset() is used or if the ByteBuffer was initialized using ByteBuffer.wrap() or ByteBuffer.allocate().


kezhenxu94:
@dmsolr can you please confirm that whether the ByteBuffer passed by Avro server is always backed with an array , if that’s the case, we don’t need to fix this

PR #5616, opened 2020-10-01
fix RESOURCE_LEAK

x22x22:
use  try-with-resources

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

RESOURCE_LEAK

- How to fix?

use  try-with-resources

___


apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/conf/IgnoreConfigInitializer.java
```diff
@@ -89,10 +88,12 @@ private static void overrideConfigBySystemProp() throws IllegalAccessException {
     private static InputStream loadConfigFromAgentFolder() throws AgentPackageNotFoundException, ConfigNotFoundException {
         File configFile = new File(AgentPackagePath.getPath(), CONFIG_FILE_NAME);
         if (configFile.exists() && configFile.isFile()) {
-            try {
+            try(var configFileStream = new FileInputStream(configFile)) {
```
dmsolr, line 91: It is closed in L50.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/loader/AgentClassLoader.java
```diff
@@ -192,9 +193,11 @@ public URL nextElement() {
                 for (String fileName : jarFileNames) {
                     try {
                         File file = new File(path, fileName);
-                        Jar jar = new Jar(new JarFile(file), file);
-                        jars.add(jar);
-                        LOGGER.info("{} loaded.", file.toString());
+                        try(var jarFile = new JarFile(file)){
```
dmsolr, line 196: JarFile need to be resident in memory. @wu-sheng @acurtain

dmsolr (changes requested):

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/loader/AgentClassLoader.java
wu-sheng: `var` is not acceptable in the JDK8 compiler. Why use this?

x22x22: There will be no effect, because the following is "new Jar(jarFile, file);"
不会有影响, 因为后面是"new Jar(jarFile, file);"

wu-sheng: We don't recommend abusing `lombok`. Its annotation-based usages are enough. Don't put `var` in the codes.

wu-sheng: We are using Java, rather than Scala or Kotlin today, we don't want to make codes simpler, but want to make the codes easier to read for everyone. At here, you are just trying to fix the muse recommendation, so, please don't bring any related thing in.

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/conf/IgnoreConfigInitializer.java
x22x22: ok, is false positive.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/loader/AgentClassLoader.java
x22x22: is false positive, I have restored the original code

wu-sheng (commented):
Codestyle issue.

oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/handler/SpanProcessor.java
```diff
@@ -39,37 +39,36 @@ public SpanProcessor(SourceReceiver receiver) {
...
     }
 
     private InputStream getInputStream(HttpServletRequest request) throws IOException {
-        InputStream requestInStream;
-
         String headEncoding = request.getHeader("accept-encoding");
         if (headEncoding != null && (headEncoding.indexOf("gzip") != -1)) {
-            requestInStream = new GZIPInputStream(request.getInputStream());
+            try(InputStream requestInStream = new GZIPInputStream(request.getInputStream());){
+                return requestInStream;
```
wu-sheng, line 67: Closing this?

x22x22: Restored code.

x22x22:
> Codestyle issue.

done

wu-sheng (approved):

dmsolr (approved):
lgtm

PR #5618, opened 2020-10-01
Fix thread safety violation

x22x22:

- [x] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

Fix thread safety violation

- How to fix?

add synchronized or use Atomic Class
___


apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumeDriver.java
```diff
@@ -91,7 +91,7 @@ public void begin(Channels channels) {
     }
 
     @Override
-    public boolean isRunning(Channels channels) {
+    public synchronized boolean isRunning(Channels channels) {
         return running;
     }
```
dmsolr, line 96: I think that to change to a `volatile` variable is enough. I mean replacing `boolean running` with `volatile boolean running`.

x22x22: ok

PR #5644, opened 2020-10-10
Provide Thrift Plugin

### LINKED ISSUE #1534: Does it support thrift?

- [☑️] Question or discussion
- [ ] Bug
- [ ] Requirement

___
### Question
- What do you want to know?

My WebApp calls ServerApp through thrift and I tried to trace the call chain using skywalking.I found WebApp only in topology,but no ServerApp.However I could trace both WebApp and ServerApp when I used springcloud instead of thrift.Does the "Java auto instrumentation agent" support thrift?

dmsolr:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues
closes #1534

___
## Provide Thrift Plugin

Because Thrift cannot carry `metadata` to transport Trace Header in the original API, we transport those by wrapping `TProtocolFactory` to do that.

### Principle about transport Trace Header

Thrift allows us to append any additional field in the `Message` even if the receiver doesn't deal with them. This data is going to be skipped while no one reads. Base on this, we take the 8888th field of Message to store Trace Header(or metadata) and to transport. That means the message with 8888 [field id](http://thrift.apache.org/docs/idl#field-id) must be revered.

Anyway, I don't think have any message big in reality.

### Snapshot
image

image

image


apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/TBaseProcessorInterceptor.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+import org.apache.skywalking.apm.plugin.thrift.wrapper.Context;
+import org.apache.skywalking.apm.plugin.thrift.wrapper.ServerInProtocolWrapper;
+import org.apache.thrift.ProcessFunction;
+import org.apache.thrift.TBaseProcessor;
+
+/**
+ * @see TBaseProcessor
```
wu-sheng, line 34: Should provide the full class name here. Also, I don't think only a `@see` is a reasonable comment.

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/client/TAsyncClientInterceptor.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.skywalking.apm.agent.core.logging.api.LogManager;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
+import org.apache.skywalking.apm.plugin.thrift.commons.ReflectionUtils;
+import org.apache.skywalking.apm.plugin.thrift.wrapper.ClientOutProtocolWrapper;
+import org.apache.thrift.async.TAsyncClient;
+import org.apache.thrift.protocol.TProtocolFactory;
+
+/**
+ * @see TAsyncClient
```
wu-sheng, line 31: See should be after the real comments, not before.

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/client/TAsyncMethodCallInterceptor.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    @Override
+    public void beforeMethod(EnhancedInstance objInst,
+                             Method method,
+                             Object[] allArguments,
+                             Class<?>[] argumentsTypes,
+                             MethodInterceptResult result) throws Throwable {
+        // TODO: It is hard to get arguments of method. So, currently, we don't do that. If you have a good idea, please fix it.
```
wu-sheng, line 79: What do you mean TODO?

wu-sheng:
> Thrift allows us to append any additional field in the Message even if the receiver doesn't deal with them. This data is going to be skipped while no one reads.

According to this, do we have a chance to get the number of using fields in the message? I think that is more reasonable?
And, does `8888` field index cost any performance concern? How does Thrift deal with the blank between using and the 8888, such as using 10 fields, then what are in 11 ~ 8888.

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/wrapper/ServerInProtocolWrapper.java
```diff
@@ -0,0 +1,110 @@
...
+            try {
+                TMap tMap = super.readMapBegin();
+                Map<String, String> header = new HashMap(tMap.size);
+
+                for (int i = 0; i < tMap.size; i++) {
+                    header.put(readString(), readString());
+                }
+
+                super.readMessageEnd();
+                super.readFieldEnd();
```
kezhenxu94, line 70: I think this should be in the `finally` block, otherwise, if it throws exception, there may be an infinite recursive call of `readFieldBegin`.

test/plugin/scenarios/thrift-scenario/thrift-client/src/main/java/org/apache/skywalking/apm/testcase/thrift/client/controller/CaseController.java
```diff
@@ -0,0 +1,99 @@
...
+    private static final Logger logger = LogManager.getLogger(CaseController.class);
+
+    private static final String SUCCESS = "Success";
+    private IClient async;
+    private IClient sync;
+
+    private IClient hasync;
+
+    private final AtomicInteger status = new AtomicInteger(0);
+    private final CountDownLatch initialed = new CountDownLatch(1);
```
kezhenxu94, line 49: `initialed` -> `initialized`

kezhenxu94 (commented):
And I strongly suggest to fix the problems that @muse-dev pointed out, returning in the `finally` block is not a good practice

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/client/TAsyncMethodCallInterceptor.java
```diff
@@ -0,0 +1,108 @@
...
+        objInst.setSkyWalkingDynamicField(async);
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst,
+                              Method method,
+                              Object[] allArguments,
+                              Class<?>[] argumentsTypes,
+                              Object ret) throws Throwable {
+        if (ContextManager.isActive()) {
```
wu-sheng, line 92: This check is unusual, could you explain why?

wu-sheng: Same question for all `ContextManager.isActive()` check.

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/client/TServiceClientReceiveInterceptor.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public void handleMethodException(final EnhancedInstance enhancedInstance,
+                                      final Method method,
+                                      final Object[] objects,
+                                      final Class<?>[] classes,
+                                      final Throwable throwable) {
+        if (ContextManager.isActive()) {
+            ContextManager.activeSpan().errorOccurred().log(throwable);
```
wu-sheng, line 61:
```suggestion
            ContextManager.activeSpan().log(throwable);
```

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/wrapper/ServerInProtocolWrapper.java
```diff
@@ -0,0 +1,108 @@
...
+                }
+
+                AbstractSpan span = ContextManager.createEntrySpan(
+                    context.getOperatorName(), createContextCarrier(header));
+                span.start(context.startTime);
+                span.tag(TAG_ARGS, context.getArguments());
+                span.setComponent(ComponentsDefine.THRIFT_SERVER);
+                SpanLayer.asRPCFramework(span);
+            } catch (Throwable throwable) {
+                LOGGER.error("Failed to resolve header or create EntrySpan.", throwable);
```
wu-sheng, line 76: All catches are not necessary in the interceptor scope, as the core has try/catch-log already.

dmsolr: These wrapping objects are executed outside of interceptor.

wu-sheng:
Your plugin test fails, please recheck

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/client/TServiceClientReceiveInterceptor.java
```diff
@@ -0,0 +1,62 @@
...
+                             final MethodInterceptResult ret) throws Throwable {
+    }
+
+    @Override
+    public Object afterMethod(final EnhancedInstance objInst,
+                              final Method method,
+                              final Object[] objects,
+                              final Class<?>[] classes,
+                              final Object ret) throws Throwable {
+        if (ContextManager.isActive() && Objects.nonNull(objInst.getSkyWalkingDynamicField())) {
```
wu-sheng, line 48: Same question, why need this?

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/client/TServiceClientReceiveInterceptor.java
```diff
@@ -0,0 +1,62 @@
...
+import java.lang.reflect.Method;
+import java.util.Objects;
+import org.apache.skywalking.apm.agent.core.context.ContextManager;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+
+/**
+ * TAsyncClient will create a TServiceClient for receiving the response. So splitting the this method,
+ * TAsyncClient#receiveBase(...), into here for more efficiency.
```
wu-sheng, line 30:
```suggestion
 * TAsyncClient creates a TServiceClient for receiving the response. So splitting this method,
 * TAsyncClient#receiveBase(...), into here for more efficiency.
```

I am not sure what `So splitting this method, TAsyncClient#receiveBase(...), into here for more efficiency.` this sentence means.

wu-sheng:
@kezhenxu94 Any new concern?

kezhenxu94:
> @kezhenxu94 Any new concern?

Some of the review comments are not addressed, I'm checking with @dmsolr by DM

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/TBaseAsyncProcessorInterceptor.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+import org.apache.skywalking.apm.plugin.thrift.wrapper.AsyncContext;
+import org.apache.skywalking.apm.plugin.thrift.wrapper.ServerInProtocolWrapper;
+import org.apache.thrift.AsyncProcessFunction;
+import org.apache.thrift.TBaseAsyncProcessor;
+import org.apache.thrift.protocol.TProtocol;
+import org.apache.thrift.server.AbstractNonblockingServer;
+
+/**
+ * To transparent the AsyncProcessFunction for getting arguments of method.
```
kezhenxu94, line 36: I guess `transparent` should be `transform` or `wrap`, `transparent` is an adjective, same as the other places where `transparent` appeared

kezhenxu94 (approved):
Looks good generally. It doesn't covers some corner cases as the review comments, but it works generally in most cases, I'll leave the corner cases to the community users who really encounter issues in reality

kezhenxu94 (approved):

PR #5648, opened 2020-10-11
Provide meter agent test tool support

mrproliu:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

___
- Provide meter agent test tool support and implementation of the toolkit.


docs/en/guides/Plugin-test.md
```diff
@@ -294,6 +294,48 @@ The verify description for SegmentRef
 | networkAddress | The peer value of parent exit span.
 | refType | Ref type, options, CrossProcess or CrossThread.
 
+**Meter verify description format**
+```yml
+meterItems:
+-
+  serviceName: SERVICE_NAME(string)
+  meterSize: SEGMENT_SIZE(int)
```
wu-sheng, line 302:
```suggestion
  meterSize: METER_SIZE(int)
```

wu-sheng: What is the size? Do you mean how many names of meter?

docs/en/guides/Plugin-test.md
```diff
@@ -294,6 +294,48 @@ The verify description for SegmentRef
...
+**Meter data verify description format**
+
+```yml
+    meterId: 
+        name: NAME(string)
+        tags:
+        - {name: TAG_NAME(string), value: TAG_VALUE(string)}
+    singleValue: SINGLE_VALUE(double)
+    histogram:
+    - {bucket: HISTOGRAM_BUCKET(double), count: HISTOGRAM_BUCKET_COUNT(long)}
```
wu-sheng, line 322: How could you know the count? the values are generated in the runtime, I have doubt that it is possible to be accurate.

test/plugin/scenarios/apm-toolkit-trace-scenario/src/main/java/org/apache/skywalking/apm/toolkit/meter/BaseBuilder.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+package org.apache.skywalking.apm.toolkit.meter;
+
+import java.util.Objects;
+
+/**
+ * Help to build the meter
+ */
+public abstract class BaseBuilder<BUILDER extends BaseBuilder, METER extends BaseMeter> {
```
wu-sheng, line 26: Could you use the snapshot for now to replace the duplicated codes?

docs/en/guides/Plugin-test.md
```diff
@@ -294,6 +294,48 @@ The verify description for SegmentRef
...
 
+**Meter verify description format**
+```yml
+meterItems:
+-
+  serviceName: SERVICE_NAME(string)
+  meterSize: SEGMENT_SIZE(int)
+  meters:
+  - ...
+```
```
wu-sheng, line 305: I think you missed this part of the document for this, https://github.com/apache/skywalking/blob/dba60957d9cf55a5bdf8c50eaeb9be76eeb401b9/docs/en/guides/Plugin-test.md#the-example-process-of-writing-expected-data

docs/en/guides/Plugin-test.md
```diff
@@ -485,6 +527,80 @@ SegmentB span list should like following
...
+```yaml
+- meterId:
+    name: test_histogram
+    tags:
+      - {name: hk1, value: hv1}
+  histogram:
+    - {bucket: 0.0, count: ge 0}
+    - {bucket: 1.0, count: ge 0}
+    - {bucket: 5.0, count: ge 0}
+    - {bucket: 10.0, count: ge 0}
```
wu-sheng, line 601: I have a little concern about the practicality. When the codes try to use the histogram to measure the performance, it seems hard to tell which bucket(s) would have value. Could you try to do a bucket check(w/o count) only, and only check at least not all `count`(s) are 0?

@dmsolr What do you think?

mrproliu: Follow the https://github.com/apache/skywalking-agent-test-tool/pull/26, change the histogram meter validation way.

wu-sheng (approved):

wu-sheng (approved):

PR #5661, opened 2020-10-14
Add health checker for cluster management module

### LINKED ISSUE #5633: Add health check for Storage/Cluster management module
In the previous releases, @hanahmily added the health check module and its implementation for the storage modules(H2, MySQL, and ES), we need to add a similar thing for InfluxDB.

Also, for cluster management, we need to add this too. If 
1. the registration fails 
1. can't get the instance list 
1. can't get itself, then
1. get multiple itself(such as multiple 127.0.0.1/0.0.0.0) in the instance list
The cluster management should be listed as an error 

The health check could be verified through CLI command, https://github.com/apache/skywalking-cli#checkhealth

@xbkaishui Are you going to do this?

xbkaishui:

- [ ] Bug fix
- [x] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___
resolved #5633


wu-sheng (changes requested):
From my understanding, `DelegatedHealthChecker` is designed for classes without accessibility to the telemetry API, such as ElasticSearch client is hosted in the `library-client`.

In the cluster case, the telemetry API(`HealthCheckMetrics`) is better to use directly.

xbkaishui:
ok will change

xbkaishui:
done

oap-server/server-cluster-plugin/pom.xml
```diff
@@ -47,5 +47,10 @@
             <artifactId>library-util</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.skywalking</groupId>
+            <artifactId>library-client</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
wu-sheng, line 54: You should remove this dependency totally.

xbkaishui: done

wu-sheng (commented):
Please remove `import org.apache.skywalking.oap.server.library.util.HealthChecker`. As in the module provider, we don't need that.

xbkaishui:
Hi @wu-sheng ,  seems the failed e2e test are caused by unhealthy status, I can't find any error log about this, it seems the health check is not stable,  how to fix this ? 
```
14:03:08.217 INFO  🐳 [docker-compose].processLine (35) - ERROR: for oap2  Container "dcc4c103dba6" is unhealthy.
14:03:08.217 INFO  🐳 [docker-compose].processLine (35) - ERROR: for provider1  Container "dcc4c103dba6" is unhealthy.
```

wu-sheng:
>  seems the failed e2e test are caused by unhealthy status, I can't find any error log about this, it seems the health check is not stable, how to fix this ?

I have rerun, maybe just unstable test.

oap-server/server-cluster-plugin/cluster-consul-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/consul/ConsulCoordinator.java
```diff
@@ -27,19 +27,25 @@
...
 import org.apache.skywalking.oap.server.core.cluster.RemoteInstance;
+import org.apache.skywalking.oap.server.core.cluster.ServiceQueryException;
 import org.apache.skywalking.oap.server.core.cluster.ServiceRegisterException;
 import org.apache.skywalking.oap.server.core.remote.client.Address;
+import org.apache.skywalking.oap.server.library.client.healthcheck.HealthCheckable;
 import org.apache.skywalking.oap.server.library.util.CollectionUtils;
+import org.apache.skywalking.oap.server.library.util.HealthChecker;
 
-public class ConsulCoordinator implements ClusterRegister, ClusterNodesQuery {
+public class ConsulCoordinator implements ClusterRegister, ClusterNodesQuery, HealthCheckable {
```
wu-sheng, line 42: You are still using the ` org.apache.skywalking.oap.server.library.util.HealthChecker` and `org.apache.skywalking.oap.server.library.client.healthcheck.HealthCheckable`.

xbkaishui: done

oap-server/server-cluster-plugin/cluster-consul-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/consul/ConsulCoordinator.java
```diff
@@ -49,22 +56,32 @@ public ConsulCoordinator(ClusterModuleConsulConfig config, Consul client) {
...
+            // Discover only "passing" nodes
+            List<ServiceHealth> nodes = healthClient.getHealthyServiceInstances(serviceName).getResponse();
+            if (CollectionUtils.isNotEmpty(nodes)) {
+                nodes.forEach(node -> {
+                    if (!Strings.isNullOrEmpty(node.getService().getAddress())) {
+                        Address address = new Address(node.getService().getAddress(), node.getService().getPort(), false);
+                        if (address.equals(selfAddress)) {
+                            address.setSelf(true);
+                        }
+                        remoteInstances.add(new RemoteInstance(address));
```
wu-sheng, line 71: If you found `127.0.0.1` or `localhost` exists in the node list(list#size > 1), then the node should be labeled as unhealthy.
FYI @hanahmily @kezhenxu94 Make sense?

wu-sheng: Same thing should be checked in every coordinator implementation. I noted that the users usually set up the cluster mode incorrectly, then, blame SkyWalking insert duplicated data or statistic not accurate.

xbkaishui: got it

xbkaishui: Done, current implementation will ignore coordinator  standalone mode

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/HealthCheckUtil.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ */
+
+package org.apache.skywalking.oap.server.library.util;
+
+import com.google.common.collect.Sets;
+
+import java.util.Set;
+
+public class HealthCheckUtil {
```
wu-sheng, line 25: Please move this class to the same folder of `ClusterModule`, as where it belongs. Then rename it to `OAPNodeChecker`.

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/HealthCheckUtil.java
```diff
@@ -0,0 +1,34 @@
...
+    private static final Set<String> UN_HEALTH_ADDRESS = Sets.newHashSet("127.0.0.1", "localhost");
```
wu-sheng, line 26:
```suggestion
    private static final Set<String> ILLEGAL_NODE_ADDRESS_IN_CLUSTER_MODE = Sets.newHashSet("127.0.0.1", "localhost");
```

xbkaishui: done

oap-server/server-cluster-plugin/cluster-etcd-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/etcd/EtcdCoordinator.java
```diff
@@ -51,6 +59,9 @@
 
     private static final Integer KEY_TTL = 45;
 
+    @Setter
+    private HealthCheckMetrics healthChecker;
```
wu-sheng, line 63: As this is required in the process, please don't use the `setter`. Initial this in the constructor.

oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/cluster/OAPNodeCheckerTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+import static org.hamcrest.CoreMatchers.is;
+
+public class OAPNodeCheckerTest {
+
+    @Test
+    public void hasUnHealthAddressFalse() {
+        Set<String> address = Sets.newHashSet("123.23.4.2");
+        boolean flag = OAPNodeChecker.hasUnHealthAddress(address);
+        Assert.assertThat(flag, is(false));
```
wu-sheng, line 35: There is `Assert.assertFalse()`

oap-server/server-cluster-plugin/cluster-etcd-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/etcd/EtcdCoordinator.java
```diff
@@ -73,15 +83,30 @@ public EtcdCoordinator(ClusterModuleEtcdConfig config, EtcdClient client) {
...
+            if (remoteInstances.size() > 1) {
+                Set<String> remoteAddressSet = remoteInstances.stream().map(remoteInstance ->
+                        remoteInstance.getAddress().getHost()).collect(Collectors.toSet());
+                boolean hasUnHealthAddress = OAPNodeChecker.hasUnHealthAddress(remoteAddressSet);
+                if (hasUnHealthAddress) {
+                    this.healthChecker.unHealth(new ServiceQueryException("found 127.0.0.1 or localhost in cluster mode"));
+                } else {
+                    List<RemoteInstance> selfInstances = remoteInstances.stream().
+                            filter(remoteInstance -> remoteInstance.getAddress().isSelf()).collect(Collectors.toList());
+                    if (CollectionUtils.isNotEmpty(selfInstances) && selfInstances.size() == 1) {
```
wu-sheng, line 98: Why `&& selfInstances.size() == 1`? Are you planning to check duplicate addresses? If so, please move this login into `OAPNodeChecker`, all coordinator implementations should check this too, right?

xbkaishui: Yes, changed

oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/cluster/OAPNodeCheckerTest.java
xbkaishui: done

oap-server/server-cluster-plugin/cluster-etcd-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/etcd/EtcdCoordinator.java
xbkaishui: done

wu-sheng (commented):
Recommend to read `RemoteClientManager#L123`, the duplicated node will be distinct. So, the duplicate instance nodes would not happen in the core.
NOTE, for many coordinators like ZK, if the OAP reboot, there is a short of time, the register information would be duplicated, as the old temp node has not been removed/expired.

xbkaishui:
> Recommend to read `RemoteClientManager#L123`, the duplicated node will be distinct. So, the duplicate instance nodes would not happen in the core.

I am not checking the duplicated node, maybe the method name is not clear , I just checking belows condtions: 
get multiple itself(such as multiple 127.0.0.1/0.0.0.0) in the instance list

wu-sheng:
> get multiple itself(such as multiple 127.0.0.1/0.0.0.0) in the instance list

You have two checking methods, this is just one thing.

xbkaishui:
Sorry, can you explain more, I am not sure what's is wrong

xbkaishui:
> Recommend to read `RemoteClientManager#L123`, the duplicated node will be distinct. So, the duplicate instance nodes would not happen in the core.
> NOTE, for many coordinators like ZK, if the OAP reboot, there is a short of time, the register information would be duplicated, as the old temp node has not been removed/expired.

Done, please help to review again, thanks

wu-sheng:
We begin the 8.2.0 release process. The PR will be held a little longer.

xbkaishui:
ok, got it

xbkaishui:
Hi @wu-sheng , can you help to review again?

wu-sheng:
I am on vacation, will be back on this at Sunday.

xbkaishui:
got it

wu-sheng (approved):
LGTM. @xbkaishui Could you prepare another pull request to enhance the setup document? Talking about using CLI health check command?

wu-sheng (changes requested):
@xbkaishui @kezhenxu94 Please recheck on the e2e, seems this change breaks the tests. I have re-run several times.

xbkaishui:
Ok, the error log is not so clear, how can I run the e2e test in my local env?   may need your help

xbkaishui:
> setup

OK, will do this later, thanks

kezhenxu94:
@xbkaishui this may be the cause

image

xbkaishui:
Thanks, it seems in prepare stage, I can't call getManager().find method, will fix this

xbkaishui:
I found the dependency circle problem, the core module remote manager dependency on cluster module, when called start method

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/cluster/OAPNodeChecker.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Check the remote instance healthiness, set health to false for bellow conditions:
+     * 1.can't get the instance list
+     * 2.can't get itself
+     * 3.check for illegal node in cluster mode such as 127.0.0.1, localhost
+     *
+     * @param remoteInstances all the remote instances from cluster
+     * @return true health false unHealth
+     */
+    public static boolean isHealth(List<RemoteInstance> remoteInstances) {
```
wu-sheng, line 49: Make the return value as an object. then you could get a boolean with the reason.
The reason could include which illegal address is found.

wu-sheng: Then add a new method in `HealthCheckMetrics`, `unHealth(String message)`(note, the message is just for log like exception). Creating exception cause unnecessary resource, and the `HealthQueryService` and `HealthCheckerProvider#notifyAfterCompleted` don't care the exception at all.

wu-sheng (commented):
Generally good, need a little more polish.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/cluster/OAPNodeChecker.java
xbkaishui: Ok, sounds better to me

xbkaishui: done, please help to check

wu-sheng:
You have conflicting, please recheck.

xbkaishui:
rebase has some issues, I have submit a new pr https://github.com/apache/skywalking/pull/5784 
will close this

Issue #5591, opened 2020-09-29
Metrics about latency of consuming queue


wu-sheng:
Hi @apache/skywalking-committers and @EvanLjp

I created this issue to track the discussion originally started here, https://github.com/apache/skywalking/issues/5584#issuecomment-700793580. The question of #5584  actually is asking for metrics of MQ consuming latency.

Here it is my proposal

The only high-performance way is to propagate the sending time(header injecting timestamp) in the extension header, https://github.com/apache/skywalking/blob/master/docs/en/protocols/Skywalking-Cross-Process-Propagation-Headers-Protocol-v3.md#extension-header-item.

If you are going to implement this, we should consider making it optional, such as
- `sw8-x` = `0-mq_send_timestamp` or `-mq_send_timestamp`
- `sw8-x` = `0`

Then, if when the upstream receiving this header, it could be readable by the entry span, the plugin could calculate the latency of this message, add as a new tag as `mq.consum_latency`. Then the OAP could do statistics with a new source.

NOTE, this new field of extension header should be readonly, but wouldn't propagate it because once it is read from the header, it means this message has been consumed.

EvanLjp:
This is enough for me.

wu-sheng:
@EvanLjp I think we still miss part of this feature. Could you complete this?

EvanLjp:
Ok, I would finish it in May.

wu-sheng:
@EvanLjp Do we have expect date to finish this?

EvanLjp:
@wu-sheng I'm worried about whether I can finish this, so I'm going to close this first.

PR #5666, opened 2020-10-14
add transmission latency for MQ case

EvanLjp:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues
Relate to https://github.com/apache/skywalking/issues/5591
___
### Bug fix
- Bug description.

- How to fix?

___


EvanLjp:
@wu-sheng  how about to transfer the send timestamp for every exit span rather than works on the MQ span. It doesn't seem to increase the amount of data being transmitted

For current codes, it is hard to control the transfer only works on Mq exit spans. Because the layer method is split with the create span method. So we cannot know the layer of the exit span in the inject method.

wu-sheng (commented):
I recommend you changing `ContextCarrier` APIs, to add instance methods.
1. `ContextCarrier#ExtensionInjector#injectSendingTimestamp()` for setting the timestamp to propagate.
1. `ContextCarrier#ExtensionExtractor#readSendingTimestamp()` for reading the propagated timestamp.
Then this timestamp could be injected by specific plugins, such as all MQ plugins.

What do you think? If you need, I could help with implementing these APIs.

And, please remember, you should update `documentation` about the protocol and plugin dev, no matter what is changed.

EvanLjp:
> I recommend you changing `ContextCarrier` APIs, to add instance methods.
> 
> 1. `ContextCarrier#ExtensionInjector#injectSendingTimestamp()` for setting the timestamp to propagate.
> 2. `ContextCarrier#ExtensionExtractor#readSendingTimestamp()` for reading the propagated timestamp.
>    Then this timestamp could be injected by specific plugins, such as all MQ plugins.
> 
> What do you think? If you need, I could help with implementing these APIs.
> 
> And, please remember, you should update `documentation` about the protocol and plugin dev, no matter what is changed.

thx for your recommend

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/AbstractSpan.java
```diff
@@ -124,4 +124,10 @@
      * Should skip analysis in the backend.
      */
     void skipAnalysis();
+
+    /**
+     * Get the start time stamp of this span.
+     */
+    long getStartTime();
```
wu-sheng, line 131: This should be removed.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ExtensionInjector.java
```diff
@@ -0,0 +1,39 @@
...
+
+    ExtensionInjector(final ExtensionContext extensionContext) {
+        this.extensionContext = extensionContext;
+    }
+
+    public void injectSendingTimestamp() {
+        extensionContext.setSendingTimestamp(System.currentTimeMillis());
+    }
+
+    public long readSendingTimestamp() {
```
wu-sheng, line 36: `long` -> `Long`, as it could be NULL if don't exist in the in-wire context.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/MultiScopesAnalysisListener.java
```diff
@@ -234,6 +234,12 @@ public void parseExit(SpanObject span, SegmentObject segmentObject) {
 
     private void setPublicAttrs(SourceBuilder sourceBuilder, SpanObject span) {
         long latency = span.getEndTime() - span.getStartTime();
+        for (final KeyStringValuePair keyStringValuePair : span.getTagsList()) {
+            if (keyStringValuePair.getKey().equals(SpanTags.TRANSMISSION_LATENCY)) {
+                latency += Integer.parseInt(keyStringValuePair.getValue());
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
wu-sheng: Why rewrite the latency of this? We never said the latency of span/segment/service including the network when we talked about the server-side latency.

test/plugin/scenarios/kafka-scenario/config/expectedData.yaml
```diff
@@ -156,6 +157,7 @@ segmentItems:
         tags:
           - {key: mq.broker, value: 'kafka-server:9092'}
           - {key: mq.topic, value: test.}
+          - {key: transmission.latency, value: not null}
```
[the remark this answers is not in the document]
EvanLjp: ok

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ExtensionContext.java
```diff
@@ -33,13 +36,20 @@
...
+    private long sendingTimestamp = 0;
+
     /**
      * Serialize this {@link ExtensionContext} to a {@link String}
      *
      * @return the serialization string.
      */
     String serialize() {
-        return skipAnalysis ? "1" : "0";
+        return skipAnalysis ? "1" : "0" + "-" + sendingTimestamp;
```
mrproliu, line 52:
```suggestion
        return (skipAnalysis ? "1" : "0") + "-" + sendingTimestamp;
```

wu-sheng (approved):
LGTM. @EvanLjp at the same time, I think this PR should not close #5591, as you have finished the backend part. Make sense?

EvanLjp:
> LGTM. @EvanLjp at the same time, I think this PR should not close #5591, as you have finished the backend part. Make sense?

yes, I would like to finish the backend part later.

wu-sheng (changes requested):
Seems Kafka tests keep failing, could you recheck?

EvanLjp:
image
@wu-sheng I don't know why to log the same keys in 2 lines.  I debug the code to make sure that only once was passed in.
And MAP<string, string> tags in the span is Impossible to store the same key. I am not very familiar with the test Tool code,  is there anyone to help this ?

wu-sheng:
Take a look at `AbstractTracingSpan#tags`. It is never being a Map, it is a list for much better performance from day 1 of SkyWalking :)

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/tag/Tags.java
```diff
@@ -77,7 +77,11 @@ private Tags() {
      */
     public static final StringTag MQ_TOPIC = new StringTag(9, "mq.topic");
 
-    public static final StringTag TRANSMISSION_LATENCY = new StringTag(15, "transmission.latency");
+    /**
+     * The latency of transmission. If there were more than one downstream endpoints, multiple tags will be recorded,
```
wu-sheng, line 81:
```suggestion
     * The latency of transmission. When there are more than one downstream parent/segment-ref(s), multiple tags will be recorded,
```

test/plugin/scenarios/pulsar-scenario/config/expectedData.yaml
```diff
@@ -83,9 +83,9 @@ segmentItems:
       spanType: Entry
       peer: ''
       tags:
-      - {key: mq.broker, value: not null}
+        - {key: transmission.latency, value: not null}
+        - {key: mq.broker, value: not null}
```
wu-sheng, line 87: Is this format correct?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ExtensionContext.java
```diff
@@ -28,18 +33,30 @@
...
     /**
      * Serialize this {@link ExtensionContext} to a {@link String}
      *
      * @return the serialization string.
      */
     String serialize() {
-        return skipAnalysis ? "1" : "0";
+        String res = skipAnalysis ? "1" : "0";
+        if (sendingTimestamp != 0) {
+            res += "-" + sendingTimestamp;
```
kezhenxu94, line 57: Maybe we should always serialize this part even `if sendingTimestamp != 0`? The extension fields are order-sensitive, we always need a real value or placeholder, otherwise, if the downstream agent plugin don't have this field but have their own extended fields(now for now, though), it will be trouble, WDYT? @wu-sheng

wu-sheng: Yes, my previous comment is about don't serialize the 0, but - should be included. 
It seems to be misunderstood.

EvanLjp: I have similar concerns about this, so the zero is a placeholder before. WDYT? @wu-sheng

wu-sheng: I don't like `0`, as `0` is a real value, you are going to represent no value. So you could use `0-` as timestamp doesn't exist.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ExtensionContext.java
```diff
@@ -69,14 +65,12 @@ void deserialize(String value) {
         final String[] extensionParts = value.split("-");
         // All parts of the extension header are optional.
         // only try to read it when it exist.
-        if (extensionParts.length > 0) {
+        if (extensionParts.length == 2) {
```
wu-sheng, line 68: Again, don't use `2`.

EvanLjp: ok

EvanLjp: Because all parameters are passing rather than optional fields, the length of extensionParts is fixed.

now, the passing data maybe "0-" or "0-1602743904804", length of extensionParts must be 2.
in the feature, if add another field, the propagate data may be "0--" or "0-xxx-xxx". And the length of extensionParts must be 3.

I think logic greater than 0 is redundant

wu-sheng: Don't require `must`, that is the key. If the new agent propagates `0-xxx-xxx`, I hope the older agent still works, that is why I wrote this part as `>0`, and ask you to do `>1`.

EvanLjp: I finally understand the key: the compatibility.

wu-sheng: Yes, this kind of PR shows me the case is, the extension header may be used more in the future.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ExtensionContext.java
```diff
@@ -49,11 +64,23 @@ void deserialize(String value) {
         if (StringUtil.isEmpty(value)) {
             return;
         }
-        final String[] extensionParts = value.split("-");
+        String[] extensionParts = value.replaceAll("-", "- ").split("-");
```
wu-sheng, line 67: Please inject ` `(whitespace), rather than this strange `replaceAll`.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ExtensionContext.java
```diff
@@ -28,18 +33,28 @@
...
      * Tracing Mode. If true means represents all spans generated in this context should skip analysis.
      */
     private boolean skipAnalysis;
 
+    /**
+     * The sending timestamp of the exit span.
+     */
+    @Getter
+    @Setter
+    private long sendingTimestamp;
```
wu-sheng, line 47: Please use `Long` to separate real 0 and NULL.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ExtensionContext.java
EvanLjp: I may have misunderstood what you meant, thinking you meant to pass as few characters as possible. fix it later.

wu-sheng: The size is not a concern, as we have injected a much bigger `sw8` header.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ExtensionContext.java
```diff
@@ -49,11 +73,23 @@ void deserialize(String value) {
         if (StringUtil.isEmpty(value)) {
             return;
         }
-        final String[] extensionParts = value.split("-");
+        String[] extensionParts = value.split(SEPARATOR);
```
wu-sheng, line 76: I think you misunderstand me, I mean no value should be represented by ` `, the separator should always be `-`.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ExtensionContext.java
```diff
@@ -100,11 +142,11 @@ public boolean equals(Object o) {
         if (o == null || getClass() != o.getClass())
             return false;
         ExtensionContext that = (ExtensionContext) o;
-        return skipAnalysis == that.skipAnalysis;
+        return skipAnalysis == that.skipAnalysis && sendingTimestamp == that.sendingTimestamp;
```
wu-sheng, line 145: I think this equal is incorrect, as it is an object for now.

docs/en/protocols/Skywalking-Cross-Process-Propagation-Headers-Protocol-v3.md
```diff
@@ -43,5 +43,7 @@ The current value includes fields.
 1. Tracing Mode. empty, 0 or 1. empty or 0 is default. 1 represents all spans generated in this context should skip analysis,
 `spanObject#skipAnalysis=true`. This context should be propagated to upstream in the default, unless it is changed in the 
 tracing process.
+2. The Exit span sending timestamp.  Space is the default. If the downstream timestamp was not space, the latency cost between downstream and upstream services
+   should be tagged in the entry spans of the upstream services.
```
wu-sheng, line 47: 2. The timestamp of sending at the client side. This is used in async RPC such as MQ. Once it is set, the consumer side would calculate the latency between sending and receiving, and tag it in the span by using key `transmission.latency ` automatically.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ExtensionContext.java
```diff
@@ -100,11 +143,11 @@ public boolean equals(Object o) {
         if (o == null || getClass() != o.getClass())
             return false;
         ExtensionContext that = (ExtensionContext) o;
-        return skipAnalysis == that.skipAnalysis;
+        return skipAnalysis == that.skipAnalysis && sendingTimestamp.equals(that.sendingTimestamp);
```
kezhenxu94, line 146: `sendingTimestamp` may be `null`?

```suggestion
        return skipAnalysis == that.skipAnalysis && Objects.equals(sendingTimestamp, that.sendingTimestamp);
```

docs/en/protocols/Skywalking-Cross-Process-Propagation-Headers-Protocol-v3.md
```diff
@@ -43,5 +43,6 @@ The current value includes fields.
 1. Tracing Mode. empty, 0 or 1. empty or 0 is default. 1 represents all spans generated in this context should skip analysis,
 `spanObject#skipAnalysis=true`. This context should be propagated to upstream in the default, unless it is changed in the 
 tracing process.
+2. The timestamp of sending at the client side. This is used in async RPC such as MQ. Once it is set, the consumer side would calculate the latency between sending and receiving, and tag it in the span by using key transmission.latency automatically.
```
wu-sheng, line 46:
```suggestion
2. The timestamp of sending at the client-side. This is used in async RPC such as MQ. Once it is set, the consumer side would calculate the latency between sending and receiving, and tag the latency in the span by using key `transmission.latency` automatically.
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ExtensionInjector.java
kezhenxu94: This comment is not addressed yet, I mark it as unresolved. @EvanLjp please only mark the conversation as "resolved" after you have actually resolved it or have reasons not to resolve, I don't see either.

Because `sendingTimestamp` is `Long`, returning it from `public long readSendingTimestamp()` will lead to an auto-unboxing and `NullPointerException`

wu-sheng (approved):
LGTM. Wait for your backend enhancement to measure the latency of MQ consumption.

wu-sheng:
We should consider adding a metric to represent `avg latency of transmission` at the endpoint level, @EvanLjp does this make sense to your scenario?

wu-sheng (commented):
> Error:  Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.0:check (validate) on project apm-agent-core: You have 1 Checkstyle violation. -> [Help 1]

@EvanLjp I accepted @kezhenxu94 's change suggestion about the equal NPE. I am not sure which change causes the style issue. Please recheck and fix.

kezhenxu94:
@EvanLjp this one breaks the CI.

image

And please also address this comment

image

EvanLjp:
@wu-sheng @kezhenxu94 I am sorry for the bugs in the code details, and I will pay more attention to the code quality in the future.

wu-sheng (approved):

PR #5675, opened 2020-10-15
Set up ALS E2E test

kezhenxu94:
Also:

- Remove istio-mixer-ci.yaml
- Don't run test on git tags
- Not run plugin tests on push event

Now I check the services and topology by name prefix, such as `ratings-v1.*`, so it should also work when refactoring the list pods API with informer, as @hanahmily suggested


wu-sheng (approved):
Cool, we finally have this :)

wu-sheng (approved):

arugal:
A year ago, I try the kubernetes cluster E2E test, but failed :smile:

PR #5694, opened 2020-10-20
Add mqtt plugin

GerryYuan:

- [ ] Bug fix
- [ ] New feature provided
- [ ] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___


apm-sniffer/apm-sdk-plugin/mqtt-3.x-plugin/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>apm-mqtt-3.x-plugin</artifactId>
+    <name>mqtt-3.x-plugin</name>
+    <packaging>jar</packaging>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <checkstyle.skip>true</checkstyle.skip>
```
kezhenxu94, line 18: Don't skip this

apm-sniffer/apm-sdk-plugin/mqtt-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mqtt/v3/MqttEnhanceRequiredInfo.java
```diff
@@ -0,0 +1,70 @@
...
+    
+    private String brokerServers;
+    
+    private String topics;
+    
+    private String qos;
+    
+    private long startTime;
+    
+    public void setBrokerServers(String[] brokerServers) {
```
kezhenxu94, line 38: Use `lombok.Getter/Setter` annotations for this class, please

kezhenxu94 (commented):
Please:

- remove all the author information from the source codes, like `Created by yuanguohua on 2020/8/21 10:02`
- plugin test is required for every plugin, refer to https://github.com/apache/skywalking/blob/master/docs/en/guides/Plugin-test.md
- documentations should be updated as well, including the [supported list](https://github.com/apache/skywalking/blob/master/docs/en/setup/service-agent/java-agent/Supported-list.md), https://github.com/apache/skywalking/blob/master/docs/en/setup/service-agent/java-agent/Plugin-list.md
- add a logo/icon to the [UI repo](https://github.com/apache/skywalking-rocketbot-ui)  ([example](https://github.com/apache/skywalking-rocketbot-ui/pull/323))

wu-sheng (commented):
MQTT is an OASIS standard, as this, it means there are implementations for supporting this spec. https://mqtt.org/software/

What is this plugin targeting? Whether all implementations of MQTT would be traced if we have this plugin? 

Also, note, there are MQs like ActiveMQ and RabbitMQ, which we have plugins already. What will happen we have both plugins in the runtime?

apm-sniffer/apm-sdk-plugin/mqtt-3.x-plugin/pom.xml
GerryYuan: OK, I got it

kezhenxu94:
@GerryYuan can you please address [@wu-sheng 's concerns](https://github.com/apache/skywalking/pull/5694#pullrequestreview-512372706)?

GerryYuan:
> MQTT is an OASIS standard, as this, it means there are implementations for supporting this spec. https://mqtt.org/software/
> 
> What is this plugin targeting? Whether all implementations of MQTT would be traced if we have this plugin?
> 
> Also, note, there are MQs like ActiveMQ and RabbitMQ, which we have plugins already. What will happen we have both plugins in the runtime?

1. The goal of the plug-in is to provide the corresponding tracking for the implementers who are using the mqtt protocol. If we have this plug-in, the plug-ins that communicate with the mqtt broker through eclipse PAHO will be tracked. For example, eclipse PAHO client publish message to broker and client consumer message, and then process the user's business logic can be traced, as shown in the following figure:
image
image

2. If we have other MQ plug-ins and add mqtt plug-ins to run at the same time, there will be no problem. They are independent of each other, as shown in the following figure:
image

apm-sniffer/apm-sdk-plugin/mqtt-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mqtt/v3/define/MqttConsumerInstrumentation.java
```diff
@@ -0,0 +1,68 @@
...
+                public boolean isOverrideArgs() {
+                    return false;
+                }
+            }
+        };
+    }
+
+    @Override
+    protected ClassMatch enhanceClass() {
+        return byHierarchyMatch(new String[] {ENHANCE_CLASS});
```
kezhenxu94, line 66: Redundant array creation for calling varargs method

apm-sniffer/apm-sdk-plugin/mqtt-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mqtt/v3/MqttNetworkInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+
+public class MqttNetworkInterceptor implements InstanceMethodsAroundInterceptor {
+
+    @Override
+    public void beforeMethod(EnhancedInstance enhancedInstance, Method method, Object[] allArguments,
+                             Class<?>[] classes, MethodInterceptResult methodInterceptResult) throws Throwable {
+        NetworkModule[] networkModules = (NetworkModule[]) allArguments[0];
+        MqttEnhanceRequiredInfo requiredInfo = new MqttEnhanceRequiredInfo();
+        String[] serverURIs = Arrays.stream(networkModules)
+                                    .map(networkModule -> networkModule.getServerURI())
```
kezhenxu94, line 38: Lambda can be replaced with method reference

apm-sniffer/apm-sdk-plugin/mqtt-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mqtt/v3/MqttNetworkInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public Object afterMethod(EnhancedInstance enhancedInstance, Method method, Object[] objects, Class<?>[] classes,
+                              Object o) throws Throwable {
+        return o;
+    }
+
+    @Override
+    public void handleMethodException(EnhancedInstance enhancedInstance, Method method, Object[] objects,
+                                      Class<?>[] classes, Throwable throwable) {
+        ContextManager.activeSpan().errorOccurred().log(throwable);
```
kezhenxu94, line 55: Don't need `errorOccurred()` now, it's included in the `log` method

```suggestion
        ContextManager.activeSpan().log(throwable);
```

apm-sniffer/apm-sdk-plugin/mqtt-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mqtt/v3/MqttProducerInterceptor.java
```diff
@@ -0,0 +1,75 @@
...
+        if (ContextManager.isActive()) {
+            ContextManager.stopSpan();
+        }
+        return o;
+    }
+
+    @Override
+    public void handleMethodException(EnhancedInstance enhancedInstance, Method method, Object[] objects,
+                                      Class<?>[] classes, Throwable throwable) {
+        ContextManager.activeSpan().errorOccurred().log(throwable);
```
kezhenxu94, line 73:
```suggestion
        ContextManager.activeSpan().log(throwable);
```

kezhenxu94 (commented):
Looks good to me in code level, just some nits, waiting for other reviews' confirmation

apm-sniffer/apm-sdk-plugin/mqtt-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mqtt/v3/MqttNetworkInterceptor.java
GerryYuan: Received, modified

apm-sniffer/apm-sdk-plugin/mqtt-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mqtt/v3/define/MqttConsumerInstrumentation.java
GerryYuan: How to do that?
eg:
org.apache.skywalking.apm.plugin.kafka.define.CallbackInstrumentation#enhanceClass
org.apache.skywalking.apm.plugin.hystrix.v1.define.HystrixCommandInstrumentation#enhanceClass
....

kezhenxu94:
```suggestion
        return byHierarchyMatch(ENHANCE_CLASS);
```

`byHierarchyMatch` originally only accepts an array as argument(like what you do now and your examples), now we refactor it to accept variable arguments so that you don't need to create an array explicitly, anyway, it's not a big deal :)

GerryYuan:
@kezhenxu94 
I need some help, 'mqtt plugin' always fails in testing 'plugin test'. ` expectedData.yaml `How to write it? I will take out the actual scope of the local test and configure it in the` expectedData.yaml `File. How can I pass' skywalking '- validator.jar `Verification, thank you!
image

apm-sniffer/apm-sdk-plugin/mqtt-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mqtt/v3/define/MqttConsumerInstrumentation.java
GerryYuan: Thank you for your review. I'll change it right away, which will help me write good code

wu-sheng:
To my context of the point of this plugin, why there is another node in the topology when we have both plugins?

wu-sheng:
To my context of the point of this plugin, why there is another node in the topology when we have both plugins?

GerryYuan:
> To my context of the point of this plugin, why there is another node in the topology when we have both plugins?

Thank you for your question, but I don't understand what the other node is. I am answering "what is this plugin targeting? What all implementation of mqtt would be traced if we have this "Plugin?" my simulation scenario is that the user sends an HTTP request. After receiving the request, the server sends a message to rabbitmq in the current thread, and then sends a message to mqtt broker at the same time. Therefore, there will be a rabbitmq node and an mqtt broker node in the topology diagram. I don't know if it's your question.

wu-sheng:
Yes, the question is about that,but my concern is more about, could rabbitmq and mqtt share one request? Sorry, I am far from understanding what mqtt did.

GerryYuan:
> Yes, the question is about that,but my concern is more about, could rabbitmq and mqtt share one request? Sorry, I am far from understanding what mqtt did.

Rabbitmq itself also supports the mqtt protocol, but it needs to be introduced as a plug-in, which is relatively complex. At present, we have independently deployed the mqtt broker for link tracking. In the IOT business scenario, this deployment method is more common. Of course, if rabbitmq installs the mqtt plug-in, it will not have the problem of tracking links now. Because our mqtt plugin enhances and tracks the mqtt client (eclipse PAHO) of Java, there is no coexistence problem.

kezhenxu94:
> @kezhenxu94
> I need some help, 'mqtt plugin' always fails in testing 'plugin test'. `expectedData.yaml`How to write it? I will take out the actual scope of the local test and configure it in the`expectedData.yaml`File. How can I pass' skywalking '- validator.jar `Verification, thank you!
> image

I've checked your `expectedData.yaml`, it's basically correct, but the `Mqtt/skywalking-agent/Consumer` span may not appear because the request finished before the consumer consumes the message, please refer to other similar test case

GerryYuan:
@wu-sheng @kezhenxu94 I have some problems and need your help. MQTT Because the protocol header is fixed, it is impossible to pass the contextcarrier header, and the consumer cannot get it. As a result, the whole call chain producer and consumer are disconnected. I have come up with a solution. I put the contextcarrier header in the mqtt payload, and then get the set from the consumer to the carrieritem. However, there is a problem. If the user can install the If our mqtt plugin is installed, but the consumer side is another service. If the mqtt plugin is not installed, the message will be changed by string instead of the message delivered by the user. Is there any other way to implement the header transmission of contextcarrier.

kezhenxu94:
> @wu-sheng @kezhenxu94 I have some problems and need your help. MQTT Because the protocol header is fixed,

What does it mean? You can not add header items into the protocol header, right?

If that's the case, I don't think modifying the payload is a good idea, that requires the agent to be installed in both producer and consumer side. 

But it sounds weird that we cannot extend the headers, can you paste some links to the docs about this?

GerryYuan:
> > @wu-sheng @kezhenxu94 I have some problems and need your help. MQTT Because the protocol header is fixed,
> 
> What does it mean? You can not add header items into the protocol header, right?
> 
> If that's the case, I don't think modifying the payload is a good idea, that requires the agent to be installed in both producer and consumer side. In such case, the plugin must be located at the `optional-plugins` folder.
> 
> But it sounds weird that we cannot extend the headers, can you paste some links to the docs about this?

Okay, I also think this is a better idea. Next I will move the plugin to ʻoptional-plugins` and improve the documentation

wu-sheng:
> If the user can install the If our mqtt plugin is installed, but the consumer side is another service. 

If this is happening, this plugin should never exist in the official plugin, even it is optional. The reason for this is, in the real production environment, it is nearly impossible to make sure the plugin has been installed on both sides always. Also, this makes the agent's features, including uninstallable and harmless, are gone.

We need a deep discussion about this.

GerryYuan:
> If this is happening, this plugin should never exist in the official plugin, even it is optional. The reason for this is, in the real production environment, it is nearly impossible to make sure the plugin has been installed on both sides always. Also, this makes the agent's features, including uninstallable and harmless, are gone.

There are currently 2 schemes:
One is an optional plug-in that supports the entire mqtt tracing function, but the premise is that the agent must be installed on both sides
Another solution is that the user is not required to install mqtt-plugin on both sides, but the tracing of producer and consumer is disconnected

Is there a better way to solve this problem?
 
Now, the Mqtt.md file is described as follows：
## Support mqtt3.X
Here is an optional plugin `apm-mqtt-3.x-plugin`

## Introduce
The Mqtt header cannot be expanded, and the Agent access requires attention
### Notice:
- Agent consumer and producer must be installed at the same time!!!
- Agent installation sequence, consumer first, producer second, is very important!!!
- Copy `apm-mqtt-3.x-plugin.jar` to `agent/plugins`, restarting the `agent` can effect the plugin.

wu-sheng:
Either breaking the tracing or the not very practical is not the official way.  From my perspective, I wouldn't support this kind of plugin exists in the official release, it is quite misleading the users.

This thing happened before, I hope our principle is consistent. Read #1543, it takes over 2 years to find a lower risk way.

GerryYuan:
> > If the user can install the If our mqtt plugin is installed, but the consumer side is another service.
> 
> If this is happening, this plugin should never exist in the official plugin, even it is optional. The reason for this is, in the real production environment, it is nearly impossible to make sure the plugin has been installed on both sides always. Also, this makes the agent's features, including uninstallable and harmless, are gone.
> 
> We need a deep discussion about this.

GerryYuan:
> Either breaking the tracing or the not very practical is not the official way. From my perspective, I wouldn't support this kind of plugin exists in the official release, it is quite misleading the users.
> 
> This thing happened before, I hope our principle is consistent. Read #1543, it takes over 2 years to find a lower risk way.

Understand, but is it possible? We provide users with trcing at one end, because the common application scenario of the MQTT protocol is on IOT. At this time, our terminals are usually IOT devices, so in this scenario we provide cloud mqtt Whether monitoring is more valuable, we can monitor the trcing of the data received by the entire cloud, and the trcing of the data sent to the terminal. This is my opinion, but if you feel that mqtt-plugin does not need to exist, it is acceptable.

PR #5697, opened 2020-10-20
Add change logs for 8.2.0

wu-sheng:
@hanahmily @ascrutae @xbkaishui If you want to add new features, please update the changelogs.

PR #5699, opened 2020-10-20
Add syncBulkActions config to avoid the large amount of metrics data written ES in a single sync bulk request.

Ax1an:

- [ ] Bug fix
- [ ] New feature provided
- [x] Improve performance

- Related issues

___
### Bug fix
- Bug description.

- How to fix?

___

   Add syncBulkActions config to avoid the large amount of metrics data written ES in a single bulk request.

image
image

### DIFF
--- a/changes/changes-8.2.0.md
+++ b/changes/changes-8.2.0.md
@@ -87,6 +87,7 @@
 * Fix end time bug in the query process.
 * Fix `Exporter INCREMENT mode` is not working.
 * Fix an error when executing startup.bat when the log directory exists
+* Add syncBulkActions config to avoid the large amount of metrics data written ES in a single sync bulk request.
 
 #### UI
 * Add browser dashboard.
--- a/changes/changes-8.2.0.md
+++ b/changes/changes-8.2.0.md
@@ -87,7 +87,7 @@
 * Fix end time bug in the query process.
 * Fix `Exporter INCREMENT mode` is not working.
 * Fix an error when executing startup.bat when the log directory exists
-* Add syncBulkActions config to avoid the large amount of metrics data written ES in a single sync bulk request.
+* Add syncBulkActions configuration to set up the batch size of the metrics persistent.
 
 #### UI
 * Add browser dashboard.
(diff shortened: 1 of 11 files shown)


zifeihan (commented):
good.

wu-sheng:
Follow #5697 to update two change log files.

wu-sheng:
If you plan to merge this for now.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java
```diff
@@ -52,6 +52,7 @@
     private int superDatasetIndexShardsFactor = 5;
     private int indexRefreshInterval = 2;
     private int bulkActions = 2000;
+    private int syncBulkActions = 10000;
```
wu-sheng, line 55: Why put this in 10k? Is there any testing, which could prove this provides better performance?

Ax1an: In our production environment, it's 100k. However, considering that most people try skywalking, the machine configuration is not so high, and this is a configurable item, so the initial value is set to 10k.

changes/changes-8.2.0.md
```diff
@@ -87,6 +87,7 @@
 * Fix end time bug in the query process.
 * Fix `Exporter INCREMENT mode` is not working.
 * Fix an error when executing startup.bat when the log directory exists
+* Add syncBulkActions config to avoid the large amount of metrics data written ES in a single sync bulk request.
```
wu-sheng, line 90: I think this description is not correct, you just separate the config, can't avoid that thing

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java
wu-sheng: I am a little concern about the size, whether it is too big for a normal user.

changes/changes-8.2.0.md
wu-sheng: Could change this to 

> Add syncBulkActions configuration to set up the batch size of the trace/log persistent.

Ax1an: Done.

wu-sheng (approved):

PR #5722, opened 2020-10-25
Improve Kubernetes service registry for ALS analysis

### LINKED ISSUE #5706: Improve ALS k8s analysis

- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement

___

The current implementation of envoy ALS k8s analysis is based on the hierarchy, pod -> statefulset -> deployment, statefulset, or others. It's freaky and different from the istio kubernetes registry. 

The new path is pod -> endpoint -> service, and we should leverage informer API instead of raw kubernetes API.

kezhenxu94:
### Improve Kubernetes service registry for ALS analysis

The current implementation of Envoy ALS K8S analysis is based on the hierarchy, pod -> StatefulSet -> deployment, StatefulSet, or others. It's freaky and different from the Istio Kubernetes registry. And the service name pattern changed in recent Kubernetes versions, which generates weird service names in the topology.

The new path is pod -> endpoint -> service, and we should leverage Informer API instead of raw Kubernetes API.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #5706.
- [x] Tests(E2E) are adjusted to verify the improvement.


CHANGES.md
```diff
@@ -9,6 +9,7 @@ Release Notes.
 #### Java Agent
 
 #### OAP-Backend
+* Improve Kubernetes service registry fo ALS analysis.
```
wu-sheng, line 12: fo? Typo?

kezhenxu94: > fo? Typo?

🤣 definitely, fixed

kezhenxu94:
FYI, @wu-sheng @hanahmily What it differs from the previous version is that the different versions of a same service are now collapsed into one service, e.g. in the previous version, reviews-v1, reviews-v2, reviews-v3 are 3 different services, but in this version, they are all collapsed into `reviews`, as a single service, with different version `tag`, (labels: `version=v1`, `version=v2`, etc.)

wu-sheng (commented):
Like the idea. Wait for @hanahmily code level review.

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/K8SServiceRegistry.java
```diff
@@ -0,0 +1,217 @@
...
+                service.setServiceInstanceName(String.format("%s.%s", podMeta.getName(), podMeta.getNamespace()));
+                service.setTags(transformLabelsToTags(podMeta.getLabels()));
+            }
+        );
+    }
+
+    private void addEndpoints(final V1Endpoints endpoints) {
+        log.debug("Adding endpoints {}", endpoints);
+
+        final String serviceName = requireNonNull(endpoints.getMetadata()).getName();
```
hanahmily, line 173: We had better introduce a service name rule to format it. For instance, if some users want to append a version to the service name, they have to issue an expression, like `${service.name}-${pod.labels.version}`

If they deploy `bookinfo`, the service name of the service productpage might be `productpage-v1`, `producatpage-v2` and etc.

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/K8SServiceRegistry.java
```diff
@@ -0,0 +1,217 @@
...
+        final ApiClient apiClient = Config.defaultClient();
+        apiClient.setHttpClient(apiClient.getHttpClient()
+                                         .newBuilder()
+                                         .readTimeout(0, TimeUnit.SECONDS)
+                                         .build());
+
+        final CoreV1Api coreV1Api = new CoreV1Api(apiClient);
+        final SharedInformerFactory factory = new SharedInformerFactory(executor);
+
+        listenEndpointsEvents(coreV1Api, factory);
```
hanahmily, line 73: [Endpoint slice](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/) resources should be listened to either.

hanahmily (changes requested):

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/K8SServiceRegistry.java
wu-sheng: I think the key is providing the format expression. Defaut should be only service name with version, correct?

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/K8SServiceRegistry.java
kezhenxu94: Even the [latest kubernetes-client](https://mvnrepository.com/artifact/io.kubernetes/client-java/10.0.0)(10.0.0, 27th, Oct, 2020) doesn't support to listen to the EndpointSlice events. There is no such API to do this 😢 Let's postponed this resource until the newer version can do this.

wu-sheng: If no listener, are you reading the data periodically like the old way?

kezhenxu94: > If no listener, are you reading the data periodically like the old way?

Not now, as the doc says, EndpointSlice is to improve the performance of Endpoints, so I think it's OK to just ignore this kind of event for now, (p.s. EndpointSlice is still in beta now)

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/K8SServiceRegistry.java
kezhenxu94: Done. I’ll add some more tests of this later. Please review again @hanahmily , thanks

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/K8SServiceRegistry.java
hanahmily: If java client doesn't support it right now, feel free to leave it alone. But we should mention it in our document and leave `todo` market in codes.

wu-sheng: How about submitting a request to k8s java client repo about listen to the EndpointSlice ?

wu-sheng: They may need a scenario about which java codes need this.

kezhenxu94:
All review comments should be addressed, if I missed anything, please let me know

hanahmily (approved):
LGTM

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/K8SServiceRegistry.java
kezhenxu94: > How about submitting a request to k8s java client repo about listen to the EndpointSlice ?

I'm thinking that they don't support it only because EndpointSlice is still in beta(after stabilization, they will), and the methods to list/watch that resources are unstable now, (e.g. need the `apiGroup`, `version`, which are changing for an alpha/beta feature).

IMO, ignoring EndpointSlice for now is safe because the docs says

> Although the EndpointSlice API is providing a newer and more scalable alternative to the Endpoints API, the Endpoints API will continue to be considered generally available and stable.

kezhenxu94: We can revamp this after the EndpointSlice API is stabilised

wu-sheng: My point is only creating an issue to track this todo as backlog.

PR #5739, opened 2020-10-27
support oal list includes and excludes & add tags to some source

EvanLjp:
1. enhance oal filter to support includes or excludes.
2. add tags in oal sources


EvanLjp:
the manual tests have been already passed. If there are autotests for OAL, I would be happy to add them.

wu-sheng (commented):
I am no vocation, so say in simple direction first.
I don't think we should config which tags are need, as this is very low resource cost. Do you have any reason to do so?
Then tags should be only relate to entry span for service, instance, and endpoint, right? And which are for relationship?

EvanLjp:
> And which are for relationship?

The reason I added tags is because of cost, If this cost can be ignored I will remove the parts.
The relationship should not add tags  , u are right.

wu-sheng:
Tags in the source level is low cost fields, due to never serialization and propagation, or persistent. 
If the relationship has no tags, please make sure the documents of sources are clear about this.

EvanLjp:
> Tags in the source level is low cost fields, due to never serialization and propagation, or persistent.
> If the relationship has no tags, please make sure the documents of sources are clear about this.

ok

docs/en/concepts-and-designs/oal.md
```diff
@@ -34,7 +34,7 @@ Read [Scope Definitions](scope-definitions.md), you can find all existing Scopes
 Use filter to build the conditions for the value of fields, by using field name and expression. 
 
 The expressions support to link by `and`, `or` and `(...)`. 
-The OPs support `==`, `!=`, `>`, `<`, `>=`, `<=`, `in [...]` ,`like %...`, `like ...%` and `like %...%`, with type detection based of field type. Trigger compile
+The OPs support `==`, `!=`, `>`, `<`, `>=`, `<=`, `in [...]` ,`like %...`, `like ...%` , `like %...%` , `inclueds` and `excludes`, with type detection based of field type. Trigger compile
```
wu-sheng, line 37: Hi, there is a word chosen game here. :) Take a look about this, https://ell.stackexchange.com/questions/47706/using-contain-vs-include-vs-consist-of-appropriately#:~:text=The%20distinction%20between%20'includes'%20and,container%2C%20and%20contains%20the%20hats.&text=Inside%20the%20box%20is%20a,and%20consists%20of)%20the%20hats.

And check the JDK API. For a list, we should use `contain`. And don't use `contains`, if you look at the OAL scripts, `tags contains` is wrong in English grammar.

wu-sheng: Do you face trouble if use `not contain` as an operator?

EvanLjp: yes,  some segments will not have tags sometimes. So I provided excludes solving this case.

EvanLjp: I actually first using Contains rather than includes. but not contains have 2 words. so I choose excludes and includes as last.

wu-sheng: `not contain` seems very clear, in SQL, you have `not in`, right?

EvanLjp: ok, I will change the words to contain and not contain now.

wu-sheng (approved):

kezhenxu94 (approved):

wu-sheng:
Please hold this PR a little while, @EvanLjp said, he wants to do another round check, as we don't have the e2e for this OAL grammar, only UTs. @kezhenxu94 Any suggestion besides manual tests?

EvanLjp:
test is passed

kezhenxu94:
> test is passed

You didn’t wait for the image to upload successfully , we can’t see the image

wu-sheng:
Can't see the image either, but I think the previous commit passed, https://github.com/apache/skywalking/pull/5739/checks?sha=c864f057e58f97dd36935656a1382b61df2733dd.

wu-sheng (approved):

kezhenxu94 (approved):

EvanLjp:
image
@wu-sheng @kezhenxu94

PR #5754, opened 2020-10-29
fix transaction too large error when use TiDB as storage

### LINKED ISSUE #5749: Transaction is too large when use TiDB as storage

- [X] Question or discussion
- [ ] Bug
- [ ] Requirement

___
### Question
I use the TIDB as the storage.When beginning remove the expire metrics from the storage,transaction is too large exception is thrown. I have adjusted the limit to 1GB, but it seems still not enough.
Should I let go of restrictions or some recommend value?

___
### Bug
SkyWalking 8.1.0

moonsphere:
Closes #5749

TiDB can not delete many rows in one transaction.

so maybe we can use @xxxlxy2008 's genius idea --- split one `delete` stmt into multiple `delete` stmt with `limit` and loop until all wanted data has be deleted

to detect affect-row, users need ensure "useAffectedRows=true" must be configured in jdbc url.

to don't affect MySQL logic, we can extend a new `TiDBStorageProvider`, TiDB user can use

```
storage:
  tidb:
```

to replace

```
storage:
  mysql:
```

to use this patch.

> remain question: `delete ... limit N` will became slower due to TiDB need scan delete tombstone again again, but maybe it can be improve in future, this PR just handle "transaction too large error"


oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/mysql/TiDBHistoryDeleteDAO.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.skywalking.oap.server.library.client.jdbc.hikaricp.JDBCHikariCPClient;
+import org.apache.skywalking.oap.server.storage.plugin.jdbc.SQLBuilder;
+import org.apache.skywalking.oap.server.storage.plugin.jdbc.h2.dao.H2HistoryDeleteDAO;
+import org.joda.time.DateTime;
+
+import java.io.IOException;
+import java.sql.Connection;
+import java.sql.SQLException;
+
+public class TiDBHistoryDeleteDAO extends H2HistoryDeleteDAO {
```
wu-sheng, line 14: Please add comments about why you need this.

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/mysql/TiDBStorageProvider.java
```diff
@@ -0,0 +1,25 @@
...
+import org.apache.skywalking.oap.server.core.storage.IHistoryDeleteDAO;
+import org.apache.skywalking.oap.server.library.module.ServiceNotProvidedException;
+
+/**
+ * TiDB storage enhanced and came from MySQLStorageProvider to support TiDB.
+ *
+ * caution: need add "useAffectedRows=true" to jdbc url.
+ */
+@Slf4j
+public class TiDBStorageProvider extends MySQLStorageProvider {
```
wu-sheng, line 13: If you have a new provider, we need a new maven module for tidb storage, and make it compiled as a separate jar.

wu-sheng (commented):
Seems you miss the update of default application.yaml in the bootstrap module, configuration vocabulary doc after you changed the YAML, and storage setup doc.

wu-sheng:
@moonsphere I have a direct talk with PingCap's friends, right now, as a minor change, this deletion way seems the best choice.

wu-sheng:
@moonsphere Are you still working on this?

wu-sheng:
As no update, this is going to be replaced by #5844

PR #5755, opened 2020-10-30
Make the duration error message more specific

fgksgf:
### Improve the error message in [DurationUtils.java](https://github.com/apache/skywalking/blob/418bdff277179b4660e578b1837f21fd3db514cb/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/DurationUtils.java)
- [x] Make the duration error message more specific, the previous error message was confusing.


kezhenxu94 (approved):
LGTM, it was kind of confusing when we debug the CLI before

PR #5764, opened 2020-11-01
Polish the context related codes.

wu-sheng:
- [x] Make some polish after #5666.
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/c2141978d1039375598a32418b75161a78da22c1/CHANGES.md). Don't need to, I just polish codes.

Basically, ideas are
1. Closing unnecessary methods in the ContextCarrier, as it is much more important than I first designed this.
1. Remove unnecessary object creation(class ExtensionInjector) to reduce GC(very tiny).
1. Make the comments more clear.

@EvanLjp PTAL


EvanLjp:
@wu-sheng LGTM

kezhenxu94 (approved):

PR #5767, opened 2020-11-02
Polish the context and optimize the tag analyze

wu-sheng:

1. Polish context, including TracingContext, ContextCarrier, CorrelationContext, and ExtensionContext. With the complexity increasing, the codes seem like breaking, even the functions are good. I reorganized the codes and make it clear. cc @EvanLjp 
1. The current generated searchable tags could be duplicated, I use a HashSet to remove those.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextCarrier.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextCarrier.java
@@ -24,7 +24,6 @@ import lombok.Getter;
 import lombok.Setter;
 import org.apache.skywalking.apm.agent.core.base64.Base64;
 import org.apache.skywalking.apm.agent.core.conf.Constants;
-import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
 import org.apache.skywalking.apm.util.StringUtil;
 
 /**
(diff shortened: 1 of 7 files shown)


kezhenxu94 (approved):

wu-sheng:
Plugin tests of async seem to fail, I am not sure whether I break this.

EvanLjp:
@wu-sheng LGTM

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextCarrier.java
```diff
@@ -90,18 +89,21 @@ public ExtensionInjector extensionInjector() {
...
-     * Extract the extension context to the given span
+     * Extract the extension context to tracing context
      */
-    void extractExtensionTo(AbstractSpan span) {
-        this.extensionContext.handle(span);
-
+    void extractExtensionTo(TracingContext tracingContext) {
+        tracingContext.getExtensionContext().extract(this);
+        // The extension context could have field not to propagate further, so, must use the this.* to process.
+        this.extensionContext.handle(tracingContext.activeSpan());
```
wu-sheng, line 97: > Seems to break the plugin tests

@kezhenxu94 Yes, the current status of context is complex. `this` and `tracingContext.getExtensionContext()` are not the same thing.

Right now, it should be good.

kezhenxu94 (approved):

PR #5774, opened 2020-11-02
remove config ,class  not exist 

hsoftxl:
class not exist


wu-sheng (approved):
LGTM

wu-sheng:
After codes update, this has been deleted.

PR #5775, opened 2020-11-02
Fix deadlock problem when using elasticsearch-client-7.0.0

zifeihan:
### Fix deadlock problem when using es

Hello, a few days ago, there was a problem that es data could not be written in our production environment. We dump the thread stack of the production environment. After analysis, it is found that es client has a deadlock problem when writing data.thread dump [thread dump](https://thread.share.perfma.com/detail/1969879/overview). In order to solve the problem, I checked many related materials, and found that this problem has been fixed in elasticsearch-client version 7.5.0. [Prevent deadlock by using separate schedulers](https://github.com/elastic/elasticsearch/pull/48697)
 so i update the elasticsearch client 7.0.0 to elasticsearch client 7.5.0.
In order to avoid es compatibility issues, I use elasticsearch server 7.0.0 as data storage. Tested whether the ui function is normal, and checked whether the OAP is abnormal during operation. The test results show that everything is normal. and the current elasticsearch server in our production environment is 7.2.0, and we are also using elasticsearch client 7.5.0. It has been running stably for 3 days without any compatibility issues.


tools/dependencies/known-oap-backend-dependencies-es7.txt
```diff
@@ -29,14 +29,18 @@ curator-client-4.0.1.jar
...
+elasticsearch-core-7.5.0.jar
+elasticsearch-geo-7.5.0.jar
+elasticsearch-rest-client-7.5.0.jar
+elasticsearch-rest-high-level-client-7.5.0.jar
+elasticsearch-secure-sm-7.5.0.jar
+elasticsearch-x-content-7.5.0.jar
+lang-mustache-client-7.5.0.jar
+mapper-extras-client-7.5.0.jar
+parent-join-client-7.5.0.jar
+rank-eval-client-7.5.0.jar
```
wu-sheng, line 43: You need to update LICENSE including
1. Versions of `lang-mustache-client` and `parent-join-client` 
1. Add `mapper-extras-client` with license as we missed before
1. Add `rank-eval-client` with license as we missed before, both 6.3.2(es6 package) and 7.5.0(you updated)

wu-sheng:
@kezhenxu94 What ES version do we use to test?

wu-sheng (commented):
And you missed the change log update.

zifeihan:
> And you missed the change log update.

done.

kezhenxu94:
> @kezhenxu94 What ES version do we use to test?

6.3.2 and 7.4.2

wu-sheng:
@zifeihan CI logs show, the dependency changes are not accurate, please recheck. You could use the same scripts we used in the CI, https://github.com/apache/skywalking/tree/master/tools/dependencies.

zifeihan:
> @zifeihan CI logs show, the dependency changes are not accurate, please recheck. You could use the same scripts we used in the CI, https://github.com/apache/skywalking/tree/master/tools/dependencies.

done.

kezhenxu94 (dismissed):
LGTM

wu-sheng:
@zifeihan I have asked @kezhenxu94 to adjust e2e to text 7.0.0 and higher version, like 7.9.3(latest) to verify the compatibility. Let's wait for that.

zifeihan:
> @zifeihan I have asked @kezhenxu94 to adjust e2e to text 7.0.0 and higher version, like 7.9.3(latest) to verify the compatibility. Let's wait for that.

> ok, got it.

dist-material/release-docs/LICENSE
```diff
@@ -233,9 +233,13 @@ The text of each license is the standard Apache 2.0 license.
     Google: proto-google-common-protos 1.17.0: https://github.com/googleapis/googleapis , Apache 2.0
     Google: jsr305 3.0.2: http://central.maven.org/maven2/com/google/code/findbugs/jsr305/3.0.0/jsr305-3.0.0.pom , Apache 2.0
     Elasticsearch BV (Elasticsearch) 6.3.2: https://www.elastic.co/products/elasticsearch , Apache 2.0
-    Elasticsearch BV (Elasticsearch) 7.0.0: https://www.elastic.co/products/elasticsearch , Apache 2.0
-    lang-mustache-client 5.5.0: https://github.com/elastic/elasticsearch/tree/master/modules/lang-mustache , Apache 2.0
-    parent-join-client 5.5.0: https://github.com/elastic/elasticsearch/tree/master/modules/parent-join , Apache 2.0
+    Elasticsearch BV (Elasticsearch) 7.5.0: https://www.elastic.co/products/elasticsearch , Apache 2.0
```
wu-sheng, line 236:
```
Elasticsearch BV (Elasticsearch) 6.3.2, 7.5.0: https://www.elastic.co/products/elasticsearch , Apache 2.0
```
For multiple versions, should be like this.

wu-sheng (changes requested):
There are several other libs updated in this according to known-oap-backend-dependencies-es7.txt change log.
So, the LICENSE still needs some polish. In the current, we missed ES7 dependency version, such as `hppc 0.7.1`, which is ES6 depending.

@zifeihan Could you help on adding the ES6 dependency versions to the LICENSE file like I commented for the ES client.

dist-material/release-docs/LICENSE
zifeihan: done.

wu-sheng (approved):
It is much better now.

kezhenxu94 (approved):

libinglong:
Notice this issue will cause the log "Grpc server thread pool is full, rejecting the task".

zifeihan:
> "Grpc server thread pool is full, rejecting the task

Thanks submit this, why it cause this error log?  our production environment processes tens of billions of data every day, and we have never found such logs.

libinglong:
> > "Grpc server thread pool is full, rejecting the task
> 
> Thanks submit this, why it cause this error log? our production environment processes tens of billions of data every day, and we have never found such logs.

I found a lot of such logs in the oap log.
The thread stacks show all the thread of grpcServerPool is WAITING. 

```text
"grpcServerPool-1-thread-14" #360 prio=5 os_prio=0 tid=0x00007f1bb8041800 nid=0x104d6 waiting for monitor entry [0x00007f219d2f1000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at org.elasticsearch.action.bulk.BulkProcessor.internalAdd(BulkProcessor.java:319)
	- waiting to lock <0x00000000e2c38d28> (a org.elasticsearch.action.bulk.BulkProcessor)
	at org.elasticsearch.action.bulk.BulkProcessor.add(BulkProcessor.java:304)
	at org.elasticsearch.action.bulk.BulkProcessor.add(BulkProcessor.java:290)
	at org.apache.skywalking.oap.server.storage.plugin.elasticsearch.base.BatchProcessEsDAO.asynchronous(BatchProcessEsDAO.java:56)
```

libinglong:
@zifeihan I don't mean our scene is the same. The bug of the es client also cause other problems.

zifeihan:
@llibing

> @zifeihan I don't mean our scene is the same. The bug of the es client also cause other thread-block problems which this merge will fix.

Have you upgraded to version 8.3.0 or higher? If so, please dump thread to analyze the root cause of the problem. If not, I think this problem may be caused by the deadlock of es-client.

zifeihan:
@llibing

> @zifeihan I don't mean our scene is the same. The bug of the es client also cause other thread-block problems which this merge will fix.

Have you upgraded to version 8.3.0 or higher? If so, please dump thread to analyze the root cause of the problem. If not, I think this problem may be caused by the deadlock of es-client.

libinglong:
> @libinglong
> 
> > @zifeihan I don't mean our scene is the same. The bug of the es client also cause other thread-block problems which this merge will fix.
> 
> Have you upgraded to version 8.3.0 or higher? If so, please dump thread to analyze the root cause of the problem. If not, I think this problem may be caused by the deadlock of es-client.

I will upgrade to 8.3.0 soon. Yes, I am sure It's the problem of es-client.

PR #5776, opened 2020-11-02
Simplifies canceling workflow

potiuk:
Instead of multiple workflow-runs generated for different workflows
this change replaces it with single workflow, cancelling all others.
It is triggered by only one CI-IT workflow, because this one is
always triggered when all others are. It also sets parallelism
of the matrix to 4, so that there are not too many jobs in
parallel taken when this workflow runs.

### DIFF
--- a/.github/workflows/cancel.yaml
+++ b/.github/workflows/cancel.yaml
@@ -18,22 +18,44 @@ name: Cancel
 
 on:
   workflow_run:
-    workflows:
-      - CI AND IT
-      - E2E
-      - PluginsTest
+    workflows: [CI AND IT]
     types:
       - requested
 
 jobs:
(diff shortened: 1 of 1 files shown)


potiuk:
You can see the cancel workflow running here: https://github.com/potiuk/skywalking/runs/1343437661?check_suite_focus=true#step:2:71

potiuk:
An alternative to that will be manual copy&pasting the action step as many times as many workflows you have to cancel and hard-code the workflowFileName as parameter for each step.

This has the added benefit that all steps will run on one machine and there will be only one job started to run them all (at the expense of copy&pasted code). 

I can do it if you think it's a better idea :)

kezhenxu94 (approved):

PR #5777, opened 2020-11-03
Fixed unit of time mixing problem in kubernetes-client

EvanLjp:
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number.https://github.com/kubernetes-client/java/pull/1344(https://github.com/apache/skywalking/blob/c2141978d1039375598a32418b75161a78da22c1/CHANGES.md).

in Kubernetes-java-client, the millisecond is used to deliver to the API server.
but the unit in the API server is Second.
Because the feature was fixed in the last few days, so choose manually pass the time out parameter.

### DIFF
--- a/oap-server/server-cluster-plugin/cluster-kubernetes-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/kubernetes/NamespacedPodListInformer.java
+++ b/oap-server/server-cluster-plugin/cluster-kubernetes-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/kubernetes/NamespacedPodListInformer.java
@@ -83,7 +83,7 @@ public enum NamespacedPodListInformer {
         SharedIndexInformer<V1Pod> podSharedIndexInformer = factory.sharedIndexInformerFor(
             params -> coreV1Api.listNamespacedPodCall(
                 podConfig.getNamespace(), null, null, null, null,
-                podConfig.getLabelSelector(), Integer.MAX_VALUE, params.resourceVersion, params.timeoutSeconds,
+                podConfig.getLabelSelector(), Integer.MAX_VALUE, params.resourceVersion, 300,
                 params.watch, null
             ),
             V1Pod.class, V1PodList.class
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@ Release Notes.
 * Support choose files to active the meter analyzer.
 * Improve Kubernetes service registry for ALS analysis.
 * Improve the queryable tags generation. Remove the duplicated tags to reduce the storage payload.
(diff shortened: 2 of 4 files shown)


kezhenxu94 (commented):
Could you please also fix those in class `org.apache.skywalking.oap.server.receiver.envoy.als.k8s.K8SServiceRegistry`

kezhenxu94 (approved):
LGTM

oap-server/server-cluster-plugin/cluster-kubernetes-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/kubernetes/NamespacedPodListInformer.java
```diff
@@ -83,7 +83,7 @@ private void doStartPodInformer(ClusterModuleKubernetesConfig podConfig) throws
         SharedIndexInformer<V1Pod> podSharedIndexInformer = factory.sharedIndexInformerFor(
             params -> coreV1Api.listNamespacedPodCall(
                 podConfig.getNamespace(), null, null, null, null,
-                podConfig.getLabelSelector(), Integer.MAX_VALUE, params.resourceVersion, params.timeoutSeconds,
+                podConfig.getLabelSelector(), Integer.MAX_VALUE, params.resourceVersion, 300,
```
wu-sheng, line 86: 300 seconds? I am a little confused.

EvanLjp: image
Follow the Kubernetes default configuration

kezhenxu94: https://github.com/kubernetes-client/java/pull/1344/files

The default value is indeed 300s

wu-sheng: OK, got it now. Could you submit an issue to track this? 

I think we need to upgrade the client to 10.1(maybe? the Last release is 10.0 already), and change this back to the right way.

EvanLjp: https://github.com/apache/skywalking/issues/5778

wu-sheng (approved):

PR #5779, opened 2020-11-03
test: E2E Elasticsearch 7.x highest and lowest versions

kezhenxu94:
### E2E tests ElasticSearch 7.0.0 and 7.9.3

Add ElasticSearch 7.0.0 and 7.9.3 to E2E tests, and fix a minor problem that some E2E tests are not executed because `paths` doesn't take effect.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Ref #5775.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -18,7 +18,9 @@ name: E2E
 
 on:
   pull_request:
+    # Do not run this workflow when and only when `*.md` files are modified
     paths:
+      - '**'
       - '!**.md'
   push:
     branches:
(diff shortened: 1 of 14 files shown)


.github/workflows/e2e.yaml
```diff
@@ -18,7 +18,9 @@ name: E2E
 
 on:
   pull_request:
+    # Do not run this workflow when and only when `*.md` files are modified
     paths:
+      - '**'
```
kezhenxu94, line 23: without `**`, GitHub Actions won't run these workflows kinda of misleading

wu-sheng: Got it.

wu-sheng (approved):

PR #5782, opened 2020-11-03
The Cancel step is now optimised in one job

potiuk:
The cancel steps were tried in separate jobs but it causes
an unnecessary increase in the number of jobs blocking the queue
even if we limited parallelism. This PR combines all the cancel
actions into a single job with multiple steps.

While this is a bit more copy&pasting and less DRY it is also
far more optimised and "aggressive" - i.e. once any cancel
job starts to be running it will immediately cancel all duplicates
for all workflows.

### DIFF
--- a/.github/workflows/cancel.yaml
+++ b/.github/workflows/cancel.yaml
@@ -28,34 +28,144 @@ jobs:
   cancel-outdated-builds:
     runs-on: ubuntu-18.04
     timeout-minutes: 10
-    strategy:
-      max-parallel: 4
-      matrix:
-        workflowFileName:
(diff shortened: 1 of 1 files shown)


potiuk:
Example here: https://github.com/potiuk/skywalking/runs/1346617869?check_suite_focus=true

potiuk:
It takes 15 s to run all the cancel steps in a single job.

wu-sheng (approved):
Seems good to me. @kezhenxu94 recheck please.

kezhenxu94 (approved):

PR #5784, opened 2020-11-03
Cluster manager health v2

### LINKED ISSUE #5633: Add health check for Storage/Cluster management module

xbkaishui:
- [x] Closes #5633 add health check for cluster manager


oap-server/server-cluster-plugin/cluster-zookeeper-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/zookeeper/ZookeeperCoordinator.java
```diff
@@ -22,29 +22,40 @@
...
 
     private static final String REMOTE_NAME_PATH = "remote";
 
     private final ClusterModuleZookeeperConfig config;
     private final ServiceDiscovery<RemoteInstance> serviceDiscovery;
     private final ServiceCache<RemoteInstance> serviceCache;
     private volatile Address selfAddress;
+    @Setter
+    private HealthCheckMetrics healthChecker;
+    private final ModuleDefineHolder manager;
```
wu-sheng, line 54: Move final to following other `final` fields in the top of the list.

oap-server/server-cluster-plugin/cluster-zookeeper-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/zookeeper/ZookeeperCoordinator.java
```diff
@@ -22,29 +22,40 @@
...
 public class ZookeeperCoordinator implements ClusterRegister, ClusterNodesQuery {
-    private static final Logger LOGGER = LoggerFactory.getLogger(ZookeeperCoordinator.class);
 
     private static final String REMOTE_NAME_PATH = "remote";
 
     private final ClusterModuleZookeeperConfig config;
     private final ServiceDiscovery<RemoteInstance> serviceDiscovery;
     private final ServiceCache<RemoteInstance> serviceCache;
     private volatile Address selfAddress;
+    @Setter
```
wu-sheng, line 52: You don't need the setter I think. Please don't open unnecessary access.

wu-sheng (commented):
Two code style issues, please fix them in all implementations. Others look good to me.

xbkaishui:
Greate, will fix now

xbkaishui:
Done

wu-sheng (approved):
LGTM

Issue #5785, opened 2020-11-03
storage-jdbc isExists not set dbname


FoolRabbit:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement
___
### Bug
- Which version of SkyWalking, OS and JRE?
8.2.0
- Which company or project?

- What happened?
If possible, provide a way to reproduce the error. e.g. demo application, component version.
when i has two or more schemas, can't auto create tables

```
    @Override
    protected boolean isExists(Model model) throws StorageException {
        TableMetaInfo.addModel(model);
        JDBCHikariCPClient h2Client = (JDBCHikariCPClient) client;
        try (Connection conn = h2Client.getConnection()) {
            // getTables need set catalog = conn.getCatalog()
            try (ResultSet rset = conn.getMetaData().getTables(null, null, model.getName(), null)) {
                if (rset.next()) {
                    return true;
                }
            }
        } catch (SQLException | JDBCClientException e) {
            throw new StorageException(e.getMessage(), e);
        }
        return false;
    }
```

wu-sheng:
@JaredTan95 Could you take a look?

JaredTan95:
I will check it soon.

PR #5786, opened 2020-11-04
fix storage-jdbc isExists not set dbname

JaredTan95:
fix https://github.com/apache/skywalking/issues/5785


wu-sheng:
I think you ignore the pull request template. You should update the changes.md in the root.

JaredTan95:
> I think you ignore the pull request template. You should update the changes.md in the root.

ok

wu-sheng (approved):

PR #5796, opened 2020-11-05
Add branch protection and status check require

kezhenxu94:
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/c2141978d1039375598a32418b75161a78da22c1/CHANGES.md). NO NEED

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -31,3 +31,15 @@ github:
     squash:  true
     merge:   false
     rebase:  false
+  protected_branches:
+    master:
+      required_status_checks:
+        strict: true
+        contexts:
+          - CI
+          - Single
+          - Storage
+          - Cluster
+      required_pull_request_reviews:
+        dismiss_stale_reviews: true
+        required_approving_review_count: 1


kezhenxu94:
Please review according to this https://cwiki.apache.org/confluence/display/INFRA/git+-+.asf.yaml+features#git.asf.yamlfeatures-BranchProtection

kezhenxu94:
As per the infra team, this should be a bug if we never have a `protected_branches` section in the `.asf.yaml`

https://issues.apache.org/jira/browse/INFRA-21073

wu-sheng (approved):

.asf.yaml
```diff
@@ -31,3 +31,15 @@ github:
...
     rebase:  false
+  protected_branches:
+    master:
+      required_status_checks:
+        strict: true
+        contexts:
+          - CI
+          - Single
+          - Storage
+          - Cluster
```
kezhenxu94, line 42: These contexts are not exactly the same as what we see in the PR page, thus I fork and set in my forked repo and use GitHub API to see what the `context` is, hopefully it's right

image
image

PR #5799, opened 2020-11-05
Fix SnakeYAML CVE. Make doc and sharing receiver codes consistent.

wu-sheng:

* Fix CVE in the alarm module, when activating the dynamic configuration feature.
* Make the codes and doc consistent in sharding server and core server.


wu-sheng:
The CVE is caused by our previous usage, which could be found from here, https://www.appmarq.com/public/security,1039056,Avoid-insecure-use-of-YAML-deserialization-when-using-SnakeYaml-JEE

kezhenxu94 (approved):
LGTM

PR #5802, opened 2020-11-06
Fix CVE in `UninstrumentedGatewaysConfig`

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>. Follow #5799 

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,6 +32,7 @@ Release Notes.
 * Fix `searchService` bug in the InfluxDB storage implementation.
 * Fix CVE in the alarm module, when activating the dynamic configuration feature.
 * Fix CVE in the endpoint grouping, when activating the dynamic configuration feature.
+* Fix CVE in the uninstrumented gateways configs, when activating the dynamic configuration feature.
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -30,8 +30,7 @@ Release Notes.
 * Fix deadlock problem when using elasticsearch-client-7.0.0.
 * Fix storage-jdbc isExists not set dbname.
 * Fix `searchService` bug in the InfluxDB storage implementation.
-* Fix CVE in the alarm module, when activating the dynamic configuration feature.
-* Fix CVE in the endpoint grouping, when activating the dynamic configuration feature.
+* Fix CVE of SnakeYAML when activating the dynamic configuration feature.
```
kezhenxu94, line 33: I merge them into one, they are essential one type of fix IMO

wu-sheng: CVE needs to be explicit. Even we don't report the CVE-ID(s), but basically, these CVEs are different, just the same in the code style.

wu-sheng (approved):
Please reply on the private mail about the progress.

PR #5803, opened 2020-11-06
Avoided to interrupt the process of the application in the `SWExecutionHookWrapper` wrapper class.

zifeihan:
Fix interrupt the process of the application in the `SWExecutionHookWrapper` wrapper class.

Because this class wraps the user's real class, under certain circumstances, the user's real business fails because of this mechanism. Therefore, we need to catch our abnormal operation here to avoid causing business application failure.


wu-sheng (commented):
Please don't change other files

zifeihan:
> Please don't change other files

done.

wu-sheng (approved):

PR #5813, opened 2020-11-09
Change the intercept bootstrap class to the same way as intercept the normal class

zifeihan:
### Change the intercept boot class to the same way as intercept the regular class
- [ ] Explain briefly about why the bug exists and how to fix it.

We now use the bootstrap interceptor to generate an interceptor of xxx_internal, and then call the real interceptor in the interceptor. For example
image

After the modification, we will not generate the xxx_internal class, For example

image
image

1. In order to reduce the use of metaspace space, avoid java.lang.OutOfMemoryError: Metaspace
2. For the readability of the code, it is easy to modify in the future.
3. For the next plugin dynamic switch.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/ClassEnhancePluginDefine.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/ClassEnhancePluginDefine.java
@@ -32,7 +32,11 @@ import org.apache.skywalking.apm.agent.core.logging.api.LogManager;
 import org.apache.skywalking.apm.agent.core.plugin.AbstractClassEnhancePluginDefine;
 import org.apache.skywalking.apm.agent.core.plugin.EnhanceContext;
 import org.apache.skywalking.apm.agent.core.plugin.PluginException;
-import org.apache.skywalking.apm.agent.core.plugin.bootstrap.BootstrapInstrumentBoost;
+import org.apache.skywalking.apm.agent.core.plugin.bootstrap.interceptor.BootstrapConstructorInter;
+import org.apache.skywalking.apm.agent.core.plugin.bootstrap.interceptor.BootstrapInstMethodsInter;
+import org.apache.skywalking.apm.agent.core.plugin.bootstrap.interceptor.BootstrapInstMethodsInterWithOverrideArgs;
+import org.apache.skywalking.apm.agent.core.plugin.bootstrap.interceptor.BootstrapStaticMethodsInter;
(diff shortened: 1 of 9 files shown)


wu-sheng (commented):
We must generate the internal interceptor, which is how SkyWalking plugin system works. This makes us a chance to inject any necessary logic in the future. And the classloader of plugin interceptor controllable by us.
Metadata wouldn't be such fragile, and small. Bootstrap level enhancement would be very limited.
As, I can't see why this blocks the plugin dynamic off. The internal interceptor and plugin interceptor are in different class loaders, the classloader of the plugin interceptor is easier to access the Config class in the core.

zifeihan:
> We must generate the internal interceptor, which is how SkyWalking plugin system works. This makes us a chance to inject any necessary logic in the future. And the classloader of plugin interceptor controllable by us.
> Metadata wouldn't be such fragile, and small. Bootstrap level enhancement would be very limited.
> As, I can't see why this blocks the plugin dynamic off. The internal interceptor and plugin interceptor are in different class loaders, the classloader of the plugin interceptor is easier to access the Config class in the core.

I think the kernel should be kept simple and easy to modify. By modifying the TARGET_INTERCEPTOR field, to generate the interceptor class. Would it be a bit too complicated? Will this writing make it difficult for others to understand?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/ClassEnhancePluginDefine.java
```diff
@@ -130,14 +134,13 @@
                 if (isBootstrapInstrumentation()) {
                     newClassBuilder = newClassBuilder.constructor(constructorInterceptPoint.getConstructorMatcher())
                                                      .intercept(SuperMethodCall.INSTANCE.andThen(MethodDelegation.withDefaultConfiguration()
-                                                                                                                 .to(BootstrapInstrumentBoost
-                                                                                                                     .forInternalDelegateClass(constructorInterceptPoint
-                                                                                                                         .getConstructorInterceptor()))));
+                                                                                                                 .to(new BootstrapConstructorInter(
+                                                                                                                     constructorInterceptPoint.getConstructorInterceptor()))));
```
wu-sheng, line 138: ByteBuddy doesn't recommend the instance-delegation explicitly for the bootstrap process. Its author recommends the static level delegate. So I changed to the delegator generation solution.
I wouldn't have the same experiences as he does.

wu-sheng: I don't have a full understanding of bootstrap level instrumentation. I would like to follow what he recommended.

PR #5815, opened 2020-11-09
sync ui: label fix in topology chart + dart-sass replacing node-sass

zeaposs:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>.
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/c2141978d1039375598a32418b75161a78da22c1/CHANGES.md).

UI submodule update. List of changes:
- fix: Removed Les Miserables from radial chart in topology. (apache/skywalking-rocketbot-ui#367)
- feat: remove node-sass, use dart-sass (apache/skywalking-rocketbot-ui#369)

PR #5816, opened 2020-11-10
Fix the expired link of the changelog file.

wu-sheng:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>. Follow #5815
- [ ] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/c2141978d1039375598a32418b75161a78da22c1/CHANGES.md).

The previous link leads to 8.2 markdown document only, which is quite confusing.


hanahmily (approved):
LGTM

PR #5818, opened 2020-11-10
Fix that chunked string is incorrect while the tag contains colon

ascrutae:


hanahmily (approved):
LGTM

kezhenxu94 (commented):
What if the colon exists in the tag value?

wu-sheng:
> What if the colon exists in the tag value?

Nothing happens. This is a special case of this function, the key is a combination, label + `:` + rank(percentile).

PR #5823, opened 2020-11-10
Update FileWriter.java

mainColorZwh:
file "apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/FileWriter.java",add  volatile to field "FileWriter INSTANCE" to avoid Instruction reordering.if no volatile , client may get an "FileWriter INSTANCE" that have not been constructed compeletly.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/FileWriter.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/FileWriter.java
@@ -42,7 +42,7 @@ import org.apache.skywalking.apm.util.RunnableWithExceptionProtection;
  * The <code>FileWriter</code> support async file output, by using a queue as buffer.
  */
 public class FileWriter implements IWriter {
-    private static FileWriter INSTANCE;
+    private volatile static FileWriter INSTANCE;
     private static final Object CREATE_LOCK = new Object();
     private FileOutputStream fileOutputStream;
     private ArrayBlockingQueue logBuffer;


wu-sheng:
This constructor should be called as soon as the agent main entrance loaded, as `SnifferConfigInitializer` has `LOGGER = LogManager.getLogger(SnifferConfigInitializer.class);`. Which case could cause the issue you describe? Do you face any real-world issue?

mainColorZwh:
> This constructor should be called as soon as the agent main entrance loaded, as `SnifferConfigInitializer` has `LOGGER = LogManager.getLogger(SnifferConfigInitializer.class);`. Which case could cause the issue you describe? Do you face any real-world issue?

1.This constructor first called at 'PluginBootstrap.loadPlugins()' method (en... I have debugged for that). 
2.FileWriter use singleton of double-check type to prevent concurrent problem ,  I just find that it is not a standard double-check type.It may be hard to produce the scenario because of low probability.
3.git hub dont support insert a picture ,so I attach a picture to explain more details at link [https://blog.csdn.net/dshf_1/article/details/109614612](blog)

wu-sheng:
I could understand officially this is not standard, yes, you are correct. But, the issue is, if you add `volatile`, there would be unnecessary performance cost. Actually, the whole double-check is not necessary :) Like you have debugged, there is no concurrency case in the boot stage.

mainColorZwh:
> I could understand officially this is not standard, yes, you are correct. But, the issue is, if you add `volatile`, there would be unnecessary performance cost. Actually, the whole double-check is not necessary :) Like you have debugged, there is no concurrency case in the boot stage.

mainColorZwh:
en ...thanks for your particutar explain:)

PR #5824, opened 2020-11-11
fix searchTag not work

EvanLjp:
image
the search tag is invalid because Graphql need no-arg constructor


wu-sheng (approved):
Thanks for fixing this before release.

PR #5827, opened 2020-11-11
Improve ES query speed by using traceId

zifeihan:
### Improve ES query speed by using traceId
- [ ] The benchmark result.
```text
<Paste the benchmark results here>
```

Because the third part of traceId has time information, we can search within the specified time slice when querying. When there are many fragments and the amount of data is large, this can greatly improve the query speed.

### DIFF
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/TraceQueryEsDAO.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/TraceQueryEsDAO.java
@@ -163,11 +163,22 @@ public class TraceQueryEsDAO extends EsDAO implements ITraceQueryDAO {
         sourceBuilder.query(QueryBuilders.termQuery(SegmentRecord.TRACE_ID, traceId));
         sourceBuilder.size(segmentQueryMaxSize);
 
-        String traceIdThirdPart = traceId.split("\\.")[2];
-        long timeBucket = TimeBucket.getRecordTimeBucket(Long.valueOf(traceIdThirdPart) / 10000);
+        SearchResponse response = null;
+        final String[] traceIdPart = traceId.split("\\.");
 
-        SearchResponse response = getClient().search(
-            new TimeRangeIndexNameMaker(SegmentRecord.INDEX_NAME, timeBucket, timeBucket), sourceBuilder);
+        /**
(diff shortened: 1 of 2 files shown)


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/TraceQueryEsDAO.java
```diff
@@ -162,7 +163,11 @@ public TraceBrief queryBasicTraces(long startSecondTB,
         sourceBuilder.query(QueryBuilders.termQuery(SegmentRecord.TRACE_ID, traceId));
         sourceBuilder.size(segmentQueryMaxSize);
 
-        SearchResponse response = getClient().search(SegmentRecord.INDEX_NAME, sourceBuilder);
+        String traceIdThirdPart = traceId.split("\\.")[2];
+        long timeBucket = TimeBucket.getRecordTimeBucket(Long.valueOf(traceIdThirdPart) / 10000);
+
+        SearchResponse response = getClient().search(
+            new TimeRangeIndexNameMaker(SegmentRecord.INDEX_NAME, timeBucket, timeBucket), sourceBuilder);
```
wu-sheng, line 170: I am afraid this is not working. We don't have this timestamp-based trace-id rule anymore.

wu-sheng: Take a look at other agent implementations, such as LUA.

zifeihan: > Take a look at other agent implementations, such as LUA.

Ok let me take a look.

zifeihan: @wu-sheng I modified some content, can you help me see if this is feasible?

wu-sheng: From my personal understanding, we don't have the format rule of traceid anymore. Any string is acceptable, and even it is combined with 3 parts, doesn't mean there is a timestamp included.

PR #5836, opened 2020-11-13
Add logic endpoint section in the agent setup doc.

wu-sheng:

The logic endpoint concept is added in recent releases. I add this doc to make the concept and usage more clear. Previously, it only exists in the development doc, which is hard to find and be noted for the end-users.

@hailin0 Please review the doc, as many naming format rules are built by you.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,6 +48,7 @@ Release Notes.
 
 #### Documentation
 * Add VNode FAQ doc.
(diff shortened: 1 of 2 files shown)


kezhenxu94 (approved):

hailin0:
good！no problem

hailin0 (approved):

PR #5839, opened 2020-11-13
Fix request url traceId sometimes shows "Ignored_Trace"

huliangdream:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,7 +19,7 @@ Release Notes.
 * Fix NPE in the nutz plugin.
 * Provide Apache Commons DBCP 2.x plugin.
 * Add the plugin for mssql-jtds 1.x plugin.
-* Fix request url traceId sometimes shows "Ignored_Trace".
+* Fix the default ignore mechanism isn't accurate enough bug.
 
 #### OAP-Backend
 * Add the `@SuperDataset` annotation for BrowserErrorLog.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@ Release Notes.
 * Add the plugin for async-http-client 2.x
 * Fix NPE in the nutz plugin.
 * Provide Apache Commons DBCP 2.x plugin.
+* Fix request url traceId sometimes shows "Ignored_Trace".
 
 #### OAP-Backend
 * Add the `@SuperDataset` annotation for BrowserErrorLog.
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManagerExtendService.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManagerExtendService.java
@@ -32,7 +32,7 @@ import java.util.Arrays;
 @DefaultImplementor
 public class ContextManagerExtendService implements BootService, GRPCChannelListener {
     
-    private String[] IGNORE_SUFFIX_ARRAY = new String[0];
+    private String[] ignoreSuffixArray = new String[0];
     
     private volatile GRPCChannelStatus status = GRPCChannelStatus.DISCONNECT;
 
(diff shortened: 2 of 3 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManagerExtendService.java
```diff
@@ -61,7 +63,7 @@ public AbstractTracerContext createTraceContext(String operationName, boolean fo
         }
 
         int suffixIdx = operationName.lastIndexOf(".");
-        if (suffixIdx > -1 && Config.Agent.IGNORE_SUFFIX.contains(operationName.substring(suffixIdx))) {
+        if (suffixIdx > -1 && Arrays.stream(Config.Agent.IGNORE_SUFFIX.split(",")).anyMatch(a -> a.equals(operationName.substring(suffixIdx)))) {
```
wu-sheng, line 66: The `split` should happen in the `boot` stage, and only do match in the runtime.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManagerExtendService.java
```diff
@@ -27,8 +27,13 @@
 import org.apache.skywalking.apm.agent.core.remote.GRPCChannelStatus;
 import org.apache.skywalking.apm.agent.core.sampling.SamplingService;
 
+import java.util.Arrays;
+
 @DefaultImplementor
 public class ContextManagerExtendService implements BootService, GRPCChannelListener {
+    
+    private static String[] IGNORE_SUFFIX_ARRAY = Config.Agent.IGNORE_SUFFIX.split(",");
```
wu-sheng, line 35: Put the initialization process in the `boot` method, rather than there. It is safer and keeps the initialization sequence more clear.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManagerExtendService.java
```diff
@@ -27,8 +27,13 @@
 import org.apache.skywalking.apm.agent.core.remote.GRPCChannelStatus;
 import org.apache.skywalking.apm.agent.core.sampling.SamplingService;
 
+import java.util.Arrays;
+
 @DefaultImplementor
 public class ContextManagerExtendService implements BootService, GRPCChannelListener {
+    
+    private static String[] IGNORE_SUFFIX_ARRAY = new String[0];
```
wu-sheng, line 35: ContextManagerExtendService's instance is controlled by the agent core. This field should not be a `static` one, like `status`.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManagerExtendService.java
```diff
@@ -27,8 +27,13 @@
 import org.apache.skywalking.apm.agent.core.remote.GRPCChannelStatus;
 import org.apache.skywalking.apm.agent.core.sampling.SamplingService;
 
+import java.util.Arrays;
+
 @DefaultImplementor
 public class ContextManagerExtendService implements BootService, GRPCChannelListener {
+    
+    private String[] IGNORE_SUFFIX_ARRAY = new String[0];
```
wu-sheng, line 35: Compiling local, this name format would not pass the CI, because of the format.

CHANGES.md
```diff
@@ -19,6 +19,7 @@ Release Notes.
 * Fix NPE in the nutz plugin.
 * Provide Apache Commons DBCP 2.x plugin.
 * Add the plugin for mssql-jtds 1.x plugin.
+* Fix request url traceId sometimes shows "Ignored_Trace".
```
wu-sheng, line 22:
```suggestion
* Fix the default ignore mechanism isn't accurate enough bug.
```

wu-sheng (approved):

kezhenxu94 (approved):
LGTM

PR #5844, opened 2020-11-14
add Tidb provider module.

### LINKED ISSUE #5749: Transaction is too large when use TiDB as storage

JaredTan95:
### Fix https://github.com/apache/skywalking/pull/5754, https://github.com/apache/skywalking/issues/5749
- [x] Add a e2e test to verify it works.

- [x] Closes #5749.


wu-sheng (commented):
1. Seems you missed the license.
1. I think you accidentally updated the UI submodule?
1. Please add e2e for TiDB storage provider.

oap-server/server-storage-plugin/storage-tidb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/tidb/TiDBHistoryDeleteDAO.java
```diff
@@ -0,0 +1,76 @@
...
+
+    public TiDBHistoryDeleteDAO(JDBCHikariCPClient client) {
+        this.client = client;
+    }
+
+    @Override
+    public void deleteHistory(Model model, String timeBucketColumnName, int ttl) throws IOException {
+        SQLBuilder dataDeleteSQL = new SQLBuilder("delete from " + model.getName() + " where ")
+            .append(timeBucketColumnName).append("<= ? and ")
+            .append(timeBucketColumnName).append(">= ?")
```
wu-sheng, line 43: Why need >= `minDate`? What is the scenario?

xbkaishui:
Due to tidb not do well on batch delete,  is it better we use partition table? 
tiflash is suitable for olap related task. do we need integrate it in e2e test case?

oap-server/server-storage-plugin/storage-tidb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/tidb/TiDBHistoryDeleteDAO.java
JaredTan95: this could be deleted.https://github.com/apache/skywalking/pull/5845

wu-sheng:
@JaredTan95 Please resolveconflict.

JaredTan95:
There is some problem creating a new database when there starting with tidb. :-P

kezhenxu94:
> There is some problem creating a new database when there starting with tidb. :-P

Need any help?

JaredTan95:
> > There is some problem creating a new database when there starting with tidb. :-P
> 
> Need any help? Feel free to ping me if anything I can help

Yes, There is some problem in e2e test, due to we have not found a way to create sw db during tidb container creating.

wu-sheng:
@kezhenxu94 Anyway to inject an initial container?

kezhenxu94 (approved):
Tests passed. LGTM except for one nit inline. You may also want to add a case into the ttl case matrix

wu-sheng (commented):
Could we add a TTL e2e for tidb too?

wu-sheng (approved):
Once e2e passed, this is good for me.

JaredTan95:
> Could we add a TTL e2e for tidb too?

added.

PR #5846, opened 2020-11-15
fix flaky test in comsumerTest

zcai2:
Fixed flaky test that caused by ConsumerTest.java under skywaling-apm-datacarrier

Existing tests are flaky because the testConsumerMoreThanChannel doesn't clear channel at the beginning of this test case.

Both testConsumerLessThanChannel and testConsumerMoreThanChannel use the same carrier. In testConsumerLessThanChannel, it produces some data and store in the channel. When we run testConsumerMoreThanChannel, the data introduced by testConsumerLessThanChannel will affect this test case. 

The flaky tests are found by running https://github.com/idflakies/iDFlakies

### DIFF
--- a/apm-commons/apm-datacarrier/src/test/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumerTest.java
+++ b/apm-commons/apm-datacarrier/src/test/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumerTest.java
@@ -35,6 +35,7 @@ public class ConsumerTest {
 
     @Test
     public void testConsumerLessThanChannel() throws IllegalAccessException {
+        BUFFER.drainTo(new ArrayList<SampleData>());
         final DataCarrier<SampleData> carrier = new DataCarrier<SampleData>(2, 100);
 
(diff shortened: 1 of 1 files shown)


apm-commons/apm-datacarrier/src/test/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumerTest.java
```diff
@@ -35,6 +35,7 @@
 
     @Test
     public void testConsumerLessThanChannel() throws IllegalAccessException {
+        BUFFER.drainTo(new ArrayList<SampleData>());
```
kezhenxu94, line 38: Seems `BUFFER` is actually not used in `testConsumerLessThanChannel`? Can `BUFFER` be simply moved into `testConsumerMoreThanChannel` as a local variable?

zcai2: good catch! removed

kezhenxu94 (approved):

PR #5849, opened 2020-11-16
Support `sideCar.internalErrorCode` in the sources

wu-sheng:

The sidecar is a common concept today, as the service mesh has been the next-generation infrastructure, no matter in k8s or out. I found we need to support this in the mesh telemetry internal format to support defining metrics in the mesh case.
There are 2 major changes in the PR
1. Nested classes definition is not supported in the previous OAL core, so I added them. It should support any nested class.
2. Support `sideCar.internalErrorCode` in the Service, ServiceInstance, Endpoint, ServiceRelation, and ServiceInstanceRelation sources, then you could have OAL scripts like this

```
service_cpm_sidecar = from(Service.*).filter(sideCar.internalErrorCode != "").cpm();
service_sla_sidecar = from(Service.*).percent(sideCar.internalErrorCode != "");
```

The current Istio+Envoy mesh telemetry's `sideCar.internalErrorCode` is mapping to envoy's v2 ALS definition, https://www.envoyproxy.io/docs/envoy/latest/api-v2/data/accesslog/v2/accesslog.proto#data-accesslog-v2-responseflags.

At last, I didn't change anything in the default OAL, all these two features are open to the users. If they want to this, they could add new OAL scripts, and set up the UI.

### DIFF
--- a/docs/en/concepts-and-designs/scope-definitions.md
+++ b/docs/en/concepts-and-designs/scope-definitions.md
@@ -29,6 +29,7 @@ Calculate the metrics data from each request of the service.
 | responseCode | Represent the response code of HTTP response, if this request is the HTTP call | | int|
 | type | Represent the type of each request. Such as: Database, HTTP, RPC, gRPC. | | enum |
 | tags | Represent the labels of each request and each value is made up with the `TagKey:TagValue` in the segment. | | `List<String>` |
+| sideCar.internalErrorCode | Represent the sidecar/proxy internal error code, the value bases on the implementation. | | string|
 
 ### SCOPE `ServiceInstance`
 
@@ -45,6 +46,7 @@ Calculate the metrics data from each request of the service instance.
 | responseCode | Represent the response code of HTTP response, if this request is the HTTP call. | | int |
 | type | Represent the type of each request. Such as: Database, HTTP, RPC, gRPC. | | enum |
 | tags | Represent the labels of each request and each value is made up with the `TagKey:TagValue` in the segment. | | `List<String>` |
+| sideCar.internalErrorCode | Represent the sidecar/proxy internal error code, the value bases on the implementation. | | string|
 
 #### Secondary scopes of `ServiceInstance` 
 
@@ -117,6 +119,7 @@ Calculate the metrics data from each request of the endpoint in the service.
 | responseCode | Represent the response code of HTTP response, if this request is the HTTP call. | | int |
 | type | Represent the type of each request. Such as: Database, HTTP, RPC, gRPC. | | enum |
 | tags | Represent the labels of each request and each value is made up with the `TagKey:TagValue` in the segment. | | `List<String>` |
+| sideCar.internalErrorCode | Represent the sidecar/proxy internal error code, the value bases on the implementation. | | string|
 
 ### SCOPE `ServiceRelation`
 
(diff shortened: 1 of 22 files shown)


oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/LogEntry2MetricsAdapter.java
```diff
@@ -164,4 +169,52 @@ protected static String parseTLS(final TLSProperties properties) {
...
+                return "downstream_connection_termination";
+            } else if (responseFlags.getUpstreamRetryLimitExceeded()) {
+                return "upstream_retry_limit_exceeded";
+            } else if (responseFlags.getStreamIdleTimeout()) {
+                return "stream_idle_timeout";
+            } else if (responseFlags.getInvalidEnvoyRequestHeaders()) {
+                return "invalid_envoy_request_headers";
+            } else if (responseFlags.getDownstreamProtocolError()) {
+                return "downstream_protocol_error";
+            }
```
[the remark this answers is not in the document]
wu-sheng: To @ZackButcher too.

kezhenxu94 (dismissed):
LGTM

docs/en/concepts-and-designs/scope-definitions.md
```diff
@@ -29,6 +29,7 @@ Calculate the metrics data from each request of the service.
 | responseCode | Represent the response code of HTTP response, if this request is the HTTP call | | int|
 | type | Represent the type of each request. Such as: Database, HTTP, RPC, gRPC. | | enum |
 | tags | Represent the labels of each request and each value is made up with the `TagKey:TagValue` in the segment. | | `List<String>` |
+| sideCar.internalErrorCode | Represent the sidecar/proxy internal error code, the value bases on the implementation. | | string|
```
hanahmily, line 32: FWIK, "proxy" is a more general name than `sidecar`. The latter one is often put together with "gateway" as Istio does, and therefore I prefer to use "proxy.internalErrorCode" instead of "sidecar.internalErrorCode".

wu-sheng: In the Istio concept, yes, I agree. But note here, we are adding a description for a service. If this is a gateway, as you mentioned, the service itself is the gateway, rather than this.
Also, for the gateway case, actually, there is no point to indicate whether sidecar issue or not, all errors are gateway's issue.

wu-sheng: As the attachment, I don't want it to be a `proxy`.

hanahmily: > Also, for the gateway case, actually, there is no point to indicate whether sidecar issue or not, all errors are gateway's issue.

Definitely, a gateway is abstracted to service, but it still needs `sidecar.xxxx` to represent this dimension. It looks a little tricky, especially, the remarks emphasize `the sidecar/proxy internal error code`. if that, `the sidecar/gateway proxy internal error code` makes more sense to me.

wu-sheng: Fixed the comments and docs. Others are unchanged.

hanahmily (approved):
LGTM

PR #5851, opened 2020-11-17
Support group in the service traffic and service list query.

### LINKED ISSUE #5841: Add Service Group concept for service filter

- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement

___

SkyWalking has the capability to observe thousands of services, which have different types, such as agent-observed, service-mesh, Istio control panel, Envoy data panel, ShardingSphere proxy(etc.). All of them should have different dashboards due to we enhanced in the past 2 versions.
Now, I intend to add this. 
1. Service Group is a string name-based concept, added to the service list. 
1. Provide a new optional condition(group) in the GraphQL. If provided, use it to filter the service list, otherwise, like we did today.
1. The group name would not be updated due to the ServiceTraffic is `supportUpdate = false`. So, it is only updated(actually added) once the TTL reached and data has been removed.
1. UI should be updated to add the current filter condition in the GraphQL. cc @Jtrust 

If #5835 merged fast, this would be added in the 8.4.0.

This concept is helpful for supported group based topology in the future, but note, this is not the goal of this PR.

wu-sheng:
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #5841. Related to #5835

This feature doc is as below. @hanahmily need to 
1. Adjust #5835 to adopt this as the lastest naming rule. 
1. Update UI's template page. `Service Filter` should be renamed to `Service Group`(including the Chinese version)
1. Update the self-observability dashboard configuration(maybe codes too).

# Service Auto Grouping
SkyWalking supports various default and customized dashboard templates. 
Each template provides the reasonable layout for the services in the particular field. 
Such as, services with a language agent installed 
could have different metrics with service detected by the service mesh observability solution, 
and different with SkyWalking's self-observability metrics dashboard.

Therefore, since 8.3.0, SkyWalking OAP would generate the group based on this simple naming format.

### ${service name} = [${group name}:]${logic name}

Once the service name includes `:`(colon), the literal string before the colon would be considered as the group name.
In the latest GraphQL query, the group name has been provided as an option parameter.
> getAllServices(duration: Duration!, group: String): [Service!]!

RocketBot UI dashboards(`Standard` type) support the `group name` for default and custom configurations.

___

FYI @hailin0 I think you need to consider this in your DolphinScheduler integration. I am not sure whether you would use a new dashboard for the DolphinScheduler Dashboard.

### DIFF
--- a/docs/en/setup/backend/backend-setup.md
+++ b/docs/en/setup/backend/backend-setup.md
@@ -108,6 +108,7 @@ or 3rd party configuration management system.
 1. [Uninstrumented Gateways](uninstrumented-gateways.md). Configure gateways/proxies that are not supported by SkyWalking agent plugins,
 to reflect the delegation in topology graph.
 1. [Apdex threshold](apdex-threshold.md). Configure the thresholds for different services if Apdex calculation is activated in the OAL.
+1. [Service Grouping](service-auto-grouping.md). An automatic grouping mechanism for all services based on name..
 1. [Group Parameterized Endpoints](endpoint-grouping-rules.md). Configure the grouping rules for parameterized endpoints,
 to improve the meaning of the metrics.
 1. [Meter Analysis](backend-meter.md). Set up the backend analysis rules, when use [SkyWalking Meter System Toolkit](../service-agent/java-agent/README.md#advanced-features) 
--- a/docs/en/setup/backend/backend-setup.md
+++ b/docs/en/setup/backend/backend-setup.md
@@ -108,7 +108,7 @@ or 3rd party configuration management system.
 1. [Uninstrumented Gateways](uninstrumented-gateways.md). Configure gateways/proxies that are not supported by SkyWalking agent plugins,
 to reflect the delegation in topology graph.
 1. [Apdex threshold](apdex-threshold.md). Configure the thresholds for different services if Apdex calculation is activated in the OAL.
-1. [Service Grouping](service-auto-grouping.md). An automatic grouping mechanism for all services based on name..
+1. [Service Grouping](service-auto-grouping.md). An automatic grouping mechanism for all services based on name.
 1. [Group Parameterized Endpoints](endpoint-grouping-rules.md). Configure the grouping rules for parameterized endpoints,
 to improve the meaning of the metrics.
 1. [Meter Analysis](backend-meter.md). Set up the backend analysis rules, when use [SkyWalking Meter System Toolkit](../service-agent/java-agent/README.md#advanced-features) 
--- a/docs/en/setup/backend/service-auto-grouping.md
+++ b/docs/en/setup/backend/service-auto-grouping.md
@@ -1,0 +1,16 @@
+# Service Auto Grouping
+SkyWalking supports various default and customized dashboard templates. 
+Each template provides the reasonable layout for the services in the particular field. 
+Such as, services with a language agent installed 
+could have different metrics with service detected by the service mesh observability solution, 
+and different with SkyWalking's self-observability metrics dashboard.
+
+Therefore, since 8.3.0, SkyWalking OAP would generate the group based on this simple naming format.
+
+### ${service name} = [${group name}:]${logic name}
+
+Once the service name includes `:`(colon), the literal string before the colon would be considered as the group name.
+In the latest GraphQL query, the group name has been provided as an option parameter.
+> getAllServices(duration: Duration!, group: String): [Service!]!
+
+RocketBot UI dashboards(`Standard` type) support the `group name` for default and custom configurations.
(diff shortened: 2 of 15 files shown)


docs/en/setup/backend/service-auto-grouping.md
```diff
@@ -0,0 +1,16 @@
+# Service Auto Grouping
+SkyWalking supports various default and customized dashboard templates. 
+Each template provides the reasonable layout for the services in the particular field. 
+Such as, services with a language agent installed 
+could have different metrics with service detected by the service mesh observability solution, 
+and different with SkyWalking's self-observability metrics dashboard.
+
+Therefore, since 8.3.0, SkyWalking OAP would generate the group based on this simple naming format.
+
+### ${service name} = [${group name}:]${logic name}
```
hanahmily, line 10: Why not double ":", typo?

wu-sheng: Typo.

docs/en/setup/backend/backend-setup.md
```diff
@@ -108,6 +108,7 @@ or 3rd party configuration management system.
 1. [Uninstrumented Gateways](uninstrumented-gateways.md). Configure gateways/proxies that are not supported by SkyWalking agent plugins,
 to reflect the delegation in topology graph.
 1. [Apdex threshold](apdex-threshold.md). Configure the thresholds for different services if Apdex calculation is activated in the OAL.
+1. [Service Grouping](service-auto-grouping.md). An automatic grouping mechanism for all services based on name..
```
dmsolr, line 111: ".." on the end

docs/en/setup/backend/service-auto-grouping.md
wu-sheng: Done.

dmsolr (approved):
LGTM without a minor typo.

docs/en/setup/backend/backend-setup.md
wu-sheng: Done.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/service/ServiceTraffic.java
```diff
@@ -93,14 +101,23 @@ public ServiceTraffic map2Data(final Map<String, Object> dbMap) {
...
+            final String serviceName = storageData.getName();
+            if (NodeType.Normal.equals(storageData.getNodeType())) {
+                int groupIdx = serviceName.indexOf(DOUBLE_COLONS_SPLIT);
+                if (groupIdx > 0) {
+                    storageData.setGroup(serviceName.substring(0, groupIdx));
+                }
+            }
             Map<String, Object> map = new HashMap<>();
-            map.put(NAME, storageData.getName());
+            map.put(NAME, serviceName);
```
hailin0, line 118: need to keep group as a service name prefix?

wu-sheng: Yes, we don't remove it. As the name is a part of metrics key. Changing it will break many things.

hailin0: ok

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/TableMetaInfo.java
```diff
@@ -62,8 +62,9 @@ public static void addModel(Model model) {
                 || EndpointTraffic.INDEX_NAME.equals(model.getName())) {
                 storageAndTagMap.put(EndpointTraffic.SERVICE_ID, InfluxConstants.TagName.SERVICE_ID);
             } else {
-                // service_traffic  name, node_type
+                // service_traffic  name, node_type, group
                 storageAndTagMap.put(ServiceTraffic.NODE_TYPE, InfluxConstants.TagName.NODE_TYPE);
+                storageAndTagMap.put(ServiceTraffic.GROUP, InfluxConstants.TagName.NODE_TYPE);
```
dmsolr, line 67: TagName.GROUP

wu-sheng: Done.

dmsolr (approved):

hanahmily (dismissed):
LGTM. these changes save istio relevant feats, thanks

dmsolr (approved):

PR #5886, opened 2020-11-24
fix gateway plugin async finish repeatedly when fallback configured

### LINKED ISSUE #5814: Can not do async finish for the span repeatedly

- [ ] Question or discussion
- [x] Bug
- [ ] Requirement

___
### Question
- When skywalking is used in spring cloud gateway 2.2.4, the following exceptions will occur in hystrix or timeout
### Bug
- JDK8U201
- SkyWalking8.2.0
- Centos7 
- SpringCloudGateway 2.2.4

2020-11-09 20:20:09.358 ERROR 8 --- [or-http-epoll-6] reactor.core.publisher.Operators         : Operator called default onErrorDropped

java.lang.RuntimeException: Can not do async finish for the span repeatedly.
	at org.apache.skywalking.apm.agent.core.context.trace.AbstractTracingSpan.asyncFinish(AbstractTracingSpan.java:325)
	at org.apache.skywalking.apm.plugin.spring.webflux.v5.DispatcherHandlerHandleMethodInterceptor.lambda$afterMethod$0(DispatcherHandlerHandleMethodInterceptor.java:96)
	at reactor.core.publisher.FluxDoFinally$DoFinallySubscriber.runFinally(FluxDoFinally.java:156)
	at reactor.core.publisher.FluxDoFinally$DoFinallySubscriber.onComplete(FluxDoFinally.java:139)
	at reactor.core.publisher.MonoFlatMap$FlatMapMain.secondComplete(MonoFlatMap.java:189)
	at reactor.core.publisher.MonoFlatMap$FlatMapInner.onComplete(MonoFlatMap.java:260)
	at reactor.core.publisher.Operators$MultiSubscriptionSubscriber.onComplete(Operators.java:2016)
	at reactor.core.publisher.FluxOnAssembly$OnAssemblySubscriber.onComplete(FluxOnAssembly.java:395)
	at reactor.core.publisher.MonoFlatMap$FlatMapMain.secondComplete(MonoFlatMap.java:189)
	at reactor.core.publisher.MonoFlatMap$FlatMapInner.onComplete(MonoFlatMap.java:260)
	at reactor.core.publisher.MonoPeekTerminal$MonoTerminalPeekSubscriber.onComplete(MonoPeekTerminal.java:292)
	at reactor.core.publisher.MonoFlatMap$FlatMapMain.secondComplete(MonoFlatMap.java:189)
	at reactor.core.publisher.MonoFlatMap$FlatMapInner.onComplete(MonoFlatMap.java:260)
	at reactor.core.publisher.MonoIgnoreElements$IgnoreElementsSubscriber.onComplete(MonoIgnoreElements.java:81)
	at reactor.core.publisher.FluxConcatArray$ConcatArraySubscriber.onComplete(FluxConcatArray.java:191)
	at reactor.core.publisher.FluxContextStart$ContextStartSubscriber.onComplete(FluxContextStart.java:115)
	at reactor.core.publisher.MonoFlatMap$FlatMapMain.secondComplete(MonoFlatMap.java:189)
	at reactor.core.publisher.MonoFlatMap$FlatMapInner.onComplete(MonoFlatMap.java:260)
	at reactor.netty.FutureMono$FutureSubscription.operationComplete(FutureMono.java:180)
	at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:577)
	at io.netty.util.concurrent.DefaultPromise.notifyListeners0(DefaultPromise.java:570)
	at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:549)
	at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:490)
	at io.netty.util.concurrent.DefaultPromise.setValue0(DefaultPromise.java:615)
	at io.netty.util.concurrent.DefaultPromise.setSuccess0(DefaultPromise.java:604)
	at io.netty.util.concurrent.DefaultPromise.trySuccess(DefaultPromise.java:104)
	at io.netty.util.internal.PromiseNotificationUtil.trySuccess(PromiseNotificationUtil.java:48)
	at io.netty.channel.ChannelOutboundBuffer.safeSuccess(ChannelOutboundBuffer.java:717)
	at io.netty.channel.ChannelOutboundBuffer.remove(ChannelOutboundBuffer.java:272)
	at io.netty.channel.ChannelOutboundBuffer.removeBytes(ChannelOutboundBuffer.java:352)
	at io.netty.channel.epoll.AbstractEpollChannel.doWriteBytes(AbstractEpollChannel.java:363)
	at io.netty.channel.epoll.AbstractEpollStreamChannel.writeBytes(AbstractEpollStreamChannel.java:260)
	at io.netty.channel.epoll.AbstractEpollStreamChannel.doWriteSingle(AbstractEpollStreamChannel.java:471)
	at io.netty.channel.epoll.AbstractEpollStreamChannel.doWrite(AbstractEpollStreamChannel.java:429)
	at io.netty.channel.AbstractChannel$AbstractUnsafe.flush0(AbstractChannel.java:930)
	at io.netty.channel.epoll.AbstractEpollChannel$AbstractEpollUnsafe.flush0(AbstractEpollChannel.java:519)
	at io.netty.channel.AbstractChannel$AbstractUnsafe.flush(AbstractChannel.java:897)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.flush(DefaultChannelPipeline.java:1372)
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:750)
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush(AbstractChannelHandlerContext.java:742)
	at io.netty.channel.AbstractChannelHandlerContext.flush(AbstractChannelHandlerContext.java:728)
	at io.netty.channel.CombinedChannelDuplexHandler$DelegatingChannelHandlerContext.flush(CombinedChannelDuplexHandler.java:531)
	at io.netty.channel.ChannelOutboundHandlerAdapter.flush(ChannelOutboundHandlerAdapter.java:125)
	at io.netty.channel.CombinedChannelDuplexHandler.flush(CombinedChannelDuplexHandler.java:356)
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:750)
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush(AbstractChannelHandlerContext.java:742)
	at io.netty.channel.AbstractChannelHandlerContext.flush(AbstractChannelHandlerContext.java:728)
	at io.netty.channel.ChannelDuplexHandler.flush(ChannelDuplexHandler.java:127)
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:750)
	at io.netty.channel.AbstractChannelHandlerContext.invokeWriteAndFlush(AbstractChannelHandlerContext.java:765)
	at io.netty.channel.AbstractChannelHandlerContext$WriteTask.run(AbstractChannelHandlerContext.java:1071)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:164)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:472)
	at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:384)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.lang.Thread.run(Thread.java:748)

yujiaxinlong:
### Fix gateway plugin async finish repeatedly when fallback configured #5814 
if fallback url is configured in springcloud gateway, 
image

when a request is timeout or other exceptions are thrown, hystrix will trigger another request like /fallback/default, this invokes webflux DispatcherHandler#handle() again, and the ServerWebExchange here is the same nested one with the previous request. 

A new span will be put in ServerWebExchange attribute, then doFinally will work twice, both asyn finish the later fallback span, thus lead to the bug. The span for actual request will never be asyn finished.

Move the get span code out of mono fix this bug easily.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #5814 


wu-sheng:
Please update the changelog.

wu-sheng:
Closing as no update.

yujiaxinlong:
@wu-sheng  
kind of forgot about this PR,  this issue seems still exist in latest version, is there anything I can do or just waiting for review of 
@ascrutae

wu-sheng (approved):

wu-sheng:
> @wu-sheng
> kind of forgot about this PR, this issue seems still exist in latest version, is there anything I can do or just waiting for review of
> @ascrutae

I synced your codes, and open CIs to run.

wu-sheng (changes requested):
The plugin test fails, could you check the expected data with your codes?

wu-sheng:
@yujiaxinlong Could you take a look?

yujiaxinlong:
@wu-sheng  I will fix this asap

wu-sheng:
image

Still fail. @yujiaxinlong

wu-sheng:
Webflux case fails.

yujiaxinlong:
@wu-sheng  got stucked when aoviding the orginal url overwritten by fallback url , should works fine now

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/DispatcherHandlerHandleMethodInterceptor.java
```diff
@@ -77,13 +77,14 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
...
     }
     
-    private void setPattern(AbstractSpan span, ServerWebExchange exchange) {
+    private void maybeSetPattern(AbstractSpan span, ServerWebExchange exchange) {
         if (span != null) {
-            Object pathPattern = exchange.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE);
-            if (pathPattern != null) {
-                span.setOperationName(((PathPattern) pathPattern).getPatternString());
+            PathPattern pathPattern = exchange.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE);
+            if (pathPattern != null && pathPattern.matches(exchange.getRequest().getPath().pathWithinApplication())) {
```
wu-sheng, line 83: What does `pathPattern.matches(exchange.getRequest().getPath().pathWithinApplication())` mean? Could you add comments about this?

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/DispatcherHandlerHandleMethodInterceptor.java
```diff
@@ -93,14 +94,11 @@ public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allA
 
         AbstractSpan span = (AbstractSpan) exchange.getAttributes().get("SKYWALING_SPAN");
         
-        return ((Mono) ret).flatMap(s -> {
-                    setPattern(span, exchange);
-                    return s;
-                 })
-                .doOnError(s -> setPattern(span, exchange))
+                return ((Mono) ret)
```
wu-sheng, line 97: Is this the right format? I am feeling not. Could you confirm? Rather than waste CI resources.

yujiaxinlong: slightly changed code style so to doesn't looks weird now。

yujiaxinlong:
added comment， it might a little bit lengthy to explain the best matching thing, so I am going to talk about it briefly here.

The short version is that without this check, the span of origin request will have url of fallback as it's name.

When there is a fallback in webflux( hystrix fallback or a webflux fallback method that involve visiting a new url).  This interceptor would be invoked twice, one for origin request and another one for fallback request.  The ServerWebExchange for these two aren't exactly the same one, but the former one is a delegate inside of the latter one, thus, they **share the same attributes.**  

The problem of the old code is that get the span from attributes within the execution of Mono would get span for latter request twice.   so just move this part of code out could solve the problem.

Then there is a new question about the operation name. Because  _BEST_MATCHING_PATTERN_ATTRIBUTE_  is also in attributes,  leave the setOperationName in  Mono make both two spans use the fallback url as operationName. But move these codes out would change the recording result of requests with path parameters, from /annotation/{test} to /annotation/foo or /annotation/loo with the parameter shown because the best matching pattern is calculated after mono are subscribed.  And that's what makes the plugin test fail before.

I tried several ways on  adjusting the order of execution  to solve this,  but it seems imposibble to make the mono chain work like expected. So instead I check whether the best matching pattern really match the url.  This check code is how webflux generating the best match pattern, so should not lead to any problem.

wu-sheng:
> The problem of the old code is that get the span from attributes within the execution of Mono would get span for latter request twice. so just move this part of code out could solve the problem.

That makes sense. My another question(not about this fix), should we tag the failback behavior with span's logs? Every log could include one-time try behavior. What do you think? Of course, this should be done in another pull request.

> I tried several ways on adjusting the order of execution to solve this, but it seems imposibble to make the mono chain work like expected. So instead I check whether the best matching pattern really match the url. This check code is how webflux generating the best match pattern, so should not lead to any problem.

That is good enough.

wu-sheng (approved):

yujiaxinlong:
> That makes sense. My another question(not about this fix), should we tag the failback behavior with span's logs? Every log could include one-time try behavior. What do you think? Of course, this should be done in another pull request.
> 

you mean just put the fallback behavior in the span of origin request rather than in a separate span? 
I think that's might be better for most cases , since most of the fallback just do very simple work and return.  Just need to find a way to distinguish between normal requests and fallback requests.

Another benefit  is that this way we might be able to mark the origin request with a fallback as errorOccurred. (not sure if this is a good idea)

wu-sheng:
> you mean just put the fallback behavior in the span of origin request rather than in a separate span?
> I think that's might be better for most cases , since most of the fallback just do very simple work and return. Just need to find a way to distinguish between normal requests and fallback requests.

Yes, for most case, only one exit span should be generated to save the cost of memory and GC, due to new spans are meaningless. The failback actually is a part of the original exit span's behaviors. 

> 
> Another benefit is that this way we might be able to mark the origin request with a fallback as errorOccurred. (not sure if this is a good idea)

We should, if fallback eventually went into fail.

PR #5908, opened 2020-11-26
Include SAN to set mTLS property

### LINKED ISSUE #5907: Including SAN to set the mTLS property of an ALS event

- [ ] Question or discussion
- [x] Bug
- [ ] Requirement

___
### Bug
`subject_alt_name` should be included in below pieces.

https://github.com/apache/skywalking/blob/9c486a05ee3ad8947011168e0039f5b48922e4ab/oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/LogEntry2MetricsAdapter.java#L155-L169

hanahmily:
### Fix #5907 
- [x] CHANGES.md

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -57,6 +57,7 @@ Release Notes.
 * Add otel rules to ui template to observe Istio control plane.
 * Remove istio mixer
 * Support close influxdb batch write model.
+* Check SAN in the ALS (m)TLS process.
 
 #### UI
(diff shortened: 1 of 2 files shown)


wu-sheng (approved):

wu-sheng (changes requested):
Seems to having a compiling issue?

wu-sheng (approved):

Issue #5904, opened 2020-11-26
[UI Enhancement] Elasticsearch exception type=index_not_found_exception


mritd:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [x] Feature or performance improvement
___
### Requirement or improvement
- Please describe your requirements or improvement suggestions.

When using ElasticSearch as the storage backend, no data on the date selected by the UI will cause oap-server to throw a large number of `index_not_found_exception` exceptions, and the UI interface will not display any error prompts.

![image](https://user-images.githubusercontent.com/13043245/100319411-84b95500-2ffa-11eb-9918-aa4fbc05f47c.png)

![image](https://user-images.githubusercontent.com/13043245/100319447-93077100-2ffa-11eb-8874-a42af847c35d.png)

**Therefore, should the UI date selection component restrict the selectable dates to ensure that the user selects a date without data and the aop server cannot return data?**

wu-sheng:
@EvanLjp Could you recheck? I remember we ignore the case `index not exist` in the codes.

EvanLjp:
I would recheck it today, thx for your notice @mritd

mritd:
@EvanLjp Thanks for your reply, my operating environment:

OS: Kubernetes
oap-server: `apache/skywalking-oap-server:8.2.0-es7`
skywalking-ui: `apache/skywalking-ui:8.2.0`
Elasticsearch: `docker.elastic.co/elasticsearch/elasticsearch:7.9.2`

EvanLjp:
@mritd I'm sorry that I only added the codes in the ES6 client and ignored the ES7 client. Thx a lot.

mritd:
@EvanLjp Okay, I will wait 8.3.0 fixed. 😁

PR #5911, opened 2020-11-27
fix index not found on trace query when using es7

EvanLjp:
the bug issue 
https://github.com/apache/skywalking/issues/5904


wu-sheng (approved):
LGTM.

kezhenxu94 (approved):

PR #5915, opened 2020-11-27
Make meter receiver support MAL

mrproliu:
### Make agent side meter receiver support MAL


wu-sheng:
FYI @ascrutae @hailin0 as you are doing something relating to this.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/SampleFamily.java
```diff
@@ -233,12 +234,29 @@ public SampleFamily tag(Closure<?> cl) {
         }).toArray(Sample[]::new));
     }
 
+    /**
+     * Ignore histogram decrease bucket value, at agent side meter, need to ignore decrease histogram bucket value
+     */
+    public SampleFamily ignoreHistogramDecrease() {
```
hanahmily, line 240: From https://en.wikipedia.org/wiki/Histogram#Mathematical_definition, there are two types of histogram bin:
 * Cumulative: prometheus, otel.
 * Ordinary: agent meter.

Let's provide two functions to set it, and defaults to cumulative.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/SampleFamily.java
```diff
@@ -233,12 +234,29 @@ public SampleFamily tag(Closure<?> cl) {
...
+        this.context.ignoreHistogramDecrease = true;
+        return this;
+    }
+
+    /**
+     * Change default histogram bucket value unit,
+     * use second on the Prometheus, use millisecond on the meter side meter
+     */
+    public SampleFamily defaultHistogramBucketUnit(TimeUnit unit) {
```
hanahmily, line 249: My preference is to add a new factory to build SampleFamily with those two settings. If we define functions in SampleFamily, that means end-users could access them, that will cause some potential issues. 

and they represent the different behaviors of the eco-system, I don't find any reason to access them in a MAL expression.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/SampleFamily.java
hanahmily: What I mean is `histogramDecrease` is not clear and very confusing to ppl. Let's opt for the notions that is well-known for ppl.

mrproliu: Created `HistogramType` enum to adapt this.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/SampleFamily.java
mrproliu: Have created `SampleFactoryBuilder` to help create `SampleFactory` with these two settings.

oap-server/analyzer/agent-analyzer/pom.xml
```diff
@@ -39,8 +39,9 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.codehaus.groovy</groupId>
-            <artifactId>groovy</artifactId>
+            <groupId>org.apache.skywalking</groupId>
```
kezhenxu94, line 42: this is the only place where groovy is used? if so, you need to update/remove the license and the dependency in the `oap-server/pom.xml`

wu-sheng:
@mrproliu Could you consider to write a blog about the MAL?

mrproliu:
> @mrproliu Could you consider to write a blog about the MAL?

Sure, I may start at the weekend, at that time, I will synchronize the progress with you.

hanahmily (approved):
LGTM, nice work @mrproliu

PR #5917, opened 2020-11-27
Fix okhttp-3.x and async-http-client-2.x did not overwrite the old trace header.

zifeihan:
### Fix okhttp-3.x and async-http-client-2.x did not overwrite the old trace header.

Fix okhttp-3.x and async-http-client-2.x did not overwrite the old trace header, May cause data confusion in some scenarios such as gateway

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ Release Notes.
 
 #### Java Agent
 
+
 #### UI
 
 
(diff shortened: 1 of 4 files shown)


wu-sheng (commented):
How about let's make this a part of 8.3.0 release? If so, update the changelog in the `changes` folder.

zifeihan:
> How about let's make this a part of 8.3.0 release? If so, update the changelog in the `changes` folder.

done.

wu-sheng (approved):
LGTM. Please send the feedback to the mail list.

PR #5921, opened 2020-11-29
Fix illegal istio-controlplane.yaml

hanahmily:
Remove redundant fields.


wu-sheng (approved):
I will cherry-pick this to the `8.3.0-release-fix` branch.

PR #5932, opened 2020-12-01
Fix influx response format

### LINKED ISSUE #5931: influxdb response format error
###Requirement

we use influxdb srelay(https://github.com/toni-moreno/influxdb-srelay) for influxdb ha.but influxdb-sreay not support msgpack protocol only support JSON format.

image

zhangjianweibj:
Fix https://github.com/apache/skywalking/issues/5931

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #5931.


docs/en/setup/backend/configuration-vocabulary.md
```diff
@@ -138,6 +138,7 @@ core|default|role|Option values, `Mixed/Receiver/Aggregator`. **Receiver** mode
 | - | - | database | Database of InfluxDB. | SW_STORAGE_INFLUXDB_DATABASE | skywalking |
 | - | - | actions | The number of actions to collect. | SW_STORAGE_INFLUXDB_ACTIONS | 1000 |
 | - | - | duration | The time to wait at most (milliseconds). | SW_STORAGE_INFLUXDB_DURATION | 1000|
+| - | - | responseFormat | Influxdb response format option. | SW_STORAGE_INFLUXDB_RESPONSE_FORMAT | MSGPACK|
```
wu-sheng, line 141: What options do they have? Could you add a reference link?

wu-sheng:
One question for the context about the repo, https://github.com/toni-moreno/influxdb-srelay. Is this under maintaining fork version?
It makes me feel strange, why you are using a fork version of a fork version. What is the status of https://github.com/influxdata/influxdb-relay?

The reason I am asking is, as an Apache project, we usually only support widely used thing, the 2 PRs you submitted are acceptable, so we could merge. But if this kind of thing keeps happening, I have concerns we will face a conflict.

Could you explain more to make us following the context,?

docs/en/setup/backend/configuration-vocabulary.md
zhangjianweibj: https://javadoc.io/static/org.influxdb/influxdb-java/2.17/org/influxdb/InfluxDB.ResponseFormat.html

https://github.com/influxdata/influxdb/issues/8330

zhangjianweibj:
https://github.com/influxdata/influxdb-relay is not a cluster solution,we used for a year and encountered many serious problems (eg two influx nodes frequent data loss,if one node breakdown then start,data will be permanently missing ).

docs/en/setup/backend/configuration-vocabulary.md
wu-sheng: I find some issues may affect the SkyWalking, such as 
> JSON does not support integers above 2^53

SkyWalking has many long fields, which in Java could be 2^63-1

wu-sheng:
> https://github.com/influxdata/influxdb-relay is not a cluster solution, we used for a year and encountered many serious problems

My point, and concerns from ASF TLP, are both whether the project you choose is really widely used? And who and which community could make the project usable and maintainable for a long time?

zhangjianweibj:
yeah,thanks for your suggestion.before skywalking project,we used prometheus with influxdb many years.during this period,influxdb-relay encountered many serious problems.we hava to look for alternatives solution(uber/m3,victoriaMetrics,srelay).as your point,srelay maintenance is a important issue.there are not many maintainers of the project so far.but influxdb with srelay is a reasonable solution util now.this solution also presented to the community.prometheus has many reasonable solution like m3,victoriaMetrics,thanos.we tried various skywalking storage plugins.tsdb is the most reasonable option so far.

wu-sheng:
> prometheus has many reasonable solution like m3,victoriaMetrics,thanos.we tried various skywalking storage plugins.tsdb is the most reasonable option so far.

For Prometheus, yes, TSDB makes sense. But honestly, I am not sure whether TSDB is the best option for SkyWalking.

wu-sheng:
> srelay maintenance is a important issue

This would be a big issue, we can't accept to make all things configurable but just for a lack of maintenance project.

docs/en/setup/backend/configuration-vocabulary.md
wu-sheng: @dmsolr told me, there would be issues when using JSON. Same reason as the above number format issue.

zhangjianweibj:
not only srelay use JSON format as response format.native influxdb also support JSON format.

zhangjianweibj:
elasticsearch is not a reasonable solution for skywalking.agent may push many metrics,each metric has a indics with timestamp 'yyyy-mm-dd'.so many indics and shards that elasticsearch can not manage.on the other hand,the reasonable size of shards in es is 20G-50G. and skywalking storage size is difficult to estimate.how to set configuration of elasticsearch index?
skywalking used many time series data.obviously this data is not suitable for storage in a realtional database like pgsql,h2.

wu-sheng:
> elasticsearch is not a reasonable solution for skywalking.agent may push many metrics,each metric has a indics with timestamp 'yyyy-mm-dd'.so many indics and shards that elasticsearch can not manage

I am not sure why you say ES can't manage them. 100+TB even PB+ level storages are common the user environments, as far as I know.  They may try to find the way to make the balance better in the cluster, but don't have the management issue.

> not only srelay use JSON format as response format.native influxdb also support JSON format.

I was not blaming using JSON. I know InfluxDB is supporting it. My feedback is from the original author of this feature, @dmsolr , he told me there would be issues using this format.

zhangjianweibj:
image

@dmsolr hello,we deploy skywalking with influxdb response json format,but no any error.

dmsolr:
Hi @zhangjianweibj , thanks!
I am meeting many strange issues, such as typecast error, during developed it.
In my idea, I don't really recommend that. Because none will configure this normally.
Finally, you have to make sure it works well except for errors in logs.

wu-sheng:
@zhangjianweibj Could you provide a recorded video to verify all things work? We have several e2e, but they don't cover 100% cases.

zhangjianweibj:
unit tests and integration test are designed to check for such problems.as a apache project,it is hard to accept checking this feature by a recorded video.
very thanks,i will close this pr.

wu-sheng:
Sorry, we have to check. Manual tests are always the last option. We have thousands of tests, which definitely helps but we can't say it is perfect. Honestly, I don't understand why a recorded video is an issue, as it is an easy way to check all UI components work well. If we asked you to fill all missing tests in the InfluxDB cases, you would have to suffer more workloads.
But anyway, we respect your choice.

Issue #5945, opened 2020-12-03
If set influxdb connection response format as InfluxDB.ResponseFormat.JSON，there will be some error


cuiweiwei:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
If set influxdb connection response format as InfluxDB.ResponseFormat.JSON，the start time of trace will always be 1970-01-01 08:00:00，also cannot see details of profiled trace segments.

___
### Bug
- Which version of SkyWalking, OS and JRE?
Skywalking version：8.2.0.RELEASE
OS：linux or windows

- What happened?
When use influxdb as storage of Skywalking, due to certain limitations, only can use json as influxdb connection response format .Some errors will occur, as follows.
![image](https://user-images.githubusercontent.com/15830188/101014244-8659be80-35a0-11eb-8fac-fe774d52cb8b.png)

wu-sheng:
Read #5932. That PR we asked for proving is knowing this could happen.

wu-sheng:
Quick answer is, don't use JSON mode.

PR #5946, opened 2020-12-03
fix influxdb bug while setting influxdb connection response format as InfluxDB.ResponseFormat.JSON

cuiweiwei:
### Fix <#5945>
When use influxdb as storage of Skywalking, due to certain limitations, only can use json as influxdb connection response format .Some errors will occur as show in issue.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<#5945>.


wu-sheng (commented):
Read the PR I pointed to you. There is more details besides this code change.

cuiweiwei:
When set json as influxdb response format, it will use scientific counting to convert long type to double, but BasicTrace use String type as its startTime, then error occured. The changed code in this pr can solve this problem, and it doesn't affect anything while using msgpack.

wu-sheng:
Could you make CI passed?

wu-sheng:
I want to remind you is this,

> JSON does not support integers above 2^53
SkyWalking has many long fields, which in Java could be 2^63-1

wu-sheng:
How do you activate the JSON format? I remember there is hardcode about `msgpack`. Please recheck.

cuiweiwei:
I have the needs to activate json ,maybe other people have the same need，so i changed the code and re-packaged the project to build image. To say the least, even if someone uses msgpack instead of json, the changed code in this pr will have a positive effect. Compare TraceQuery class to ProfileThreadSnapshotQuery class, the set method of SegmentRecord takes different params, perhaps the author was negligent at the time.

wu-sheng:
To be more clear, I mean this, https://github.com/apache/skywalking/blob/6ff562334044446657ba4df44b98b5f3897864ef/oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/InfluxClient.java#L78

If this is hard coded, how JSON format works?

cuiweiwei:
I know you mean here. However, users may find that the default implementation will fail when they use a special version of influxDB, and error logs finally point to this : InfluxDB.ResponseFormat.MSGPACK. Then they find InfluxDBFactory.connect method support both JSON and MSGPACK, and JSON is even the default. so they changed MSGPACK to JSON，error occured. Users may wonder, clearly org. Influxdb. InfluxDBFactory support json, why is wrong

wu-sheng:
My question is, is JSON format working already? If so, this PR should make that part configurable like the previous I linked. And related doc should be updated too.

cuiweiwei:
Yes, according to my tests, it is ok. I have made that part configurable.

wu-sheng:
Same as before, we need to verify. The easiest way to uploading the screenshots of every pages, or a recorded videos.

cuiweiwei:
image
image
image
image
image
image
image
image
image
image

wu-sheng:
Pending on @dmsolr 's review.

dmsolr (approved):
LGTM, Thanks.

Issue #5952, opened 2020-12-04
[Test] Recheck E2E / Meter


wu-sheng:
This test suddenly becomes very unstable, fails repeated in https://github.com/apache/skywalking/pull/5946/checks?check_run_id=1499245325 and master branch https://github.com/apache/skywalking/runs/1496049540

Both changes have no relationship with this feature. We need to recheck locally and try to find out what is going on.

wu-sheng:
Have been proved by #5955, this is a so11y e2e error.

PR #5954, opened 2020-12-05
[TEST] Add logs to help locate meter e2e failed reason

mrproliu:
follow #5952, add logs to locate e2e failed reason


wu-sheng:
Have been proved by #5955, this is a so11y e2e error.

PR #5955, opened 2020-12-05
test: extract so11y test and add logs

kezhenxu94:


wu-sheng:
@hanahmily We found this in the logs, 

image

By logging in here

image

wu-sheng:
We removed the so11y in the e2e for now. Please check and fix this ASAP. We need to locate the boundary.

test/e2e/e2e-test/src/test/java/org/apache/skywalking/e2e/SO11yE2E.java
```diff
@@ -0,0 +1,136 @@
...
+    protected HostAndPort swWebappHostPort;
+
+    @BeforeAll
+    void setUp() {
+        queryClient(swWebappHostPort);
+    }
+
+    @AfterAll
+    public void tearDown() {
+        trafficController.stop();
```
hanahmily, line 100: @kezhenxu94 this line throws NPE

hanahmily:
After increasing the time range of several metrics, cases seem passed. But in the teardown phase, an NPE is thrown, needing @kezhenxu94 's help.

wu-sheng:
Why time range matters? Could you explain a little?

hanahmily:
> Why time range matters? Could you explain a little?

The key point is the average of the increase of relevant metrics. for example, in a single minute, the increase of level 1 aggregation is 1, 1, 0. The average of them is 0, less than 1.After increasing time range, we might get 2,2,1, the result of average is more than 1.

wu-sheng (approved):

PR #5958, opened 2020-12-06
Fix jdk-http and okhttp-3.x plugin did not overwrite the old trace header.

zifeihan:
### Fix jdk-http and okhttp-3.x plugin did not overwrite the old trace header.
- [ ] Explain briefly about why the bug exists and how to fix it.

Fix jdk-http and okhttp-3.x plugin did not overwrite the old trace header.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@ Release Notes.
 * Chore: adapt `create_source_release.sh` to make it runnable on Linux.
 
 #### Java Agent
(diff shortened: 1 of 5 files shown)


wu-sheng:
Please update the CHANGES.md.

zifeihan:
> Please update the CHANGES.md.

done.

wu-sheng (approved):

PR #5960, opened 2020-12-06
fix springmvc reactive api can't collect HTTP statusCode

### LINKED ISSUE #5905: skywalking 8.3 add  springmvc plugin reactive api support , but that cant collect HTTP statusCode

- [ ] Feature improvement

skywalking 8.3 add  springmvc plugin reactive api [support](https://github.com/apache/skywalking/pull/5360) , but that cant collect HTTP statusCode, hope somebody  can improve it

haoyann:
     the framework change httpCode after `org.apache.skywalking.apm.plugin.spring.mvc.commons.interceptor.AbstractMethodInterceptor`, so modify  `org.apache.skywalking.apm.plugin.spring.mvc.v5.InvokeInterceptor` to collect HTTP statusCode.

- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #5905.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/v5/InvokeInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/v5/InvokeInterceptor.java
@@ -18,13 +18,24 @@
 package org.apache.skywalking.apm.plugin.spring.mvc.v5;
 
 import java.lang.reflect.Method;
+import java.util.List;
+
+import org.apache.skywalking.apm.agent.core.context.CarrierItem;
+import org.apache.skywalking.apm.agent.core.context.ContextCarrier;
 import org.apache.skywalking.apm.agent.core.context.ContextManager;
+import org.apache.skywalking.apm.agent.core.context.tag.Tags;
+import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
+import org.apache.skywalking.apm.agent.core.context.trace.SpanLayer;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+import org.apache.skywalking.apm.network.trace.component.ComponentsDefine;
 import org.apache.skywalking.apm.plugin.spring.mvc.commons.ReactiveRequestHolder;
 import org.apache.skywalking.apm.plugin.spring.mvc.commons.ReactiveResponseHolder;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
 import org.springframework.web.server.ServerWebExchange;
+import reactor.core.publisher.Mono;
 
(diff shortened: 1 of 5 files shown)


apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/v5/InvokeInterceptor.java
```diff
@@ -41,6 +52,23 @@ public void beforeMethod(final EnhancedInstance objInst,
...
+        HttpHeaders headers = exchange.getRequest().getHeaders();
+        while (next.hasNext()) {
+            next = next.next();
+            List<String> header = headers.get(next.getHeadKey());
+            if (header != null && header.size() > 0) {
+                next.setHeadValue(header.get(0));
+            }
+        }
+
+        AbstractSpan span = ContextManager.createEntrySpan(exchange.getRequest().getURI().getPath(), carrier);
```
wu-sheng, line 67: Why are you creating a new span?

haoyann: beacuse spring webflux chang httpCode after org.apache.skywalking.apm.plugin.spring.mvc.commons.interceptor.AbstractMethodInterceptor, so need create new span to collect httpCode.

wu-sheng: This is not the right way to change.  If you create 2 entry span(including the existing one), one request would be counted two, and the latency would be wrong.

haoyann: Yes, this is a problem. And I found that tomcat and springmvc plugin have the same problem. How should we fix. createLocalSpan in the springmvc plugin?

wu-sheng: No, tomcat and springmvc in one thread, the core will ignore the tomcat automatically.

wu-sheng: If one component goes to async mode, use async API to close span in another thread. Read doc, please.

horizonzy: > If one component goes to async mode, use async API to close span in another thread. Read doc, please.

Maybe delivery the threadlocal when the thread switch can solve this problem.

wu-sheng: Please don't guess. We have supported other reactor mode long time ago, there are explicit document in the plugin dev doc to explain the APIs, and we have plugins for async already.

haoyann: Thanks, this is not the right change.

PR #5961, opened 2020-12-06
Add alarm tests in the e2e

### LINKED ISSUE #4530: Add alarm tests in the e2e

mrproliu:
### Fix multiple same scope alarms, create at the same time.
- Add `ruleName` field to the AlarmRecord bean, and let the `ruleName` field be part of the ID.

### Add alarm tests in the e2e
- Support alarm query.
- Check the Alarm webhook. 
- Check silence config.

- [x] resolves #4530 .


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/alarm/AlarmRecord.java
```diff
@@ -46,10 +47,11 @@
     public static final String ID1 = "id1";
     public static final String START_TIME = "start_time";
     public static final String ALARM_MESSAGE = "alarm_message";
+    public static final String RULE_NAME = "rule_name";
 
     @Override
     public String id() {
-        return getTimeBucket() + Const.ID_CONNECTOR + scope + Const.ID_CONNECTOR + id0 + Const.ID_CONNECTOR + id1;
+        return getTimeBucket() + Const.ID_CONNECTOR + scope + Const.ID_CONNECTOR + ruleName + Const.ID_CONNECTOR + id0 + Const.ID_CONNECTOR + id1;
```
wu-sheng, line 54: I think once you have the `ruleName`, `scope` is not necessary? RuleName matches the metrics name already, right?

mrproliu: That's true, deleted.

wu-sheng (approved):
The new alarm rule looks good to me. E2E is pending on @kezhenxu94 's review.

PR #5987, opened 2020-12-10
Fix logger-plugin module directory structure problem(Maven cannot trigger junit testing).

zifeihan:
### Fix Fix logger-plugin module directory structure problem(Maven cannot trigger junit testing).

@vcjmhg 
Fix logger-plugin module directory structure problem(Maven cannot trigger junit testing), and format the test code.


CHANGES.md
```diff
@@ -12,6 +12,7 @@ Release Notes.
 * The operation name of quartz-scheduler plugin, has been changed as the `quartz-scheduler/${className}` format.
 * Fix jdk-http and okhttp-3.x plugin did not overwrite the old trace header.
 * Support collecting logs of log4j, log4j2, and logback in the tracing context with a new `logger-plugin`.
+* Fix logger-plugin module directory structure problem(Maven cannot trigger junit testing).
```
```suggestion
```

This fix is not related to the changelog, as the bug codes were introduced in the master only. Please remove.

wu-sheng (dismissed):

wu-sheng (approved):

wu-sheng (approved):

apm-sniffer/optional-plugins/logger-plugin/src/test/java/org/apache/skywalking/apm/plugin/logger/ContextConfigDefaultTest.java
```diff
@@ -24,7 +25,7 @@
 public class ContextConfigDefaultTest {
     @Test
     public void testDefaultConfig() {
-        ContextConfig config = ContextConfig.getInstance();
+        ContextConfig config = ContextConfig.getLatestConfig();
```
vcjmhg, line 28: @zifeihan Why we should write another method named `getLastsConfig()` for UT. It seems that when any UT starts, `initContextConfig()` will be executed. Just curious :smile:

zifeihan: @vcjmhg Hello, because maven test is in a jvm instance, the loading configuration file content is in singleton mode, so if the file content is modified in the test case, it will not take effect.

zifeihan: @vcjmhg you can test success for every UT by manual, but it will failed in maven automatic test.

apm-sniffer/optional-plugins/logger-plugin/src/main/java/org/apache/skywalking/apm/plugin/logger/ContextConfig.java
```diff
@@ -58,6 +58,10 @@ public static ContextConfig getInstance() {
         return HolderContextConfig.INSTANCE;
     }
 
+    public static ContextConfig getLatestConfig() {
```
wu-sheng, line 61:
```suggestion
    /**
     * For testing use only.
     * Avoid to use config cache. Load the configuration every time. 
     */
    static ContextConfig getLatestConfig() {
```

Should like this?

apm-sniffer/optional-plugins/logger-plugin/src/test/java/org/apache/skywalking/apm/plugin/logger/ContextConfigDefaultTest.java
vcjmhg: ok , I got it. Thank u!

wu-sheng (approved):

PR #5989, opened 2020-12-10
Fix thrift trace broken and wrong arg collected.

ZS-Oliver:
fix thrift

bug1:
For methods without parameters, the collected ARGS are not correct,
For example, hi() will become hi）

bug2:
If services A, B and C communicate via RPC
like A -> B -> C
A, C mounts agent, B does not mount agent
The phenomenon is that A has trace, B and C don't
Fix broken link, Now C can also trace


dmsolr:
Thanks, @ZS-Oliver .
Please fix the checkstyle first.

wu-sheng:
@ZS-Oliver 2 things first
1. The code style control file is in the root folder.
1. Update the changelog(CHANGES.md) to describe what you did.

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/wrapper/ServerInProtocolWrapper.java
```diff
@@ -57,10 +64,11 @@ public void initial(AbstractContext context) {
     @Override
     public TField readFieldBegin() throws TException {
         final TField field = super.readFieldBegin();
+        Map<String, String> header = new HashMap<>(1);
```
dmsolr, line 67: The variable is visible in the if statement block. So we don't need to move it out of the if statement block.

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/wrapper/ServerInProtocolWrapper.java
```diff
@@ -81,6 +90,20 @@ public TField readFieldBegin() throws TException {
             }
             return readFieldBegin();
         }
+        if (field.type == TType.STOP && !HAVE_CREATED.get()) {
```
dmsolr, line 93: Would you try to do it without `ThreadLocal`?

dmsolr: I don't recommend to use `ThreadLocal` here.

wu-sheng: @ZS-Oliver Could you explain why do you need this? Are there many wrapper instances or something?

wu-sheng:
@ZS-Oliver Any update?

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/wrapper/ServerInProtocolWrapper.java
ZS-Oliver: sorry，I took leave on Friday and didn't go to work，I will update on Monday.  happy weekend !

ZS-Oliver: I saw ServerInProtocolWrapper was initialized in TServerInterceptor.onConstruct.
The method onConstruct was called when an intercept construction method is called.
I use ThreadLocal to do the isolation, because I think this behavior will only be invoked once during the enhancement.
Are there any better way to implement this ?

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/wrapper/ServerInProtocolWrapper.java
ZS-Oliver: done

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/wrapper/ServerInProtocolWrapper.java
dmsolr: LGTM, Thanks @ZS-Oliver

dmsolr:
Leave a description about this pr did in `changes.md`, please.

ZS-Oliver:
> Leave a description about this pr did in `changes.md`, please.

ZS-Oliver:
sorry, I made a mistake.

ZS-Oliver:
> Leave a description about this pr did in `changes.md`, please.

done

dmsolr (approved):
LGTM

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/wrapper/ServerInProtocolWrapper.java
```diff
@@ -45,6 +45,7 @@
     private static final ILog LOGGER = LogManager.getLogger(ServerInProtocolWrapper.class);
     private static final StringTag TAG_ARGS = new StringTag("args");
     private AbstractContext context;
+    private static ThreadLocal<Boolean> HAVE_CREATED = ThreadLocal.withInitial(() -> Boolean.FALSE);
```
wu-sheng, line 48: TracingContext has native running context for this case, could you consider to use that? It could be auto cleared when the tracing context finished.

wu-sheng (changes requested):

CHANGES.md
```diff
@@ -12,6 +12,7 @@ Release Notes.
 * The operation name of quartz-scheduler plugin, has been changed as the `quartz-scheduler/${className}` format.
 * Fix jdk-http and okhttp-3.x plugin did not overwrite the old trace header.
 * Support collecting logs of log4j, log4j2, and logback in the tracing context with a new `logger-plugin`.
+* Fix thrift plugin trace link broken when intermediate service does not mount agent, and collect wrong args when method without parameters.
```
```suggestion
* Fix thrift plugin trace link broken when intermediate service does not mount agent
* Fix thrift plugin collects wrong args when the method without parameter.
```

wu-sheng: 2lines of changelog seem more clear. Agree?

wu-sheng:
I just updated the title of this PR. Please note, the title will become git commit log, so make it meaningful.

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/wrapper/ServerInProtocolWrapper.java
ZS-Oliver: It seems like the life cycle of a TraceContext is exactly one trace. Thinking of such a circumstance that we have A, B, C three services and they are inter-communicated through RPC like A->B->C, with both A and C mounted the Agent while B not. Then, in the above situation, C will start a new trace since the trace from A is stopped at B. Hence, C could not obsever A with the use of TraceContext.

CHANGES.md
ZS-Oliver: 👍

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/wrapper/ServerInProtocolWrapper.java
wu-sheng: How is your description relating to what I asked? I am talking about don't your own threadlocal, use the one core provided. Do you misunderstand somehow?

dmsolr: SkyWalking provides a safe `RuntimeContext` which is alive during a tracing. So we prefer to use this rather than directly ThreadLocal. 
More details see `ContextManager#getRuntimeContext()`, please.

wu-sheng:
@ZS-Oliver Are you still following this?

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/wrapper/ServerInProtocolWrapper.java
ZS-Oliver: Sorry,  I misunderstand in some aspect.
Thanks for your advice.

ZS-Oliver:
> @ZS-Oliver Are you still following this?

done.

wu-sheng:
@ZS-Oliver You missed keeping this change, https://github.com/apache/skywalking/pull/5989/commits/c87bf69677ab9d7a7c96347da6707426aba8a762#r542432841. I add it back.

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/wrapper/ServerInProtocolWrapper.java
```diff
@@ -81,6 +84,24 @@ public TField readFieldBegin() throws TException {
             }
             return readFieldBegin();
         }
+        if (field.type == TType.STOP) {
+            boolean haveCreatedSpan =
+                    (boolean) ContextManager.getRuntimeContext().get(HAVE_CREATED_SPAN);
```
wu-sheng, line 89: I think there is an NPE risk.
```suggestion
            Boolean haveCreatedSpan =
                    ContextManager.getRuntimeContext().get(HAVE_CREATED_SPAN);
```
And you should check `haveCreatedSpan != null && !haveCreatedSpan`, agree?

wu-sheng: If this context doesn't exist, then it could not change to `boolean`.

ZS-Oliver: I think I forgot to initialize, Otherwise haveCreatedSpan cannot be false.
I think Object cannot auto cast to Boolean.

wu-sheng: I don't say any object. I was saying null casting to Boolean rather than boolean.

ZS-Oliver: 👍 , done

wu-sheng:
You reverted the changelog and caused the conflict again. Please note your local env, and pay attention to git merge.

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/wrapper/ServerInProtocolWrapper.java
```diff
@@ -81,6 +85,24 @@ public TField readFieldBegin() throws TException {
             }
             return readFieldBegin();
         }
+        if (field.type == TType.STOP) {
+            Boolean haveCreatedSpan =
+                    (Boolean) ContextManager.getRuntimeContext().get(HAVE_CREATED_SPAN);
+            if (!haveCreatedSpan) {
```
wu-sheng, line 91: This is still an NPE risk. Take a look at my previous comment. If you are using `Boolean`, you need to `haveCreatedSpan != null &&` first.

If you are not familiar with auto-unboxing in Java, take a look at this, https://stackoverflow.com/questions/55100449/unexpected-nullpointerexception-while-autoboxing-unboxing-java-long-type-as-re

ZS-Oliver: I think I've initialize "haveCreatedSpan" in initial(), so it would't be NPE.
But It is necessary to add a precaution. Good advice.

wu-sheng (approved):
LGTM. Once CI passed, it is ready to merge. Thanks for your patient.

apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/wrapper/ServerInProtocolWrapper.java
wu-sheng: Yes, just be preciseness. :)

PR #5998, opened 2020-12-12
chore: remove unused lagecy codes

kezhenxu94:


oap-server/server-receiver-plugin/skywalking-mesh-receiver-plugin/src/main/java/org/apache/skywalking/aop/server/receiver/mesh/TelemetryDataDispatcher.java
```diff
@@ -103,7 +101,6 @@ public static void process(ServiceMeshMetric.Builder data) {
     static void doDispatch(ServiceMeshMetric.Builder metrics) {
         long minuteTimeBucket = TimeBucket.getMinuteTimeBucket(metrics.getStartTime());
 
-        heartbeat(metrics, minuteTimeBucket);
```
wu-sheng, line 106: Why remove this? If no heartbeat, instances will be gone, and not available on the ui.

kezhenxu94: Check the method, it did nothing but create several POJOs, and I removed it altogether

wu-sheng (approved):

PR #6003, opened 2020-12-14
Checkstyle plugin checks test codes

### LINKED ISSUE #4377: Checkstyle plugin checks test codes too

vcjmhg:
### Checkstyle plugin can check test codes alse.
- [X] If this is non-trivial feature, paste the links/URLs to the design doc. NO NEED
- [X] Update the documentation to include this new feature. NO NEED
- [X] If it's UI related, attach the screenshots below. NO NEED

- [X] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #4377.
- [X] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md). NOT NEED


vcjmhg:
Need the codes in `test/plugin/scenarios` be checked by the checkestyle plugin ?

kezhenxu94:
> Need the codes in `test/plugin/scenarios` be checked by the checkestyle plugin ?

I hope so

vcjmhg:
> I hope so

@kezhenxu94 ,because `scenarios` is not a maven project, there are two ways for checking the codes with the Chekcstyle plugin:
1. add `<sourceDirectory>scenario-name<sourceDirectory>` into `test/plugin/pom.xml`. Just like this:
```xml
 <sourceDirectories>
                        <sourceDirectory>${project.build.sourceDirectory}</sourceDirectory>
                        <sourceDirectory>${project.build.testSourceDirectory}</sourceDirectory>
                        <sourceDirectory>scenarios/activemq-scenario</sourceDirectory>
</sourceDirectories>
```
Meanwhile, in order to make new test-case projects will be checked with the Checkestyle plugin, I will modify the `generator.sh`. And ensure that when creating a new project through the `generator.sh`, a tag like  `<sourceDirectory>new-scenario-name<sourceDirectory>` will be added in `test/plugin/pom.xml`.
2. make `scenarios` to be a maven project and make all projects in its folder become its subprojects
like method one, I also can modify the  `generator.sh`.
Which method would be better?

kezhenxu94:
> because scenarios is not a maven project

I don't quite remember the context why we didn't make it a maven project before, @wu-sheng @dmsolr do you remember the reason? If there is no particular reason, I personally prefer method 2, to make it a maven project.

- import them all at once: I occasionally wanted to check some of the scenarios codes but need to import them one by one manually
- dependency management: there are tons of dependencies that are shared in these scenarios, like Spring, etc, with a maven project, we can share the same dependencies in the parent `pom.xml`

WDYT?

wu-sheng:
The original idea is keeping them separately, as in test cases, we don't want them share anything. Dependency management is the thing we want to avoid, and every scenario will be packaged mutile times for using different library versions.

So, I would say, don't merge them into one to avoid complexity, but add the check plugin for evryone of them. I remember @dmsolr provides a scenario project structure generator? If we change there too, the new scenarios should be easier to follow this new requirement.

vcjmhg:
Okay, I will adopt option one.

kezhenxu94 (commented):
@vcjmhg is this PR ready for review? I see there are some commented lines

vcjmhg:
> @vcjmhg is this PR ready for review? I see there are some commented lines

Not yet, some code formatting work has not been completed. I will finish as soon as possible

test/plugin/pom.xml
```diff
@@ -99,6 +103,151 @@
...
+                <version>${maven-checkstyle-plugin.version}</version>
+                <configuration>
+                    <configLocation>${maven.multiModuleProjectDirectory}/apm-checkstyle/checkStyle.xml</configLocation>
+                    <headerLocation>${maven.multiModuleProjectDirectory}/apm-checkstyle/CHECKSTYLE_HEAD</headerLocation>
+                    <encoding>UTF-8</encoding>
+                    <consoleOutput>true</consoleOutput>
+                    <includeTestSourceDirectory>true</includeTestSourceDirectory>
+                    <failOnViolation>${checkstyle.fails.on.error}</failOnViolation>
+                    <sourceDirectories>
+                        <sourceDirectory>${project.build.sourceDirectory}</sourceDirectory>
```
wu-sheng, line 117: Due to https://github.com/apache/skywalking/blob/753c73ab8ba68717d67dfa5306a7d224a6f26805/test/plugin/generator.sh#L79-L84, could we give rid of this?

vcjmhg: Because `scenarios` is not a maven project, if only configure `<sourceDirectory>${project.build.sourceDirectory}</sourceDirectory>`, the maven project in `scenarios` can't  be checked.

wu-sheng: I know, I mean you are using this script to generate these configurations, right? But the hard-code configurations are still there, https://github.com/apache/skywalking/blob/753c73ab8b/test/plugin/pom.xml#L116-L222

Do I misunderstand something?

kezhenxu94: The `generator.sh` is to generate **NEW** scenario project, and it will add a new line of that project to https://github.com/apache/skywalking/blob/753c73ab8b/test/plugin/pom.xml#L116-L222

wu-sheng: Why limit it to new only? Why don't we do that before running the tests and check the code style first?

vcjmhg: It will be ok to write this script into `run.sh`. If we do so, it will ensure that the code style is checked before running.

wu-sheng: Yes, then this part could be removed and don't require the new contributor to add.

vcjmhg: There are some problems that should be considered, if we write the scripts into `run.sh`:
1. We execute scenarios in a different order and the content of pom.xml may be different, which breaks idempotence and may cause conflicts when merging.
2. Some scenarios may need to be handled separately to check the code style. For example `jdk14-with-gson-scenario` cannot be checked by adding its name in `plugin/pom.xml` because of different jdk version.

wu-sheng:
> We execute scenarios in a different order and the content of pom.xml may be different, which breaks idempotence and may cause conflicts when merging.

The changed pom.xml wouldn't be submitted or merged. It is updated in the CI process only.

> Some scenarios may need to be handled separately to check the code style. For example jdk14-with-gson-scenario cannot be checked by adding its name in plugin/pom.xml because of different jdk version.

Could you be more clear about what does this mean? Why does code style relate to JDK version?

test/plugin/pom.xml
vcjmhg: > The changed pom.xml wouldn't be submitted or merged. It is updated in the CI process only.

Add the scenario name into `pom.xml` before run, and delete the added content after execution. Is it like this? 



> Could you be more clear about what does this mean? Why does code style relate to JDK version?

When I add ` <sourceDirectory>scenarios/jdk14-with-gson-scenario</sourceDirectory>` into `plugin/pom.xml`, We can see the following verification results：
`Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.0:check (validate) on project apache-skywalking-test-plugin: Failed during checkstyle configuration
`
So, I guess the reason may be caused by different jdk versions and I handle it separately. If my understanding is wrong or you have a better solution please let me know

wu-sheng: With your description, how about adding the check style plugin into every scenario's pom.xml, rather than governing the parent pom.xml?

wu-sheng: The generator.sh you changed clearly has the chance to add codestyle check when setting a new plugin test scenario, right?

vcjmhg: Yes,  it will add `<sourceDirectory>scenario_name<\/sourceDirectory>` into pom.xml . It works well when I test on my pc.

wu-sheng: I don't mean this, `Yes, it will add <sourceDirectory>scenario_name<\/sourceDirectory> into pom.xml`.

I prefer you consider to add the plugin configuration into file like this one, https://github.com/apache/skywalking/blob/9c48d0686a/test/plugin/scenarios/activemq-scenario/pom.xml

vcjmhg: ok，I got it. I will have a try!

test/plugin/generator.sh
```diff
@@ -74,4 +74,11 @@ ${MVNW} archetype:generate \
...
-    -Dpackage=${package}
\ No newline at end of file
+    -Dpackage=${package}
+
+# add scenario_name into plugin/pom.xml
+isWrite=$(cat ./pom.xml | grep $scenario_name)
+if [[ -z  $isWrite ]]
+then
+  sed -i '/<\/sourceDirectories>/i <sourceDirectory>'"$scenario_name"'<\/sourceDirectory>' ./pom.xml
+fi
```
wu-sheng, line 84: I think you forget to remove this.

wu-sheng:
Seems the checks passed? Please check the logs to verify the style check really invoked.

vcjmhg:
> Seems the checks passed? Please check the logs to verify the style check really invoked.

When we test the plugin with `bash run.sh -f`, the code can be checkstyle-plugin. Just like this:
image
In the CI, it seems that run the scenario with `bash run.sh`, there may be some hidden problems and I will fix it later.

vcjmhg:
> Seems the checks passed? Please check the logs to verify the style check really invoked.

When we test the plugin with `bash run.sh -f`, the code can be checkstyle-plugin. Just like this:
image
In the CI, it seems that run the scenario with `bash run.sh`, there may be some hidden problems and I will fix it later.

vcjmhg:
@wu-sheng @kezhenxu94 Sorry, some time ago, the progress was a bit slow because I was preparing to the Dissertation Proposal. 
Let me briefly talk about a basic idea to solve this problem at present.
**Currently**, it still add `<sourceDirectory>scenarios/$scenario_name<\/sourceDirectory>` into plugin/pom.xml that make the code can be checked by the `checkstyle-plugin`.
**The problems we encountered before**:
>jdk14-with-gson-scenario cannot be checked by adding its name in plugin/pom.xml because of different JDK versions.

I fix the problem by [Upgrading Checkstyle at Runtime](http://maven.apache.org/plugins/maven-checkstyle-plugin/examples/upgrading-checkstyle.html). And Because `maven-checkstyle-plugin 3.1.10` not support `checkstyle 8.38`, I update it to  `maven-checkstyle-plugin 3.1.11`.
**The only problem at the moment is how to format `retransform-class-tomcat-scenario`:**
I have tried several times including modifying the package name from `test.org.apache.skywalking.apm.testcase.retransform_class` to `test.apache.skywalking.apm.testcase.retransform` and changing the content in `spring-mvc-servlet.xml.xml`. But the ci will fail because the `serviceName` and ` spanName` has changed.
image
**So I want to know when the serviceName  was set to `retransform-class-tomcat-scenario` and the `spanName` was changed.** I can't find the related code in `retransform-class-tomcat-scenario`  and there seem to be no relevant settings in the configuration file and environment variables.

wu-sheng:
From my guessing, do you add some root path to the scenario? Prefix seems related to that.

vcjmhg:
> From my guessing, do you add some root path to the scenario? Prefix seems related to that.

:smile_cat: :smile_cat:  , it can be passed in ci. It seems that there is a problem with my local environment before.

.github/workflows/plugins-jdk14-test.0.yaml
```diff
@@ -53,6 +53,8 @@ jobs:
       - uses: actions/setup-java@v1
         with:
           java-version: 8
+      - name: Add checkstyle-plugin
+        run: sed -i '/<\/sourceDirectories>/i <sourceDirectory>scenarios\/jdk14-with-gson-scenario<\/sourceDirectory>' test/plugin/pom.xml
```
wu-sheng, line 57: @kezhenxu94 @dmsolr Are you ok about this?

kezhenxu94: > @kezhenxu94 @dmsolr Are you ok about this?

Works for me

wu-sheng: @vcjmhg Could you change this yaml as a matrix based? In case we forget we hard coded `jdk14-with-gson-scenario` here.

vcjmhg: ok

test/plugin/run.sh
```diff
@@ -140,11 +140,19 @@ if [[ ! -d ${agent_home} ]]; then
     echo "[WARN] SkyWalking Agent not exists"
     ${mvnw} --batch-mode -f ${home}/../../pom.xml -Pagent -DskipTests clean package
 fi
+# if it fails last time, relevant information will be deleted
+sed -i '/<sourceDirectory>scenarios\/'"$scenario_name"'<\/sourceDirectory>/d' ./pom.xml
+# add scenario_name into plugin/pom.xml
+echo check code with the checkstyle-plugin
+sed -i '/<\/sourceDirectories>/i <sourceDirectory>scenarios\/'"$scenario_name"'<\/sourceDirectory>' ./pom.xml
+
```
wu-sheng, line 148: You inject the style check in the CI control file, what is the purpose of this?

vcjmhg: It ensures that the code will be checked when we run the scenario locally.

wu-sheng: I see. But I feel a little concern about the dirty config(added but not removed somehow) will be pushed.

wu-sheng (approved):
LGTM

wu-sheng:
Thanks for finishing this task. It has waited for a very long time. :)

test/plugin/run.sh
nisiyong: I check out the newest code, 8.5.0-SNAPSHOT.
Run this script failed in my MacBook. How to fix it?
ref: [local-test-and-pull-request-to-the-upstream](https://github.com/apache/skywalking/blob/master/docs/en/guides/Plugin-test.md#local-test-and-pull-request-to-the-upstream)

```bash
➜  skywalking git:(master) bash ./test/plugin/run.sh hbase-scenario
+++ dirname ./test/plugin/run.sh
++ cd ./test/plugin
++ pwd
+ home=/Users/nisiyong/CodeProjects/skywalking/test/plugin
+ scenario_name=
+ force_build=off
+ cleanup=off
+ debug_mode=
+ mvnw=/Users/nisiyong/CodeProjects/skywalking/test/plugin/../../mvnw
+ agent_home=/Users/nisiyong/CodeProjects/skywalking/test/plugin/../../skywalking-agent
+ jacoco_home=/Users/nisiyong/CodeProjects/skywalking/test/plugin/../jacoco
+ scenarios_home=/Users/nisiyong/CodeProjects/skywalking/test/plugin/scenarios
+ num_of_testcases=
+ image_version=jdk8-1.0.0
++ date +%s
+ start_stamp=1616572417
+ parse_commandline hbase-scenario
+ test 1 -gt 0
+ _key=hbase-scenario
+ case "$_key" in
+ scenario_name=hbase-scenario
+ shift
+ test 0 -gt 0
+ [[ off == \o\n ]]
+ test -z hbase-scenario
+ [[ ! -d /Users/nisiyong/CodeProjects/skywalking/test/plugin/../../skywalking-agent ]]
+ sed -i '/<sourceDirectory>scenarios\/hbase-scenario<\/sourceDirectory>/d' ./pom.xml
sed: 1: "./pom.xml": invalid command code .
```

PR #6006, opened 2020-12-14
[Plugin] added component ID for Python aiohttp

tom-pytel:
Added component ID for Python plugin for module `aiohttp`.

### Add an agent plugin to support Python aiohttp module

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,9 @@ Release Notes.
 * Support collecting logs of log4j, log4j2, and logback in the tracing context with a new `logger-plugin`.
 * Fix the unexpected RunningContext recreation in the Tomcat plugin.
 
+#### Python Agent
+* Added plugin for aiohttp module requester and server.
+
 #### OAP-Backend
 * Make meter receiver support MAL.
 * Support influxDB connection response format option. Fix some error when use JSON as influxDB response format.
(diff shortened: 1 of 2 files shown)


kezhenxu94 (dismissed):
Thanks

wu-sheng (commented):
I think changlog should be updated as new component id added.

tom-pytel:
> I think changlog should be updated as new component id added.

Done

CHANGES.md
```diff
@@ -14,6 +14,9 @@ Release Notes.
 * Support collecting logs of log4j, log4j2, and logback in the tracing context with a new `logger-plugin`.
 * Fix the unexpected RunningContext recreation in the Tomcat plugin.
 
+#### Python Agent
+* Added plugin for aiohttp module requester and server.
```
wu-sheng, line 18: This changlog belongs to the OAP-backend. 
And added->add.

tom-pytel: I am confused, so what do you want me to add and where?

kezhenxu94: > I am confused, so what do you want me to add and where?

Done for you

wu-sheng: Sorry, review from cell phone, was trying to type less words.
1. Move the change into OAP-backend part, as the file you changed belongs there. 
2. As this
```suggestion
* Add component id for Python aiohttp plugin.
```

The changelog file in the main repo is tracking the its own change. New python plugin belongs to skywalking-python repo's changelog.

wu-sheng (approved):

PR #6007, opened 2020-12-15
Add null-conditional operators on these 3 files

raybi-asus:
###  
Fix the potential NPE. Due to some special database drivers(eg. ojdbc), different classes have same method name. For example , all of T4CPreparedStatement, OraclePreparedStatementWrapper and OraclePreparedStatement have setString method. If we create a InterceptPoint for setString, all of these 3 classes can be intercepted, but only OraclePreparedStatementWrapper(Outside invocation) have connection info and StatementEnhanceInfo. Hence, sometimes when ojdbc invoke executeQuery method with setMethod for SQL parameters, setMethod of OraclePreparedStatementWrapper runs well , but both  T4CPreparedStatement and OraclePreparedStatement  would get a NullPoniterException. We can find it in agent log. To aviod this issue in a easy way , we can try to add a null-conditional operator for statementEnhanceInfos, to prevent the null object invoke setParameter(index, parameter); 


wu-sheng (commented):
Why the statementEnhanceInfos could be null? I am feeling this could be a bug.

raybi-asus:
> Why the statementEnhanceInfos could be null? I am feeling this could be a bug.

Hello , the case is for PrepareStatement operation and Trace_SQL_Parameter=true. 
Please refer to the attachment 1
Usually we use 'prepareStament' for methods matcher :
agent_connectionInstrumentation

In ojdbc driver PhysicalConnection has prepareStatement method with 2 arguments like below , it will return OraclePreparedStatementWrapper object: 
OraclePreparedStatementWrapper

Then , when we invoke afterMethod , the code would set connection and StatementEnhanceInfos to the object which was returned from last step
CreateInterceptor

Hence , if we try to intercept setString method , the operation for OraclePreparedStatementWrapper runs well , but for internal invocations(Who also have setString method) such as 'T4CPreparedStatement' and 'OraclePreparedStatement', both of them don't have StatementEnhanceInfos , the NPE would be coming up.

kezhenxu94:
> Hence , if we try to intercept setString method , the operation for OraclePreparedStatementWrapper runs well , but for internal invocations(Who also have setString method) such as 'T4CPreparedStatement' and 'OraclePreparedStatement', both of them don't have StatementEnhanceInfos , the NPE would be coming up.

I understand what you said, just curious about how the 'T4CPreparedStatement' and 'OraclePreparedStatement' are created and why they are not enhanced.

Also, please update the changelog.md

wu-sheng:
Are you saying there are cases we don't trace?

raybi-asus:
I uploaded a screenshot of error log. 
NPE of Agent

Actually the fix is not related to the trace. I think they are all traced, but 'T4CPreparedStatement' and 'OraclePreparedStatement' should be filtered out, because they weren't enhanced by 'correct' purpose.

From ConnectionInstrumentation class in Oracle Plugin , I saw it does enhance PhysicalConnection,T4CConnection and intercept prepareStatement method. 
As my preceding post , OraclePreparedStatementWrapper is the return value of 'prepareStatement' method in PhysicalConnection which was already enhanced in ConnectionInstrumentation.  Below is the method name in PhysicalConnection:
public synchronized PreparedStatement prepareStatement(String paramString, Properties paramProperties) throws SQLException

So , i guess OraclePreparedStatementWrapper was set StatementEnhanceInfos when the code intercept 'prepareStatement' method of PhysicalConnection. 

About the question why 'T4CPreparedStatement' and 'OraclePreparedStatement' are not enhanced.  
'T4CPreparedStatement' and 'OraclePreparedStatement' were actually enhanced. That's why they still can be intercepted by 'JDBCPreparedStatementSetterInterceptor.beforeMethod'. But they were intercepted by setString interceptor of ' T4CPreparedStatement, OraclePreparedStatementWrapper and OraclePreparedStatement' in 'OraclePrepareStatementInstrumentation' rather than 'PhysicalConnection.prepareStatement' , so they didn't inject StatementEnhanceInfos . That's the cause of NPE.

That's my opinion.
Please let me know if you have any concerns.

wu-sheng:
The current change is ignoring to collect the parameters.Is that right? If the info is missing, I prefer we could propagate.

raybi-asus:
> The current change is ignoring to collect the parameters.Is that right? If the info is missing, I prefer we could propagate.

The current change is only to aviod the NPE.  I won't ignore to collect the parameters.
The trace_sql_parameter is still working for the interceptor who has StatementEnhanceInfos and ConnectionInfo.

From Oracle plugin case, OraclePreparedStatementWrapper was enhanced and intercepted , the parameters can be traced successfully.  The null-conditional operator will fliter out 'T4CPreparedStatement' and 'OraclePreparedStatement'. Because they may cause the NPE. Below is the test trace list in my local env
trace
Parameter

wu-sheng (approved):

ascrutae (approved):

PR #6009, opened 2020-12-15
fix: Non-atomic operation on volatile field

nisiyong:
### Fix Non-atomic operation on volatile field
- [x] Fix #5606, this PR didn't fix the bug.

Recently I dive into SkyWalking sources code, and I find some bugs in the `DataCarrier` module.
It runs with something wrong although they won't affect the main feature. 

IntelliJ IDEA shows the warning as follow:
> Inspection info: Reports any non-atomic operations on volatile fields. Non-atomic operations on volatile fields are operations where the field is read and the value is used to update the field. It is possible for the value of the field to change between the read and the write, possibly invalidating the operation. The non-atomic operation can be avoided by surrounding it with a synchronized block or by making use of one of the classes from the java.util.concurrent.atomic package.

### DIFF
--- a/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/MultipleChannelsConsumer.java
+++ b/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/MultipleChannelsConsumer.java
@@ -20,6 +20,8 @@ package org.apache.skywalking.apm.commons.datacarrier.consumer;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicLong;
+
 import org.apache.skywalking.apm.commons.datacarrier.buffer.Channels;
 import org.apache.skywalking.apm.commons.datacarrier.buffer.QueueBuffer;
 
@@ -30,13 +32,12 @@ import org.apache.skywalking.apm.commons.datacarrier.buffer.QueueBuffer;
 public class MultipleChannelsConsumer extends Thread {
     private volatile boolean running;
     private volatile ArrayList<Group> consumeTargets;
-    @SuppressWarnings("NonAtomicVolatileUpdate")
-    private volatile long size;
+    private final AtomicLong size = new AtomicLong();
     private final long consumeCycle;
 
     public MultipleChannelsConsumer(String threadName, long consumeCycle) {
         super(threadName);
-        this.consumeTargets = new ArrayList<Group>();
+        this.consumeTargets = new ArrayList<>();
         this.consumeCycle = consumeCycle;
     }
 
@@ -94,17 +95,14 @@ public class MultipleChannelsConsumer extends Thread {
     public void addNewTarget(Channels channels, IConsumer consumer) {
         Group group = new Group(channels, consumer);
         // Recreate the new list to avoid change list while the list is used in consuming.
-        ArrayList<Group> newList = new ArrayList<Group>();
-        for (Group target : consumeTargets) {
-            newList.add(target);
-        }
+        ArrayList<Group> newList = new ArrayList<>(consumeTargets);
         newList.add(group);
         consumeTargets = newList;
-        size += channels.size();
+        size.addAndGet(channels.size());
     }
 
     public long size() {
-        return size;
+        return size.get();
     }
 
     void shutdown() {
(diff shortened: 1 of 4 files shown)


apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/MultipleChannelsConsumer.java
```diff
@@ -30,13 +32,12 @@
 public class MultipleChannelsConsumer extends Thread {
     private volatile boolean running;
     private volatile ArrayList<Group> consumeTargets;
-    @SuppressWarnings("NonAtomicVolatileUpdate")
-    private volatile long size;
+    private final AtomicLong size = new AtomicLong();
```
wu-sheng, line 35: Why need atomic long? We don't do concurrency consuming in any case

nisiyong: Yes, there has another way to fix this warning. Just add `synchronized` on the `MultipleChannelsConsumer#addNewTarget`, but it has already added on `BulkConsumePool#add` which invoke `MultipleChannelsConsumer#addNewTarget`.

IDE shows warning because the volatile filed has non-atomic operations on this Class, the non-atomic operations should be synchronized. Or we just use AtomicLong.

wu-sheng: I don't find the real case, the reason for warning is this could happen, but in our case, there isn't.

nisiyong: It doesn't mean there has a wrong logic in the release code, maybe it runs well. But I just want to indicate that was a bad practice to use `volatile` in this way. 

If there are multi-threads invoke the `MultipleChannelsConsumer#addNewTarget` method and the field `size` still uses `volatile`,  the `size` maybe a wrong value. Because 
> The Java `volatile` keyword guarantees visibility only! 

If you don't want to use AtomicLong, I am willing to just add `synchronized` on the `MultipleChannelsConsumer#addNewTarget` method.

wu-sheng: We are trying to avoid any possible lock, it is also slower, more or less. That is why it may be not a good idea.
SkyWalking codes are for product environment and high performance, we have no plan to use the codes to guide people how to write codes.

So, the key is, if the current version of codes has a bug, we are open to the solution.

nisiyong: Okay, I can revert `MultipleChannelsConsumer` first. 
But the `SimpleRollingPartitioner` has a bug, and I have added a test case. Please take a look.

apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/partition/SimpleRollingPartitioner.java
```diff
@@ -18,16 +18,18 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
 /**
  * use normal int to rolling.
  */
 public class SimpleRollingPartitioner<T> implements IDataPartitioner<T> {
-    @SuppressWarnings("NonAtomicVolatileUpdate")
-    private volatile int i = 0;
+
+    private final AtomicInteger i = new AtomicInteger();
```
wu-sheng, line 28: Let's talk about this. I know your point of bug, it would not increase every time. But it has higher performance with no lock, in the partition, there is a lock to avoid the real issue. So this is actually not a bug in the real world.
This is another typical thing, codes are not by the book.

nisiyong: > in the partition, there is a lock to avoid the real issue.

Could you tell me where the lock is? I want to take a look.

wu-sheng: Channel index. There is a atomic integer as index pointer.

PR #6037, opened 2020-12-21
Update byte-buddy to 1.10.19

lujiajing1126:

For the custom plugins using annotation with primitive array members, the current byte-buddy version (`1.10.16`) cannot find proper target to enhance.

See the issue in the upstream repo https://github.com/raphw/byte-buddy/issues/971
and a full example to reproduce the issue
https://github.com/lujiajing1126/bytebuddy-issue971


wu-sheng:
You need to update the changelog, and version in the license file, https://github.com/apache/skywalking/blob/40a98ee14b397850376f5362a339aa66275a4fd7/dist-material/release-docs/LICENSE#L228

wu-sheng:
Also, the asm version should be updated.

lujiajing1126:
I have updated the change log and LICENSE files.

wu-sheng (approved):

PR #6038, opened 2020-12-21
Fix some mistakes in the comments

wankai123:


kezhenxu94 (approved):

apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/v4/define/AbstractControllerInstrumentation.java
```diff
@@ -26,6 +26,7 @@
 import org.apache.skywalking.apm.agent.core.plugin.match.ClassAnnotationMatch;
 import org.apache.skywalking.apm.agent.core.plugin.match.ClassMatch;
 import org.apache.skywalking.apm.plugin.spring.mvc.commons.Constants;
+import org.apache.skywalking.apm.plugin.spring.mvc.commons.interceptor.RequestMappingMethodInterceptor;
```
wu-sheng, line 29: Please run the maven check locally, comments related thing should not be in import.
Same as below.
That is why CI fails.

wu-sheng (approved):
Seems correct now, waiting for the CI.

Issue #6056, opened 2020-12-22
use influxdb identifier cause error


zhe1926:
### improvement
searchService method use influxdb identifier in sql when we use storage-influxdb-plugin.Put influxdb identifier in the middle of double quotation marks can solve this question.

![image](https://user-images.githubusercontent.com/13587931/102866269-f1950300-4471-11eb-85b5-b31d67364824.png)

wu-sheng:
Could you submit a pull request directly?

PR #6057, opened 2020-12-22
fix influx query error #6056

zhe1926:
image

SearchService method use influxdb identifier in sql when we use storage-influxdb-plugin.Put influxdb identifier in the middle of double quotation marks can solve this question.


wu-sheng:
Could you update the `CHANGES.md` in the root folder.

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/MetadataQuery.java
```diff
@@ -125,7 +125,7 @@ public Service searchService(String serviceCode) throws IOException {
         final WhereQueryImpl<SelectQueryImpl> whereQuery = select(ID_COLUMN, NAME, ServiceTraffic.GROUP)
             .from(client.getDatabase(), ServiceTraffic.INDEX_NAME)
             .where(eq(TagName.NODE_TYPE, String.valueOf(NodeType.Normal.value())));
-        whereQuery.and(eq(ServiceTraffic.NAME, serviceCode));
+        whereQuery.and(eq("\"" + ServiceTraffic.NAME + "\"", serviceCode));
```
dmsolr, line 128: We have a constant with quotes, `InfluxConstants.NAME`.

zhe1926: thanks,i submit again.

wu-sheng (approved):
LGTM

dmsolr (approved):

PR #6083, opened 2020-12-26
I found that when using the ehcache1.* version, the agent will throw NPE in the plugin, so I fixed it

YczYanchengzhe:
### Fix < using the ehcache1.* version, the agent will throw NPE>
- [ ] try catch it and give it default value


wu-sheng:
The codes you are changing is ehcache 2.x plugin, we never said this supports 1.x. If you want to support more versions, adding NULL check is not a good idea. You should find why the required info is not missing and add it back.

By the way, we provided the code style file in the root folder, please don't format the codes randomly.

PR #6087, opened 2020-12-27
Update cancel.yaml to use ASF GHA

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>. NO NEED

### DIFF
--- a/.github/workflows/cancel.yaml
+++ b/.github/workflows/cancel.yaml
@@ -29,140 +29,140 @@ jobs:
     runs-on: ubuntu-18.04
     timeout-minutes: 10
     steps:
-      - uses: potiuk/cancel-workflow-runs@v4_7
+      - uses: apache/airflow-cancel-workflow-runs@f4745c8ce1955c28676b5afe129a88a61aa743b9
         name: Cancel Outdated Builds ci-it.yaml
         with:
           cancelMode: allDuplicates
           token: ${{ secrets.GITHUB_TOKEN }}
           sourceRunId: ${{ github.event.workflow_run.id }}
           workflowFileName: ci-it.yaml
-      - uses: potiuk/cancel-workflow-runs@v4_7
+      - uses: apache/airflow-cancel-workflow-runs@f4745c8ce1955c28676b5afe129a88a61aa743b9
         name: Cancel Outdated Builds docker-ci.yaml
         with:
           cancelMode: allDuplicates
           token: ${{ secrets.GITHUB_TOKEN }}
           sourceRunId: ${{ github.event.workflow_run.id }}
           workflowFileName: docker-ci.yaml
-      - uses: potiuk/cancel-workflow-runs@v4_7
+      - uses: apache/airflow-cancel-workflow-runs@f4745c8ce1955c28676b5afe129a88a61aa743b9
         name: Cancel Outdated Builds e2e.cluster.yaml
         with:
           cancelMode: allDuplicates
           token: ${{ secrets.GITHUB_TOKEN }}
           sourceRunId: ${{ github.event.workflow_run.id }}
           workflowFileName: e2e.cluster.yaml
-      - uses: potiuk/cancel-workflow-runs@v4_7
+      - uses: apache/airflow-cancel-workflow-runs@f4745c8ce1955c28676b5afe129a88a61aa743b9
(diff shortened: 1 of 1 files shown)


kezhenxu94:
Thanks @potiuk for quick response, btw, do you need an ASF release for that GHA to be used?

wu-sheng (approved):
Cool!!! Thanks @potiuk for updating this so quickly.

potiuk:
No need for a release @kezhenxu94. This is not software officially "released" by ASF the idea is that its origin must be controlled by the ASF committers.

wu-sheng:
> do you need an ASF release for that GHA to be used?

From my understanding, if we use internally(inside ASF), this is not a strict required. As this is not included in a source release :) But if they are going to post on the airflow website as a thing called official release(or similar), then it is .

potiuk (approved):

potiuk:
One comment here. I thought about the new policy and consequences, and I think it makes much more sense for each projects to keep their own copies of the actions. I am still waiting for the infra/Greg to comment on it but I raised my concerns here about using such actions cross-project:

 https://lists.apache.org/thread.html/rfc3269829edc672f928d8304d7b0aa38d1d2a5f5cb177af95e98a1e5%40%3Cusers.infra.apache.org%3E

Happy if you also join the discussion, I'd love to hear what others think about it.

kezhenxu94:
> One comment here. I thought about the new policy and consequences, and I think it makes much more sense for each projects to keep their own copies of the actions. I am still waiting for the infra/Greg to comment on it but I raised my concerns here about using such actions cross-project:
> 
> https://lists.apache.org/thread.html/rfc3269829edc672f928d8304d7b0aa38d1d2a5f5cb177af95e98a1e5%40%3Cusers.infra.apache.org%3E
> 
> Happy if you also join the discussion, I'd love to hear what others think about it.

Thanks. I'm also watching that thread, will reply there

wu-sheng:
@potiuk I think your(airflow) fork is fine, then SkyWalking doesn't need a new one, from my understanding of ASF policy. 
2 key points,
1. INFRA is making sure the action is security, and doesn't include write operation.
2. Airflow and SkyWalking are in the same foundation, we are just 2 groups of people interesting in different tech directions, but basically, we are sharing branding(ASF) and source code IPs.

potiuk:
> 1. INFRA is making sure the action is security, and doesn't include write operation.

This is not happening now. INFRA is not involved in review of the actions. And some of those actions actually perform write operations.

> 2. Airflow and SkyWalking are in the same foundation, we are just 2 groups of people interesting in different tech directions, but basically, we are sharing branding(ASF) and source code IPs.

The problem is that I (more any on the commiters/PMCs of Airflow) we are not reviewing all the commits in those repos. I simply cloned someone else's action repo (in this case it is mine, so I am pretty sure I know what I am doing but there are other actions that aren't mine). So I just cloned the repo, In my case, I am reviewing only the particular commit I am referring to in 'apache-airflow' but someone could add malicious code in any of the previous commits - I have no physical possibility to review all that code, but I want to keep the history so that I can bring new changes, review and switch to new versions.

wu-sheng:
> This is not happening now. INFRA is not involved in review of the actions. And some of those actions actually perform write operations.

Yes, I know they haven't. I mean, all these things are related to this purpose :)

> The problem is that I (more any on the commiters/PMCs of Airflow) we are not reviewing all the commits in those repos. I simply cloned someone else's action repo (in this case it is mine, so I am pretty sure I know what I am doing but there are other actions that aren't mine). So I just cloned the repo, In my case, I am reviewing only the particular commit I am referring to in 'apache-airflow' but someone could add malicious code in any of the previous commits - I have no physical possibility to review all that code, but I want to keep the history so that I can bring new changes, review and switch to new versions.

We(SkyWalking) are only using the `cancel` task for now, so, I could say this is safe, right :). That is my point of now doing a duplicated clone, specifically for this case.

potiuk:
> We(SkyWalking) are only using the cancel task for now, so, I could say this is safe, right :). That is my point of now doing a duplicated clone, specifically for this case.

For now yes I assume you reviewed my code :). I am more into creating a good policy that anyone could follow.

PR #6093, opened 2020-12-28
Added statement access frequency statistics

jaredhu:
### Added statement access frequency statistics
image

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/TopNRecordsQueryService.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/TopNRecordsQueryService.java
@@ -50,4 +50,9 @@ public class TopNRecordsQueryService implements Service {
         return getTopNRecordsQueryDAO().readSampledRecords(
             condition, ValueColumnMetadata.INSTANCE.getValueCName(condition.getName()), duration);
     }
+
+    public List<SelectedRecord> readSampledRecordsMetric(TopNCondition condition, Duration duration) throws IOException {
+        return getTopNRecordsQueryDAO().readSampledRecordsMetric(
+                condition, ValueColumnMetadata.INSTANCE.getValueCName(condition.getName()), duration);
+    }
 }
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/ITopNRecordsQueryDAO.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/ITopNRecordsQueryDAO.java
@@ -20,6 +20,7 @@ package org.apache.skywalking.oap.server.core.storage.query;
 
 import java.io.IOException;
 import java.util.List;
+
 import org.apache.skywalking.oap.server.core.analysis.Stream;
 import org.apache.skywalking.oap.server.core.analysis.worker.TopNStreamProcessor;
 import org.apache.skywalking.oap.server.core.query.input.Duration;
@@ -36,4 +37,6 @@ public interface ITopNRecordsQueryDAO extends Service {
     List<SelectedRecord> readSampledRecords(TopNCondition condition,
                                             final String valueColumnName,
                                             Duration duration) throws IOException;
+
+    List<SelectedRecord> readSampledRecordsMetric(TopNCondition condition, String valueCName, Duration duration) throws IOException;
 }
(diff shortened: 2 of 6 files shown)


wu-sheng:
Could you explain what are doing? You are adding new query GraphQL level service, this is a protocol level change. 
All changes related to this must be discussed and approved by the whole PMC. cc @apache/skywalking-committers

jaredhu:
Sorry, I don’t know this process, I just have this requirement and developed this function

wu-sheng:
What issue do you want to resolve? What is the meaning of the metrics of a sampled data set? Such as `Slow statement access`.

jaredhu:
I just adding new query GraphQL level service, I want to see the access frequency of the statements of various business systems, find out the high-frequency statements of the system, and see if caching or other methods can be used to reduce access frequency, reduce database pressure, and improve system stability Sex, and just wrote elasticsearch, but nothing else 
@wu-sheng @apache/skywalking-committers

wu-sheng:
> I just adding new query GraphQL level service, I want to see the access frequency of the statements of various business systems, find out the high-frequency statements of the system, and see if caching or other methods can be used to reduce access frequency, reduce database pressure, and improve system stability Sex, and just wrote elasticsearch, but nothing else

Could you take a look what I said above? This statistic is not accurate, actually, far from accurate. Have you checked how SkyWalking inserts the data you queried.

jaredhu:
> > 我只是添加新的查询GraphQL级服务，我想查看各种业务系统的语句的访问频率，找出系统的高频语句，并查看是否可以使用缓存或其他方法来降低访问频率，减轻数据库压力，提高系统稳定性性别，只写了elasticsearch，但没有别的
> 
> 你能看一下我上面说的吗？实际上，此统计数据不准确。您是否检查过SkyWalking如何插入查询的数据。

jaredhu:
> > I just adding new query GraphQL level service, I want to see the access frequency of the statements of various business systems, find out the high-frequency statements of the system, and see if caching or other methods can be used to reduce access frequency, reduce database pressure, and improve system stability Sex, and just wrote elasticsearch, but nothing else
> 
> Could you take a look what I said above? This statistic is not accurate, actually, far from accurate. Have you checked how SkyWalking inserts the data you queried.

It is based on the data of Slow statement access. This does not need to be absolutely accurate, just a relative magnitude.

wu-sheng:
If the `magnitude` is correct, they this is meaningful, but sadly, the way you implemented isn't.

jaredhu:
How should it be implemented? I accidentally closed it

wu-sheng:
Could you share the idea first about why do you need this count? Aren't the slowest SQLs the biggest issues you need to resolve?

jaredhu:
Yes, the slowest SQLs is my first priority, but the SQLs with high access frequency is also a problem I want to care about. High-frequency SQL is also one of the reasons that affect database performance, but the [No statement] shown in my screenshot is not What I care about, this will block the writing of this data in actual use

wu-sheng:
> High-frequency SQL is also one of the reasons that affect database performance, but the [No statement] shown in my screenshot is not What I care about, this will block the writing of this data in actual use

Let's separate your issues to be care what need to do.
1. High-frequency SQL may or may not an issue. How to statistic all SQLs? Some SQLs are very long, we saw 7k length. So, if you want to discuss this deeply, you need to know what is the ID of a SQL. Because statistic is targeting for the entity which requires an ID.
2. Ignoring `[No statement]` is quite simple. Isn't it? Change the `MultiScopesAnalysisListener` to exclude this case, and provide a configuration.(Ignore could be default, I think).
3. All things you want to resolve are irrelative with this query. That is why in the open source collaboration, you should discuss first.

jaredhu:
Yes, discussion is really necessary
1. I really want to be simple, and I didn't think about this problem carefully. As for the problem of sql id, I have no idea about this for the time being. Can you provide any suggestions?
2. I agree with this, it can be developed to ignore the configuration items, this is also easy to implement
3. So, close this pr first

wu-sheng:
> I really want to be simple, and I didn't think about this problem carefully. As for the problem of sql id, I have no idea about this for the time being. Can you provide any suggestions?

First, in the reality, the SQL must be limited by length. But to be honest, I am not sure whether this is helpful. What would you do if you know a top frequency SQL list in the service level? 

> I agree with this, it can be developed to ignore the configuration items, this is also easy to implement

This could be sent by another pull request.

> So, close this pr first

Agree.

PR #6098, opened 2020-12-29
Fix CVE of UninstrumentedGateways in Dynamic Configuration activation.

wu-sheng:

CVE fix as title. I copied the solution from ShardingSphere. 

To ShardingSphere PMC members, @tristaZero @terrymanu, I think as we are in the same foundation, no LICENSE update is required. So I copied this class, which is a good solution.
https://github.com/apache/shardingsphere/blob/5cda8ed914530ac5a9300890995971049ef7ff8d/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/yaml/engine/constructor/ClassFilterConstructor.java

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -38,6 +38,7 @@ Release Notes.
 * Add component ID for NodeJS Axios plugin.
 * Fix searchService method error in storage-influxdb-plugin.
 * Add JavaScript component ID.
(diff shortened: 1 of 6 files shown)


kezhenxu94 (approved):

terrymanu:
Get, I will have a look, thank you for reminder.

PR #6103, opened 2020-12-30
new feature add witness method

libinglong:
### <Feature description>
     
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>. No issue. Original discussion happened on the mail list, https://lists.apache.org/thread.html/r4f67a7ed880a493e4e82bd2e9a0c26a6e43c37bbfcdaebe7b1a340dc%40%3Cdev.skywalking.apache.org%3E


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/WitnessMethod.java
```diff
@@ -0,0 +1,49 @@
...
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+import java.util.StringJoiner;
+
+/**
+ * @author binglongli217932
+ * <a href="mailto:libinglong9@gmail.com">libinglong:libinglong9@gmail.com</a>
+ * @since 2020/12/30
+ */
```
kezhenxu94, line 30: Remove this

CHANGES.md
```diff
@@ -22,6 +22,7 @@ Release Notes.
 * Fix DataCarrier's `org.apache.skywalking.apm.commons.datacarrier.buffer.Buffer` implementation isn't activated in `IF_POSSIBLE` mode.
 * Fix ArrayBlockingQueueBuffer's useless `IF_POSSIBLE` mode list
 * Support building gRPC TLS channel but CA file is not required.
+* add witness method for plugin develop.
```
kezhenxu94, line 25: Capitalize this sentence

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/AbstractClassEnhancePluginDefine.java
```diff
@@ -70,6 +75,25 @@
...
+                TypePool.Resolution resolution = WitnessClassFinder.INSTANCE.getResolution(witnessMethod.declaringClassName, classLoader);
+                if (!resolution.isResolved()) {
+                    LOGGER.warn("enhance class {} by plugin {} is not working. Because declaringClass {} of the witness method is not existed.", transformClassName, interceptorDefineClassName, witnessMethod.declaringClassName);
+                    return null;
+                }
+                boolean empty = resolution.resolve()
+                        .getDeclaredMethods()
+                        .filter(witnessMethod.elementMatcher)
+                        .isEmpty();
+                if(empty){
```
kezhenxu94, line 90: code style
```suggestion
                if (empty) {
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/AbstractClassEnhancePluginDefine.java
```diff
@@ -104,6 +128,10 @@
         return new String[] {};
     }
 
+    protected List<WitnessMethod> witnessMethods() {
+        return Collections.emptyList();
```
wu-sheng, line 132:
```suggestion
        return null;
```
Will be better performance.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/AbstractClassEnhancePluginDefine.java
```diff
@@ -70,6 +75,25 @@
...
+        if (!CollectionUtil.isEmpty(witnessMethods)) {
+            for (WitnessMethod witnessMethod : witnessMethods) {
+                TypePool.Resolution resolution = WitnessClassFinder.INSTANCE.getResolution(witnessMethod.declaringClassName, classLoader);
+                if (!resolution.isResolved()) {
+                    LOGGER.warn("enhance class {} by plugin {} is not working. Because declaringClass {} of the witness method is not existed.", transformClassName, interceptorDefineClassName, witnessMethod.declaringClassName);
+                    return null;
+                }
+                boolean empty = resolution.resolve()
+                        .getDeclaredMethods()
+                        .filter(witnessMethod.elementMatcher)
```
wu-sheng, line 88: Please provide the `getElementMatcher()` method, don't access the fields of another class.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/AbstractClassEnhancePluginDefine.java
```diff
@@ -70,6 +75,25 @@
...
+        List<WitnessMethod> witnessMethods = witnessMethods();
+        if (!CollectionUtil.isEmpty(witnessMethods)) {
+            for (WitnessMethod witnessMethod : witnessMethods) {
+                TypePool.Resolution resolution = WitnessClassFinder.INSTANCE.getResolution(witnessMethod.declaringClassName, classLoader);
+                if (!resolution.isResolved()) {
+                    LOGGER.warn("enhance class {} by plugin {} is not working. Because declaringClass {} of the witness method is not existed.", transformClassName, interceptorDefineClassName, witnessMethod.declaringClassName);
+                    return null;
+                }
+                boolean empty = resolution.resolve()
+                        .getDeclaredMethods()
```
wu-sheng, line 87: `getDeclaredMethods` doesn't include the inherited methods, right?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/AbstractClassEnhancePluginDefine.java
```diff
@@ -70,6 +75,25 @@
                 }
             }
         }
+        List<WitnessMethod> witnessMethods = witnessMethods();
```
wu-sheng, line 78: This existing logic should move into the `WitnessClassFinder` from my understanding. And you should add UTs for WitnessClassFinder.

CHANGES.md
```diff
@@ -22,6 +22,7 @@ Release Notes.
 * Fix DataCarrier's `org.apache.skywalking.apm.commons.datacarrier.buffer.Buffer` implementation isn't activated in `IF_POSSIBLE` mode.
 * Fix ArrayBlockingQueueBuffer's useless `IF_POSSIBLE` mode list
 * Support building gRPC TLS channel but CA file is not required.
+* Add witness method for plugin develop.
```
wu-sheng, line 25:
```suggestion
* Add witness method mechanism in the agent plugin core.
```

wu-sheng: The plugin development doc is required to update too.

wu-sheng (commented):
Comments inline.

wu-sheng:
@candyleer If you have time, could you recheck this agent core-level feature?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/AbstractClassEnhancePluginDefine.java
libinglong: Right.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/AbstractClassEnhancePluginDefine.java
libinglong: I think The WitnessClassFinder class name should be changed If the logic is moved.
What's the meaning of UTs here?

wu-sheng: `WitnessClassFinder` -> `WitnessFinder` should be enough.
UT means Unit Test, to verify the method you added.

libinglong:
Do I need to care about code review of muse-dev?

docs/en/guides/Java-Plugin-Development-Guide.md
```diff
@@ -299,6 +299,30 @@ The following sections will tell you how to implement the interceptor.
 tomcat-7.x/8.x=TomcatInstrumentation
 ```
 
+4. If a plugin should be activated or not in different condition.You can Override the witnessClasses or witnessMethods.
```
wu-sheng, line 302:
```suggestion
4. Set up `witnessClasses` and/or `witnessMethods` if the instrumentation should be activated in specific versions.
```

docs/en/guides/Java-Plugin-Development-Guide.md
```diff
@@ -299,6 +299,30 @@ The following sections will tell you how to implement the interceptor.
...
+   @Override
+   protected String[] witnessClasses() {
+     return new String[] {
+       "foo.Bar"
+     };
+   }
+   
+   //the plugin is activated only when the foo.Bar#hello method exists.
+   @Override
+   protected List<WitnessMethod> witnessClasses() {
```
wu-sheng, line 317:
```suggestion
   protected List<WitnessMethod> witnessMethods() {
```
Copy/paste issue?

apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/plugin/witness/WitnessTest.java
```diff
@@ -0,0 +1,27 @@
...
+    private String className = "org.apache.skywalking.apm.agent.core.plugin.witness.WitnessTest";
+
+    @Test
+    public void testWitnessClass(){
+        Assert.assertTrue(WitnessFinder.INSTANCE.exist(className, this.getClass().getClassLoader()));
+    }
+
+    @Test
+    public void testWitnessMethod(){
+        WitnessMethod witnessMethod = new WitnessMethod(className, ElementMatchers.named("testWitnessMethod"));
```
wu-sheng, line 23: Please add more tests also could be as demo about how to set witness methods with parameter type and return value.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/WitnessFinder.java
```diff
@@ -46,9 +57,24 @@ public boolean exist(String witnessClass, ClassLoader classLoader) {
...
-        return witnessClassResolution.isResolved();
+        return typePool.describe(witnessClass);
+    }
+
+    /**
+     * @param classLoader for finding the witness method
+     * @return true, if the given witness method exists, through the given classLoader.
+     */
+    public boolean exist(WitnessMethod witnessMethod, ClassLoader classLoader) {
+        TypePool.Resolution resolution = WitnessFinder.INSTANCE.getResolution(witnessMethod.getDeclaringClassName(), classLoader);
```
wu-sheng, line 68:
```suggestion
        TypePool.Resolution resolution = this.getResolution(witnessMethod.getDeclaringClassName(), classLoader);
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/WitnessMethod.java
```diff
@@ -0,0 +1,61 @@
...
+                .toString();
+    }
+
+    public String getDeclaringClassName() {
+        return declaringClassName;
+    }
+
+    public ElementMatcher<? super MethodDescription.InDefinedShape> getElementMatcher() {
+        return elementMatcher;
+    }
```
wu-sheng, line 59: Use `@Getter`.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/AbstractClassEnhancePluginDefine.java
```diff
@@ -64,12 +67,21 @@
         String[] witnessClasses = witnessClasses();
         if (witnessClasses != null) {
             for (String witnessClass : witnessClasses) {
-                if (!WitnessClassFinder.INSTANCE.exist(witnessClass, classLoader)) {
+                if (!WitnessFinder.INSTANCE.exist(witnessClass, classLoader)) {
```
wu-sheng, line 70: We could move `WitnessFinder.INSTANCE` as a local variable.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/WitnessClassFinder.java
```diff
@@ -36,6 +36,17 @@
      * @return true, if the given witnessClass exists, through the given classLoader.
      */
     public boolean exist(String witnessClass, ClassLoader classLoader) {
+        return getResolution(witnessClass, classLoader)
```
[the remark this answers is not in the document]
wu-sheng: @muse-dev The `hashmap` modification is already in the double-check and `synchronized`. I don't think this is an issue.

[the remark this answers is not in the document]
wu-sheng: @TomMD Could you double-check about this?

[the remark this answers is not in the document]
wu-sheng: > Do I need to care about code review of muse-dev?

@libinglong I am not sure whether this is an issue. Pinged the muse-dev team. But at least, this is an issue brought by this PR.

wu-sheng:
Your codes fail in CI, maybe code style issue? Please fix it with other comment suggestions.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/WitnessClassFinder.java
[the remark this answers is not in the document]
TomMD: @wu-sheng I haven't looked at the actual trace produced by the tool but here are my thoughts - please let me know if they seem off and I can ask the tool for its explanation.

The message here is due to the same core issue as you see in the `getResolution` comment below.  When calling `getResolution` then without synchronization we execute:

```
 TypePool typePool = poolMap.get(mappingKey);
```

And that can race with a write to `poolMap` occurring in another thread.  There might not be a `mappingKey` inside of `poolMap` because `poolMap` could be modified prior to calling `get`.  If you'd like to be sure the element exists then you'd need the synchronize block to encompass both the `get` and the `containsKey` call.

We're planning on making traces accessible from the comments with a commands to the bot - this would illuminate the supposed bugs and help developers decide if the trace is feasible and worth changing.  We welcome any thoughts you have about bug explanation and elaboration.

wu-sheng:
Could you fix the CI tasks? Then I could review your PR again.

apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/plugin/witness/WitnessTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private String className = "org.apache.skywalking.apm.agent.core.plugin.witness.WitnessTest";
+
+    @Test
+    public void testWitnessClass() {
+        Assert.assertTrue(WitnessFinder.exist(className, this.getClass().getClassLoader()));
+    }
+
+    @Test
+    public void testWitnessMethod() {
```
wu-sheng, line 27: Please keep the previous version too. (name match only). It is better to have 2 test cases.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/WitnessClassFinder.java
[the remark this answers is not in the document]
wu-sheng: @TomMD Typical race condition of HashMap should like this, https://stackoverflow.com/questions/7830791/java-hashmap-race-condition.

I can't find the `get` conflicts with `put`. What do I miss?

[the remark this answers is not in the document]
libinglong: Maybe It should not be assumed that the put method for a key has no effect on the get method for another key In concurrent environment as long as the data struct(hashmap) is thread unsafe.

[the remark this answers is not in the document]
TomMD: @wu-sheng I don't believe you miss anything, it's merely a difference of what one considers a race.  If you `get` from a variable outside of synchronized, and that variable is sometimes used in a `synchronized` block then you receive this read/write race warning.

Your read might race with a write because the lock isn't held.  You are probably unconcerned and can quickly dismiss the warning in this case.  The link you posted suggests you are concern is about races regarding the handling of the hashmap itself (i.e. the collision and resize issues in the answer) and it suffices to avoid write/write races.

I believe in practice read/write races are concerning almost exclusively in multi-variable situations where there are invariants between the variables that are broken within the writers so the readers must ensure they see a consistent state, usually with a lock.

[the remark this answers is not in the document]
wu-sheng: @TomMD My understanding is, Java's HashMap is only facing write/write race, but here, it shows me the `HashMap#get` will need a lock. Is this correct?

apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/plugin/witness/WitnessTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.skywalking.apm.agent.core.plugin.WitnessMethod;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * witness test
+ */
```
wu-sheng, line 34: This comment seems strange. Please fix.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/WitnessMethod.java
```diff
@@ -0,0 +1,48 @@
...
+import net.bytebuddy.matcher.ElementMatcher;
+
+/**
+ * Witness Method for plugin activation
+ */
+@ToString
+public class WitnessMethod {
+
+    /**
+     * java.lang.reflect.Method#getDeclaringClass()
```
wu-sheng, line 33: What does this comment means?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/WitnessFinder.java
```diff
@@ -0,0 +1,82 @@
...
+
+package org.apache.skywalking.apm.agent.core.plugin;
+
+import java.util.HashMap;
+import java.util.Map;
+import net.bytebuddy.pool.TypePool;
+
+/**
+ * The <code>WitnessFinder</code> represents a pool of {@link TypePool}s, each {@link TypePool} matches a {@link
+ * ClassLoader}, which helps to find the class define existed or not.
```
wu-sheng, line 27:
```suggestion
 * ClassLoader}, which helps to find the class declaration existed or not.
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/WitnessMethod.java
libinglong: I have changed this comment.

apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/plugin/witness/WitnessTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * unit test for WitnessFinder
+ */
+public class WitnessTest {
+
+    private String className = "org.apache.skywalking.apm.agent.core.plugin.witness.WitnessTest";
```
wu-sheng, line 37: Using `WitnessTest#getClass` should be better?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/AbstractClassEnhancePluginDefine.java
```diff
@@ -64,12 +67,21 @@
         String[] witnessClasses = witnessClasses();
         if (witnessClasses != null) {
             for (String witnessClass : witnessClasses) {
-                if (!WitnessClassFinder.INSTANCE.exist(witnessClass, classLoader)) {
+                if (!WitnessFinder.exist(witnessClass, classLoader)) {
```
wu-sheng, line 70: I was not saying change this to static method. My meaning is you could get `var = WitnessFinder.INSTANCE before all loops.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/WitnessFinder.java
```diff
@@ -0,0 +1,82 @@
...
+public enum WitnessFinder {
+    INSTANCE;
+
+    private final Map<ClassLoader, TypePool> poolMap = new HashMap<ClassLoader, TypePool>();
+
+    /**
+     * @param classLoader for finding the witnessClass
+     * @return true, if the given witnessClass exists, through the given classLoader.
+     */
+    public static boolean exist(String witnessClass, ClassLoader classLoader) {
```
wu-sheng, line 38: Don't expose static method, including all methods below.

apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/plugin/witness/WitnessTest.java
libinglong: I am not sure If I can use WitnessTest.class.
I think the witness class should not be loaded before the app start.

wu-sheng: I am not asking you to change the mechanism. My point is this test class. This field doesn't need to be a hard coded string. You could use `getClass.getName` to instead of, right?

libinglong: Yes,you are right.
But it's good for the unit test, not good as demo for users.We should find a banlance.

wu-sheng: Good point, let's keep this unchanged.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/WitnessMethod.java
```diff
@@ -0,0 +1,48 @@
...
+@ToString
+public class WitnessMethod {
+
+    /**
+     * the class or interface name where the witness method is declared.
+     */
+    @Getter
+    private final String declaringClassName;
+    /**
+     * mather fo match the witness method
```
wu-sheng, line 38:
```suggestion
     * matcher to match the witness method
```
typos

docs/en/guides/Java-Plugin-Development-Guide.md
```diff
@@ -299,6 +299,31 @@ The following sections will tell you how to implement the interceptor.
 tomcat-7.x/8.x=TomcatInstrumentation
 ```
 
+4. Set up `witnessClasses` and/or `witnessMethods` if the instrumentation should be activated in specific versions.
+
+   Example:
+
+   ```java
+   //the plugin is activated only when the foo.Bar class exists.
```
wu-sheng, line 307:
```suggestion
   // The plugin is activated only when the foo.Bar class exists.
```

docs/en/guides/Java-Plugin-Development-Guide.md
```diff
@@ -299,6 +299,31 @@ The following sections will tell you how to implement the interceptor.
...
+   @Override
+   protected String[] witnessClasses() {
+     return new String[] {
+       "foo.Bar"
+     };
+   }
+   
+   //the plugin is activated only when the foo.Bar#hello method exists.
```
wu-sheng, line 315:
```suggestion
   // The plugin is activated only when the foo.Bar#hello method exists.
```

docs/en/guides/Java-Plugin-Development-Guide.md
```diff
@@ -299,6 +299,31 @@ The following sections will tell you how to implement the interceptor.
...
+   @Override
+   protected List<WitnessMethod> witnessMethods() {
+     List<WitnessMethod> witnessMethodList = new ArrayList<>();
+     WitnessMethod witnessMethod = new WitnessMethod("foo.Bar", ElementMatchers.named("hello"));
+     witnessMethodList.add(witnessMethod);
+     return witnessMethodList;
+   }
+   ```
+   For more example, See [WitnessTest.java](/apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/plugin/witness/WitnessTest.java)
```
wu-sheng, line 324:
```suggestion
   For more example, see [WitnessTest.java](/apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/plugin/witness/WitnessTest.java)
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/WitnessFinder.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * get TypePool.Resolution of the witness class
+     * @param witnessClass class name
+     * @param classLoader classLoader for finding the witnessClass
+     * @return TypePool.Resolution
+     */
+    private TypePool.Resolution getResolution(String witnessClass, ClassLoader classLoader) {
+        ClassLoader mappingKey = classLoader == null ? NullClassLoader.INSTANCE : classLoader;
+        if (!INSTANCE.poolMap.containsKey(mappingKey)) {
```
wu-sheng, line 51: `INSTANCE` is not necessary, I think. Or use `this`.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/WitnessMethod.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * matcher to match the witness method
+     */
+    @Getter
+    private final ElementMatcher<? super MethodDescription.InDefinedShape> elementMatcher;
+
+    public WitnessMethod(String declaringClassName, ElementMatcher<? super MethodDescription.InDefinedShape> elementMatcher) {
+        this.declaringClassName = declaringClassName;
+        this.elementMatcher = elementMatcher;
+    }
```
wu-sheng, line 46: Use `@RequiredArgsConstructor` annotation.

candyleer:
should fix CI first.

wu-sheng (approved):

wu-sheng (approved):

PR #6115, opened 2021-01-01
Fix the uuid field in GRPCConfigWatcherRegister is not updated

zifeihan:
### Fix the uuid field in GRPCConfigWatcherRegister is not updated.

The uuid field is not set, and is the same value every time request, modified to update uuid after each request is completed.


wu-sheng (approved):
LGTM

PR #6124, opened 2021-01-03
Fix bug that endpoint-name-grouping.yml can not be customized in docker / helm

kezhenxu94:
### Fix bug that `endpoint-name-grouping.yml` can not be customized in docker
The file is packaged into the jar hence there is no chance that it can be overrode by users through volume / config map.


wu-sheng (approved):

wu-sheng:
Why this causes all CIs fail?

wu-sheng (changes requested):
Seems you don't add `metadata-service-mapping.yaml` in the dist package binary.xml?

wu-sheng (approved):

PR #6131, opened 2021-01-05
Unify the service name rule in `metadata-service-mapping.yml` of `mx-…

kezhenxu94:
…mesh` analyzer and Envoy metrics service

### DIFF
--- a/.github/workflows/e2e.istio.yaml
+++ b/.github/workflows/e2e.istio.yaml
@@ -71,7 +71,9 @@ jobs:
           cd skywalking-kubernetes
           git reset --hard dd749f25913830c47a97430618cefc4167612e75
           cd chart
-          mkdir -p skywalking/files/conf.d/oap/ && echo 'serviceName: e2e::${LABELS.app}\nserviceInstanceName: ${NAME}' > skywalking/files/conf.d/oap/metadata-service-mapping.yaml
+          mkdir -p skywalking/files/conf.d/oap/ && \
+              echo 'serviceName: e2e::${LABELS.app}' > skywalking/files/conf.d/oap/metadata-service-mapping.yaml && \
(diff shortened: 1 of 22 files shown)


.github/workflows/e2e.istio.yaml
```diff
@@ -71,13 +71,15 @@ jobs:
...
           cd chart
+          mkdir -p skywalking/files/conf.d/oap/ && echo 'serviceName: e2e::${LABELS.app}\nserviceInstanceName: ${NAME}' > skywalking/files/conf.d/oap/metadata-service-mapping.yaml
           helm dep up skywalking
           helm -n istio-system install skywalking skywalking \
                --set fullnameOverride=skywalking \
                --set elasticsearch.replicas=1 \
                --set elasticsearch.minimumMasterNodes=1 \
                --set elasticsearch.imageTag=7.5.1 \
                --set oap.env.SW_ENVOY_METRIC_ALS_HTTP_ANALYSIS=${{ matrix.analyzer }} \
+               --set oap.env.K8S_SERVICE_NAME_RULE='e2e::${service.metadata.name}' \
```
wu-sheng, line 82: Is this for unifying the ip-mapping e2e for the same prefix?

wu-sheng (dismissed):

.github/workflows/e2e.istio.yaml
kezhenxu94: This config only takes effect in the `k8s-mesh` case, because we share the same `expected.data.yaml` between `k8s-mesh` and `mx-mesh`, so I just add the group to `k8s-mesh` case

wu-sheng (approved):

wu-sheng:
Seems `mx-mesh` still failing :P

wu-sheng (approved):

Issue #6080, opened 2020-12-25
mssql error


arronbin:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS, and JRE?

- Which company or project?

- What happened?
If possible, provide a way to reproduce the error. e.g. demo application, component version.

___
### Requirement or improvement
- Please describe your requirements or improvement suggestions.
Caused by: org.apache.ibatis.executor.ExecutorException: Error preparing statement. Cause: java.lang.ClassCastException: org.apache.skywalking.apm.plugin.jdbc.trace.SWPreparedStatement cannot be cast to com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement
at org.apache.ibatis.executor.statement.BaseStatementHandler.prepare(BaseStatementHandler.java:97)
at org.apache.ibatis.executor.statement.RoutingStatementHandler.prepare(RoutingStatementHandler.java:59)
at org.apache.ibatis.executor.SimpleExecutor.prepareStatement(SimpleExecutor.java:85)
at org.apache.ibatis.executor.SimpleExecutor.doUpdate(SimpleExecutor.java:49)
at org.apache.ibatis.executor.BaseExecutor.update(BaseExecutor.java:117)
at org.apache.ibatis.executor.CachingExecutor.update(CachingExecutor.java:76)
at sun.reflect.GeneratedMethodAccessor405.invoke(Unknown Source)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.lang.reflect.Method.invoke(Method.java:498)
at org.apache.ibatis.plugin.Plugin.invoke(Plugin.java:63)
at com.sun.proxy.$Proxy320.update(Unknown Source)
at sun.reflect.GeneratedMethodAccessor405.invoke(Unknown Source)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.lang.reflect.Method.invoke(Method.java:498)
at org.apache.ibatis.plugin.Plugin.invoke(Plugin.java:63)
at com.sun.proxy.$Proxy320.update(Unknown Source)
... 290 more

wu-sheng:
I am not familiar with the ibatis codes, why it is trying to casting?

[...]

PR #6133, opened 2021-01-05
Fix mssql-plugin occur ClassCastException when call the method of return generate key

zifeihan:
### Fix mssql-plugin occur ClassCastException when call the method of return generate key
     
fix https://github.com/apache/skywalking/issues/6080, 
Previously, I wrapped the return value of the prepareStatement method of SWPreparedStatement type, but in the method `com.microsoft.sqlserver.jdbc.SQLServerConnection#prepareStatement(java.lang.String, int, com.microsoft.sqlserver.jdbc.SQLServerStatementColumnEncryptionSetting)` , Type conversion, a `java.lang.ClassCastException: org.apache.skywalking.apm.plugin.jdbc.trace.SWPreparedStatement cannot be cast to com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement`, fix method: using the interceptor to replace wrapper.


apm-sniffer/apm-sdk-plugin/mssql-jdbc-plugin/src/main/resources/skywalking-plugin.def
```diff
@@ -16,3 +16,5 @@
 
 mssql-jdbc=org.apache.skywalking.apm.plugin.mssql.jdbc.define.DriverInstrumentation
 mssql-jdbc=org.apache.skywalking.apm.plugin.mssql.jdbc.define.ConnectionInstrumentation
+mssql-jdbc=org.apache.skywalking.apm.plugin.mssql.jdbc.define.PreparedStatementInstrumentation
+mssql-jdbc=org.apache.skywalking.apm.plugin.mssql.jdbc.define.StatementInstrumentation
```
wu-sheng, line 20: mssql plugin doesn't instrument statement before? I am a little confused.

wu-sheng:
You need to explain what you are doing and why you are doing in the PR, please fix this. 
From my understanding, your chance is using the interceptor to replace wrapper, if so, why these interceptors are not in the jdbc-common, instead, you created msql-common?

apm-sniffer/apm-sdk-plugin/mssql-jdbc-plugin/src/main/resources/skywalking-plugin.def
zifeihan: In the previous packaging mode, statement was packaged by `org.apache.skywalking.apm.plugin.jdbc.JDBCStatementInterceptor`.

zifeihan:
> You need to explain what you are doing and why you are doing in the PR, please fix this.
> From my understanding, your chance is using the interceptor to replace wrapper, if so, why these interceptors are not in the jdbc-common, instead, you created msql-common?

```
    public PreparedStatement prepareStatement(String sql,
            int flag,
            SQLServerStatementColumnEncryptionSetting stmtColEncSetting) throws SQLServerException {
        loggerExternal.entering(getClassNameLogging(), "prepareStatement", new Object[] {sql, new Integer(flag), stmtColEncSetting});
        checkClosed();
        SQLServerPreparedStatement ps = (SQLServerPreparedStatement) prepareStatement(sql, ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY,
                stmtColEncSetting);
        ps.bRequestedGeneratedKeys = (flag == Statement.RETURN_GENERATED_KEYS);
        loggerExternal.exiting(getClassNameLogging(), "prepareStatement", ps);
        return ps;
    }
```
For example, in this method com.microsoft.sqlserver.jdbc.SQLServerConnection#prepareStatement(java.lang.String, int, com.microsoft.sqlserver.jdbc.SQLServerStatementColumnEncryptionSetting), the return value of the prepareStatement method is forcibly converted. Because mssql-jtds and mssql-jdbc have certain similarities, a new mssql-commons module has been added.

wu-sheng:
You forget to update the changelog. Please fix.

wu-sheng (dismissed):

wu-sheng (approved):
Once the plugin test passed, it is good to me, as the expected data isn't changed.

PR #6120, opened 2021-01-02
Log collecting and query implementation

arugal:
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>.
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md).

https://github.com/apache/skywalking/issues/5930

PR #6139, opened 2021-01-06
Fix some influxdb plugin bug

cuiweiwei:
* If use JSON as InfluxDB.ResponseFormat then NumberFormatException maybe occur.
* Before set "time_bucket" query parameter in influxdb storage plugin，no zero value judgment.


dmsolr (dismissed):
LGTM

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/LogQuery.java
```diff
@@ -98,9 +98,10 @@ public Logs queryLogs(String metricName, int serviceId, int serviceInstanceId, S
         if (!Strings.isNullOrEmpty(stateCode)) {
             recallQuery.and(eq(STATUS_CODE, stateCode));
         }
-        recallQuery.and(gte(AbstractLogRecord.TIME_BUCKET, startTB))
-                   .and(lte(AbstractLogRecord.TIME_BUCKET, endTB));
-
+        if (startTB != 0 && endTB != 0) {
+            recallQuery.and(gte(AbstractLogRecord.TIME_BUCKET, startTB))
+                       .and(lte(AbstractLogRecord.TIME_BUCKET, endTB));
+        }
```
wu-sheng, line 104: Please revert this. I can't see the point of change this. Log feature is provided #6120

wu-sheng (changes requested):

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/LogQuery.java
cuiweiwei: In log.graphqls，LogQueryCondition.queryDuration is an optional field with annotation "The time range of log happened". If I understand correctly, when this field is empty it will query for all the logs. But the original code didn't do that, it can't find any log.

wu-sheng: I think the comment is very clear, https://github.com/apache/skywalking-query-protocol/blob/0eb55a7de358d35a3f8982bdd39ba7d21dde4f9c/log.graphqls#L53-L54

wu-sheng: The key is, your PR doesn't relate to this implementation. This query has been used in any place, before the PR I pointed to you.

cuiweiwei: My local code is behind https://github.com/apache/skywalking-query-protocol/blob/0eb55a7de358d35a3f8982bdd39ba7d21dde4f9c/log.graphqls#L53-L54，I only got the L52. So above changes are unnecessary? But i find that other storage plugin have the same zero value judgment, such as H2LogQueryDAO.java

wu-sheng: All log query(s) are not used, take a look  #6120. It is being introduced recently(haven't merged). The codes you are seeing are not working in any case.
Please revert change and avoid unnecessary conflict.

wu-sheng (approved):

PR #6147, opened 2021-01-06
Fix istio control plane ui template metric type mismatch issue

hanahmily:

as indicated by the title, UI metric type mismatches backend metric type.


wu-sheng (approved):
LGTM.

wu-sheng:
Please don't sync the PR by yourself, we are facing limited CI resources. The PMC will take care of the approved PR one by one.

Issue #6136, opened 2021-01-06
ReadWriteSafeCache simultaneous read-write support


andyzzl:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [x] Feature or performance improvement

___

### Requirement or improvement
- Please describe your requirements or improvement suggestions.

Currently, `ReadWriteSafeCache` holds two buffer pointers to read-buffer and write-buffer respectively. Writer thread write data to write-buffer, while reader thread read data from read-buffer. Pointers will be swapped after reading finished. All operations are controlled by one single `ReentrantLock`. As follows:

```java
public void write(List<T> data) {
    lock.lock();
    try {
        data.forEach(writeBufferPointer::accept);
    } finally {
        lock.unlock();
    }
}

public List<T> read() {
    lock.lock();
    try {
        // Switch the read and write pointers, when there is no writing.
        BufferedData<T> tempPointer = writeBufferPointer;
        writeBufferPointer = readBufferPointer;
        readBufferPointer = tempPointer;

        return readBufferPointer.read();
    } finally {
        lock.unlock();
    }
}
```

I'm confused that why two buffers are used here. It seems like one single buffer can do the same job. Maybe the switchable buffers here are used for simultaneous read-write. But it's not possible with only one lock, cause all operations to `ReadWriteSafeCache` are blocked.

To support simultaneous read-write, I synchronize the operations with two locks, and move the swapping operations to writing method, just after writing operation is finished, only if read-buffer is empty. As follows:

```
public void write(List<T> data) {
    writeBufferLock.lock();
    try {
        data.forEach(writeBufferPointer::accept);
        try {
            readBufferLock.lock();
            if (readBufferPointer.isEmpty()) {
                switchPointers();
            }
        } finally {
            readBufferLock.unlock();
        }
    } finally {
        writeBufferLock.unlock();
    }
}

public List<T> read() {
    readBufferLock.lock();
    try {
        return readBufferPointer.read();
    } finally {
        readBufferLock.unlock();
    }
}

private void switchPointers() {
    BufferedData<T> tempPointer = writeBufferPointer;
    writeBufferPointer = readBufferPointer;
    readBufferPointer = tempPointer;
}
```

Only swapping operation need to acquire two lock. Reading and writing operation just need to acquire their own lock. Empty check avoids unnecessary swapping and disorder problem of data.

[...]

PR #6149, opened 2021-01-07
ReadWriteSafeCache concurrency read-write

andyzzl:
### Improve the performance of ReadWriteSafeCache
- [ ] The benchmark result.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<6136>.

Made adjustments to read() method of ReadWriteSafeCache to support concurrency read-write. Issue: #6136 .

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/data/ReadWriteSafeCache.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/data/ReadWriteSafeCache.java
@@ -34,13 +34,9 @@ public class ReadWriteSafeCache<T> {
      */
     private volatile BufferedData<T> writeBufferPointer;
     /**
-     * Write-buffer lock.
-     *
-     * Access of read-buffer is no need to be locked,
-     * because only {@link org.apache.skywalking.oap.server.core.storage.PersistenceTimer}
-     * thread reads data from read-buffer. No concurrency reading exists.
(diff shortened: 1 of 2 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/data/ReadWriteSafeCache.java
```diff
@@ -34,9 +34,13 @@
...
     /**
-     * Read/Write lock.
+     * Write-buffer lock.
+     *
+     * Access of read-buffer is no need to be locked,
+     * because only {@link org.apache.skywalking.oap.server.core.storage.PersistenceTimer}
+     * thread reads data from read-buffer. No concurrency reading exists.
      */
-    private final ReentrantLock lock;
+    private final ReentrantLock writeLock;
```
wu-sheng, line 43: Why rename this field? Basically, we don't change anything related to this.

wu-sheng: The lock is still being used in read and write, please revert the name.

andyzzl: Sorry, I'll revert it.

wu-sheng (approved):
LGTM

PR #6153, opened 2021-01-07
Fix `timeBucket` not taking effect in EqualsAndHashCode annotation.

zifeihan:
### Fix timeBucket not taking effect in EqualsAndHashCode annotation

Fix `timeBucket` not taking effect in EqualsAndHashCode annotation, Because the hashcode and equals methods generated by EqualsAndHashCode do not contain the properties of the parent class, the data related to the topology map is inaccurate during aggregation. use https://github.com/zifeihan/friday look runtime class info in jvm.

before fix:
image
after fix:
image
image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,7 @@ Release Notes.
 * Fix ArrayBlockingQueueBuffer's useless `IF_POSSIBLE` mode list
 * Support building gRPC TLS channel but CA file is not required.
 * Add Dolphinscheduler plugin definition.
+* Make sampling still works when the trace ignores plug-in activation.
 * Fix mssql-plugin occur ClassCastException when call the method of return generate key.
 
 #### OAP-Backend
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -49,6 +49,7 @@ Release Notes.
 * Fix the priority setting doesn't work of the ALS analyzers.
 * Fix bug that `endpoint-name-grouping.yml` is not customizable in Dockerized case.
 * Fix bug that istio version metric type on UI template mismatches the otel rule.
+* Fix `entityId` not taking effect in EqualsAndHashCode annotation
 
 #### UI
 * Fix un-removed tags in trace query.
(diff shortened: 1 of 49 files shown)


CHANGES.md
```diff
@@ -49,6 +49,7 @@ Release Notes.
 * Fix the priority setting doesn't work of the ALS analyzers.
 * Fix bug that `endpoint-name-grouping.yml` is not customizable in Dockerized case.
 * Fix bug that istio version metric type on UI template mismatches the otel rule.
+* Fix `entityId` not taking effect in EqualsAndHashCode annotation
```
wu-sheng, line 52: I think it is `timestamp` is not working.

wu-sheng:
Please add the UTs to verify the `equals`

zifeihan:
> Please add the UTs to verify the `equals`

Ok, I will add test cases.

CHANGES.md
zifeihan: done.

oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/analysis/manual/relation/endpoint/EndpointCallRelation.java
```diff
@@ -0,0 +1,36 @@
+package org.apache.skywalking.oap.server.core.analysis.manual.relation.endpoint;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class EndpointCallRelation {
```
wu-sheng, line 6:
```suggestion
public class EndpointCallRelationTest {
```

oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/analysis/manual/relation/endpoint/EndpointCallRelationTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+package org.apache.skywalking.oap.server.core.analysis.manual.relation.endpoint;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class EndpointCallRelationTest {
+    @Test
+    public void testServiceInstanceRelationClientSideMetricsEquals() {
+        EndpointRelationServerSideMetrics thisObject = new EndpointRelationServerSideMetrics();
```
wu-sheng, line 27: I think the codes are incorrect. Please fix.

zifeihan: Thank you, fixed.

wu-sheng (approved):

Issue #6114, opened 2021-01-01
Dynamic Configuration for the agent


wu-sheng:
Hi @xuanyu66, as https://github.com/apache/skywalking/issues/6105#issuecomment-753248303 mentioned, we start the discussion about the dynamic configuration for the agent. FYI all @apache/skywalking-committers , if any of you wants to work on this, please reply.

We have command system at the agent side, we just need to extend a new command for dynamic configuration. Like we did in the backend side, the configurations should be separated from the the initial agent.config definition. The implementation decides whether they support the configuration dynamically changes.

From my preference, we may need a new gRPC service for the agent for dynamic configuration. Note, we need a ID(UUID or timestamp) to represent the configuration to avoid responding the configuration to the agent repeatedly. Also, the configurations of different services are different.

mrproliu:
I'm interested in this.

kezhenxu94:
> From my preference, we may need a new gRPC service for the agent for dynamic configuration.

In terms of gRPC implementation, what's the idea of modifying the configurations? If I understand correctly, this is going to be similar with the backend dynamic configurations, one should expose a gRPC service providing the latest configs to the OAP, and the OAP responds these to the agent, right?

```text

| user-defined gRPC service (emitting agent configs) |
                       |
                       | fetch (OAP behaves like a "client" side)
                       |
                       V
                    | OAP |
                       | 
                       | response via command system
                       |
                       V
                   | agent |

```

If that's the case, I would suggest reversing the roles (server / client) of OAP and "user-defined gRPC service", which makes it possible for the CLI to update the configs without a long-running process.

wu-sheng:
>   In terms of gRPC implementation, what's the idea of modifying the configurations? If I understand correctly, this is going to be similar with the backend dynamic configurations, one should expose a gRPC service providing the latest configs to the OAP, and the OAP responds these to the agent, right?

@kezhenxu94 The agent configuration isn't the agent dynamic configuration. They may logically be supported in the same features, but they are plugins' choices. 

>  If that's the case, I would suggest reversing the roles (server / client) of OAP and "user-defined gRPC service", which makes it possible for the CLI to update the configs without a long-running process.

The agent wouldn't accept any remote request, because of 2 reasons
1. High security leak risk. Users could accidentally open this port on the internal(even intranet is same), then the app could be controlled remotely.
2. Dynamic configuration of the agent should be driven and managed by the configuration center(s), as we have implemented the integration with most popular tools.

[...]

PR #6164, opened 2021-01-10
Support trace path ignore with patterns match.

JaredTan95:
### Bring agent trace ignore feature to OAP and support dynamic update patterns.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -56,6 +56,7 @@ Release Notes.
 * Improve ReadWriteSafeCache concurrency read-write performance
 * Fix bug that if use JSON as InfluxDB.ResponseFormat then NumberFormatException maybe occur.
 * Fix `timeBucket` not taking effect in EqualsAndHashCode annotation of some relationship metrics.
+* Support trace path ignore with patterns match.
 
 #### UI
 * Fix un-removed tags in trace query.
--- a/apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/matcher/FastPathMatcher.java
+++ b/apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/matcher/FastPathMatcher.java
@@ -16,7 +16,7 @@
  *
  */
 
-package org.apache.skywalking.apm.plugin.trace.ignore.matcher;
+package org.apache.skywalking.apm.util.matcher;
 
 public class FastPathMatcher implements TracePathMatcher {
     @Override
--- a/apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/matcher/TracePathMatcher.java
+++ b/apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/matcher/TracePathMatcher.java
@@ -16,7 +16,7 @@
  *
  */
 
-package org.apache.skywalking.apm.plugin.trace.ignore.matcher;
+package org.apache.skywalking.apm.util.matcher;
 
 public interface TracePathMatcher {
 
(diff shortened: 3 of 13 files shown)


wu-sheng:
Before reviewing the codes, I want to ask for the scenarios about why need this at the backend. What is the case we are opening the backend-side ignorance feature?

JaredTan95:
like sampling logic, we can sample on the agent-side or backend-side. 
So, we can ignore trace on agent source, or we can do this on the backend side. The 2nd scenario is that when the storage resources are limited and there is no way to dynamically update the agent side configurations to ignore more traces currently. 
And I looking forward to the [agent side dynamic update feature](https://github.com/apache/skywalking/issues).

wu-sheng:
@mrproliu Do you have any progress to share about #6114 ? 

@JaredTan95 Will this feature still exist if 6114 completes? This seems a little duplicate.

mrproliu:
> @mrproliu Do you have any progress to share about #6114 ?
> 
> @JaredTan95 Will this feature still exist if 6114 completes? This seems a little duplicate.

Not yet. I'm still working on the Zabbix adaptor. After the Zabbix adaptor finished, I could start the agent side dynamic configuration.

JaredTan95:
> @mrproliu Do you have any progress to share about #6114 ?
> 
> @JaredTan95 Will this feature still exist if 6114 completes? This seems a little duplicate.

I don't think there's a duplicate, the scope of the agent side is a single instance, and the scope of the backend is the entire system. Both of them can exist.

wu-sheng:
> I don't think there's a duplicate, the scope of the agent side is a single instance, and the scope of the backend is the entire system. Both of them can exist.

That is another thing I want to discuss. The current implementation has performance concern. Because basically, ignorance should be service-oriented. For this feature, we need to be very carefully about the performance impact. 
Because `cost` * N agent instances ==> `cost` of M OAP instances. N is much bigger than M.

wu-sheng:
And once you have the rules across services, the rules must check all endpoints of all services.

wu-sheng:
Note, base on CI
```
2021-01-10T13:12:44.7190051Z [INFO] Results:
2021-01-10T13:12:44.7190577Z [INFO] 
2021-01-10T13:12:44.7191080Z [ERROR] Failures: 
2021-01-10T13:12:44.7192309Z [ERROR]   TraceIgnorePathWatcherTest.testDynamicUpdate:64 
2021-01-10T13:12:44.7194630Z Expected: is "/ignore/path/**"
2021-01-10T13:12:44.7195246Z      but: was ""
```
The test is not working. But let's make the design clear first.

JaredTan95:
> Note, base on CI
> 
> ```
> 2021-01-10T13:12:44.7190051Z [INFO] Results:
> 2021-01-10T13:12:44.7190577Z [INFO] 
> 2021-01-10T13:12:44.7191080Z [ERROR] Failures: 
> 2021-01-10T13:12:44.7192309Z [ERROR]   TraceIgnorePathWatcherTest.testDynamicUpdate:64 
> 2021-01-10T13:12:44.7194630Z Expected: is "/ignore/path/**"
> 2021-01-10T13:12:44.7195246Z      but: was ""
> ```
> 
> The test is not working. But let's make the design clear first.

okay

wu-sheng:
@JaredTan95 It seems CDS(https://github.com/apache/skywalking-data-collect-protocol/blob/master/language-agent/ConfigurationDiscoveryService.proto) is on the way faster than I thought.

Refer #6114 

If we have that, do you think still be necessary for the backend ignorance mechanism?

JaredTan95:
I'm also thinking about this logic these days, as you said, service-oriented.

like providing a configuration for the service and service's ignore trace path as follows:

```yaml
servcie-a:
    - /your/path/1/**
    - /your/path/2/**
servcie-b:
    - /your/path/1/**
```

But, this way may have a performance impact when items too much in this config file. If CDS is WIP,  I prefer that filter at the source of the agent data. So, I'll close this PR.

Issue #6172, opened 2021-01-11
Java.lang.ClassFormatError: Duplicate interface name "org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/EnhancedInstance"


zifeihan:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?

___
### Bug
- Which version of SkyWalking, OS, and JRE?
SkyWalking version:8.4.0-SNAPSHOT, 
jdk version:jdk1.8
os version: mac 10.14.2
- Which company or project?

- What happened?
If possible, provide a way to reproduce the error. e.g. demo application, component version.
When there is an app to use skywalking, application report this error, and crash.
```
2021-01-11 11:17:53.580|[main]|ERROR|org.springframework.boot.SpringApplication - Application run failed
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'reCreateChatRoomJob' defined in URL [jar:file:/Users/echo/Downloads/target/target.jar!/BOOT-INF/classes!/xxx.xxx.xxx/bg/job/ReCreateChatRoomJob.class]: Initialization of bean failed; nested exception is org.springframework.aop.framework.AopConfigException: Could not generate CGLIB subclass of class com.zhangmen.yun.csw.marketlesson.bg.job.ReCreateChatRoomJob$$EnhancerBySpringCGLIB$$5a04f5ab: Common causes of this problem include using a final class or a non-visible class; nested exception is org.springframework.cglib.core.CodeGenerationException: java.lang.ClassFormatError-->Duplicate interface name "org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/EnhancedInstance" in class file xxx.xxx.xxx/bg/job/ReCreateChatRoomJob$$EnhancerBySpringCGLIB$$f3689c3c
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:601)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:515)
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:320)
[...]

PR #6173, opened 2021-01-11
Fix Java.lang.ClassFormatError: Duplicate interface name "org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/EnhancedInstance"

zifeihan:
### Fix Java.lang.ClassFormatError: Duplicate interface name "org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/EnhancedInstance"

The cause and fix of the problem are as follows, https://github.com/apache/skywalking/issues/6172


wu-sheng (approved):
LGTM. Let's wait for the all plugin tests to verify.

PR #6174, opened 2021-01-11
SharingServerConfig properties is not correct, contextPath -> restCon…

xingren23:
SharingServerConfig properties is not correct, contextPath -> restConnextPath


wu-sheng (approved):
Good catch. LGTM.

Issue #5980, opened 2020-12-09
[Core] Add new concept, EVENT


wu-sheng:
Hi SkyWalking team

I have been thinking about this concept for a long time, I am going to officially propose this concept now.
Right now, we have tracing and metrics fully supported, logging is on the way due to the browser provided and satellite is going to do more in this field. We have the typical 3 concepts, logging/tracing/metrics, in the observability.
But in the real world, it is more complex than those 3. **EVENT** is one of the most typical things. 

System performance could be changed due to different variables, such as upgrading, rebooting, chaos testing, hosted VM overload. Some of them are detectable in some tools, such as rebooting in k8s, upgrading in CI/CD pipeline. But at the same time, they are only accessible through agent tech. We call this the event.

**Event** should include owner(affected instance/service/etc.) with time range(start/end). This would be easy to implement in OAP as actually, it is a new kind of record only. The major difference is, one event would include 2 records, typically, one for the start, the other for ending, linked by an UUID. Because an event could take a long time, 2 records could make the event started shows up ASAP.

- gRPC/HTTP interface should be provided to report events. CLI should follow this to give a command to report events too.
- GraphQL interface should be provided for UI to visualize the event.

As a special request to UI, @Fine0830 
![image](https://user-images.githubusercontent.com/5441976/101591380-67877c00-3a27-11eb-9719-89f4ba135727.png)
We need to add a timeline based on the selected time range, showing the event dot(if even hasn't stopped), or a bar(covering start to end timestamp). When the mouse clicks the dot or bar, all charts(if in the dashboard), should show a vertical line(or range) to highlight the metrics values at this special time points.

kezhenxu94:
Who will be the owner of this feature? I’d like to take this if no one is working on it yet.

wu-sheng:
We don't have yet.

kezhenxu94:
> We don't have yet.

OK. Let me kick it off by some further discussions (later)

wu-sheng:
@BFergerson You pinged me about the similar thing, if you like, join the discussion here.

kezhenxu94:
@wu-sheng @BFergerson I just drafted a simple doc about the event concept, please take a look at https://github.com/apache/skywalking/pull/6183

kezhenxu94:
[...]

PR #6183, opened 2021-01-12
Add a new concept "Event" and its implementations to collect events

kezhenxu94:
### Add a new concept "Event" and its implementations to collect events
- [x] If this is non-trivial feature, paste the links/URLs to the design doc.  https://github.com/apache/skywalking/issues/5980
- [x] If it's UI related, attach the screenshots below. UI is not yet implemented, should be in another repo.
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes partially #5980.


docs/en/concepts-and-designs/README.md
```diff
@@ -23,3 +23,5 @@ SkyWalking already support.
   - [Overview](ui-overview.md). A simple brief about SkyWalking UI.
 - CLI
   - [SkyWalking CLI](https://github.com/apache/skywalking-cli). A command line interface for SkyWalking.
+- Events
+  - [Events](event.md). Introduce the concept of "Event" in SkyWalking.
```
wu-sheng, line 27: I think this should be listed here? https://github.com/apache/skywalking/blob/845312f28b61bf51beba5a3205dd04e38c0ae08b/docs/en/concepts-and-designs/overview.md#why-use-skywalking

kezhenxu94: I'll list it there once the doc is settled down

docs/en/concepts-and-designs/event.md
```diff
@@ -0,0 +1,53 @@
...
+
+The start time of the event. This field is mandatory when an event is reported.
+
+### End Time
+
+The end time of the event. This field may be empty if the event has not stopped yet, otherwise it should be a valid timestamp after `startTime`.
+
+**NOTE:** When reporting an event, you typically call the report function twice, one for starting of the event and the other one for ending of the event, with the same UUID.
+There are also cases where you have both the start time and end time already, for example, when exporting events from a 3rd-party system, the start time and end time are already known so that you can
+call the report function only once.
```
kezhenxu94, line 53: @wu-sheng here is a little different from what you proposed in #5980 , where you prefer 2 records for each event, but because we have a UUID for each event, maybe we can just leave the end time to be empty when the event is not finished, and update it with the same UUID when it's done?

docs/en/concepts-and-designs/event.md
```diff
@@ -0,0 +1,53 @@
...
+while `Error` is considered as unexpected operations, such as `Crash` events, therefore we can mark them with different colors to be easier identified.
+
+### Message
+
+The detail of the event that describes why this event happened. This should be a one-line message that briefly describes why the event is reported. Examples of a `Crash` event may be something like `Run out of memory`.
+It's NOT encouraged to include the detailed logs of this event, such as the exception stack trace.
+
+### Start Time
+
+The start time of the event. This field is mandatory when an event is reported.
```
wu-sheng, line 45:
```suggestion
The start time of the event. This field is mandatory when an event occurs.
```

kezhenxu94: Let me apply the suggestions in batch after all suggestions are submitted

docs/en/concepts-and-designs/event.md
wu-sheng: If you want to update, it is fine. I was considering the duration of the event could be long, so session(storage level) mechanism could fail. But after all, the traffic of events should be limited. It is fine to do update.

BFergerson, line 53:
```suggestion
There are also cases where you have both the start time and end time already, for example, when exporting events from a 3rd-party system, the start time and end time are already known so that you can call the report function only once.
```

docs/en/concepts-and-designs/event.md
```diff
@@ -0,0 +1,53 @@
...
+The object that the event occurs on. In the concepts of SkyWalking, the object is typically service, service instance, etc.
+
+### Name
+
+The name of the event. For example, `Start`, `Stop`, `Crash`, `Reboot`, `Upgrade` etc.
+
+### Type
+
+The type of the event. This field is friendly for UI visualization, where events of type `Normal` is considered as normal operations,
+while `Error` is considered as unexpected operations, such as `Crash` events, therefore we can mark them with different colors to be easier identified.
```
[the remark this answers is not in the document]
wu-sheng: Why `are`? I think this is `Error` type only?

docs/en/concepts-and-designs/event.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Message
+
+The detail of the event that describes why this event happened. This should be a one-line message that briefly describes why the event is reported. Examples of a `Crash` event may be something like `Run out of memory`.
+It's NOT encouraged to include the detailed logs of this event, such as the exception stack trace.
```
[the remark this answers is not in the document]
wu-sheng: Are you asking for collaboration-obs-entity? Such as traceid, logid(s), etc.

docs/en/concepts-and-designs/event.md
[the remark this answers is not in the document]
BFergerson: Must have read wrong. It's fine as-is.

docs/en/concepts-and-designs/event.md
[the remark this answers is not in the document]
BFergerson: I'm not sure what collaboration-obs-entity means. I'm just saying there should be a way to allow events to have user data. Besides the use case I mentioned above, the use case I was imagining was when the event is of type `Upgrade`; you could add `from.version=1.2`/`to.version=1.3` in the user data somewhere. That way UI/CLI/etc that display the events can show additional information related to the event.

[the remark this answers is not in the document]
wu-sheng: I think `from.version=1.2/to.version=1.3` is reasonable in the message. @kezhenxu94 WDYT? I think the typical thing should not be included  is `exception stack`, which belongs to logs or trace/span.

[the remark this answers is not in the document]
kezhenxu94: > I think `from.version=1.2/to.version=1.3` is reasonable in the message. @kezhenxu94 WDYT? I think the typical thing should not be included  is `exception stack`, which belongs to logs or trace/span.

Exactly. `from.version=1.2/to.version=1.3` or something like `Upgrade from 1.2 to 1.3` is valid in my mind.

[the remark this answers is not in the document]
BFergerson: If you guys don't feel it's valid then disregard it. My thinking here was mainly to provide a way for static typing. I suppose you could always just encode something in JSON and put it in `message` if you need something more structural.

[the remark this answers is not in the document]
wu-sheng: The message structure is really a little complex. We could offer a field with JSON type to describe it as parameters, which provides a structure type, but we should provide recommended data format examples, for upgrading, OOM, kill, etc.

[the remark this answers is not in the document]
BFergerson: I agree about the recommended data formats. I just don't see how you do that with something as simple as `Run out of memory`. That's good for a summary but IMO it would be nice to have the ability to attribute additional information. I don't mind if it's through some JSON field, meta-map, etc.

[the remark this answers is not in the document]
kezhenxu94: > I just don't see how you do that with something as simple as `Run out of memory`. That's good for a summary but IMO it would be nice to have the ability to attribute additional information. I don't mind if it's through some JSON field, meta-map, etc.

Having a JSON / meta-map field sounds good to me, but because we don't have static typings for that field, users are free to report random text of that JSON / meta-map field.

I'm also wondering how will the structured field be used. There are many events (some may be unknown for now), whose JSON field properties are indeterministic.

If we want to provide a JSON field, we may need to list (recommended) properties of the JSON field for some common events, right?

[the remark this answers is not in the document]
wu-sheng: > If we want to provide a JSON field, we may need to list (recommended) properties of the JSON field for some common events, right?

Yes. We need to. But we could add this one by one, such as in my mind, an upgrade event could be
```
- type: Upgrade
- message: Upgrade from ${prev_version} to ${latest_version}.
- parameter: {"prev_version" : "1.2"; "latest_version" : "1.3" }
```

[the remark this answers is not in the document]
BFergerson: The above would be my preferred solution as well.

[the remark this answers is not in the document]
kezhenxu94: I added the `parameters` field

docs/en/concepts-and-designs/event.md
kezhenxu94: I finally end up with the origin proposal in #5980 , 1 event 2 records

kezhenxu94:
E2E tests are added, this can be reviewed somehow

BFergerson:
I gave this a look and try. The only things I've noticed that seem problematic is that `uuid` is not guaranteed to be unique. You can re-use the same `uuid` and the event system will accept it. Also, the `queryEvents` endpoint allows you to not define the start/end/step. I don't think other endpoints allow you not to specify the time range. Seems like it should be required.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/event/EventRecord.java
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.skywalking.oap.server.core.storage.StorageBuilder;
+import org.apache.skywalking.oap.server.core.storage.annotation.Column;
+
+import static org.apache.skywalking.oap.server.core.source.DefaultScopeDefine.EVENT;
+
+@Getter
+@Setter
+@ScopeDeclaration(id = EVENT, name = "Event")
+@Stream(name = EventRecord.INDEX_NAME, scopeId = EVENT, builder = EventRecord.Builder.class, processor = RecordStreamProcessor.class)
+public class EventRecord extends Record {
```
wu-sheng, line 40: Based on https://github.com/apache/skywalking/pull/6183#issuecomment-771260649, I want to discuss about, whether should be recognized as `Record` or a `Metric`.

As a metric, we could provide 
```
@Stream(name = EVENT, scopeId = EVENT,
    builder = Event.Builder.class, processor = MetricsStreamProcessor.class)
@MetricsExtension(supportDownSampling = false, supportUpdate = true)
```
The advantages of being this are, you could write custom `combine`/`merge` logic, and no downsampling, also still available to query manually with new GraphQL.
As the merge logic is controlled, rather than record(pure adding), we could ignore(but not reject @BFergerson ) the duplicate UUID events, or duplicate begin or stop commands.

kezhenxu94:
> I gave this a look and try. The only things I've noticed that seem problematic is that `uuid` is not guaranteed to be unique. You can re-use the same `uuid` and the event system will accept it. 

At OAP side, we choose to adopt the strategy "two records for one event" because we don't want to do updates for time series data like this. "Two records for one event" results in that no guarantee on the uniqueness of uuid at db level. If the users use the same UUIDs then the events should be considered as a single event according to our protocol. 

> Also, the `queryEvents` endpoint allows you to not define the start/end/step. I don't think other endpoints allow you not to specify the time range. Seems like it should be required.

I'm not sure about the fact of what you said. Requiring those is easy at protocol level, though we may not want to use them when there is no necessity (e.g. when you have a uuid while the time is unsure but you have to give one). 

> Also, I don't seem to be able to update events. The following two commands:
> 
> ```
> 
> ./swctl-latest-linux-amd64 event r --uuid uuid3 --service service --name name --message message --startTime 1612225542333
> 
> ./swctl-latest-linux-amd64 event r --uuid uuid3 --service service --name name --message message --startTime 1612225542333 --endTime 1612225542333
> 
> ```
> 
> 
> 
> Results in:
> 
> ```
> 
> {
> 
>   "data": {
> 
>     "result": {
> 
>       "events": [
> 
>         {
> 
>           "name": "name",
> 
>           "source": {
> 
>             "service": "service",
> 
>             "serviceInstance": "",
> 
>             "endpoint": ""
> 
>           },
> 
>           "startTime": 1612225542333,
> 
>           "endTime": 0,
> 
>           "message": "message",
> 
>           "parameters": [],
> 
>           "uuid": "uuid3"
> 
>         }
> 
>       ]
> 
>     }
> 
>   }
> 
> }
> 
> ```

That must be a bug if it is really the case, which I don't think so, if you are using ElasticSearch as storage, attention that the writes are async and may need some time to flush them to ES, can you try some time later after the CLI commands? (e.g. 30 seconds or so).

> 
> I'm using skywalking-cli @ https://github.com/apache/skywalking-cli/commit/ae085deb32434a0859e49c45819fed7a5c4010b5. It also seems a bit odd that you have to set `Required flags "uuid, service, name, message, startTime"` even when updating an event. All that should be necessary is the `uuid` as the other fields may have already been set.
> 

@fgksgf can you please re-organize the flags' verification? Let's remove all `required` and do composite verifications later.

> Finally, there does not seem to be a way to set event parameters via skywalking-cli.

The parameters are set as arguments, i.e. `./swctl-latest-linux-amd64 event r --uuid uuid3 --service service --name name --message message --startTime 1612225542333 --endTime 1612225542333 key1=value key2=value2`

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/event/EventRecord.java
kezhenxu94: Well. Seems metrics type is more suitable for events!

wu-sheng: `@MetricsExtension` is a very tricky thing :P I added it when we adopted the metadata as a kind of metrics.

BFergerson:
@kezhenxu94, my thinking was that if you don't include the `uuid` then you have to include other conditions. If having the ability to return all events is something that's at least non-detrimental then go ahead. It just seemed odd that you can't get all logs, all traces, or all metrics so it seemed to make sense not to be able to get all events.

As for the non-updating issue, I'm not using ES. I'm able to produce that issue on H2.

kezhenxu94:
@BFergerson pre-verification "time field is required when uuid is absent" is now added when querying

@wu-sheng the event record is now changed to metrics type

wu-sheng:
> @BFergerson pre-verification "time field is required when uuid is absent" is now added when querying
> 
> @wu-sheng the event record is now changed to metrics type

CI fails somehow?

oap-server/server-receiver-plugin/skywalking-event-receiver-plugin/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  ~
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>server-receiver-plugin</artifactId>
+        <groupId>org.apache.skywalking</groupId>
+        <version>8.4.0-SNAPSHOT</version>
```
BFergerson, line 24:
```suggestion
        <version>8.5.0-SNAPSHOT</version>
```

oap-server/server-receiver-plugin/skywalking-event-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/event/listener/TrafficAnalysisListener.java
```diff
@@ -0,0 +1,121 @@
...
+import org.apache.skywalking.oap.server.core.source.SourceReceiver;
+import org.apache.skywalking.oap.server.library.module.ModuleManager;
+
+import static java.util.Objects.nonNull;
+
+/**
+ * Generate service, service instance and endpoint traffic by event data.
+ */
+@RequiredArgsConstructor
+public class TrafficAnalysisListener implements EventAnalysisListener {
```
wu-sheng, line 42: I am thinking whether do we need this as event actually is an attachment to existing entity. So, if there is no mate data of service/instance/endpoint existing now, we have no place to show event. And manually creating the traffic(aka meta) seems unnecessary. WDYT?

docs/en/concepts-and-designs/README.md
wu-sheng: Some other documents should be updated
1. Receiver documentation
2. [Protocol doc](https://github.com/apache/skywalking/blob/master/docs/en/protocols/README.md), should add whole new section in `Probe Protocols`, as `Event Report Protocol` before `3rd-party instrument protocol`

oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -422,3 +422,7 @@ configuration-discovery:
   selector: ${SW_CONFIGURATION_DISCOVERY:default}
   default:
     disableMessageDigest: ${SW_DISABLE_MESSAGE_DIGEST:false}
+
+receiver-event:
+  selector: ${SW_RECEIVER_EVENT:default}
```
wu-sheng, line 427: This config should be added in `configuration-vocabulary`

oap-server/server-receiver-plugin/skywalking-event-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/event/service/EventAnalysisService.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.apache.skywalking.apm.network.event.v3.Event;
+import org.apache.skywalking.oap.server.library.module.Service;
+
+/**
+ * Analyze the collected event data.
+ */
+public interface EventAnalysisService extends Service {
+    void analyze(Event event);
+}
```
wu-sheng, line 29: Why do we need this as a `Service`? `Service` is the internal concept for across-module interaction, as a receiver and especially event receiver, I don't think this is the case.
And if this will(for event, may not), you could find there are `*-analyzer`(s) module exposing the interface.

wu-sheng:
@kezhenxu94 I think you also should remove `close xxx` from your PR, due to we need UI part works after merge this.

wu-sheng:
Changelog has not been updated.

wu-sheng (commented):
Could we reduce the number of `org.apache.skywalking.oap.server.receiver.event`'s sub packages?
Most of them seems having 1 and 2 classes in there. From my understanding, only grpc handler should be in separated folder as we may need to add more handler such as for HTTP.

____
And I want to discuss a little about EventService, should we consider having an event report mechanism inside the OAP server? Such as, when the service relation metrics causes alarm at the server side, could we trigger the event of downstream indicating `Your upstream is failing`

If you agree this case makes sense, we should move this service at the core level. The `listener`/`factory`/`analyzer` should be able to be easier than right now. It just generate a manual metric with 5-6 classes. But basically, it is a simple logic.

kezhenxu94:
> And I want to discuss a little about EventService, should we consider having an event report mechanism inside the OAP server? Such as, when the service relation metrics causes alarm at the server side, could we trigger the event of downstream indicating `Your upstream is failing`

Yes this is necessary, but let's keep this PR minimal to reducing the workload of reviewers, I'll record a todo list in the issue later

wu-sheng:
> Yes this is necessary, but let's keep this PR minimal to reducing the workload of reviewers, I'll record a todo list in the issue later

I don't mean we need to support that event detection, I agree, that is for longer term. That is just my example of proposing, move the event service into the core level, and make the codes simpler.

docs/en/concepts-and-designs/event.md
```diff
@@ -0,0 +1,56 @@
...
+SkyWalking backend supports three protocols to collect events, gRPC, HTTP, and Kafka. Any agent or CLI that implements one of these protocols can report events to SkyWalking.
+Currently, the officially supported clients to report events are:
+
+- [ ] Java Agent Toolkit: Use the Java agent toolkit to report events from inside the applications.
+- [x] SkyWalking CLI: Use the CLI to report events from the command line interface.
+- [ ] Kubernetes Event Exporter: Deploy an event exporter to refine and report Kubernetes events.
+
+## Event Definition
+
+An event contains the following fields. The definitions of event can be found at the [protocol repo](#TODO)
```
wu-sheng, line 19: Please finish this TODO

oap-server/analyzer/event-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/event/listener/EventRecordAnalyzerListener.java
```diff
@@ -0,0 +1,79 @@
...
+            event.setServiceInstance(namingControl.formatInstanceName(source.getServiceInstance()));
+            event.setEndpoint(namingControl.formatEndpointName(source.getService(), source.getEndpoint()));
+        }
+
+        event.setName(e.getName());
+        event.setType(e.getType().name());
+        event.setMessage(e.getMessage());
+        event.setParameters(GSON.toJson(e.getParametersMap()));
+        event.setStartTime(e.getStartTime());
+        event.setEndTime(e.getEndTime());
```
wu-sheng, line 62: You never `#setTimeBucket`, I think you need to check manually when use ElasticSearch storage.
As the time-series data, you need the unchanged `time bucket` as rowid(not entity id) to update the real record. This feature is hard to test in the e2e, as it would cross 2 days(a.k.a. 2 entities).

Basically, you need to read the event base on event UUID, and merge it first, to make sure, the start timestamp exists as time bucket.

kezhenxu94: Confirmed

oap-server/analyzer/event-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/event/listener/EventRecordAnalyzerListener.java
```diff
@@ -0,0 +1,81 @@
...
+        event.setTimeBucket(TimeBucket.getMinuteTimeBucket(e.getStartTime()));
```
wu-sheng, line 64: Do you have start time when reporting end time? I remember you and @BFergerson discussed this, this should be only optional. And from my understanding, it isn't just optional, it is immutable.

kezhenxu94: > Do you have start time when reporting end time? I remember you and @BFergerson discussed this, this should be only optional.

No, also that's not mandatory. It's optional when reporting an `endTime`.

> And from my understanding, it isn't just optional, it is immutable.

It's indeed immutable, we won't touch it when merging two records, only `endTime` and `message` (`parameters`) will be merged

wu-sheng: But what happens if someone reports an `end` event without `start` event`? So someone, the `start` event lost or blocked in some channel?

kezhenxu94: > But what happens if someone reports an end event without start event? Such as, the start event lost or blocked in some channel?

Do you think we should tackle the case where the `end` event comes before the `start` event, (i.e. the `start` and `end` are unordered), or simply abandon the `end` event if there is no `start` event already?

wu-sheng: Usually `end` should be later at least several seconds before `start`, but it may be very close. Such as rebooting could be quickly enough. 
The abandon mechanism is the thing we should discuss, where to do. In my mind, there are 2 places
1. Query before sending the metrics into the stream.
2. Adjust merge return value (`void`->`boolean`) to support `abandon` at the last minute. But this would cause wider changes

kezhenxu94: Method 2 seems to be more elegant. What about adjusting the signature of `combine` like this?

```java
    /**
     * Merge the given metrics instance, these two must be the same metrics type.
     *
     * @param metrics to be merged
     * @return {@code true} if the combined metrics should be continuously processed, {@code false} it should be abandoned.
     */
    public abstract boolean combine(Metrics metrics);
```

wu-sheng: Could you submit another PR to change that first? This change makes sense to me, but will affect a lot of codes, I hope the commit log is clear to everyone.

kezhenxu94: > Could you submit another PR to change that first? This change makes sense to me, but will affect a lot of codes, I hope the commit log is clear to everyone.

https://github.com/apache/skywalking/pull/6318

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/event/Event.java
```diff
@@ -0,0 +1,231 @@
...
+        if (StringUtil.isNotBlank(event.getType())) {
+            setType(event.getType());
+        }
+        if (StringUtil.isNotBlank(event.getMessage())) {
+            setType(event.getMessage());
+        }
+        if (StringUtil.isNotBlank(event.getParameters())) {
+            setParameters(event.getParameters());
+        }
+        return true;
```
wu-sheng, line 138: So, we still always return `true`, right? timebucket will be there always.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/event/Event.java
```diff
@@ -0,0 +1,231 @@
...
+    private String parameters;
+
+    @Column(columnName = START_TIME)
+    private long startTime;
+
+    @Column(columnName = END_TIME)
+    private long endTime;
+
+    @Override
+    public boolean combine(final Metrics metrics) {
```
wu-sheng, line 108: Let's update this protocol doc a little to map this logic, https://github.com/apache/skywalking-query-protocol/blob/master/event.graphqls#L26-L29

oap-server/analyzer/event-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/event/listener/EventRecordAnalyzerListener.java
```diff
@@ -0,0 +1,83 @@
...
+
+        event.setName(e.getName());
+        event.setType(e.getType().name());
+        event.setMessage(e.getMessage());
+        if (e.getParametersCount() > 0) {
+            event.setParameters(GSON.toJson(e.getParametersMap()));
+        }
+        event.setStartTime(e.getStartTime());
+        event.setEndTime(e.getEndTime());
+        event.setTimeBucket(TimeBucket.getMinuteTimeBucket(e.getStartTime()));
```
wu-sheng, line 66: I think you need some start time / end time / bucket check logic here. If this event(same UUID) reports at the first time, and only once, the `combine` wouldn't execute.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/event/Event.java
kezhenxu94: Yes

oap-server/analyzer/event-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/event/listener/EventRecordAnalyzerListener.java
kezhenxu94: done

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/event/Event.java
kezhenxu94: Submodule sync is also included in this PR

oap-server/analyzer/event-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/event/listener/EventRecordAnalyzerListener.java
```diff
@@ -0,0 +1,87 @@
...
+        if (e.getParametersCount() > 0) {
+            event.setParameters(GSON.toJson(e.getParametersMap()));
+        }
+        event.setStartTime(e.getStartTime());
+        event.setEndTime(e.getEndTime());
+        if (e.getStartTime() > 0) {
+            event.setTimeBucket(TimeBucket.getMinuteTimeBucket(e.getStartTime()));
+        } else if (e.getEndTime() > 0) {
+            event.setTimeBucket(TimeBucket.getMinuteTimeBucket(e.getEndTime()));
+        }
```
wu-sheng, line 70: With all of our long discussion about start time and end time, we may need to set up a set of validators to check the trace/metric/log/event data if some necessary fields are illegal or NULL.

Just FYI. If you agree, we could add a new issue to track this.

wu-sheng (approved):
LGTM.

oap-server/analyzer/event-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/event/listener/EventRecordAnalyzerListener.java
kezhenxu94: > With all of our long discussion about start time and end time, we may need to set up a set of validators to check the trace/metric/log/event data if some necessary fields are illegal or NULL.
> 
> 
> 
> Just FYI. If you agree, we could add a new issue to track this.

Agree

wu-sheng (approved):

PR #6187, opened 2021-01-12
Fix thread leakage caused by es client connection creation

YczYanchengzhe:
Since the es plugin creates a connection when acquiring the peer, but does not close it, this will cause thread leakage problems


wu-sheng:
2 things
1. Please fix the code format. Error from the logs.
> 2021-01-12T11:58:43.9816407Z [INFO] There is 1 error reported by Checkstyle 8.19 with /home/runner/work/skywalking/skywalking/apm-checkstyle/checkStyle.xml ruleset.
2021-01-12T11:58:43.9833897Z [ERROR] src/main/java/org/apache/skywalking/apm/plugin/elasticsearch/v6/interceptor/RestHighLevelClientConInterceptor.java:[35,31] (naming) ConstantName: Name 'logger' must match pattern '(^[A-Z][A-Z0-9]*(_[A-Z0-9]+)*$)'.

2. Please update `CHANGES.md` in the root folder.

YczYanchengzhe:
I fixed the problem you mentioned, what should I do next, do I check the ci again~

wu-sheng (approved):
Once plugin test passed, good to merge.

PR #6188, opened 2021-01-12
Make source builder set service name and endpoint name in right order

### LINKED ISSUE #6185: Endpoint name group doesn't work due to service name not set in SourceBuilder
### Please answer these questions before submitting your issue.

- [ ] Question or discussion
- [x] Bug
       When segment span reference list is not empty, for MultiScopesAnalysisListener.parseEntry method, the SourceBuilder sets endpoint name before service name is set which makes the endpoint name grouping doesn't work due to miss service name.
- [ ] Requirement

___
### Question
- What do you want to know?

___
### Bug
   SkyWalking 8.3.0, Linux, JDK 11
- Which company or project?

- What happened?
If possible, provide a way to reproduce the error. e.g. demo application, component version.

___
   Just set the service name before setting endpoint in source builder.

buxingzhe:
### Fix https://github.com/apache/skywalking/issues/6185
       `EndpointGroupingRule#format`  depends on service name to getting the rule, but we don't set service name in advance before setting endpoint name, which leads to fail to get the endpoint group rules of the service.
      The fix solution is the `SourceBuilder` class provides a method naming `prepare` to do the name formatting and endpoint grouping action, and delay the `prepare` operation in `MultiScopesAnalysisListener#build`.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6185 .


oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/MultiScopesAnalysisListener.java
```diff
@@ -310,6 +312,7 @@ public void build() {
         slowDatabaseAccesses.forEach(sourceReceiver::receive);
```
wu-sheng, line 312: I think you miss this one.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/config/NamingControl.java
```diff
@@ -95,6 +95,10 @@ public String formatInstanceName(String instanceName) {
      * @return the string, which length less than or equals {@link #endpointNameMaxLength};
      */
     public String formatEndpointName(String serviceName, String endpointName) {
+        if (endpointName == null) {
+            return null;
+        }
```
wu-sheng, line 100: I think you forget the null/empty check for service Name.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/MultiScopesAnalysisListener.java
buxingzhe: > I think you miss this one.

`slowDatabaseAccesses` is list of Metrics not SourceBuilders.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/config/NamingControl.java
buxingzhe: OK, I add the `service name` checking logic.

buxingzhe:
@wu-sheng I'd like to remove all the `setter` methods in `SourceBuilder` and replaced with Lombok `Setter` annotation, how do you think of it?

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/MultiScopesAnalysisListener.java
wu-sheng: It should be. This is another bug. DB service has over length risk too.

wu-sheng:
> @wu-sheng I'd like to remove all the `setter` methods in `SourceBuilder` and replaced with Lombok `Setter` annotation, how do you think of it?

Make sense. Go for it.

wu-sheng:
And resolve conflict please.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/MultiScopesAnalysisListener.java
buxingzhe: > It should be. This is another bug. DB service has over length risk too.
So what fields have such risk, `databaseServiceId` or `statement`?

wu-sheng: I think should be service id. We need to move service name and type inside this, then add the prepare method.

buxingzhe:
> And resolve conflict please.

I have merged the latest changes, but they are included in my current commit, it seems a little ugly.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/MultiScopesAnalysisListener.java
```diff
@@ -198,11 +198,9 @@ public void parseExit(SpanObject span, SegmentObject segmentObject) {
         if (sourceBuilder.getType().equals(RequestType.DATABASE)) {
```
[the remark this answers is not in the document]
buxingzhe: Resolved by changing to `if (RequestType.DATABASE.equals(sourceBuilder.getType())`

kezhenxu94:
> > And resolve conflict please.
> 
> I have merged the latest changes, but they are included in my current commit, it seems a little ugly.

It doesn't matter, we will squash the commits into a single one when merging into master

wu-sheng:
The CI seems still failing, could you fix them?

wu-sheng:
@buxingzhe The CI are failing, you need to read the logs and fix them.

buxingzhe:
> @buxingzhe The CI are failing, you need to read the logs and fix them.

Yeah, I have found a break due to changes to `DataSlowStatement`. Now I have changed to build `DatabaseSlowStatement` via `SourceBuilder` too.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/MultiScopesAnalysisListener.java
```diff
@@ -62,7 +61,7 @@
 public class MultiScopesAnalysisListener implements EntryAnalysisListener, ExitAnalysisListener, LocalAnalysisListener {
     private final List<SourceBuilder> entrySourceBuilders = new ArrayList<>(10);
     private final List<SourceBuilder> exitSourceBuilders = new ArrayList<>(10);
-    private final List<DatabaseSlowStatement> slowDatabaseAccesses = new ArrayList<>(10);
+    private final List<SourceBuilder> slowDatabaseAccessesBuilders = new ArrayList<>(10);
```
wu-sheng, line 64: Don't change this. My meaning is, `DatabaseSlowStatement` should have related name fields(used to have IDs only), and do `#prepare` too.

buxingzhe: @wu-sheng My question is any risk or breakout for my change? I think it' good to handle the sources in the same way.

wu-sheng: They are not the same thing, we created `DatabaseSlowStatement` intently, SourceBuilder basically relates to the real agent-instralled services.

buxingzhe: ok, I get your point.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/DatabaseSlowStatement.java
```diff
@@ -21,16 +21,26 @@
...
+import org.apache.skywalking.oap.server.core.analysis.IDManager;
+import org.apache.skywalking.oap.server.core.analysis.NodeType;
+import org.apache.skywalking.oap.server.core.config.NamingControl;
 
 import static org.apache.skywalking.oap.server.core.source.DefaultScopeDefine.DATABASE_SLOW_STATEMENT;
 
 @ScopeDeclaration(id = DATABASE_SLOW_STATEMENT, name = "DatabaseSlowStatement")
 public class DatabaseSlowStatement extends Source {
+    @Setter
+    private NamingControl namingControl;
```
wu-sheng, line 33: Take a look at `SourceBuilder`, how we initialize this there.

buxingzhe: Actually I'd like to do like `SourceBuilder`, but `DatabaseSlowStatement` must have the default construstor, just referring to `DispatcherManager.addIfAsSourceDispatcher` line#97.
Another way is I declare two constructors in `DatabaseSlowStatement`, i.e. the default and with `NamingControl`, what do you think of that?

wu-sheng: I see, how about adding a `DatabaseSlowStatementBuilder`?

buxingzhe: That's all right. Done!

wu-sheng (approved):
LGTM. Appreciate your patient.

PR #6190, opened 2021-01-12
Remove redundant metrics and polish panel layout

hanahmily:
As the title indicated, these changes fix some issues about istio control plane dashboard..

 - Remove redundant metrics from pilot errors.
 - Set correct metric type to pilot errors.
 - Reset the width of some panels to fill an entire row.


wu-sheng (approved):
Could you share the latest screenshot?

Issue #6127, opened 2021-01-04
Collect Log with java agent


EvanLjp:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
___
### discussion
Collect the java logs with the java agent.

There is a point in the feature, which it should have 2 gRPC channels to send data. one is the original gRPC channel to directly send trace data to the OAP server, another channel is the optional gRPC client to send the log to the Satellite.  
@elk-g

zifeihan:
There is a question, will it cause relatively large performance overhead or network bandwidth problems for business applications? Or we need to define the level of collecting logs on the agent?

wu-sheng:
@zifeihan @EvanLjp Could you take a look at existing log collection plugin, recently added.
We should enhance it rather than building a new one. Speaking about the channel, I prefer separate channel is an optional feature for the plugin. Because not all people are going to use satellite.

EvanLjp:
> There is a question, will it cause relatively large performance overhead or network bandwidth problems for business applications? Or we need to define the level of collecting logs on the agent?

1. The extension plugins is an appender plugin for log framework, such as Logback. The collected logs could be controlled by logback.xml by loggers or level. So the agent should not care about which logs should be collected. 
2. For the performance, In the future Satellite will support file reading, but for cloud disk mount, there will still be network consumption, so the first version prefers to use localhost network transfer to deliver the log data. I have no relevant experience in network performance testing, which I will learn about in the near future.

zifeihan:
> @zifeihan @EvanLjp Could you take a look at existing log collection plugin, recently added.
> We should enhance it rather than building a new one. Speaking about the channel, I prefer separate channel is an optional feature for the plugin. Because not all people are going to use satellite.

Okay, let me take a look first.

zifeihan:
> > There is a question, will it cause relatively large performance overhead or network bandwidth problems for business applications? Or we need to define the level of collecting logs on the agent?
> 
> 1. The extension plugins is an appender plugin for log framework, such as Logback. The collected logs could be controlled by logback.xml by loggers or level. So the agent should not care about which logs should be collected.
> 2. For the performance, In the future Satellite will support file reading, but for cloud disk mount, there will still be network consumption, so the first version prefers to use localhost network transfer to deliver the log data. I have no relevant experience in network performance testing, which I will learn about in the near future.

[...]

PR #6194, opened 2021-01-13
Collect Log with java agent (#6127)

elk-g:


apm-sniffer/config/agent.config
```diff
@@ -110,3 +110,10 @@ plugin.mount=${SW_MOUNT_FOLDERS:plugins,activations}
 
 # Match spring bean with regex expression for classname
 # plugin.springannotation.classname_match_regex=${SW_SPRINGANNOTATION_CLASSNAME_MATCH_REGEX:}
+
+#grpc-log client configuration
+grpclog.service_host=${SW_GRPC_LOG_SERVER_HOST:localhost}
+grpclog.service_port=${SW_GRPC_LOG_SERVER_PORT:8000}
+grpclog.max_message_size=${SW_GRPC_LOG_MAX_MESSAGE_SIZE:10485760}
+grpclog.log_body_type=${SW_GRPC_LOG_LOG_BODY_TYPE:2}
+grpclog.upstream_timeout=${SW_GRPC_LOG_GRPC_UPSTREAM_TIMEOUT:30}
```
wu-sheng, line 119: All these are optional, should move to the documentation.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -325,4 +325,32 @@
          */
         public static String AUTO_TAG_KEYS = "";
     }
+
+    public static class GRPCLog {
```
wu-sheng, line 329: This doesn't belong to the core, please move to the plugin.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/log/LogReportServiceClient.java
```diff
@@ -0,0 +1,161 @@
...
+import org.apache.skywalking.apm.network.logging.v3.LogData;
+import org.apache.skywalking.apm.network.logging.v3.LogReportServiceGrpc;
+
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
+import io.grpc.StatusRuntimeException;
+import io.grpc.stub.StreamObserver;
+
+@DefaultImplementor
+public class LogReportServiceClient implements BootService, IConsumer<LogData> {
```
wu-sheng, line 46: Log collecting service belongs to the plugin, please move.

wu-sheng (commented):
I can't see any update about the documentation. We need them.

EvanLjp:
@elk-g  I would help u  to finish this PR.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/log/LogReportServiceClient.java
EvanLjp: There is a core problem is gRPC lib would be shaded in the core. when having gRPC reference in optional plugins, the exception call `class not found ` would be occurring. @wu-sheng Do you have any good ideas?

wu-sheng: Do the shade and exclusion at the same time. It should be able to avoid this.

EvanLjp: It sounds great.

apm-application-toolkit/apm-toolkit-log4j-2.x/src/main/java/org/apache/skywalking/apm/toolkit/log/log4j/v2/x/log/GRPCLogClientAppender.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.apache.logging.log4j.core.Filter;
+import org.apache.logging.log4j.core.Layout;
+import org.apache.logging.log4j.core.LogEvent;
+import org.apache.logging.log4j.core.appender.AbstractAppender;
+
+public class GRPCLogClientAppender extends AbstractAppender {
+
+    protected GRPCLogClientAppender(String name, Filter filter, Layout<? extends Serializable> layout) {
+        super(name, filter, layout);
```
EvanLjp, line 31: why add it?

apm-application-toolkit/apm-toolkit-logback-1.x/src/main/java/org/apache/skywalking/apm/toolkit/log/logback/v1/x/log/GRPCLogClientAppender.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.toolkit.log.logback.v1.x.log;
+
+import ch.qos.logback.core.AppenderBase;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
```
EvanLjp, line 24: please remove it.

apm-sniffer/config/agent.config
```diff
@@ -110,3 +110,10 @@ plugin.mount=${SW_MOUNT_FOLDERS:plugins,activations}
 
 # Match spring bean with regex expression for classname
 # plugin.springannotation.classname_match_regex=${SW_SPRINGANNOTATION_CLASSNAME_MATCH_REGEX:}
+
+#grpc-log client configuration
+# plugin.grpclog.service_host=${SW_GRPC_LOG_SERVER_HOST:localhost}
+# plugin.grpclog.service_port=${SW_GRPC_LOG_SERVER_PORT:8000}
+# plugin.grpclog.max_message_size=${SW_GRPC_LOG_MAX_MESSAGE_SIZE:10485760}
+# plugin.grpclog.log_body_type=${SW_GRPC_LOG_LOG_BODY_TYPE:2}
+# plugin.grpclog.upstream_timeout=${SW_GRPC_LOG_GRPC_UPSTREAM_TIMEOUT:30}
```
wu-sheng, line 119: Please remove these from the default agent.config.

And you should update the toolkit doc about how to use the new appenders.

apm-sniffer/apm-toolkit-activation/apm-toolkit-log4j-1.x-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/log/log4j/v1/x/log/GRPCLogAppenderInterceptor.java
```diff
@@ -0,0 +1,114 @@
...
+    private LogData transform(LoggingEvent event) {
+        if (Objects.isNull(event)) {
+            return null;
+        }
+        LogData.Builder logBuilder = LogData.newBuilder()
+                .setTimestamp(event.getTimeStamp())
+                .setService(Config.Agent.SERVICE_NAME)
+                .setServiceInstance(Config.Agent.INSTANCE_NAME)
+                .setTraceContext(TraceContext.newBuilder()
+                        .setTraceId(ContextManager.getGlobalTraceId()).build())
```
wu-sheng, line 85: Why we don't inject segment id and span id?

apm-sniffer/apm-toolkit-activation/apm-toolkit-common/src/main/java/org/apache/skywalking/apm/toolkit/common/log/GRPCLogConfig.java
```diff
@@ -0,0 +1,54 @@
...
+
+            /**
+             * The max size of message to send to server.Default is 10 MB
+             */
+            public static int MAX_MESSAGE_SIZE = 10 * 1024 * 1024;
+
+            /**
+             * The log body type to send to the server, includes text(2),json(3),yaml(4).Default is text.
+             */
+            public static int LOG_BODY_TYPE = 2;
```
wu-sheng, line 46: Why use `int` rather than `Enum`?

apm-sniffer/apm-toolkit-activation/apm-toolkit-log4j-1.x-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/log/log4j/v1/x/log/GRPCLogAppenderInterceptor.java
```diff
@@ -0,0 +1,114 @@
...
+                .addTags(KeyStringValuePair.newBuilder()
+                        .setKey("level").setValue(event.getLevel().toString()).build())
+                .addTags(KeyStringValuePair.newBuilder()
+                        .setKey("logger").setValue(event.getLoggerName()).build())
+                .addTags(KeyStringValuePair.newBuilder()
+                        .setKey("thread").setValue(event.getThreadName()).build());
+
+        LogDataBody.Builder logDataBodyBuilder = LogDataBody.newBuilder();
+        LogDataBody.ContentCase contentType = Optional.ofNullable(LogDataBody.ContentCase.forNumber(
+                GRPCLogConfig.Plugin.GRPCLog.LOG_BODY_TYPE)).orElse(LogDataBody.ContentCase.TEXT);
```
wu-sheng, line 95: Doesn't log framework provide the content format? Right now, it seems based on manual configuration.

apm-sniffer/apm-toolkit-activation/apm-toolkit-log4j-1.x-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/log/log4j/v1/x/log/GRPCLogAppenderInterceptor.java
EvanLjp: waiting https://github.com/apache/skywalking/pull/6213

apm-sniffer/apm-toolkit-activation/apm-toolkit-log4j-1.x-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/log/log4j/v1/x/log/GRPCLogAppenderInterceptor.java
EvanLjp: The log framework is format the log content in console appended or file appended. In the current position, we got the real text rather than the formatted log.

elk-g: In the current position, we get the real text, and do we need to add an encoder in appender and send it with formatting log content to the client ?

wu-sheng:
You still have UT failure, please fix.

CHANGES.md
```diff
@@ -34,6 +34,7 @@ Release Notes.
 * Fix duplicated `EnhancedInstance` interface added.
 * Fix thread leaks caused by the elasticsearch-6.x-plugin plugin.
 * Support reading segmentId and spanId with toolkit.
+* Support collecting log by agent.
```
wu-sheng, line 37:
```suggestion
* Support collecting logs and forwarding through gRPC.
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/LogReportServiceClient.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.List;
+
+import org.apache.skywalking.apm.agent.core.boot.BootService;
+import org.apache.skywalking.apm.agent.core.boot.DefaultImplementor;
+import org.apache.skywalking.apm.commons.datacarrier.consumer.IConsumer;
+import org.apache.skywalking.apm.network.logging.v3.LogData;
+
+@DefaultImplementor
+public class LogReportServiceClient implements BootService, IConsumer<LogData> {
```
wu-sheng, line 29:
```suggestion
public class LogReportServiceClient implements BootService {
```
You don't need to implement this interface from my understanding.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/LogReportServiceClient.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    public void init() {
+
+    }
+
+    public void produce(LogData logData) {
+
+    }
```
wu-sheng, line 58: Why this?

EvanLjp:
image
@elk-g  please use `mvn clean package -Pall` to find the error units.

docs/en/concepts-and-designs/README.md
```diff
@@ -23,3 +23,5 @@ SkyWalking already support.
   - [Overview](ui-overview.md). A simple brief about SkyWalking UI.
 - CLI
   - [SkyWalking CLI](https://github.com/apache/skywalking-cli). A command line interface for SkyWalking.
+- Events
+  - [Events](event.md). Introduce the concept of "Event" in SkyWalking.
```
wu-sheng, line 27: I think this is a merging issue? You are not touching the `event` related things.

docs/en/concepts-and-designs/event.md
```diff
@@ -0,0 +1,56 @@
+# Events
```
wu-sheng, line 1: Same here. This is not your thing.

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -159,6 +159,10 @@ property key | Description | Default |
 `plugin.kafka.topic_profilings` | Specify which Kafka topic name for Thread Profiling snapshot to report to. | `skywalking_profilings` |
 `plugin.kafka.topic_management` | Specify which Kafka topic name for the register or heartbeat data of Service Instance to report to. | `skywalking_managements` |
 `plugin.springannotation.classname_match_regex` |  Match spring beans with regular expression for the class name. Multiple expressions could be separated by a comma. This only works when `Spring annotation plugin` has been activated. | `All the spring beans tagged with @Bean,@Service,@Dao, or @Repository.` |
+`plugin.grpclog.service_host` | Specify which grpc server's host for log data to report to. | `127.0.0.1` |
+`plugin.grpclog.service_port` | Specify which grpc server's port for traces data to report to. | `8000` |
+`plugin.grpclog.max_message_size` | Specify the maximum size of log data for grpc client to report to. | `10MB` |
+`plugin.grpclog.upstream_timeout` | How long grpc client will timeout in sending data to upstream. Unit is second.|`30` seconds|
```
wu-sheng, line 165: The config hierarchy seems doesn't make sense. You are putting reporter at level one, but nothing related toolkit. They should use `toolkit.log.grpc.reporter` for all these RPC related things.

apm-sniffer/apm-toolkit-activation/apm-toolkit-logback-1.x-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/log/logback/v1/x/log/GRPCLogAppenderInterceptor.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private LogData transform(ILoggingEvent event) {
+        return LogData.newBuilder()
+                .setTimestamp(event.getTimeStamp())
+                .setService(Config.Agent.SERVICE_NAME)
+                .setServiceInstance(Config.Agent.INSTANCE_NAME)
+                .setTraceContext(TraceContext.newBuilder()
+                        .setTraceId(ContextManager.getGlobalTraceId())
+                        .setSpanId(ContextManager.getSpanId())
+                        .setTraceSegmentId(ContextManager.getSegmentId())
```
wu-sheng, line 79: @EvanLjp As we are always sending this, should we still keep trace id as empty string, if there is no tracing context? What do you think?

apm-sniffer/apm-toolkit-activation/apm-toolkit-common/src/main/java/org/apache/skywalking/apm/toolkit/common/log/GRPCLogConfig.java
```diff
@@ -0,0 +1,49 @@
...
+             */
+            public static String SERVER_HOST = "127.0.0.1";
+
+            /**
+             * The port of gRPC log server.
+             */
+            public static int SERVER_PORT = 8000;
+
+            /**
+             * The max size of message to send to server.Default is 10 MB
```
EvanLjp, line 39: The max size of the message to send to the server. Default is 10 MB.

apm-sniffer/apm-toolkit-activation/apm-toolkit-common/src/main/java/org/apache/skywalking/apm/toolkit/common/log/GRPCLogConfig.java
```diff
@@ -0,0 +1,49 @@
...
+             */
+            public static int MAX_MESSAGE_SIZE = 10 * 1024 * 1024;
+
+            /**
+             * How long grpc client will timeout in sending data to upstream.
```
EvanLjp, line 44: please notice the timeunit.

docs/en/setup/service-agent/java-agent/Application-toolkit-log4j-1.x.md
```diff
@@ -18,3 +18,32 @@ log4j.appender.CONSOLE.layout.ConversionPattern=%d [%T] %-5p %c{1}:%L - %m%n
 ```
 
 * When you use `-javaagent` to active the sky-walking tracer, log4j will output **traceId**, if it existed. If the tracer is inactive, the output will be `TID: N/A`.
+
+# grpc log log4j plugin
+
+The grpc log plugin support report log to grpc Server.
```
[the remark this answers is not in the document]
wu-sheng: When you write documents, please keep in your mind, the reader will be like you when you first time read SkyWalking's documents. If you keep too many things in the dark, they will be very confused.

docs/en/setup/service-agent/java-agent/README.md
wu-sheng: There is another place of document should be updated

```md
    * If you want to print trace context(e.g. traceId) in your logs, choose the log frameworks, [log4j](Application-toolkit-log4j-1.x.md), 
[log4j2](Application-toolkit-log4j-2.x.md), [logback](Application-toolkit-logback-1.x.md)
```
Should change to 
> If you want to print trace context(e.g. traceId) in your logs, or collect logs, ...

apm-sniffer/apm-toolkit-activation/apm-toolkit-logback-1.x-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/log/logback/v1/x/log/GRPCLogAppenderInterceptor.java
EvanLjp: when no tracing context, the traceId maybe `N/A` rather than an empty string.

apm-sniffer/apm-toolkit-activation/apm-toolkit-common/src/main/java/org/apache/skywalking/apm/toolkit/common/log/GRPCLogReportServiceClient.java
```diff
@@ -0,0 +1,142 @@
...
+            carrier.shutdownConsumers();
+        } catch (Throwable t) {
+            LOGGER.error(t.getMessage(), t);
+        }
+    }
+
+    @Override
+    public void produce(LogData logData) {
+        if (Objects.nonNull(logData)) {
+            if (!carrier.produce(logData)) {
```
EvanLjp, line 87:
```
if (Objects.nonNull(logData)) &&  !carrier.produce(logData) {
    
                LOGGER.error("LogReportServiceClient discard logData: " + logData.toString());
            }
        }
```

apm-sniffer/apm-toolkit-activation/apm-toolkit-common/src/main/java/org/apache/skywalking/apm/toolkit/common/log/GRPCLogReportServiceClient.java
```diff
@@ -0,0 +1,142 @@
...
+                    waitStatus.finished();
+                }
+            });
+
+            for (final LogData logData : dataList) {
+                reportStreamObserver.onNext(logData);
+            }
+        } catch (Throwable e) {
+            if (!(e instanceof StatusRuntimeException)) {
+                LOGGER.error(e, "Report grpc log to server fail.");
```
EvanLjp, line 133: report log failure with the gRPC client.

apm-sniffer/apm-toolkit-activation/apm-toolkit-logback-1.x-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/log/logback/v1/x/log/GRPCLogAppenderInterceptor.java
wu-sheng: Do you want to send `N/A` always?

apm-sniffer/apm-toolkit-activation/apm-toolkit-common/src/main/java/org/apache/skywalking/apm/toolkit/common/log/GRPCLogReportServiceClient.java
wu-sheng: `!carrier.produce(logData)` is not a warning, it is only considered a warning.

apm-sniffer/config/agent.config
```diff
@@ -110,3 +110,10 @@ plugin.mount=${SW_MOUNT_FOLDERS:plugins,activations}
 
 # Match spring bean with regex expression for classname
 # plugin.springannotation.classname_match_regex=${SW_SPRINGANNOTATION_CLASSNAME_MATCH_REGEX:}
+
+#grpc-log client configuration
```
arugal, line 114:
```suggestion
# grpc-log client configuration
```

apm-sniffer/apm-toolkit-activation/apm-toolkit-common/pom.xml
```diff
@@ -0,0 +1,102 @@
...
+        <shade.io.grpc.target>${shade.package}.${shade.io.grpc.source}</shade.io.grpc.target>
+        <shade.io.netty.source>io.netty</shade.io.netty.source>
+        <shade.io.netty.target>${shade.package}.${shade.io.netty.source}</shade.io.netty.target>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.skywalking</groupId>
+            <artifactId>apm-network</artifactId>
+            <version>${project.version}</version>
```
arugal, line 43: Hi @elk-g [apm-agent-core](https://github.com/apache/skywalking/blob/9021e6791229e8dd81ffaf7b77cb9f567723d0b6/apm-sniffer/apm-agent-core/pom.xml#L58-L62) already dependency on `apm-network`, is it repeated? If I miss something, please let me know :)

apm-sniffer/apm-toolkit-activation/apm-toolkit-logback-1.x-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/log/logback/v1/x/log/GRPCLogAppenderInterceptor.java
EvanLjp: we can reduce the cost on the network, @arugal I think the oap server query mode is compatible with this way. Is that right?

apm-sniffer/apm-toolkit-activation/apm-toolkit-common/pom.xml
EvanLjp: @elk-g tell me that the shade plugin depends on the direct reference lib. @arugal

apm-sniffer/apm-toolkit-activation/apm-toolkit-logback-1.x-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/log/logback/v1/x/log/GRPCLogAppenderInterceptor.java
arugal: https://github.com/apache/skywalking/blob/9021e6791229e8dd81ffaf7b77cb9f567723d0b6/oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/RecordAnalysisListener.java#L85-L92

Now `N/A` is not filtered, I will submit a PR fix.

apm-sniffer/apm-toolkit-activation/apm-toolkit-common/pom.xml
arugal: https://github.com/elk-g/skywalking/blob/c6fc54a1e38adc77e5619135586644cd0f880f7a/apm-sniffer/apm-toolkit-activation/apm-toolkit-common/pom.xml#L49-L99

Is it here? I don't see the difference with [apm-agent-core](https://github.com/apache/skywalking/blob/9021e6791229e8dd81ffaf7b77cb9f567723d0b6/apm-sniffer/apm-agent-core/pom.xml#L171-L232).

apm-sniffer/apm-toolkit-activation/apm-toolkit-logback-1.x-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/log/logback/v1/x/log/GRPCLogAppenderInterceptor.java
wu-sheng: Don't fix for now. I am still feeling we should not send them.

wu-sheng: I am thinking, either core return empty even in tracing context, or the plugin filters this. After all, my key concern is network cost.

apm-sniffer/apm-toolkit-activation/apm-toolkit-common/pom.xml
elk-g: I tried to remove this dependency,  but it reported an incompatible error,  like `io.grpc.ManagedChannel` cannot convert to `org.apache.skywalking.apm.dependencies.io.grpc.Channel`. However when I add it, everything is ok. But I don't know the specific reason.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/LogReportServiceClient.java
elk-g: `GRPCLogReportServiceClient` needs to produce `LogData` to `DataCarrier` and consume it by sending it to the gRPC server. In the `apm-toolkit-activation` module, `GRPCLogClientAppender` produce `LogData` by `LogReportServiceClient`, so add this method in here and  implement the interface `IConsumer<LogData>`.

apm-sniffer/apm-toolkit-activation/apm-toolkit-common/pom.xml
arugal: @elk-g  I got it, maybe can move `GRPCLogReportServiceClient` to `apm-agent-core`, what do you think? @wu-sheng @EvanLjp

wu-sheng: It may be caused by the `core` including the shade. It is fine to depend this directly.

apm-sniffer/apm-toolkit-activation/apm-toolkit-common/pom.xml
```diff
@@ -0,0 +1,102 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>apm-toolkit-activation</artifactId>
+        <groupId>org.apache.skywalking</groupId>
+        <version>8.4.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>apm-toolkit-common</artifactId>
```
kezhenxu94, line 28: This name is too generic, consider renaming to `apm-toolkit-logging-common`

apm-sniffer/apm-toolkit-activation/apm-toolkit-common/src/main/java/org/apache/skywalking/apm/toolkit/common/log/GRPCLogReportServiceClient.java
```diff
@@ -0,0 +1,140 @@
...
+            carrier.shutdownConsumers();
+        } catch (Throwable t) {
+            LOGGER.error(t.getMessage(), t);
+        }
+    }
+
+    @Override
+    public void produce(LogData logData) {
+        if (Objects.nonNull(logData) && !carrier.produce(logData)) {
+                LOGGER.warn("LogReportServiceClient discard logData: " + logData.toString());
```
kezhenxu94, line 87: Format the indents

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -159,6 +159,10 @@ property key | Description | Default |
 `plugin.kafka.topic_profilings` | Specify which Kafka topic name for Thread Profiling snapshot to report to. | `skywalking_profilings` |
 `plugin.kafka.topic_management` | Specify which Kafka topic name for the register or heartbeat data of Service Instance to report to. | `skywalking_managements` |
 `plugin.springannotation.classname_match_regex` |  Match spring beans with regular expression for the class name. Multiple expressions could be separated by a comma. This only works when `Spring annotation plugin` has been activated. | `All the spring beans tagged with @Bean,@Service,@Dao, or @Repository.` |
+`plugin.grpclog.service_host` | Specify which grpc server's host for log data to report to. | `127.0.0.1` |
+`plugin.grpclog.service_port` | Specify which grpc server's port for traces data to report to. | `8000` |
+`plugin.grpclog.max_message_size` | Specify the maximum size of log data for grpc client to report to. | `10MB` |
```
kezhenxu94, line 164: The default value `10MB` misleads users that they can set it to something like `20MB`, please use `10485760`

apm-sniffer/apm-toolkit-activation/apm-toolkit-common/src/main/java/org/apache/skywalking/apm/toolkit/common/log/GRPCLogReportServiceClient.java
wu-sheng: I think print all discarded logs would be very dangerous, which could impact the system performance a lot. Please take a look how trace sender did. Don't print the log content.

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -159,6 +159,10 @@ property key | Description | Default |
 `plugin.kafka.topic_profilings` | Specify which Kafka topic name for Thread Profiling snapshot to report to. | `skywalking_profilings` |
 `plugin.kafka.topic_management` | Specify which Kafka topic name for the register or heartbeat data of Service Instance to report to. | `skywalking_managements` |
 `plugin.springannotation.classname_match_regex` |  Match spring beans with regular expression for the class name. Multiple expressions could be separated by a comma. This only works when `Spring annotation plugin` has been activated. | `All the spring beans tagged with @Bean,@Service,@Dao, or @Repository.` |
+`plugin.grpclog.service_host` | Specify which grpc server's host for log data to report to. | `127.0.0.1` |
+`plugin.grpclog.service_port` | Specify which grpc server's port for traces data to report to. | `8000` |
```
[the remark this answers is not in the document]
wu-sheng: Why port is `8000`? @elk-g I think the default should be `11800`.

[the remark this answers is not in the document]
kezhenxu94: > Why port is `8000`? @elk-g I think the default should be `11800`.

Seems this port is of https://github.com/apache/skywalking-satellite

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/LogReportServiceClient.java
wu-sheng: I know you need to. My question is, why need this in core default empty service?

docs/en/setup/service-agent/java-agent/README.md
wu-sheng: Above are still missing. @elk-g

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -209,6 +213,7 @@ The tag is, key=`x-le` and value = `{"logic-span":true}`.
     * If you require customized metrics, try [SkyWalking Meter System Toolkit](Application-toolkit-meter.md).
     * If you want to print trace context(e.g. traceId) in your logs, choose the log frameworks, [log4j](Application-toolkit-log4j-1.x.md), 
 [log4j2](Application-toolkit-log4j-2.x.md), [logback](Application-toolkit-logback-1.x.md)
+    * If you want to report your logs by gRPC, choose the log frameworks, [log4j](Application-toolkit-log4j-1.x.md), [log4j2](Application-toolkit-log4j-2.x.md), [logback](Application-toolkit-logback-1.x.md)
```
wu-sheng, line 216: If you want to separate this, please separate the detail makrdown. It will be strange that both are pointing to `Application-toolkit-log4j-1.x.md`, etc.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/LogReportServiceClient.java
elk-g: @wu-sheng  In toolkit module, only LogReportServiceClient can be obtained by `ServiceManager.INSTANCE.findService()`. There is a need for class LogReportServiceClient to provide a method for producing data, so I add this default empty service. Is there something wrong with the way I use it?

wu-sheng: My thought is, you should be able to `findService` and cast to your `override` implementation.

elk-g: I tried it by using `findService(LogReportServiceClient.class)` to obtain the client and then cast it to `GRPCLogReportServiceClient`, but  it occurs an error  like `GRPCLogReportServiceClient cannot be cast to GRPCLogReportServiceClient`. The way doesn't seem to work.

wu-sheng:
> 2021-01-17T09:08:19.9531484Z [ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.1.1:shade (default) on project apm-toolkit-logging-common: Error creating shaded jar: duplicate entry: META-INF/services/org.apache.skywalking.apm.dependencies.io.grpc.ServerProvider -> [Help 1]

Seems you are facing package issue?

elk-g:
> > 2021-01-17T09:08:19.9531484Z [ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.1.1:shade (default) on project apm-toolkit-logging-common: Error creating shaded jar: duplicate entry: META-INF/services/org.apache.skywalking.apm.dependencies.io.grpc.ServerProvider -> [Help 1]
> 
> Seems you are facing package issue?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/LogReportServiceClient.java
wu-sheng: I mean remove the interface, your previous error is caused by `LogReportServiceClient` doesn't exist.

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -159,6 +159,10 @@ property key | Description | Default |
 `plugin.kafka.topic_profilings` | Specify which Kafka topic name for Thread Profiling snapshot to report to. | `skywalking_profilings` |
 `plugin.kafka.topic_management` | Specify which Kafka topic name for the register or heartbeat data of Service Instance to report to. | `skywalking_managements` |
 `plugin.springannotation.classname_match_regex` |  Match spring beans with regular expression for the class name. Multiple expressions could be separated by a comma. This only works when `Spring annotation plugin` has been activated. | `All the spring beans tagged with @Bean,@Service,@Dao, or @Repository.` |
+`plugin.grpclog.service_host` | Specify which grpc server's host for log data to report to. | `127.0.0.1` |
+`plugin.grpclog.service_port` | Specify which grpc server's port for log data to report to. | `11800` |
+`plugin.grpclog.max_message_size` | Specify the maximum size of log data for grpc client to report to. | `10485760` |
+`plugin.grpclog.upstream_timeout` | How long grpc client will timeout in sending data to upstream. Unit is second.|`30` seconds|
```
wu-sheng, line 165: Should be `toolkit.log.grpc.reporter.*`

elk-g:
> > 2021-01-17T09:08:19.9531484Z [ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.1.1:shade (default) on project apm-toolkit-logging-common: Error creating shaded jar: duplicate entry: META-INF/services/org.apache.skywalking.apm.dependencies.io.grpc.ServerProvider -> [Help 1]
> 
> Seems you are facing package issue?

Yes, there are some problems in using maven-shade-plugin in toolkit. When I add maven-shade-plugin in toolkit module,  it will occur `Error creating shaded jar: duplicate entry`. But  when I remove this plugin,  other checks will fail and occur a compilation problem.  How should I use the maven-shade-plugin in toolkit module?

apm-sniffer/apm-toolkit-activation/apm-toolkit-logging-common/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+                                    <artifact>com.google.protobuf:protobuf-java</artifact>
+                                    <excludes>
+                                        <exclude>google/protobuf/*.proto</exclude>
+                                        <exclude>google/protobuf/compiler/*.proto</exclude>
+                                    </excludes>
+                                </filter>
+                            </filters>
+                            <transformers>
+                                <transformer
+                                        implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
```
wu-sheng, line 91: Do you choose the right transformer?

wu-sheng: I think you don't have service resource to transfer?

Try this? https://maven.apache.org/plugins/maven-shade-plugin/examples/resource-transformers.html

apm-sniffer/apm-toolkit-activation/apm-toolkit-logback-1.x-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/log/logback/v1/x/log/GRPCLogAppenderInterceptor.java
wu-sheng: @EvanLjp What do you prefer? Filter `N/A` here?

EvanLjp: @wu-sheng  I think the trace context should be removed, but the log content still should be sent.

wu-sheng: It is fine to send the logs, we at least have the service and service instance name.

wu-sheng: I mean the segment id and span id only.

docs/en/setup/service-agent/java-agent/README.md
elk-g: Using `@PluginConfig` to set config, isn't the prefix of key  fixed to plugin plus classname?

wu-sheng: Take a look at this, https://github.com/apache/skywalking/pull/6194#discussion_r558848862. I mean, that should be changed.

elk-g: I'm so sorry to miss it. Do you means using `Toolkit` as the class name and adding `log_grpc_reporter` as the prefix of the field, so the key willi like plugin.toolkit.log_grpc_reporter*? Or using a long class name? Or something else？

wu-sheng: `toolkit.log.grpc.reporter` means you have deeper nested static class. Not a long class name. 
The nested inner class will be transformed as the key automatically by the agent core.

apm-sniffer/apm-toolkit-activation/apm-toolkit-logback-1.x-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/log/logback/v1/x/log/GRPCLogAppenderInterceptor.java
wu-sheng: @EvanLjp @elk-g Let's finalize this, others are good now. There are `N/A`, `Ignored_Trace`, ``(EMPTY_STRING`) as trace id and segment id.  -1 as span id, which all means no available trace context. Could we filter these out? At least base on `trace_id`?

apm-sniffer/apm-toolkit-activation/apm-toolkit-logging-common/src/main/java/org/apache/skywalking/apm/toolkit/logging/common/log/GRPCLogReportServiceClient.java
```diff
@@ -0,0 +1,143 @@
...
+                .withMaxOutboundMessageSize(ToolkitConfig.Plugin.Toolkit.Log.GRPC.Reporter.MAX_MESSAGE_SIZE);
+    }
+
+    @Override
+    public void shutdown() {
+        try {
+            if (channel != null) {
+                channel.shutdownNow();
+            }
+            carrier.shutdownConsumers();
```
kezhenxu94, line 78: Should shut down consumers first

```suggestion
            carrier.shutdownConsumers();
            if (channel != null) {
                channel.shutdownNow();
            }
```

apm-sniffer/apm-toolkit-activation/apm-toolkit-logging-common/src/main/java/org/apache/skywalking/apm/toolkit/logging/common/log/GRPCLogReportServiceClient.java
```diff
@@ -0,0 +1,143 @@
...
+                public void onNext(Commands commands) {
+                }
+
+                @Override
+                public void onError(Throwable t) {
+                    waitStatus.finished();
+                    if (!disconnected.get()) {
+                        LOGGER.error("Send log to gRPC server fail with an internal exception.", t);
+                        disconnected.set(true);
+                    }
```
kezhenxu94, line 114: `disconnected` may be unnecessary, but if it's really needed, it should be

```suggestion
                    if (disconnected.compareAndSet(false, true)) {
                        LOGGER.error("Send log to gRPC server fail with an internal exception.", t);
                    }
```

apm-sniffer/apm-toolkit-activation/apm-toolkit-logging-common/src/main/java/org/apache/skywalking/apm/toolkit/logging/common/log/GRPCLogReportServiceClient.java
```diff
@@ -0,0 +1,143 @@
...
+
+                    LOGGER.error(t, "Try to send {} log data to collector, with unexpected exception.",
+                            dataList.size());
+                }
+
+                @Override
+                public void onCompleted() {
+                    if (disconnected.get()) {
+                        disconnected.set(false);
+                    }
```
kezhenxu94, line 124: ditto

```suggestion
                    disconnected.compareAndSet(true, false)
```

wu-sheng (dismissed):
LGTM. @kezhenxu94 @arugal Could you recheck and confirm too?

arugal (approved):
LGTM

wu-sheng:
Let's merge #6214 first, it waits for a little long.

apm-sniffer/apm-toolkit-activation/apm-toolkit-logging-common/src/main/resources/META-INF/services/org.apache.skywalking.apm.agent.core.boot.BootService
```diff
@@ -0,0 +1,22 @@
+#
+# Copyright (C) 2021 Baidu, Inc. All Rights Reserved.
```
EvanLjp, line 2: please remove it

wu-sheng: @kezhenxu94 This is a hit, why the SkyWalking eye doesn't detect this?

kezhenxu94: > @kezhenxu94 This is a hit, why the SkyWalking eye doesn't detect this?
 
The Apache license header is there correctly, but "Copyright (C) 2021 Baidu, Inc. All Rights Reserved." is an extra line, skywalking-eyes doesn't take this into consideration for now

kezhenxu94 (approved):

PR #6200, opened 2021-01-13
chore: set up license-eye to check license header

kezhenxu94:

### DIFF
--- a/.github/actions/e2e-test/action.yml
+++ b/.github/actions/e2e-test/action.yml
@@ -1,0 +1,66 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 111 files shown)


.github/actions/e2e-test/action.yml
```diff
@@ -0,0 +1,65 @@
...
+        rm -rf skywalking-eyes
+        docker run --rm -v $(pwd):/github/workspace license-eye:latest header check
+        echo "::endgroup::"
+    - name: Build Docker Image
+      shell: bash
+      env:
+        SKIP_TEST: "true"
+      run: |
+        echo "::group::Build Docker Image"
+        make docker && ES_VERSION=es7 TAG=latest-es7 make docker.oap
```
wu-sheng, line 44: Are we always using ES7?

wu-sheng (commented):
Could we keep `Check License Header` in one task only? It seems repeatedly check the headers.

kezhenxu94:
> Could we keep `Check License Header` in one task only? It seems repeatedly check the headers.

This is intentional, it serves as a barrier to fail fast when there is file missing license header.

.github/actions/e2e-test/action.yml
kezhenxu94: > Are we always using ES7?

Not necessary, if we already build a ES6 image, it costs only very little time to build ES7 too, so I wanted to make it simple

wu-sheng (commented):
`skywalking-ui` submodule has been updated, please recheck. If this is your plan, please update the changes file.

kezhenxu94:
> `skywalking-ui` submodule has been updated, please recheck. If this is your plan, please update the changes file.

Updated

wu-sheng (approved):

PR #6202, opened 2021-01-13
added an optional plugin which control sampling by specifying a "percentage" instead of "agent.sample_n_per_3_secs"

### LINKED ISSUE #5997: agent side percentage sampling config option

- [X] Question or discussion
- [ ] Bug
- [ ] Requirement

___
### Question
- What do you want to know?
For sampling, it seems currently (v8.2.0/v8.3.0) there are primarily 2 ways to control trace sampling. One is in server side using `sampleRate` but the bandwidth is still wasted and may not be very flexible if has many multiple sources of agent. The other is via agent side `agent.sample_n_per_3_secs` option but the use case is kind of counter-intuitive. In many scenarios one may want a percentage sampling rate option in agent side. What's the reason of not providing that but using a "traces in 3 seconds" mechanism instead? This looks a bit odd and from looking at the `SamplingService` code a percentage way would not be difficult to implement.   
___
### Bug
v8.2.0
openjdk version "1.8.0_265"

___
- provide an agent config similar to `agent.sample_n_per_3_secs` but in percentage (like 10 in every 10000). We can try to implement this and open a PR if that's OK.

Frefreak:
This trivial plugin provides a `plugin.sampling.sample_rate` config to control sampling rate (n out of 10000). It overrides default `SamplingService`. Our team is using it to work around what had been discussed in #5997 and thought this might be useful to some others.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #5997.

### DIFF
--- a/apm-sniffer/optional-plugins/percentage-sampling-plugin/README.md
+++ b/apm-sniffer/optional-plugins/percentage-sampling-plugin/README.md
@@ -1,0 +1,12 @@
+# Overview
+
+This plugin adds another sampling mechanism which determines if it should "sample"
+according to a specified rate parameter instead of the original `agent.sample_n_per_3_secs`.
+
+# Usage
+
+1. To enable this plugin, drop the apm-percentage-sampling-plugin-<version>.jar into plugins folder
+2. Add a config to agentArgs using key `plugin.sampling.sample_rate` to sample n out of 10000.
+For example `plugin.sampling.sample_rate = 100` means 1% (100 out of 10000). Do note that:
+    - because PRNG is involved, the actual result will not be precisely the specified amount.
+    - just like other plugins, config can be given using system properties and environment variables.
--- a/apm-sniffer/optional-plugins/percentage-sampling-plugin/pom.xml
+++ b/apm-sniffer/optional-plugins/percentage-sampling-plugin/pom.xml
@@ -1,0 +1,30 @@
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
+  ~ contributor license agreements.  See the NOTICE file distributed with
+  ~ this work for additional information regarding copyright ownership.
+  ~ The ASF licenses this file to You under the Apache License, Version 2.0
+  ~ (the "License"); you may not use this file except in compliance with
+  ~ the License.  You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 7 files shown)


wu-sheng:
I am confused, during the discussion, I said https://github.com/apache/skywalking/issues/5997#issuecomment-743729015. I can't see the question had been answered and provided reasonable explanation.

Frefreak:
For example, imagine in a scenario where there is an operation's QPS changes during the day, say 10000 in the evening and 100 in the morning. If I want a portion of all the samples to be sampled, a fixed `sample_n_per_3_secs` would not work very well because it's kind of hard to determine what value should be used for it. If I set it relative to QPS = 100, too less samples would be sampled when QPS = 10000. If I set it relative to QPS = 10000, all the traces in other time might be sampled. Besides I need know the current QPS in order to set it right. In this case if one can just specify a number indicating the percentage rate, it would be way easier to achieve this.

Also we do understand that the sample rate would not be accurate, that's expected. This plugin doesn't intend to be accurate either.
In summary this plugin just provides another way to specify how often one would like traces to be sampled, please consider its use cases. Not all users need this of course, but maybe some users would consider this useful.

wu-sheng:
> If I set it relative to QPS = 100, too less samples would be sampled when QPS = 10000. If I set it relative to QPS = 10000, all the traces in other time might be sampled.

Personally, I can't see the real value of this. Because of the inaccurate issue you have mentioned, the agent-side sampling only makes sense when you are facing high-traffic-load. Which percent-sampling mechanism doesn't really help. When the traffic goes to over 10+ times than it usually is, sampling rate keeps unchanged, and much more data has been sampled.

pierre94:
> Personally, I can't see the real value of this. Because of the inaccurate issue you have mentioned, the agent-side sampling only makes sense when you are facing high-traffic-load. Which percent-sampling mechanism doesn't really help. When the traffic goes to over 10+ times than it usually is, sampling rate keeps unchanged, and much more data has been sampled.

I think sampling by specifying a "percentage" can be a optional configuration for users. 

Jaeger provides a similar configuration[`Probabilistic`](https://www.jaegertracing.io/docs/1.21/sampling/#client-sampling-configuration)

wu-sheng:
`Jaeger` option belongs to them. We are asking for use case, rather than referencing. SkyWalking should always focus on the use case, what benefits users could have, rather than what configurations other projects have.

wu-sheng:
We are moving fast on Agent CDS(Config Discovery Service) https://github.com/apache/skywalking/issues/6114.
With that, it also reduces the requirement of using percentage mechanism to do sampling. Actually, you could do automatical sampling easily through that.

Frefreak:
Yes, dynamic configuration will do. We're recently also experimenting with that but official support would be optimal. Looking forward to that. Thanks!

PR #6211, opened 2021-01-14
Fix memory leak that tag may cause

YczYanchengzhe:

I think it is meaningless for tags to have multiple tags with the same key and value, so I suggest changing the tag type to set.
Problem scenario :  If there is a thread that has always existed, the method called by the thread is intercepted by us. When creating a tag, it will cause the risk of memory leaks.


wu-sheng:
First I think this is not a leak. And have you run the benchmark between ArrayList and HashSet?

YczYanchengzhe:
The problem we encountered is this: we developed a custom plug-in to intercept a certain method, in which we set a tag, which is not overridable, and we have enabled cross-thread interception, and there is a thread pool for the business We kept calling our intercepted method. After running for a period of time, there was a memory overflow, and the same tags in the memory. I think this is a problem that may bring security risks. What do you mean by the benchmark test? I don’t know much about this~ what do I need to do

YczYanchengzhe:
Do you mean the performance impact of comparing arraylist and hashset in high concurrency scenarios?

wu-sheng:
> in which we set a tag, which is not overridable

This mostly likely is caused by your tag declaration. If you want override supported, use this constructor to create the tag, and give it a unique `id`. See `status_code` tag.
```java
    public StringTag(int id, String tagKey, boolean canOverWrite) {
        super(id, tagKey, canOverWrite);
    }
```

PR #6214, opened 2021-01-15
fix recording url tag with wrong port

ralphgj:
The default port is -1 in `URI`, now use empty string instead of the default value.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,7 @@ Release Notes.
 * Fix duplicated `EnhancedInstance` interface added.
 * Fix thread leaks caused by the elasticsearch-6.x-plugin plugin.
 * Support reading segmentId and spanId with toolkit.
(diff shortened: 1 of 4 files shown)


wu-sheng (approved):
Confirmed through `java.net.URL` source code. Good catch.

ralphgj:
sorry, I had a mistake and recommit now. please review again, thx.

wu-sheng:
> sorry, I had a mistake and recommit now. please review again, thx.

You just did a force push, I can't see what you changed. Could you be clear about that?

wu-sheng (approved):

wu-sheng (changes requested):
Please check the plugin test logs. Your change breaks the tests.

wu-sheng (approved):
When all plugin tests passed, this should be good.

PR #6215, opened 2021-01-16
Set up test for Istio 1.8.2 to verify Envoy V3 protocol

kezhenxu94:

### DIFF
--- a/.github/workflows/e2e.istio.yaml
+++ b/.github/workflows/e2e.istio.yaml
@@ -27,7 +27,6 @@ on:
 env:
   SKIP_TEST: true
   ES_VERSION: es7
-  ISTIO_VERSION: 1.7.1
   TAG: ${{ github.sha }}
   SCRIPTS_DIR: test/e2e-mesh/e2e-istio/scripts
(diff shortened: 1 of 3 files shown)


kezhenxu94:
Verified via https://github.com/kezhenxu94/skywalking/actions/runs/489408211

.github/workflows/e2e.istio.yaml
```diff
@@ -134,7 +136,13 @@ jobs:
   metrics-service:
     runs-on: ubuntu-16.04
     timeout-minutes: 60
-    name: MetricsService
+    name: Istio(${{ matrix.istio_version }})+MetricsService
```
wu-sheng, line 139: This includes meta exchange,  right?

kezhenxu94: This is "Metrics Service", Metrics Service has nothing to do with Meta Exchange. Check line 171 "# disable the metadata-exchange extension intentionally to make sure metrics service doesn't rely on it".

As for Meta Exchange verification, it's in line 40-42

wu-sheng (approved):

PR #6216, opened 2021-01-16
change spanId toolkit api type

EvanLjp:
change spanId toolkit api type

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManager.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManager.java
@@ -88,7 +88,7 @@ public class ContextManager implements BootService {
     /**
      * @return the current span id if needEnhance. Otherwise, the value is -1.
      */
-    public static Integer getSpanId() {
+    public static int getSpanId() {
         AbstractTracerContext context = CONTEXT.get();
         return Objects.nonNull(context) ? context.getSpanId() : -1;
     }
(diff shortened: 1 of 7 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManager.java
```diff
@@ -70,27 +70,27 @@ private static AbstractTracerContext get() {
...
         AbstractTracerContext context = CONTEXT.get();
         return Objects.nonNull(context) ? context.getSegmentId() : EMPTY_TRACE_CONTEXT_ID;
     }
 
     /**
-     * @return the current span id if needEnhance. Otherwise, "N/A".
+     * @return the current span id when tracing. Otherwise, the value is -1.
      */
-    public static String getSpanId() {
+    public static Integer getSpanId() {
```
wu-sheng, line 91: The integer makes me feel, if not tracing, should null?
-1 makes sense if the return type is int.

EvanLjp: I struggle with that too, choose -1 to avoid null pointer at last. I would change the type to int

wu-sheng: It is fine to use -1 and int as return.

wu-sheng (approved):

kezhenxu94 (approved):

PR #6239, opened 2021-01-21
add pulsar reporter & fetcher

pierre94:
### <Feature description>


wu-sheng:
Hi, could you share why add this to the upstream?

pierre94:
I want to provide optional pulsar reporter and fetcher  to adapt to different data infrastructure.

wu-sheng:
@pierre94 Yes, I understand Pulsar could be the transport layer. My question is more that, whether should this be included in the main repo.

FYI @apache/skywalking-committers , especially @hanahmily @dmsolr @kezhenxu94 @arugal @EvanLjp  What do you think about this policy? The reporter plugin is much different with the other framework plugin, which hosts very little codes(less than 10 files usually). But as the reporter, we have to shade and re-distribute the whole dependencies. I think we should consider hosting codes in other places(including new repo) in or out of ASF.

EvanLjp:
> @pierre94 Yes, I understand Pulsar could be the transport layer. My question is more that, whether should this be included in the main repo.
> 
> FYI @apache/skywalking-committers , especially @hanahmily @dmsolr @kezhenxu94 @arugal @EvanLjp What do you think about this policy? The reporter plugin is much different with the other framework plugin, which hosts very little codes(less than 10 files usually). But as the reporter, we have to shade and re-distribute the whole dependencies. I think we should consider hosting codes in other places(including new repo) in or out of ASF.
Sounds Great,  If this is possible, logging-agent-receiver would not cost so much time to solve the shade problems.

wu-sheng:
We could consider to have a repo called SkyWalking-Transporter(in ASF) or Transporter-For-SkyWalking(in SkyAPM).

wu-sheng:
> > @pierre94 Yes, I understand Pulsar could be the transport layer. My question is more that, whether should this be included in the main repo.
> > FYI @apache/skywalking-committers , especially @hanahmily @dmsolr @kezhenxu94 @arugal @EvanLjp What do you think about this policy? The reporter plugin is much different with the other framework plugin, which hosts very little codes(less than 10 files usually). But as the reporter, we have to shade and re-distribute the whole dependencies. I think we should consider hosting codes in other places(including new repo) in or out of ASF.
> 
> Sounds Great, If this is possible, logging-agent-receiver would not cost so much time to solve the shade problems.

I am afraid that will not change, due to the classloader issue keeps anyway. No matter what we do.

EvanLjp:
> I am afraid that will not change, due to the classloader issue keeps anyway. No matter what we do.

that's right, but we can avoid doing repeat shade operations in core or plugins. ^-^

wu-sheng:
@pierre94 Do you want to work with us to move this to SkyAPM org, and only host the doc link in the Apache repo?

dmsolr:
For now, placement in SkyAPM Repo will be a good choice. Once have many people need this, we are easy to take it back.
We notice that the jar of Pulsar client is so big. Which more than 19m without dependencies.

pierre94:
> @pierre94 Do you want to work with us to move this to SkyAPM org, and only host the doc link in the Apache repo?

This is a good solution. Let's do it !

wu-sheng:
@dmsolr @pierre94 Repo created, https://github.com/SkyAPM/skywalking-transporter-plugin.

@dmsolr Could you lead on this project? Setting up the maven project structure and CI etc. Maybe later we need e2e too.

dmsolr:
> @dmsolr @pierre94 Repo created, https://github.com/SkyAPM/skywalking-transporter-plugin.
> 
> @dmsolr Could you lead on this project? Setting up the maven project structure and CI etc. Maybe later we need e2e too.

OK.

dmsolr:
Hi @pierre94 
The sub repo got ready (CI environment). Please move it to the new repo.

wu-sheng:
FYI @sijie, I think this is what you asked for.

PR #6243, opened 2021-01-21
Support the identification of the source host of the metric data in otel-receiver-plugin

wankai123:
### Add node information from the Openensus proto to the labels of the samples, to support the identification of the source host of the metric data

When I used the otel-receiver-plugin to receive prometheus node_exporter data forwarded from the otel-collector, 
I found that there is no label of the data source host in the samples, so I can't use MAL to do aggregation analysis by host,
unless I configure additional labels in otel-collector config.yaml, But then the label will be on each metrics.

In <opencensus.proto.agent.common.v1.Node>, node information is already carried, included  the host name.

So I add a labelkey 'node_identifier_host_name' to the collected data samples, the labelvalue is from 
<Node.identifier.host_name>, to support the identification of the source host of the metric data.


docs/en/setup/backend/backend-receivers.md
```diff
@@ -182,7 +182,9 @@ receiver-otel:
     enabledHandlers: ${SW_OTEL_RECEIVER_ENABLED_HANDLERS:"oc"}
     enabledOcRules: ${SW_OTEL_RECEIVER_ENABLED_OC_RULES:"istio-controlplane"}
 ```
-
+The receiver will add a labelkey 'node_identifier_host_name' to the collected data samples，
+the labelvalue is from opencensus.proto.agent.common.v1.Node.identifier.host_name, 
+to support the identification of the source of the metric data.
```
wu-sheng, line 187:
```suggestion
The receiver adds a label with `key=node_identifier_host_name` to the collected data samples，
and value from `Node.identifier.host_name` defined in opencensus agent proto, 
to be the identification of the metric data.
```

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/oc/OCMetricHandler.java
```diff
@@ -107,6 +111,15 @@
         return result;
     }
 
+    private static Map<String, String> buildLabelsWithNodeInfo(Node node,
```
wu-sheng, line 114:
```suggestion
    private static Map<String, String> buildLabelsFromNodeInfo(Node node,
```

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/oc/OCMetricHandler.java
```diff
@@ -107,6 +111,15 @@
...
+                                                               List<LabelKey> keys,
+                                                               List<LabelValue> values) {
+        Map<String, String> result = buildLabels(keys, values);
```
wu-sheng, line 117: Take a look at this comment in the original proto
```
  // This is required only in the first message on the stream or if the
  // previous sent ExportMetricsServiceRequest message has a different Node (e.g.
  // when the same RPC is used to send Metrics from multiple Applications).
  opencensus.proto.agent.common.v1.Node node = 1;
```

You need to hold the previous node, if the current `request#hasNode()` is false, use the previous one.

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/oc/OCMetricHandler.java
wu-sheng, line 114: Please don't hide `buildLabels` in the `buildLabelsWithNodeInfo`, it is misleading the reader. Do like this `buildLabelsFromNodeInfo(buildLables(...))`.

wu-sheng:
Next time, please use a pure new branch to send the pull request. Otherwise, it always brings the previous commit log.

docs/en/setup/backend/backend-receivers.md
wankai123: done

wankai123:
done

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/oc/OCMetricHandler.java
```diff
@@ -57,11 +58,21 @@
     @Override public StreamObserver<ExportMetricsServiceRequest> export(
         StreamObserver<ExportMetricsServiceResponse> responseObserver) {
         return new StreamObserver<ExportMetricsServiceRequest>() {
-            @Override public void onNext(ExportMetricsServiceRequest request) {
+            private Node node;
+
+            @Override
+            public void onNext(ExportMetricsServiceRequest request) {
+                if (request.hasNode())
+                    node = request.getNode();
```
wu-sheng, line 66: I am not sure whether without `{}` is allowed. We will see.

hanahmily: In case of NPE, I prefer to assign a default value to the `Node`, even node is sent in the first message.

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/oc/OCMetricHandler.java
```diff
@@ -107,6 +118,13 @@
         return result;
     }
 
+    private static Map<String, String> buildLabelsFromNodeInfo(Node node,
+                                                               Map<String, String> buildLabelsResult) {
+        if (node != null)
+            buildLabelsResult.put("node_identifier_host_name", node.getIdentifier().getHostName());
```
hanahmily, line 124: Could we add other fields more than `hostname` in `ProcessIdenifier`?

wu-sheng:
```
// Identifier that uniquely identifies a process within a VM/container.
message ProcessIdentifier {

  // The host name. Usually refers to the machine/container name.
  // For example: os.Hostname() in Go, socket.gethostname() in Python.
  string host_name = 1;

  // Process id.
  uint32 pid = 2;

  // Start time of this ProcessIdentifier. Represented in epoch time.
  google.protobuf.Timestamp start_timestamp = 3;
}
```

I think there is only `pid` makes sense? Do you think `_` is good, or should consider `.`? Such as `node.identifier.host_name`.

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/oc/OCMetricHandler.java
wu-sheng: Do you mean build an node with mock values of fields we need?

hanahmily: Exactly, `Optional.orElse()` is what I like. Some defensive programming style.

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/oc/OCMetricHandler.java
hanahmily: MAL has a capacity to transform `.` to `_`, which means no matter what we choose, the final result must be `node_identifier_host_name `.

wu-sheng: OK, then `_` is fine.

hanahmily: A common use case about the `start_timestamp` is to get `uptime` metric, the expression should leverage `start_timestamp` and `MAL` 's `now()` function as below:

`now()-start_timestamp`

wu-sheng: I am just wondering whether this is the meaning, as it is reported every time, rather than a separated metric.

wankai123:
I added some defensive codes, and added a label with `key = node_identifier_pid` and value from `Node.identifier.pid` defined in opencensus agent proto.
I  don't add `start_timestamp` to the labels yet,  I think the `start_timestamp`  is a value and in MAL ladel is used for filter or aggregation, not for count.

wu-sheng (approved):
Generally good to me. @hanahmily Recheck please.

hanahmily (approved):
LGTM

PR #6249, opened 2021-01-24
add text analyzer for es

EvanLjp:

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
@@ -62,17 +62,16 @@ public @interface Column {
      * @return the length of this column, this is only for {@link String} column. The usage of this depends on the
      * storage implementation.
      *
-     * Notice, different lengths may cause different types.
-     * Such as, over 16383 would make the type in MySQL to be MEDIUMTEXT, due to database varchar max=16383
-     *
+     * Notice, different lengths may cause different types. Such as, over 16383 would make the type in MySQL to be
+     * MEDIUMTEXT, due to database varchar max=16383
      * @since 7.1.0
      */
     int length() default 200;
 
     /**
      * The return name of system environment could provide an override value of the length limitation.
-     * @return the variable name of system environment.
      *
+     * @return the variable name of system environment.
      * @since 8.2.0
      */
     String lengthEnvVariable() default "";
@@ -86,6 +85,27 @@ public @interface Column {
      * @since 8.0.0
      */
     ValueDataType dataType() default ValueDataType.NOT_VALUE;
+
+    /**
+     * The storage analyzer mode.
+     *
+     * @since 8.4.0
+     */
+    AnalyzerType analyzer() default AnalyzerType.OAP_ANALYZER;
+
+    /**
+     * The analyzer declares the text analysis mode.
+     */
+    enum AnalyzerType {
+        /**
+         * The default analyzer.
+         */
+        OAP_ANALYZER,
+        /**
+         * The log analyzer.
+         */
+        OAP_LOG_ANALYZER,
+    }
 
     /**
      * ValueDataType represents the data structure of value column. The persistent way of the value column determine the
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
@@ -62,16 +62,17 @@ public @interface Column {
      * @return the length of this column, this is only for {@link String} column. The usage of this depends on the
      * storage implementation.
      *
-     * Notice, different lengths may cause different types. Such as, over 16383 would make the type in MySQL to be
-     * MEDIUMTEXT, due to database varchar max=16383
+     * Notice, different lengths may cause different types.
+     * Such as, over 16383 would make the type in MySQL to be MEDIUMTEXT, due to database varchar max=16383
+     *
      * @since 7.1.0
      */
     int length() default 200;
 
     /**
      * The return name of system environment could provide an override value of the length limitation.
-     *
      * @return the variable name of system environment.
+     *
      * @since 8.2.0
      */
     String lengthEnvVariable() default "";
@@ -100,11 +101,18 @@ public @interface Column {
         /**
          * The default analyzer.
          */
-        OAP_ANALYZER,
+        OAP_ANALYZER("oap_analyzer"),
         /**
          * The log analyzer.
          */
-        OAP_LOG_ANALYZER,
+        OAP_LOG_ANALYZER("oap_log_analyzer");
+
+        @Getter
+        private final String name;
+
+        AnalyzerType(final String name) {
+            this.name = name;
+        }
     }
 
     /**
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/Model.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/Model.java
@@ -19,9 +19,11 @@
 package org.apache.skywalking.oap.server.core.storage.model;
 
 import java.util.List;
+import java.util.Set;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import org.apache.skywalking.oap.server.core.analysis.DownSampling;
+import org.apache.skywalking.oap.server.core.storage.annotation.Column;
 
 /**
  * The model definition of a logic entity.
@@ -37,6 +39,7 @@ public class Model {
     private final boolean record;
     private final boolean superDataset;
     private final boolean isTimeSeries;
+    private final Set<Column.AnalyzerType> analyzer;
 
     public Model(final String name,
                  final List<ModelColumn> columns,
@@ -44,7 +47,8 @@ public class Model {
                  final int scopeId,
                  final DownSampling downsampling,
                  final boolean record,
-                 final boolean superDataset) {
+                 final boolean superDataset,
+                 final Set<Column.AnalyzerType> analyzer) {
         this.name = name;
         this.columns = columns;
         this.extraQueryIndices = extraQueryIndices;
@@ -53,5 +57,6 @@ public class Model {
         this.isTimeSeries = !DownSampling.None.equals(downsampling);
         this.record = record;
         this.superDataset = superDataset;
+        this.analyzer = analyzer;
     }
 }
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/Model.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/Model.java
@@ -19,11 +19,9 @@
 package org.apache.skywalking.oap.server.core.storage.model;
 
 import java.util.List;
-import java.util.Set;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import org.apache.skywalking.oap.server.core.analysis.DownSampling;
-import org.apache.skywalking.oap.server.core.storage.annotation.Column;
 
 /**
  * The model definition of a logic entity.
@@ -39,7 +37,6 @@ public class Model {
     private final boolean record;
     private final boolean superDataset;
     private final boolean isTimeSeries;
-    private final Set<Column.AnalyzerType> analyzer;
 
     public Model(final String name,
                  final List<ModelColumn> columns,
@@ -47,8 +44,7 @@ public class Model {
                  final int scopeId,
                  final DownSampling downsampling,
                  final boolean record,
-                 final boolean superDataset,
-                 final Set<Column.AnalyzerType> analyzer) {
+                 final boolean superDataset) {
         this.name = name;
         this.columns = columns;
         this.extraQueryIndices = extraQueryIndices;
@@ -57,6 +53,5 @@ public class Model {
         this.isTimeSeries = !DownSampling.None.equals(downsampling);
         this.record = record;
         this.superDataset = superDataset;
-        this.analyzer = analyzer;
     }
 }
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ModelColumn.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ModelColumn.java
@@ -39,7 +39,8 @@ public class ModelColumn {
                        boolean matchQuery,
                        boolean storageOnly,
                        boolean isValue,
-                       int length, final Column.AnalyzerType analyzer) {
+                       int length,
+                       Column.AnalyzerType analyzer) {
         this.columnName = columnName;
         this.type = type;
         this.genericType = genericType;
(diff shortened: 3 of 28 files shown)


EvanLjp:
let's define a default log text analyzer first, would add doc later. @wu-sheng @arugal

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
```diff
@@ -62,17 +62,16 @@
...
+     * MEDIUMTEXT, due to database varchar max=16383
      * @since 7.1.0
      */
     int length() default 200;
 
     /**
      * The return name of system environment could provide an override value of the length limitation.
-     * @return the variable name of system environment.
      *
+     * @return the variable name of system environment.
```
wu-sheng, line 74: I think you should not format this part.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -135,17 +155,19 @@ protected void createTable(Model model) throws StorageException {
...
-                originalColumn.put("type", columnTypeEsMapping.transform(columnDefine.getType(), columnDefine.getGenericType()));
+                originalColumn.put(
+                    "type", columnTypeEsMapping.transform(columnDefine.getType(), columnDefine.getGenericType()));
                 originalColumn.put("copy_to", matchCName);
                 properties.put(columnDefine.getColumnName().getName(), originalColumn);
 
                 Map<String, Object> matchColumn = new HashMap<>();
                 matchColumn.put("type", "text");
-                matchColumn.put("analyzer", "oap_analyzer");
+                matchColumn.put("analyzer", columnDefine.getAnalyzer().name().toLowerCase());
```
wu-sheng, line 165: Please provide a String value Analyzer, rather than `toLowerCase`.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/AnalyzerSetting.java
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * An optional array of built-in or customised token filters.
+     */
+    private Map<String, Object> filter = new HashMap<>();
+    /**
+     * The custom analyzers.
+     */
+    private Map<String, Object> analyzer = new HashMap<>();
+
+    public void combine(AnalyzerSetting analyzerSetting) {
```
wu-sheng, line 49: Why do we need `combine`? Could you share the use case?

EvanLjp: image
one index setting has multi analyzers

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java
```diff
@@ -20,6 +20,7 @@
 
 import lombok.Getter;
 import lombok.Setter;
+import org.apache.skywalking.oap.server.core.storage.annotation.Column;
```
wu-sheng, line 23: This is illegal import.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
```diff
@@ -87,6 +86,27 @@
...
+     * The storage analyzer mode.
+     *
+     * @since 8.4.0
+     */
+    AnalyzerType analyzer() default AnalyzerType.OAP_ANALYZER;
+
+    /**
+     * The analyzer declares the text analysis mode.
+     */
+    enum AnalyzerType {
```
wu-sheng, line 99: As an enum type you should put more information in the type, rather than separating them in the codes.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java
EvanLjp: it's used in link, may not understand your suggestion to modify this.
image

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
EvanLjp: fixed

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: fixed

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
EvanLjp: fixed

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
wu-sheng: You should add `getSettings` method to return the settings.

EvanLjp: I would add a functional interface to refactor it.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -113,14 +115,32 @@ protected void createTable(Model model) throws StorageException {
         setting.put("index.refresh_interval", model.isRecord()
             ? TimeValue.timeValueSeconds(10).toString()
             : TimeValue.timeValueSeconds(config.getFlushInterval()).toString());
-        setting.put("analysis.analyzer.oap_analyzer.type", "stop");
+        setting.put("analysis", getAnalyzerSetting(model.getAnalyzer()));
```
wu-sheng, line 118: I think we should separate settings based on model's columns. Then we don't need `combine` anymore. Also, I think it should be easier to understand.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
wu-sheng: As an enum, I think a method is enough. We would use the interface to anywhere else.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: Analyzer is for columns, so it should be for column configuration, and setting needs to define the configuration of multiple columns in combination at the same time, so I think the current structure is clear. 

this is a test example: 
```json
{
  "log" : {
    "order" : 0,
    "index_patterns" : [
      "log-*"
    ],
    "settings" : {
      "index" : {
        "refresh_interval" : "10s",
        "analysis" : {
          "filter" : {
            "cjk_log_stop_filter" : {
              "type" : "stop",
              "stopwords" : "_none_"
            },
            "cjk_log_bigram_filter" : {
              "output_unigrams" : "true",
              "type" : "cjk_bigram"
            },
            "cjk_log_word_delimiter" : {
              "type" : "word_delimiter",
              "preserve_original" : "true",
              "split_on_numerics" : "false"
            }
          },
          "analyzer" : {
            "oap_log_analyzer" : {
              "filter" : [
                "cjk_width",
                "lowercase",
                "cjk_log_bigram_filter",
                "cjk_log_word_delimiter",
                "cjk_log_stop_filter"
              ],
              "type" : "custom",
              "tokenizer" : "standard"
            },
            "oap_analyzer" : {
              "type" : "stop"
            }
          }
        },
        "number_of_shards" : "5",
        "number_of_replicas" : "0"
      }
    },
    "mappings" : {
      "properties" : {
        "trace_id" : {
          "type" : "keyword"
        },
        "unique_id" : {
          "type" : "keyword"
        },
        "content_match" : {
          "analyzer" : "oap_log_analyzer",
          "type" : "text"
        },
        "span_id" : {
          "type" : "integer"
        },
        "endpoint_name" : {
          "copy_to" : "endpoint_name_match",
          "type" : "keyword"
        },
        "endpoint_id" : {
          "type" : "keyword"
        },
        "service_instance_id" : {
          "type" : "keyword"
        },
        "content" : {
          "copy_to" : "content_match",
          "type" : "keyword"
        },
        "endpoint_name_match" : {
          "analyzer" : "oap_analyzer",
          "type" : "text"
        },
        "tags" : {
          "type" : "keyword"
        },
        "trace_segment_id" : {
          "type" : "keyword"
        },
        "content_type" : {
          "index" : false,
          "type" : "integer"
        },
        "tags_raw_data" : {
          "index" : false,
          "type" : "binary"
        },
        "service_id" : {
          "type" : "keyword"
        },
        "time_bucket" : {
          "type" : "long"
        },
        "is_error" : {
          "type" : "integer"
        },
        "timestamp" : {
          "type" : "long"
        }
      }
    },
    "aliases" : {
      "log" : { }
    }
  }
}
```

wu-sheng: My point is, when do you need `oap_analyzer` and `oap_log_analyzer` at the same time?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java
wu-sheng: I think you will see `maven` failure. You need to use the full name in the `@link`.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: Maybe only used in the log mode. So you want to differentiate based on type? That means we maybe need a new annotation. Is it a little heavier than now?

wu-sheng: You have changed `Column` annotation, doesn't you? In there you have known the analyzer, and here, you have `Model model`, which provided the information you need. 

There are 2 ways to enhance the analyzer mechanism
1. What you do right now, need to support combine. 
2. Every index has one analyzer based on Column declaration. No combine needed.

My question is, do you have any case to support 2 analyzer(s) for one index.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/AnalyzerSetting.java
```diff
@@ -0,0 +1,119 @@
...
+        private final Column.AnalyzerType type;
+        private final GenerateAnalyzerSettingFunc func;
+
+        Generator(final Column.AnalyzerType type,
+                  final GenerateAnalyzerSettingFunc func) {
+            this.type = type;
+            this.func = func;
+        }
+
+        public GenerateAnalyzerSettingFunc GetGenerateFunc() {
```
wu-sheng, line 99: Why these methods have first letter in upper case?

EvanLjp: fixed, recently, keep writing golang, so.....

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/Model.java
```diff
@@ -37,14 +39,16 @@
...
 
     public Model(final String name,
                  final List<ModelColumn> columns,
                  final List<ExtraQueryIndex> extraQueryIndices,
                  final int scopeId,
                  final DownSampling downsampling,
                  final boolean record,
-                 final boolean superDataset) {
+                 final boolean superDataset,
+                 final Set<Column.AnalyzerType> analyzer) {
```
wu-sheng, line 51: `List<ModelColumn> columns` should include the analyzer(s) already, we don't need to add this new parameter.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ModelColumn.java
```diff
@@ -30,19 +31,21 @@
...
+    private final Column.AnalyzerType analyzer;
 
     public ModelColumn(ColumnName columnName,
                        Class<?> type,
                        Type genericType,
                        boolean matchQuery,
                        boolean storageOnly,
                        boolean isValue,
-                       int length) {
+                       int length, final Column.AnalyzerType analyzer) {
```
wu-sheng, line 42: Please use a new line to follow the format. And make all parameters `as final` or none. Don't mix the style

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/AnalyzerSetting.java
```diff
@@ -0,0 +1,119 @@
...
+                  final GenerateAnalyzerSettingFunc func) {
+            this.type = type;
+            this.func = func;
+        }
+
+        public GenerateAnalyzerSettingFunc getGenerateFunc() {
+            return this.func;
+        }
+
+        public static Generator GetGenerator(Column.AnalyzerType type) throws StorageException {
```
wu-sheng, line 103: Still have upper case issue.

wu-sheng: I think style-check should fail, please check locally.

wu-sheng (approved):
LGTM

PR #6255, opened 2021-01-25
Fix that the dynamic configuration is deleted on the server side but does not take effect on the agent.

zifeihan:
### Fix that the dynamic configuration is deleted on the server side but does not take effect on the agent.
- [ ] Explain briefly why the bug exists and how to fix it.

Fix that the dynamic configuration is deleted on the server side but does not take effect on the agent.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/dynamic/ConfigurationDiscoveryService.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/dynamic/ConfigurationDiscoveryService.java
@@ -18,15 +18,18 @@
 
 package org.apache.skywalking.apm.agent.core.conf.dynamic;
 
+import com.google.common.collect.Lists;
 import io.grpc.Channel;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/dynamic/ConfigurationDiscoveryService.java
```diff
@@ -168,6 +172,30 @@ public void handleConfigurationDiscoveryCommand(ConfigurationDiscoveryCommand co
...
+     */
+    private List<KeyStringValuePair> readConfig(ConfigurationDiscoveryCommand configurationDiscoveryCommand) {
+        Map<String, KeyStringValuePair> commandConfigs = configurationDiscoveryCommand.getConfig()
+                                                                                      .stream()
+                                                                                      .collect(Collectors.toMap(
+                                                                                          KeyStringValuePair::getKey,
+                                                                                          Function.identity()
+                                                                                      ));
+        List<KeyStringValuePair> configList = Lists.newArrayList();
+        for (final String name : register.keys()) {
```
wu-sheng, line 190: So you are building the list with all keys, right? That is what we missed today.

zifeihan: Yes, sorry, I didn't think about it before.

wu-sheng (approved):
LGTM

kezhenxu94 (approved):

PR #6263, opened 2021-01-27
Remove duplicate codes in Envoy ALS handler

kezhenxu94:

They are already included in https://github.com/apache/skywalking/blob/972145c2083936d83d8d83270e2ffa84cdb04dd5/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/Analyzer.java#L115

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ Release Notes.
 * Add ElasticSearch 7.10 to test matrix and verify it works.
 * Replace Apache RAT with skywalking-eyes to check license headers.
 * Set up test of Envoy ALS / MetricsService under Istio 1.8.2 to verify Envoy V3 protocol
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -11,6 +11,7 @@ Release Notes.
 * Add ElasticSearch 7.10 to test matrix and verify it works.
 * Replace Apache RAT with skywalking-eyes to check license headers.
 * Set up test of Envoy ALS / MetricsService under Istio 1.8.2 to verify Envoy V3 protocol
+* Chore:  Remove duplicate codes in Envoy ALS handler
```
wu-sheng, line 14: Please move this to the OAP part.

wu-sheng (approved):

PR #6266, opened 2021-01-27
Support dynamic change agent.trace.ignore_path on java agent within CDS.

JaredTan95:
Support config `agent.trace.ignore_path` can be changed in the runtime.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/pull/6164


JaredTan95:
Local tested with Apollo configuration:

image

image

CHANGES.md
```diff
@@ -37,7 +37,7 @@ Release Notes.
 * Support collecting logs and forwarding through gRPC.
 * Support config `agent.sample_n_per_3_secs` can be changed in the runtime.
 * Support DNS periodic resolving mechanism to update backend service.
-
+* Support config `agent.trace.ignore_path` can be changed in the runtime.
```
wu-sheng, line 40:
```suggestion
* Support config `agent.trace.ignore_path` can be changed in the runtime.

```

wu-sheng (commented):
Please update the agent side document about what are the formats of key and value.

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/TraceIgnorePatternWatcher.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    private void activeSetting(String config) {
+        if (LOGGER.isDebugEnable()) {
+            LOGGER.debug("Updating using new static config: {}", config);
+        }
+        try {
+            this.traceIgnorePathPatterns.set(config);
+            traceIgnoreExtendService.handleTraceIgnorePatternsChanged();
+        } catch (NumberFormatException ex) {
```
kezhenxu94, line 46: no `NumberFormatException` will be thrown in this case

JaredTan95: yes, copied from SampleRateWatcher,I will delete id.

kezhenxu94 (approved):

zifeihan (approved):
LGTM

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/TraceIgnoreExtendService.java
```diff
@@ -76,4 +85,10 @@ public boolean trySampling(final String operationName) {
     public void forceSampled() {
         super.forceSampled();
     }
+
+    void handleTraceIgnorePatternsChanged() {
+        if (StringUtil.isNotBlank(traceIgnorePatternWatcher.getTraceIgnorePathPatterns())) {
```
wu-sheng, line 90: What happens if I set the value as blank? I think the PR misses some logic here.

wu-sheng (changes requested):

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/TraceIgnoreExtendService.java
kezhenxu94: Blank string is considered as `DELETE` event, and it will use the default value (`TraceIgnorePatternWatcher#51`) of `IgnoreConfig.Trace.IGNORE_PATH`, so `traceIgnorePatternWatcher.getTraceIgnorePathPatterns()` may be effectively not blank unless `IgnoreConfig.Trace.IGNORE_PATH` is set to empty

wu-sheng (approved):

wu-sheng:
With limited CI resource, let's merge #6268 first, then we could begin the pre-release testing.

PR #6270, opened 2021-01-27
Fix a legal metric query adoption bug.

weiqiang-w:
if ids is null ,the function query the values but never add to return list.


wu-sheng (changes requested):
Please describe what you are doing, and update the changelog. Your PR is not complete.

wu-sheng (dismissed):
I updated the changelog for you.

wu-sheng (approved):

wu-sheng (approved):

PR #6272, opened 2021-01-27
Support dynamic change agent.ignore_suffix on java agent.

Ax1an:
Support config agent.ignore_suffix can be changed in the runtime.

- [x] Update the CHANGES log.

Local tested：
image
image


zifeihan (commented):
Wait https://github.com/apache/skywalking/pull/6266, Some doc may conflict.

wu-sheng:
@zifeihan You could review the codes, we are facing limited CI resources, so merge would be very slow.

zifeihan (approved):

zifeihan:
> @zifeihan You could review the codes, we are facing limited CI resources, so merge would be very slow.
 
I double check, except that the documents will conflict, everything else is good for me.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManagerExtendService.java
```diff
@@ -84,4 +92,10 @@ public AbstractTracerContext createTraceContext(String operationName, boolean fo
     public void statusChanged(final GRPCChannelStatus status) {
         this.status = status;
     }
+
+    void handleIgnoreSuffixPatternsChanged() {
+        if (StringUtil.isNotBlank(ignoreSuffixPatternsWatcher.getIgnoreSuffixPatterns())) {
+            ignoreSuffixArray = ignoreSuffixPatternsWatcher.getIgnoreSuffixPatterns().split(",");
```
wu-sheng, line 98: @kezhenxu94 @zifeihan I want to discuss about this detail, including the previous implementation of `TraceIgnoreExtendService`.
The `ignoreSuffixArray` doesn't have `volatile` key word, due to another thread is watching and updating the value, the latest value will not be affected in some threads immediately.
According to this kind of update is not urgent, it should be fine. But I am to check with you two, have we agreed this is our conclusion?

zifeihan: I agree with you, since this is not an urgent operation, I think it should be fine. What do you think? @kezhenxu94

kezhenxu94: > The ignoreSuffixArray doesn't have volatile key word, due to another thread is watching and updating the value, the latest value will not be affected in some threads immediately. According to this kind of update is not urgent, it should be fine.

Since most of the notifications of config centers are not real-time (have delay) themselves, the fields not being `volatile` may be acceptable if it has no side effects.

> According to this kind of update is not urgent, it should be fine. But I am to check with you two, have we agreed this is our conclusion?

I don't remember we have this conclusion before. TBH, I didn't notice this part indeed 🤦🏻‍♂️

wu-sheng: This isn't a conclusion yet, but I noticed this, 1st dynamic configuration is sampling rate, which is an AtomicReference, so it is fine. Then, ignore plugin, it has been approved and merged, with no `volatile`. And I can see @zifeihan approved this too.

So, I want to ask, what do you think about this? Sometimes, no `volatile` is tricky about when it effects in the runtime.

wu-sheng: Once we have this agreement, I want this to be noticed in a new GitHub issue and mail list for searching in the future.

kezhenxu94: I prefer to add `volatile` or similar semantics (like `AtomicXxxx`)

zifeihan: > I prefer to add `volatile` or similar semantics (like `AtomicXxxx`)

Agree with you.

wu-sheng: @zifeihan OK, then you need to fix your previous merged PR ASAP, we are closing the release.

docs/en/setup/service-agent/java-agent/configuration-discovery.md
```diff
@@ -25,6 +25,7 @@ Java agent supports the following dynamic configurations.
 |        Config Key         |                      Value Description                       | Value Format Example  | Required Plugin(s) |
 | :-----------------------: | :----------------------------------------------------------: | :-------------------: | :----------------: |
 | agent.sample_n_per_3_secs |          The number of sampled traces per 3 seconds          |          -1           | - |
+| agent.ignore_suffix       |          Ignore the segments if their operation names end with these suffix. multiple paths should be separated by `,`        |          `.txt,.log`         | - |
```
wu-sheng, line 28: The description is not correct, take a look at `Config#Agent#IGNORE_SUFFIX`. This can't affect every operation name.

> If the operation name of the first span is included in this set, this segment should be ignored.

Ax1an: Updated.

wu-sheng (approved):

Issue #3661, opened 2019-10-20
"java -ext.dirs" is not supported


jackiepon31:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ 1] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?
I just want to fix something to the AgentClassLoader
 
___
### Bug
- Which version of SkyWalking, OS and JRE?

- Which company or project?

- What happen?
In some cases, especially when the target java app is packaged in an old way, i mean the depency libs directory is out of the target application jar.  We need to add some VM options to make sure the java applications starts well, like "java -cp **/lib   -ext.dirs **/lib" . 
But when we add these VM options, the target ClassLoaders may not be able to load any classes from skywalking-agent.jar cause it's out of the path  i have defined previously.  The target app classloader or any of its parent classloader will load the class with failure and push it down to AgentClassLoader to load the class ,and finally ,the not class found exception will stop the application from start.
___
### Requirement or improvement
- Please describe about your requirements or improvement suggestions.

I suggest to add SkyWalking-agent.jar or its dirs to classpath to make the load mechanism compatible with old packaing way instead of spring-boot fat jar.

wu-sheng:
Yes. ext-dir is not supported today. But I think -cp should work. If you want to fix this, welcome. We may not have slot to fix this shortly as low priority, little people use this way to package anymore.

wu-sheng:
@candyleer If you have the interest, welcome to consider this.

jackiepon31:
thanks a lot for your answer, yeah，really ware to see this old style of packing, 
i'm trying to build an APM system for my company , after 5 years of iteration, the Architecture of backend is really really complex and drive me to crazy ,lol
so, I'm right now doing some pre-study work on skywalking and i am gonna create some new pluggings and new storage implementations . it's realling chanllenging !! 
I'm hoping for your support,maybe in the nearby future.

wu-sheng:
Plugin development is pretty easy in SkyWalking. We have detailed documents for that, read that, you should be fine. Just in open source, there are dozens, even over 100+ contributors are/were working on SkyWalking plugins, so, don't consider it is complex. SkyWalking's core has taken care of the complex part, it is just a more powerful AOP style core already.

jackiepon31:
cool, thanks for your advice , how can i join u guys as a contributors ?

wu-sheng:
[...]

PR #6273, opened 2021-01-27
[FAQ doc]How to load outside jar to classpath

HendSame:
[FAQ doc]How to load outside jar to classpath


wu-sheng:
Please update the CHANGES.md in the root.

wu-sheng:
Relate to #3661

wu-sheng:
Please don't sync the branch by yourself. We are facing limited CI resource issue due to too many PRs in the Apache.
We will update when this PR has been reviewed and ready to merge. Thanks.

HendSame:
> Please don't sync the branch by yourself. We are facing limited CI resource issue due to too many PRs in the Apache.
> We will update when this PR has been reviewed and ready to merge. Thanks.

OK

docs/en/FAQ/Load-outside-jar-to-classpath.md
```diff
@@ -0,0 +1,24 @@
+# Load outside jar to classpath 
+ 
+### Problem
+In some cases, especially when the target java app is packaged in an old way,that mean the depency libs directory is out of the target application jar. We need to add some VM options to make sure the java applications starts well, like "java -cp **/lib -ext.dirs **/lib" .
```
wu-sheng, line 4:
```suggestion
In some cases, the application is not package as a urber jar(a.k.a. fat0jar). Also they don't only use `-cp:` to add all dependencies, instead they use like "java -cp **/lib -ext.dirs **/lib" to boot .
```

docs/en/FAQ/Load-outside-jar-to-classpath.md
```diff
@@ -0,0 +1,24 @@
+# Load outside jar to classpath
```
wu-sheng, line 1:
```suggestion
# How to resolve the issue, "java -ext.dirs" is not supported
```

docs/en/FAQ/Load-outside-jar-to-classpath.md
```diff
@@ -0,0 +1,24 @@
+# Load outside jar to classpath 
+ 
+### Problem
+In some cases, especially when the target java app is packaged in an old way,that mean the depency libs directory is out of the target application jar. We need to add some VM options to make sure the java applications starts well, like "java -cp **/lib -ext.dirs **/lib" .
+But when we add these VM options, the target ClassLoaders may not be able to load any classes from skywalking-agent.jar cause it's out of the path i have defined previously. The target app classloader or any of its parent classloader will load the class with failure and push it down to AgentClassLoader to load the class ,and finally ,the not class found exception will stop the application from start.
```
wu-sheng, line 5:
```suggestion
But when we use `--ext.dirs` options, the target ClassLoader(s) may not be able to load any classes from skywalking-agent.jar due to your codes are loaded by Extension Class Loader, but the agent class is going to be loaded in Application Class Loader as it is a Java agent. 
SkyWalking agents provides native ways to support Bootstrap Class Loader to avoid similar issue, but doesn't support this case.
```

docs/en/FAQ/Load-outside-jar-to-classpath.md
```diff
@@ -0,0 +1,24 @@
...
+
+### Resolve
+"-Xbootclasspath/a:" can solve this problem.
```
wu-sheng, line 8:
```suggestion
Use "-Xbootclasspath/a:" JVM option to change the class loader in order to resolve this issue.
```

docs/en/FAQ/Load-outside-jar-to-classpath.md
```diff
@@ -0,0 +1,24 @@
...
+By the way ,we can use the BootstrapClassLoader to load the them.
```
wu-sheng, line 9: Use `BootstrapClassLoader`? Is this working? I remember all agent codes(except few JDK instrumentation) are loaded by Application ClassLoader only, please confirm.

docs/en/FAQ/README.md
```diff
@@ -27,6 +27,7 @@ These are known and common FAQs. We welcome you to contribute yours.
 * [Compatible with other javaagent bytecode processing](Compatible-with-other-javaagent-bytecode-processing.md)
 * [**Java agent memory leak** when enhance `Worker thread` at use Thread Pool](Memory-leak-enhance-Worker-thread.md)
 * [Thrift plugin](thrift-plugin.md)
+* ["java -ext.dirs" is not supported,So how to load outside jar to classpath?](Load-outside-jar-to-classpath.md)
```
wu-sheng, line 30:
```suggestion
* [How to resolve the issue, "java -ext.dirs" is not supported](Load-outside-jar-to-classpath.md)
```

Please change the name of markdown file to `resolve-ext-dir-issue.md`.

wu-sheng (commented):
I did some polish for your doc, with a question, please check.
And all changes should be done in your local env, and commit with one commit to reduce the cost of CI resources.

docs/en/FAQ/Load-outside-jar-to-classpath.md
HendSame: Yes, that will use BootstrapClassLoader. I have a question : Why the target ClassLoader(s) need to load classes from skywalking-agent.jar? I think I got it wrong.

wu-sheng: I think we don't need `BootstrapClassLoader`, we need application's codes could access agent's classes, which are loaded by application class loader due to `-javaagent:agent.jar`.
That is why I am confused about your writing.

HendSame: Got it. What if we can get application's codes and  agent's codes to load  by application class loader.  For example: Instrumentation.**_appendToSystemClassLoaderSearch_**(application's  jar or application's dependency jar) ? Is this working?

wu-sheng: I am not sure. But basically, I don't recommend you to manipulate the class loader of users' codes. It is hard to tell whether they are using some private class loader mechanism, which could be failure due to this change.
This kind of things could work in your own practice, but it may not work for anyone else.

HendSame: I see.So I close this PR ?

wu-sheng: Agree, I think we need some other ideas to resolve this issue. :)

PR #6277, opened 2021-01-29
Support Prometheus node-exporter VM metrics monitoring

### LINKED ISSUE #6108: Adopt the INFRA VM/Machine monitoring
Since @hanahmily brings the OpenTelemetry receiver in the SkyWalking's landscape, now we could move on to the new territory, this time, VM/Machine Infrastructure.

@wankai123 pinged me about he wants to take part in this. FYI @hanahmily @kezhenxu94 cc @apache/skywalking-committers 

In the implementation, I prefer, Prometheus eco-system agents -> OpenTelemetry collector --(OC format)--> OAP server. 
Such as Prometheus node exporter, https://prometheus.io/docs/guides/node-exporter/
@hanahmily agree?

---
~~At the backend side, we need to add a new `vm_traffic` to represent VMs and Machines, which should include~~
~~1. name, entity id. IDManager should implement this.~~
~~2. tags (JsonObject properties), like InstanceTraffic implemented.~~

~~The key point of this is adding aggregation capability for MAL core. From my understanding the key point is adding `SampleFamily#vm` for new entity context, also `Analyzer#generateTraffic` should be enhanced to support the new aggregation type.~~

~~After above finish, the UI(cc @Fine0830) will add a new kind of dashboard called `Infrastructure `, which just query `VM` list. NOTE, this is a query-protocol level change.~~

---
Based on https://github.com/apache/skywalking/issues/6108#issuecomment-758311084, the change proposal in the above block is not necessary. Updated at Jan. 12th 2021.

@wankai123 If you have any question, please discuss at here.

wankai123:
### Support Prometheus node-exporter VM metrics monitoring
- [X] If it's UI related, attach the screenshots below.
     

- [X] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6108.

Since we define the VM entity as a service in OAP, I use `vm::`  as a prefix in MAL to identify.
### VM Metrics flow
Prometheus node-exporter ---(pull)---> OpenTelemetry-collector ---(push)---> (otel-receiver)OAP server
### Monitored metrics
- CPU Average Used
- CPU Load
- Memory RAM
- Memory Swap
- File System
- Disk R/W
- Network Bandwidth Usage
- Network Status
- Filefd Allocated

The above metrcs support Linux, has bean tested with `node_exporter-1.0.1.linux-amd64` in Ubuntu16 & CentOS7


wu-sheng:
Let's add a table in this part of documentation, https://github.com/apache/skywalking/blob/master/docs/en/setup/backend/backend-receivers.md#opentelemetry-receiver

| Rule Name | Description | Configuration File | Data Source |
|----|----|-----|----|
|The name should be used in enabledOcRules| What metrics included| file name and path| Who could send this kind of metrics|

oap-server/server-bootstrap/src/main/resources/otel-oc-rules/vm.yaml
```diff
@@ -0,0 +1,93 @@
...
+    exp: node_load15 * 100
+
+    #node Memory
+  - name: memory_total
+    exp: node_memory_MemTotal_bytes
+  - name: memory_available
+    exp: node_memory_MemAvailable_bytes
+  - name: memory_used
+    exp: node_memory_MemTotal_bytes - node_memory_MemAvailable_bytes
+  - name: memory_SwapFree
```
hanahmily, line 51: The mix of camel style and sake style seems not good practice to the metric name. Could you make them consistent?

hanahmily:
Would we add some `single number` panels to show CPU percentage, cores' number, total mem to provide a glance at the global status of VM?


> ### Screenshots
> 
>

wu-sheng:
Also, please help on adding table to this meter receiver part, https://github.com/apache/skywalking/blob/master/docs/en/setup/backend/backend-receivers.md#meter-receiver

wu-sheng:
@wankai123 Let me polish the document structure a little, I will send a new pull request. I think the current structure about OTEL is not clear enough

wankai123:
> Would we add some `single number` panels to show CPU percentage, cores' number, total mem to provide a glance at the global status of VM?
> 
> > ### Screenshots
> > 
> > 

Yep, I tried to do this. But the core's info in the label of the metrics, not a value， MAL can't calculate the labels now.
So I can't get core' number and total CPU uesed percentage, I’m prepare to add a  function `count()` to calculate the
labels number，then we can do this

wu-sheng:
I did the polish here #6278, you could sync and update your part after it is merged.

oap-server/server-bootstrap/src/main/resources/otel-oc-rules/vm.yaml
wankai123: The metric name in exp is from 'node-exporter' can't be modify， you mean make the  expression `name ` consistent？

wu-sheng: Yes, I think so

wankai123: done

wu-sheng:
Doc updated, and synced. Please update.

wu-sheng:
@hanahmily Should we add an issue to track `count` func in MAL?

wankai123:
doc updated @wu-sheng

wu-sheng:
> Would we add some `single number` panels to show CPU percentage, cores' number, total mem to provide a glance at the global status of VM?
> 
> > ### Screenshots
> > 
> > 

@wankai123 I think you misunderstood Gao's point. He was asking about the panel, the 2nd level of the dashboard to avoid `vertical scrolling`. I am fine will both ways.

wu-sheng:
@wankai123 Please post the latest screenshots here.

wankai123:
> Would we add some `single number` panels to show CPU percentage, cores' number, total mem to provide a glance at the global status of VM?
> 
Add some single number panels, change 'File System' to 'File System Mountpoint Usage' @wu-sheng @hanahmily

wu-sheng (dismissed):
Good to me. @hanahmily recheck?

hanahmily (dismissed):
LGTM

wu-sheng (approved):

PR #6278, opened 2021-01-29
Polish doc structure.

wu-sheng:

FYI @wankai123

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,21 +25,24 @@ The core features are following.
 - Slow services and endpoints detected
 - Performance optimization
 - Distributed tracing and context propagation
-- Database access metrics. Detect slow database access statements(including SQL statements).
+- Database access metrics. Detect slow database access statements(including SQL statements)
 - Alarm
 - Browser performance monitoring
+- Infrastructure(VM, network, disk etc.) monitoring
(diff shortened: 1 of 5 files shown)


kezhenxu94 (dismissed):

README.md
```diff
@@ -25,21 +25,24 @@ The core features are following.
 - Slow services and endpoints detected
 - Performance optimization
 - Distributed tracing and context propagation
-- Database access metrics. Detect slow database access statements(including SQL statements).
+- Database access metrics. Detect slow database access statements(including SQL statements)
 - Alarm
 - Browser performance monitoring
+- Infrastructure(VM, network, disk etc.) monitoring
+- Collaboration across metrics, traces, and logs
```
kezhenxu94, line 32: I just want to make sure this is not a typo of "Correlation"

wu-sheng: I was struggling in both too... I was thinking, we are going to generate metrics from logs, and is generating metrics from traces, rather than simply Correlation(such as trace id in logs). But I am not sure whether this choice causes confusing.
What do you think?

kezhenxu94 (approved):
LGTM

PR #6285, opened 2021-01-30
Fix missed group in UI template

wu-sheng:

We should add group in the template to exclude the normal services from VM services.
@wankai123 Could you test locally?


kezhenxu94 (approved):

wankai123 (approved):
Tested in local env
WechatIMG1158

PR #6295, opened 2021-01-31
Multiple exclude and include query conditions don't work as expected.

wu-sheng:

Besides title said, I also polish the codes as mixed used `boolQueryBuilder.must()` and `mustQueryList`(choose this only).


arugal (approved):
LGTM

EvanLjp (approved):

PR #6303, opened 2021-02-02
Remove invalid mysql configuration in agent.config.

Ax1an:
Remove invalid mysql configuration in agent.config.

- [x] Update the CHANGES log.


wu-sheng:
Good catch, please update the change log too.

wu-sheng (approved):

PR #6304, opened 2021-02-02
chore: add @Override tag on some methods

nisiyong:

There are a lot of methods that haven't annotate with the `@Override` tag, but they should be.
I use some tools to find all methods that miss the `@Override` tag and fix it. PTAL.

### DIFF
--- a/apm-checkstyle/checkStyle.xml
+++ b/apm-checkstyle/checkStyle.xml
@@ -105,6 +105,7 @@
         <module name="TypeName">
             <property name="format" value="(^[A-Z][a-zA-Z0-9]*$)|(^[A-Z][A-Z0-9]*(_[A-Z0-9]+)*$)"/>
         </module>
+        <module name="MissingOverride"/>
 
         <!--whitespace-->
         <module name="GenericWhitespace"/>
--- a/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/common/AtomicRangeInteger.java
+++ b/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/common/AtomicRangeInteger.java
@@ -54,18 +54,22 @@ public class AtomicRangeInteger extends Number implements Serializable {
         return this.values.get(VALUE_OFFSET);
     }
 
+    @Override
     public int intValue() {
         return this.values.get(VALUE_OFFSET);
     }
 
+    @Override
     public long longValue() {
         return this.values.get(VALUE_OFFSET);
     }
 
+    @Override
     public float floatValue() {
         return this.values.get(VALUE_OFFSET);
     }
 
+    @Override
     public double doubleValue() {
         return this.values.get(VALUE_OFFSET);
     }
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
@@ -176,6 +176,7 @@ public class CorrelationContext {
     /**
      * Clone the context data, work for capture to cross-thread.
      */
+    @Override
     public CorrelationContext clone() {
         final CorrelationContext context = new CorrelationContext();
         context.data.putAll(this.data);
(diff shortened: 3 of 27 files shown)


wu-sheng:
I am wondering, why we need all of them?

wu-sheng:
If this is only related code style, you need to polish the maven code style file too. If there is no mechanism to check in the CI process, this kind of thing could be missed again and again in the future.

nisiyong:
> I am wondering, why we need all of them?

I think it will be helpful when reading source code.

nisiyong:
> If this is only related code style, you need to polish the maven code style file too. If there is no mechanism to check in the CI process, this kind of thing could be missed again and again in the future.

Okay, I will polish the code style file.

nisiyong:
I just have updated codeStyle.xml and checkStyle.xml.

But should aware that the checkStyle.xml has a limitation to verify annotation @Override according to [#MissingOverride](https://checkstyle.sourceforge.io/config_annotation.html#MissingOverride).
> Verifies that the @Override annotation is present when the @inheritDoc javadoc tag is present.

Those missing `@Override` methods mostly has added to source code for a while. Maybe at that time, the IDE didn't show the warning message by default. Now IDE will automate add @Override annotation when you override a method,.

kezhenxu94:
> But should aware that the checkStyle.xml has a limitation to verify annotation @override according to [#MissingOverride](https://checkstyle.sourceforge.io/config_annotation.html#MissingOverride).
> 
> >

This doesn't make much sense actually, we seldom use `{@inheritDoc}` in our code

nisiyong:
> > But should aware that the checkStyle.xml has a limitation to verify annotation @override according to [#MissingOverride](https://checkstyle.sourceforge.io/config_annotation.html#MissingOverride).
> > >
> 
> This doesn't make much sense actually, we seldom use `{@inheritDoc}` in our code

Yes, it is. But there was the only config I found in CheckStyle Reference.

wu-sheng:
> > But should aware that the checkStyle.xml has a limitation to verify annotation @override according to [#MissingOverride](https://checkstyle.sourceforge.io/config_annotation.html#MissingOverride).
> > >
> 
> This doesn't make much sense actually, we seldom use `{@inheritDoc}` in our code

I can not open this sourceforge website, what does it say there? Could you provide a screenshot here?

wu-sheng:
@kezhenxu94 What do you think about this codeStyle change? Personally, I am +0 to this kind of change.

kezhenxu94 (approved):
> @kezhenxu94 What do you think about this codeStyle change? Personally, I am +0 to this kind of change.

I’m +1 to add this, what I meant was we still need to find some other methods to enforce this.

kezhenxu94:
> {@inheritDoc}

It says it only checks this method commented with `/** {@inheritDoc} */` because `checkstyle` is static analyzer, it cannot see a class’s parent class when analyzing, hence doesn’t know whether the method is overriding its parent’s method

wu-sheng:
> I’m +1 to add this, what I meant was we still need to find some other methods to enforce this.

Does current checkStyle change enforce this?

> It says it only checks this method commented with /** {@inheritDoc} */ because checkstyle is static analyzer, it cannot see a class’s parent class when analyzing, hence doesn’t know whether the method is overriding its parent’s method

What does this mean to `@Override`? I read the missing `@Override` could cause CI failure in the style check plugin doc. I am not sure I am following.

kezhenxu94:
> > I’m +1 to add this, what I meant was we still need to find some other methods to enforce this.
> 
> Does current checkStyle change enforce this?

Not really. The added rule will only fail the CI if the method actually overrides its parent’s method, AND is commented (JavaDoc) with `/** {@inheritDoc} */’, this doesn’t make sense at all b/c we seldom use that JavaDoc and if you add that JavaDoc, why will you forget to add `@Override`. 

> 
> > It says it only checks this method commented with /** {@inheritdoc} */ because checkstyle is static analyzer, it cannot see a class’s parent class when analyzing, hence doesn’t know whether the method is overriding its parent’s method
> 
> What does this mean to `@Override`? I read the missing `@Override` could cause CI failure in the style check plugin doc. I am not sure I am following.

See 👆。

I approve the code changes in this PR (add the missing `@Override`), but this rule only take very limited effects, does no harm though.

wu-sheng:
> See 👆。
> 
> I approve the code changes in this PR (add the missing `@Override`), but this rule only take very limited effects, does no harm though. I prefer to investigate some other dedicated tools professional at code analysis (Sonar, CodeQL, etc.) to enforce this.

Agree, I just don't want this kind of PR keeps coming, because even this is correct from coding style perspective, it is not really very useful as nearly 100% people use IDE rather than text.editor to write Java codes.
I have concerned people may use this kind of PR to make them having higher rank in the contributor list(not referring this one), as some people are marketing the rank position in the top level project. I will be more cautious if there is another this kind of PR happens in the future.

PR #6309, opened 2021-02-03
add redefine listener

libinglong:
### <Feature description>

### DIFF
--- a/apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
+++ b/apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
@@ -19,7 +19,9 @@
 package org.apache.skywalking.apm.agent;
 
 import java.lang.instrument.Instrumentation;
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 
 import net.bytebuddy.ByteBuddy;
 import net.bytebuddy.agent.builder.AgentBuilder;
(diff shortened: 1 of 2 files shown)


apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
```diff
@@ -215,4 +218,23 @@ public void onError(final String typeName,
...
+
+        @Override
+        public void onBatch(int index, List<Class<?>> batch, List<Class<?>> types) {
+            /* do nothing */
+        }
+
+        @Override
+        public Iterable<? extends List<Class<?>>> onError(int index, List<Class<?>> batch, Throwable throwable, List<Class<?>> types) {
+            LOGGER.error(throwable, "index={}, batch={}, types={}", index, batch, types);
+            return Collections.emptyList();
```
wu-sheng, line 232: I am not sure about this `onError` API, do we need to return `batch` or empty list?

wu-sheng:
Please update the PR, add `permalink` of https://lists.apache.org/list.html?dev@skywalking.apache.org

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
libinglong: according to the api doc:
> Returns:
A set of classes which should be attempted to be redefined. Typically, this should be a subset of the classes contained in batch but not all classes.

We don't need to attempt to redefine again.
The default listener is net.bytebuddy.agent.builder.AgentBuilder.RedefinitionStrategy.Listener.NoOp.I think adding the error log is enough.

wu-sheng (approved):

PR #6318, opened 2021-02-04
Metrics combination API supports abandoning results

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Related #6183 .

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/Metrics.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/Metrics.java
@@ -56,7 +56,7 @@ public abstract class Metrics extends StreamData implements StorageData {
      * Merge the given metrics instance, these two must be the same metrics type.
      *
      * @param metrics to be merged
-     * @return {@code true} if the combined metrics should be continuously processed, {@code false} it should be abandoned.
+     * @return {@code true} if the combined metrics should be continuously processed. {@code false} means it should be abandoned, and the implementation needs to keep the data unaltered in this case.
      */
     public abstract boolean combine(Metrics metrics);
 
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/Metrics.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/Metrics.java
@@ -56,8 +56,9 @@ public abstract class Metrics extends StreamData implements StorageData {
      * Merge the given metrics instance, these two must be the same metrics type.
      *
      * @param metrics to be merged
+     * @return {@code true} if the combined metrics should be continuously processed, {@code false} it should be abandoned.
      */
-    public abstract void combine(Metrics metrics);
+    public abstract boolean combine(Metrics metrics);
 
     /**
      * Calculate the metrics final value when required.
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
@@ -174,7 +174,10 @@ public class MetricsPersistentWorker extends PersistenceWorker<Metrics> {
                     /*
                      * Merge metrics into cachedMetrics, change only happens inside cachedMetrics.
                      */
-                    cachedMetrics.combine(metrics);
+                    final boolean isAbandoned = !cachedMetrics.combine(metrics);
+                    if (isAbandoned) {
+                        continue;
+                    }
                     cachedMetrics.calculate();
                     prepareRequests.add(metricsDAO.prepareBatchUpdate(model, cachedMetrics));
                     nextWorker(cachedMetrics);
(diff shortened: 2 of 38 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
```diff
@@ -174,21 +175,25 @@ private void flushDataToStorage(List<Metrics> metricsList,
...
                      */
-                    cachedMetrics.combine(metrics);
-                    cachedMetrics.calculate();
-                    prepareRequests.add(metricsDAO.prepareBatchUpdate(model, cachedMetrics));
-                    nextWorker(cachedMetrics);
+                    isAbandoned = !cachedMetrics.combine(metrics);
+                    if (isAbandoned) {
+                        cachedMetrics.calculate();
+                        prepareRequests.add(metricsDAO.prepareBatchUpdate(model, cachedMetrics));
+                        nextWorker(cachedMetrics);
```
wu-sheng, line 182: Should be add `else continue` here, rather than adding `if (!isAbandoned) ` at L190?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
```diff
@@ -174,21 +175,25 @@ private void flushDataToStorage(List<Metrics> metricsList,
                     /*
                      * Merge metrics into cachedMetrics, change only happens inside cachedMetrics.
                      */
-                    cachedMetrics.combine(metrics);
-                    cachedMetrics.calculate();
-                    prepareRequests.add(metricsDAO.prepareBatchUpdate(model, cachedMetrics));
-                    nextWorker(cachedMetrics);
+                    isAbandoned = !cachedMetrics.combine(metrics);
+                    if (isAbandoned) {
```
wu-sheng, line 179:
```suggestion
                    if (!isAbandoned) {
```
Is this reverse?

wu-sheng: Or should we see `if isBandoned; continue`?

wu-sheng: This should make the metrics never updated again, from my reading.

kezhenxu94: > Or should we see `if isBandoned; continue`?

🤦🏻‍♂️ my stupid mistake, fixed

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/Metrics.java
```diff
@@ -56,8 +56,9 @@
      * Merge the given metrics instance, these two must be the same metrics type.
      *
      * @param metrics to be merged
+     * @return {@code true} if the combined metrics should be continuously processed, {@code false} it should be abandoned.
```
wu-sheng, line 59:
```suggestion
     * @return {@code true} if the combined metrics should be continuously processed. {@code false} means it should be abandoned, and keep the data unaltered in this case.
```
I add a little more to the comments, from my code reading, if the change has been made, the data will still flush into the database eventually.

wu-sheng: Also, I feel the data in the cache has refreshed repeatedly into database even without change, could you recheck about this too? If so, we need to fix that in another PR, which should be able to reduce the database load.

wu-sheng (approved):

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/Metrics.java
wu-sheng: > Also, I feel the data in the cache has refreshed repeatedly into database even without change, could you recheck about this too? If so, we need to fix that in another PR, which should be able to reduce the database load.

@kezhenxu94 ⬆️ What do you think about this?

kezhenxu94: Because only limited `Event` data falls into this case, this may be not a big problem for now. Anyway, let me check and think about how to deal with this

wu-sheng: I don't mean the change you made, I mean for all metrics, if the metrics exist in the cache, even no update, they are keeping being updated periodly.

kezhenxu94: > I don't mean the change you made, I mean for all metrics, if the metrics exist in the cache, even no update, they are keeping being updated periodly.

OK I mistakenly thought you only referred to the "abandoned" metrics, it makes sense to me now, I'll check

Issue #5778, opened 2020-11-03
kubernetes-java-client watch time unit error


EvanLjp:
Please answer these questions before submitting your issue.

- Why do you submit this issue?

- [ ] Bug
https://github.com/kubernetes-client/java/pull/1344
in Kubernetes-java-client, the millisecond is used to deliver to the API server.
but the unit in the API server is Second.

PR: https://github.com/apache/skywalking/pull/5777

wu-sheng:
We should fix this after new k8s-java-client released.

wu-sheng:
@EvanLjp I can see the 8.0.2 k8s-client-java released in May only. https://search.maven.org/artifact/io.kubernetes/client-java

It seems like 8.x is not updated, could you take a look, and evaluate, whether do we need to upgrade? @hanahmily Please take a look too.

EvanLjp:
![image](https://user-images.githubusercontent.com/31562192/105953441-236f3a00-60ae-11eb-8d1c-dccbaf3c1b05.png)
I have checked the released package recently and found only fixed in the `10.01` and `11`  version.

wu-sheng:
Yes, clearly, 8.x is not maintained. When you have time, please consider to upgrade this client lib to resolve this issue.

EvanLjp:
ok

wu-sheng:
It is not that urgent, but we should fix this :)

PR #6327, opened 2021-02-05
Add some defensive codes for NPE and bump up Kubernetes client version to expose exception stack trace

kezhenxu94:


kezhenxu94:
The Kubernetes-client swallowed the exception stack, I bump up the version (minimal) to include https://github.com/kubernetes-client/java/pull/1164 , also added some defensive codes for NPE

wu-sheng (commented):
I think you missed dependency check file, and License update

wu-sheng (approved):
LGTM.

@EvanLjp As k8s-client has been upgraded to 10.0. Your #5778 could continue.

kezhenxu94:
> LGTM.
> 
> @EvanLjp As k8s-client has been upgraded to 10.0. Your #5778 could continue.

That's not the case, I've checked that issue and the bug kubernetes-client/java#1344 is only included in 11.0, and there are API changes affected our uses in 11.0 so I just keep it under 11.0, when dealing with #5778 the k8s-client needs another upgrade and some APIs adjustments

PR #6328, opened 2021-02-05
fix bug about memory leak; Detail info @see https://github.com/apache…

gaoweijie:
…/skywalking/issues/6301

- [x] Verify that the fix works.
image
image

these 2 parameters are the case of over-design, which makes things too complex. We agree, we should remove these 2 parameters, and also create span before method, and stop after that. 
1、remove these 2 parameters（close_before_method=true & close_ after_ Method）, and also create span before method, and stop after that. 
2、add  span isActive judgment logic.
3、This will be better when we also upload some common info such as method name and request param info. The actual effect can refer to the verification picture provided above

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<6301>.


FranklinZhang1992 (approved):
This looks good to me!

kezhenxu94:
Duplicate of #6329

PR #6329, opened 2021-02-05
fix bug; Detail info see #6328

gaoweijie:
…/skywalking/issues/6301

- [x] Verify that the fix works.
image
image
image

these 2 parameters are the case of over-design, which makes things too complex. We agree, we should remove these 2 parameters, and also create span before method, and stop after that.
1、remove these 2 parameters（close_before_method=true & close_ after_ Method）, and also create span before method, and stop after that.
2、add span isActive judgment logic.
3、This will be better when we also upload some common info such as method name and request param info. The actual effect can refer to the verification picture provided above

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<6301>.


apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
```diff
@@ -194,4 +194,6 @@
     public static final OfficialComponent APACHE_CXF = new OfficialComponent(105, "Apache-CXF");
 
     public static final OfficialComponent DOLPHIN_SCHEDULER = new OfficialComponent(106, "dolphinscheduler");
+
+    public static final OfficialComponent CUSTOM_ENHANCE = new OfficialComponent(999, "custom-enhance");
```
wu-sheng, line 198: We don't need component id for local span.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/SpanLayer.java
```diff
@@ -19,7 +19,7 @@
 package org.apache.skywalking.apm.agent.core.context.trace;
 
 public enum SpanLayer {
-    DB(1), RPC_FRAMEWORK(2), HTTP(3), MQ(4), CACHE(5);
+    DB(1), RPC_FRAMEWORK(2), HTTP(3), MQ(4), CACHE(5), CUSTOM_ENHANCE(9);
```
wu-sheng, line 22: This is illegal.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/SpanLayer.java
```diff
@@ -50,4 +50,8 @@ public static void asHttp(AbstractSpan span) {
     public static void asMQ(AbstractSpan span) {
         span.setLayer(SpanLayer.MQ);
     }
+
+    public static void asCustomEnhance(AbstractSpan span) {
```
wu-sheng, line 54: Illegal.

apm-sniffer/optional-plugins/customize-enhance-plugin/src/main/java/org/apache/skywalking/apm/plugin/customize/interceptor/BaseInterceptorMethods.java
```diff
@@ -35,54 +38,63 @@
...
+        // collect request info.
+        Tags.CUSTOM.PARAMS.set(span, requestInfo);
+        // collect method info.
+        Tags.CUSTOM.METHOD.set(span, MethodConfiguration.getMethodName(configuration));
     }
 
     void afterMethod(Method method) {
-        if (!MethodConfiguration.isCloseAfterMethod(CustomizeConfiguration.INSTANCE.getConfiguration(method))) {
+        // fix ThreadLocal memory leak bug.
+        // Resolves #6301 https://github.com/apache/skywalking/issues/6301
```
wu-sheng, line 96: Don't put this issue link in the codes

oap-server/server-bootstrap/src/main/resources/component-libraries.yml
```diff
@@ -350,6 +350,10 @@ Apache-CXF:
 dolphinscheduler:
   id: 106
   languages: Java
+# The id of supported Java component library can be increased in order (usually less than 999) and written here.
```
wu-sheng, line 353: Illegal.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/tag/Tags.java
```diff
@@ -98,6 +98,12 @@ private Tags() {
         public static final StringTag HEADERS = new StringTag(14, "http.headers");
     }
 
+    public static final class CUSTOM {
```
wu-sheng, line 101: If you want to add new feature, you need a new PR and issue.

wu-sheng (changes requested):
I think you miss
1. document not update.
2. plugin test cases not update
3. changlog not update

Key point, you should not put unrelated and undiscussed things in one pull reques. Especially, you should not change design without PMC approval.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/SpanLayer.java
gaoweijie: Does this mean that new enumeration values are not allowed to be added,  or must the code of the added enumeration value be incremented（CUSTOM_ENHANCE(6)  ？）

PR #6330, opened 2021-02-05
add debug info when type not match in instrumentation process

libinglong:
### <Feature description>

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/ProtectiveShieldMatcher.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/ProtectiveShieldMatcher.java
@@ -44,7 +44,11 @@ public class ProtectiveShieldMatcher<T> extends ElementMatcher.Junction.Abstract
     @Override
     public boolean matches(T target) {
         try {
-            return this.matcher.matches(target);
+            boolean match = this.matcher.matches(target);
+            if (LOGGER.isDebugEnable()) {
+                LOGGER.debug("type={}, match={}", target, match);
+            }
+            return match;
         } catch (Throwable t) {
             if (LOGGER.isDebugEnable()) {
                 LOGGER.debug(t, "Byte-buddy occurs exception when match type.");
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ Release Notes.
 #### Java Agent
 * Remove invalid mysql configuration in agent.config.
 * Add net.bytebuddy.agent.builder.AgentBuilder.RedefinitionStrategy.Listener to show detail message when redefine errors occur
+* Add debug info when type not match in the instrumentation process.
 
 
 #### OAP-Backend


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/ProtectiveShieldMatcher.java
```diff
@@ -44,7 +44,11 @@ public ProtectiveShieldMatcher(ElementMatcher<? super T> matcher) {
     @Override
     public boolean matches(T target) {
         try {
-            return this.matcher.matches(target);
+            boolean match = this.matcher.matches(target);
+            if (LOGGER.isDebugEnable()) {
+                LOGGER.debug("type={}, match={}", target, match);
+            }
+            return match;
```
wu-sheng, line 51: Why add log here? From my understanding, there will cause a lot of logs in every plugin.

libinglong: We can only log when match is false.

In some scene, the result of match is not as we think. 
Let's see the paragraph in https://github.com/apache/skywalking-website/pull/209/files#
> Now we should decide how to do this. You can use the RunnableWrapper in the maven artifact org.apache.skywalking:apm-toolkit-trace to wrap the param, but you will face another problem. This RunnableWrapper has a plugin whose active condition is checking if there is @TraceCrossThread. Byte buddy in
skywalking will use net.bytebuddy.pool.TypePool.Default.WithLazyResolution.LazyTypeDescription to find the annotations of a class. The LazyTypeDescription finds annotations by using a URLClassLoader with no urls if the classloader is null(bootstrap classloader). So it can not find the @TraceCrossThread class unless you change the LocationStrategy of skywalking java agent builder.

When I inject the RunnableWrapper in the maven artifact org.apache.skywalking:apm-toolkit-trace into the bootstrap classloader. It can not be enhanced properly because the byte buddy can not find it's  @TraceCrossThread annotation.

wu-sheng: This kind of issue will exist always, but we can't add logs in every place, which causes logs unreadable too. This kind of issue could be located by agent debug easily.

wu-sheng:
Principle, we should not add log for helping developing. The logs are for product env debug only.

libinglong:
> Principle, we should not add log for helping developing. The logs are for product env debug only.

I still do not think the log is just for product env. It's a worth as long as it can help any people resolve their problems and save their time. :)

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/match/ProtectiveShieldMatcher.java
libinglong: People usually do not see log line by line. They search.

wu-sheng:
> > Principle, we should not add log for helping developing. The logs are for product env debug only.
> 
> I still do not think the log is just for product env. It's a worth as long as it can help any people resolve their problems and save their time. :) And I think adding logs here is cheap.

Then a project has to face countless PR to add logs, because any line of codes could have potential risk, people could ask to log out anything. Then your system breaks.
All internal systems are being added logs randomly, one way, it is a bad thing, and also, you only face limited developers, so damages are controllable. But in the open source, especially like SkyWalking, we have 400+ code contributors, and more in potential. We cant afford to argue with every one, this log is acceptable, and others are not.
The easy, clear, and affordable principle could be only, you will need this in the runtime, or are facing the requirements.

libinglong:
> Then a project has to face countless PR to add logs, because any line of codes could have potential risk, people could ask to log out anything. Then your system breaks.
> All internal systems are being added logs randomly, one way, it is a bad thing, and also, you only face limited developers, so damages are controllable. But in the open source, especially like SkyWalking, we have 400+ code contributors, and more in potential. We cant afford to argue with every one, this log is acceptable, and others are not.
> The easy, clear, and affordable principle could be only, you will need this in the runtime, or are facing the requirements.

How many contributors is agent contributors of the 400+ code contributors?
I think this log is important because I feel it. Actually I merely log anything in my project. I am not a crazy logger.

> The easy, clear, and affordable principle could be only, you will need this in the runtime, or are facing the requirements.

This is a acceptable reason but still good for me. :)
Anyway, thank you for spending you energy on this pr. :)

wu-sheng:
> How many contributors is agent contributors of the 400+ code contributors?

Over 70% focused on or worked on agent, AFAIK. Agent side clearly has more plugins than the server side, and easier.

> I think this log is important because I feel it. Actually I merely log anything in my project. I am not a crazy logger.

You may not, but how the community should answer the question, when another people want to add logs and quote this PR? How could I prove this is useful than another log? :) It is hard for the community, and hard to understand to the new contributors.

libinglong:
I know where I was wrong.
I should requset a more valuable pr, add this log in passing.

libinglong:
Should we add a troubleshoutting at the plugin develop doc?

wu-sheng:
> Should we add a troubleshoutting at the plugin develop doc?

That depends, how it could be written. It is not easy to provide such kind of documentation. Usually it is a presentation of showcases, but documentation will require it more like a book.

PR #6334, opened 2021-02-06
Support multiple implementations of StorageBuilder in different storage implementations - stage 1.

wu-sheng:

To improve the performance to flush data into the database, and prepare the adoption to the native BanyanDB, I make `StorageBuilder` getting rid of `Map`. 
In the stage 2, I will try to add features of providing override mechanism to builder. Although, this wouldn't be easy, because some metrics's builder are OAL generated, we can't override it manually. But provide a mechanism to support trace and log are much easier.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,7 +17,6 @@ Release Notes.
 * Metrics combination API supports abandoning results.
 * Add a new concept "Event" and its implementations to collect events.
 * Add some defensive codes for NPE and bump up Kubernetes client version to expose exception stack trace.
-* Refactor StorageBuilder to remove the dependency of Map.
 
 #### UI
 Update selector scroller to show in all pages.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,7 @@ Release Notes.
 * Metrics combination API supports abandoning results.
 * Add a new concept "Event" and its implementations to collect events.
 * Add some defensive codes for NPE and bump up Kubernetes client version to expose exception stack trace.
+* Refactor StorageBuilder to remove the dependency of Map.
 
 #### UI
 Update selector scroller to show in all pages.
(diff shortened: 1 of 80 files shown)


hanahmily (approved):
LGTM

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/type/StorageBuilder.java
```diff
@@ -16,18 +16,16 @@
...
+ * Converter between the give T and K.
  *
  * @param <T> A storage entity implementation.
+ * @param <K> The data structure recognized by the storage implementation.
  */
-public interface StorageBuilder<T extends StorageData> {
-
-    T map2Data(Map<String, Object> dbMap);
+public interface StorageBuilder<T, K> {
+    T storage2Entity(K dbMap);
```
kezhenxu94, line 28: Also get rid of the map at interface definition?

```suggestion
    T storage2Entity(K storageData);
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/type/StorageBuilder.java
```diff
@@ -16,18 +16,16 @@
...
 
-    Map<String, Object> data2Map(T storageData);
+    K entity2Storage(T storageData);
```
kezhenxu94, line 30:
```suggestion
    K entity2Storage(T entity);
```

CHANGES.md
```diff
@@ -17,6 +17,7 @@ Release Notes.
 * Metrics combination API supports abandoning results.
 * Add a new concept "Event" and its implementations to collect events.
 * Add some defensive codes for NPE and bump up Kubernetes client version to expose exception stack trace.
+* Refactor StorageBuilder to remove the dependency of Map.
```
kezhenxu94, line 20: I've been thinking whether we should include (nearly) all kinds of changes here, IMO the changes is targeted to our end users, who don't care about our code lever refactor at all, as well as some chores like CI settings, etc.

wu-sheng: I am not sure either. I just want to make this kind of change clear to all developer. Do you want me to remove this?

wu-sheng: Let me remove this.

kezhenxu94 (approved):

kezhenxu94 (approved):

PR #6335, opened 2021-02-06
Update the timestamp field type for LogQuery

arugal:

image
image


test/e2e/e2e-data/src/main/java/org/apache/skywalking/e2e/log/LogMatcher.java
```diff
@@ -70,7 +70,7 @@ public void verify(final Log log) {
             doVerify(getTraceId(), log.getTraceId());
         }
         if (nonNull(getTimestamp())) {
-            doVerify(getTimestamp(), log.getTimestamp());
+            doVerify(getTimestamp(), String.valueOf(log.getTimestamp()));
```
wu-sheng, line 73: Verification doesn't support `long` comparison? I think changing the expected data makes more sense.

wu-sheng:
And please fix CI.

test/e2e/e2e-data/src/main/java/org/apache/skywalking/e2e/log/LogMatcher.java
arugal: > Verification doesn't support `long` comparison?

Has supported.

https://github.com/apache/skywalking/blob/fc23dabf37a85cfc27482dbaf25b0109114f6ef7/test/e2e/e2e-data/src/main/java/org/apache/skywalking/e2e/verification/AbstractMatcher.java#L49-L56

wu-sheng (approved):

PR #6340, opened 2021-02-07
Skip CI in some cases to save resources

kezhenxu94:

We have conditional execution for plugin tests, but for E2E tests, we want to make them "required" for status check while saving resources for some changes like "\*.md" and "skywalking-ui/*", this patch check the changed files of a pull request and set `SKIP_CI` to skip the workflows.

### DIFF
--- a/.github/actions/skip/action.yml
+++ b/.github/actions/skip/action.yml
@@ -1,0 +1,76 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+name: "Set Skip Env Var"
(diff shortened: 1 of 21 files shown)


.github/actions/skip/action.yml
```diff
@@ -0,0 +1,76 @@
...
+#
+name: "Set Skip Env Var"
+description: "Action to set the SKIP_CI environment variable indicating that we should skip CI jobs"
+inputs:
+  paths:
+    description: >-
+      Set the SKIP_CI environment variable when and only when all the changed files located in one of the path,
+      the paths is shell-style pattern.
+    required: false
+    default: >-
```
wu-sheng, line 27: So md and ui submodule changes wouldn't trigger any CI and e2e tasks anymore, right? Do we need to set more required tasks?

kezhenxu94: > So md and ui submodule changes wouldn't trigger any CI and e2e tasks anymore, right?

Yes, effectively, no CI task will be run when only `*.md` and `skywalking-ui` are updated.

> Do we need to set more required tasks?

We can set all E2E tests as required, WDYT?

test/e2e/e2e-service-provider/pom.xml
```diff
@@ -34,7 +34,7 @@
     <artifactId>e2e-service-provider</artifactId>
 
     <properties>
-        <sw.snapshot.version>8.4.0-SNAPSHOT</sw.snapshot.version>
+        <sw.snapshot.version>8.4.0</sw.snapshot.version>
```
wu-sheng, line 37: The variable name is not correct.

.github/actions/skip/action.yml
wu-sheng: > We can set all E2E tests as required, WDYT?

I want to, but meter system tests are not stable from time to time. I am not sure why.

kezhenxu94: > > We can set all E2E tests as required, WDYT?
> 
> I want to, but meter system tests are not stable from time to time. I am not sure why.

Well, let's keep it as is for now

wu-sheng (approved):

Issue #6360, opened 2021-02-10
Improve UT, ZabbixProtocolHandlerTest


wu-sheng:
Right now, `org.apache.skywalking.oap.server.receiver.zabbix.provider.protocol.ZabbixProtocolHandlerTest#testErrorProtocol`  takes 20+ seconds, even it is blocked like this.

![image](https://user-images.githubusercontent.com/5441976/107535291-92df4080-6bfb-11eb-8210-037c315a3f38.png)

I am going to remove this case for now.

wu-sheng:
Update
1. jdk1.8.0_91.jdk FAIL, no response.
2. adoptopenjdk-11.jdk passed, still 22s.

PR #6361, opened 2021-02-10
Temporarily remove `testErrorProtocol` case. 

wu-sheng:
Can't pass on my MacOS 11.2, with 8 core Intel CPU. But when I run this case separately, it passed in 20+ sec.
I am wondering whether test or real codes have faced face condition. @mrproliu Please recheck to reply on #6360


mrproliu (approved):
My fault, I will recode the error protocol UT.

PR #6362, opened 2021-02-10
Simplify the Zabbix UT, reduce use time

### LINKED ISSUE #6360: Improve UT, ZabbixProtocolHandlerTest

mrproliu:
### Fix #6360
- [x] Using mock handler and encoder/decoder to UT. Not using the starting a TCP server to do this.


wu-sheng (approved):
My local tests look like, JDK8 passed too.

PR #6363, opened 2021-02-11
Change the Apdex metric combine calculator

mrproliu:
### Update the Apdex metric combine calculator
- Follow the [Apdex document in The New Relic](https://docs.newrelic.com/docs/apm/new-relic-apm/apdex/apdex-measure-user-satisfaction), I think we need to change the range between Apdex levels.


wu-sheng (approved):

wu-sheng:
@hanahmily Seems so 
image

PR #6248, opened 2021-01-24
Enhancement log-recerive e2e test

arugal:
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>.
- [ ] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md).

PR #6385, opened 2021-02-17
Supplement the E2E test for log reporter

EvanLjp:
Supplement the E2E test for log reporter


wu-sheng:
@arugal Does the previous Log E2E have some log generator? I remember we didn't report logs from the real agents.

wu-sheng (approved):

arugal:
> @arugal Does the previous Log E2E have some log generator? 

From https://github.com/apache/skywalking/pull/6248, the log is generated through the agent

arugal (approved):

PR #6398, opened 2021-02-19
Meter system enhancement

### LINKED ISSUE #6356: Meter system enhancement.
MAL is widely used in different meter/metrics receivers today, we are facing a challenge, which is, when we share the same metrics(name and data type) over different MAL script definition, the `MeterSystem#create` fails.

We need to make `MeterSystem#create` to return `true` w/o failing when `metricsName`/`functionName`/`type` are same.

mrproliu:
### Fix #6356 
- [x] Support create metric with same `metricName` / `function` / `scope`.


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/MeterSystem.java
```diff
@@ -187,6 +187,22 @@ public MeterSystem(final ModuleManager manager) {
...
         final String className = formatName(metricsName);
+
+        /**
+         * Check is already define the metric
+         */
+        try {
+            CtClass alreadyMetric = classPool.get(METER_CLASS_PACKAGE + className);
+            if (alreadyMetric.getSuperclass() != parentClass || type != meterPrototypes.get(metricsName).getScopeType()) {
+                throw new IllegalArgumentException("Already define metric " + metricsName
+                                                       + ", but calculate function or scope type is not same.");
```
kezhenxu94, line 198:
```suggestion
                throw new IllegalArgumentException("Already defined metrics class " + metricsName
                                                       + ", but calculate function or scope type is different.");
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/MeterSystem.java
```diff
@@ -187,6 +187,21 @@ public MeterSystem(final ModuleManager manager) {
             throw new IllegalArgumentException("Function " + functionName + " can't be found by javaassist.");
         }
         final String className = formatName(metricsName);
+
+        /**
+         * Check whether the metrics class is already defined or not
+         */
+        try {
+            CtClass alreadyMetric = classPool.get(METER_CLASS_PACKAGE + className);
+            if (alreadyMetric.getSuperclass() != parentClass || type != meterPrototypes.get(metricsName).getScopeType()) {
```
wu-sheng, line 196: `!=` -> `equals`?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/MeterSystem.java
```diff
@@ -187,6 +187,21 @@ public MeterSystem(final ModuleManager manager) {
             throw new IllegalArgumentException("Function " + functionName + " can't be found by javaassist.");
         }
         final String className = formatName(metricsName);
+
+        /**
+         * Check whether the metrics class is already defined or not
+         */
+        try {
+            CtClass alreadyMetric = classPool.get(METER_CLASS_PACKAGE + className);
```
wu-sheng, line 195:
```suggestion
            CtClass existingMetric = classPool.get(METER_CLASS_PACKAGE + className);
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/MeterSystem.java
```diff
@@ -187,6 +187,21 @@ public MeterSystem(final ModuleManager manager) {
...
+            if (alreadyMetric.getSuperclass() != parentClass || type != meterPrototypes.get(metricsName).getScopeType()) {
+                throw new IllegalArgumentException("Already defined metrics class " + metricsName
+                                                       + ", but calculate function or scope type is different.");
```
wu-sheng, line 198:
```suggestion
                                                       + ", but calculate function or/are scope type is/are different.");
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/MeterSystem.java
```diff
@@ -187,6 +187,21 @@ public MeterSystem(final ModuleManager manager) {
...
+            }
+            log.debug("Metric {} is already define, so skip the metric creation.", metricsName);
```
wu-sheng, line 200:
```suggestion
            log.error("Metric {} is already define, so skip the metric creation.", metricsName);
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/MeterSystem.java
wu-sheng, line 198: log -> Metric class + metricsName + has been defined, but calculate function or/are scope type is/are different.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/MeterSystem.java
mrproliu: Their data type is `Class`, I think we don't need to the use `equals` method?

wu-sheng: Make sense, they share the same classPool, so same class loader. Technically, yes.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/MeterSystem.java
mrproliu: Such as the `OTEL` and `Zabbix` plugin, use the same metric name, I think it's a normal log. Why need to upgrade to the error level?

wu-sheng: I see your point. Keep it in INFO at least, this is important.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/MeterSystem.java
```diff
@@ -187,6 +185,20 @@ public MeterSystem(final ModuleManager manager) {
...
+
+        /**
+         * Check whether the metrics class is already defined or not
+         */
+        try {
+            CtClass existingMetric = classPool.get(METER_CLASS_PACKAGE + className);
+            if (existingMetric.getSuperclass() != parentClass || type != meterPrototypes.get(metricsName).getScopeType()) {
+                throw new IllegalArgumentException(metricsName + " has been defined, but calculate function or/are scope type is/are different.");
+            }
+            log.info("Metric {} is already define, so skip the metric creation.", metricsName);
```
wu-sheng, line 197:
```suggestion
            log.info("Metric {} is already defined, so skip the metric creation.", metricsName);
```

wu-sheng (approved):
Once tests passed, good to me.

PR #6400, opened 2021-02-19
do not direct send after the first aggregation to reduce the network …

whfjam:
Do not direct send after the first aggregation to reduce the network  traffic and reduce the oap-server's load
In my own project i made it a core config , but i don't know how to do it in the new version as my skywalking version is 8.0.1

### Improve the performance of <org.apache.skywalking.oap.server.core.analysis.worker.MetricsAggregateWorker>
- [x] The benchmark result.
```text
In my situation, about 150 applications with 4 oap-server(Mixed mode) nodes(8G16C) on line, and 700G segments one day , it would reduce my network traffic from 200Mbps to 100-150Mbps and reduce the oap-server's load from 7 to 4
```

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsAggregateWorker.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsAggregateWorker.java
@@ -46,6 +46,8 @@ public class MetricsAggregateWorker extends AbstractWorker<Metrics> {
     private final DataCarrier<Metrics> dataCarrier;
     private final MergableBufferedData<Metrics> mergeDataCache;
     private CounterMetrics aggregationCounter;
+    public static final long SEND_AFTER_AGGREGATION_INTERVAL = 300;
+    private long lastSendTime = 0;
 
     MetricsAggregateWorker(ModuleDefineHolder moduleDefineHolder, AbstractWorker<Metrics> nextWorker,
                            String modelName) {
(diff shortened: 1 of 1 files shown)


wu-sheng:
Hi, this solution was discussed years ago, I can't remember where. 

**We don't do this now**, because the data would not be refreshed to 2nd level aggregation, then last dozens packages of metrics would NOT show up anymore until there is new data coming, which could be very confusing, especially when you do baseline testing or local testing.
Think you this way, you are doing a local test, and send 10+ segment to OAP, then nothing happens.

wu-sheng:
If you want to hold data for longer, you have to use some timer mechanism to refresh the cache if there is no incoming data recently. This would not be an easy coding model, and you have to add some kinds of lock, like Persistent Timer doing.

whfjam:
> you

Yes I see, I have not thought about that，still I think it would be a great improvement about performance , I will try to make it work about what you had mentioned.

honganan:
> > you
> 
> Yes I see, I have not thought about that，still I think it would be a great improvement about performance , I will try to make it work about what you had mentioned.

I'm very interested in your solution and I'm facing the same bottleneck. Would you share your code if it has been verified works by PR or else type?

whfjam:
> bottleneck

in my production environment i just change it not to send directly after aggregation，just use interval send such as 500ms， it won't meat that problem because of in production environment the visit will not stop

honganan:
> > bottleneck
> 
> in my production environment i just change it not to send directly after aggregation，just use interval send such as 500ms， it won't meat that problem because of in production environment the visit will not stop

thanks for your idea.

wu-sheng:
As this PR has no update for a long time, I am going to take over the ownership and finish the remaining half.

PR #6403, opened 2021-02-20
Enhance gRPC log appender to allow layout pattern

kezhenxu94:

The gRPC log appender only grasps the users' log message without supporting layout / encoder, which makes no sense in real use cases, this patch enhances the appender to support layout / encoder, the appenders now behave similar to the ConsolerAppender in the corresponding logging frameworks.

- logback

```xml
    <appender name="grpc-log" class="org.apache.skywalking.apm.toolkit.log.logback.v1.x.log.GRPCLogClientAppender">
        <encoder class="ch.qos.logback.core.encoder.LayoutWrappingEncoder">
            <layout class="org.apache.skywalking.apm.toolkit.log.logback.v1.x.mdc.TraceIdMDCPatternLogbackLayout">
                <Pattern>%d{yyyy-MM-dd HH:mm:ss.SSS} [%X{tid}] [%thread] %-5level %logger{36} -%msg%n</Pattern>
            </layout>
        </encoder>
    </appender>
```

- log4j2

```xml
        <GRPCLogClientAppender name="grpc-log">
            <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
        </GRPCLogClientAppender>
```

- log4j

```properties
log4j.appender.CustomAppender=org.apache.skywalking.apm.toolkit.log.log4j.v1.x.log.GRPCLogClientAppender
log4j.appender.CustomAppender.layout=org.apache.log4j.PatternLayout
log4j.appender.CustomAppender.layout.ConversionPattern=[%t] %-5p %c %x - %m%n
```

### DIFF
--- a/.github/actions/e2e-test/action.yml
+++ b/.github/actions/e2e-test/action.yml
@@ -59,7 +59,8 @@ runs:
         echo "::endgroup::"
 
         echo "::group::Run E2E Test ${{ inputs.test_class }}"
-        ./mvnw --batch-mode -f test/e2e/pom.xml -am -DfailIfNoTests=false verify -Dit.test=${{ inputs.test_class }}
+        SW_VERSION=$(./mvnw -q -DforceStdout -N org.apache.maven.plugins:maven-help-plugin:3.2.0:evaluate -Dexpression=project.version)
+        ./mvnw --batch-mode -f test/e2e/pom.xml -am -DfailIfNoTests=false -Dsw.version=${SW_VERSION} verify -Dit.test=${{ inputs.test_class }}
         echo "::endgroup::"
     - name: Report Coverage
       shell: bash
--- a/.github/actions/e2e-test/action.yml
+++ b/.github/actions/e2e-test/action.yml
@@ -32,7 +32,7 @@ runs:
         sudo apt install -y -q xmlstarlet
         SW_VERSION=$(xmlstarlet sel -N pom=http://maven.apache.org/POM/4.0.0 -t -v "/pom:project/pom:properties/pom:sw.version" test/e2e/pom.xml)
         if [[ "$(echo $(echo $SW_VERSION))" != "" ]]; then
-          echo "Please don't submitt the change of sw.version in test/e2e/pom.xml"
+          echo "Please don't submit the change of sw.version in test/e2e/pom.xml"
           exit 1
         fi
         echo "::endgroup::"
--- a/.github/actions/e2e-test/action.yml
+++ b/.github/actions/e2e-test/action.yml
@@ -25,6 +25,17 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Check Unintended Changes
+      shell: bash
+      run: |
+        echo "::group::Check sw.version"
+        sudo apt install -y -q xmlstarlet
+        SW_VERSION=$(xmlstarlet sel -N pom=http://maven.apache.org/POM/4.0.0 -t -v "/pom:project/pom:properties/pom:sw.version" test/e2e/pom.xml)
+        if [[ "$(echo $(echo $SW_VERSION))" != "" ]]; then
+          echo "Please don't submitt the change of sw.version in test/e2e/pom.xml"
+          exit 1
+        fi
+        echo "::endgroup::"
     - name: Check License
       shell: bash
       run: |
--- a/.github/actions/e2e-test/action.yml
+++ b/.github/actions/e2e-test/action.yml
@@ -53,6 +53,11 @@ runs:
     - name: Run E2E Test
       shell: bash
       run: |
+        echo "::group::Install SNAPSHOT apm-application-toolkit"
+        ./mvnw -DskipTests -N install
+        ./mvnw -f apm-application-toolkit -DskipTests -am install
+        echo "::endgroup::"
+
         echo "::group::Run E2E Test ${{ inputs.test_class }}"
         ./mvnw --batch-mode -f test/e2e/pom.xml -am -DfailIfNoTests=false verify -Dit.test=${{ inputs.test_class }}
         echo "::endgroup::"
--- a/test/e2e/pom.xml
+++ b/test/e2e/pom.xml
@@ -44,7 +44,14 @@
     </modules>
 
     <properties>
-        <sw.version></sw.version> <!-- Please manually set the SkyWalking version here to build in the IDE (if you build from command line, just use ./mvnw -Dsw.version=xxx), but make sure not to check it into the code base / git -->
+        <sw.version>
+            <!--
+            Please manually set the SkyWalking version here to build in the IDE
+            (if you build from command line, just use ./mvnw -Dsw.version=x.y.z),
+            but make sure not to check it into the code base / git
+            -->
+            8.5.0-SNAPSHOT
+        </sw.version>
 
         <java.version>1.8</java.version>
         <maven.compiler.source>${java.version}</maven.compiler.source>
--- a/test/e2e/pom.xml
+++ b/test/e2e/pom.xml
@@ -44,7 +44,7 @@
     </modules>
 
     <properties>
-        <sw.version></sw.version> <!-- Please manually set the SkyWalking version here to build in the IDE, but make sure not to check it into the code base / git -->
+        <sw.version></sw.version> <!-- Please manually set the SkyWalking version here to build in the IDE (if you build from command line, just use ./mvnw -Dsw.version=xxx), but make sure not to check it into the code base / git -->
 
         <java.version>1.8</java.version>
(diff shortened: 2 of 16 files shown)


EvanLjp:
we already have the structured logging protocol to collect level, time, or thread, why we also need the pattern layout.

kezhenxu94:
> we already have the structured logging protocol to collect level, time, or thread, why we also need the pattern layout.

`level`, `time`, and `thread`.

Our structured protocol is limited (statically compiled), the users can have many other customized fields, (e.g. using MDC in logback, or other similar to set fields, like request id, user id, response status code, etc.), the later can be used in LAL for more useful analysis.

EvanLjp:
> > we already have the structured logging protocol to collect level, time, or thread, why we also need the pattern layout.
> 
> `level`, `time`, and `thread`.
> 
> Our structured protocol is limited (statically compiled), the users can have many other customized fields, (e.g. using MDC in logback, or other similar to set fields, like request id, user id, response status code, etc.), the later can be used in LAL for more useful analysis.

When the custom field could be changed to TAG in the collection end, it will reduce the computing pressure of the server.

wu-sheng:
This feature makes sense to me.

.github/actions/e2e-test/action.yml
```diff
@@ -53,6 +53,11 @@ runs:
     - name: Run E2E Test
       shell: bash
       run: |
+        echo "::group::Install SNAPSHOT apm-application-toolkit"
+        ./mvnw -DskipTests -N install
+        ./mvnw -f apm-application-toolkit -DskipTests -am install
+        echo "::endgroup::"
```
wu-sheng, line 59: This is not enough, because after the 8.5.0, all related tests will fail again :)

I think once the e2e passed once, we could revert this and changes of e2e case back to normal.

kezhenxu94: > This is not enough, because after the 8.5.0, all related tests will fail again :)
> 
> I think once the e2e passed once, we could revert this and changes of e2e case back to normal.

I consider the property `sw.version` in e2e pom.xml should be updated together when starting the next iteration

wu-sheng: How? Our release process doesn't do this automatically.

kezhenxu94: > How? Our release process doesn't do this automatically.

Automation may be not a restriction IMO because there is only [one place](https://github.com/apache/skywalking/blob/6c6a8a7109c9a5a0dc173d8424ce45dec1da9f9e/test/e2e/e2e-service-provider/pom.xml#L37) to update. Also, forgetting to update it can be discovered (fail the CI) early when PR is opened, instead of failing someday after the newer version is released (we had this before), also we don't bother to release SNAPSHOTs just for testing. WDYT?

kezhenxu94: If automation is the only concern, I may make it into the release automatical process later

wu-sheng: The concern is the process, snapshot->release version is automatically push and tag. Then, we will face the test will fail in official release tag.

kezhenxu94: > Then, we will face the test will fail in official release tag.

Today we are already facing this case, [example](https://github.com/apache/skywalking/pull/6380). I just can see any downside of this PR but benefits I mentioned above

wu-sheng: I know we faced, that is why I don't want it happens again. As we are not real testing this feature in E2E(format I mean), we don't have to risk this again

kezhenxu94: > that is why I don't want it happens again.

There is no guarantee that it won't happened again unless we decided not to do SNAPSHOT releases for testing. Also, using a previous version (8.4.0) in PR tests makes no sense to me, even if we don't test format here, it acts as a regression test that it doesn't break the non-format logs.

wu-sheng: Try this in the root, you will find the thing we are all happy

> printf 'VERSION=${project.version}\n0\n' | mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate | grep '^VERSION'

test/e2e/pom.xml
```diff
@@ -44,6 +44,8 @@
     </modules>
 
     <properties>
+        <sw.version>8.5.0-SNAPSHOT</sw.version>
```
wu-sheng, line 47:
```suggestion
        <sw.version>${env.SKYWALKING_VERSION}</sw.version>
```

How about this? And update the E2E doc a little to match this?

kezhenxu94: > ${env.SKYWALKING_VERSION}

With this, the local IDE won't work unless you manually replace this with the existing version, you cannot even set environment variable in the IDE (see my screenshot below).

image

If the developers have to replace `${env.SKYWALKING_VERSION}` with an actual version, it's highly possible it would be checked into the codebase.

wu-sheng: Strange thing, I read this from https://books.sonatype.com/mvnref-book/reference/resource-filtering-sect-properties.html

kezhenxu94: > Strange thing, I read this from https://books.sonatype.com/mvnref-book/reference/resource-filtering-sect-properties.html

It works from command line, but not in IDEA

wu-sheng: Doesn't this work? 

image

kezhenxu94: > Doesn't this work?
> 
> image

This is running configuration (take effect in Java application or JUnit test case), not in maven's lifecycle, and, the pom.xml takes effect at compile phase.

kezhenxu94: I can adopt `${env.SKYWALKING_VERSION}` and update the doc, but we need to pay attention to PRs that when developers modify `${env.SKYWALKING_VERSION}` to something like `8.5.0-SNAPSHOT`, it's not checked into the codebase.

wu-sheng: > but we need to pay attention to PRs that when developers modify ${env.SKYWALKING_VERSION} to something like 8.5.0-SNAPSHOT

How about we add a step in the CI to make sure the version is `${env.SKYWALKING_VERSION}`(or EMPTY?), before we set it by ourselves.

kezhenxu94: No need to do that, check my latest commit

test/e2e/pom.xml
```diff
@@ -44,6 +44,8 @@
     </modules>
 
     <properties>
+        <sw.version></sw.version> <!-- Please manually set the SkyWalking version here to build in the IDE, but make sure not to check it into the code base / git -->
```
wu-sheng, line 47:
```suggestion
        <sw.version></sw.version> <!-- Please manually set the SkyWalking version here to build in the IDE, such as -Dsw.version=x.y.z, but make sure not to check it into the code base / git -->
```
Is this correct?

kezhenxu94: `-Dsw.version=x.y.z` should be in command line, if in IDE, should just put x.y.z in <sw.version></sw.version> like <sw.version>x.y.z</sw.version>

wu-sheng: Or VM option of the IDE, right? 

I still feel, we should add a check step in the CI, before we run maven.

wu-sheng: And remember to update the e2e doc.

kezhenxu94: > Or VM option of the IDE, right?
> 
> I still feel, we should add a check step in the CI, before we run maven.

Check what? The property is already empty.

wu-sheng: Use shell to read this property before run `./mvnw`, if it is not empty, we should shut down the e2e directly with warning.

.github/actions/e2e-test/action.yml
```diff
@@ -25,6 +25,17 @@ inputs:
...
   using: "composite"
   steps:
+    - name: Check Unintended Changes
+      shell: bash
+      run: |
+        echo "::group::Check sw.version"
+        sudo apt install -y -q xmlstarlet
+        SW_VERSION=$(xmlstarlet sel -N pom=http://maven.apache.org/POM/4.0.0 -t -v "/pom:project/pom:properties/pom:sw.version" test/e2e/pom.xml)
+        if [[ "$(echo $(echo $SW_VERSION))" != "" ]]; then
+          echo "Please don't submitt the change of sw.version in test/e2e/pom.xml"
```
wu-sheng, line 35:
```suggestion
          echo "Please don't submit the change of sw.version in test/e2e/pom.xml"
```

wu-sheng (approved):

wu-sheng (commented):
I think you miss the e2e doc update?

PR #6410, opened 2021-02-20
Fix KafkaContextManagerExtendService boot error of kafka plugin.

zifeihan:
### Fix KafkaContextManagerExtendService boot error of kafka plugin.

Fix https://github.com/apache/skywalking/issues/6406, 
Because when KafkaContextManagerExtendService is used, the ignoreSuffixPatternsWatcher is not initialized by calling prepare of ContextManagerExtendService, which results in a NullPointerException. Don't worry too much, this exception will only affect ignoreSuffixArray not being dynamically updated, and will not affect the trace and other functions.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@ Release Notes.
 * Remove invalid mysql configuration in agent.config.
 * Add net.bytebuddy.agent.builder.AgentBuilder.RedefinitionStrategy.Listener to show detail message when redefine errors occur.
 * Fix ClassCastException of log4j gRPC reporter.
+* Fix KafkaContextManagerExtendService boot error of kafka plugin.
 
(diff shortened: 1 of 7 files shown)


CHANGES.md
```diff
@@ -12,6 +12,7 @@ Release Notes.
 * Remove invalid mysql configuration in agent.config.
 * Add net.bytebuddy.agent.builder.AgentBuilder.RedefinitionStrategy.Listener to show detail message when redefine errors occur.
 * Fix ClassCastException of log4j gRPC reporter.
+* Fix KafkaContextManagerExtendService boot error of kafka plugin.
```
```suggestion
* Fix NPE when Kafka reporter activated.
```

wu-sheng (approved):

PR #6415, opened 2021-02-21
fix agent hostname unknown #4520

### LINKED ISSUE #4520: Enhanced Agent Hostname detection function

- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [X] Feature or performance improvement

We know that the hostname detection method of skywalking-agent is provided by JVM API InetAddress.getLocalhost().GetHostName().

https://github.com/apache/skywalking/blob/69e3d80e24b203bef4d86bfce2ef201a961373b0/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/os/OSUtil.java#L46-L54

However, in some Linux system multiple network card scenarios (for example,container/ Alibaba Cloud ECS), the /etc/hosts file configuration is set in the container. Might look like:

Here is a case where hostname resolution is not assigned

in /etc/hosts,Only the following records:  

```shell
::1     localhost       localhost.localdomain   localhost6      localhost6.localdomain6
127.0.0.1       localhost       localhost.localdomain   localhost4      localhost4.localdomain4
```

In this case, if there is no default bound device, GetHostName() will reverse resolve the host name 127.0.0.1 to "localhost", which is usually not what we expected.

So we may be able to make some improvements and perform a second detection for the case of returning "localhost", for example, from the linux kernel information:

```java
// simple demo .When hostname is localhost, check again.
public static String getHostName() {
    if (HOST_NAME == null) {
        try {
            InetAddress host = InetAddress.getLocalHost();
            HOST_NAME = host.getHostName();
        } catch (UnknownHostException e) {
            HOST_NAME = "unknown";
        }
    }

    if (HOST_NAME == "localhost"){
        try{
            HOST_NAME = new String(Files.readAllBytes(Paths.get("/proc/sys/kernel/hostname")), StandardCharsets.UTF_8);
        } catch (Exception e) {
            HOST_NAME = "unknown";
        }
    }
    return HOST_NAME;
}

```

Maybe things will change. Looking forward to feedback : ) .

BobDu:
fix agent hostname unknown when in aliyun ecs , ....
If no dns reslove hostname.
https://github.com/apache/skywalking/issues/4520

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #4520.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/os/OSUtil.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/os/OSUtil.java
@@ -50,7 +50,7 @@ public class OSUtil {
                 HOST_NAME = host.getHostName();
             } catch (UnknownHostException e) {
                 HOST_NAME = "unknown";
-                String host = uhe.getMessage();
+                String host = e.getMessage();
                 if (host != null) {
                     int colon = host.indexOf(':');
                     if (colon > 0) {
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/os/OSUtil.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/os/OSUtil.java
@@ -50,6 +50,13 @@ public class OSUtil {
                 HOST_NAME = host.getHostName();
             } catch (UnknownHostException e) {
                 HOST_NAME = "unknown";
+                String host = uhe.getMessage();
+                if (host != null) {
+                    int colon = host.indexOf(':');
+                    if (colon > 0) {
+                        HOST_NAME = host.substring(0, colon);
+                    }
+                }
             }
         }
         return HOST_NAME;
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@ Release Notes.
 * Fix ClassCastException of log4j gRPC reporter.
 * Fix NPE when Kafka reporter activated.
 * Enhance gRPC log appender to allow layout pattern.
+* Fix agent get hostname is unknown if no DNS resolve hostname.
 
 #### OAP-Backend
 * Allow user-defined `JAVA_OPTS` in the startup script.


wu-sheng:
Please make your codes following code style, and compiling passed locally.
Then changes.md is required to change too.

BobDu:
Sorry, I reformat my code in locally.
And add change log in changes.md

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/os/OSUtil.java
```diff
@@ -50,6 +50,13 @@ public static String getHostName() {
                 HOST_NAME = host.getHostName();
             } catch (UnknownHostException e) {
                 HOST_NAME = "unknown";
+                String host = e.getMessage();
```
wu-sheng, line 53: According to `UnknownHostException`, `Thrown to indicate that the IP address of a host could not be determined.`.

Could you explain why you need this? Once you set hostname <-> ip address mapping in your OS, you should be fine.

wu-sheng: The original discussion seems very clear about the solution, https://github.com/apache/skywalking/issues/4520#issuecomment-599959270. I am not sure why you are asking for code change? And at the same time, you are reading a hostname based on exception message, clearly, this is not following the JDK codes or spec. You could find many `throw new UnknownHostException(...)` in the JDK source codes w/o ip and `:`.

BobDu: In #4520, say it‘s a special case.
But i Think this following scenarios are many,
only change hostname, like `FooBar-01` in `/etc/hostname`(in linux),
usually, we don't add `192.xxx.xxx.xxx FooBar-01` in `/etc/hosts` or add it in our private dns service.
I aggre you say in #4520, 
but in this situation, agent get hostname is `FooBar-01` better than `unknown`.

wu-sheng: > But i Think this following scenarios are many,
only change hostname, like FooBar-01 in /etc/hostname(in linux),
usually, we don't add 192.xxx.xxx.xxx FooBar-01 in /etc/hosts or add it in our private dns service.

Actually, special case means, this could be done in a standard way. Rather than changing codes. 
And please read the second half of my comments, the way you are using to process the exception message, is really not an official way.

wu-sheng:
No update.

PR #6416, opened 2021-02-21
Fix the segment tag is not synchronized when updating the correlation value.

mrproliu:
- [x] When put the same correlation key, the value will update. In the auto tag case, the segment tag value not override. 

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
@@ -78,6 +78,11 @@ public class CorrelationContext {
         // already contain key
         if (data.containsKey(key)) {
             final String previousValue = data.put(key, value);
+
+            // update tag if need
+            if (AUTO_TAG_KEYS.contains(key) && ContextManager.isActive()) {
+                ContextManager.activeSpan().tag(new StringTag(-1, key, true), value);
+            }
             return Optional.of(previousValue);
         }
 
@@ -86,7 +91,7 @@ public class CorrelationContext {
             return Optional.empty();
         }
         if (AUTO_TAG_KEYS.contains(key) && ContextManager.isActive()) {
-            ContextManager.activeSpan().tag(new StringTag(key), value);
+            ContextManager.activeSpan().tag(new StringTag(-1, key, true), value);
         }
         // setting
         data.put(key, value);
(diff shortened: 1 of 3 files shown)


wu-sheng:
Please fix the conflict.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/CorrelationContext.java
```diff
@@ -78,6 +78,11 @@ public CorrelationContext() {
         // already contain key
         if (data.containsKey(key)) {
             final String previousValue = data.put(key, value);
+
+            // update tag if need
+            if (AUTO_TAG_KEYS.contains(key) && ContextManager.isActive()) {
+                ContextManager.activeSpan().tag(new StringTag(-1, key, true), value);
```
wu-sheng, line 84: Why -1 as key?

wu-sheng: I am not sure about your meaning of update. Transitionally, we don't encourage tag update, if it is really needed.

mrproliu: `-1` is the tag `id`. It uses to compare with the same tag. In the `StringTag` constructor, the id is `-1` default.

wu-sheng: My point is, why override? Overrideable tags are limited. What is the use case?

mrproliu: > I am not sure about your meaning of update. Transitionally, we don't encourage tag update, if it is really needed.

I think the correlation tag value is overridden, the same segment tag should be synchronized.

wu-sheng: Do you mean same tag key? If so, no, it should not override. Most keys in current plugins are not overriding.

mrproliu: I see. This PR will be closed.

wu-sheng: OK, I see what you are missing. CorrelationContext is not relating to or connecting with the span of segment. Instead, CorrelationContext is totally separated with tracing context, but leveraging the context propagation mechanism for assisting the users.
The codes here are for a very specific cases, which mean, there is a context propagated from remote peer, and need to inject into new context(mostly because there is an entry span), then the tags will be set. That is all. No further interaction.

PR #6427, opened 2021-02-23
Add `sum` function in meter system

kezhenxu94:

The MAL didn't implement the `sum` function and the E2E test of the previous #6388 actually falls into the `avg` function, causing all `log_count_info` is `1`, this PR adds the real implementation of `sum` function and adjust the E2E test.

image

### DIFF
--- a/docs/en/concepts-and-designs/mal.md
+++ b/docs/en/concepts-and-designs/mal.md
@@ -177,16 +177,16 @@ MAL should instruct meter-system how to do downsampling for metrics. It doesn't 
 Down sampling operations are as global function in MAL:
 
  - avg
+ - sum
  - latest (TODO)
  - min (TODO)
  - max (TODO)
  - mean (TODO)
- - sum (TODO)
  - count (TODO)
 
 The default one is `avg` if not specific an operation.
 
-If user want get latest time from `last_server_state_sync_time_in_seconds`:
+If user want to get the latest time from `last_server_state_sync_time_in_seconds`:
 
 ```
 latest(last_server_state_sync_time_in_seconds.tagEqual('production', 'catalog'))
--- a/docs/en/concepts-and-designs/mal.md
+++ b/docs/en/concepts-and-designs/mal.md
@@ -178,7 +178,7 @@ Down sampling operations are as global function in MAL:
 
  - avg
  - sum
- - latest (TODO)
+ - latest
  - min (TODO)
  - max (TODO)
  - mean (TODO)
@@ -189,11 +189,7 @@ The default one is `avg` if not specific an operation.
 If user want to get the latest time from `last_server_state_sync_time_in_seconds`:
 
 ```
-latest(last_server_state_sync_time_in_seconds.tagEqual('production', 'catalog'))
-
-or
-
-latest last_server_state_sync_time_in_seconds.tagEqual('production', 'catalog')
+last_server_state_sync_time_in_seconds.tagEqual('production', 'catalog').latest()
 ```
 
 ## Metric level function
--- a/docs/en/concepts-and-designs/mal.md
+++ b/docs/en/concepts-and-designs/mal.md
@@ -174,22 +174,22 @@ Examples:
 MAL should instruct meter-system how to do downsampling for metrics. It doesn't only refer to aggregate raw samples to 
 `minute` level, but also hints data from `minute` to higher levels, for instance, `hour` and `day`. 
 
-Down sampling operations are as global function in MAL:
+Down sampling function is called `downsampling` in MAL, it accepts the following types:
 
- - avg
- - sum
- - latest
- - min (TODO)
- - max (TODO)
- - mean (TODO)
- - count (TODO)
+ - AVG
+ - SUM
+ - LATEST
+ - MIN (TODO)
+ - MAX (TODO)
+ - MEAN (TODO)
+ - COUNT (TODO)
 
-The default one is `avg` if not specific an operation.
+The default type is `AVG`.
 
-If user want to get the latest time from `last_server_state_sync_time_in_seconds`:
+If users want to get the latest time from `last_server_state_sync_time_in_seconds`:
 
 ```
-last_server_state_sync_time_in_seconds.tagEqual('production', 'catalog').latest()
+last_server_state_sync_time_in_seconds.tagEqual('production', 'catalog').downsampling(LATEST)
 ```
 
 ## Metric level function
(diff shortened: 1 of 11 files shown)


oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/SampleFamily.java
```diff
@@ -175,7 +176,10 @@ public SampleFamily min(List<String> by) {
     }
 
     public SampleFamily avg(List<String> by) {
-        ExpressionParsingContext.get().ifPresent(ctx -> ctx.aggregationLabels.addAll(by));
+        ExpressionParsingContext.get().ifPresent(ctx -> {
+            ctx.aggregationLabels.addAll(by);
+            ctx.downsampling = DownsamplingType.AVG;
```
wu-sheng, line 181: AVG seems default in `ExpressionParsingContext#create`. Is this for making codes more clear or functional need?

kezhenxu94: > AVG seems default in `ExpressionParsingContext#create`. Is this for making codes more clear or functional need?

`avg` being default had no problem before since it was only one function, now since I added another one, if the expression contains multiple function, and the `avg` being last one, it may be wrong.  For example, say the expression is `log_count.sum(by: ['service', 'instance']).avg(['level'])`, (this expression may make no sense, just for demonstration), the final metrics should be aggregated by `avg` instead of `sum`.

wu-sheng: Got it. A suggestion, could you use this PR to format `SampleFamily` a little better? Such as put `avg` in this way
```
    public SampleFamily avg(List<String> by) {
        ExpressionParsingContext.get().ifPresent(ctx -> ctx.aggregationLabels.addAll(by));
        if (this == EMPTY) {
            return EMPTY;
        }
        if (by == null) {
            double result = Arrays.stream(samples).mapToDouble(s -> s.value).average().orElse(0.0D);
            return SampleFamily.build(this.context, newSample(ImmutableMap.of(), samples[0].timestamp, result));
        }
        return SampleFamily.build(
            this.context,
            Arrays.stream(samples)
                  .map(sample -> 
                           Tuple.of(
                               by.stream()
                                 .collect(toImmutableMap(
                                     labelKey -> labelKey, 
                                     labelKey -> sample.labels.getOrDefault(labelKey, ""))
                                 ), 
                               sample)
                  )
                  .collect(groupingBy(Tuple2::_1, mapping(Tuple2::_2, toList())))
                  .entrySet().stream()
                  .map(entry -> 
                           newSample(entry.getKey(), 
                                     entry.getValue().get(0).timestamp, 
                                     entry.getValue().stream().mapToDouble(s -> s.value).average().orElse(0.0D)))
                  .toArray(Sample[]::new)
        );
    }
```

wu-sheng: Including several other methods. The codes mixed Tuple, it is hard to read.

docs/en/concepts-and-designs/mal.md
```diff
@@ -177,23 +177,19 @@ MAL should instruct meter-system how to do downsampling for metrics. It doesn't
...
-If user want get latest time from `last_server_state_sync_time_in_seconds`:
+If user want to get the latest time from `last_server_state_sync_time_in_seconds`:
 
 ```
-latest(last_server_state_sync_time_in_seconds.tagEqual('production', 'catalog'))
-
-or
-
-latest last_server_state_sync_time_in_seconds.tagEqual('production', 'catalog')
+last_server_state_sync_time_in_seconds.tagEqual('production', 'catalog').latest()
```
kezhenxu94, line 192: This may be a small breaking change, the form of `latest xxxx` is now NOT supported, instead, one should use `xxx.latest()`. The form of `latest xxx` is error-prone b/c something like

```
latest (foo - 1).tagEqual('bar', '1').sum(['tt']).irate().histogram().histogram_percentile([50,99]).service(['rr'])
```

should be considered the `latest` value of `(foo - 1).tagEqual('bar', '1').sum(['tt']).irate().histogram().histogram_percentile([50,99]).service(['rr'])`, but it's actually the `histogram_percentile` of `latest((foo - 1).tagEqual('bar', '1')).sum(['tt']).irate().histogram()`.

wu-sheng: This looks good to me, @hanahmily any idea?

hanahmily: LGTM. and you have to search our mal relevant rules to adapt to this change.

kezhenxu94: Done, only unit tests used this

wu-sheng (dismissed):

wu-sheng:
Test fails, labels missing
>2021-02-24T12:37:06.7762470Z [ERROR]   ExpressionParsingTest.test:124 
2021-02-24T12:37:06.7764030Z Expected: is <ExpressionParsingContext(samples=[foo], isHistogram=true, percentiles=[50, 99], aggregationLabels=[], scopeLabels=[rr], downsampling=LATEST, scopeType=SERVICE)>
2021-02-24T12:37:06.7766583Z      but: was <ExpressionParsingContext(samples=[foo], isHistogram=true, percentiles=[50, 99], aggregationLabels=[tt], scopeLabels=[rr], downsampling=LATEST, scopeType=SERVICE)>

wu-sheng:
Duplicated adding? The error log seems missing label. What do I miss?

kezhenxu94:
> Duplicated adding? The error log seems missing label. What do I miss?

It WAS missing, after I added them in the tests, it shows up multiple times because I added multi-aggregation-functions cases (e.g. `xxx.avg(['a']).sum(['a'])`, causing `aggregationLabels` being `['a', 'a']`

wu-sheng:
I am just being informed this change will cause a serious bug. The previous `sum`(with avg downsampling) is correct, such as calculating the disk usage, CPU etc. Then, the SampleFamily have 30%(for example), and 35% next time, then we really need 33% usage. `Sum function with avg downsampling` is widely used.

wu-sheng (approved):
This seems very clear to me. Thanks for polishing this.

wu-sheng:
FYI @wankai123 @hanahmily

PR #6432, opened 2021-02-24
Fix apm-dubbo-2.7.x-plugin memory leak due to some Dubbo RpcExceptions.

osiriswd:
### Fix apm-dubbo-2.7.x-plugin memory leak due to some Dubbo RpcExceptions
Dubbo Version: 2.7.5
Skywalking Version: 8.3.0

My Java app ran into an endless FullGC state some days before. After some analysis, I found this.
When Dubbo RPC Chain acts like: A-->dubbo://B-->dubbo://C, and in some situation server C slow down or suddenly crush, Dubbo-thread(Provider) on Server B may triggers an 'Consumer is shutting down' Exception, org.apache.dubbo.rpc.AsyncRpcResult.getAppResponse won't get any result(dubbo code notes: '// This should not happen in normal request process;').
Then result.getException in afterMethod throws exception in Skywalking java agent as below(won't log into span.log):

> ERROR 2021-01-27 15:06:19:290 Dubbo-thread-388 InstMethodsInter : class[class org.apache.dubbo.monitor.support.MonitorFilter] after method[invoke] intercept failure
> org.apache.dubbo.rpc.RpcException: java.util.concurrent.ExecutionException: java.lang.IllegalStateException: Consumer is shutting down and this call is going to be stopped without receiving any result, usually this is called by a slow provider instance or bad service implementation.
  >    at org.apache.dubbo.rpc.AsyncRpcResult.getAppResponse(AsyncRpcResult.java:148)
  >    at org.apache.dubbo.rpc.AsyncRpcResult.getException(AsyncRpcResult.java:107)
  >    at org.apache.skywalking.apm.plugin.asf.dubbo.DubboInterceptor.afterMethod(DubboInterceptor.java:110)
  >    at org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstMethodsInter.intercept(InstMethodsInter.java:97)
  >    at org.apache.dubbo.monitor.support.MonitorFilter.invoke(MonitorFilter.java)
  >    at org.apache.dubbo.rpc.protocol.ProtocolFilterWrapper$1.invoke(ProtocolFilterWrapper.java:81)
  >    at org.apache.dubbo.rpc.protocol.dubbo.filter.FutureFilter.invoke(FutureFilter.java:49)
  >    at org.apache.dubbo.rpc.protocol.ProtocolFilterWrapper$1.invoke(ProtocolFilterWrapper.java:81)

StopSpan() won't work, TraceSegment will be kept by ThreadLocal when dubbo runs in threadpool mode.
While this thread creating new entryspans later, the old TraceSegment will be added into refs.(memory leak)

This is a JVM heapdump after the memory leak happened for several days(the thread is holding 768332 spans):
mat

I thinks maybe there are some 'unnormal' RpcExceptions which can't be obtained by result.getException(). Perhaps we could catch them when calling it from afterMethod.
Built an new dubbo-2.7.x-plugin on my machine and it looks good now.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@ Release Notes.
 * Fix ClassCastException of log4j gRPC reporter.
 * Fix NPE when Kafka reporter activated.
 * Enhance gRPC log appender to allow layout pattern.
+* Fix apm-dubbo-2.7.x-plugin memory leak due to some Dubbo RpcExceptions.
 
 #### OAP-Backend
 * Allow user-defined `JAVA_OPTS` in the startup script.
--- a/apm-sniffer/apm-sdk-plugin/dubbo-2.7.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/asf/dubbo/DubboInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/dubbo-2.7.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/asf/dubbo/DubboInterceptor.java
@@ -24,6 +24,7 @@ import org.apache.dubbo.rpc.Invocation;
 import org.apache.dubbo.rpc.Invoker;
 import org.apache.dubbo.rpc.Result;
 import org.apache.dubbo.rpc.RpcContext;
+import org.apache.dubbo.rpc.RpcException;
 import org.apache.skywalking.apm.agent.core.context.CarrierItem;
 import org.apache.skywalking.apm.agent.core.context.ContextCarrier;
 import org.apache.skywalking.apm.agent.core.context.ContextManager;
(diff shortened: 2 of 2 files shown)


wu-sheng (approved):
According to this, https://github.com/apache/dubbo/blob/5ae160ad7b6b1ec67dad14bce819e5d27353cf21/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AsyncRpcResult.java#L148, yes, the leak could happen.

We could accept this PR, but to be honest, I would say, this is a wrong design from a RPC framework perspective. They provide `hasException` and `getException`, which both of them end in another exception?

Issue #6443, opened 2021-02-25
lettucev5 plugin bug when redis sentinel mode  can't get hosts value


DeadLion:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [√] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___

SkyWalking :8.4.0   
Centos7.0+
java version "1.8.0_191"

lettucev5 plugin bug when redis sentinel mode  can't get hosts value,so ui tolopogy will display null:0 , like below pic
![image](https://user-images.githubusercontent.com/2594907/109130182-65160200-778c-11eb-990e-480fa06882d1.png)

because sentinel mode the server nodes info will save in sentinels field, below is debug info, it's clear to understand why this bug happen.
![image](https://user-images.githubusercontent.com/2594907/109130915-3b110f80-778d-11eb-92cd-65936b8fb3ca.png)

i already fixed this , i will make a pr.

wu-sheng:
Yes, please.

PR #6453, opened 2021-02-26
Bugfix/lettucev5 plugin

DeadLion:
### Fix lettuce-5.x-plugin bug  <https://github.com/apache/skywalking/issues/6443>
noraml we could get redis node info from redisURI.host and redisURI.port ,but when redis cluster mode is sentinel ,we need to get info from redisURI.sentinels. so  i add a condition  when sentinelMasterId is not blank,splicing sentinels node info,  finally nodeInfoString would be like masterId[ip:port,ip1:port], otherwise same with before.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/lettuce-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/lettuce/v5/RedisClientConstructorInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/lettuce-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/lettuce/v5/RedisClientConstructorInterceptor.java
@@ -22,6 +22,9 @@ import io.lettuce.core.RedisClient;
 import io.lettuce.core.RedisURI;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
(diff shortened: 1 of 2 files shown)


apm-sniffer/apm-sdk-plugin/lettuce-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/lettuce/v5/RedisClientConstructorInterceptor.java
```diff
@@ -30,6 +33,14 @@ public void onConstruct(EnhancedInstance objInst, Object[] allArguments) {
         RedisURI redisURI = (RedisURI) allArguments[1];
         RedisClient redisClient = (RedisClient) objInst;
         EnhancedInstance optionsInst = (EnhancedInstance) redisClient.getOptions();
-        optionsInst.setSkyWalkingDynamicField(redisURI.getHost() + ":" + redisURI.getPort());
+        StringBuilder redisInfoSb = new StringBuilder();
```
wu-sheng, line 36:
```suggestion
        StringBuilder redisPeer = new StringBuilder();
```
You should make the variable more clear.

DeadLion: done

wu-sheng (approved):
LGTM

PR #6467, opened 2021-02-27
Refine Compatible-with-other-javaagent-bytecode-processing.md and rename to Compatibility-with-other-Java-agent-bytecode-processes.md

wingwong-knh:


wu-sheng (commented):
Hi, Wing, I think you accidentally removed the `.md` suffix.

wingwong-knh:
> Hi, Wing, I think you accidentally removed the `.md` suffix.

@wu-sheng Oops sorry! Should I submit a new pull request on this edit?

kezhenxu94:
> > Hi, Wing, I think you accidentally removed the `.md` suffix.
> 
> @wu-sheng Oops sorry! Should I submit a new pull request on this edit?

@wingwong-knh you can edit in this branch directly, [here](https://github.com/wingwong-knh/skywalking/edit/patch-1/docs/en/FAQ/Compatibility-with-other-Java-agent-bytecode-processes), 

image

kezhenxu94 (approved):

kezhenxu94:
File `./docs/en/setup/service-agent/java-agent/README.md` contains a reference to this file, which is renamed, can you please also correct that link in the aforementioned file?

wu-sheng:
Hi, if there is no big grammar error in the file name, we should keep the old one. Because as github URL and website URL is widely used in other blocks, our changes could cause 404.

wingwong-knh:
> Hi, if there is no big grammar error in the file name, we should keep the old one. Because as github URL and website URL is widely used in other blocks, our changes could cause 404.

@wu-sheng  Ah OK, I agree. I think it's fine to keep the file name as it is. Let's revert this edit?

wu-sheng:
> > Hi, if there is no big grammar error in the file name, we should keep the old one. Because as github URL and website URL is widely used in other blocks, our changes could cause 404.
> 
> @wu-sheng Ah OK, I agree. I think it's fine to keep the file name as it is. Let's revert this edit?

Make sense, let me keep it. And keep the old name could make reviewers more clear about what you changed. Otherwise, we see this, purely add a new file.
image

If we should rename a file name in the future, let's use a separated PR to change the file name only.

wu-sheng (approved):

PR #6470, opened 2021-02-28
Fix `DatabaseSlowStatementBuilder` statement maybe null

### LINKED ISSUE #6455: Database Slow Statements return error:Cannot return null for non-nullable type: 'String' within parent 'SelectedRecord' (/readSampledRecords[0]/name)

- [ ] Question or discussion
- [x] Bug
- [ ] Requirement

___

___
### Bug
   SkyWalking:8.3.0, OS:Centos, java:1.8.0_221
- What happened?
  The error only returned by ES database,Mysql is ok. Is there any additional configuration I need to add? 
```json
{
	"errors": [{
		"message": "Cannot return null for non-nullable type: 'String' within parent 'SelectedRecord' (/readSampledRecords[0]/name)"
	}, {
		"message": "Cannot return null for non-nullable type: 'String' within parent 'SelectedRecord' (/readSampledRecords[1]/name)"
	}]
}
```

haoyann:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6455.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -37,6 +37,7 @@ Release Notes.
 * Remove the experimental Zipkin span analyzer.
 * Upgrade the Zipkin Elasticsearch storage from 6 to 7.
 * Require Zipkin receiver must work with `zipkin-elasticsearch7` storage option.
+* Fix `DatabaseSlowStatementBuilder` statement maybe null.
 
 #### UI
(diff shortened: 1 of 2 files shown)


wu-sheng (approved):
LGTM. cc @daiqianshi

PR #6477, opened 2021-03-03
Enhance the LAL to allow easily skipping logs with malformed formats

kezhenxu94:
The `json` / `yaml` parsers will benefit from option `abortOnFailure` in usage:

```groovy
json {
  abortOnFailure true
  // ... we will have other options here
}
```

instead of (imaginary, not implemented actually)

```groovy
if (!json({
  // ... we will have other options here
})) {
  abort {}
}
```

> P.S. 8.5.0 is the first version introducing LAL so I didn't update the change log file about this enhancement.

### DIFF
--- a/docs/en/concepts-and-designs/lal.md
+++ b/docs/en/concepts-and-designs/lal.md
@@ -64,6 +64,7 @@ See examples below.
 ```groovy
 filter {
     json {
+        abortOnFailure true // this is optional because it's default behaviour
     }
 }
 ```
@@ -73,6 +74,7 @@ filter {
 ```groovy
 filter {
     yaml {
+        abortOnFailure true // this is optional because it's default behaviour
     }
 }
 ```
@@ -90,6 +92,7 @@ all the captured groups can be used later in the extractors or sinks.
 ```groovy
 filter {
     text {
+        abortOnFailure true // this is optional because it's default behaviour
         // this is just a demo pattern
         regexp "(?<timestamp>\\d{8}) (?<thread>\\w+) (?<level>\\w+) (?<traceId>\\w+) (?<msg>.+)"
     }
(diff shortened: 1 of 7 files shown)


docs/en/concepts-and-designs/lal.md
```diff
@@ -55,15 +54,35 @@ types of parsers at the moment, namely `json`, `yaml`, and `text`.
...
 Property `parsed` is typically a map, containing all the fields parsed from the raw logs, for example, if the parser
 is `json` / `yaml`, `parsed` is a map containing all the key-values in the `json` / `yaml`, if the parser is `text`
-, `parsed` is a map containing all the captured groups and their values (for `regexp` and `grok`). See examples below.
+, `parsed` is a map containing all the captured groups and their values (for `regexp` and `grok`).
+
+All parsers share the following options:
+
+| Option | Type | Description | Default Value |
+| ------ | ---- | ----------- | ------------- |
+| `abortOnFailure` | `boolean` | Whether the filter chain should abort if the parser failed to parse / match the logs | `false` |
```
wu-sheng, line 63:
```suggestion
| `abortOnFailure` | `boolean` | Whether the filter chain should abort if the parser failed to parse / match the logs | `true` |
```

Any reason doesn't use `true` as default?

docs/en/concepts-and-designs/lal.md
```diff
@@ -69,7 +64,6 @@ See examples below.
 ```groovy
 filter {
     json {
-        abortOnFailure true
```
wu-sheng: I think you still should add these docs, but committed and say, this is default behavior.

wu-sheng (dismissed):

wu-sheng (approved):

PR #6478, opened 2021-03-03
Update license

Fine0830:

Sync ui(Update echarts version to 5.0.2) and update license.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -49,6 +49,7 @@ Release Notes.
 * Search endpoints with keywords.
 * Fix pageSize on logs page.
 * Update echarts version to 5.0.2.
+
 #### Documentation
 * Polish documentation due to we have covered all tracing, logging, and metrics fields.
 * Adjust documentation about Zipkin receiver.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,7 +48,7 @@ Release Notes.
 * Fix trace id by clear search conditions.
 * Search endpoints with keywords.
 * Fix pageSize on logs page.
-* update echarts version to 5.0.2.
+* Update echarts version to 5.0.2.
 #### Documentation
 * Polish documentation due to we have covered all tracing, logging, and metrics fields.
 * Adjust documentation about Zipkin receiver.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,7 +48,7 @@ Release Notes.
 * Fix trace id by clear search conditions.
 * Search endpoints with keywords.
 * Fix pageSize on logs page.
-
+* update echarts version to 5.0.2.
 #### Documentation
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -48,7 +48,7 @@ Release Notes.
 * Fix trace id by clear search conditions.
 * Search endpoints with keywords.
 * Fix pageSize on logs page.
-
+* Update echarts version to 5.0.2.
```
wu-sheng, line 51:
```suggestion
* Update echarts version to 5.0.2.

```

wu-sheng (dismissed):

dist-material/release-docs/LICENSE
```diff
@@ -498,8 +499,8 @@ d3-tip          0.9.1:  https://github.com/Caged/d3-tip BSD-3-Clause
 d3-transition   1.1.3:  https://github.com/d3/d3-transition     BSD-3-Clause
 d3-voronoi      1.1.4:  https://github.com/d3/d3-voronoi        BSD-3-Clause
 d3-zoom 1.7.3:  https://github.com/d3/d3-zoom   BSD-3-Clause
-zrender 4.0.4:  https://github.com/ecomfe/zrender   BSD-3-Clause
-vue-i18n 8.10.0:  https://github.com/kazupon/vue-i18n   MIT
+zrender 5.0.4:  https://github.com/ecomfe/zrender   BSD-3-Clause
+tslib 2.0.3 https://github.com/libts/tslib LGPL-2.1
```
kezhenxu94, line 503: IMO the real "tslib" should be https://github.com/Microsoft/tslib instead of https://github.com/libts/tslib, also, LGPL-2.1 is category X in Apache projects, please confirm

kezhenxu94 (changes requested):

dist-material/release-docs/LICENSE
kezhenxu94: The latest version of https://github.com/libts/tslib is 1.22, far away from 2.0.3

wu-sheng: Yes, I checked, repo actually is https://www.npmjs.com/package/tslib.

wu-sheng: According to this, it is BSD Zero(https://github.com/microsoft/tslib/blob/master/LICENSE.txt), which belongs to Catalog A. We are good to go. Just need to update the license file. Thanks for catching this, otherwise it will be super confused we included LGPL-2

wu-sheng: @Fine0830 Please be advised, GPL, LGPL are not allowed to be used in the ASF. Please be careful when you add new dependency and edit LICENSE file.

wu-sheng:
@kezhenxu94 I updated, please recheck.

kezhenxu94 (approved):

PR #6479, opened 2021-03-03
Support k8s monitoring

### LINKED ISSUE #6289: Monitor Pod(k8s) performance
We have finished the VM monitoring in the 8.4.0, now, we could move forward to the pod. @wankai123 will continue to work on this field to get the performance data for the pods.

Once either #6260 or #6209, we have chance to ship pods' logs to the SkyWalking too.

wankai123:
### Support k8s cluster monitoring

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6289.

Defined the k8s-cluster as a service in OAP, use k8s-cluster:: as a prefix in MAL to identify.
Defined the k8s-node as an instance in OAP,  the name is k8s node name.
Defined the k8s-service as an endpoint in OAP,  the name is `$serviceName.$namespace`.

### Metrics flow
cAdvisor & K8s kube-state-metrics -> OpenTelemetry Collector --OC format--> SkyWalking OAP Server
### Monitored metrics
**Cluster:**
- Node Total
- Namespace Total
- Deployment Total
- Service Total
- Pod Total
- Container Total
- CPU Resources
- Memory Resources
- Storage Resources
- Node Status
- Deployment Status
- Deployment Spec Replicas
- Service Status
- Pod Status Not Running
- Pod Status Waiting
- Pod Status Terminated

**Node:**
- Pod Total
- Node Status
- CPU Resources
- Memory Resources
- Storage Resources
- CPU Usage
- Memory Usage
- Network I/O

**Service**
- Service Pod Total
- Service Pod Status
- Service CPU Resources
- Service Memory Resources
- Pod CPU Usage
- Pod Memory Usage
- Pod Waiting
- Pod Terminated
- Pod Restarts
- Pod Network Receive
- Pod Network Transmit
- Pod Storage Usage

k8s server version: v1.18.12-gke.1210
kube-state-metrics version: 2.0.0-beta
### Screenshots
image
image
image


wu-sheng:
Suggestions 
1. The graph layout should be from larger concept to smaller
2. Separate current panel into two panels, <1> is the cluster and nodes status, which basically the brief and physical perspective, <2> Service, Deployment and Pod, which are related to your images and code packages. WDYT?
3. If you want to show the avg number, then choose some of them on the top of the panel. Current number of nodes and pods showing up seems strange.

wankai123:
> Suggestions
> 
> 1. The graph layout should be from larger concept to smaller
> 2. Separate current panel into two panels, <1> is the cluster and nodes status, which basically the brief and physical perspective, <2> Service, Deployment and Pod, which are related to your images and code packages. WDYT?
> 3. If you want to show the avg number, then choose some of them on the top of the panel. Current number of nodes and pods showing up seems strange.

1. Yes, the order is Cluster -> Nodes -> Namespace -> Departments -> Services -> Containers(image) -> Pods
2. My idea is the `Cluster` panel show the global info, from Cluster to Pods, used for quick monitor, and the Cluster metric which scraped can't provide more details. So I'm working on another panel `Node` that will be add as the second panel, the  `Node` defined as the Cluster's instance, it is related to Cluster and can be selected to show more details. These metrics are scraped from the other collector but separated by nodes.
3. I just put them in the same level and group, and the order is the count of the entity -> the status of the entity.
@wu-sheng

oap-server/server-bootstrap/src/main/resources/otel-oc-rules/k8s-cluster.yaml
```diff
@@ -0,0 +1,90 @@
...
+  - name: node_count
+    exp: kube_node_info.sum(['cluster'])
+  - name: node_status
+    exp: kube_node_status_condition.tagMatch('status' , 'true|unknown').sum(['cluster' , 'node' ,'condition'])
+
+  - name: namespace_count
+    exp: kube_namespace_labels.sum(['cluster' , 'namespace'])
+
+  - name: deployment_count
+    exp: kube_deployment_labels.sum(['cluster' , 'deployment'])
```
kezhenxu94, line 68: These `*_count` (`.sum`) need `.downsampling(SUM)` now?

wu-sheng:
I discussed with @wankai123 directly, it seems we need to enhance the `endpoint` function in the SampleFamily to support extract multiple records from one SampleFamily, due to how k8s metrics work.
We will feedback  more details here.

oap-server/server-bootstrap/src/main/resources/otel-oc-rules/k8s-cluster.yaml
wankai123: No, the metrics provides the total count each time, so it's still `.downsampling(AVG)` ：） like:
`
kube_namespace_labels{namespace="kube-system"} 1
kube_namespace_labels{namespace="monitoring"} 1
kube_namespace_labels{namespace="bar"} 1
kube_namespace_labels{namespace="default"} 1
kube_namespace_labels{namespace="foo"} 1
kube_namespace_labels{namespace="kube-node-lease"} 1
kube_namespace_labels{namespace="kube-public"} 1
`

wankai123:
PR updated. @wu-sheng @hanahmily @kezhenxu94

wu-sheng:
Why node status is a bar graph?

And could you fill the missing units? Such as CPU usage, should 100 or 1 be expected as 100%?

wankai123:
> **Why node status is a bar graph?**
Status is a label value，use bar graph can see how many nodes in there, and if status was changed we could see different colors.  If issue #6569 finished, we could have another choice to show status more directly.
image

> **And could you fill the missing units? Such as CPU usage, should 100 or 1 be expected as 100%?**
In k8s, the CPU usually used cores as the units, such as `1 core, 200m core(about 0.2core )`, I think we can discuss it.

wankai123:
Added `Cores` as the units of CPU usage and resources.

hanahmily:
> > **Why node status is a bar graph?**
> > Status is a label value，use bar graph can see how many nodes in there, and if status was changed we could see different colors.  If issue #6569 finished, we could have another choice to show status more directly.
> > image
> 
> > **And could you fill the missing units? Such as CPU usage, should 100 or 1 be expected as 100%?**
> > In k8s, the CPU usually used cores as the units, such as `1 core, 200m core(about 0.2core )`, I think we can discuss it.
> > @hanahmily @wu-sheng

For CPU request and limit metrics, the unit should be `m` which represents an absolute quantity, never as a relative quantity; 0.1 is the same amount of CPU on a single-core, dual-core, or 48-core machine. And I prefer to use `m` due to avoiding floating-point numbers. Any number which is finer than 1m, for example, 0.1m, will be rounded up to 1m under the hood.

If we want a percentage of the requested CPU, `cpu_cores_requests/cpu_cores`could get the desired result.

wankai123:
Add `m` as the unit of the CPU resources. Updated screenshots.

wu-sheng:
Is this waiting for @Fine0830 next UI version?

wankai123:
> Is this waiting for @Fine0830 next UI version?

yes，if it‘s possible

wankai123:
Update: In service-related meters, filter the samples in which the pod can't match any service.

hanahmily (approved):
LGTM

wankai123:
PR updated, included the table panel. @wu-sheng

wu-sheng (approved):

PR #6493, opened 2021-03-05
Support foreground startup，Default daemon mode.

Snile826:

Support foreground startup，Default daemon mode.


Snile826:
Foreground startup helps containerization.

kezhenxu94:
We have provided the official docker images, and, we have dedicated startup script for containerization, those scripts (for example, `dist-material/bin/oapService.sh`) are not meant to be used in containers

wu-sheng:
> We have provided [official docker images](https://hub.docker.com/search?q=skywalking&type=image), and, we have [dedicated startup script for containerization](https://github.com/apache/skywalking/blob/master/docker/oap/docker-entrypoint.sh), those scripts (for example, `dist-material/bin/oapService.sh`) are not meant to be used in containers

Agree with @kezhenxu94

Issue #6411, opened 2021-02-20
[Core Feature] Merge metrics indexes of ElasticSearch based on MetricsFunction/MeterFunction annotation


wu-sheng:
Hi

This is a core feature/improvement proposal. Currently, this is a theoretically solution.

## Background
SkyWalking today has a huge extendibility in log/metric/trace analysis. So, many metrics are being or will be added, rather than the old days, we rely on the trace analysis mostly.
Then, right now, every metric of SkyWalking generated logically maps to one index of ElasticSearch. Even we have merged day and hour level into minute, and provided dayStep to reduce the number of indexes. Still, the number is huge, and most importantly, it is increasing.
Also, at the same time, the size of metric index is much less than the segments, such as
```
skywalking-online_service_cpm-20210213                                 FB8Yym7aTDqgAgpYJrG_NQ  1 1    941083    2612 191.8mb  95.8mb 
skywalking-online_endpoint_cpm-20210130                                V8SZg-1qSO6mhvA0m7cs9A  1 1   8777005   36984   3.5gb   1.7gb
skywalking-online_service_instance_relation_client_cpm-20210123        Qb-cxpTGS7-Qpz02n1T8LQ  1 1  95699175  563518  60.1gb    30gb
skywalking-online_segment-20210218                                     5aOr9YGeTq2_xfOPypCQUQ 30 0 353400759       0 389.3gb 389.3gb
```

## Proposal
If we merge `service_cpm`, `endpoint_cpm` and `service_instance_relation_client_cpm` into an index, let's call it, `cpm-*`, the performance impact of query is very limited, because basically the size of data set still ~30GB(60GB because of one replication). 
To separate these metrics for the SkyWalking query, we need to add metric name in the `ROWID` and as a separated column, which is also not a big issue. The only side-effect is causing 10-20 chars as ROWID.
Then, consider this grouping mechanism is 100% automatically, so users would feel confused.

At last, speaking from ElasticSearch storage perspective, as the metric/metric function sharing, most columns are still same. But as we need `StorageEsInstaller#createMapping`, we need the unison set for all possible columns of all metrics in one kind. In the previous case, endpoint_cpm has one or two more columns.
As a result, we need to change `createTable(Model model)` to `createTable(List<Model> models)`, for providing the storage implementation a chance to merge logic model than creating.

@apache/skywalking-committers I am willing to hear your feedback. I talked with @EvanLjp privately, and think this is worth to try.

PR #6499, opened 2021-03-06
[Incompatible Enhancement]New index policy of ElasticSearch storage option

EvanLjp:
https://github.com/apache/skywalking/issues/6411


docs/en/setup/backend/backend-storage.md
```diff
@@ -72,6 +72,8 @@ storage:
     oapAnalyzer: ${SW_STORAGE_ES_OAP_ANALYZER:"{\"analyzer\":{\"oap_analyzer\":{\"type\":\"stop\"}}}"} # the oap analyzer.
     oapLogAnalyzer: ${SW_STORAGE_ES_OAP_LOG_ANALYZER:"{\"analyzer\":{\"oap_log_analyzer\":{\"type\":\"standard\"}}}"} # the oap log analyzer. It could be customized by the ES analyzer configuration to support more language log formats, such as Chinese log, Japanese log and etc.
     advanced: ${SW_STORAGE_ES_ADVANCED:""}
+    storageMode: ${SW_STORAGE_ES_MODE:FUNCTIONAL_AGGREGATION}#There are 2 mode for indices storage, which are FUNCTIONAL_AGGREGATION and NO_AGGREGATION. The Storage behavior using NO_AGGREGATION is same with the storage behavior before 8.5.0 version. And, the Storage behavior using FUNCTIONAL_AGGREGATION means the telemetry data, that generated by OAL or MAL, would be partitioned by the functions of the OAL or MAL.
```
wu-sheng, line 75: I think we could do permanent change rather than 2 modes. This new one is always better and recommended. Right?

EvanLjp: We users should expect the storage to be compatible, giving them time to migrate to the new scheme.

wu-sheng: Actually, this wouldn't help. Because OAP can't write data in 2 kinds of storage implementations. At any point the switch happens, data lost forever.

wu-sheng: And you will make end users confused about choose what solution is better. They don't read codes.

EvanLjp: Or we can add some notice comments here to show the new schema is better than now. Or remove history storage mode in 9.0.0.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StoragePartitioner.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.skywalking.apm.util.StringUtil;
+import org.apache.skywalking.oap.server.core.storage.model.Model;
+
+@Slf4j
+public enum StoragePartitioner {
+    /**
+     * The telemetry data, that generated by OAL or MAL, would be partitioned to storage by the functions of the OAL or
+     * MAL. And, the other record data would be insulated storage by themselves definitions.
```
wu-sheng, line 34: All data in SkyWalking is telemetry data. Don't use this words. INSTANCE should not have comments. All comments should on the class level.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StoragePartitioner.java
```diff
@@ -0,0 +1,104 @@
...
+     */
+    INSTANCE;
+    private final Base64.Encoder encoder = Base64.getEncoder();
+    public static final String LOGIC_TABLE_NAME = "logic_table";
+
+    /**
+     * The relations of the logic table and the physical table.
+     */
+    private final Map<String, String> tableRelations = new ConcurrentHashMap<>();
```
wu-sheng, line 43: `ConcurrentHashMap`? Is there a race condition?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StoragePartitioner.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * BASE64 encode the original logic table name.
+     */
+    public String getLogicTableColumnVal(String logicTableName) {
+        return encoder.encodeToString(logicTableName.getBytes(StandardCharsets.UTF_8));
+    }
+
+    public String getPhysicialTableName(String logicName) {
```
wu-sheng, line 97: `Physicial` typo.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -38,27 +41,50 @@
...
 public class StorageEsInstaller extends ModelInstaller {
     private final Gson gson = new Gson();
-
     private final StorageModuleElasticsearchConfig config;
     protected final ColumnTypeEsMapping columnTypeEsMapping;
 
+    /**
+     * The relation is between physical table and the specific structure definition.
+     */
+    private final ConcurrentHashMap<String, Map<String, Object>> tables;
```
wu-sheng, line 50: I think should use `StoragePartitioner` to hold, and put a class called `PhysicalIndex` the value of `StoragePartitioner#tableRelations`. 
According to your code style, you should isolate these two.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -38,27 +41,50 @@
...
+
     public StorageEsInstaller(Client client,
                               ModuleManager moduleManager,
-                              final StorageModuleElasticsearchConfig config) {
+                              StorageModuleElasticsearchConfig config) throws StorageException {
         super(client, moduleManager);
         this.columnTypeEsMapping = new ColumnTypeEsMapping();
         this.config = config;
+        this.tables = new ConcurrentHashMap<>(loadHistoryTables(client));
```
wu-sheng, line 58: I don't recommend to do this. Reading all template impact the user side **security policy**.
And according to existing codes, if this template and index exist, we will treat them as a correct version.

wu-sheng: You should read template settings when you face the creation event only.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -100,6 +139,48 @@ protected void createTable(Model model) throws StorageException {
...
+        if (Objects.isNull(mapping) || mapping.size() == 0) {
+            return new HashMap<>();
+        }
+        return (Map<String, Object>) ((Map<String, Object>) mapping.get(ElasticSearchClient.TYPE)).get("properties");
+    }
+
+    /**
+     * Whether the tables contains the input mapping with the same table name key.
+     */
+    private boolean containsTemplateMapping(String tableName, Map<String, Object> mapping) {
```
wu-sheng, line 175: `containsTemplateMapping`->`isTemplateMappingCompatible`

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -68,26 +94,39 @@ protected boolean isExists(Model model) throws StorageException {
...
-                    log.info(
-                        "create {} index template finished, isAcknowledged: {}", model.getName(), isAcknowledged);
+            if (model.isTimeSeries()) {
+                Map<String, Object> template = esClient.getTemplate(tableName);
+                if (!template.isEmpty()) {
+                    appendTemplateMapping(tableName, (Map<String, Object>) template.get("mappings"));
+                }
+                if (!containsTemplateMapping(tableName, mapping)) {
+                    Map<String, Object> templateMapping = appendTemplateMapping(tableName, mapping);
+                    boolean isAcknowledged = esClient.createTemplate(tableName, settings, templateMapping);
```
wu-sheng, line 114: Are you meaning we could recreate a template w/o deleting the old one?

wu-sheng: You need to polish this new creating logic, it is not clear to me.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -68,26 +94,39 @@ protected boolean isExists(Model model) throws StorageException {
...
-                        throw new StorageException("create " + model.getName() + " index template failure, ");
+                        throw new StorageException("create " + tableName + " index template failure, ");
                     }
                 }
                 indexName = TimeSeriesUtils.latestWriteIndexName(model);
+            } else {
+                indexName = tableName;
+            }
+
+            if (esClient.isExistsIndex(indexName) && esClient.getDocNumber(indexName) == 0) {
```
wu-sheng, line 125: Why adding this? How does this relate to this PR?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StoragePartitioner.java
```diff
@@ -0,0 +1,104 @@
...
+            return columns;
+        }
+        columns.put(LOGIC_TABLE_NAME, this.getLogicTableColumnVal(model.getName()));
+        return columns;
+    }
+
+    /**
+     * BASE64 encode the original logic table name.
+     */
+    public String getLogicTableColumnVal(String logicTableName) {
```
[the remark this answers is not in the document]
EvanLjp: ok

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StoragePartitioner.java
EvanLjp: I am not clear about this. If the Create Processor is linear, it should not have a race condition.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: Just merge together and can be isolatedd again

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StoragePartitioner.java
wu-sheng: No race condition, hashmap is enough. This is a small issue. Let's focus on creating process.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
wu-sheng: This is not my point. I mean you keep related data in different places.

wu-sheng: StoragePartitioner is managing the metadata of ES storage module, then all related cache and metadata should be there.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: now we are facing more complex conditions. Since the template is created dynamically, we do not know when to terminate. It is not enough to judge by the existence of index and template.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: create template using PUT method, so don't need to delete

wu-sheng: Then let's polish the logic like this
1. Check template existing. 
2. If no template, creating.
3. If yes, check mapping compatible, then merging and update(put).
4. Create latest index.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
wu-sheng: You need to check all, but don't scan all templates. This is not allowed due to security reason.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/MeterFunction.java
```diff
@@ -29,6 +30,7 @@
  */
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
+@Inherited
```
wu-sheng, line 33: I think this will affect the class scan, such as MeterSystem, functionRegister could be wrong.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -286,6 +288,24 @@ public boolean isExistsIndex(String indexName) throws IOException {
         return client.indices().exists(request);
     }
 
+    public long getDocNumber(String indexName) throws IOException {
```
wu-sheng, line 291: You forget to remove this permnantly?

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -302,8 +322,8 @@ public boolean isExistsTemplate(String indexName) throws IOException {
         }
     }
 
-    public boolean createTemplate(String indexName, Map<String, Object> settings,
-                                  Map<String, Object> mapping) throws IOException {
+    public boolean putTemplate(String indexName, Map<String, Object> settings,
```
wu-sheng, line 325:
```suggestion
    public boolean createOrUpdateTemplate(String indexName, Map<String, Object> settings,
```

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -68,40 +78,87 @@ protected boolean isExists(Model model) throws StorageException {
...
                     }
                 }
                 indexName = TimeSeriesUtils.latestWriteIndexName(model);
+            } else {
+                indexName = tableName;
             }
-            if (!esClient.isExistsIndex(indexName)) {
-                boolean isAcknowledged = esClient.createIndex(indexName);
+
+            if (esClient.isExistsIndex(indexName)) {
```
wu-sheng, line 105: If not exist, don't create a new index?

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
EvanLjp: please review the installer again. used in isExist method.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
EvanLjp: fixed

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: fixed

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
wu-sheng: I read, and don't know why. We are checking template and index, whether having doc matters?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
wu-sheng: Still not correct, I think. If you are going to delete an existing index, all data will be deleted in every rebooting process. I think you have 2 options
1. Don't recreate index, like we used to. Make the deleting as the user's responsibility
2. Deleting index only when columns don't match as expected like template. But still, you could make the OAP accidentally removes all existing data, when users are not aware.

So, I recommend (1). What do you think?

wu-sheng:
Note, you have ES 7.10 e2e failing.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/PhysicalIndices.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+public class PhysicalIndices {
+
+    /**
+     * The relations of the logic table and the physical table.
+     */
+    private static final Map<String, String> RELATIONS = new ConcurrentHashMap<>();
+
+    public static String getPhysicalTableName(String logicName) {
```
wu-sheng, line 32: Logic as below exists in many places
```
   String tableName = PhysicalIndices.getPhysicalTableName(condition.getName());
   boolean aggregationMode = !tableName.equals(condition.getName());
```

Actually, you should use `RELATIONS#containsKey` to provide a new `isLogicIndex` method to get the result quickly, w/o doing the `String#equals`.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/PhysicalIndices.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ */
+
+package org.apache.skywalking.oap.server.storage.plugin.elasticsearch.base;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class PhysicalIndices {
```
wu-sheng, line 25: Please explain differences and boundaries between `PhysicalIndexManager` and `PhysicalIndices`.
Right now, PhysicalIndexManager becomes a util, as it doesn't include any new fields. These two should be merged?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/PhysicalIndices.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * The relations of the logic table and the physical table.
+     */
+    private static final Map<String, String> RELATIONS = new ConcurrentHashMap<>();
```
wu-sheng, line 30:
```suggestion
    private static final Map<String, String> LOGIC_INDECES_CATALOG = new ConcurrentHashMap<>();
```
No matter how you merge PhysicalIndexManager and PhysicalIndices, this field should be more meaningful.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/PhysicalIndices.java
wu-sheng: I am thing about
```
public enum IndexController {
   ....
   public class LogicIndicesRegister {
        private static final Map<String, String> LOGIC_INDECES_CATALOG = new HashMap<>();
   }
}
```
How about this?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: i add a tag template called install_sw_template_finished to avoid this condition.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
EvanLjp: I think this is important for preserving historical versions of the data.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
wu-sheng: I am confused about you are saying. I mean you are deleting the index every time no matter it is old version or new version.

wu-sheng: The index you created is an empty index, all column related things are in the template(such as mapping), am I right? 
Then why do we need to keep deleting/recreating? Even more for the indices not time series?

wu-sheng: The whole `INSTALL_TEMPLATE_FINISHED_TAG` thing should be removed/reverted, from my understanding. It is quite confusing. I can't find a reason to delete an index from OAP perspective, except for TTL case.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/PhysicalIndexer.java
```diff
@@ -0,0 +1,76 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.apache.skywalking.apm.util.StringUtil;
+import org.apache.skywalking.oap.server.core.storage.model.Model;
+
+/**
+ * The metrics data, that generated by OAL or MAL, would be partitioned to storage by the functions of the OAL or MAL.
+ * And, the other record data would be insulated storage by themselves definitions.
+ */
+@Slf4j
+public enum PhysicalIndexer {
```
wu-sheng, line 31:
```suggestion
public enum PhysicalIndices {
```

Typo.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/PhysicalIndices.java
wu-sheng: I think your latest change still separating `PhysicalIndexManager and PhysicalIndices` as `PhysicalIndices` and `PhysicalIndexer`. Could you explain why?
And again on coding style, you have `PhysicalIndexer` having all instance level methods, but actually no instance field, and at the same time, you have PhysicalIndices with a static field, with all static methods. These 2 typically should be reverted. No instance method means no point to maintain the instance. Why?

Please read my above suggestion. If you want a enum to provide an API set, it is fine, but please keep consistent.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/PhysicalIndexer.java
EvanLjp: There are 2 classes, one is PhysicalIndices, another is PhysicalIndexer.
PhysicalIndices stores the relations between the logic name and the physical name.
PhysicalIndexer is an indexer to help create some meta data.

wu-sheng: Let's merge 2 into IndexControl, https://github.com/apache/skywalking/pull/6499#discussion_r588883890

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
wu-sheng: We need use ESInstaller to check whether this index was created through the current booting process,  if yes, it is safe to delete and recreate. Otherwise, please terminate the booting with an error.
The compatibility check should be on index and template. If existing template and index have the required columns and mapping, then  booting could continue. Extra columns are not the isssue.
We could set a checker, in the storage completed stage, to check whether all physical indices matching the column. Only check, don't manipulate.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/PhysicalIndices.java
wu-sheng: Right now, I am thinking about this
```
public enum IndexController {
   // Move methods in PhysicalIndexer here
   ....
   public class LogicIndicesRegister {
        // Move PhysicalIndices methods into here
        private final Map<String, String> LOGIC_INDECES_CATALOG = new HashMap<>();
   }

  public class Checker {
       // Check the template and index compatible 
       // Note, the compatible check require column name, type, mapping, analyzer, etc. are all same
  }
}
```

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
wu-sheng: Read https://github.com/apache/skywalking/pull/6499#discussion_r588958394 for code structure recommendation.

EvanLjp:
@wu-sheng I use put index mapping to avoid delete index. https://www.elastic.co/guide/en/elasticsearch/reference/7.12/indices-put-mapping.html

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -38,68 +42,165 @@
...
-                    TimeSeriesUtils.latestWriteIndexName(model));
-            } else {
-                return esClient.isExistsIndex(model.getName());
+            if (!model.isTimeSeries()) {
+                return esClient.isExistsIndex(tableName);
             }
+            boolean exist = esClient.isExistsTemplate(tableName)
+                && esClient.isExistsIndex(TimeSeriesUtils.latestWriteIndexName(model));
+            if (RunningMode.isInitMode() && IndexController.INSTANCE.isAggregationMode(model)) {
+                appendTemplateMapping(tableName, (Map<String, Object>) esClient.getTemplate(tableName).get("mappings"));
```
wu-sheng, line 74: The return value is never used.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -38,68 +42,165 @@
...
+            boolean updateTemplate = !esClient.isExistsTemplate(tableName);
+            if (!updateTemplate) {
+                appendTemplateMapping(tableName, (Map<String, Object>) esClient.getTemplate(tableName)
+                                                                               .get("mappings"));
+                updateTemplate = !isTemplateMappingCompatible(tableName, mapping);
             }
-            if (!esClient.isExistsIndex(indexName)) {
-                boolean isAcknowledged = esClient.createIndex(indexName);
-                log.info("create {} index finished, isAcknowledged: {}", indexName, isAcknowledged);
+            if (updateTemplate) {
```
wu-sheng, line 123: Use `else` directly?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexController.java
```diff
@@ -0,0 +1,103 @@
...
+     * Generate the index doc ID.
+     */
+    public String generateDocId(String logicTableName, String originalID) {
+        return logicTableName + originalID;
+    }
+
+    /**
+     * Check the mode of the Model definition.
+     */
+    public boolean isAggregationMode(Model model) {
```
wu-sheng, line 61:
```suggestion
    public boolean isAggregationModel(Model model) {
```

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: there is a field called tables, the line is to append the history template to tables and then to compare.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: image
cannot use else, there is an optional condition.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
wu-sheng: Got it. Then `appendTemplateMapping` should be renamed.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -38,68 +42,165 @@
...
+        Map<String, Object> mapping = createEmptyMapping();
+        Map<String, Object> properties = getColumnProperties(mapping);
+        newFields.forEach(properties::put);
+        return mapping;
+    }
+
+    /**
+     * Append the mapping to the tables with the same table name key.
+     */
+    private Map<String, Object> appendTemplateMapping(String tableName, Map<String, Object> mapping) {
```
wu-sheng, line 169: `appendTemplateMapping` -> `resolveTemplateMapping`.
Please add comments about return value.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -38,68 +42,165 @@
...
 public class StorageEsInstaller extends ModelInstaller {
     private final Gson gson = new Gson();
-
     private final StorageModuleElasticsearchConfig config;
     protected final ColumnTypeEsMapping columnTypeEsMapping;
 
+    /**
+     * The mappings of the template .
+     */
+    private final Map<String, Map<String, Object>> tables;
```
wu-sheng, line 51: We still need to polish this part. `Map<String, Map<String, Object>>`->`Map<String, IndexFields>`.
`class IndexFields` could include a map to represent the data structure originally from ES.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -38,68 +42,165 @@
...
+            createTimeSeriesTable(model);
+        } else {
+            createNormalTable(model);
+        }
+    }
+
+    private void createNormalTable(Model model) throws StorageException {
         ElasticSearchClient esClient = (ElasticSearchClient) client;
+        String tableName = IndexController.INSTANCE.getTableName(model);
+        IndexController.LogicIndicesRegister.registerRelation(model.getName(), tableName);
```
wu-sheng, line 95: This has been called in the `isExists` stage. Don't need duplicate creation.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -38,68 +42,165 @@
...
+            }
+        } catch (IOException e) {
+            throw new StorageException("cannot create the normal index", e);
+        }
+    }
 
+    private void createTimeSeriesTable(Model model) throws StorageException {
+        ElasticSearchClient esClient = (ElasticSearchClient) client;
+        String tableName = IndexController.INSTANCE.getTableName(model);
+        IndexController.LogicIndicesRegister.registerRelation(model.getName(), tableName);
```
wu-sheng, line 112: This has been called in the `isExists` stage. Don't need duplicate creation.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -38,68 +42,165 @@
...
+            if (!updateTemplate) {
+                appendTemplateMapping(tableName, (Map<String, Object>) esClient.getTemplate(tableName)
+                                                                               .get("mappings"));
+                updateTemplate = !isTemplateMappingCompatible(tableName, mapping);
             }
-            if (!esClient.isExistsIndex(indexName)) {
-                boolean isAcknowledged = esClient.createIndex(indexName);
-                log.info("create {} index finished, isAcknowledged: {}", indexName, isAcknowledged);
+            if (updateTemplate) {
+                Map<String, Object> templateMapping = appendTemplateMapping(tableName, mapping);
```
wu-sheng, line 124: I think we should use `get` directly? The existing fields should have been resolved, right?

wu-sheng: This `append` method return existing columns, rather than new column. Which do you need here?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -38,68 +42,165 @@
...
-                if (!esClient.isExistsTemplate(model.getName())) {
-                    boolean isAcknowledged = esClient.createTemplate(model.getName(), settings, mapping);
-                    log.info(
-                        "create {} index template finished, isAcknowledged: {}", model.getName(), isAcknowledged);
-                    if (!isAcknowledged) {
-                        throw new StorageException("create " + model.getName() + " index template failure, ");
-                    }
-                }
-                indexName = TimeSeriesUtils.latestWriteIndexName(model);
+            boolean updateTemplate = !esClient.isExistsTemplate(tableName);
```
wu-sheng, line 117:
```suggestion
            boolean shouldUpdateTemplate = !esClient.isExistsTemplate(tableName);
```

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -38,68 +42,165 @@
...
+                    }
+                }
+            }
         } catch (IOException e) {
-            throw new StorageException(e.getMessage());
+            throw new StorageException("cannot create " + tableName + " index template", e);
+        }
+    }
+
+    private Map<String, Object> extractAppendMapping(final Map<String, Object> latestMapping,
```
wu-sheng, line 151: I hope you could shape/polish the comparing and extract extra fields/mappings in a better way. Such as using the new `class IndexFields`. Existing status and new fields should be appended in creating stage should be managed inside it.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/FunctionCategory.java
```diff
@@ -0,0 +1,63 @@
...
+            return (METER.getName() + Const.LINE + ((MeterFunction) annotation).functionName()).toLowerCase();
+        }
+        annotation = doGetAnnotation(aClass, MetricsFunction.class);
+        if (annotation != null) {
+            return (METRICS.getName() + Const.LINE + ((MetricsFunction) annotation).functionName()).toLowerCase();
+        }
+        return "";
+    }
+
+    public static Annotation doGetAnnotation(Class<?> clazz, Class<? extends Annotation> annotationClass) {
```
hanahmily, line 51: this method could be private since no other pieces refer to it.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: image
already checked , the 2 lines only for register the relation of the logic table and the physical table

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: same to above

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: For the mapping with same tableName in the tables, would append the new columns to it and return it.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
wu-sheng: My point is, you have done this in the `isExists` methods.
image

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
wu-sheng: You are returning the existMapping, what do I miss?
image

EvanLjp: u are right

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/FunctionCategory.java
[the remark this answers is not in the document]
EvanLjp: @wu-sheng said this way would affect package scan.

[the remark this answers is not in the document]
wu-sheng: We don't focus on efficiency, instead, accurate is important. We need to read the exact class declaring the annotation.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -38,67 +39,118 @@
...
-                indexName = TimeSeriesUtils.latestWriteIndexName(model);
+            boolean shouldUpdateTemplate = !esClient.isExistsTemplate(tableName);
+            if (IndexController.INSTANCE.isAggregationModel(model)) {
+                shouldUpdateTemplate = shouldUpdateTemplate || !structures.containsStructure(tableName, mapping);
             }
-            if (!esClient.isExistsIndex(indexName)) {
-                boolean isAcknowledged = esClient.createIndex(indexName);
-                log.info("create {} index finished, isAcknowledged: {}", indexName, isAcknowledged);
+            if (shouldUpdateTemplate) {
+                structures.resolveStructure(tableName, mapping);
```
wu-sheng, line 123: Why `resolveStructure` again?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexStructures.java
```diff
@@ -0,0 +1,157 @@
...
+
+    public Map<String, Object> getMapping(String tableName) {
+        return wrapper.wrapper(
+            structures.containsKey(tableName) ? structures.get(tableName).properties : new HashMap<>());
+    }
+
+    /**
+     * Add or append field when the current structures don't contain the input structure or having new fields in it.
+     */
+    public void resolveStructure(String tableName, Map<String, Object> mapping) {
```
wu-sheng, line 64: I think this method should be `idempotent`, could we use `structures.put(tableName, fields);` every time?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: this is the first time to resolve structure, could u check again?

wu-sheng: One in `isExist`, this is second time when creating.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexStructures.java
wu-sheng: `resolveStructure` could be called multiple times, such as `no-init` mode waiting, like I mentioned in https://github.com/apache/skywalking/pull/6499#discussion_r590203776

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: image
only check contains

EvanLjp: the above calling is for check exist template. but the current operation is for add new columns for template.
structures.resolveStructure(
                    tableName, (Map<String, Object>) esClient.getTemplate(tableName).get("mappings")
                );
structures.resolveStructure(tableName, mapping);

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexStructures.java
EvanLjp: renamed the function

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
EvanLjp: renamed the function

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -38,67 +39,118 @@
...
-                return esClient.isExistsTemplate(model.getName()) && esClient.isExistsIndex(
-                    TimeSeriesUtils.latestWriteIndexName(model));
-            } else {
-                return esClient.isExistsIndex(model.getName());
+            if (!model.isTimeSeries()) {
+                return esClient.isExistsIndex(tableName);
+            }
+            boolean exist = esClient.isExistsTemplate(tableName)
+                && esClient.isExistsIndex(TimeSeriesUtils.latestWriteIndexName(model));
+            if (RunningMode.isInitMode() && exist && IndexController.INSTANCE.isAggregationModel(model)) {
```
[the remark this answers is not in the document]
EvanLjp: thx

wu-sheng:
@EvanLjp Have your local tests passed for both ElasticSearch 6 and 7? If so, @kezhenxu94 please help with e2e testing.

wu-sheng:
The codes are much better now, I prefer to wait for the e2e passed, and run a deeper review. All @apache/skywalking-committers , welcome to join the review. This would be the most important change since the 8.0 upgrade.

EvanLjp:
> @EvanLjp Have your local tests passed for both ElasticSearch 6 and 7? If so, @kezhenxu94 please help with e2e testing.

yes, already local test it

kezhenxu94:
> yes, already local test it

Maybe you just tested it in IDE? The OAP failed to start up in docker container

```
oap_1       | 2021-03-10 01:15:12,503 - org.apache.skywalking.oap.server.storage.plugin.elasticsearch.base.StorageEsInstaller -18119 [main] INFO  [] - create metrics-percentile index template finished, isAcknowledged: true
oap_1       | 2021-03-10 01:15:12,732 - org.apache.skywalking.oap.server.starter.OAPServerBootstrap -18348 [main] ERROR [] - Elasticsearch exception [type=illegal_argument_exception, reason=cannot create index with name [metrics-percentile-20210310], because it matches with template [metrics] that creates data streams only, use create data stream api instead]
oap_1       | org.elasticsearch.ElasticsearchStatusException: Elasticsearch exception [type=illegal_argument_exception, reason=cannot create index with name [metrics-percentile-20210310], because it matches with template [metrics] that creates data streams only, use create data stream api instead]
oap_1       | 	at org.elasticsearch.rest.BytesRestResponse.errorFromXContent(BytesRestResponse.java:177) ~[elasticsearch-7.5.0.jar:7.5.0]
oap_1       | 	at org.elasticsearch.client.RestHighLevelClient.parseEntity(RestHighLevelClient.java:1793) ~[elasticsearch-rest-high-level-client-7.5.0.jar:7.5.0]
oap_1       | 	at org.elasticsearch.client.RestHighLevelClient.parseResponseException(RestHighLevelClient.java:1770) ~[elasticsearch-rest-high-level-client-7.5.0.jar:7.5.0]
oap_1       | 	at org.elasticsearch.client.RestHighLevelClient.internalPerformRequest(RestHighLevelClient.java:1527) ~[elasticsearch-rest-high-level-client-7.5.0.jar:7.5.0]
oap_1       | 	at org.elasticsearch.client.RestHighLevelClient.performRequest(RestHighLevelClient.java:1499) ~[elasticsearch-rest-high-level-client-7.5.0.jar:7.5.0]
oap_1       | 	at org.elasticsearch.client.RestHighLevelClient.performRequestAndParseEntity(RestHighLevelClient.java:1466) ~[elasticsearch-rest-high-level-client-7.5.0.jar:7.5.0]
oap_1       | 	at org.elasticsearch.client.IndicesClient.create(IndicesClient.java:131) ~[elasticsearch-rest-high-level-client-7.5.0.jar:7.5.0]
oap_1       | 	at org.apache.skywalking.oap.server.storage.plugin.elasticsearch7.client.ElasticSearch7Client.createIndex(ElasticSearch7Client.java:123) ~[storage-elasticsearch7-plugin-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 	at org.apache.skywalking.oap.server.storage.plugin.elasticsearch.base.StorageEsInstaller.createTimeSeriesTable(StorageEsInstaller.java:144) ~[storage-elasticsearch-plugin-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 	at org.apache.skywalking.oap.server.storage.plugin.elasticsearch.base.StorageEsInstaller.createTable(StorageEsInstaller.java:88) ~[storage-elasticsearch-plugin-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 	at org.apache.skywalking.oap.server.core.storage.model.ModelInstaller.whenCreating(ModelInstaller.java:56) ~[server-core-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 	at org.apache.skywalking.oap.server.core.storage.model.StorageModels.addModelListener(StorageModels.java:90) ~[server-core-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 	at org.apache.skywalking.oap.server.storage.plugin.elasticsearch7.StorageModuleElasticsearch7Provider.start(StorageModuleElasticsearch7Provider.java:224) ~[storage-elasticsearch7-plugin-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 	at org.apache.skywalking.oap.server.library.module.BootstrapFlow.start(BootstrapFlow.java:49) ~[library-module-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 	at org.apache.skywalking.oap.server.library.module.ModuleManager.init(ModuleManager.java:62) ~[library-module-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 	at org.apache.skywalking.oap.server.starter.OAPServerBootstrap.start(OAPServerBootstrap.java:43) [server-bootstrap-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 	at org.apache.skywalking.oap.server.starter.OAPServerStartUp.main(OAPServerStartUp.java:27) [server-starter-es7-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 	Suppressed: org.elasticsearch.client.ResponseException: method [PUT], host [http://es:9200], URI [/metrics-percentile-20210310?master_timeout=30s&timeout=30s], status line [HTTP/1.1 400 Bad Request]
oap_1       | {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"cannot create index with name [metrics-percentile-20210310], because it matches with template [metrics] that creates data streams only, use create data stream api instead"}],"type":"illegal_argument_exception","reason":"cannot create index with name [metrics-percentile-20210310], because it matches with template [metrics] that creates data streams only, use create data stream api instead"},"status":400}
oap_1       | 		at org.elasticsearch.client.RestClient.convertResponse(RestClient.java:283) ~[elasticsearch-rest-client-7.5.0.jar:7.5.0]
oap_1       | 		at org.elasticsearch.client.RestClient.performRequest(RestClient.java:261) ~[elasticsearch-rest-client-7.5.0.jar:7.5.0]
oap_1       | 		at org.elasticsearch.client.RestClient.performRequest(RestClient.java:235) ~[elasticsearch-rest-client-7.5.0.jar:7.5.0]
oap_1       | 		at org.elasticsearch.client.RestHighLevelClient.internalPerformRequest(RestHighLevelClient.java:1514) ~[elasticsearch-rest-high-level-client-7.5.0.jar:7.5.0]
oap_1       | 		at org.elasticsearch.client.RestHighLevelClient.performRequest(RestHighLevelClient.java:1499) ~[elasticsearch-rest-high-level-client-7.5.0.jar:7.5.0]
oap_1       | 		at org.elasticsearch.client.RestHighLevelClient.performRequestAndParseEntity(RestHighLevelClient.java:1466) ~[elasticsearch-rest-high-level-client-7.5.0.jar:7.5.0]
oap_1       | 		at org.elasticsearch.client.IndicesClient.create(IndicesClient.java:131) ~[elasticsearch-rest-high-level-client-7.5.0.jar:7.5.0]
oap_1       | 		at org.apache.skywalking.oap.server.storage.plugin.elasticsearch7.client.ElasticSearch7Client.createIndex(ElasticSearch7Client.java:123) ~[storage-elasticsearch7-plugin-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 		at org.apache.skywalking.oap.server.storage.plugin.elasticsearch.base.StorageEsInstaller.createTimeSeriesTable(StorageEsInstaller.java:144) ~[storage-elasticsearch-plugin-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 		at org.apache.skywalking.oap.server.storage.plugin.elasticsearch.base.StorageEsInstaller.createTable(StorageEsInstaller.java:88) ~[storage-elasticsearch-plugin-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 		at org.apache.skywalking.oap.server.core.storage.model.ModelInstaller.whenCreating(ModelInstaller.java:56) ~[server-core-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 		at org.apache.skywalking.oap.server.core.storage.model.StorageModels.addModelListener(StorageModels.java:90) ~[server-core-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 		at org.apache.skywalking.oap.server.storage.plugin.elasticsearch7.StorageModuleElasticsearch7Provider.start(StorageModuleElasticsearch7Provider.java:224) ~[storage-elasticsearch7-plugin-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 		at org.apache.skywalking.oap.server.library.module.BootstrapFlow.start(BootstrapFlow.java:49) ~[library-module-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 		at org.apache.skywalking.oap.server.library.module.ModuleManager.init(ModuleManager.java:62) ~[library-module-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 		at org.apache.skywalking.oap.server.starter.OAPServerBootstrap.start(OAPServerBootstrap.java:43) [server-bootstrap-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]
oap_1       | 		at org.apache.skywalking.oap.server.starter.OAPServerStartUp.main(OAPServerStartUp.java:27) [server-starter-es7-8.5.0-SNAPSHOT.jar:8.5.0-SNAPSHOT]

```

EvanLjp:
when I start the docker with init mode, the process is finally succeeded.
image

wu-sheng:
I run local testing on ElasticSearch 7.6.2(no docker), it works. @kezhenxu94 

@EvanLjp I notice a tiny issue, the entity ID missing a `_` between metrics and timebucket. It look like this `service_resp_time202103101352_bW9ja19jX3NlcnZpY2U=.1`

EvanLjp:
> I run local testing on ElasticSearch 7.6.2(no docker), it works. @kezhenxu94
> 
> @EvanLjp I notice a tiny issue, the entity ID missing a `_` between metrics and timebucket. It look like this `service_resp_time202103101352_bW9ja19jX3NlcnZpY2U=.1`

I don't think this delimiter has much meaning for the final ID, so I don't add it

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexController.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    public static class LogicIndicesRegister {
+
+        /**
+         * The relations of the logic table and the physical table.
+         */
+        private static final Map<String, String> LOGIC_INDICES_CATALOG = new ConcurrentHashMap<>();
+
+        public static final String LOGIC_TABLE_NAME = "logic_table";
```
wu-sheng, line 85: Let's rename this to `metric_name`.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexController.java
```diff
@@ -82,7 +83,7 @@ public boolean isAggregationModel(Model model) {
          */
         private static final Map<String, String> LOGIC_INDICES_CATALOG = new ConcurrentHashMap<>();
 
-        public static final String LOGIC_TABLE_NAME = "logic_table";
```
wu-sheng: Forget to change `LOGIC_TABLE_NAME`.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchProvider.java
```diff
@@ -246,4 +250,11 @@ public String convert(final String indexName) {
             return indexName;
         }
     }
+
+    private static class SkyWalkingTagConverter implements IndexNameConverter {
+        @Override
+        public String convert(final String indexName) {
+            return "sw" + "_" + indexName;
+        }
+    }
```
wu-sheng, line 259: We just need to use `sw` as the default namespace, please set it in the `#prepare` method. We don't need to add this new converter. And more importantly, this name is super confusing.

wu-sheng:
I feel good about the current version now. @kezhenxu94 @arugal @dmsolr Could you run another round review?

wu-sheng (approved):
LGTM

wu-sheng:
I am going to merge this first, @ascrutae please deploy this on the demo env, and let's see how this works in days.

Issue #6481, opened 2021-03-03
NullPointerException when using KafkaProducer.send(record)


yaojingguo:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [X] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Bug
- Which version of SkyWalking, OS, and JRE?

1. SkyWalking 8.4.0
1. OS: macOS 10.15.5
1. JDK: Oracle JDK 1.8.0_162
1. kafka_2.13-2.7.0 server
1. `spring-kafka:2.2.6.RELEASE` which depends on `kafka-clients:2.6.0`

- What happened?
Use the following to send a message to Kafka:
```
  Future<RecordMetadata> result = kafkaProducer.send(record);
  result.get();
```

After running the above code, I found the following exception in `agent/logs/skywalking-api.log`:
```
ERROR 2021-03-03 23:40:36:098 kafka-producer-network-thread | producer-1 InstMethodsInter : class[class org.apache.kafka.clients.producer.KafkaProducer$InterceptorCallback] before method[onCompletion] intercept failure 
java.lang.NullPointerException
	at org.apache.skywalking.apm.plugin.kafka.CallbackInterceptor.getSnapshot(CallbackInterceptor.java:83)
	at org.apache.skywalking.apm.plugin.kafka.CallbackInterceptor.beforeMethod(CallbackInterceptor.java:44)
	at org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstMethodsInter.intercept(InstMethodsInter.java:76)
	at org.apache.kafka.clients.producer.KafkaProducer$InterceptorCallback.onCompletion(KafkaProducer.java)
	at org.apache.kafka.clients.producer.internals.ProducerBatch.completeFutureAndFireCallbacks(ProducerBatch.java:228)
	at org.apache.kafka.clients.producer.internals.ProducerBatch.done(ProducerBatch.java:197)
	at org.apache.kafka.clients.producer.internals.Sender.completeBatch(Sender.java:653)
	at org.apache.kafka.clients.producer.internals.Sender.completeBatch(Sender.java:634)
	at org.apache.kafka.clients.producer.internals.Sender.handleProduceResponse(Sender.java:554)
	at org.apache.kafka.clients.producer.internals.Sender.lambda$sendProduceRequest$0(Sender.java:743)
[...]

PR #6503, opened 2021-03-07
Add test cases for kafka-plugin

yaojingguo:
1. Add plugin test for KafkaProducer.send(record).

Since an exception in CallbackInterceptor can't be checked by a plugin test. The new test does not check exceptions thrown by  CallbackInterceptor. But the test does cover more KafkaProducer usages.

2. Add exception checking test cases for CallbackInterceptor.

These unit test cases check exceptions thrown by CallbackInterceptor.

See #6481.

@zifeihan


wu-sheng (commented):
Please revert `skywalking-ui` to match the current master.. This is an unrelated change.

yaojingguo:
A new commit after the change to `skywalking-ui` is reverted.

wu-sheng (approved):

zifeihan (approved):
LGTM

Issue #6498, opened 2021-03-05
Propose to optimize the self monitoring grafana dashboard.


Ax1an:
- Why do you submit this issue?
- [x] Feature or performance improvement

### Requirement or improvement
  I propose to optimize the self monitoring grafana dashboard.

  Now, the dashboard is not only troublesome to import, but also lacks some important metrics. So I made two new dashboards that I wanted to contribute to the SkyWalking Community.

  The new dashboard is not only easy to import, but also has more practical metrics. Our company engineer uses it every day.

  I believe it can help more people who use SkyWalking. Whether it's self-monitoring or optimizing performance.

  Here is a comparison between the old and new dashboards.

  SkyWalking offical dashboard:
  
  ![image](https://user-images.githubusercontent.com/28091237/110111565-82317d00-7deb-11eb-855c-4797365411df.png)

  My optimized dashboard:

  1. SkyWalking Cluster Monitor Dashboard
  
  ![image](https://user-images.githubusercontent.com/28091237/110111644-a2f9d280-7deb-11eb-994b-674ac9e93351.png)

  2. SkyWalking Standalone Monitor Dashboard

  ![image](https://user-images.githubusercontent.com/28091237/110111705-b5740c00-7deb-11eb-9b3d-889a6ef55259.png)

wu-sheng:
> SkyWalking Standalone Monitor Dashboard

Should this be called, OAP metrics per instance? Standalone means there is only one node of OAP, please confirm.

Ax1an:
Yes，you are right. If the community accepts my propose, I will adjust the name later.




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "apache/skywalking"                                                                                    <notifications@github.com&gt;;
发送时间:&nbsp;2021年3月5日(星期五) 晚上7:52
收件人:&nbsp;"apache/skywalking"<skywalking@noreply.github.com&gt;;
抄送:&nbsp;"Ax1an"<ax1an@foxmail.com&gt;;"Author"<author@noreply.github.com&gt;;
主题:&nbsp;Re: [apache/skywalking] Propose to optimize the self monitoring grafana dashboard. (#6498)





  
SkyWalking Standalone Monitor Dashboard
  
Should this be called, OAP metrics per instance? Standalone means there is only one node of OAP, please confirm.

wu-sheng:
@apache/skywalking-committers I will treat this as lazy consensus, if no objection, then good to go.

[...]

PR #6516, opened 2021-03-08
Add telemetry data about metrics in, metrics scraping and trace in metrics to zipkin receiver.

Ax1an:
#6498 

I find that telemetry data about log in metrics have been added. So I added telemetry data about metrics in, metrics scraping and trace in metrics to zipkin receiver.

### DIFF
--- a/oap-server/server-fetcher-plugin/prometheus-fetcher-plugin/src/main/java/org/apache/skywalking/oap/server/fetcher/prometheus/provider/PrometheusFetcherProvider.java
+++ b/oap-server/server-fetcher-plugin/prometheus-fetcher-plugin/src/main/java/org/apache/skywalking/oap/server/fetcher/prometheus/provider/PrometheusFetcherProvider.java
@@ -77,10 +77,10 @@ public class PrometheusFetcherProvider extends ModuleProvider {
                 .provider()
                 .getService(MetricsCreator.class);
         histogram = metricsCreator.createHistogramMetric(
-                "metrics_scratch_latency", "The process latency of metrics scratching",
+                "metrics_fetcher_latency", "The process latency of metrics scratching",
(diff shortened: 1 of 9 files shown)


oap-server/server-fetcher-plugin/prometheus-fetcher-plugin/src/main/java/org/apache/skywalking/oap/server/fetcher/prometheus/provider/PrometheusFetcherProvider.java
```diff
@@ -62,8 +67,22 @@
...
+
+    private final CounterMetrics errorCounter;
+
     public PrometheusFetcherProvider() {
         config = new PrometheusFetcherConfig();
+        MetricsCreator metricsCreator = getManager().find(TelemetryModule.NAME)
+                .provider()
+                .getService(MetricsCreator.class);
+        histogram = metricsCreator.createHistogramMetric(
+                "metrics_scratch_latency", "The process latency of metrics scratching",
```
wu-sheng, line 80:
```suggestion
                "metrics_fetcher_latency", "The process latency of metrics scratching",
```
`scratch` -> `fetcher`.

wu-sheng:
I think somehow you break tests?

Ax1an:
@wu-sheng I probably know what the problem is and I'll try to fix it.

kezhenxu94 (commented):
Seems `scratch` / `scratching` is a typo of `scrap` / `scraping`, please correct them in both title, commit message, codes, comments, etc.

Ax1an:
@kezhenxu94 Done.

kezhenxu94 (dismissed):

wu-sheng (approved):

PR #6519, opened 2021-03-09
Set up code quality check and fix some

kezhenxu94:
The scan results of `master` branch will be in [Security tab](https://github.com/apache/skywalking/security), and the results in a PR will be in the "Checks" tab and in the PR page, context name is "Code scanning results / CodeQL".

### DIFF
--- a/.github/actions/skip/action.yml
+++ b/.github/actions/skip/action.yml
@@ -34,6 +34,8 @@ inputs:
       "docs/menu.yml"
       ".github/actions/skip/action.yml"
       "dist-material/release-docs"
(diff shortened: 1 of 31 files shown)


wu-sheng:
Should we merge this?

kezhenxu94:
> Should we merge this?

Yes, I think it's good enough to merge, from the official GitHub Action, there are some basic checks out of box that may avoid obvious issues https://github.com/github/codeql/tree/main/java/ql/src , and we can add our own rules to the check list in the future if needed

wu-sheng (approved):

PR #6538, opened 2021-03-11
resolve "receiver don't need to get itself when healthCheck #6515 "

### LINKED ISSUE #6515: receiver don't need to get itself when healthCheck

- [X] Question or discussion

___
### Question
- What do you want to know?
If OapNode‘s role is receiver, the OAPNodeChecker'method isHealth() will return unHealth.
```
 public static ClusterHealthStatus isHealth(List<RemoteInstance> remoteInstances) {
        if (CollectionUtils.isEmpty(remoteInstances)) {
            return ClusterHealthStatus.unHealth("can't get the instance list");
        }
        List<RemoteInstance> selfInstances = remoteInstances.stream().
                filter(remoteInstance -> remoteInstance.getAddress().isSelf()).collect(Collectors.toList());
        if (CollectionUtils.isEmpty(selfInstances)) {
            return ClusterHealthStatus.unHealth("can't get itself");
        }
        if (remoteInstances.size() > 1 && hasIllegalNodeAddress(remoteInstances)) {
            return ClusterHealthStatus.unHealth("find illegal node in cluster mode such as 127.0.0.1, localhost");
        }
        return ClusterHealthStatus.HEALTH;
    }
```

I think OAPNodeChecker should check its role and then decide whether to get itself.
But how can I get role?

s00373198:
- [x] Closes #6515.
1、Add variable private static string role in OAPNodeChecker, and set it in CoreModuleProvider#start stage
2、check private static string role before CoreModuleProvider#isHealth check itself
Fix receiver don't need to get itself when healthCheck


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/cluster/OAPNodeChecker.java
```diff
@@ -28,6 +30,9 @@
 public class OAPNodeChecker {
     private static final Set<String> ILLEGAL_NODE_ADDRESS_IN_CLUSTER_MODE = Sets.newHashSet("127.0.0.1", "localhost");
 
+    @Setter
+    private static String ROLE;
```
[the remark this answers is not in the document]
s00373198: Got it，review it again please

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleProvider.java
```diff
@@ -315,6 +316,8 @@ public void start() throws ModuleStartException {
                 .registerRemote(gRPCServerInstance);
         }
 
+        OAPNodeChecker.setROLE(CoreModuleConfig.Role.valueOf(moduleConfig.getRole()));
```
wu-sheng, line 319: This is not a perfect way to do this. Take a look at how L304 works. We should add this as a part of that IF. 
Your way in here could cause `IllegalArgumentException` if a user sets the value as `mixed1`, rather than `Mixed`. The whole OAP will fail. If you use the better way, then it just fails back to the Mixed mode.

s00373198: Glad to receive you reply, i undersatand the problem, but i cann't understand L304 relation with it, but i find a solution and PR again, is this ok?

s00373198: I think about it， Did you mean create a new enum class to store current server role?

s00373198:
need i solve the problem
image
 
the codecov result seems not right

wu-sheng (approved):

PR #6549, opened 2021-03-15
Optimize the self monitoring grafana dashboard.

Ax1an:
#6498 

Optimize the self monitoring grafana dashboard.

SkyWalking OAP Cluster Monitor Dashboard: 

image

SkyWalking OAP Instance Monitor Dashboard: 

image


kezhenxu94:
# Some screenshots with more metrics (not perfectly complete)
<details>
  <summary>Click to expand</summary>
  
Screen Shot 2021-03-19 at 13 39 06
Screen Shot 2021-03-19 at 13 39 13
Screen Shot 2021-03-19 at 13 39 25
Screen Shot 2021-03-19 at 13 39 31
Screen Shot 2021-03-19 at 13 44 38
Screen Shot 2021-03-19 at 13 44 46
Screen Shot 2021-03-19 at 13 44 52
Screen Shot 2021-03-19 at 13 44 58
Screen Shot 2021-03-19 at 13 45 03
</details>


Generally LGTM, @Ax1an can you please take a look at this 👇 the unit (25 min) looks unreasonable.

Screen Shot 2021-03-19 at 13 49 14

Ax1an:
@kezhenxu94 
I've looked at your question carefully, but in fact, the default unit after I import is seconds. Have you adjusted the chart unit after importing?

From your screenshot, I found that there is something wrong with the legend of some charts, which has been adjusted.

image

image

kezhenxu94:
I didn't change anything

Ax1an:
@kezhenxu94 

According to your screenshot, I found that the cluster dashboard is good, there is a problem with the instance dashboard. 

Could you edit these two panels to see if their units are different?

image

image

wu-sheng:
I am a little confused, what `Minute` and `Second` suffixes mean? I can see the Y axle doesn't match in many cases.

@Ax1an The question is, how the latency could be 25mins.

wu-sheng:
I think @kezhenxu94 's concern is there is a bug existing about converting the unit.

wu-sheng:
I think most latency in the second when report.

https://github.com/apache/skywalking/blob/25214e910a193fbf22511ce6048964f4ce4d3048/oap-server/server-telemetry/telemetry-api/src/main/java/org/apache/skywalking/oap/server/telemetry/api/HistogramMetrics.java#L51

kezhenxu94 (approved):
Well, must be something wrong in my previous env, tested in a new env and it works.

Ax1an:
@wu-sheng 

I use the rate function to represent the number of seconds.

I use the increase function to represent the increments per minute.

For example, let's assume that there are two persistence_timer_bulk_prepare operations between 09: 59 and 10: 00, which take 20 seconds and 10 seconds respectively. Then, it took 2 minutes to do next persistence_timer_bulk_prepare operation.

This panel should look like this: 

image

This may be difficult to understand at first. It works better when you get used to it. It can help us to intuitively see some high-latency single operation.

Ax1an:
For all time related metrics, the default unit I set is second, because I have also read the part of the code you said.

You can see this in the configuration file.

image

wu-sheng:
> @wu-sheng
> 
> I use the rate function to represent the trend in number per second.
> 
> I use the increase function to represent the increments per minute.
> 
> For example, let's assume that there are two persistence_timer_bulk_prepare operations between 09: 59 and 10: 00, which take 20 seconds and 10 seconds respectively. Then, it took 2 minutes to do next persistence_timer_bulk_prepare operation.
> 
> This panel should look like this:
> 
> image
> 
> This may be difficult to understand at first. It works better when you get used to it. It can help us to intuitively see some high-latency single operation. In order to optimize it.

So, basically, you are showing trend.

wu-sheng (approved):

PR #6552, opened 2021-03-15
Remove group concept from AvgHistogramFunction

wu-sheng:

Remove group concept from AvgHistogramFunction. Heatmap(function result) doesn't support labels. FYI @wankai123 


wu-sheng:
If labels left(not fully aggregated) when calling histogram, the new implementation will ignore them(in the group field), and aggregate them as multiple records.

hanahmily (approved):
LGTM

kezhenxu94 (approved):

PR #6559, opened 2021-03-16
Collect and report agent starting / shutdown events

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>. Related to #5980 

This is a follow-up of #5980

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/ServiceInstanceGenerator.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/ServiceInstanceGenerator.java
@@ -28,8 +28,6 @@ import static org.apache.skywalking.apm.util.StringUtil.isEmpty;
 
 @Getter
 public class ServiceInstanceGenerator implements BootService {
-    private volatile boolean isGenerated = false;
-
     @Override
     public void prepare() throws Throwable {
         if (!isEmpty(Config.Agent.INSTANCE_NAME)) {
@@ -37,7 +35,6 @@ public class ServiceInstanceGenerator implements BootService {
         }
 
         Config.Agent.INSTANCE_NAME = UUID.randomUUID().toString().replaceAll("-", "") + "@" + OSUtil.getIPV4();
-        isGenerated = true;
     }
 
     @Override
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/ServiceInstanceGenerator.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/ServiceInstanceGenerator.java
@@ -21,6 +21,7 @@ package org.apache.skywalking.apm.agent.core;
 import java.util.UUID;
 import lombok.Getter;
 import lombok.experimental.Accessors;
+import org.apache.skywalking.apm.agent.core.boot.BootService;
 import org.apache.skywalking.apm.agent.core.conf.Config;
 import org.apache.skywalking.apm.agent.core.os.OSUtil;
 
@@ -28,17 +29,36 @@ import static org.apache.skywalking.apm.util.StringUtil.isEmpty;
 
 @Getter
 @Accessors(fluent = true)
-public enum ServiceInstanceGenerator {
-    SINGLETON;
-
+public class ServiceInstanceGenerator implements BootService {
     private volatile boolean isGenerated = false;
 
-    public synchronized void generateIfNotSpecified() {
+    @Override
+    public void prepare() throws Throwable {
         if (!isEmpty(Config.Agent.INSTANCE_NAME)) {
             return;
         }
 
         Config.Agent.INSTANCE_NAME = UUID.randomUUID().toString().replaceAll("-", "") + "@" + OSUtil.getIPV4();
         isGenerated = true;
+    }
+
+    @Override
+    public void boot() throws Throwable {
+
+    }
+
+    @Override
+    public void onComplete() throws Throwable {
+
+    }
+
+    @Override
+    public void shutdown() throws Throwable {
+
+    }
+
+    @Override
+    public int bootOrder() {
+        return Integer.MIN_VALUE;
     }
 }
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/ServiceInstanceGenerator.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/ServiceInstanceGenerator.java
@@ -16,27 +16,29 @@
  *
  */
 
-package org.apache.skywalking.apm.agent.core.boot;
+package org.apache.skywalking.apm.agent.core;
 
-import org.apache.skywalking.apm.agent.core.remote.GRPCChannelManager;
+import java.util.UUID;
+import lombok.Getter;
+import lombok.experimental.Accessors;
+import org.apache.skywalking.apm.agent.core.conf.Config;
+import org.apache.skywalking.apm.agent.core.os.OSUtil;
 
-/**
- * The <code>BootService</code> is an interface to all remote, which need to boot when plugin mechanism begins to work.
- * {@link #boot()} will be called when <code>BootService</code> start up.
- */
-public interface BootService {
-    void prepare() throws Throwable;
+import static org.apache.skywalking.apm.util.StringUtil.isEmpty;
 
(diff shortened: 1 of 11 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/ServiceInstanceGenerator.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.Getter;
+import lombok.experimental.Accessors;
+import org.apache.skywalking.apm.agent.core.conf.Config;
+import org.apache.skywalking.apm.agent.core.os.OSUtil;
+
+import static org.apache.skywalking.apm.util.StringUtil.isEmpty;
+
+@Getter
+@Accessors(fluent = true)
+public enum ServiceInstanceGenerator {
```
wu-sheng, line 31: This should be a service, rather than an enum, then user could override this by a plugin.

kezhenxu94: Done

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/ServiceInstanceGenerator.java
```diff
@@ -0,0 +1,64 @@
...
+import lombok.Getter;
+import lombok.experimental.Accessors;
+import org.apache.skywalking.apm.agent.core.boot.BootService;
+import org.apache.skywalking.apm.agent.core.conf.Config;
+import org.apache.skywalking.apm.agent.core.os.OSUtil;
+
+import static org.apache.skywalking.apm.util.StringUtil.isEmpty;
+
+@Getter
+@Accessors(fluent = true)
```
wu-sheng, line 31: Why suddenly we use this style? I have a concern about this. This could be a new style start.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/ServiceInstanceGenerator.java
```diff
@@ -0,0 +1,64 @@
...
+
+    }
+
+    @Override
+    public void shutdown() throws Throwable {
+
+    }
+
+    @Override
+    public int bootOrder() {
```
wu-sheng, line 61:
```suggestion
    public int priority() {
```

Does the current mode fit, the booting and shut down follow FILO? If so, one priority method works.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/EventReportServiceClient.java
```diff
@@ -0,0 +1,193 @@
...
+
+    private ServiceInstanceGenerator serviceInstanceGenerator;
+
+    @Override
+    public void prepare() throws Throwable {
+        ServiceManager.INSTANCE.findService(GRPCChannelManager.class).addChannelListener(this);
+
+        serviceInstanceGenerator = ServiceManager.INSTANCE.findService(ServiceInstanceGenerator.class);
+
+        if (serviceInstanceGenerator.isGenerated()) {
```
wu-sheng, line 64: Let's change it to `serviceInstanceGenerator#generate(callback {}).

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/EventReportServiceClient.java
```diff
@@ -0,0 +1,193 @@
...
+
+    @Override
+    public void onComplete() throws Throwable {
+        reportStartingEvent();
+    }
+
+    @Override
+    public void shutdown() throws Throwable {
+        if (serviceInstanceGenerator.isGenerated()) {
+            // If the agent service instance name is randomly generated, ignore the shutdown signal.
```
wu-sheng, line 104: Do you mean, UUID-based instance doesn't report a shutdown event?

wu-sheng: Why? My feeling is, this is especially should have the event. When you see its metrics(such as CPM) suddenly goes down as 0, and find out the shut event exists, you know, this is not a performance accident.

kezhenxu94: > Do you mean, UUID-based instance doesn't report a shutdown event?

Yes, UUID-based instance won't report start / shutdown events.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/EventReportServiceClient.java
kezhenxu94: > Let's change it to `serviceInstanceGenerator#generate(callback {}).

`serviceInstanceGenerator#generate` is removed now

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/EventReportServiceClient.java
kezhenxu94: > Why? My feeling is, this is especially should have the event. When you see its metrics(such as CPM) suddenly goes down as 0, and find out the shut event exists, you know, this is not a performance accident.

OK, this makes sense to me, I was overthinking about this

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/EventReportServiceClient.java
```diff
@@ -0,0 +1,178 @@
...
+
+            @Override
+            public void onCompleted() {
+                latch.countDown();
+            }
+        });
+
+        collector.onNext(shutdownEvent.build());
+        collector.onCompleted();
+        latch.await();
```
wu-sheng, line 133: Consider using a timeout mechanism. This way could block the shutdown process.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/EventReportServiceClient.java
```diff
@@ -130,7 +131,7 @@ public void onCompleted() {
 
         collector.onNext(shutdownEvent.build());
         collector.onCompleted();
-        latch.await();
+        latch.await(1, TimeUnit.SECONDS);
```
wu-sheng, line 134: Do you have tested? Is 1s enough?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/EventReportServiceClient.java
kezhenxu94: > Consider using a timeout mechanism. This way could block the shutdown process.

No, the gRPC service would either success (`onCompleted` called) or timeout (`onError` called), either of them should call `latch.countDown()` and `latch.await` won't block the shutdown process

wu-sheng: I can't see you are using `withDeadlineAfter`, take a look at `TraceSegmentServiceClient` about how to do timeout in gRPC.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/EventReportServiceClient.java
```diff
@@ -0,0 +1,183 @@
...
+
+        reportStartingEvent();
+    }
+
+    private void reportStartingEvent() {
+        if (reported.compareAndSet(false, true)) {
+            return;
+        }
+
+        startingEvent.setEndTime(System.currentTimeMillis());
```
wu-sheng, line 159: In `onError`, you will set `reported = false` again, then it could cause the end time of starting event will be reset in the next report. I think you don't want that.

wu-sheng (approved):

PR #6565, opened 2021-03-17
support jedis pipeline

### LINKED ISSUE #6434: jedis not support pipeline

- [X] Feature or performance improvement

___
### Question
- What do you want to know?
I look into jedis-2.x plugin, find that enhanceClass is redis.clients.jedis.Jedis.
But pipeline will directly invoke jedis client.
Jedis --invoke---> client
pipeline ---invoke-->client

There are two solution to solve this  problem
1. just enhance client
2. enhance both Jedis and Pipeline

xuanyu66:
### <Feature description>

- [X] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6434 .


wu-sheng:
Read plugin development doc, one step is missing. We need the plugin e2e test.

xuanyu66:
> Read plugin development doc, one step is missing. We need the plugin e2e test.

OK

wu-sheng (approved):
LGTM, once all tests passed.

wu-sheng (changes requested):
Plugin tests fail, please recheck.

wu-sheng:
I think you should run the test locally, there is no actual span generated. You could read the logs by yourself.

xuanyu66:
> I think you should run the test locally, there is no actual span generated. You could read the logs by yourself.

image
It look just OK for me

xuanyu66:
> I think you should run the test locally, there is no actual span generated. You could read the logs by yourself.

I ran the test ,found out it's compatible with version 2.9.0 2.8.2 2.7.3, but not compatible with version 2.6.3 and before.
Before 2.6.3, the parent class of pipeline has package private access, so it will bring java.lang.NoSuchMethodError

wu-sheng:
> I ran the test ,found out it's compatible with version 2.9.0 2.8.2 2.7.3, but not compatible with version 2.6.3 and before.
> Before 2.6.3, the parent class of pipeline has package private access, so it will bring java.lang.NoSuchMethodError

Have you tried to find other intercepting points? Such as intercept the parent methods directly? SkyWalking enhancement doesn't care it is a public or private method.

xuanyu66:
@wu-sheng It's done now

wu-sheng (approved):

Issue #6572, opened 2021-03-18
hbase-1.x-plugin is not compatible with 2.x hbase-client


nisiyong:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [x] Feature or performance improvement

___
#### Environment

- **SkyWalking Agent**: 8.2.0
- **HBase Client**: 2.1.0 (1.x works fine)

#### Agent Log

```
ERROR 2021-03-17 18:31:53:844 http-nio-8080-exec-4 InstMethodsInter : class[class org.apache.hadoop.hbase.client.HTable] before method[delete] intercept failure
java.lang.IllegalStateException: Exit span doesn't include meaningful peer information.
	at org.apache.skywalking.apm.agent.core.context.TracingContext.inject(TracingContext.java:163)
	at org.apache.skywalking.apm.agent.core.context.TracingContext.inject(TracingContext.java:143)
	at org.apache.skywalking.apm.agent.core.context.ContextManager.createExitSpan(ContextManager.java:126)
	at org.apache.skywalking.apm.plugin.hbase.HTableInterceptor.beforeMethod(HTableInterceptor.java:72)
	at org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstMethodsInter.intercept(InstMethodsInter.java:76)
	at org.apache.hadoop.hbase.client.HTable.delete(HTable.java)
	at org.apache.skywalking.apm.testcase.hbase.controller.HBaseController.hbaseCase(HBaseController.java:91)
```

## 1.x HTable contructor (from version 1.0.0 to 1.6.0)
source code: [HTable.java](https://github.com/apache/hbase/blob/rel/1.6.0/hbase-client/src/main/java/org/apache/hadoop/hbase/client/HTable.java)
```java
@InterfaceAudience.Private
public HTable(TableName tableName, final ClusterConnection connection,
    final ConnectionConfiguration tableConfig,
    final RpcRetryingCallerFactory rpcCallerFactory,
    final RpcControllerFactory rpcControllerFactory,
    final ExecutorService pool) throws IOException {
    ...
}
```

## 2.x HTable contructor (from version 2.0.0 to 2.1.9)
source code: [HTable.java](https://github.com/apache/hbase/blob/rel/2.0.0/hbase-client/src/main/java/org/apache/hadoop/hbase/client/HTable.java)
```java
@InterfaceAudience.Private
[...]

PR #6577, opened 2021-03-18
Update the agent HBase plugin to support HBase Client 2.x

nisiyong:
### Update the agent HBase plugin to support HBase Client 2.x

- [x] Relate the issue #6572 

In HBase Client 2.x, the `HTable` constructor function has a little difference after `2.2.0`.
So the agent plugin should have 3 `ConstructorInterceptPoint`, I have tested it all in `test/plugin/scenarios/hbase-scenario`, just change the dependency version.

The e2e test cases haven't committed, I will commit it later. Please take a look first, thanks.


apm-sniffer/apm-sdk-plugin/hbase-1.x-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hbase/HTable220Interceptor.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.Properties;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.client.Connection;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+import org.apache.skywalking.apm.util.StringUtil;
+
+public class HTable220Interceptor extends HTableInterceptor {
```
wu-sheng, line 29: Could you share the naming rule of 100, 200, and 220?

docs/en/setup/service-agent/java-agent/Plugin-list.md
```diff
@@ -23,7 +23,7 @@
 - grpc-1.x
 - gson-2.8.x
 - h2-1.x
-- hbase-1.x
+- hbase-1.x/2.x
```
wu-sheng, line 26: Is this correct? Not `1.x-2.x`?

wu-sheng (commented):
1. Supported-list.md should be updated
2. Plugin e2e tests should be updated.

apm-sniffer/apm-sdk-plugin/hbase-1.x-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hbase/HTable220Interceptor.java
nisiyong: 100 -> 1.0.0, means adapt version [1.0.0,1.6.0), 1.6.0 is the latest 1.x currently.
200 -> 2.0.0, means adapt version [2.0.0, 2.1.9), 2.1.9 is the latest 2.1.x currently.
220 -> 2.2.0, means adapt version [2.2.0, 2.4.1), 2.4.1 is the latest 2.x currently.

I use the start version as the name because we don't know what is the newest future version

docs/en/setup/service-agent/java-agent/Plugin-list.md
nisiyong: Just like `tomcat-7.x/8.x` in the same file.

apm-sniffer/apm-sdk-plugin/hbase-1.x-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hbase/define/HTableInstrumentation.java
```diff
@@ -47,11 +47,15 @@
 import static net.bytebuddy.matcher.ElementMatchers.named;
 import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
 import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
+import static org.apache.skywalking.apm.agent.core.plugin.bytebuddy.ArgumentTypeNameMatch.takesArgumentWithType;
 
 public class HTableInstrumentation extends ClassInstanceMethodsEnhancePluginDefine {
```
wu-sheng, line 52: Add the comments here about https://github.com/apache/skywalking/pull/6577#discussion_r596931939. We need to be clear about the naming rules.

wu-sheng:
Please don't close and reopen, You are making all CI process rerun.

nisiyong:
@wu-sheng 

After reading these docs as follows:
* [Contributing Guides](https://skywalking.apache.org/docs/main/v8.4.0/en/guides/readme/)
* [Plugin Development Guide](https://skywalking.apache.org/docs/main/v8.4.0/en/guides/java-plugin-development-guide/)
* [Apache SkyWalking Agent Test Tool Suite](https://github.com/apache/skywalking-agent-test-tool)

You said that update the e2e test, does it meaning I just need to update the `test/plugin/scenarios/hbase-scenario`?
And consider the version, I could remove this module, and add 3 new scenario:
- hbase-1.x-scenario (just rename hbase-scenario)
- hbase-2.0.x-scenario
- hbase-2.2.x-scenario

wu-sheng:
> You said that update the e2e test, does it meaning I just need to update the test/plugin/scenarios/hbase-scenario?
And consider the version, I could remove this module, and add 3 new scenario:

It depends on which way you prefer? Does the client API change much? Which makes you have to separate the test cases?

nisiyong:
> Does the client API change much? 

NO, but the HTable constructor function change in different versions.

> Which makes you have to separate the test cases?

According to the difference constructor, we need different versions of maven dependency, or we will miss some scenarios.

wu-sheng:
> NO, but the HTable constructor function change in different versions.

So, you can't use the same test codebase for different versions, right? If so, feel free to separate the cases.

nisiyong:
> you can't use the same test codebase for different versions, right?

The test code will be the same or similar, just the maven dependency version is different. We need to cover the HBase agent plugin for the different logic. In my opinion, it still should separate the cases.

wu-sheng:
> > you can't use the same test codebase for different versions, right?
> 
> The test code will be the same or similar, just the maven dependency version is different. We need to cover the HBase agent plugin for the different logic. In my opinion, it still should separate the cases.

Then you could add versions into the here, https://github.com/apache/skywalking/blob/master/test/plugin/scenarios/hbase-scenario/support-version.list. The test framework will cover the versions.

nisiyong:
> We need to be clear about the naming rules.

> 1. Supported-list.md should be updated
> 2. Plugin e2e tests should be updated.

I had added the comment about naming rules and update the `support-version.list`.
I had already tested the plugin in my local environment, and it works fine. 
@wu-sheng PTAL.

wu-sheng:
> 1. Supported-list.md should be updated

This missed.

wu-sheng (approved):

PR #6579, opened 2021-03-18
SkyWalking support Seata plugin

zhaoyuguang:
### Add an agent plugin to support Seata

https://github.com/seata/seata/issues/714
https://github.com/seata/seata/issues/884


wu-sheng:
@zhaoyuguang Could you share what is progress happens on the Seata side? This thing began and failed. In the previous discussion, we agree that we should have a plugin, and integrated agent+plugin on the Seata side(source codes and distribution).

zhaoyuguang:
> @zhaoyuguang Could you share what is progress happens on the Seata side? This thing began and failed. In the previous discussion, we agree that we should have a plugin, and integrated agent+plugin on the Seata side(source codes and distribution).

The old version of Seata did not have a transfer protocol similar to HTTP header, and it was easy to be incompatible when the client and server did not have probes together. After my communication with @slievrly, Seata now has a good header protocol in 1.4.x in synchronous mode.

wu-sheng:
> > @zhaoyuguang Could you share what is progress happens on the Seata side? This thing began and failed. In the previous discussion, we agree that we should have a plugin, and integrated agent+plugin on the Seata side(source codes and distribution).
> 
> The old version of Seata did not have a transfer protocol similar to HTTP header, and it was easy to be incompatible when the client and server did not have probes together. After my communication with @slievrly, Seata now has a good header protocol in 1.4.x in batch-mode.

This is good, and that was why @kezhenxu94 gave up. But even we have that, still we need Seata server-side plugin and installed agent to propagate the tracing context. From the previous discussion, those plugins should be hosted in the Seata repo.

wu-sheng:
Such as, in Apache DolphinScheduler's next release branch, https://github.com/apache/incubator-dolphinscheduler/tree/1.3.6-prepare/ext/skywalking. They have `ext/skywalking` folder for source codes. And their distribution includes our agent and necessary plugins.

docs/en/setup/service-agent/java-agent/Plugin-list.md
```diff
@@ -113,3 +113,4 @@
 - mssql-jtds-1.x
 - mssql-jdbc
 - apache-cxf-3.x
+- seata-1.4.x
```
wu-sheng, line 116: I think you don't need this. This list is a native plugin list. You should update `supported-list.md`.

wu-sheng:
When do you need this gets merge?

zhaoyuguang:
> When do you need this gets merge?

 Wait for me to finish the plugin in seata community.

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -123,6 +123,9 @@ metrics based on the tracing data.
   * [log4j](https://github.com/apache/log4j) 2.x
   * [log4j2](https://github.com/apache/logging-log4j2) 1.2.x 
   * [logback](https://github.com/qos-ch/logback) 1.2.x
+* Distributed Transaction Framework
+  * [seata](https://github.com/seata/seata) seata 1.4.x
+
```
wu-sheng, line 128: Remove this for now? This is going to be on the website directly.

zhaoyuguang:
@wu-sheng  Please check this MR, so Seata can use release version

docs/en/setup/service-agent/java-agent/Supported-list.md
zhaoyuguang: OK, had removed

CHANGES.md
```diff
@@ -12,6 +12,7 @@ Release Notes.
 * Improve `GlobalIdGenerator` performance.
 * Add an agent plugin to support elasticsearch7.
 * Add `jsonrpc4j` agent plugin.
+* Support Seata plugin.
```
```suggestion
* Add Seata in the component definition. Seata plugin hosts on Seata project.
```

wu-sheng (approved):

PR #6623, opened 2021-03-25
Timestamp in GlobalIdGenerator is implemented using ThreadLocalRandom

CalvinKirs:
- [ ] The benchmark result.
This is the official description (https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadLocalRandom.html), maybe I don’t need to add JMH Test for this


wu-sheng (approved):
As comments said, this random is nearly never used, but no harm.

PR #6631, opened 2021-03-28
Remove binary files and download them in CI process instead

kezhenxu94:
I found that the binary files are included in the source release, which should not


wu-sheng:
Should we put the version at the beginning of the file? I think we will need to upgrade one day or another.

wu-sheng (approved):

PR #6632, opened 2021-03-28
Remove the close_before_method and close_after_method parameters of custom-enhance-plugin to avoid memory leaks

### LINKED ISSUE #6301: custom-enhance-plugin bug

- [ ] Question or discussion
- [ ] Bug

___
### Question
- Q1:  There is a bug in version 8.3 of SW. If contextmanager creates span but does not execute "ContextManager.stopSpan ()", then threadLocal object's remove method will be skiped,  there  maybe existed a risk of ThreadLocal memory leak.

If so, there is a great risk:
In the customize-enhance-plugin, there is a set of configuration: close_ before_ method & close_ after_ method.
If user set the field  close_before_method=true，close_ after_ Method = false, this maybe cause "ContextManager.stopSpan ()" don't execute, and then ThreadLocal doesn't execute the remove method. After a long time, there will be a memory leak, right?

- Q2: What is the meaning of these two parameters(close_before_method & close_ after_ Method)? Is it necessary to exist?

liqiangz:
Remove the close_before_method and close_after_method parameters of custom-enhance-plugin to avoid memory leaks

If user set the field `close_before_method=false`，`close_ after_ Method = true`, this maybe cause `ContextManager.stopSpan()`don't execute.
Refer to #6301

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6301.


wu-sheng (commented):
I think document and plugin test should be updated too.

liqiangz:
@wu-sheng Now there are no two parameters in the document and in the test. 
image
image

wu-sheng:
I see. This is not documented, how so many people find this config. 
Anyway, I expect to wait for the confirmation from assigned reviewers.

zhaoyuguang:
It is true that these two parameters `close_before_method` `close_after_method` are not described in the documentation and unit tests. 
These two parameters allow for before span and after span not to be done in the same method, but at the same time, they introduce the risk of `memory leak`.
Removing this part of the code that is not described in the documentation reduces the risk of `memory leak`, LGTM

zhaoyuguang (approved):
LGTM

BFergerson (commented):
LGTM

Issue #6472, opened 2021-03-01
Evaluate the possibility to make PersistenceTimer running in multi threads


wu-sheng:
I am reviewing the core of OAP recently, one side to evaluate the requirements for the BanyanDB, and find the chance to improve the performance. #6411 is the first, this is the second one.

PersistenceTimer takes the core responsibility to assemble the requests and execute them to the storage. I noticed, the stage of assembly and execution is one-thread model. I remember these 2 stage causes many resources and latency. Considering the payload of OAP is usually not high, we could increase the number of thread to improve this.

@EvanLjp Could you share some stats data of self observability about these two?

FYI @apache/skywalking-committers

EvanLjp:
![image](https://user-images.githubusercontent.com/31562192/109500783-a5012000-7ad1-11eb-8ae5-5aa3b5062cbc.png)
![image](https://user-images.githubusercontent.com/31562192/109500788-a7fc1080-7ad1-11eb-9e8f-3052aa10a523.png)

wu-sheng:
@EvanLjp Back to you, have you tried to change `PersistenceTimer#extractDataAndSave`'s execution stage into the concurrency mode, such as using 2-4 threads to compare the performance? The 20s seem a little long.

EvanLjp:
let me have a try.

xuanyu66:
I want to konw, why metrics need to persist synchronously, can we just persist like record

wu-sheng:
> @wu-sheng I want to konw, why metrics need to persist synchronously, can we just persist like record

Simple answer is, NO, they need to be in the sync way.. And we can't afford the load to explain how the source codes work.

zifeihan:
@wu-sheng @xuanyu66 @EvanLjp I have also encountered the problem of es holding a lot of objects. in addition to multi-threaded writing, I also made an optimization here. I rewritten the buildBatchRequests method in MetricsPersistentWorker and added a parameter for batch writing. , Is to avoid frequent updates of hourly and day indicators. In addition, I made a time window to write one minute of data asynchronously (window period). If the current one minute has passed, then use synchronous writing. We have observed that the effect is good. hope some suggestions.

wu-sheng:
I am not following the codes. Could you submit a real design with a test benchmark?
This is a part of core, if we want to change this, it should be evaluted and reviewed carefully.

PR #6642, opened 2021-03-29
make sync metrics concurrency

EvanLjp:
https://github.com/apache/skywalking/issues/6472

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,6 +25,7 @@ Release Notes.
 * Support jedis pipeline in jedis-2.x-plugin.
 * Fix apm-toolkit-log4j-2.x-activation no trace Id in async log.
 * Replace hbase-1.x-plugin with hbase-1.x-2.x-plugin to adapt hbase client 2.x
+* Remove the close_before_method and close_after_method parameters of custom-enhance-plugin to avoid memory leaks.
 * Fix bug that springmvn-annotation-4.x-plugin, witness class does not exist in some versions.
 
 #### OAP-Backend
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -62,7 +62,7 @@ Release Notes.
 * Optimize the self monitoring grafana dashboard.
 * Enhance the export service.
 * Add function `retagByK8sMeta` and opt type `K8sRetagType.Pod2Service` in MAL for k8s to relate pods and services.
-* Make the flushing metrics operation concurrency.
+* Make the flushing metrics operation concurrent.
 
 #### UI
 * Update selector scroller to show in all pages.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,27 +4,21 @@ Release Notes.
 
 8.5.0
 ------------------
-
 #### Project
-
-* **Incompatible Change**. Indices and templates of ElasticSearch(6/7, including zipkin-elasticsearch7) storage option
-  have been changed.
+* **Incompatible Change**. Indices and templates of ElasticSearch(6/7, including zipkin-elasticsearch7) storage option have been changed.
 * Update frontend-maven-plugin to 1.11.0, for Download node x64 binary on Apple Silicon.
 * Add E2E test for VM monitoring that metrics from Prometheus node-exporter.
 * Upgrade lombok to 1.18.16.
 
 #### Java Agent
-
 * Remove invalid mysql configuration in agent.config.
-* Add net.bytebuddy.agent.builder.AgentBuilder.RedefinitionStrategy.Listener to show detail message when redefine errors
-  occur.
+* Add net.bytebuddy.agent.builder.AgentBuilder.RedefinitionStrategy.Listener to show detail message when redefine errors occur.
 * Fix ClassCastException of log4j gRPC reporter.
 * Fix NPE when Kafka reporter activated.
 * Enhance gRPC log appender to allow layout pattern.
 * Fix apm-dubbo-2.7.x-plugin memory leak due to some Dubbo RpcExceptions.
 * Fix lettuce-5.x-plugin get null host in redis sentinel mode.
-* Fix ClassCastException by making CallbackAdapterInterceptor to implement EnhancedInstance interface in the
-  spring-kafka plugin.
+* Fix ClassCastException by making CallbackAdapterInterceptor to implement EnhancedInstance interface in the spring-kafka plugin.
 * Fix NullPointerException with KafkaProducer.send(record).
 * Support config `agent.span_limit_per_segment` can be changed in the runtime.
 * Collect and report agent starting / shutdown events.
@@ -35,7 +29,6 @@ Release Notes.
 * Fix bug that springmvn-annotation-4.x-plugin, witness class does not exist in some versions.
 
 #### OAP-Backend
-
 * Allow user-defined `JAVA_OPTS` in the startup script.
 * Metrics combination API supports abandoning results.
 * Add a new concept "Event" and its implementations to collect events.
@@ -72,7 +65,6 @@ Release Notes.
 * Make the flushing metrics operation concurrency.
 
 #### UI
-
 * Update selector scroller to show in all pages.
 * Implement searching logs with date.
 * Add nodejs 14 compiling.
@@ -90,7 +82,6 @@ Release Notes.
 * Refactor the route mapping, Dynamically import routing components, Improve first page loading performance.
 
 #### Documentation
-
 * Polish documentation due to we have covered all tracing, logging, and metrics fields.
 * Adjust documentation about Zipkin receiver.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,6 +25,7 @@ Release Notes.
 * Support jedis pipeline in jedis-2.x-plugin.
 * Fix apm-toolkit-log4j-2.x-activation no trace Id in async log.
 * Replace hbase-1.x-plugin with hbase-1.x-2.x-plugin to adapt hbase client 2.x
+* Fix bug that springmvn-annotation-4.x-plugin, witness class does not exist in some versions.
 
 #### OAP-Backend
 * Allow user-defined `JAVA_OPTS` in the startup script.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,7 @@ Release Notes.
 * Replace hbase-1.x-plugin with hbase-1.x-2.x-plugin to adapt hbase client 2.x
 * Remove the close_before_method and close_after_method parameters of custom-enhance-plugin to avoid memory leaks.
 * Fix bug that springmvn-annotation-4.x-plugin, witness class does not exist in some versions.
+* Fix NullPointerException with `ReactiveRequestHolder.getHeaders`.
 
 #### OAP-Backend
 * Allow user-defined `JAVA_OPTS` in the startup script.
@@ -63,6 +64,9 @@ Release Notes.
 * Enhance the export service.
 * Add function `retagByK8sMeta` and opt type `K8sRetagType.Pod2Service` in MAL for k8s to relate pods and services.
 * Make the flushing metrics operation concurrent.
+* Fix ALS K8SServiceRegistry didn't remove the correct entry.
+* Using "service.istio.io/canonical-name" to replace "app" label to resolve Envoy ALS service name
+* Append the root slash(/) to getIndex and getTemplate requests in ES client
 
 #### UI
 * Update selector scroller to show in all pages.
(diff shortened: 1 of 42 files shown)


docs/en/setup/backend/configuration-vocabulary.md
```diff
@@ -39,6 +39,8 @@ core|default|role|Option values, `Mixed/Receiver/Aggregator`. **Receiver** mode
 | - | - | maxSizeOfNetworkAddressAlias|Max size of network address detected in the be monitored system.| - | 1_000_000|
 | - | - | maxPageSizeOfQueryProfileSnapshot|The max size in every OAP query for snapshot analysis| - | 500 |
 | - | - | maxSizeOfAnalyzeProfileSnapshot|The max number of snapshots analyzed by OAP| - | 12000 |
+| - | - | syncThreads|The number of threads used to synchronously refresh the metrics data to the storage. When the value is negative, the number of processors will used as the default value.| SW_CORE_SYNC_THREADS | -1 |
```
kezhenxu94, line 42:
```suggestion
| - | - | syncThreads|The number of threads used to synchronously refresh the metrics data to the storage. When the value is zero or negative, the number of processors will used as the default value.| SW_CORE_SYNC_THREADS | -1 |
```

oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -102,6 +102,10 @@ core:
     searchableTracesTags: ${SW_SEARCHABLE_TAG_KEYS:http.method,status_code,db.type,db.instance,mq.queue,mq.topic,mq.broker}
     # Define the set of log tag keys, which should be searchable through the GraphQL.
     searchableLogsTags: ${SW_SEARCHABLE_LOGS_TAG_KEYS:level}
+    # The number of threads used to synchronously refresh the metrics data to the storage. When the value is negative, the number of processors will used as the default value.
```
kezhenxu94, line 105:
```suggestion
    # The number of threads used to synchronously refresh the metrics data to the storage. When the value is zero or negative, the number of processors will used as the default value.
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleConfig.java
```diff
@@ -134,6 +134,24 @@
     @Getter
     private String searchableLogsTags = "";
 
+    /**
+     * The number of threads used to synchronously refresh the metrics data to the storage. When the value is negative,
```
kezhenxu94, line 138:
```suggestion
     * The number of threads used to synchronously refresh the metrics data to the storage. When the value is zero or negative,
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleConfig.java
```diff
@@ -134,6 +134,24 @@
...
+     * the number of processors will used as the default value.
+     *
+     * @since 8.5.0
+     */
+    @Getter
+    @Setter
+    private int syncThreads = 10;
```
kezhenxu94, line 145: If the user doesn't specify this, for example, use an old `application.yaml` file (without this key), the default thread count is 10, should it be processor count here too?

CHANGES.md
```diff
@@ -48,20 +48,21 @@ Release Notes.
...
-* Merge indices by Metrics Function and Meter Function in Elasticsearch Storage. 
+* Merge indices by Metrics Function and Meter Function in Elasticsearch Storage.
 * Fix receiver don't need to get itself when healthCheck
 * Remove group concept from AvgHistogramFunction. Heatmap(function result) doesn't support labels.
 * Support metrics grouped by scope labelValue in MAL, no need global same labelValue as before.
 * Add functions in MAL to filter metrics according to the metric value.
 * Optimize the self monitoring grafana dashboard.
 * Enhance the export service.
 * Add function `retagByK8sMeta` and opt type `K8sRetagType.Pod2Service` in MAL for k8s to relate pods and services.
+* Make the flushing metrics operation concurrency.
```
kezhenxu94, line 65:
```suggestion
* Make the flushing metrics operation concurrent.
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -71,7 +77,11 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
             "persistence_timer_bulk_execute_latency", "Latency of the execute stage in persistence timer",
             MetricsTag.EMPTY_KEY, MetricsTag.EMPTY_VALUE
         );
-
+        syncOperationThreadsNum = moduleConfig.getSyncThreads() > 0
+            ? moduleConfig.getSyncThreads()
+            : Runtime.getRuntime().availableProcessors();
```
kezhenxu94, line 82: I'd say this is better to be located in the `moduleConfig.getSyncThreads()`,

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -118,9 +128,17 @@ private void extractDataAndSave(IBatchDAO batchDAO) {
...
             try {
-                if (CollectionUtils.isNotEmpty(prepareRequests)) {
-                    batchDAO.synchronous(prepareRequests);
+                List<List<PrepareRequest>> partitions = Lists.partition(prepareRequests, maxSyncoperationNum);
+                CountDownLatch countDownLatch = new CountDownLatch(partitions.size());
+                for (final List<PrepareRequest> partition : partitions) {
+                    executorService.submit(() -> {
+                        if (CollectionUtils.isNotEmpty(partition)) {
+                            batchDAO.synchronous(partition);
+                        }
```
kezhenxu94, line 137: Let's enclose this inside a `try-catch` block, in case the `latch` is never counted down when exception occurs.

oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -102,6 +102,10 @@ core:
     searchableTracesTags: ${SW_SEARCHABLE_TAG_KEYS:http.method,status_code,db.type,db.instance,mq.queue,mq.topic,mq.broker}
     # Define the set of log tag keys, which should be searchable through the GraphQL.
     searchableLogsTags: ${SW_SEARCHABLE_LOGS_TAG_KEYS:level}
+    # The number of threads used to synchronously refresh the metrics data to the storage. When the value is zero or negative, the number of processors will used as the default value.
```
wu-sheng, line 105: Same expression issue.

docs/en/setup/backend/configuration-vocabulary.md
```diff
@@ -39,6 +39,8 @@ core|default|role|Option values, `Mixed/Receiver/Aggregator`. **Receiver** mode
 | - | - | maxSizeOfNetworkAddressAlias|Max size of network address detected in the be monitored system.| - | 1_000_000|
 | - | - | maxPageSizeOfQueryProfileSnapshot|The max size in every OAP query for snapshot analysis| - | 500 |
 | - | - | maxSizeOfAnalyzeProfileSnapshot|The max number of snapshots analyzed by OAP| - | 12000 |
+| - | - | syncThreads|The number of threads used to synchronously refresh the metrics data to the storage. When the value is zero or negative, the number of processors will used as the default value.| SW_CORE_SYNC_THREADS | -1 |
```
wu-sheng, line 42: Your default value is -1. So your explanation doesn't make sense. What exactly threads are you going to use in default?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleConfig.java
```diff
@@ -134,10 +134,31 @@
...
+    @Getter
+    @Setter
+    private int maxSyncOperationNum = 50000;
+
     public CoreModuleConfig() {
         this.downsampling = new ArrayList<>();
     }
 
+    public int getSyncThreads() {
+        return this.syncThreads > 0 ? this.syncThreads : Runtime.getRuntime().availableProcessors();
```
wu-sheng, line 159: Why choose the number of the CPU core? Any benchmark to support this?

wu-sheng: We have the straming based on this, including level 1 and level 2.
Why this is determined by this too?

wu-sheng:
@EvanLjp Codes basically are good. I am a little focusing on the default value, which is a good one. I hope you could have some data. Or @zifeihan could you?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleConfig.java
EvanLjp: I just set the maximum parallelism for this. But I have no experience with this.

wu-sheng: But as default value, mostly would be the real configuration for many users. I would say this is dangerous.

This value should relate to ES server CPU, memory and scale, rather than oap. Right?

EvanLjp: You are right. Any suggestions? Maybe the default value is 1 or 2?

wu-sheng: I am thinking about 1. Let's wait for @zifeihan feedback. As he seems having some private version experimenting this.

zifeihan (commented):
I think buildbatchRequests can also use multithreaded

zifeihan:
> @EvanLjp Codes basically are good. I am a little focusing on the default value, which is a good one. I hope you could have some data. Or @zifeihan could you?

Our oap is 4c, and the thread pool size is the Runtime.getRuntime().availableProcessors(), current value is not associated with es, Overall, the batch persistence speed during the peak period is several times faster.

wu-sheng:
> > @EvanLjp Codes basically are good. I am a little focusing on the default value, which is a good one. I hope you could have some data. Or @zifeihan could you?
> 
> Our oap is 4c, and the thread pool size is the Runtime.getRuntime().availableProcessors(), current value is not associated with es, Overall, the batch persistence speed during the peak period is several times faster.

I know it is helpful, that is why I raised the issue. The key here is what is a good default way. And whether we need multi threads in prepare stage. From @EvanLjp provided data, we don't. @zifeihan Do you have some recommends about this? And any data to prove?

zifeihan:
> I know it is helpful, that is why I raised the issue. The key here is what is a good default way. And whether we need multi threads in prepare stage. From @EvanLjp provided data, we don't. @zifeihan Do you have some recommends about this? And any data to prove?

According to my data, We have more than 10 billion segment and more than 80 billion metrics data every day, before the modification, at the peak, the preparation time was close to one minute, and some reached several minutes. After the modification, most of the time is less than 10 seconds.  perpare's thread pool and batch save are one, and the number of threads is Runtime.getRuntime().availableProcessors()

EvanLjp:
I think we could use many threads to parallel in the preparation phase. But How many threads do we need in the storage phase? @zifeihan May be should be different.
image

wu-sheng:
> I think we could use many threads to parallel in the preparation phase. But How many threads do we need in the storage phase? @zifeihan May be should be different.
> image

Let's make this as 2. And not according to the CPU core. @zifeihan any objection?

EvanLjp:
> > I think we could use many threads to parallel in the preparation phase. But How many threads do we need in the storage phase? @zifeihan May be should be different.
> > image
> 
> Let's make this as 2. And not according to the CPU core. @zifeihan any objection?

Do we need this thread to do some preprocesses?

wu-sheng:
> > > I think we could use many threads to parallel in the preparation phase. But How many threads do we need in the storage phase? @zifeihan May be should be different.
> > > image
> > 
> > 
> > Let's make this as 2. And not according to the CPU core. @zifeihan any objection?
> 
> Do we reuse these threads to do some preprocesses?

From your statistics, it seems don't need. Maybe consider to adding later if it is really needed.

zifeihan:
> Let's make this as 2. And not according to the CPU core. @zifeihan any objection?

No problem.

EvanLjp:
> From your statistics, it seems don't need. Maybe consider to adding later if it is really needed.

no problem

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleConfig.java
```diff
@@ -134,10 +134,30 @@
...
+     */
+    @Getter
+    @Setter
+    private int maxSyncOperationNum = 50000;
+
     public CoreModuleConfig() {
         this.downsampling = new ArrayList<>();
     }
 
+    public int getSyncThreads() {
```
wu-sheng, line 157: I thinl this get logic should be removed?

wu-sheng (approved):
@kezhenxu94 Please recheck.

zifeihan (approved):
LGTM

kezhenxu94 (approved):
LGTM

Issue #6648, opened 2021-03-29
NullPointerException with nonexistent Header when call "ReactiveRequestHolder.getHeaders"


mm23504570:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ * ] Bug
- [ ] Requirement
- [ ] Feature or performance improvement
___
### Bug
- Which version of SkyWalking, OS, and JRE?
8.4.0
- Which company or project?

- What happened?
If possible, provide a way to reproduce the error. e.g. demo application, component version.

- What do you want to know?
code
```
public class ReactiveRequestHolder implements RequestHolder {
@Override
    public Enumeration<String> getHeaders(final String headerName) {
        return Collections.enumeration(this.serverHttpRequest.getHeaders().get(headerName));
    }
}
```

```
java.lang.NullPointerException: null
    at java.util.Collections$3.<init>(Collections.java:5314)
    at java.util.Collections.enumeration(Collections.java:5313)
    at org.apache.skywalking.apm.plugin.spring.mvc.commons.ReactiveRequestHolder.getHeaders(ReactiveRequestHolder.java:40)
    at org.apache.skywalking.apm.plugin.spring.mvc.commons.interceptor.AbstractMethodInterceptor.lambda$collectHttpHeaders$0(AbstractMethodInterceptor.java:257)
    at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:176)
    at java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
    at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
    at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
    at java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
    at java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
    at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
    at java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:497)
    at org.apache.skywalking.apm.plugin.spring.mvc.commons.interceptor.AbstractMethodInterceptor.collectHttpHeaders(AbstractMethodInterceptor.java:258)
[...]

PR #6651, opened 2021-03-29
Fix NullPointerException when call "ReactiveRequestHolder.getHeaders"

mm23504570:
#6648


apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-commons/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/commons/ReactiveRequestHolder.java
```diff
@@ -37,7 +38,11 @@ public String getHeader(final String headerName) {
 
     @Override
     public Enumeration<String> getHeaders(final String headerName) {
-        return Collections.enumeration(this.serverHttpRequest.getHeaders().get(headerName));
+        List<String> values = this.serverHttpRequest.getHeaders().get(headerName);
+        if (values == null) {
+            return null;
```
wu-sheng, line 43: Is null expected  when use this method? You used to say returning an empty list.

wu-sheng (approved):

PR #6654, opened 2021-03-30
Resolve envoy service name from a different position

### LINKED ISSUE #6647: Envoy ALS mx-mesh should resolve service name from another position

- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
___
Since Istio injects ` service.istio.io/canonical-name` to the pod as the service name, OAP Envoy receiver should update `https://github.com/apache/skywalking/blob/master/oap-server/server-bootstrap/src/main/resources/metadata-service-mapping.yaml` to replace `LABELS.app` with this canonical label.

hanahmily:
* Using "service.istio.io/canonical-name" to replace "app" label to resolve service name
 * Improving FieldsHelper to support dot in labels

Fixed #6647

### DIFF
--- a/oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/mx/FieldsHelper.java
+++ b/oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/mx/FieldsHelper.java
@@ -39,6 +39,9 @@ import org.apache.skywalking.oap.server.receiver.envoy.als.ServiceMetaInfo;
 import org.yaml.snakeyaml.Yaml;
 
 @Slf4j
+/**
+ * FieldsHelper
+ */
 public enum FieldsHelper {
     SINGLETON;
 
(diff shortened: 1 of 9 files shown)


oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/mx/FieldsHelper.java
```diff
@@ -82,7 +85,24 @@ public void init(final InputStream inputStream,
             final StringBuffer serviceNamePattern = new StringBuffer();
             while (m.find()) {
                 final String property = m.group("property");
-                flatBuffersFieldNames.add(Splitter.on('.').omitEmptyStrings().splitToList(property));
+                List<String> tokens = Splitter.on('.').omitEmptyStrings().splitToList(property);
```
kezhenxu94, line 88: Just did a quick Google search, replacing the `.` with this delimiter `\.(?=([^"]*"[^"]*")*[^"]*$)` can do the trick, but you can make the call whether to replace it or not

kezhenxu94 (dismissed):
LGTM if tests pass

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/mx/FieldsHelper.java
wu-sheng: I prefer the current way. It is more clear.

wu-sheng:
@hanahmily Code style issue? Please recheck.

wu-sheng (approved):

PR #6665, opened 2021-04-01
Fix Agent Plugin "asynchttpclient" not record the status code of the …

darcydai:
…response

### Fix <agent asynchttpclient plugin not record the response status code>
     add a tag to async span when status received method be invoked


wu-sheng (commented):
You need to update the changelog.

darcydai:
> You need to update the changelog.
 Hi sheng, I have added change log

ascrutae (approved):

ascrutae (commented):
You should also changed the expected data of async http client scenario. Link:   https://github.com/apache/skywalking/blob/master/test/plugin/scenarios/asynchttpclient-scenario/config/expectedData.yaml

ascrutae (changes requested):
You should also changed the expected data of async http client scenario. Link: https://github.com/apache/skywalking/blob/master/test/plugin/scenarios/asynchttpclient-scenario/config/expectedData.yaml

darcydai:
> You should also changed the expected data of async http client scenario. Link: https://github.com/apache/skywalking/blob/master/test/plugin/scenarios/asynchttpclient-scenario/config/expectedData.yaml

ok I will do it tomorrow，and run 'asynchttpclient-scenario' in local env

ascrutae (approved):
LGTM

PR #6670, opened 2021-04-01
Allow exclusive mode for WitnessMethod

lujiajing1126:
### Allow exclusive mode for WitnessMethod
- [x] It is trivial.
- [x] Not related to UI

From skywalking 8.4.0, witness method has been introduced. This feature enables those plugins to conduct enhancements only if some specific methods are found.

In our scenarios, we would like to enhance some classes in Kafka client module, but only if `public Object metricValue()` can be found in the `KafkaMetric` class (which means Kafka version >= 1.0.0) since we are using this method in plugin.

Also, we want to make enhancements targeting legacy Kafka versions (< 1.0.0) with other different interceptors.

The most intuitive way is to do the opposite, searching for `KafkaMetric` class but without `public Object metricValue()` method. However, this method is not support yet in Skywalking agent. Thus, this PR is to provide this possibility.

Currently, we can do this in the following ways,

- use `MethodUtil.isMethodExist` during the runtime like what we have done in Tomcat, Jetty plugins for servlet version compatibility.
- in this specific case, we can check `@Deprecation` annotation on `@Deprecated public double value()` method. But this method will be totally removed in the next major release due to [KAFKA-12573](https://github.com/apache/kafka/pull/10425)

I suppose this PR can make improvement to the `WitnessMethod` feature.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/AbstractClassEnhancePluginDefine.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/AbstractClassEnhancePluginDefine.java
@@ -67,7 +67,7 @@ public abstract class AbstractClassEnhancePluginDefine {
         String[] witnessClasses = witnessClasses();
         if (witnessClasses != null) {
             for (String witnessClass : witnessClasses) {
-                if (!finder.exist(witnessClass, classLoader)) {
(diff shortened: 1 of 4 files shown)


wu-sheng:
If you have searched the issue and mail list, this debate existed for a long time. 
We are not going to accept exclusive rules. to keep codes clear.
Because like your case, exclusive seems an easy way, but the newer versions still be easy to break plugins.

PR #6671, opened 2021-04-01
Fix springmvc reactive api can't collect HTTP statusCode.

### LINKED ISSUE #5905: skywalking 8.3 add  springmvc plugin reactive api support , but that cant collect HTTP statusCode

liqiangz:
    The framework change httpCode after AbstractMethodInterceptor.  Replace it with 
    org.apache.skywalking.apm.plugin.spring.mvc.v5.InvokeInterceptor.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #5905.


apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/v5/InvokeInterceptor.java
```diff
@@ -19,12 +19,16 @@
...
+import org.apache.skywalking.apm.agent.core.context.tag.Tags;
+import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
 import org.apache.skywalking.apm.plugin.spring.mvc.commons.ReactiveRequestHolder;
 import org.apache.skywalking.apm.plugin.spring.mvc.commons.ReactiveResponseHolder;
+import org.springframework.http.HttpStatus;
 import org.springframework.web.server.ServerWebExchange;
+import reactor.core.publisher.Mono;
```
wu-sheng, line 31: What is this import for?
I notice you added one more dependency, is this always used in Spring classpath?

liqiangz: I will change the referenced package to spring-boot-starter-webflux

ascrutae: That will woks in spring boot webflux, But I am afraid that it causes ClassNotFoundException exception if someone only uses spring web framework.

liqiangz: @ascrutae InvokeInterceptor enhance `org.springframework.web.reactive.result.method.InvocableHandlerMethod`. This class only appears in the webflux package.

liqiangz: done

wu-sheng (commented):
Still, I don't think Spring 5 has to depend on webflux. Tomcat is available, isn't it?

liqiangz:
@wu-sheng InvokeInterceptor enhance org.springframework.web.reactive.result.method.InvocableHandlerMethod. This class only appears in the webflux package.  This method will only be executed when the user uses webflux, if it is not used, it should not affect the original process.

wu-sheng:
> @wu-sheng `InvokeInterceptor` enhance `org.springframework.web.reactive.result.method.InvocableHandlerMethod`. This class only appears in the webflux package. This method will only be executed when the user uses webflux, if it is not used, it should not affect the original process.

@ascrutae Could you verify the codes?

ascrutae:
Could you move the [webflux plugin](https://github.com/apache/skywalking/tree/master/apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin) into the plugins folder to resolve this issue?

liqiangz:
@ascrutae  When using springmvc reactive api, two spans will be created by mistake. 
 I think we can only  webflux plugin  when not  using spring mvc api
image

ascrutae:
> @ascrutae When using springmvc reactive api, two spans will be created by mistake.
> I think we can only activate webflux plugin when not using spring mvc reactive api.
> And when use webflux plugin, the trace is broken.
> image
> image

Could you give me a simple demo to reproduce this issue ?

liqiangz:
@ascrutae Move the webflux plugin into the plugins folder and run springmvc-reactive-scenario demo

liqiangz:
@ascrutae I feel that springmvc 5 plugin and webflux plugin should be two levels. Just like springmvc5 plugin and tomcat plugin. So the problem of the springmvc5 plugin that cannot collect http code should also be fixed. This issue should have nothing to do with the webflux plugin, right?

ascrutae:
You're right,  the spring plugin should work single.

test/plugin/scenarios/springmvc-reactive-scenario/config/expectedData.yaml
```diff
@@ -44,7 +73,31 @@ segmentItems:
       startTime: not null
       endTime: not null
       componentId: 14
+      isError: false
+      tags:
+      - {key: url, value: not null}
+      - {key: http.method, value: GET}
```
ascrutae, line 79: Does it miss the status_code  tag?

liqiangz: @ascrutae  `status_code` is recorded When `httpStatus >= 400` in spring plugins.   `{GET}/testcase/error`  will `record status_code=500`
```java
            HttpStatus httpStatus = exchange.getResponse().getStatusCode();
            if (httpStatus != null && httpStatus.isError()) {
                span.errorOccurred();
                Tags.STATUS_CODE.set(span, Integer.toString(httpStatus.value()));
            }
```

ascrutae (approved):
LGTM

PR #6678, opened 2021-04-03
doc: added contributor over time

juzhiyuan:
Hi, I have not seen any section that matches this PR, then I remove all of them 🤔 

This PR aims to add the Contributor Over Time for Apache SkyWalking since 2015-11-09, I hope SkyWalking will like it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,9 @@ Find the [demo](https://skywalking.apache.org/#demo) and [screenshots](https://s
 Hundreds of companies and organizations use SkyWalking for research, production, and commercial product.
 Visit our [website](http://skywalking.apache.org/#users) to find the user page.
 
+# Contributor Over Time
+
+![image](https://user-images.githubusercontent.com/2106987/113481423-0bf86700-94cc-11eb-97e9-239deb0ae9f5.png)
 
 # Landscapes
 


wu-sheng:
I think you should add this to the website, if you like the trend graph from time to time.
https://skywalking.apache.org/team/

wu-sheng:
The team page will be updated every day and once new PR merged in the website repo, like we did for the document.

juzhiyuan:
ok, let me do it.

wu-sheng:
When you do this, please make sure it is either attched on every project, or aggregating all projects data into one graph, considering SkyWalking has many subprojects.
Also, we need this to rebuild every time.

juzhiyuan:
> When you do this, please make sure it is either attched on every project, or aggregating all projects data into one graph, considering SkyWalking has many subprojects.
> Also, we need this to rebuild every time.

Good catch! Let me generate a graph contains all subprojects :)

And we will support one link target to a dynamic graph, so we don't have to update that graph every time.

juzhiyuan:
image

Yes, when I try to add all projects, I find that it has too many projects 😂 Could we attach this graph for every project? (on the README file instead of website).

juzhiyuan:
It's not clear for repos that don't have too many contributors when combining all of the projects.

wu-sheng:
Do you mean api7 host service? Be careful anout it.
ASF TLP relies on commercial vendor website, which could be a concern by PMC, even ASF at policy level.

juzhiyuan:
Yes, this graph is generated by API7, and for now, I will generate every graph by that service, without any commercial information actually 🤔 

> And we will support one link target to a dynamic graph, so we don't have to update that graph every time.

This feature is still in progress :)

wu-sheng:
> Yes, this graph is generated by API7, but for now, I will generate every graph by that service, without any commercial information actually 🤔
> 
> > And we will support one link target to a dynamic graph, so we don't have to update that graph every time.
> 
> This feature is still in progress :)

Yes, but you will add it into the website pipeline or accessing api7 service address every time when user accesses this team page.
Either way, you need the PMC approval(vote) to accept this.

wu-sheng:
Also, I am guessing you mean the 2nd way, then your service speed to global people matters.

wu-sheng:
> image
> 
> Yes, when I try to add all projects, I find that it has too many projects 😂 Could we attach this graph for every project? (on the README file instead of website).

We removed this kind of graph including star trend from the readme. SkyWalking is focusing on GitHub for development, website for users. @Jtrust is leading all things, we have no plan to add them back.

wu-sheng:
> It's not clear for repos that don't have too many contributors when combining all of the projects.

I prefer to add this to the whole ecosystem, rather than a repo. For SkyWalking, contributions to any repo are same, and we appreciated them in same.

juzhiyuan:
😂 ok, got it. But making a chart that contains multiple repos' trends is a little hard in this style. Let us reconsider.

wu-sheng:
> 😂 ok, got it. But making a chart that contains multiple repos' trends is a little hard in this style. Let us reconsider.

I just recommend to use one line in the graph. I am feeling you are thinking one line per repo

PR #6685, opened 2021-04-05
Support @Trace, @Tag and @Tags work for static methods

### LINKED ISSUE #6640: apm-toolkit-trace: @Trace, @Tag and @Tags do not work for static methods

- [x] Question or discussion
- [ ] Bug
- [ ] Requirement

___
I added `Trace` annotation to a static method. The result shows that `Trace` annotation does not work for static methods. A browse of `apm-toolkit-trace-activiation` module shows that `TagAnnotationActivation` and `TraceAnnotationActivation` only do method interception for static methods.

Does this work as designed? If not, I am willing to add support for static methods.

liqiangz:
@Trace, @Tag and @Tags do not work for static methods. Add StaticMethodsAroundInterceptor to support static methods.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6640.


wu-sheng:
Hi, thanks for submitting. I think there is one decision to be made. Today, we don't support to use `this.field` as tag value, so, trace/tag instance and static methods look the same. But consider those cases, they are not.
What do you think about separating these 2 cases, and we may need to add support for `this.field` one day.

wu-sheng:
Reference doc, http://skywalking.apache.org/docs/main/v8.4.0/en/setup/service-agent/java-agent/application-toolkit-trace/. We only support using parameter and returnObj as tag value.

liqiangz:
@wu-sheng  In most cases, these two annotations are the same. I feel separating the two cases for `this.filed`  increases the cost of use. If we need to support `this.field`  one day, maybe we can simply set tag value to `null` for static methods.

wu-sheng:
> I feel separating the two cases for this.filed will increase the cost of use

OK, make sense to me. Let's focus on the details.

apm-sniffer/apm-toolkit-activation/apm-toolkit-trace-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/trace/TagStaticAnnotationActivation.java
```diff
@@ -0,0 +1,71 @@
...
+import static org.apache.skywalking.apm.agent.core.plugin.match.logical.LogicalMatchOperation.not;
+import static org.apache.skywalking.apm.agent.core.plugin.match.logical.LogicalMatchOperation.or;
+import static org.apache.skywalking.apm.toolkit.activation.trace.TagAnnotationActivation.TAGS_ANNOTATION;
+import static org.apache.skywalking.apm.toolkit.activation.trace.TagAnnotationActivation.TAG_ANNOTATION;
+import static org.apache.skywalking.apm.toolkit.activation.trace.TraceAnnotationActivation.TRACE_ANNOTATION;
+
+/**
+ * The static of tag annotation activation.
+ */
+public class TagStaticAnnotationActivation extends ClassStaticMethodsEnhancePluginDefine {
```
wu-sheng, line 40: You don't need a new activation, you could enhance the existing one by `extends ClassEnhancePluginDefine`.

liqiangz: done

wu-sheng:
FYI @yaojingguo

wu-sheng (approved):
LGTM. Wait for @ascrutae to recheck.

ascrutae (approved):

PR #6715, opened 2021-04-08
Add `trace_segment_ref_limit_per_span` configuration mechanism to avoid OOM

zifeihan:
### Add `trace_segment_ref_limit_per_span` configuration mechanism to avoid OOM

There maybe an exception in some plugins, which causes the span to fail to exit, which results in the ThreadLocal not being cleaned up, and this data of TraceSegmentRef is increasing. This configuration will constrain the memory size occupied by TraceSegmentRef and avoid OOM. Note that this configuration is only used to avoid oom, and cannot solve the problem that ThreadLocal cannot be cleaned from the root cause.


wu-sheng:
It will be hold until 8.5.0 release tagged.

CHANGES.md
```diff
@@ -33,6 +33,7 @@ Release Notes.
 * Fix bug that asynchttpclient plugin does not record the response status code.
 * Fix spanLayer is null in optional plugin(gateway-2.0.x-plugin gateway-2.1.x-plugin).
 * Support @Trace, @Tag and @Tags work for static methods.
+* Add `trace_segment_ref_limit_per_segment` configuration mechanism to avoid OOM.
```
wu-sheng, line 36: We need to clear the changelog for 8.6.0

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ids/DistributedTraceIds.java
```diff
@@ -37,6 +38,13 @@ public void append(DistributedTraceId distributedTraceId) {
         if (relatedGlobalTraces.size() > 0 && relatedGlobalTraces.getFirst() instanceof NewDistributedTraceId) {
             relatedGlobalTraces.removeFirst();
         }
+        /*
+         * There maybe an exception in some plugins, which causes the span to fail to exit,
+         * which results in the ThreadLocal not being cleaned up, and this data is increasing.
+         */
+        if (relatedGlobalTraces.size() >= Config.Agent.TRACE_SEGMENT_REF_LIMIT_PER_SEGMENT) {
```
wu-sheng, line 45: Check the codes of `getRelatedGlobalTraces`, it seems we need one reference here only?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/TraceSegment.java
```diff
@@ -89,6 +89,13 @@ public void ref(TraceSegmentRef refSegment) {
         if (refs == null) {
             refs = new LinkedList<>();
         }
+        /*
+         * There maybe an exception in some plugins, which causes the span to fail to exit,
+         * which results in the ThreadLocal not being cleaned up, and this data is increasing.
+         */
+        if (refs.size() >= Config.Agent.TRACE_SEGMENT_REF_LIMIT_PER_SEGMENT) {
```
wu-sheng, line 96: Similar to https://github.com/apache/skywalking/pull/6715/files#r610564498, check the code, please.

wu-sheng:
If you want to avoid ref-OOM, take a look at `AbstractTracingSpan#refs`, which is the real place hosting the reference obj

zifeihan:
> If you want to avoid ref-OOM, take a look at `AbstractTracingSpan#refs`, which is the real place hosting the reference obj

Get it, I will try to optimize this.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -66,6 +66,12 @@
          */
         public static String IGNORE_SUFFIX = ".jpg,.jpeg,.js,.css,.png,.bmp,.gif,.ico,.mp3,.mp4,.html,.svg";
 
+        /**
+         * The max number of TraceSegmentRef in a single span. Through this config item, SkyWalking keep your application
+         * memory cost estimated.
```
wu-sheng, line 71:
```suggestion
         * The max number of TraceSegmentRef in a single span to keep memory cost estimatable.
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/TraceSegment.java
```diff
@@ -38,30 +38,15 @@
...
 
-    /**
-     * The <code>relatedGlobalTraces</code> represent a set of all related trace. Most time it contains only one
-     * element, because only one parent {@link TraceSegment} exists, but, in batch scenario, the num becomes greater
-     * than 1, also meaning multi-parents {@link TraceSegment}. <p> The difference between
-     * <code>relatedGlobalTraces</code> and {@link #refs} is: {@link #refs} targets this {@link TraceSegment}'s direct
-     * parent, <p> and <p> <code>relatedGlobalTraces</code> targets this {@link TraceSegment}'s related call chain, a
-     * call chain contains multi {@link TraceSegment}s, only using {@link #refs} is not enough for analysis and ui.
-     */
-    private DistributedTraceIds relatedGlobalTraces;
```
wu-sheng, line 64: You removed important comments but add nothing. You need to fix this.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/TraceSegment.java
```diff
@@ -86,19 +70,14 @@ public TraceSegment() {
...
     }
 
     /**
-     * Establish the line between this segment and all relative global trace ids.
+     * Establish the line between this segment and the relative global trace id.
      */
-    public void relatedGlobalTraces(DistributedTraceId distributedTraceId) {
-        relatedGlobalTraces.append(distributedTraceId);
+    public void relatedGlobalTrace(DistributedTraceId distributedTraceId) {
+        this.relatedGlobalTraceId = distributedTraceId;
```
wu-sheng, line 80: Check how old logic works, this overriding is incorrect.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/TraceSegment.java
```diff
@@ -86,19 +70,14 @@ public TraceSegment() {
      * @param refSegment {@link TraceSegmentRef}
      */
     public void ref(TraceSegmentRef refSegment) {
-        if (refs == null) {
-            refs = new LinkedList<>();
-        }
-        if (!refs.contains(refSegment)) {
-            refs.add(refSegment);
-        }
+        this.traceSegmentRef = refSegment;
```
wu-sheng, line 73: Check how old logic works, this overriding is incorrect.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/TraceSegment.java
```diff
@@ -172,7 +154,7 @@ public SegmentObject transform() {
 
     @Override
     public String toString() {
-        return "TraceSegment{" + "traceSegmentId='" + traceSegmentId + '\'' + ", refs=" + refs + ", spans=" + spans + ", relatedGlobalTraces=" + relatedGlobalTraces + '}';
+        return "TraceSegment{" + "traceSegmentId='" + traceSegmentId + '\'' + ", traceSegmentRef=" + traceSegmentRef + ", spans=" + spans + ", relatedGlobalTraceId=" + relatedGlobalTraceId + '}';
```
wu-sheng, line 157: Remove `relatedGlobalTraceId` from `toString`, as it only keeps one.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/TraceSegment.java
```diff
@@ -38,30 +38,15 @@
...
-    /**
-     * The <code>relatedGlobalTraces</code> represent a set of all related trace. Most time it contains only one
-     * element, because only one parent {@link TraceSegment} exists, but, in batch scenario, the num becomes greater
-     * than 1, also meaning multi-parents {@link TraceSegment}. <p> The difference between
-     * <code>relatedGlobalTraces</code> and {@link #refs} is: {@link #refs} targets this {@link TraceSegment}'s direct
-     * parent, <p> and <p> <code>relatedGlobalTraces</code> targets this {@link TraceSegment}'s related call chain, a
-     * call chain contains multi {@link TraceSegment}s, only using {@link #refs} is not enough for analysis and ui.
-     */
-    private DistributedTraceIds relatedGlobalTraces;
+    private TraceSegmentRef traceSegmentRef;
```
wu-sheng, line 47: Why rename? And put a duplicated `traceSegment` in the name?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/AbstractTracingSpan.java
```diff
@@ -301,6 +302,13 @@ public void ref(TraceSegmentRef ref) {
         if (refs == null) {
             refs = new LinkedList<>();
         }
+        /*
+         * There maybe an exception in some plugins, which causes the span to fail to exit,
+         * which results in the ThreadLocal not being cleaned up, and the refs data is increasing.
```
wu-sheng, line 307:
```suggestion
         * Provide the OOM protection if the entry span hosts too many references.
```
You should not do coding for bugs. This OOM protection could happen w/o bug, such as consuming 1k+ message as a batch, isn't it?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/TraceSegment.java
```diff
@@ -39,13 +39,13 @@
...
-     * The refs of parent trace segments, except the primary one. For most RPC call, {@link #refs} contains only one
+     * The refs of parent trace segments, except the primary one. For most RPC call, {@link #traceSegmentRef} contains only one
      * element, but if this segment is a start span of batch process, the segment faces multi parents, at this moment,
-     * we use this {@code #refs} to link them.
+     * we only related the first parent TraceSegment.
      * <p>
      * This field will not be serialized. Keeping this field is only for quick accessing.
      */
-    private List<TraceSegmentRef> refs;
+    private TraceSegmentRef traceSegmentRef;
```
wu-sheng, line 48: Again, why rename?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/TraceSegment.java
```diff
@@ -39,13 +39,13 @@
     private String traceSegmentId;
 
     /**
-     * The refs of parent trace segments, except the primary one. For most RPC call, {@link #refs} contains only one
+     * The refs of parent trace segments, except the primary one. For most RPC call, {@link #traceSegmentRef} contains only one
      * element, but if this segment is a start span of batch process, the segment faces multi parents, at this moment,
-     * we use this {@code #refs} to link them.
+     * we only related the first parent TraceSegment.
```
wu-sheng, line 44:
```suggestion
     * we only cached the first parent segment reference.
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/TraceSegment.java
```diff
@@ -121,16 +116,19 @@ public String getTraceSegmentId() {
...
     }
 
-    public boolean hasRef() {
-        return !(refs == null || refs.size() == 0);
-    }
-
+    /**
+     * For testing use only.
+     */
     public List<TraceSegmentRef> getRefs() {
```
wu-sheng, line 122: Why this API doesn't change? You are optimizing the implementation, you should clear the legacy codes once for all. Otherwise, people are going to be confused when reading this next time.

zifeihan: done.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/AbstractTracingSpan.java
zifeihan: done.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/TraceSegment.java
zifeihan: fixed.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/TraceSegment.java
zifeihan: done

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/TraceSegment.java
```diff
@@ -39,13 +38,13 @@
     private String traceSegmentId;
 
     /**
-     * The refs of parent trace segments, except the primary one. For most RPC call, {@link #refs} contains only one
+     * The refs of parent trace segments, except the primary one. For most RPC call, {@link #ref} contains only one
      * element, but if this segment is a start span of batch process, the segment faces multi parents, at this moment,
-     * we use this {@code #refs} to link them.
+     * we only cached the first parent segment reference.
```
wu-sheng, line 43:
```suggestion
     * we only cache the first parent segment reference.
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/AbstractTracingSpan.java
```diff
@@ -301,6 +302,12 @@ public void ref(TraceSegmentRef ref) {
         if (refs == null) {
             refs = new LinkedList<>();
         }
+        /*
+         * Provide the OOM protection if the entry span hosts too many references.
+         */
+        if (refs.size() >= Config.Agent.TRACE_SEGMENT_REF_LIMIT_PER_SPAN) {
```
wu-sheng, line 308:
```suggestion
        if (refs.size() == Config.Agent.TRACE_SEGMENT_REF_LIMIT_PER_SPAN) {
```
What is the case should be `>`? Unless there is a bug here, or faces face condition, but actually race condition is a bug

zifeihan: Fixed.

wu-sheng (approved):
LGTM. @kezhenxu94 @ascrutae Do a recheck please.

kezhenxu94 (approved):
LGTM

ascrutae (approved):
LGTM

PR #6729, opened 2021-04-11
GlobalIdGenerator will produce a negative id when the time shift back

### LINKED ISSUE #6728: GlobalIdGenerator will produce a negative id when the time shift back

- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement

___
8.4.0 MacOS JDK8

Random will return a negative value, while this is an easy way to solve this.
https://github.com/apache/skywalking/blob/f3ca0d0ea52bf05df87e9b1e268b7f5d69423a88/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ids/GlobalIdGenerator.java#L78

nickwongwong:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6728.


wu-sheng:
I am still waiting for you to explain why positive or negative matters.

nickwongwong:
Here are my comments. #6729

nickwongwong:
Here are my comments. 
https://github.com/apache/skywalking/issues/6728

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ids/GlobalIdGenerator.java
```diff
@@ -73,9 +73,19 @@ private long timestamp() {
             long currentTimeMillis = System.currentTimeMillis();
 
             if (currentTimeMillis < lastTimestamp) {
+                long overTime = lastTimestamp - currentTimeMillis;
+                if (overTime <= 5) {
+                    try {
+                        Thread.sleep(overTime);
```
wu-sheng, line 79: We can't afford this sleep.

nickwongwong: image
Change `overTime` to 1ms.
Use `LockSupport.parkNanos` instead of `Thread.sleep`.
image

Actually, the `NTP` time offset(time shift) is very small in Linux servers.
`1ms` may cover most cases, while keeping good performance.
image

If ok, I will modify this PR.

wu-sheng: No, `1ms` is not acceptable. The agent can't block the user process, no matter in any case.

wu-sheng: APM is a low priority system, `1ms` is not in a good performance.

nickwongwong: > APM is a low priority system, `1ms` is not in a good performance.

This is a very good reason to avoid `sleep` indeed.
Inspired by your comments, there is an elegant solution without `sleep` or `random`.

General idea:
Use increment instead of `random`, which avoid conflict and improve performance
Old:
image
New:
image

If ok, I will modify this PR.

wu-sheng: This could face concurrency status, `lastShiftValue++` could be the same in different threads.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ids/GlobalIdGenerator.java
```diff
@@ -74,11 +73,11 @@ private long timestamp() {
 
             if (currentTimeMillis < lastTimestamp) {
                 // Just for considering time-shift-back by Ops or OS. @hanahmily 's suggestion.
-                if (runRandomTimestamp != currentTimeMillis) {
-                    lastRandomValue = ThreadLocalRandom.current().nextInt();
-                    runRandomTimestamp = currentTimeMillis;
+                if (lastShiftTimestamp != currentTimeMillis) {
+                    lastShiftValue++;
```
wu-sheng, line 77: Again, this has a race condition. And no guarantee in concurrency

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ids/GlobalIdGenerator.java
nickwongwong: image
image

`ThreadLocal` means every thread hold different `IDContext`.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ids/GlobalIdGenerator.java
nickwongwong: image

wu-sheng: So? ThreadLocal is just holding a value? Then it could be as same as `N+1` in different threads.

nickwongwong: 对于每一次调用`GlobalIdGenerator.generate()`，都是获取当前线程的`ThreadLocal `对应的`IDContext`，也就说每个线程获取的`IDContext`都是自己独有的，不会存在并发问题。
`IDContext`不存在并发问题的话，显然，其中的`lastShiftValue`也不会。

For each call to `GlobalIdGenerator.generate()`, the `IDContext` corresponding to the `ThreadLocal` of the current thread is obtained. In other words, the `IDContext` obtained by each thread is unique to itself, and there will be no concurrency problems. .
If there is no concurrency problem with `IDContext`, obviously, the `lastShiftValue` in it will not.

nickwongwong: 我的理解没有问题吧？我看下面的的`nextThreadSeq++`也是没有问题
There is no problem with my understanding, right? I see the `nextThreadSeq++` below is also no problem
image

wu-sheng:
Generally good, please update the changes.md

nickwongwong:
> Generally good, please update the changes.md

done

wu-sheng (dismissed):
LGTM

wu-sheng (approved):

PR #6748, opened 2021-04-14
[Plugin] add Lua kong component ID

dmsolr:
- [X] Add a component id in [the component-libraries.yml](https://github.com/apache/skywalking/blob/master/oap-server/server-bootstrap/src/main/resources/component-libraries.yml)
- [X] Add a logo in [the UI repo](https://github.com/apache/skywalking-rocketbot-ui/tree/master/src/views/components/topology/assets)

### DIFF
--- a/.github/actions/skip/action.yml
+++ b/.github/actions/skip/action.yml
@@ -36,6 +36,7 @@ inputs:
       ".github/workflows/codeql.yaml"
       "dist-material/release-docs"
       "test/plugin/*"
+      "*/component-libraries.yml"
 
 runs:
   using: "composite"
--- a/oap-server/server-bootstrap/src/main/resources/component-libraries.yml
+++ b/oap-server/server-bootstrap/src/main/resources/component-libraries.yml
@@ -458,7 +458,9 @@ GoMicroServer:
 Nginx:
   id: 6000
   languages: Lua
-
+Kong:
+  id: 6001
+  languages: Lua
 
 # [7000, 8000) are reserved for Python components
 Python:


wu-sheng (dismissed):

kezhenxu94 (approved):
Adding component id is common case so I added the file into skip list

wu-sheng (approved):

wu-sheng:
> Adding component id is common case so I added the file into skip list

The side effect is, if a PR breaks the YAML format, we will face all tests failure on all next PRs :P

kezhenxu94:
> > Adding component id is common case so I added the file into skip list
> 
> The side effect is, if a PR breaks the YAML format, we will face all tests failure on all next PRs :P

Yes indeed, this is a tradeoff to reduce the CI resource load while increasing the reviewers' attention, since it's skipped when only `component-libraries.yml` is modified, it's acceptable to review its format as well, let's keep this until we really encounter the case you said.

wu-sheng:
> > > Adding component id is common case so I added the file into skip list
> > 
> > 
> > The side effect is, if a PR breaks the YAML format, we will face all tests failure on all next PRs :P
> 
> Yes indeed, this is a tradeoff to reduce the CI resource load while increasing the reviewers' attention, since it's skipped when only `component-libraries.yml` is modified, it's acceptable to review its format as well, let's keep this until we really encounter the case you said.

Make sense to me.

wu-sheng:
> > > Adding component id is common case so I added the file into skip list
> > 
> > 
> > The side effect is, if a PR breaks the YAML format, we will face all tests failure on all next PRs :P
> 
> Yes indeed, this is a tradeoff to reduce the CI resource load while increasing the reviewers' attention, since it's skipped when only `component-libraries.yml` is modified, it's acceptable to review its format as well, let's keep this until we really encounter the case you said.

Do we have a node type check in the topology e2e? If so, we could make sure we will find something wrong in the next PRs.

kezhenxu94:
- For Java libraries' component IDs, we have this test to verify https://github.com/apache/skywalking/blob/2b9de2007f4ae2004765bbbf0437ee9fe4bcb227/oap-server/server-bootstrap/src/test/java/org/apache/skywalking/oap/server/starter/ComponentLibrariesTest.java#L39 , but in this case, we usually need to modify more file than `component-libraries.yml`, so the CI checks won't be skipped.

- For non-Java components, we have no chance to check them at all actually, for example, malformed YAML like 

```yaml
Nginx:
  id: 6000
  languages: Lua
Kong:
  id: 6001
   languages: Lua # has one more space here
```

can be detected when parsing the file, but YAML like 

```yaml
Nginx:
  id: 6000
  languages: Lua
  Kong: # the entire section is indented as a child of Nginx section
    id: 6001
    languages: Lua # has one more space here
```

cannot be detected at all because it's also valid YAML and since we only read `Nginx.id` and `Nginx.languages`, and we don't know there is `Kong`, hence the other properties `Kong` is ignored.

Therefore, for non-Java component IDs, malformed YAML cannot be detected unless developers add the E2E tests and put component id assertions in the tests, which I believe we have that currently, for example, Go2Sky E2E, PHP E2E, Python E2E, etc.

kezhenxu94:
That said, I prefer non-Java component IDs should be added along with the E2E tests.

Issue #6749, opened 2021-04-14
About the value of service_instance_id in receiver_zipkin


klboke:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ 1] Bug
- [ ] Requirement
- [ ] Feature or performance improvement
___
### Bug
- Which version of SkyWalking, OS, and JRE?

1. OS : mac os 10.15.7
2. JRE : jdk1.8
3. skywalking version ：8.5.0
4. elasticsearch version ：7.10.2

- Which company or project?
TapTap
- What happened?
I tried to use receiver_zipkin in spring cloud and istio environment. But there are bugs. When ui query trace, this exception will appear: `Cannot return null for non-nullable type:'ID' within parent'Span' (/trace/spans[0]/serviceInstanceName`. For example:
![image](https://user-images.githubusercontent.com/18591662/114656226-42ca4a80-9d20-11eb-839b-ea898efd1f0f.png)

I have located the reason, which is caused by the lack of ServiceInstanceId in the span entity that was queried. I went to look at the processing logic of zipkin when receiving data, and found that there is no processing logic for ServiceInstanceId
___
### Requirement or improvement
- Please describe your requirements or improvement suggestions.
I learned that there is no definition of ServiceInstanceId in the zipkin data model, and then found that there is spring.InstanceId in the zipkin tags in the spring-cloud environment. There is node.id in the tags of istio's zipkin. These are very suitable to be used as ServiceInstanceId.

istio zipkin data
![image](https://user-images.githubusercontent.com/18591662/114656875-70fc5a00-9d21-11eb-81b6-91be420693ed.png)

spring-cloud  zipkin data
![image](https://user-images.githubusercontent.com/18591662/114657028-ba4ca980-9d21-11eb-8342-57610c4586f3.png)

wu-sheng:
We could adpot Spring if you like, and set an empty as name in default. Welcome to send a pull request to fix

klboke:
No problem, I plan to first be compatible with the naming of zipkin tags in the istio and spring environments. Then if there is not, set a default empty, you see if this is ok. If there is no problem, I will start to develop

wu-sheng:
Generally, you could set a list of tag keys and try to set an instance name. If nothing found, set a default value.

klboke:
Okay, design to get the naming corresponding to service_instance_id in tags from receiver_zipkin. I will follow up on this issue and submit pr

PR #6763, opened 2021-04-15
fix: Receiver_zipkin missing ServiceInstanceID bug #6749

klboke:
fix : https://github.com/apache/skywalking/issues/6749


oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/trace/SpanForward.java
```diff
@@ -29,15 +30,18 @@
...
 import org.apache.skywalking.oap.server.library.util.BooleanUtils;
+import org.apache.skywalking.oap.server.receiver.zipkin.ZipkinReceiverConfig;
 import org.apache.skywalking.oap.server.receiver.zipkin.handler.SpanEncode;
 import org.apache.skywalking.oap.server.storage.plugin.zipkin.ZipkinSpan;
 import zipkin2.Span;
 import zipkin2.codec.SpanBytesEncoder;
 
 @RequiredArgsConstructor
 public class SpanForward {
+    private static final String DEFAULT_SERVICE_INSTANCE_NAME = "Unknown_instance";
```
wu-sheng, line 41:
```suggestion
    private static final String DEFAULT_SERVICE_INSTANCE_NAME = " unknown_instance";
```

oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/ZipkinReceiverConfig.java
```diff
@@ -33,4 +36,5 @@
     private long jettyIdleTimeOut = 30000;
     private int jettyAcceptorPriorityDelta = 0;
     private int jettyAcceptQueueSize = 0;
+    private List<String> instanceNameRule = new ArrayList<>();
```
wu-sheng, line 39: Do OAP core support putting list type?

klboke: support

wu-sheng: Could you add a unit test to verify? I can't find the test about this for now. And existing codes use string, and then convert to list.

wu-sheng: @klboke I can't see the change, why do you resolve this?

oap-server/server-storage-plugin/storage-zipkin-elasticsearch7-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/zipkin/elasticsearch/ZipkinTraceQueryEs7DAO.java
```diff
@@ -211,9 +214,14 @@ public TraceBrief queryBasicTraces(long startSecondTB,
             });
             if (StringUtil.isNotEmpty(serviceId)) {
                 final IDManager.ServiceID.ServiceIDDefinition serviceIDDefinition = IDManager.ServiceID.analysisId(
-                    serviceId);
+                        serviceId);
                 swSpan.setServiceCode(serviceIDDefinition.getName());
             }
+            if (StringUtil.isNotEmpty(serviceInstanceId)) {
```
wu-sheng, line 220: This seems `serviceInstanceName` still could be NULL.

wu-sheng: If you are 100% saying this is not null, why need `if (StringUtil.isNotEmpty(serviceInstanceId))`

klboke:
```
type Span {
    traceId: ID!
    segmentId: ID!
    spanId: Int!
    parentSpanId: Int!
    refs: [Ref!]!
    serviceCode: String!
    serviceInstanceName: ID!
    startTime: Long!
    endTime: Long!
    endpointName: String
    # There are three span types: Local, Entry and Exit
    type: String!
    # Peer network id, e.g. host+port, ip+port
    peer: String
    component: String
    isError: Boolean
    # There are 5 layers: Unknown, Database, RPCFramework, Http, MQ and Cache
    layer: String
    tags: [KeyValue!]!
    logs: [LogEntity!]!
}
```
The processing logic here is just to be consistent with the serviceId

oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/ZipkinReceiverConfig.java
klboke: I am going to add a test case. Is there a similar test case for reference?

klboke: https://github.com/apache/skywalking/pull/6763/commits/0245b00cd3303b6222de9a2d6bd5206986e41c3b

oap-server/server-storage-plugin/storage-zipkin-elasticsearch7-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/zipkin/elasticsearch/ZipkinTraceQueryEs7DAO.java
wu-sheng: I am saying from the code perspective. What do you want to say to refer the GraphQL?

klboke: What I mean is that GraphQL also defines that the serviceCode cannot be empty, but it also makes null processing when processing the serviceId. I am not very clear about the reason, so I refer to the processing logic of the serviceId

oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/ZipkinReceiverConfig.java
wu-sheng: Test case to prove the property with list type working with the current framework.

klboke: I didn’t understand what you mean, is there any example you can refer to?

wu-sheng: Take a look at `ModuleManagerTest`.

oap-server/server-bootstrap/src/test/java/org/apache/skywalking/oap/server/starter/config/ApplicationConfigLoaderTestCase.java
```diff
@@ -47,4 +51,13 @@ public void testLoadConfig() {
         Properties properties = (Properties) providerConfig.get("properties");
         assertThat(properties.get("jdbcUrl"), is("jdbc:mysql://localhost:3306/swtest"));
     }
+
+    @Test
+    public void testLoadZipkinReceiverConfig() {
```
wu-sheng, line 56: You are putting codes in the config test, why name it related to Zipkin? You are testing the `list type` config injection.

wu-sheng (approved):

PR #6774, opened 2021-04-16
feat: add enhance pulsar MessageListener instance

### LINKED ISSUE #6730: pulsar plugin breaks trace on the consumer size

- [ ] Question or discussion
- [x] Bug
- [ ] Requirement

___
### Question
- What do you want to know?

___
### Bug
SkyWalking：8.4.0
OS：MacOS
JRE：11
- Which company or project?

- What happened?
If possible, provide a way to reproduce the error. e.g. demo application, component version.

Pulsar plugin is only enhance `org.apache.pulsar.client.impl.ConsumerImpl` class, however pulsar will use `org.apache.pulsar.client.impl.MultiTopicsConsumerImpl` class to consume when topic's partition size greater than 0. So the trace will be breaked on consumer side. 

In `org.apache.pulsar.client.impl.MultiTopicsConsumerImpl` class, pulsar will new some  `org.apache.pulsar.client.impl.ConsumerImpl` class instances to receive messages, but consumption process is in the internal executor while callback to listener in the external executor.

The entry method is `org.apache.pulsar.client.impl.PulsarClientImpl#doSingleTopicSubscribeAsync`.
___

wallezhang:
### Fix <#6730>

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6730 .


wu-sheng:
Should be reviewed by @codelipenghui 

And you should update https://github.com/apache/skywalking/tree/master/test/plugin/scenarios/pulsar-scenario to verify your case. How to write plugin test doc is in the plugin dev doc already.

wallezhang:
> Should be reviewed by @codelipenghui
> 
> And you should update https://github.com/apache/skywalking/tree/master/test/plugin/scenarios/pulsar-scenario to verify your case. How to write plugin test doc is in the plugin dev doc already.

OK, I will update pulsar-scenario according to the doc

wu-sheng:
If you add a new consumer, you need to update the expected data file.

wu-sheng:
And Changes.md should be updated to describe what you changed.

wu-sheng:
`pulsar` plugin test fails, please verify the tests locally first.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarConsumerListenerInterceptor.java
```diff
@@ -0,0 +1,86 @@
...
+ * <pre>
+ *  1. Get @{@link ContextCarrier} from message
+ *  2. Create a local span when call <code>received</code> method
+ *  3. Extract trace information from context carrier
+ *  4. Stop the local span when <code>received</code> method finished.
+ * </pre>
+ */
+public class PulsarConsumerListenerInterceptor implements InstanceMethodsAroundInterceptor {
+
+    public static final String CONSUMER_OPERATE_NAME = "Pulsar/Consumer/MessageListener";
```
codelipenghui, line 49: Shall we need to introduce a new operation name? The message listener just a way to consume messages, for the tracing context, we just need to know the consumer got the message.

And, the message listener can't work together with the receive() method, so I think we can use the existing operation name.

test/plugin/scenarios/pulsar-scenario/src/main/java/test/apache/skywalking/apm/testcase/pulsar/controller/CaseController.java
```diff
@@ -51,38 +51,37 @@ public String pulsarCase() throws PulsarClientException, InterruptedException {
...
         String topic = "test";
 
+        CountDownLatch latch = new CountDownLatch(1);
+
         PulsarClient pulsarClient = PulsarClient.builder().serviceUrl(PULSAR_DOMAIN + serviceUrl).build();
 
         Producer<byte[]> producer = pulsarClient.newProducer().topic(topic).create();
 
-        Consumer<byte[]> consumer = pulsarClient.newConsumer().topic(topic).subscriptionName("test").subscribe();
+        Consumer<byte[]> consumer = pulsarClient.newConsumer().topic(topic)
```
codelipenghui, line 60: I noticed you are replacing the old one(based on the receive method) with the message listener one. We should cover all of these two cases in the test. Could you please add a new consumer or add a new test for this case. I think you can add a new consumer with a different subscription name which will get the same messages from the topic.

codelipenghui (changes requested):

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarConsumerListenerInterceptor.java
wallezhang: Before entering the `MessageListener` callback method, the `receive()` method will be called to receive the message first, so I think introduce a new operation name can distinguish these two spans

test/plugin/scenarios/pulsar-scenario/src/main/java/test/apache/skywalking/apm/testcase/pulsar/controller/CaseController.java
wallezhang: Yes, because the `receive()` method will be called first before entering the `MessageListener` callback, so I directly replace the old one, and I think this can cover all cases in the test. I just add new expect data in `expectedData.yml` and reserve the `receive()` method segment. So based on this consideration, do I still need to create a new consumer to cover the new case?

wallezhang:
@codelipenghui I found that byte buddy can not instrument lambda, however most cases when we set a `MessageListener`, we use lambda expression, so I wonder if there is a better way to instrument the `MessageListener`.

wu-sheng:
> I found that byte buddy can not instrument lambda

This is not related to byte-buddy. It can do, but SkyWalking doesn't want you to do that. The recommended way to do is enhancing the method using the Lambda, and create another new lambda as a wrapper to do the instrument.

wallezhang:
> > I found that byte buddy can not instrument lambda
> 
> This is not related to byte-buddy. It can do, but SkyWalking doesn't want you to do that. The recommended way to do is enhancing the method using the Lambda, and create another new lambda as a wrapper to do the instrument.

OK, Just like what the class `RunnableWrapper` does?

wu-sheng:
> > > I found that byte buddy can not instrument lambda
> > 
> > 
> > This is not related to byte-buddy. It can do, but SkyWalking doesn't want you to do that. The recommended way to do is enhancing the method using the Lambda, and create another new lambda as a wrapper to do the instrument.
> 
> OK, Just like what the class `RunnableWrapper` does?

Yes, but use a new lambda. I remember we have this in some plugins, but can't tell which. Sorry, too many plugins.

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarConsumerListenerInterceptor.java
codelipenghui: It's not the same concepts here I think, the `Pulsar/Producer` means the client sends out the message, the `Pulsar/Producer/Callback` means the broker sends out the publish receipt to the client, if not callback from the broker, this means the message does not written succeed. 

But for the message listener, it's a way at the client side to get messages, both `Pulsar/Consumer/MessageListener` and `Pulsar/Consumer` are indicate the messages are arrivals the consumer, and we are not support the `Pulsar/Consumer/Acknowledge` for now to indicate the consumer processed the message and acked it.

The reason that I think we don't need a new operation here is we might have different get message methods in the client side such as batchReceive etc., I think the trace is telling users that the message arrives the consumer is enough here, If we want to add more details about how the users get the messages, I think we can add some labels in the trace context, but not a new operation name.

BTW, do we need to add a new span for handling this case? Looks like

```
--- Pulsar/Consumer
      --- Pulsar/Consumer/MessageListener
```
If there is a way to set the existing span to the message listener thread, seems more reasonable here.

test/plugin/scenarios/pulsar-scenario/src/main/java/test/apache/skywalking/apm/testcase/pulsar/controller/CaseController.java
codelipenghui: The consumer might disabled the message listener, it not a required feature for users to consume messages.

wallezhang: OK, I will add a new consumer with message listener to test new case

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/PulsarConsumerListenerInterceptor.java
wallezhang: I know what you mean, but in most cases, we will write the business code directly in the `MessageListener`, therefore it is not a span of message reception, but more like a process of message handling, such as
```
// span Pulsar/Consumer start
Message message = consumer.receive();
// List<Message> messages = consumer.batchReceive();
// span Pulsar/Consumer end

// business code, maybe start a new span named HandleMessage
// do some business  logic with message...
consumer.acknowledge(message);
// business code end, and the HandleMessage span is end
```

And with `MessageListener`, we will do the business logic like this
```
pulsarClient.newConsumer().topic("test").subscriptionName("test").messageListener((consumer,message) -> {
    // span Pulsar/Consumer/MessageListener start
    // do some business logic with message...
    consumer.acknowledge(message);
    // span Pulsar/Consumer/MessageListener end
})
```
Span `Pulsar/Consumer` is finished in client internal, and the `Pulsar/Consumer/MessageListener` is a `Local` span that represent the business logic processing. So I agree with adding a new span with `Pulsar/Consumer` span as parent.

wallezhang:
@codelipenghui Sorry for late update. I instrument `getMessageListener` method in `org.apache.pulsar.client.impl.conf.ConsumerConfigurationData` class, create a new lambda expression wrap the original message listener if it exists. And instrument `Message` interface so that I can pass the trace context snapshot to message listener thread. Plz take a review.

wallezhang:
@wu-sheng During the local test, I found that if I want to run unit test locally, I had to introduce some extra libraries in the root module pom file, e.g. `powermock-module-junit4-common`, `powermock-api-support`, `mockito-core`. Did I miss some configuration?

wu-sheng:
> @wu-sheng During the local test, I found that if I want to run unit test locally, I had to introduce some extra libraries in the root module pom file, e.g. `powermock-module-junit4-common`, `powermock-api-support`, `mockito-core`. Did I miss some configuration?

About the libraries you mentioned, some are not included. But we have included some libraries about `mockito-core` is there. I am not sure why you can't use it.

And you seems to update many submodules(core proto, query-protocol, test proto and UI), which should not be. Please revert these unexpected changes.

wallezhang:
> > @wu-sheng During the local test, I found that if I want to run unit test locally, I had to introduce some extra libraries in the root module pom file, e.g. `powermock-module-junit4-common`, `powermock-api-support`, `mockito-core`. Did I miss some configuration?
> 
> About the libraries you mentioned, some are not included. But we have included some libraries about `mockito-core` is there. I am not sure why you can't use it.
> 
> And you seems to update many submodules(core proto, query-protocol, test proto and UI), which should not be. Please revert these unexpected changes.

Sorry, I have reverted these changes already

wu-sheng:
> > > @wu-sheng During the local test, I found that if I want to run unit test locally, I had to introduce some extra libraries in the root module pom file, e.g. `powermock-module-junit4-common`, `powermock-api-support`, `mockito-core`. Did I miss some configuration?
> > 
> > 
> > About the libraries you mentioned, some are not included. But we have included some libraries about `mockito-core` is there. I am not sure why you can't use it.
> > And you seems to update many submodules(core proto, query-protocol, test proto and UI), which should not be. Please revert these unexpected changes.
> 
> Sorry, I have reverted these changes already

You reverted more than expected. Please recheck the file changes. Many things are being removed.

wallezhang:
> > > > @wu-sheng During the local test, I found that if I want to run unit test locally, I had to introduce some extra libraries in the root module pom file, e.g. `powermock-module-junit4-common`, `powermock-api-support`, `mockito-core`. Did I miss some configuration?
> > > 
> > > 
> > > About the libraries you mentioned, some are not included. But we have included some libraries about `mockito-core` is there. I am not sure why you can't use it.
> > > And you seems to update many submodules(core proto, query-protocol, test proto and UI), which should not be. Please revert these unexpected changes.
> > 
> > 
> > Sorry, I have reverted these changes already
> 
> You reverted more than expected. Please recheck the file changes. Many things are being removed.

I make some new commit, and recheck the change file list. All files are only related to this PR. 😄

wu-sheng:
I gave the green light to CI. @codelipenghui please review the codes.

codelipenghui (approved):

wu-sheng (approved):
Forward @codelipenghui 's approval, as he wrote this. And once tests passed, I am good with this PR.

wallezhang:
Great! Thank all for helping to review the code and put forward suggestions. 😄

wu-sheng:
@wallezhang Welcome to join the code contributor list. You will find your GitHub ID on https://skywalking.apache.org/team/

PR #6783, opened 2021-04-19
fix a bug that spring-mvc set an error endpoint name 

darcydai:
fix a bug that spring-mvc set an error endpoint name if the controller class annotation implements an interface.

----------------------------
recently, I find some error logs in the OAP server like this 
image

after decoding the endpoint id, I found the endpoint name is empty. after adding more logs in the trace segment receiver, I found the error from the spring-MVC plugin. 

 if the controller class annotation implements an interface, it can not get the RequestMapping. in the spring-cloud  tech stack, it is common sense, like RPC framework is feign
## example
``` java
@RequestMapping(value = "/for")
public interface A {
    @RequestMapping
    String bar();
}
```

```java
@RestController
public class B implements A{

    @Override
    public String bar() {
        return "oops";
    }
}
```  
  
  
result in oap server :
image

### Fix <spring-mvc set an error endpoint name if the controller class annotation implements an interface>


ascrutae:
You'd better add this case into the test scenarios about Spring framework.  @darcydai

ascrutae:
> You'd better add this case into the test scenarios about Spring framework. @darcydai

Here are those test scenarios about Spring framework:
1. [spring-4.3.x-scenario](https://github.com/apache/skywalking/tree/master/test/plugin/scenarios/spring-4.3.x-scenario)
2. [spring-3.1.x-scenario](https://github.com/apache/skywalking/tree/master/test/plugin/scenarios/spring-3.1.x-scenario)
3. [spring-4.1.x-scenario](https://github.com/apache/skywalking/tree/master/test/plugin/scenarios/spring-4.1.x-scenario)
4. [spring-4.3.x-scenario](https://github.com/apache/skywalking/tree/master/test/plugin/scenarios/spring-4.3.x-scenario)

darcydai:
> > You'd better add this case into the test scenarios about Spring framework. @darcydai
> 
> Here are those test scenarios about Spring framework:
> 1. [spring-4.3.x-scenario](https://github.com/apache/skywalking/tree/master/test/plugin/scenarios/spring-4.3.x-scenario)
> 2. [spring-3.1.x-scenario](https://github.com/apache/skywalking/tree/master/test/plugin/scenarios/spring-3.1.x-scenario)
> 3. [spring-4.1.x-scenario](https://github.com/apache/skywalking/tree/master/test/plugin/scenarios/spring-4.1.x-scenario)
> 4. [spring-4.3.x-scenario](https://github.com/apache/skywalking/tree/master/test/plugin/scenarios/spring-4.3.x-scenario)

I will do this later

darcydai:
> > You'd better add this case into the test scenarios about Spring framework. @darcydai
> 
> Here are those test scenarios about Spring framework:
> 
> 1. [spring-4.3.x-scenario](https://github.com/apache/skywalking/tree/master/test/plugin/scenarios/spring-4.3.x-scenario)
> 2. [spring-3.1.x-scenario](https://github.com/apache/skywalking/tree/master/test/plugin/scenarios/spring-3.1.x-scenario)
> 3. [spring-4.1.x-scenario](https://github.com/apache/skywalking/tree/master/test/plugin/scenarios/spring-4.1.x-scenario)
> 4. [spring-4.3.x-scenario](https://github.com/apache/skywalking/tree/master/test/plugin/scenarios/spring-4.3.x-scenario)
I found  a scenarios can only define one entry service,can i add more scenario in spring mvc for the special case ?  
@ascrutae  
image

wu-sheng:
> I found a scenarios can only define one entry service,can i add more scenario in spring mvc for the special case ?

You don't need to. Do a cycle call inside the entrance service. You could have as many `/PATH` as you like, unless timeout due to too slow

darcydai:
> > I found a scenarios can only define one entry service,can i add more scenario in spring mvc for the special case ?
> 
> You don't need to. Do a cycle call inside the entrance service. You could have as many `/PATH` as you like, unless timeout due to too slow

thanks, i got it

darcydai:
I found a new bug when I coding the test case for spring MVC, I has fix it in last commit.

darcydai:
DO NOT MERGE UNTIL I FINISH SCENARIOS TEST CASE

darcydai:
I found a new bug when i coding the scenarios for the spring MVC plugin. i have fixed in last commit
 `org.apache.skywalking.apm.plugin.spring.mvc.commons.interceptor.RequestMappingMethodInterceptor#getRequestURL`
1. the annotation RequestMapping with the value[value()] array length equals zore represent an empty string in spring MVC concept.
image

2. RequestMapping in spring mvc 3.x and 4.1  has no the path method [path()], if RequestMapping with  with the value[value()] array length equals zore in spring mvc 3 will case no such method error;
image

image


3. because of AliasFor annotation, no need to call path(), I have tested
image


image

darcydai:
do not  merge before i finish scenario test for spring mvc.

wu-sheng:
> do not merge before I finish the scenario test for spring MVC.

 I expect you could update the plugin test scenarios.

wu-sheng:
@darcydai Please fix your test case.

wu-sheng:
Seems still have style issues?

darcydai:
> Seems still have style issues?

seems failed scenario test is not spring MVC, how to trigger  the task run again

wu-sheng:
> > Seems still have style issues?
> 
> seems failed scenario test is not spring MVC, how to trigger the task run again

I will try to rerun, I think we have located the issue.

test/plugin/scenarios/spring-3.0.x-scenario/config/expectedData.yaml
```diff
@@ -14,52 +14,130 @@
...
+            spanLayer: Http
+            startTime: nq 0
+            endTime: nq 0
+            componentId: 14
+            isError: false
+            spanType: Entry
+            peer: ''
+            skipAnalysis: false
+            tags:
+              - {key: url, value: 'http://localhost:8080/spring-3.0.x-scenario//impl/requestmapping'}
```
ascrutae, line 52: The URL which has more '/' char looks strange, I think that `http://localhost:8080/spring-3.0.x-scenario/impl/requestmapping` prefer looks good.

test/plugin/scenarios/spring-3.0.x-scenario/config/expectedData.yaml
```diff
@@ -14,52 +14,130 @@
...
+            spanLayer: Http
+            startTime: nq 0
+            endTime: nq 0
+            componentId: 14
+            isError: false
+            spanType: Entry
+            peer: ''
+            skipAnalysis: false
+            tags:
+              - {key: url, value: 'http://localhost:8080/spring-3.0.x-scenario//impl/restmapping'}
```
ascrutae, line 74: Same as above

test/plugin/scenarios/spring-3.0.x-scenario/config/expectedData.yaml
```diff
@@ -14,52 +14,130 @@
...
+            spanLayer: Http
+            startTime: nq 0
+            endTime: nq 0
+            componentId: 13
+            isError: false
+            spanType: Exit
+            peer: localhost:8080
+            skipAnalysis: false
+            tags:
+              - {key: url, value: 'http://localhost:8080/spring-3.0.x-scenario//impl/requestmapping'}
```
ascrutae, line 108: Same as above

test/plugin/scenarios/spring-3.0.x-scenario/config/expectedData.yaml
```diff
@@ -14,52 +14,130 @@
...
+            spanLayer: Http
+            startTime: nq 0
+            endTime: nq 0
+            componentId: 13
+            isError: false
+            spanType: Exit
+            peer: localhost:8080
+            skipAnalysis: false
+            tags:
+              - {key: url, value: 'http://localhost:8080/spring-3.0.x-scenario//impl/restmapping'}
```
ascrutae, line 123: Same as above

test/plugin/scenarios/spring-3.0.x-scenario/config/expectedData.yaml
ascrutae: I found that this problem caused by `TestServiceBean#URL` has more  '/',   You can remove it to resolve the problem

ascrutae (approved):
LGTM

PR #6814, opened 2021-04-23
Pin the openjdk docker version

kezhenxu94:
Because `openjdk:8` is always pointed to the latest revision of `8` series, the latest one disabled TLS 1.0 (~16 hours ago) thus causes the mysql test to fail, this PR pins the Docker image to an older tag with specific revision so that it won't be overrode again


wu-sheng (approved):

wu-sheng:
Thanks for locating fast.

PR #6816, opened 2021-04-23
optimize alarm

lpcy:
### Fix alarm webhook server frequently receives empty messages
- [x] The alarm core does not judge whether the filteredmessages is empty before sending.


wu-sheng (approved):
Seems harmless.

PR #6832, opened 2021-04-25
Upgrade the etcd4j version to solve the memory leak problem

YczYanchengzhe:
- [x] Explain briefly why the bug exists and how to fix it. When we use etcd as the registry center of oap, netty reported a memory leak error. Check the version of etcd4j and found that this problem has been fixed in the latest version. It is recommended that skywalking can also upgrade the version of etcd4j


wu-sheng:
We are checking the dependency libraries, so, as you want to upgrade the dependency, you need to update
1. LICENSE, https://github.com/apache/skywalking/blob/master/dist-material/release-docs/LICENSE#L328
2. Dependency checklist, https://github.com/apache/skywalking/blob/master/tools/dependencies/known-oap-backend-dependencies.txt
3. Dependency checklist 2, https://github.com/apache/skywalking/blob/master/tools/dependencies/known-oap-backend-dependencies-es7.txt

About (2) and (3), please notice the tool checking all dependency libraries. If your change actually affects more than one library, (1) should be updated too.

wu-sheng:
@YczYanchengzhe And your changelogs are including unexpected things, please make that clear.

YczYanchengzhe:
Regarding the issue of dependency, I will modify it. What exactly did you mean by the malicious code? How can I fix it~

wu-sheng:
> What exactly did you mean by the malicious code? 

Where I said about `malicious codes`? Something is being misread?

wu-sheng:
Oh, I think I know what do you misread. I mean your commits https://github.com/apache/skywalking/pull/6832/commits. They include only and unrelated things. Not you had submitted real codes.
In the open-source, we usually expect the commits only to include things related to this specific PR.

wu-sheng:
@YczYanchengzhe Any update?

YczYanchengzhe:
I understand what you mean. I'll close this commit and pull one again

PR #6838, opened 2021-04-26
Add an optional agent plugin to support mybatis.

Ax1an:

image

image

### DIFF
--- a/apm-sniffer/optional-plugins/mybatis-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mybatis/MyBatisMethodMatch.java
+++ b/apm-sniffer/optional-plugins/mybatis-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mybatis/MyBatisMethodMatch.java
@@ -1,0 +1,37 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.plugin.mybatis;
+
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+public enum MyBatisMethodMatch {
+    INSTANCE;
+
+    public ElementMatcher<MethodDescription> getMyBatisMethodMatcher() {
+        return named("selectOne").or(named("selectList"))
+                                 .or(named("selectMap"))
+                                 .or(named("select"))
+                                 .or(named("insert"))
+                                 .or(named("update"))
+                                 .or(named("delete"));
+    }
+}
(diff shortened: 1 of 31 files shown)


apm-sniffer/optional-plugins/mybatis-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mybatis/SqlSessionOperationInterceptor.java
```diff
@@ -0,0 +1,71 @@
...
+            }
+            ContextManager.getRuntimeContext().put(MYBATIS_ENTRY_METHOD_NAME, operationName);
+        }
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                              Object ret) throws Throwable {
+        String operationName = objInst.getClass().getName() + "." + method.getName();
+        if (String.valueOf(ContextManager.getRuntimeContext().get(MYBATIS_ENTRY_METHOD_NAME)).equals(operationName)) {
```
zhaoyuguang, line 56: avoid String.valueOf(null)

Ax1an: image
It doesn't seem to be a big problem.

apm-sniffer/optional-plugins/mybatis-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mybatis/MyBatisMethodMatch.java
```diff
@@ -0,0 +1,37 @@
...
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+public enum MyBatisMethodMatch {
+    INSTANCE;
+
+    public ElementMatcher<MethodDescription> getMyBatisMethodMatcher() {
+        return named("selectOne").or(named("selectList"))
```
zhaoyuguang, line 30: I saw that you prevent repeated generation of localSpan code in the interceptor. It is because of such code.
image
 right? If it is, the code is ok, but there code `matcher method` can be clearer.

Ax1an: Yes, I want to avoid wasting resources by generating unimportant localSpan.
In some scenarios, selectList() is called directly.
image

zhaoyuguang: So use more API such as `takesArgument` is better

Ax1an: I think your suggestion is very good and I will optimize it.

Ax1an: Done.
image

zhaoyuguang: Mean, can use `named("selectOne").and(takesArguments(2))` instead of `named("selectOne")`?  what do you think about？

Ax1an: I don't recommend this, because sometimes it will enter a parameter's selectOne first.

zhaoyuguang: OK

zhaoyuguang (approved):
LGTM

wu-sheng:
@Ax1an Please recheck the CI, it fails somehow.

Ax1an:
@wu-sheng  I'm also a bit confused, when the latest main branch is not merged, it's good.

wu-sheng:
Let me try to rerun later. I will merge 6851 first.

Ax1an:
@wu-sheng OK, if there are still problems in the back, I'll take a closer look.

wu-sheng:
> @wu-sheng OK, if there are still problems in the back, I'll take a closer look.

running now.

wu-sheng (commented):
@kezhenxu94 Seems some shasum error of coverage?

image

PR #6843, opened 2021-04-26
Add shasum verification to Codecov Bash Uploader

kezhenxu94:
According to https://about.codecov.io/security-update/


wu-sheng (approved):

wu-sheng:
Thanks for fixing this.

PR #6851, opened 2021-04-27
Fix: Some incompatibles check codes didn't work in `PercentileFunction combine`.

wankai123:
### Fix: Some incompatibles check codes didn't work in `PercentileFunction combine`.
    
The condition codes below would always false:
image


wu-sheng (approved):
Good catch.

PR #6860, opened 2021-04-27
code clean up for function readWeLinkConfig

YunfengGao:
1.Fix spelling errors(rebootName -> robotName) in WeLinkAlarm.
2.There is a nesting IF in function `readWeLinkConfig` compare with null , then refactor a more elegant way. 
3.Some parameter are required and should not have default value, such as `clientId`,`clientSecret`,`accessTokenUrl`,`messageUrl`,`groupIds`, a fast-fail wil help user find problems.


oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RulesReader.java
```diff
@@ -257,28 +261,24 @@ private void readFeishuConfig(Rules rules) {
...
-                    welinkSettings.getWebhooks()
-                                  .add(new WeLinkSettings.WebHookUrl(clientId, clientSecret, accessTokenUrl, messageUrl,
-                                                                     rebootName, groupIds
-                                  ));
-                });
-            }
-            rules.setWelinks(welinkSettings);
+        Map<String, Object> welinkConfig = (Map<String, Object>) yamlData.getOrDefault(
+            "welinkHooks",
+            Collections.EMPTY_MAP
```
wu-sheng, line 268: Why remove IF but using the EMPTY MAP? 
No matter what is the reason, the title of PR doesn't match what you are doing. Please fix.

YunfengGao: To reduce one "if judgment" for code readability.It is equivalent before and after modification

YunfengGao: > Why remove IF but using the EMPTY MAP?

To reduce one "if judgment" for code readability. It is equivalent before and after modification.

> No matter what is the reason, the title of PR doesn't match what you are doing. Please fix.

OK.

wu-sheng (approved):

PR #6863, opened 2021-04-28
Fix possible NullPointerException in agent's ES plugin.

Ax1an:

Exception 1: 
image

Exception2: 
image


wu-sheng:
You need to explain why NPE happens, rather than add a checking. This is not the wrong way to fix.

Ax1an:
@wu-sheng OK, I'll sort out the specific reasons in my spare time.

Ax1an:
**Why NPE1 happens？**

image

First, we need to make it clear that skywalking sets dynamic fields when enhancing the `org.elasticsearch.client.transport.TransportClientNodesService.execute` method.

Secondly, we need to know that the exception occurred in the `org.apache.skywalking.apm.plugin.elasticsearch.v6.interceptor.AdapterActionFutureActionGetMethodsInterceptor` class of enhanced the `org.elasticsearch.action.support.AdapterActionFuture.actionGet` method.

Seeing this exception, I first suspect that the execution result of the `isTrace (objInst)` method in the `beforeMethod` method is false, which results in the `createLocalSpan` method not executing. And the `org.elasticsearch.action.support.AdapterActionFuture` class's instance method `actionGet` occurred exception  in the normal execution process.

image

By using arthas to observe `org.elasticsearch.action.support.AdapterActionFuture.actionGet()` method, I get the following exception.

```java
watch org.elasticsearch.action.support.AdapterActionFuture actionGet "{throwExp}" -e -x 2 -n 2

ts=2021-04-30 17:40:18; [cost=5.923075ms] result=@ArrayList[
    ConnectTransportException[[][10.111.233.68:9300] general node connection failure]; nested: IllegalStateException[Received message from unsupported version: [6.3.2] minimal compatible version is: [6.8.0]];
	at org.elasticsearch.transport.TcpTransport$ChannelsConnectedListener$1.onFailure(TcpTransport.java:957)
	at org.elasticsearch.transport.TransportHandshaker$HandshakeResponseHandler.handleResponse(TransportHandshaker.java:138)
	at org.elasticsearch.transport.TransportHandshaker$HandshakeResponseHandler.handleResponse(TransportHandshaker.java:115)
	at org.elasticsearch.transport.InboundHandler$1.doRun(InboundHandler.java:224)
	at org.elasticsearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:37)
	at org.elasticsearch.common.util.concurrent.EsExecutors$DirectExecutorService.execute(EsExecutors.java:193)
	at org.elasticsearch.transport.InboundHandler.handleResponse(InboundHandler.java:216)
	at org.elasticsearch.transport.InboundHandler.messageReceived(InboundHandler.java:141)
	at org.elasticsearch.transport.InboundHandler.inboundMessage(InboundHandler.java:105)
	at org.elasticsearch.transport.TcpTransport.inboundMessage(TcpTransport.java:660)
	at org.elasticsearch.transport.netty4.Netty4MessageChannelHandler.channelRead(Netty4MessageChannelHandler.java:62)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:359)
	... 32 more
,
]
```

After asking related colleagues, I found that they called version 6.3.2 ES cluster by using version 7.2.1 ES SDK.

Why the execution result of the `isTrace (objInst)` method in the `beforeMethod` method is false?

I find that the skywalking dynamic fields are set when the `org.elasticsearch.client.transport.TransportClientNodesService.execute()` method is executed.

image

However,  the `org.elasticsearch.client.transport.TransportClientNodesService.execute()` method is not executed every time before the `org.elasticsearch.action.support.AdapterActionFuture.actionGet()` method is executed.

NPE occurs in the following execution path:

image

`ScheduledNodeSampler` periodically sniffs cluster nodes and does not execute the `org.elasticsearch.client.transport.TransportClientNodesService.execute()` method.

NPE will not occur in the following execution path:

image

By looking at the `ActionRequestBuilder` method source code, you can know that the `org.elasticsearch.action.support.AdapterActionFuture.actionGet()` method is executed after the `org.elasticsearch.client.transport.TransportClientNodesService.execute()` method.

So this exception has nothing to do with skywalking. It is caused by developers operating ES with an incompatible version of SDK.

Ax1an:
**Why NPE2 happens？**

image

First, we need to make it clear that skywalking sets dynamic fields when enhancing the constructors in `org.elasticsearch.client.RestHighLevelClient` class.

Secondly, we need to know that the exception occurred in the `org.apache.skywalking.apm.plugin.elasticsearch.v6.interceptor.RestHighLevelClientSearchMethodsInterceptor` class of enhanced the `org.elasticsearch.client.RestHighLevelClient.search` method.

I checked the full agent log and found the following exception:

image

After asking related colleagues, I found that they used version 5.6.12 ES SDK.

However, the argument types of constructors in version 5.x are different from those in version 6.x and 7.x.

5.x constructors:

image

6.x and 7.x constructors:

image

So we need to add witness class to the `org.apache.skywalking.apm.plugin.elasticsearch.v6.define.RestHighLevelClientInstrumentation` class of elasticsearch-6.x-plugin to avoid enhancing the incompatible version.

And later, I will create a new PR to add `RestHighLevelClient` related enhancement for elaticsearch-5.x-plugin.

wu-sheng:
OK, so this witness class makes sense, we are not going to instrument es5 client, correct?

Ax1an:
@wu-sheng Yes.

wu-sheng (approved):
Good to me to restrict versions.

PR #6864, opened 2021-04-28
Fix: MAL function would miss samples name after creating new samples.

### LINKED ISSUE #6857: how can i understand cpu useage? it looks like incorrect

___
### Question

in infrastructure monitoring 
how can i understand cpu useage?
it always 0.0 
___
image
here is the config:
image

otel-collector-config:
image

wankai123:
### Fix MAL function would miss samples name after creating new samples.
- [X] Add a unit test to verify that the fix works.
- [X] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6857

When MAL do `aggregate` like `sum/max/min` or do `avg`/`histogram`, there used `InternalOps.newSample`.
But this function missed the samples name:

image

That is impacted the `rate/irate/increase` functions because when we do increase count we use sample `name` and `labels` 
as the `id` to separate the samples.

image

So, in some scenario, for example:
```yaml
  - name: cpu_total_percentage
    exp: (node_cpu_seconds_total * 100).tagNotEqual('mode' , 'idle').sum(['node_identifier_host_name']).rate('PT1M')
  - name: disk_written
    exp: node_disk_written_bytes_total.sum(['node_identifier_host_name']).rate('PT1M')
```
These 2 expressions will make the samples have the same labels if the sample name missed, the `id` above would the same.
The cache and calculate of the value would incorrect, this caused issue #6857.

### DIFF
--- a/oap-server/analyzer/meter-analyzer/src/test/java/org/apache/skywalking/oap/meter/analyzer/dsl/AggregationTest.java
+++ b/oap-server/analyzer/meter-analyzer/src/test/java/org/apache/skywalking/oap/meter/analyzer/dsl/AggregationTest.java
@@ -54,7 +54,7 @@ public class AggregationTest {
     @Parameterized.Parameters(name = "{index}: {0}")
     public static Collection<Object[]> data() {
         return Arrays.asList(new Object[][] {
-           /* {
+            {
                 "sum",
                 of("http_success_request", SampleFamilyBuilder.newBuilder(
                         Sample.builder().labels(of("idc", "t1")).value(50).name("http_success_request").build(),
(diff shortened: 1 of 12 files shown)


wu-sheng (dismissed):
@hanahmily Please recheck.

oap-server/analyzer/meter-analyzer/src/test/java/org/apache/skywalking/oap/meter/analyzer/dsl/AggregationTest.java
```diff
@@ -54,117 +54,117 @@
     @Parameterized.Parameters(name = "{index}: {0}")
     public static Collection<Object[]> data() {
         return Arrays.asList(new Object[][] {
-            {
+           /* {
```
kezhenxu94, line 57: Why comment out these cases?

wankai123: It's an omission，thanks @kezhenxu94 ，I have fixed it.

kezhenxu94 (approved):

PR #6888, opened 2021-05-02
Include event(s) to alarms.

### LINKED ISSUE #6725: Include events of the entity(s) in the alarm
I talked with @llinder yesterday, to introduce the current SkyWalking and what we are going to do about the EVENT concept.
Lance proposed a very interesting thing when SkyWalking sends an alarm, we could include the events of entity(one or two if it is a relationship level alarm) in the alarm message according to the period.

@kezhenxu94 You asked a very similar thing back in the opposite direction(https://github.com/apache/skywalking/pull/6702#discussion_r611034197).

I think we should do both ways. @kezhenxu94 @chenmudu What do you think? Of course, in a separated PR later.

chenmudu:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number.

Related to PR #6725

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/EventQueryService.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/EventQueryService.java
@@ -26,6 +26,9 @@ import org.apache.skywalking.oap.server.core.storage.query.IEventQueryDAO;
 import org.apache.skywalking.oap.server.library.module.ModuleManager;
 import org.apache.skywalking.oap.server.library.module.Service;
 
+import java.util.List;
+import java.util.Objects;
+
 import static java.util.Objects.isNull;
 import static org.apache.skywalking.apm.util.StringUtil.isBlank;
 
@@ -51,6 +54,14 @@ public class EventQueryService implements Service {
             throw new IllegalArgumentException("time field is required when uuid is absent.");
         }
         return getDao().queryEvents(condition);
+    }
+
+    public Events queryEvents(final List<EventQueryCondition> conditions) throws Exception {
+        EventQueryCondition condition = conditions.stream().filter(c -> isBlank(c.getUuid()) && isDurationInvalid(c.getTime())).findFirst().orElse(null);
+        if (Objects.nonNull(condition)) {
+            throw new IllegalArgumentException("time field is required when uuid is absent.");
+        }
+        return getDao().queryEvents(conditions);
     }
 
     boolean isDurationInvalid(final Duration duration) {
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/type/event/Source.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/type/event/Source.java
@@ -19,29 +19,16 @@
 package org.apache.skywalking.oap.server.core.query.type.event;
 
 import lombok.AllArgsConstructor;
+import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
-import org.apache.commons.lang3.StringUtils;
 
 @Data
+@Builder
 @NoArgsConstructor
 @AllArgsConstructor
 public class Source {
     private String service;
     private String serviceInstance;
     private String endpoint;
-
-    public String getSourcesStr() {
-        String sourceStr = "";
-        if (StringUtils.isNotBlank(service)) {
-            sourceStr = service;
-        }
-        if (StringUtils.isNotBlank(serviceInstance)) {
-            sourceStr = sourceStr + serviceInstance;
-        }
-        if (StringUtils.isNotBlank(endpoint)) {
-            sourceStr = sourceStr + endpoint;
-        }
-        return sourceStr;
-    }
 }
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/type/event/Source.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/type/event/Source.java
@@ -21,6 +21,7 @@ package org.apache.skywalking.oap.server.core.query.type.event;
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.NoArgsConstructor;
+import org.apache.commons.lang3.StringUtils;
 
 @Data
 @NoArgsConstructor
@@ -29,4 +30,18 @@ public class Source {
     private String service;
     private String serviceInstance;
     private String endpoint;
+
+    public String getSourcesStr() {
+        String sourceStr = "";
+        if (StringUtils.isNotBlank(service)) {
+            sourceStr = service;
+        }
+        if (StringUtils.isNotBlank(serviceInstance)) {
+            sourceStr = sourceStr + serviceInstance;
+        }
+        if (StringUtils.isNotBlank(endpoint)) {
+            sourceStr = sourceStr + endpoint;
+        }
+        return sourceStr;
+    }
 }
--- a/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
+++ b/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
@@ -22,12 +22,11 @@ import com.coxautodev.graphql.tools.GraphQLQueryResolver;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Objects;
-import java.util.concurrent.ForkJoinPool;
-import java.util.concurrent.ForkJoinTask;
-import java.util.concurrent.RecursiveTask;
+import java.util.Map;
 
-import lombok.SneakyThrows;
+import java.util.concurrent.ForkJoinPool;
+import java.util.stream.Collectors;
+
 import org.apache.skywalking.oap.server.core.CoreModule;
 import org.apache.skywalking.oap.server.core.analysis.IDManager;
 import org.apache.skywalking.oap.server.core.analysis.manual.searchtag.Tag;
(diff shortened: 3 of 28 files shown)


oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
```diff
@@ -62,11 +85,166 @@ public Alarms getAlarm(final Duration duration, final Scope scope, final String
...
         long startSecondTB = 0;
         long endSecondTB = 0;
+        EventQueryCondition condition = new EventQueryCondition();
         if (nonNull(duration)) {
             startSecondTB = duration.getStartTimeBucketInSec();
             endSecondTB = duration.getEndTimeBucketInSec();
+            condition.setTime(duration);
+        }
+        condition.setType(EventType.Error);
+        condition.setName("Alarm");
```
kezhenxu94, line 95: These conditions are specific to alarm events, which is not what we want.

Let's take an example, if there is an alarm saying that `cpm(serviceA_Instance1) == 0`, we should include all events of `serviceA_Instance1` during that period, they may be `Shutdown` event, `Crash` event, etc. so that users can make sense of the alarm from the events quickly

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
```diff
@@ -62,11 +85,166 @@ public Alarms getAlarm(final Duration duration, final Scope scope, final String
...
+        Map<String, List<Event>> mappingMap = events.getEvents().stream().collect(Collectors.groupingBy(Event::getServiceInSource));
+        alarms.getMsgs().forEach(a -> {
+            switch (a.getScopeId()) {
+                case DefaultScopeDefine.SERVICE :
+                    List<Event> serviceEvent = mappingMap.get(IDManager.ServiceID.analysisId(a.getId()).getName());
+                    if (CollectionUtils.isNotEmpty(serviceEvent)) {
+                        if (serviceEvent.size() == 1) {
+                            a.setEvents(serviceEvent);
+                        } else {
+                            Event event = serviceEvent.stream().filter(e -> StringUtils.equals(a.getMessage(), e.getMessage())).findFirst().orElse(null);
```
kezhenxu94, line 121: Ditto, filtering according to the message makes no sense as per our goal

wu-sheng:
I did a quick read, do you plan to change GraphQL? I saw changes relating to query.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
chenmudu: Yeah, which means that there are fewer filters to query Events, which means that when an alert occurs, the number of Events associated is no longer the corresponding event, but all related Events.

What I wanted to establish was whether Alarms and Event(s) had a unique connection, rather than all of the events associated with the application in a given time period.

Because this association method may not be accurate and accurate enough in special scenarios, it may produce incorrect association relationship.What do you think? @kezhenxu94  @wu-sheng .

Looking forward to your reply. Thanks a lot.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
chenmudu: > Ditto, filtering according to the message makes no sense as per our goal

Copy that.

chenmudu:
> I did a quick read, do you plan to change GraphQL? I saw changes relating to query.

Yeah, the back-end is currently associated with the corresponding Event(s) content into the Alarms.  Could you give me some better advices?

wu-sheng:
My point is, if you want to change query protocol, you need to a pull request to that repo first. Do you know how to do that?

chenmudu:
> My point is, if you want to change query protocol, you need to a pull request to that repo first. Do you know how to do that?

Yes, I know.

wu-sheng:
Then let's review the protocol first. Only requirement is, once protocol get merged, we have to ask you to finish codes ASAP.
Because otherwise, we can't adopt other protocol level enhancement.

chenmudu:
> Then let's review the protocol first. Only requirement is, once protocol get merged, we have to ask you to finish codes ASAP.
> Because otherwise, we can't adopt other protocol level enhancement.

No problem.

wu-sheng (commented):
Don't update other submodules. Those are not relating to your PR.

wu-sheng:
Also, please update all alarm notifiers. They should include events too.

chenmudu:
> Also, please update all alarm notifiers. They should include events too.

Shouldn't the current event information be forthe UI to query?
My biggest puzzle is where the data will come from if the Event is included in the Alarm Message.
I'm so sorry, perhaps I misunderstood your appeal. 
One final point of confusion: Do **all alarm notifiers** mean all Call Backs?

chenmudu:
> Don't update other submodules. Those are not relating to your PR.

This will be fixed later with the updates. Thanks for your reminder.

wu-sheng:
> Shouldn't the current event information be forthe UI to query?
My biggest puzzle is where the data will come from if the Event is included in the Alarm Message.

The event should be queried by the UI for sure. That is why I asked you to update the query protocol. The event could be from many places, such as k8s re-deployment, agent report, CI/CD system.

> One final point of confusion: Do all alarm notifiers mean all Call Backs?

Yes, once we send this alarm out, we should include events related, HTTP, gRPC, slack, anything we have.

chenmudu:
> > Shouldn't the current event information be forthe UI to query?
> > My biggest puzzle is where the data will come from if the Event is included in the Alarm Message.
> 
> The event should be queried by the UI for sure. That is why I asked you to update the query protocol. The event could be from many places, such as k8s re-deployment, agent report, CI/CD system.
> 
> > One final point of confusion: Do all alarm notifiers mean all Call Backs?
> 
> Yes, once we send this alarm out, we should include events related, HTTP, gRPC, slack, anything we have.

Yes, I see what you mean. But maybe I didn't make it clear. What I want to make sure is where to get these events and where to put these into the Alarm Message, so that it can be handled by the receiver(eg. http/slack/grpc etc.) of the Alarm.

Because I saw that when I constructed the Alarm Message, most of the attributes in it were taken directly or indirectly from the configuration file and from the Message Header.

wu-sheng:
> > > Shouldn't the current event information be forthe UI to query?
> > > My biggest puzzle is where the data will come from if the Event is included in the Alarm Message.
> > 
> > 
> > The event should be queried by the UI for sure. That is why I asked you to update the query protocol. The event could be from many places, such as k8s re-deployment, agent report, CI/CD system.
> > > One final point of confusion: Do all alarm notifiers mean all Call Backs?
> > 
> > 
> > Yes, once we send this alarm out, we should include events related, HTTP, gRPC, slack, anything we have.
> 
> Yes, I see what you mean. But maybe I didn't make it clear. What I want to make sure is where to get these events and where to put these into the Alarm Message, so that it can be handled by the receiver(eg. http/slack/grpc etc.) of the Alarm.
> 
> Because I saw that when I constructed the Alarm Message, most of the attributes in it were taken directly or indirectly from the configuration file and from the Message Header.

OK, if this is about the code structure, I have taken a look. I think you mean, you need to query existing events from database, right? @kezhenxu94 Is EVENT API going to provide an API to do that? As alarm core used to only access the core to add event, rather than reading them. I think due to current implementation, we need to read them first.

wu-sheng:
If that API takes time, we could merge this part first, but keep #6725 open to track further progress. As we need UI enhancement base on this PR. @kezhenxu94 WDYT?

kezhenxu94:
> If that API takes time, we could merge this part first, but keep #6725 open to track further progress. As we need UI enhancement base on this PR. @kezhenxu94 WDYT?

Agree to merge the UI query part to let the UI team go forward first. I’ll review this PR another round soon.

chenmudu:
If there are no other major problems, I will continue to complete the other part of #6725  in the next PR.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
```diff
@@ -62,11 +83,111 @@ public Alarms getAlarm(final Duration duration, final Scope scope, final String
...
         if (nonNull(duration)) {
             startSecondTB = duration.getStartTimeBucketInSec();
             endSecondTB = duration.getEndTimeBucketInSec();
+            condition.setTime(duration);
+        }
+        Alarms alarms = getQueryService().getAlarm(
+                scopeId, keyword, paging, startSecondTB, endSecondTB, tags);
+        Events events = null;
+        try {
+            events = getEventQueryService().queryEvents(condition);
```
wu-sheng, line 96: So, you only get to query a maximum of 100 events. SkyWalking could easily face 300+ services, 10k+instances, and countless endpoints (even we don't recommend alarm on the endpoint). This seems not the right call.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
```diff
@@ -62,11 +83,111 @@ public Alarms getAlarm(final Duration duration, final Scope scope, final String
...
+        } catch (Throwable e) {
+            LOGGER.error(e.getMessage(), e);
+            return alarms;
```
wu-sheng, line 99: Why catch exceptions here?

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
wu-sheng: I think we should the alarm list first, and query the event accordingly, @kezhenxu94 What do you think?
If you have concerns about performance, try concurrently query.

wu-sheng:
Please use the CLI tool to generate events(mock), and verify the alarm-event links in the e2e.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
chenmudu: So I'm going to go up to the `GraphQLQueryHandler` to catch it, so that's the right thing to do, right?

chenmudu:
My personal views are as follows:

The first thing we should determine is the relationship between these Events and Alarms (it seems that no one discussed with me above), and the need to find out the list of all events that were already in store when the current Alarm occurred.Or just include the corresponding `Type == Alarm Events` (which may be 1 or 2) generated under the current Alarm.

If it is determined in a certain time period with the **Scope (Service/Instance/Endpoint)** related to all events (such as the **Type ==  Start、 Shutdown、 Alarm、 and other, etc.**), it means that there could be multiple event,The amount is unimaginable.

First of all: The richness of so many events is guaranteed, but it is also filled with lists of events that are not too strongly related to the current Alarm.

Next: A list of events that are not too strongly correlated may not be as convenient as the user might expect.

Then: The first thing we should determine for performance concerns is how to query (through endpoints or not).

1. The first is to query all the Alarm Messages at once through the drop-down box as now, and then associate multiple Alarm Messages and multiple Events lists in memory.
2. The second option is to query the event list when an Alarm Message is queried and paginate it in the front end.
Only after you decide on these two query modes can you determine how to operate in the backend OAP Server and guarantee excellent performance.

In fact, the above problem is to determine how to establish an accurate and non-redundant relationship between numerous events and the current alarm, but there seems to be no excellent solution at present.

In my opinion, if we want to ensure Alarms and Events' associated data with value, we may have the simplest and most valuable way of association, which is to associate the current event list of `Type == Alarm`.Of course, there is nothing wrong with correlating all the Alarm lists that are matched under the current **Scope**.

To sum up, this is just my personal idea. There may be some mistakes in understanding. I am looking forward to @wu-sheng @kezhenxu94  reply. Thanks a lot.

wu-sheng:
@chenmudu Let me be clear about these
## The comments I made
You are querying all events, even binding to the scope still could impact the performance match.
1. If you take a look at UI, we could query all scopes' alarm messages. So this kind of filter is pointless.
2. We are only querying limited alarms at one time, so, we could definitely query related events after we have the alarm list.

## Your questions about the relationship and correlation
> The richness of so many events is guaranteed

For a specific entity, there is no chance to have too many events. Could you share why you feel this way? My arguing is you are querying all events of a given scope, rather than an entity. Take the above (1).

An alarm is according to the metrics, rather than events, so, they are NOT a direct causal relationship. Logically you could be, such as a redeployment brings the codes with a performance issue in the runtime, which reduces the process capability, then increase the latency. But only we know event and alarm binding within a certain time window and take a look at the codes, we know that for sure. This is not about the solution,
This example shows you, alarm + event is leading users in the right direction to find the issue, we are not doing it internally. So, basically, all events matters, we can't know what is happening. 

___
The conclusion is, once you have the alarm list with entities(of source scope), you could find their events easily, and as limited queries, there is not a query performance risk.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
kezhenxu94: > I think we should the alarm list first, and query the event accordingly, @kezhenxu94 What do you think?
> If you have concerns about performance, try concurrently query.

+1, please narrow the `condition` as much as possible, querying events by time range may cause critical performance problem

kezhenxu94: Also, I don't think we want to include the `Alarm` events in the alarm messages, it seems to be pointless to me.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
```diff
@@ -62,11 +83,111 @@ public Alarms getAlarm(final Duration duration, final Scope scope, final String
...
+                    if (CollectionUtils.isNotEmpty(sourceInstanceEvent)) {
+                        List<Event> filterEvents = sourceInstanceEvent.stream().filter(e -> StringUtils.equals(e.getSource().getServiceInstance(), sourceServiceInstanceName)).collect(Collectors.toList());
+                        a.setEvents(filterEvents);
+                    }
+                    if (CollectionUtils.isNotEmpty(destInstanceEvent)) {
+                        List<Event> filterEvents = destInstanceEvent.stream().filter(e -> StringUtils.equals(e.getSource().getServiceInstance(), destServiceInstanceName)).collect(Collectors.toList());
+                        a.getEvents().addAll(filterEvents);
+                    }
+                    break;
+                case DefaultScopeDefine.ENDPOINT :
```
kezhenxu94, line 158: As for the `endpoint` scope, I'm wondering what do you think about the idea that we simply fallback to the service instance level events, because there are rarely endpoint events in my opinion, but endpoint alarms are highly possible triggered by instance events. Associating endpoint alarm with endpoint events ONLY may be correct, but it's highly possible to miss the relationship between them once there missed the endpoint events.

chenmudu:
> 1. If you take a look at UI, we could query all scopes' alarm messages. So this kind of filter is pointless.

Cpoy, thanks. I understand that the incorrect place should be in the EventQueryCondition (sorry that I ignored the filter item on UI). I should enrich the EventQueryCondition according to Scope and reduce other meaningless event data from the query. 

> 2. We are only querying limited alarms at one time, so, we could definitely query related events after we have the alarm list.

Why don't we query all Alarm (there may be some other filters) and then let the user select the event list for a single Alarm in the Alarm List? Instead of including Events in each Alarm while querying the current Alarm list. I don't know if it's getting easier, but I'd like your advice.  Maybe I haven't seen the UI, so I'm not sure which is the final query mode.

Thanks a lot.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
chenmudu: Copy,  Thanks for your answer.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
chenmudu: Ok. Thanks.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
wu-sheng: @kezhenxu94 I think currently, I can't see a chance to see endpoint-level event, unless we have IDE or some code analysis tech to report there is logic changed in an endpoint scope.

> but endpoint alarms are highly possible triggered by instance events

I think you can't have endpoint-instance mapping, only could have service and service's all instances events, are you asking to get all of them?

wu-sheng:
> Why don't we query all Alarm (there may be some other filters) and then let the user select the event list for a single Alarm in the Alarm List? 
> Instead of including Events in each Alarm while querying the current Alarm list

You were proposing the query protocol, with the event included in the alarm. Why are you asking me about this? If you are thinking about 2 steps query, why we haven't discussed this at the protocol level.
Please be clear about what is your point.

chenmudu:
> You were proposing the query protocol, with the event included in the alarm. Why are you asking me about this? If you are thinking about 2 steps query, why we haven't discussed this at the protocol level.

Sorry, I just raised the other possibility. Thank you for your reply. I will pay attention next time.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
chenmudu: > You were proposing the query protocol, with the event included in the alarm. Why are you asking me about this? If you are thinking about 2 steps query, why we haven't discussed this at the protocol level.

Sorry, it was only after I received the request to narrow the scope of the query that I looked at the code and found that the current single-step query mode seemed unable to construct the query condition completely(The query criteria only have the **Scope Type**, without the  **Source values**.).
That's why I brought up the topic of a two-step query. I would like to confirm with you that we may need to roll back the query-protocol to the previous version. Sorry for my mistake.

if there are some problems in my way of asking questions, thanks for helping me to point out. I just want to do something to make SkyWalking to be better.Thanks for your reply.I'll pay attention next time.

wu-sheng:
It is fine to discuss merged thing, all things are not permanent.
But my question is more about,I can't see the issue of doing secondary query base on alarm list.
What are your concerns? Performance or something? Performance at here I think not hard to resolve. Not matter parrallel query or reorgnize the query conditions for event query. Such as Elasticsearch is good at multiple should conditions search.

wu-sheng:
I was asking about the protocol, because I thought you have made decision, rather than how we discuss questions. You don't have to go deeply about that.

chenmudu:
Okay. thanks for being so patient.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
chenmudu: I need to determine the number and types of **case branches** again here. I am not sure if I understand it correctly, it seems that the `case` judgment of `ENDPOINT_RELATION` needs to be **removed** here.

kezhenxu94: > I think you can't have endpoint-instance mapping, only could have service and service's all instances events.

OK I missed that.

> are you asking to get all of them?

If we can't get instance-endpoint mapping, listing all of the instances' events is one way, WDYT? @wu-sheng

wu-sheng: > we can't get instance-endpoint mapping

No, we can't. I prefer a service level event should be enough for the endpoint, after all, the endpoint belongs to a service. The unstable of a single instance can't impact the performance of an endpoint in most cases unless there is a load balance issue.

wu-sheng:
What is the status here? I notice no update in 3 days.

chenmudu:
E2e testing is underway and will be updated today.

wu-sheng:
e2e fails, please recheck.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
```diff
@@ -50,23 +65,199 @@ private AlarmQueryService getQueryService() {
...
+                scopeId, keyword, paging, startSecondTB, endSecondTB, tags);
+        return includeEvents2AlarmsByCondition(alarms, condition);
+    }
+
+    private Alarms includeEvents2AlarmsByCondition(Alarms alarms, EventQueryCondition condition) throws Throwable {
+        if (alarms.getTotal() < 1) {
+            return alarms;
+        }
+        SearchEventTask searchEventTask = new SearchEventTask(alarms.getMsgs(), condition);
+        ForkJoinTask<List<AlarmMessage>> queryEventTask = getForkJoinPool().submit(searchEventTask);
```
wu-sheng, line 110: A question, do we really need this? At least, do we really need this at OAP service level?
Multiple conditions match(like a `OR` in SQL) could be handled by ES `should` or `SQL's OR statement`or `SQL's Union statement`. I think most database has good optimization to do this kind of query in the DB engine.

wu-sheng: Even some storage can't, I don't know whether InfluxDB can, its DAO implementation could have the thread pool feature, provided by a util class.

chenmudu: `org.influxdb.querybuilder.WhereQueryImpl#orNested` and `org.influxdb.querybuilder.WhereNested#close` cooperate to use can be achieved like **OR** in SQL and ES the same effect as **should**. More info is [available here](https://github.com/influxdata/influxdb-java/blob/828298022b1a53c45ce84ccb2e03cb983469c38b/src/test/java/org/influxdb/querybuilder/api/BuiltQueryTest.java#L461).

wu-sheng: OK, it seems all storage could work in this case.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/EventQueryService.java
```diff
@@ -53,6 +56,14 @@ public Events queryEvents(final EventQueryCondition condition) throws Exception
         return getDao().queryEvents(condition);
     }
 
+    public Events queryEvents(final List<EventQueryCondition> conditions) throws Exception {
```
wu-sheng, line 59: Please be explicit about the condition. In the `List<EventQueryCondition>` case, should not have UUID. You should refactor to codes to make 2 `queryEvents` more clear.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
```diff
@@ -50,23 +64,167 @@ private AlarmQueryService getQueryService() {
...
     }
 
+    private EventQueryService getEventQueryService() {
+        if (eventQueryService == null) {
+            this.eventQueryService = moduleManager.find(CoreModule.NAME).provider().getService(EventQueryService.class);
+        }
+        return eventQueryService;
+    }
+
+    private ForkJoinPool getForkJoinPool() {
```
wu-sheng, line 74: Legal codes, please clean your PR.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
```diff
@@ -50,23 +64,167 @@ private AlarmQueryService getQueryService() {
...
             startSecondTB = duration.getStartTimeBucketInSec();
             endSecondTB = duration.getEndTimeBucketInSec();
+            condition.setTime(duration);
+        }
+        Alarms alarms = getQueryService().getAlarm(
+                scopeId, keyword, paging, startSecondTB, endSecondTB, tags);
+        return includeEvents2AlarmsByCondition(alarms, condition);
+    }
+
+    private Alarms includeEvents2AlarmsByCondition(Alarms alarms, EventQueryCondition condition) throws Exception {
```
wu-sheng, line 104:
```suggestion
    private Alarms findReleventEvents(Alarms alarms, EventQueryCondition condition) throws Exception {
```

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
```diff
@@ -50,23 +64,167 @@ private AlarmQueryService getQueryService() {
...
+                a.getEvents().addAll(mappingEvents.get(a.getId0SourcesStr()));
+            }
+            if (Boolean.TRUE.equals(a.getId0LinkId1Flag()) && CollectionUtils.isNotEmpty(mappingEvents.get(a.getId1SourcesStr()))) {
+                a.getEvents().addAll(mappingEvents.get(a.getId0SourcesStr()));
+            }
+        });
+        return alarms;
+    }
+
+    private List<Source> constructCurrentSource(AlarmMessage msg) {
```
wu-sheng, line 140:
```suggestion
    private List<Source> buildEventSources(AlarmMessage msg) {
```

wu-sheng:
@kezhenxu94 Do you think we should include these events in the notification messages? If so, should we include these in all channels? I think it isn't easy to format them in slack, wechat, etc.

FYI @llinder If you want to join the discussion

@chenmudu This is not a block to this PR. I just think about this and want to discuss it.

kezhenxu94:
> @kezhenxu94 Do you think we should include these events in the notification messages? If so, should we include these in all channels? I think it isn't easy to format them in slack, wechat, etc.
> 

I feel like it’s unnecessary to include events in notification channels, users tend to open the web ui for more details after receiving an alert message. I don’t think one can deal with the alert only in the chat channel.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/EventQueryService.java
chenmudu: Copy.

wu-sheng:
@kezhenxu94 Please review the e2e. I have finished one round review about code style. The way seems correct this time.

wu-sheng:
CI fails, please fix.

chenmudu:
> CI fails, please fix.

Could you help me restart CI? It seems there was a network problem that caused the mission to fail.

wu-sheng:
> > CI fails, please fix.
> 
> Could you help me restart CI? It seems there was a network problem that caused the mission to fail. Thanks a lot.

GitHub Action is down.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
```diff
@@ -50,23 +62,160 @@ private AlarmQueryService getQueryService() {
...
+        return eventQueryService;
+    }
+
     public AlarmTrend getAlarmTrend(final Duration duration) {
         return new AlarmTrend();
     }
 
     public Alarms getAlarm(final Duration duration, final Scope scope, final String keyword,
-                           final Pagination paging, final List<Tag> tags) throws IOException {
+                           final Pagination paging, final List<Tag> tags) throws Throwable {
```
wu-sheng, line 77: Why `Throwable`? This is rarely to see in the codes.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/EventQueryService.java
wu-sheng: I can't see any change about this. Could you explain more?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/type/event/Source.java
```diff
@@ -29,4 +30,18 @@
     private String service;
     private String serviceInstance;
     private String endpoint;
+
+    public String getSourcesStr() {
```
wu-sheng, line 34: I am super confused about this logic, every entity has a very clear ID rule, managed by `IDManager` in SkyWalking. Why do you create so many of this kind of method in here and `AlarmMessage`?
The codes seem to be changed randomly.

wu-sheng (changes requested):
I am not feeling well about the code style. about how you put the logic together. 
I really rarely judge a PR like this way, but the current status makes me very confused. Your codes should target to make people easier to understand rather harder, and you should not keep adding complex things for one simple feature.

@kezhenxu94 Since you know him in person or at least better than me, could you help him a little about how to put the codes in a better way?

SkyWalking usually doesn't require to restrict coding style, but this PR presents a warning status. I already have to download the codes, with IDE assistance, to understand the codes. 
This feature should not be so complex at the code level.

kezhenxu94:
> @kezhenxu94 Since you know him in person or at least better than me, could you help him a little about how to put the codes in a better way?

For sure.

llinder:
> > @kezhenxu94 Do you think we should include these events in the notification messages? If so, should we include these in all channels? I think it isn't easy to format them in slack, wechat, etc.
> 
> I feel like it’s unnecessary to include events in notification channels, users tend to open the web ui for more details after receiving an alert message. I don’t think one can deal with the alert only in the chat channel.

Makes sense to me.  Quickly opening an alert from a message to get more context is probably better UX than adding too much info in Slack.  At the same time, being able to correlate which events might be related to an alert seems really powerful for things like web hooks.  If the events aren't delivered with the web hook that might be OK but I would imagine it would be something that an automated process monitoring alerts might be interested in.  What comes to mind are things like, notify me if a recent change of my service is correlated to another teams alert etc.  Having that context in Slack or WeChat probably isn't necessary but access from custom web hooks or other integrations might make sense.

kezhenxu94:
> Having that context in Slack or WeChat probably isn't necessary but access from custom web hooks or other integrations might make sense.

I agree.

kezhenxu94:
Since @chenmudu has some emergent personal stuffs to deal with, I'm taking over this

wu-sheng (approved):

PR #6894, opened 2021-05-04
[WIP] group endpoint  in skywalking agent automatically

darcydai:
Hi all, I open this PR want to confirm that I am doing the right thing.  there is so much work that needs to do, like unit test case, scenario test case，optimize the spring MVC match logic 
In a distributed system, hundreds of services in the skywalking cluster, group endpoint name in the server by regex pattern configuration are very difficult for skywalking cluster maintainer, even impossible. using a group mechanism in the agent may be better. 
for example, the servlet container has no path variable concept, and spring mvc achieves this feature.   if a request with a path variable is rejected by a servlet filter (like authenticate filter)，the endpoint can not reset by spring mvc interceptor. 

it will cause endpoint options in skywalking UI very hard to use (endpoint traffic), more cost for metric aggregation(especial endpoint 2 endpoint relations ), and more alarm window for alarm module (even oom in OAP server). 
we can get the path mapping info when controller creating, and try to rename endpoint(use same logic like spring MVC path match) when sending to skywalking backend service. now I made the feature in an optional plugin.


darcydai:
PathMatcher.java is a copy from the spring MVC framework. and also under the Apache license. it will case license issue?

darcydai:
I create a new module in SDK plugin,  similar logic in the servlet container plugin(undertow jetty tomcat) move to this module like spring-mvc-commons module. to undertow, I think it should split into 2 parts, undertow core, and undertow servlet. undertow core can work without undertow servlet. and undertow servlet depends on undertow core.

darcydai:
Welcome to communicate face 2 face

wu-sheng:
Hi, I haven't read the codes, just glance at the changeset. You are making a change on the core level, and submit wide changes on the plugins. 
I know the parameters in the URI harming the metrics system, it was found and resolved. Have you known this? https://skywalking.apache.org/docs/main/latest/en/setup/backend/endpoint-grouping-rules/
If you have known, why do you think the client-side grouping is better than the OAP server side?

darcydai:
> Hi, I haven't read the codes, just glance at the changeset. You are making a change on the core level, and submit wide changes on the plugins.
> I know the parameters in the URI harming the metrics system, it was found and resolved. Have you known this? https://skywalking.apache.org/docs/main/latest/en/setup/backend/endpoint-grouping-rules/
> If you have known, why do you think the client-side grouping is better than the OAP server side?

yes， I have known this feature on the server-side. but config pattern rule by YAML is not convenient for use, special for large OAP cluster, we would not configuration every restful API.  in the skywalking agents, we can detect some useful information and group endpoint automatically. users no need to do configuration works.

darcydai:
that commit is not the final version, detect the spring MVC mapping info will intercept spring mvc mapping-register later, now I achieve this feature by scan the Controller objInstan annotation.

wu-sheng:
My argue is about you seems making plugins interact with each other, what is the interactive model?
In SkyWalking, if you want different components working together,we need that data flow first.
The operation name is overridable, and controlled by core.

wu-sheng:
If I go deeper into your case, some of them are not making sense.
> if a request with a path variable is rejected by a servlet filter (like authenticate filter)

If this is an authentication filter, the path and parameter don't matter in most cases. Even we unify those URIs, what metrics could you get?

> it will cause endpoint options in skywalking UI very hard to use (endpoint traffic), 

UI doesn't care about this. What do you mean hard to use? This low traffic endpoint should not be seen on the UI

> more cost for metric aggregation(especial endpoint 2 endpoint relations ), 

Since 8.0, the OAP removed the register mechanism, the possibility of URI doesn't matter anymore. It just costs a little memory of OAP cluster.

> and more alarm window for alarm module (even oom in OAP server).

Like we mentioned many times, endpoint level alarm is never recommended. We support endpoint level alarm is helping you to set rules for specific endpoints, rather than for all.

___
From all the above things, we really should focus on what are the real issues to trigger this feature. In SkyWalking, we are very focusing on the real case, rather than a feature people feeling in the first place.
Only with that, we could make sure the agent costing limited and reasonable resources, and help the system running in high traffic. In current practice, we usually support 3k~10k/s per JVM.

darcydai:
> My argue is about you seems making plugins interact with each other, what is the interactive model?
> In SkyWalking, if you want different components working together,we need that data flow first.
> The operation name is overridable, and controlled by core.

I have pushed a new commit that has no plugins interact with each other(the first commit is archiving in a simple way, not the final version). actually  the new naming plugin, not similar to others, interceptor of this is in order to detect the spring-mvc mapping info.  would you like to discuss this PR face to face  tomorrow ? I still think  group endpoint name in skywalking agent  is a right thing.

darcydai:
> My argue is about you seems making plugins interact with each other, what is the interactive model?
> In SkyWalking, if you want different components working together,we need that data flow first.
> The operation name is overridable, and controlled by core.

I have pushed a new commit that has no plugins interact with each other(the first commit is archiving in a simple way, not the final version). actually the new naming plugin, not similar to others, interceptor of this is in order to detect the spring-MVC mapping info.  would you like to discuss this PR face to face  tomorrow ? I still think  group endpoint name in skywalking agent  is a right thing.

darcydai:
> My argue is about you seems making plugins interact with each other, what is the interactive model?
> In SkyWalking, if you want different components working together,we need that data flow first.
> The operation name is overridable, and controlled by core.

I have pushed a new commit that has no plugins interact with each other(the first commit is archiving in a simple way, not the final version). actually the new naming plugin, not similar to others, interceptor of this is in order to detect the spring-MVC mapping info.  would you like to discuss this PR face to face  tomorrow ? I still think  group endpoint name in skywalking agent  is a right thing.

darcydai:
> My argue is about you seems making plugins interact with each other, what is the interactive model?
> In SkyWalking, if you want different components working together,we need that data flow first.
> The operation name is overridable, and controlled by core.

I have pushed a new commit that has no plugins interact with each other(the first commit is archiving in a simple way, not the final version). actually the new naming plugin, not similar to others, interceptor of this is in order to detect the spring-MVC mapping info.  would you like to discuss this PR face to face  tomorrow ? I still think  group endpoint name in skywalking agent  is a right thing.

wu-sheng:
> would you like to discuss this PR face to face tomorrow ? 

What do you mean face to face?

> interceptor of this is in order to detect the spring-mvc mapping info

Could you provide a graph or something to show what is going on? It is much better for me to spend much time reading codes.

>  I still think group endpoint name in skywalking agent is a right thing.

Let's not jump to the conclusion directly. We haven't been there yet.

darcydai:
> My argue is about you seems making plugins interact with each other, what is the interactive model?
> In SkyWalking, if you want different components working together,we need that data flow first.
> The operation name is overridable, and controlled by core.

I have pushed a new commit that has no plugins interact with each other(the first commit is archiving in a simple way, not the final version). actually the new naming plugin, not similar to others, interceptor of this is in order to detect the spring-MVC mapping info.  would you like to discuss this PR face to face  tomorrow ? I still think  group endpoint name in skywalking agent  is a right thing.

wu-sheng:
Any update?

darcydai:
> Any update?

I am a little busy last week and will continue work in the process by now

wu-sheng:
No update.

PR #6899, opened 2021-05-06
Add `okhttp-4.x` plugin

liqiangz:
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>.
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md).

PR #6903, opened 2021-05-06
Fix `container.log: No such file or directory` in scenarios test

liqiangz:
### Fix `container.log: No such file or directory` in scenarios test
image


dmsolr:
Hi @liqiangz 
As far as I know, `./logs` is created by `run.sh` script in https://github.com/apache/skywalking/blob/c2399c091bab5601e7afbe15689581d2d77f243a/test/plugin/run.sh#L196

Could you give more information about how to reproduce this issue?

dmsolr:
I notice the test case, you showed, works well on our [plugin-test scenario](https://github.com/apache/skywalking/pull/6899/checks?check_run_id=2516811775).

liqiangz:
@dmsolr  https://github.com/apache/skywalking/pull/6863/checks?check_run_id=2455313835
image

dmsolr:
> @dmsolr Thanks for your reply.
> https://github.com/apache/skywalking/pull/6863/checks?check_run_id=2455313835
> image

I checked it. It is successful. It should be accidental, so I think we don't fix this.
How do you think, @wu-sheng ?
image

wu-sheng:
@liqiangz If this usually works, I agree there is nothing to fix.

liqiangz:
@dmsolr @wu-sheng 
Thanks for your reply. I haven’t checked the others carefully, but this error is a must in the okhttp test.
Yean, this does not affect the results of ci. But it will cause the running log to be unable to be seen when running. It will bring some trouble to code debugging.

wu-sheng:
@dmsolr Please recheck the running logs. Whether do we face this?

dmsolr:
In fact, It is harmless and acceptable. But I prefer to fix on `run.sh` if so.

liqiangz:
> In fact, It is harmless and acceptable. But I prefer to fix on `run.sh` if so.

@dmsolr Fixing in run.sh is a better way, I will update it later.
At least fixing it can reduce the confusion of developers running tests locally.

liqiangz:
@dmsolr I found that the folder has indeed been created in run.sh. But in the scenario.sh,  it still reports that the folder cannot be found. But it is ok  when creating  a folder in compose-start-script.template. I haven't found the reason yet.  If you have time, could you please help me see why this happens?

wu-sheng:
Milestone removed, it seems we don't have the cause for now.

wu-sheng:
Closing as no root cause.

PR #6909, opened 2021-05-07
FIX: NPE when thrift field is nested

xuanyu66:

When thrift message has nested field, ServerInProtocolWrapper read Inner fieldStop and make context = null, then ServerInProtocolWrapper read outer fieldStop, which cause NPE.
So I move this logic to readMessageEnd() so that make sure it's invoked only once
```java
ServerInProtocolWrapper#readFieldBegin()
    @Override
    public TField readFieldBegin() throws TException {
       ................
        if (field.type == TType.STOP) {
            Boolean haveCreatedSpan =
                    (Boolean) ContextManager.getRuntimeContext().get(HAVE_CREATED_SPAN);
            if (haveCreatedSpan != null && !haveCreatedSpan) {
                try {
                    AbstractSpan span = ContextManager.createEntrySpan(
                            context.getOperatorName(), createContextCarrier(null));
                    span.start(context.startTime);
                    span.tag(TAG_ARGS, context.getArguments());
                    span.setComponent(ComponentsDefine.THRIFT_SERVER);
                    SpanLayer.asRPCFramework(span);
                } catch (Throwable throwable) {
                    LOGGER.error("Failed to create EntrySpan.", throwable);
                } finally {
                    context = null;
                }

        return field;
    }
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,7 @@ Release Notes.
 * Add an optional agent plugin to support mybatis.
 * Add `spring-cloud-gateway-3.x` optional plugin.
 * Add `okhttp-4.x` plugin.
+* Fix NPE when thrift field is nested in plugin `thrift`
 
 #### OAP-Backend
 * BugFix: filter invalid Envoy access logs whose socket address is empty.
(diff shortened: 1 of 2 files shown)


dmsolr (approved):
LGTM

PR #6910, opened 2021-05-07
FIX: Failed to create ExitSpan when thrift transport is not enhanced

xuanyu66:

When transport is't enhanced by plugin, TServiceClient will fail to set SkyWalkingDynamicField, which make create exit span failed.
I validate transport, if it is not enhanced, set the SkyWalkingDynamicField with "UNKNOWN"

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,7 @@ Release Notes.
 * Add an optional agent plugin to support mybatis.
 * Add `spring-cloud-gateway-3.x` optional plugin.
 * Add `okhttp-4.x` plugin.
+* Fix bug that failed to create ExitSpan when thrift transport is not enhanced
 
 #### OAP-Backend
 * BugFix: filter invalid Envoy access logs whose socket address is empty.
--- a/apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/client/TServiceClientInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/thrift-plugin/src/main/java/org/apache/skywalking/apm/plugin/thrift/client/TServiceClientInterceptor.java
@@ -56,7 +56,13 @@ public class TServiceClientInterceptor implements InstanceConstructorInterceptor
                     "oprot_",
                     new ClientOutProtocolWrapper(protocol)
             );
-            Object dynamicField = ((EnhancedInstance) protocol.getTransport()).getSkyWalkingDynamicField();
+            
+            TTransport transport = protocol.getTransport();
+            Object dynamicField = null;
+            if (transport instanceof EnhancedInstance) {
+                dynamicField = ((EnhancedInstance) transport).getSkyWalkingDynamicField();
+            }
+
             objInst.setSkyWalkingDynamicField(Objects.isNull(dynamicField) ? "UNKNOWN" : dynamicField);
         }
     }


wu-sheng:
I think the question is, why not enhanced? Which should not happen.

xuanyu66:
> I think the question is, why not enhanced? Which should not happen.

If Someone import some package which implement new transport, which is situation we met.
I think it can not avoid，we need some Safeguard

wu-sheng:
Could you explain about new transport? Why wouldn't we enhance that?

xuanyu66:
> Could you explain about new transport? Why wouldn't we enhance that?

It's jaeger that implement ThriftUdpTransport for UdpSender

```
public class UdpSender extends ThriftSender {
  public static final String DEFAULT_AGENT_UDP_HOST = "localhost";
  public static final int DEFAULT_AGENT_UDP_COMPACT_PORT = 6831;

  @ToString.Exclude private Agent.Client agentClient;
  @ToString.Exclude private ThriftUdpTransport udpTransport;

  /**
   * This constructor expects Jaeger running running on {@value #DEFAULT_AGENT_UDP_HOST}
   * and port {@value #DEFAULT_AGENT_UDP_COMPACT_PORT}
   */
  public UdpSender() {
    this(DEFAULT_AGENT_UDP_HOST, DEFAULT_AGENT_UDP_COMPACT_PORT, 0);
  }

  /**
   * @param host host e.g. {@value DEFAULT_AGENT_UDP_HOST}
   * @param port port e.g. {@value DEFAULT_AGENT_UDP_COMPACT_PORT}
   * @param maxPacketSize if 0 it will use {@value ThriftUdpTransport#MAX_PACKET_SIZE}
   */
  public UdpSender(String host, int port, int maxPacketSize) {
    super(ProtocolType.Compact, maxPacketSize);

    if (host == null || host.length() == 0) {
      host = DEFAULT_AGENT_UDP_HOST;
    }

    if (port == 0) {
      port = DEFAULT_AGENT_UDP_COMPACT_PORT;
    }

    udpTransport = ThriftUdpTransport.newThriftUdpClient(host, port);
    agentClient = new Agent.Client(protocolFactory.getProtocol(udpTransport));
  }
```

dmsolr:
Do you mean customizing transport?

xuanyu66:
> Do you mean customizing transport?

yes, some package maybe  customize transport , we need some safeguard

wu-sheng:
Please resolve conflict.

xuanyu66:
> Please resolve conflict.

done

dmsolr:
> > Do you mean customizing transport?
> 
> yes, some package maybe customize transport , we need some safeguard

This is a known issue. In this scenario, there are two things maybe we need to consider.
1. whether we can get the "peer" from transport
2. can we cover all customized cases

xuanyu66:
> > > Do you mean customizing transport?
> > 
> > 
> > yes, some package maybe customize transport , we need some safeguard
> 
> This is a known issue. In this scenario, there are two things maybe we need to consider.
> 
> 1. whether we can get the "peer" from transport
> 2. can we cover all customized cases

I think we just set UNKNOWN, if someone want to enhance some customized transport, they can contribute themselves, we should only support official transport

wu-sheng:
> > > > Do you mean customizing transport?
> > > 
> > > 
> > > yes, some package maybe customize transport , we need some safeguard
> > 
> > 
> > This is a known issue. In this scenario, there are two things maybe we need to consider.
> > 
> > 1. whether we can get the "peer" from transport
> > 2. can we cover all customized cases
> 
> I think we just set UNKNOWN, if someone want to enhance some customized transport, they can contribute themselves, we should only support official transport.
> What's your opinion?

My question would be, if this is not an official transport implementation, why should SkyWalking take care of? User could remove the thrift plugin directly.

dmsolr:
I agree with @wu-sheng said. Thrift plugin is complex enough and easy to break other cases. I am worried about this.

xuanyu66:
> > > > > Do you mean customizing transport?
> > > > 
> > > > 
> > > > yes, some package maybe customize transport , we need some safeguard
> > > 
> > > 
> > > This is a known issue. In this scenario, there are two things maybe we need to consider.
> > > 
> > > 1. whether we can get the "peer" from transport
> > > 2. can we cover all customized cases
> > 
> > 
> > I think we just set UNKNOWN, if someone want to enhance some customized transport, they can contribute themselves, we should only support official transport.
> > What's your opinion?
> 
> My question would be, if this is not an official transport implementation, why should SkyWalking take care of? User could remove the thrift plugin directly.

we want to trace thrift rpc call, and it work properly. But other dependency with customized transport cause IllegalStateException.

wu-sheng:
> > > > > > Do you mean customizing transport?
> > > > > 
> > > > > 
> > > > > yes, some package maybe customize transport , we need some safeguard
> > > > 
> > > > 
> > > > This is a known issue. In this scenario, there are two things maybe we need to consider.
> > > > 
> > > > 1. whether we can get the "peer" from transport
> > > > 2. can we cover all customized cases
> > > 
> > > 
> > > I think we just set UNKNOWN, if someone want to enhance some customized transport, they can contribute themselves, we should only support official transport.
> > > What's your opinion?
> > 
> > 
> > My question would be, if this is not an official transport implementation, why should SkyWalking take care of? User could remove the thrift plugin directly.
> 
> we want to trace thrift rpc call, and it work properly. But other dependency with customized transport cause IllegalStateException.

It seems strange. Are you saying your thrift is using 2 kinds of transport implementations in one service? Why?

xuanyu66:
> > > > > > > Do you mean customizing transport?
> > > > > > 
> > > > > > 
> > > > > > yes, some package maybe customize transport , we need some safeguard
> > > > > 
> > > > > 
> > > > > This is a known issue. In this scenario, there are two things maybe we need to consider.
> > > > > 
> > > > > 1. whether we can get the "peer" from transport
> > > > > 2. can we cover all customized cases
> > > > 
> > > > 
> > > > I think we just set UNKNOWN, if someone want to enhance some customized transport, they can contribute themselves, we should only support official transport.
> > > > What's your opinion?
> > > 
> > > 
> > > My question would be, if this is not an official transport implementation, why should SkyWalking take care of? User could remove the thrift plugin directly.
> > 
> > 
> > we want to trace thrift rpc call, and it work properly. But other dependency with customized transport cause IllegalStateException.
> 
> It seems strange. Are you saying your thrift is using 2 kinds of transport implementations in one service? Why?

We use official transport to call rpc, jaeger use its customized transport to report trace span

wu-sheng:
> > > > > > > > Do you mean customizing transport?
> > > > > > > 
> > > > > > > 
> > > > > > > yes, some package maybe customize transport , we need some safeguard
> > > > > > 
> > > > > > 
> > > > > > This is a known issue. In this scenario, there are two things maybe we need to consider.
> > > > > > 
> > > > > > 1. whether we can get the "peer" from transport
> > > > > > 2. can we cover all customized cases
> > > > > 
> > > > > 
> > > > > I think we just set UNKNOWN, if someone want to enhance some customized transport, they can contribute themselves, we should only support official transport.
> > > > > What's your opinion?
> > > > 
> > > > 
> > > > My question would be, if this is not an official transport implementation, why should SkyWalking take care of? User could remove the thrift plugin directly.
> > > 
> > > 
> > > we want to trace thrift rpc call, and it work properly. But other dependency with customized transport cause IllegalStateException.
> > 
> > 
> > It seems strange. Are you saying your thrift is using 2 kinds of transport implementations in one service? Why?
> 
> We use official transport to call rpc, jaeger use its customized transport to report trace span.
> there are two different thrift client

That is strange, why SkyWalking needs to work with Jaeger? What is the use case?

xuanyu66:
> > > > > > > > > Do you mean customizing transport?
> > > > > > > > 
> > > > > > > > 
> > > > > > > > yes, some package maybe customize transport , we need some safeguard
> > > > > > > 
> > > > > > > 
> > > > > > > This is a known issue. In this scenario, there are two things maybe we need to consider.
> > > > > > > 
> > > > > > > 1. whether we can get the "peer" from transport
> > > > > > > 2. can we cover all customized cases
> > > > > > 
> > > > > > 
> > > > > > I think we just set UNKNOWN, if someone want to enhance some customized transport, they can contribute themselves, we should only support official transport.
> > > > > > What's your opinion?
> > > > > 
> > > > > 
> > > > > My question would be, if this is not an official transport implementation, why should SkyWalking take care of? User could remove the thrift plugin directly.
> > > > 
> > > > 
> > > > we want to trace thrift rpc call, and it work properly. But other dependency with customized transport cause IllegalStateException.
> > > 
> > > 
> > > It seems strange. Are you saying your thrift is using 2 kinds of transport implementations in one service? Why?
> > 
> > 
> > We use official transport to call rpc, jaeger use its customized transport to report trace span.
> > there are two different thrift client
> 
> That is strange, why SkyWalking needs to work with Jaeger? What is the use case?

Which I want to say is that other package may implement their own customized transport to send message at the same time we use official transport to call rpc between services. Should we take this situation into account.

wu-sheng:
> > > > > > > > > > Do you mean customizing transport?
> > > > > > > > > 
> > > > > > > > > 
> > > > > > > > > yes, some package maybe customize transport , we need some safeguard
> > > > > > > > 
> > > > > > > > 
> > > > > > > > This is a known issue. In this scenario, there are two things maybe we need to consider.
> > > > > > > > 
> > > > > > > > 1. whether we can get the "peer" from transport
> > > > > > > > 2. can we cover all customized cases
> > > > > > > 
> > > > > > > 
> > > > > > > I think we just set UNKNOWN, if someone want to enhance some customized transport, they can contribute themselves, we should only support official transport.
> > > > > > > What's your opinion?
> > > > > > 
> > > > > > 
> > > > > > My question would be, if this is not an official transport implementation, why should SkyWalking take care of? User could remove the thrift plugin directly.
> > > > > 
> > > > > 
> > > > > we want to trace thrift rpc call, and it work properly. But other dependency with customized transport cause IllegalStateException.
> > > > 
> > > > 
> > > > It seems strange. Are you saying your thrift is using 2 kinds of transport implementations in one service? Why?
> > > 
> > > 
> > > We use official transport to call rpc, jaeger use its customized transport to report trace span.
> > > there are two different thrift client
> > 
> > 
> > That is strange, why SkyWalking needs to work with Jaeger? What is the use case?
> 
> Which I want to say is that other package may implement their own customized transport to send message at the same time we use official transport to call rpc between services. Should we take this situation into account.

I have to repeat the point I said before. `MAY` is not a thing the community should take care of. Some real and reasonable use cases are. It is really rare to see a service using 2 RPC protocols but sharing the same framework. And also, you mentioned, SkyWalking and Jaeger are being deployed at the same time. We have no intention to compete, but also I have no vision why you need both at the same time, especially you are talking in Java stack.
Maybe you have something that can't be shared in public, but at the same time, I am afraid, open-source community and project are responding to open information only.

xuanyu66:
> > > > > > > > > > > Do you mean customizing transport?
> > > > > > > > > > 
> > > > > > > > > > 
> > > > > > > > > > yes, some package maybe customize transport , we need some safeguard
> > > > > > > > > 
> > > > > > > > > 
> > > > > > > > > This is a known issue. In this scenario, there are two things maybe we need to consider.
> > > > > > > > > 
> > > > > > > > > 1. whether we can get the "peer" from transport
> > > > > > > > > 2. can we cover all customized cases
> > > > > > > > 
> > > > > > > > 
> > > > > > > > I think we just set UNKNOWN, if someone want to enhance some customized transport, they can contribute themselves, we should only support official transport.
> > > > > > > > What's your opinion?
> > > > > > > 
> > > > > > > 
> > > > > > > My question would be, if this is not an official transport implementation, why should SkyWalking take care of? User could remove the thrift plugin directly.
> > > > > > 
> > > > > > 
> > > > > > we want to trace thrift rpc call, and it work properly. But other dependency with customized transport cause IllegalStateException.
> > > > > 
> > > > > 
> > > > > It seems strange. Are you saying your thrift is using 2 kinds of transport implementations in one service? Why?
> > > > 
> > > > 
> > > > We use official transport to call rpc, jaeger use its customized transport to report trace span.
> > > > there are two different thrift client
> > > 
> > > 
> > > That is strange, why SkyWalking needs to work with Jaeger? What is the use case?
> > 
> > 
> > Which I want to say is that other package may implement their own customized transport to send message at the same time we use official transport to call rpc between services. Should we take this situation into account.
> 
> I have to repeat the point I said before. `MAY` is not a thing the community should take care of. Some real and reasonable use cases are. It is really rare to see a service using 2 RPC protocols but sharing the same framework. And also, you mentioned, SkyWalking and Jaeger are being deployed at the same time. We have no intention to compete, but also I have no vision why you need both at the same time, especially you are talking in Java stack.
> Maybe you have something that can't be shared in public, but at the same time, I am afraid, open-source community and project are responding to open information only.

OK,I get your point

wu-sheng:
Should we close this? Or you want to add more context?

xuanyu66:
> Should we close this? Or you want to add more context?

yes, close this

PR #6913, opened 2021-05-08
Fix possible memory leaks in mybatis plugin.

Ax1an:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6838.

When the span count in segment exceeds the span_limit_per_segment limit, it will cause some NoopSpan to not be closed and memory leakage will occur.

Before fix:
image
image

After fix:
image


apm-sniffer/optional-plugins/mybatis-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mybatis/SqlSessionOperationInterceptor.java
```diff
@@ -40,7 +41,8 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
         if (ContextManager.isActive()) {
             lastSpan = ContextManager.activeSpan();
         }
-        if (lastSpan == null || lastSpan.getComponentId() != ComponentsDefine.MYBATIS.getId()) {
+        boolean isMyBatisEntryMethod = !(lastSpan instanceof NoopSpan) && (lastSpan == null || lastSpan.getComponentId() != ComponentsDefine.MYBATIS.getId());
```
wu-sheng, line 44: Why does NoopSpan matter? Could you be more clear?

wu-sheng: NoopSpan should be 100% transparent to the plugin, there is no case we should process like this.
I know you are facing a memory leak, but what you did, I would say, it is caused by checking component id. Could you be clear why the logic exists before?

wu-sheng: The plugin wouldn't be notified if the core changes the implementation one day. So, we need to polish the logic here.

Ax1an: @wu-sheng Yes, you are right. I don't want to enhance too many overloaded methods to waste resources.

wu-sheng: This check is actually costing resources, much more than intercepting multiple methods. Because intercepting different methods doesn't happen in the runtime, so, no latency cost.

wu-sheng: I would like to recommend you intercept explicit methods rather than using this check. The existing or new ones are both not good choices. We actually should not merge this in the beginning.

Ax1an:
image

wu-sheng (approved):

PR #6916, opened 2021-05-08
Add OpenSearch as a storage option

### LINKED ISSUE #6745: Add AWS OpenSearch as storage option
@kezhenxu94 As AWS has finally announced the APLv2 distribution of ElasticSearch, we should evaluate the possibility and impact if SkyWalking use it as the storage to replace the ElasticSearch as the recommended one, especially for open-source users. We should also evaluate the possibility and practicability of removing the ElasticSearch's client if they don't release APLv2-licensed client anymore.

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6745.

### DIFF
--- a/docs/en/setup/backend/backend-storage.md
+++ b/docs/en/setup/backend/backend-storage.md
@@ -9,6 +9,7 @@ storage:
 
 Native supported storage
 - H2
+- OpenSearch
 - ElasticSearch 6, 7
 - MySQL
 - TiDB
@@ -31,14 +32,20 @@ storage:
     user: sa
 ```
 
+## OpenSearch
+
+OpenSearch storage shares the same configurations as ElasticSearch 7.
+In order to activate ElasticSearch 7 as storage, set storage provider to **elasticsearch7**.
+Please download the `apache-skywalking-bin-es7.tar.gz` if you want to use OpenSearch as storage.
+
 ## ElasticSearch
 - In order to activate ElasticSearch 6 as storage, set storage provider to **elasticsearch**
 - In order to activate ElasticSearch 7 as storage, set storage provider to **elasticsearch7**
 
 **Required ElasticSearch 6.3.2 or higher. HTTP RestHighLevelClient is used to connect server.**
 
-- For ElasticSearch 6.3.2 ~ 7.0.0 (excluded), please download the `apache-skywalking-bin.tar.gz` or `apache-skywalking-bin.zip`,
-- For ElasticSearch 7.0.0 ~ 8.0.0 (excluded), please download the `apache-skywalking-bin-es7.tar.gz` or `apache-skywalking-bin-es7.zip`.
+- For ElasticSearch 6.3.2 ~ 7.0.0 (excluded), please download the `apache-skywalking-bin.tar.gz`,
+- For ElasticSearch 7.0.0 ~ 8.0.0 (excluded), please download the `apache-skywalking-bin-es7.tar.gz`.
 
 For now, ElasticSearch 6 and ElasticSearch 7 share the same configurations, as follows:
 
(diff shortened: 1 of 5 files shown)


docs/en/setup/backend/backend-storage.md
```diff
@@ -10,6 +10,7 @@ storage:
 Native supported storage
 - H2
 - ElasticSearch 6, 7
+- OpenSearch
```
wu-sheng, line 13: This has different order with the detail.

docs/en/setup/backend/backend-storage.md
```diff
@@ -31,14 +32,20 @@ storage:
...
+
+OpenSearch storage shares the same configurations as ElasticSearch 7.
+In order to activate ElasticSearch 7 as storage, set storage provider to **elasticsearch7**.
+Please download the `apache-skywalking-bin-es7.tar.gz` if you want to use OpenSearch as storage.
+
 ## ElasticSearch
 - In order to activate ElasticSearch 6 as storage, set storage provider to **elasticsearch**
 - In order to activate ElasticSearch 7 as storage, set storage provider to **elasticsearch7**
 
 **Required ElasticSearch 6.3.2 or higher. HTTP RestHighLevelClient is used to connect server.**
```
wu-sheng, line 46: We need a NOTICE here to indicate from which versions, ElasticSearch became to be licensed by SSPL, which is not compatible with Apache 2.0 anymore.

> NOTICE: MySQL driver is NOT allowed in Apache official distribution and source codes. Please download MySQL driver by yourself. Copy the connection driver jar to oap-libs.

kezhenxu94: We don't know yet from which version, do we? Also, we're still using APLv2 licensed jars, why bother telling users to download by themselves before we have to remove ES client jars in our distribution?

wu-sheng: The client API jar's license is not all. Because the dependency is checked from the runtime. We should say after which versions(released), they changed LICENSE. No matter Elastic called these new versions, they are already making us to depend on SSPL licensed server.

kezhenxu94: I just don’t see the point to add the notice to the users because the clients jar we are using now can be safely used no matter what server version the users use. And if the users want to download more recent versions (SSPL), that’s Elastic’s or / and users’ duty to check their license issue. The reason why MySQL jar needs a notice is that IMO it needs users action explicitly otherwise our project won’t work when using MYSQL as storage.

wu-sheng: We as ASF TLP, definitely have obligation to tell end users, there is License incompatible risk.
The foundation requires us for public good, and our release should make sure of it. Apache license v2 is one of the most important things as the foundation.

wu-sheng: We write the source codes to tell users, we recommend Elasticsearch as the storage option. We shouldn't say this is your duty about legal or license usage.

wu-sheng: I wouldn't say SSPL is a bad thing, I just argue from the ASF perspective.

wu-sheng (approved):
Just a NOTICE for the future. 

In the v6 release, the current 6.8.15(latest) is still in Apache 2.0 LICENSE.
All releases in 7.11 and 7.12 are released in SSPL and Elastic License, not compatible with Apache 2.0.
We could still test the compatibility with higher version server releases.

Also, I checked, https://mvnrepository.com/artifact/org.elasticsearch.client/elasticsearch-rest-high-level-client. The client API jar 7.11 and 7.12 are already licensed in SSPL License only in the binary.

PR #6917, opened 2021-05-09
Fix wrong implementation of "Double-Checked Locking".

YunfengGao:
### Fix wrong implementation of "Double-Checked Locking".
Add volatile to avoid only parts fields of DEFAULT_LOADER are visible, reference http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/loader/AgentClassLoader.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/loader/AgentClassLoader.java
@@ -57,7 +57,7 @@ public class AgentClassLoader extends ClassLoader {
     /**
      * The default class loader for the agent.
      */
-    private static AgentClassLoader DEFAULT_LOADER;
+    private static volatile AgentClassLoader DEFAULT_LOADER;
 
     private List<File> classpath;
     private List<Jar> allJars;


wu-sheng:
Are you facing any real error? Or this is just a fix in theory perspective.

wu-sheng:
Because this initialization actually is in one and only one thread of the premain.

YunfengGao:
> Are you facing any real error? Or this is just a fix in theory perspective.

Just in theory perspective，has a very small probability of occurrence in multi-threaded scenarios.

YunfengGao:
> Because this initialization actually is in one and only one thread of the premain.

`synchronized` here confused me that there is a multi-threaded scenario, haha. 
```Java
public static void initDefaultLoader() throws AgentPackageNotFoundException {
    if (DEFAULT_LOADER == null) {
        synchronized (AgentClassLoader.class) {
            if (DEFAULT_LOADER == null) {
                DEFAULT_LOADER = new AgentClassLoader(PluginBootstrap.class.getClassLoader());
            }
        }
    }
}
```

wu-sheng:
We used to setup parallel unit tests, so used to add this, as it has no side effect in the runtime.
We will initial this in the very beginning, may use it in multiple threada due to parallel class loader.

wu-sheng:
I think it is fine to keep it as it is. Any concern?

PR #6918, opened 2021-05-09
fix pulsar plugin message listener error with multi partitions topic

wallezhang:
### Fix pulsar plugin message listener error with multi partitions topic
- [x] Related issue #6730 

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/MessageConstructorInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/MessageConstructorInterceptor.java
@@ -18,6 +18,7 @@
 
 package org.apache.skywalking.apm.plugin.pulsar;
 
+import org.apache.pulsar.client.impl.TopicMessageImpl;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ClassEnhancePluginDefine;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
@@ -34,6 +35,12 @@ public class MessageConstructorInterceptor implements InstanceConstructorInterce
 
     @Override
     public void onConstruct(EnhancedInstance objInst, Object[] allArguments) {
-        objInst.setSkyWalkingDynamicField(new MessageEnhanceRequiredInfo());
+        final Object msgArgument = allArguments[2];
+        if (objInst instanceof TopicMessageImpl && msgArgument instanceof EnhancedInstance) {
+            objInst.setSkyWalkingDynamicField(((EnhancedInstance) msgArgument).getSkyWalkingDynamicField());
+        } else {
+            objInst.setSkyWalkingDynamicField(new MessageEnhanceRequiredInfo());
+        }
+
     }
 }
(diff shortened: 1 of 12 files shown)


apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/MessageConstructorInterceptor.java
```diff
@@ -34,6 +35,12 @@
 
     @Override
     public void onConstruct(EnhancedInstance objInst, Object[] allArguments) {
-        objInst.setSkyWalkingDynamicField(new MessageEnhanceRequiredInfo());
+        final Object msgArgument = allArguments[2];
+        if (objInst instanceof TopicMessageImpl && msgArgument instanceof EnhancedInstance) {
```
wu-sheng, line 39: If you need to separate the type of current object, why not use two different instrumentation definitions?

wu-sheng (commented):
@codelipenghui Please review

apm-sniffer/apm-sdk-plugin/pulsar-plugin/src/main/java/org/apache/skywalking/apm/plugin/pulsar/MessageConstructorInterceptor.java
wallezhang: The enhanced class is an interface, I want to enhance all the implementation classes, but some of them are used for wrapping other implementation, so they are only distinguished by the parameters of the constructor instead of using another instrumentation definitions.

wu-sheng: I was talking about objInst. 
Are you saying you could give a full list of implementors? Interface based is not a stable, you don't know how people will or have extended.

wallezhang: Yes, it is indeed not stable. The specific types are also checked in the pulsar client code, and I think I can change the enhanced class into two certain types of implementors.

wu-sheng: Thanks. Targeting the specific classes is already a better idea, and the codes would be more clear and have better performance.

wallezhang: done

wu-sheng: Thanks. Let's wait for @codelipenghui 's review.

test/plugin/scenarios/pulsar-scenario/src/main/java/test/apache/skywalking/apm/testcase/pulsar/controller/CaseController.java
```diff
@@ -48,10 +53,25 @@
...
     @ResponseBody
-    public String pulsarCase() throws PulsarClientException, InterruptedException {
+    public String pulsarCase() {
 
-        String topic = "test";
+        String topicOnePartition = "test";
+        String topicMultiPartition = "testMultiPartition";
 
+        try {
+            createTopic(topicOnePartition, 1);
```
codelipenghui, line 62: From this change, the test only covered the partitioned topic, we can call create topic only for the partitions > 1?

wallezhang: OK,I will call `doSendAndReceiveMessage` method directly for non-partitioned topic.

wallezhang: done

wu-sheng:
@codelipenghui Please check again.

codelipenghui (approved):

wu-sheng (approved):
Forward @codelipenghui 's approval.

PR #6937, opened 2021-05-13
Introduce method interceptor v2

### LINKED ISSUE #6915: Introduce method interceptor API v2
Hi @apache/skywalking-committers 

We have a very stable and powerful plugin for 4 years, all of our plugins are being built on top of it. With more scenarios reported, I noticed there is a small gap we should fill.
In `InstanceMethodsAroundInterceptor` and `StaticMethodsAroundInterceptor` have `before` and `after` methods, if we want to propagate information between them in one intercepting operation, we right now have
1. Dynamic field
2. RunnintContext
3. Enhanced parameter

But when we face concurrency calling of one instance, highly performance-sensitive, and only have `primitive data types` and `string` as parameter, all above three are not good options.

So, we should consider to provide a new instance/static method interceptor API(v2) like this.
```java
/**
 * A interceptor, which intercept method's invocation. The target methods will be defined in {@link
 * ClassEnhancePluginDefine}'s subclass, most likely in {@link ClassInstanceMethodsEnhancePluginDefine}
 */
public interface InstanceMethodsAroundInterceptorV2 {
    /**
     * called before target method invocation.
     *
     * @param result change this result, if you want to truncate the method.
     */
    void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
        MethodInterceptResult result, MethodInvocationContext mic) throws Throwable;

    /**
     * called after target method invocation. Even method's invocation triggers an exception.
     *
     * @param ret the method's original return value. May be null if the method triggers an exception.
     * @return the method's actual return value.
     */
    Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
        Object ret, MethodInvocationContext mic) throws Throwable;

    /**
     * called when occur exception.
     *
     * @param t the exception occur.
     */
    void handleMethodException(EnhancedInstance objInst, Method method, Object[] allArguments,
        Class<?>[] argumentsTypes, Throwable t, MethodInvocationContext mic);
}
```

A new object of `class MethodInvocationContext` is shared per method invocation, which should be created and managed by core inter(s), such as `InstMethodsInter`.

With this, we could share status in the method static and 100% isolation with other methods and static. When we use this with RuntimeContext, in `afterMethod` and `handleMethodException`, we could be easy to know whether some variable in the RunningContext is put or updated by `beforeMethod`.

Notice, this change also affects the bootstrap instrumentation, so, we need the v2 inter templates too.

___
To anyone who wants to handle this, you should know, this is a very core level API enhancement. We expect 100% forward compatibility and will accept the new plugin implementing either in v1 or v2 APIs. v1 may not be removed ever.

This task only recommends to SkyWalking committer or you are confident you have already known SkyWalking's agent core very well. Newcomers, please don't try this.

dmsolr:
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6915 .


wu-sheng:
Do you want me to review? Seems CI has failed, and are JDK-HTTP plugin changes for testing only?

wu-sheng:
CI and related plugin tests fail, please recheck.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/v2/MethodInvocationContext.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.v2;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class MethodInvocationContext {
+    private Map<String, Object> context = new HashMap<>(1);
```
wu-sheng, line 24: I don't think we need a complex map to enhance the context, a `Object` field should be enough. If a map is really required, the plugin could build one by itself. A general map for all plugins could cost a load to GC and memory.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/v2/InstanceMethodsAroundInterceptorV2.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.v2;
+
+import java.lang.reflect.Method;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+
+public interface InstanceMethodsAroundInterceptorV2 {
```
wu-sheng, line 24: Many codes lack comments. You are writing the agent core level things.

apm-sniffer/apm-toolkit-activation/pom.xml
```diff
@@ -26,7 +26,6 @@
     <modelVersion>4.0.0</modelVersion>
     <packaging>pom</packaging>
     <modules>
-        <module>apm-toolkit-kafka-activation</module>
```
wu-sheng, line 29: Why remove this? Wrong merge?

apm-sniffer/bootstrap-plugins/jdk-http-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdk/http/HttpClientNewInstanceInterceptor.java
```diff
@@ -18,23 +18,23 @@
...
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.v2.MethodInvocationContext;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.v2.StaticMethodsAroundInterceptorV2;
 
-public class HttpClientNewInstanceInterceptor implements StaticMethodsAroundInterceptor {
+public class HttpClientNewInstanceInterceptor implements StaticMethodsAroundInterceptorV2 {
 
     @Override
     public void beforeMethod(Class clazz, Method method, Object[] allArguments, Class<?>[] parameterTypes,
-        MethodInterceptResult result) {
+                             MethodInterceptResult result, MethodInvocationContext context) {
```
wu-sheng, line 31: I am feeling `MethodInterceptResult` and `MethodInvocationContext` should be merged, as result to override the execution result is a case of MIC usages.

wu-sheng:
Besides the above, I think you also should find a plugin(not bootstrap) to verify the new APIs for AppClassLoader scope. such as the last MyBatis plugin, or something you are familiar with.

apm-sniffer/apm-toolkit-activation/pom.xml
dmsolr: yeah. Wrong pushed

apm-sniffer/apm-sdk-plugin/avro-plugin/src/main/java/org/apache/skywalking/apm/plugin/avro/AbstractRequestInterceptor.java
```diff
@@ -24,13 +24,15 @@
...
 import org.apache.skywalking.apm.agent.core.logging.api.ILog;
 import org.apache.skywalking.apm.agent.core.logging.api.LogManager;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
-import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.v2.InstanceMethodsAroundInterceptorV2;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.v2.MethodInvocationContext;
 
-public abstract class AbstractRequestInterceptor implements InstanceConstructorInterceptor, InstanceMethodsAroundInterceptor {
+public abstract class AbstractRequestInterceptor implements InstanceConstructorInterceptor, InstanceMethodsAroundInterceptorV2 {
```
wu-sheng, line 35: Why choose this plugin about changing to v2 APIs? It seems not to match the case. `MethodInvocationContext` is never used, as we mentioned, we only recommend to use v2 API when it is more suitable.

dmsolr: Here tests plugin work with AppClassLoader.

wu-sheng: Yes, it is. But it is not suitable to change. The better case is about this plugin, https://github.com/apache/skywalking/pull/6913#pullrequestreview-654964408

dmsolr: make sense.

wu-sheng:
@dmsolr Your PR is not ready to merge, you don't need to sync the branch and use exact CI resources. No one is working on anything close to the agent core level. If someone is working on this, they have lower priority than this one.

dmsolr:
> @dmsolr Your PR is not ready to merge, you don't need to sync the branch and use exact CI resources. No one is working on anything close to the agent core level. If someone is working on this, they have lower priority than this one.

I got it. I know ci tasks will be canceled when it is duplicated. I updated branch before the ci tasks running. It doesn't waste resources.

apm-sniffer/optional-plugins/mybatis-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mybatis/define/MyBatisShellMethodInstrumentation.java
```diff
@@ -21,31 +21,32 @@
...
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.v2.ClassEnhancePluginDefineV2;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.v2.InstanceMethodsInterceptV2Point;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.v2.StaticMethodsInterceptV2Point;
 import org.apache.skywalking.apm.agent.core.plugin.match.ClassMatch;
 import org.apache.skywalking.apm.plugin.mybatis.MyBatisMethodMatch;
 
 import static org.apache.skywalking.apm.agent.core.plugin.match.NameMatch.byName;
 
-public class MyBatisShellMethodInstrumentation extends ClassInstanceMethodsEnhancePluginDefine {
+public class MyBatisShellMethodInstrumentation extends ClassEnhancePluginDefineV2 {
```
wu-sheng, line 32: Why extending `ClassEnhancePluginDefineV2`? Where is `ClassInstanceEnhancePluginDefineV2`

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/v2/ClassEnhancePluginDefineV2.java
```diff
@@ -0,0 +1,185 @@
...
+import static net.bytebuddy.matcher.ElementMatchers.isStatic;
+import static net.bytebuddy.matcher.ElementMatchers.not;
+
+/**
+ * This class controls all enhance operations, including enhance constructors, instance methods and static methods. All
+ * the enhances base on three types interceptor point: {@link ConstructorInterceptV2Point}, {@link
+ * InstanceMethodsInterceptV2Point} and {@link StaticMethodsInterceptV2Point} If plugin is going to enhance constructors,
+ * instance methods, or both, {@link ClassEnhancePluginDefineV2} will add a field of {@link Object} type.
+ */
+public abstract class ClassEnhancePluginDefineV2 extends AbstractClassEnhancePluginDefine {
```
wu-sheng, line 55: You miss a layer on the top of this for instance plugin and static method plugin, then developers don't have to always return Array[0].

apm-sniffer/optional-plugins/mybatis-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/mybatis/define/MyBatisShellMethodInstrumentation.java
dmsolr: I don't provide `ClassInstanceEnhancePluginDefineV2` and `ClassStaticMethodsEnhancePluginDefineV2`.

wu-sheng: Let's wait for CI to pass, then we should enhance this part.

dmsolr: BTW, jdk-http-plugin should be revert?

wu-sheng: After all of the codes are ready, we should. But we have some codes change for now.

wu-sheng (approved):
Once all tests passed, it is good to merge. 

Notice, JDK plugin test is done and reverted. 

We need to update the documentation later. @dmsolr

PR #6943, opened 2021-05-15
fix: a path of `menu.yml`

Jtrust:


wu-sheng (approved):

wu-sheng:
FYI @wankai123 I think last time, we broke this link

PR #6964, opened 2021-05-18
fix: Fix macos test script command not found error. #6846

### LINKED ISSUE #6846: Issue running Plugin test on macos

- [ ] Question or discussion
- [X] Bug
- [ ] Requirement

### Bug
Skywalking 8.6 code on master branch, macos (Big Sur), JDK 1.8

- Which company or project?
Plugin test framework fails with run.sh failing to run sed command on macos

- What happened?
Run Plugin test on macos as per [link](https://github.com/apache/skywalking/blob/master/docs/en/guides/Plugin-test.md#local-test-and-pull-request-to-the-upstream) on macos. it fails with error such as "invalid command .". After debugging for some time found that sed command used on between 145-156 in run.sh is not compatible on macos.

Either the document needs to be updates to support linux only and remove macos or fix the run.sh script.

wallezhang:
### Fix <#6846>

Fix the `run.sh` test script command not found error on MacOS. Because the syntax of `sed` command is different between macOS and Linux. So I use different syntax according to the current os.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6846.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ Release Notes.
 #### Project
 * Add OpenSearch as storage option.
 * Upgrade Kubernetes Java client dependency to 11.0.
+* Fix plugin test script error in macOS.
 
 #### Java Agent
 * Add `trace_segment_ref_limit_per_span` configuration mechanism to avoid OOM.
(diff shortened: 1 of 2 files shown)


wu-sheng:
Please update changelog in the project section.

wallezhang:
> Please update changelog in the project section.

Done

dmsolr (approved):
LGTM

wu-sheng:
@dmsolr Have you run this on a real MacOS?

dmsolr:
> @dmsolr Have you run this on a real MacOS?

Yes. I tested it passed on my laptop.

PR #6976, opened 2021-05-19
Fetch further data only when they're requested

kezhenxu94:
Follow up #6888 , nowadays our GraphQL queries always fetch all data from DB, and only some of them are requested by frontend (web ui), this is typically obvious in #6888 so this PR determines whether we should fetch events when querying alarms.

#6888 was no released yet so I don't update change log here.

@wu-sheng do you have any other similar situations in mind that we also need to check?

### DIFF
--- a/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
+++ b/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
@@ -97,7 +97,7 @@ public class AlarmQuery implements GraphQLQueryResolver {
         Alarms alarms = getQueryService().getAlarm(
             scopeId, keyword, paging, startSecondTB, endSecondTB, tags);
 
-        final boolean selectEvents = env.getSelectionSet().get().containsKey("items/events");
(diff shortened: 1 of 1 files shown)


wu-sheng (dismissed):
Make sense as an optimization. For other cases, we could use this as a protective way, but not many real cases.

wu-sheng:
Such as don't have `value` in metric query, but who uses in this way for the product env?

kezhenxu94:
> Such as don't have `value` in metric query, but who uses in this way for the product env?

This is not to map the requested fields to DB fields.

For the example you gave, we STILL query all fields in metrics table and let the GraphQL determine which fields to return. The overhead is tiny.

But for those queries that need multiple DB queries to compose a complex (and maybe large) result set (e.g. events in alarm), this becomes necessary. 

Also, since WebApp and OAP are deployed separately and OAP doesn't usually have incompatible changes from the perspective of WebApp, users may use old web app that is not aware of events in alarm while new OAP still queries them.

wu-sheng:
> But for those queries that need multiple DB queries to compose a complex (and maybe large) result set (e.g. events in alarm), this becomes necessary.

This PR is necessary, I am sure. I was just saying, there is no typical case like this one.

wu-sheng:
This seems breaking tests?

wu-sheng (approved):

Issue #7001, opened 2021-05-24
jdk-threading-plugin cause a memory leak


muyun12:
Please answer these questions before submitting your issue.

Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Bug
- CentOS 7.5
- JRE 8
- SkyWalking 6.5

- 路歌

- What happened?
In the following scene, it will cause a memory leak:
```java
	private static final ThreadPoolExecutor TPE = new ThreadPoolExecutor(4, 4,60 , TimeUnit.SECONDS, new ArrayBlockingQueue<>(1000));
	public static void main(String[] args) {
		// do something
		TPE.execute(new Runnable() {
			@Override
			public void run() {
				// do something
			}
		});
		TPE.shutdown();
		try {
			TPE.awaitTermination(2, TimeUnit.HOURS);
		} catch (InterruptedException e) {
			// do something
		}
	}
```

ThreadingMethodInterceptor class, as below：
```java
    @Override
    public void beforeMethod(final EnhancedInstance objInst, final Method method, final Object[] allArguments,
        final Class<?>[] argumentsTypes, final MethodInterceptResult result) {
        
        // this code cause a memory leak
        AbstractSpan span = ContextManager.createLocalSpan(generateOperationName(objInst, method));
        span.setComponent(ComponentsDefine.JDK_THREADING);

        final Object storedField = objInst.getSkyWalkingDynamicField();
        if (storedField != null) {
            final ContextSnapshot contextSnapshot = (ContextSnapshot) storedField;
            ContextManager.continued(contextSnapshot);
        }

    }
```

___

PR #7003, opened 2021-05-24
fixed jdk-threading-plugin memory leak #7001

muyun12:

fixed the issues #7001 

In the following scene, it will cause a memory leak:
```java
	private static final ThreadPoolExecutor TPE = new ThreadPoolExecutor(4, 4,60 , TimeUnit.SECONDS, new ArrayBlockingQueue<>(1000));
		// do something
		TPE.execute(new Runnable() {
			@Override
			public void run() {
				// do something
			}
		});
		TPE.shutdown();
		try {
			TPE.awaitTermination(2, TimeUnit.HOURS);
			// do something
		}
	}
```

ThreadingMethodInterceptor class, as below：
```java
    @Override
    public void beforeMethod(final EnhancedInstance objInst, final Method method, final Object[] allArguments,
        final Class<?>[] argumentsTypes, final MethodInterceptResult result) {
        
        // this code cause a memory leak
        AbstractSpan span = ContextManager.createLocalSpan(generateOperationName(objInst, method));
        span.setComponent(ComponentsDefine.JDK_THREADING);

        final Object storedField = objInst.getSkyWalkingDynamicField();
        if (storedField != null) {
            final ContextSnapshot contextSnapshot = (ContextSnapshot) storedField;
        }

    }
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,7 +31,7 @@ Release Notes.
 * Fixed part of the dynamic configuration of ConfigurationDiscoveryService that does not take effect under certain circumstances.
 * Introduce method interceptor API v2
 * Fix ClassCast issue for RequestHolder/ResponseHolder.
-* fixed `jdk-threading-plugin` memory leak #7001
+* fixed `jdk-threading-plugin` memory leak.
 
 #### OAP-Backend
 * BugFix: filter invalid Envoy access logs whose socket address is empty.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,6 +31,7 @@ Release Notes.
 * Fixed part of the dynamic configuration of ConfigurationDiscoveryService that does not take effect under certain circumstances.
 * Introduce method interceptor API v2
 * Fix ClassCast issue for RequestHolder/ResponseHolder.
+* fixed `jdk-threading-plugin` memory leak #7001
 
 #### OAP-Backend
 * BugFix: filter invalid Envoy access logs whose socket address is empty.
--- a/apm-sniffer/bootstrap-plugins/jdk-threading-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdk/threading/ThreadingMethodInterceptor.java
+++ b/apm-sniffer/bootstrap-plugins/jdk-threading-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdk/threading/ThreadingMethodInterceptor.java
@@ -34,10 +34,11 @@ public class ThreadingMethodInterceptor implements InstanceMethodsAroundIntercep
     public void beforeMethod(final EnhancedInstance objInst, final Method method, final Object[] allArguments,
         final Class<?>[] argumentsTypes, final MethodInterceptResult result) {
 
+        AbstractSpan span = ContextManager.createLocalSpan(generateOperationName(objInst, method));
+        span.setComponent(ComponentsDefine.JDK_THREADING);
+
         final Object storedField = objInst.getSkyWalkingDynamicField();
         if (storedField != null) {
-            AbstractSpan span = ContextManager.createLocalSpan(generateOperationName(objInst, method));
-            span.setComponent(ComponentsDefine.JDK_THREADING);
             final ContextSnapshot contextSnapshot = (ContextSnapshot) storedField;
             ContextManager.continued(contextSnapshot);
         }
--- a/apm-sniffer/bootstrap-plugins/jdk-threading-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdk/threading/ThreadingMethodInterceptor.java
+++ b/apm-sniffer/bootstrap-plugins/jdk-threading-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdk/threading/ThreadingMethodInterceptor.java
@@ -47,12 +47,7 @@ public class ThreadingMethodInterceptor implements InstanceMethodsAroundIntercep
     @Override
     public Object afterMethod(final EnhancedInstance objInst, final Method method, final Object[] allArguments,
         final Class<?>[] argumentsTypes, final Object ret) {
-
-        final Object storedField = objInst.getSkyWalkingDynamicField();
-        if (storedField != null) {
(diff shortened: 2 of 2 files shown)


kezhenxu94 (commented):
Thanks, can you please update CHANGES.md?

apm-sniffer/bootstrap-plugins/jdk-threading-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdk/threading/ThreadingMethodInterceptor.java
```diff
@@ -34,11 +34,10 @@
     public void beforeMethod(final EnhancedInstance objInst, final Method method, final Object[] allArguments,
         final Class<?>[] argumentsTypes, final MethodInterceptResult result) {
 
-        AbstractSpan span = ContextManager.createLocalSpan(generateOperationName(objInst, method));
-        span.setComponent(ComponentsDefine.JDK_THREADING);
-
         final Object storedField = objInst.getSkyWalkingDynamicField();
         if (storedField != null) {
+            AbstractSpan span = ContextManager.createLocalSpan(generateOperationName(objInst, method));
+            span.setComponent(ComponentsDefine.JDK_THREADING);
```
wu-sheng, line 40: I am not sure this is the right fix. Why not remove `afterMethod`'s `ContextManager.stopSpan()`, as we are always creating local span in `beforeMethod`?

muyun12: Yes, `beforeMethod` is always creating local span，and `afterMethod` can not clear `ContextManager`'s `CONTEXT` and `RUNTIME_CONTEXT`, that will cause a memory leak.

Add `objInst.getSkyWalkingDynamicField()` check，the problem of memory leak can be solved by keeping the logic in beforeMethod.

In the ThreadingConstructorInterceptor class:
```java
       // in our sence ContextManager.isActive() will return false，cause 
            objInst.setSkyWalkingDynamicField(ContextManager.capture());
        }
```
In the ThreadingMethodInterceptor  class:
```
    @Override
    public void beforeMethod(final EnhancedInstance objInst, final Method method, final Object[] allArguments,
        final Class<?>[] argumentsTypes, final MethodInterceptResult result) {

        // LocalSpan is always created here and referenced by the worker of ThreadPool
        AbstractSpan span = ContextManager.createLocalSpan(generateOperationName(objInst, method));
        span.setComponent(ComponentsDefine.JDK_THREADING);

        final Object storedField = objInst.getSkyWalkingDynamicField();
        if (storedField != null) {
            final ContextSnapshot contextSnapshot = (ContextSnapshot) storedField;
        }

    }

    @Override
    public Object afterMethod(final EnhancedInstance objInst, final Method method, final Object[] allArguments,
        final Class<?>[] argumentsTypes, final Object ret) {
        // ContextManager.CONTEXT cannot be cleaned up here
        final Object storedField = objInst.getSkyWalkingDynamicField();
        if (storedField != null) {
            ContextManager.stopSpan();
        }

        return ret;
    }
`````

**Comment by muyun12** [comment]
> Thanks, can you please update CHANGES.md?

Has been updated.

File: apm-sniffer/bootstrap-plugins/jdk-threading-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdk/threading/ThreadingMethodInterceptor.java (hunk 1 above)
Comment (wu-sheng) [reply apm-sniffer/bootstrap-plugins/jdk-threading-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdk/threading/ThreadingMethodInterceptor.java:40]: Do you really read what I said? I never said there is no memory leak. I was saying what is the better way to fix.

Comment (muyun12) [reply apm-sniffer/bootstrap-plugins/jdk-threading-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdk/threading/ThreadingMethodInterceptor.java:40]: OK, I'll recheck if there's something wrong with our known scenes, then submit a new version.

Comment (wu-sheng) [reply apm-sniffer/bootstrap-plugins/jdk-threading-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdk/threading/ThreadingMethodInterceptor.java:40]: The key is, if a `runnable#run` is called, when should create a local span. Your logic is, only creating when dynamicField has value, my question is should we create it always, just with or without reference.

Comment (muyun12) [reply apm-sniffer/bootstrap-plugins/jdk-threading-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdk/threading/ThreadingMethodInterceptor.java:40]: `beforeMethod` should always create local span，otherwise the trace of some scenarios will be missing. Just make sure that `afterMethod` calls the `ContextManager.stopSpan()` method to clear the `ContextManager`'s `CONTEXT` and `RUNTIME_CONTEXT` reference, it won't cause a memory leak.

Comment (wu-sheng) [reply apm-sniffer/bootstrap-plugins/jdk-threading-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdk/threading/ThreadingMethodInterceptor.java:40]: What are you doing? You are still don't using the same condition to create and stop span. I am very confused.

Comment (muyun12) [reply apm-sniffer/bootstrap-plugins/jdk-threading-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdk/threading/ThreadingMethodInterceptor.java]: I'm sorry there was a problem with the code submission.

File: CHANGES.md [hunk 2]
```diff
@@ -31,6 +31,7 @@ Release Notes.
 * Fixed part of the dynamic configuration of ConfigurationDiscoveryService that does not take effect under certain circumstances.
 * Introduce method interceptor API v2
 * Fix ClassCast issue for RequestHolder/ResponseHolder.
+* fixed `jdk-threading-plugin` memory leak #7001
```
wu-sheng, line 34:
```suggestion
* fixed `jdk-threading-plugin` memory leak.
```
Git changelog will show the related issue. So, don't worry.

wu-sheng (approved):

PR #7005, opened 2021-05-25
Fix counter misuse in the alarm core. Alarm can't be triggered in time.

### LINKED ISSUE #7000: The meaning of 'Counter' variable in RunningRule

- [x] Question or discussion
- [ ] Bug
- [ ] Requirement

___
### Question
- What do you want to know?
  - org.apache.skywalking.oap.server.core.alarm.provider.RunningRule
		- what do variable 'Counter' mean?
		- why need a Counter variable?
		- the method 'isMatch' use the 'countThreshold' variable, but method 'checkAlarm' also use the 'countThreshold' varaible to compare, I cannot understant why compare it twice.

```java
public Optional<AlarmMessage> checkAlarm() {
            if (isMatch()) {
                /*
                 * When
                 * 1. Metrics value threshold triggers alarm by rule
                 * 2. Counter reaches the count threshold;
                 * 3. Isn't in silence stage, judged by SilenceCountdown(!=0).
                 */
                counter++;
                if (counter >= countThreshold && silenceCountdown < 1) {
                    silenceCountdown = silencePeriod;
                    return Optional.of(new AlarmMessage());
                } else {
                    silenceCountdown--;
                }
            } else {
                silenceCountdown--;
                if (counter > 0) {
                    counter--;
                }
            }
            return Optional.empty();
        }

private boolean isMatch() {
            int matchCount = 0;
            for (Metrics metrics : values) {
                if (metrics == null) {
                    continue;
                }

            // xxx
            // xxx
            
            // Reach the threshold in current bucket.
            return matchCount >= countThreshold;
        }
```

___
	- There is a example:
		- window size: 5
		- window threshold: 3
		- initial window data: 0 1 1 0 1 (0 means not matched, 1 means matched), and later data is all 0 when window move
	- this example rule will not be triggered by followed rule
	  - window size: 5
	  - window threshold: 3
	- I think it should be triggered, but it's not. So it confuse me a lot.

wu-sheng:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7000.

The counter has been used twice, inside `isMatch` and `checkAlarm`. As a result, the counter is hard to reach unexpectedly.
I removed the duplicated logic, and make it only works in `isMatch`. Also fix all UTs, which are written to adopt rather than locate this bug.

Please recheck.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -51,6 +51,7 @@ Release Notes.
 * Fix: `!=` is not supported in oal when parameters are numbers.
 * Include events of the entity(s) in the alarm.
 * Support `native-json` format log in kafka-fetcher-plugin.
+* Fix counter misuse in the alarm core. Alarm can't be triggered in time.
 
 #### UI
 * Add logo for kong plugin.
--- a/oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
+++ b/oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
@@ -29,7 +29,6 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-
 import lombok.RequiredArgsConstructor;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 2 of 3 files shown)


kezhenxu94 (approved):
LGTM, seems CHANGLOGS.md is not updated but you checked the checkbox

wu-sheng:
> LGTM, seems CHANGLOGS.md is not updated but you checked the checkbox

Forgot to submit it. Changed locally. Done.

kezhenxu94 (approved):

PR #7016, opened 2021-05-26
fix bug for kafka report plugin

darcydai:
the Kafka report plugin has a bug that when Kafka bootstrap server is unavailable, I found AdminClient will create a thread to check the Kafka bootstrap server status in a loop. As time elapses more and more thread will be created, so should close it when creating Kafka producer failed. because 8.6.0 not release ,so i am not update change log


wu-sheng (approved):
Seems good. Wait for e2e.

PR #7017, opened 2021-05-26
Optimize Class.getDeclaredField in Feign plugin

lujiajing1126:
### Improve the performance of `Class.getDeclaredField`
- [x] Add a benchmark for the improvement: Not necessary
- [x] The benchmark result: trivial
- [x] Links/URLs to the theory proof or discussion articles/blogs. https://stackoverflow.com/questions/16395881/getdeclaredfields-absolute-performance

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. No.

I am wondering if we can add `FieldUtil` and/or `ClassUtil` similar to `MethodUtil` in agent core since I've met a couple of cases where this helper is needed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,6 +32,7 @@ Release Notes.
 * Introduce method interceptor API v2
 * Fix ClassCast issue for RequestHolder/ResponseHolder.
 * fixed `jdk-threading-plugin` memory leak.
+* Optimize multiple field reflection opeartion in Fiegn plugin.
 
 #### OAP-Backend
 * BugFix: filter invalid Envoy access logs whose socket address is empty.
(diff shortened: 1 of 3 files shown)


wu-sheng (dismissed):

lujiajing1126:
> > I am wondering if we can add FieldUtil and/or ClassUtil similar to MethodUtil in agent core since I've met a couple of cases where this helper is needed.
> 
> We are trying to avoid moving things into the core for 2 reasons
> 
> 1. Once we are facing some JDK breaking level change(it happened, and definitely will happen again), the core is as safer as possible.
> 2. The core doesn't care how plugins work, if there are some things that should be shared, we could write another `plugin` to provide assistance, it doesn't have to be in the core level(agent.jar).

Thanks for the explanation! Reasonable for me.

Changelog modified.

wu-sheng (approved):

PR #7019, opened 2021-05-26
fix wrong parameterized type in MeterSystem

dmsolr:

I don't think `functionClass` inherits from MeterFunction. Actually, it was marked by @MeterFunction. 

https://github.com/apache/skywalking/blob/f6f5af1b437bd1f65d0f4452fb54429c2ee6e617/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/MeterSystem.java#L84

According to the above code, it seems like inherit from `AcceptableValue`. It works well because generic type would be removed(type erasure) at runtime.


wu-sheng (approved):
Seems yes. And compiling remove this checking, which makes no casting error.

kezhenxu94 (approved):
LGTM.

Interesting generic erasing

PR #7027, opened 2021-05-27
 Make the number of core worker in meter converter thread pool configurable

dmsolr:
…urable

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,6 @@ Release Notes.
 * Add OpenSearch as storage option.
 * Upgrade Kubernetes Java client dependency to 11.0.
 * Fix plugin test script error in macOS.
-* Make the number of core worker in meter converter thread pool configurable.
 
 #### Java Agent
 * Add `trace_segment_ref_limit_per_span` configuration mechanism to avoid OOM.
@@ -55,6 +54,7 @@ Release Notes.
 * Include events of the entity(s) in the alarm.
 * Support `native-json` format log in kafka-fetcher-plugin.
 * Fix counter misuse in the alarm core. Alarm can't be triggered in time.
+* Make the number of core worker in meter converter thread pool configurable.
 
 #### UI
 * Add logo for kong plugin.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@ Release Notes.
 * Add OpenSearch as storage option.
 * Upgrade Kubernetes Java client dependency to 11.0.
 * Fix plugin test script error in macOS.
+* Make the number of core worker in meter converter thread pool configurable.
 
 #### Java Agent
 * Add `trace_segment_ref_limit_per_span` configuration mechanism to avoid OOM.
--- a/docs/en/setup/backend/configuration-vocabulary.md
+++ b/docs/en/setup/backend/configuration-vocabulary.md
@@ -211,7 +211,7 @@ core|default|role|Option values, `Mixed/Receiver/Aggregator`. **Receiver** mode 
 | prometheus-fetcher | default | Read [fetcher doc](backend-fetcher.md) for more details | - | - |
 | - | - | active | Activate the Prometheus fetcher. | SW_PROMETHEUS_FETCHER_ACTIVE | false |
 | - | - | enabledRules | Enable rules. | SW_PROMETHEUS_FETCHER_ENABLED_RULES | self |
-| - | - | numConvertWorker | The maximize meter convert worker. | SW_PROMETHEUS_FETCHER_NUM_CONVERT_WORKER | half the number of CPU core(s) |   
+| - | - | maxConvertWorker | The maximize meter convert worker. | SW_PROMETHEUS_FETCHER_NUM_CONVERT_WORKER | -1(by default, half the number of CPU core(s)) |   
 | kafka-fetcher | default | Read [fetcher doc](backend-fetcher.md) for more details | - | - |
 | - | - | bootstrapServers | A list of host/port pairs to use for establishing the initial connection to the Kafka cluster. | SW_KAFKA_FETCHER_SERVERS | localhost:9092 |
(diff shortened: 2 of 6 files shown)


CHANGES.md
```diff
@@ -8,6 +8,7 @@ Release Notes.
 * Add OpenSearch as storage option.
 * Upgrade Kubernetes Java client dependency to 11.0.
 * Fix plugin test script error in macOS.
+* Make the number of core worker in meter converter thread pool configurable.
```
wu-sheng, line 11: This should be a backend changelog.

oap-server/server-fetcher-plugin/prometheus-fetcher-plugin/src/main/java/org/apache/skywalking/oap/server/fetcher/prometheus/provider/PrometheusFetcherConfig.java
```diff
@@ -28,13 +28,15 @@
 
 @Getter
 public class PrometheusFetcherConfig extends ModuleConfig {
+
+    private int meterConvertWorker = Runtime.getRuntime().availableProcessors() / 2;
```
wu-sheng, line 32: You will face a bug when `Runtime.getRuntime().availableProcessors()`=1, I think. 
And take a look at `BulkConsumePool`, this usually accepts a System ENV override.

kezhenxu94, line 32: the field name is different with the one in config file


```diff
- meterConvertWorker
+ maxConvertWorker
```

oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -337,7 +337,9 @@ envoy-metric:
 prometheus-fetcher:
   selector: ${SW_PROMETHEUS_FETCHER:-}
   default:
+    active: ${SW_PROMETHEUS_FETCHER_ACTIVE:false}
     enabledRules: ${SW_PROMETHEUS_FETCHER_ENABLED_RULES:"self"}
+    maxConvertWorker: ${SW_PROMETHEUS_FETCHER_NUM_CONVERT_WORKER:-}
```
kezhenxu94, line 342: I don't think we support `-` in these config items (only `selector` supports `-` I think)

kezhenxu94 (changes requested):

oap-server/server-fetcher-plugin/prometheus-fetcher-plugin/src/main/java/org/apache/skywalking/oap/server/fetcher/prometheus/provider/PrometheusFetcherConfig.java
hanahmily: I would like to default the worker number to `Runtime.getRuntime().availableProcessors()`(the min value should be 1). An env could override it.

wu-sheng: > I would like to default the worker number to `Runtime.getRuntime().availableProcessors()`(the min value should be 1). An env could override it.

Really need so many threads? This should not cost too many resources, as prom data structure is not complex and data set is not very big.

dmsolr: I will avoid this case, core == 1. But I don't think we need EnvUtil to do it, because we have provided a configuration option that can override the value.

docs/en/setup/backend/configuration-vocabulary.md
```diff
@@ -210,6 +210,8 @@ core|default|role|Option values, `Mixed/Receiver/Aggregator`. **Receiver** mode
 | - | - | maxMessageSize | Sets the maximum message size allowed to be received on the server. Empty means 4 MiB | - | 4M(based on Netty) |
 | prometheus-fetcher | default | Read [fetcher doc](backend-fetcher.md) for more details | - | - |
 | - | - | active | Activate the Prometheus fetcher. | SW_PROMETHEUS_FETCHER_ACTIVE | false |
+| - | - | enabledRules | Enable rules. | SW_PROMETHEUS_FETCHER_ENABLED_RULES | self |
+| - | - | numConvertWorker | The maximize meter convert worker. | SW_PROMETHEUS_FETCHER_NUM_CONVERT_WORKER | half the number of CPU core(s) |
```
kezhenxu94, line 214: Wrong name here too…

oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -337,7 +337,9 @@ envoy-metric:
 prometheus-fetcher:
   selector: ${SW_PROMETHEUS_FETCHER:-}
   default:
+    active: ${SW_PROMETHEUS_FETCHER_ACTIVE:false}
     enabledRules: ${SW_PROMETHEUS_FETCHER_ENABLED_RULES:"self"}
+    # maxConvertWorker: ${SW_PROMETHEUS_FETCHER_NUM_CONVERT_WORKER:8}
```
kezhenxu94, line 342: Commenting out this line is not a good idea, it makes the env var not work and the user have to edit this file. Please default this value to 0 and set it to cpuCore / 2 if it’s 0.

wu-sheng: 0 or negative, should make the codes falling back to core / 2.

kezhenxu94 (approved):

PR #7028, opened 2021-05-27
Events can be configured as alarm source

### LINKED ISSUE #7024: Make alarm rules work for events
Some events can directly and immediately trigger service failure, such as health-check failure, Pod OOM, etc., I think we can have alarm rules work for events as well, WDYT @wu-sheng , for example.

```yaml
  unhealthy_event_rule:
    metrics-name: Unhealthy
    threshold: 2
    op: ">"
    period: 10
    count: 1
```

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7024.

### DIFF
--- a/docs/en/concepts-and-designs/event.md
+++ b/docs/en/concepts-and-designs/event.md
@@ -60,8 +60,8 @@ There are also cases where you would already have both the start time and end ti
 Events are derived from metrics, and can be the source to trigger alarms. For example, if a specific event occurs for a
 certain times in a period, alarms can be triggered and sent.
 
-Every event has a default `count = 1`, when `n` events with the same name are reported, they are aggregated
-into `count = n` as follows.
+Every event has a default `value = 1`, when `n` events with the same name are reported, they are aggregated
+into `value = n` as follows.
 
 ```
 Event{name=Unhealthy, source={service=A,instance=a}, ...}
@@ -75,7 +75,7 @@ Event{name=Unhealthy, source={service=A,instance=a}, ...}
 will be aggregated into
 
 ```
-Event{name=Unhealthy, source={service=A,instance=a}, ...} <count = 6>
+Event{name=Unhealthy, source={service=A,instance=a}, ...} <value = 6>
 ```
 
 so you can configure the following alarm rule to trigger alarm when `Unhealthy` event occurs more than 5 times within 10
@@ -85,6 +85,10 @@ minutes.
 rules:
   unhealthy_event_rule:
     metrics-name: Unhealthy
+    # Healthiness check is usually a scheduled task,
+    # they may be unhealthy for the first few times,
+    # and can be unhealthy occasionally due to network jitter,
+    # please adjust the threshold as per your actual situation.
     threshold: 5
     op: ">"
     period: 10
@@ -93,6 +97,9 @@ rules:
 ```
 
 For more alarm configuration details, please refer to the [alarm doc](../setup/backend/backend-alarm.md).
+
+**Note** that the `Unhealthy` event above is only for demonstration, they are not detected by default in SkyWalking,
+however, you can use the methods in [How to Report Events](#how-to-report-events) to report this kind of events.
 
 ## Known Events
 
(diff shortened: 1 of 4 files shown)


docs/en/concepts-and-designs/event.md
```diff
@@ -55,10 +55,49 @@ The end time of the event. This field may be empty if the event has not ended ye
...
+```
+
+so you can configure the following alarm rule to trigger alarm when `Unhealthy` event occurs more than 5 times within 10
+minutes.
+
+```yaml
+rules:
+  unhealthy_event_rule:
+    metrics-name: Unhealthy
+    threshold: 5
```
wu-sheng, line 88: What threshold means? how many events per minute?

kezhenxu94: > What threshold means? how many events per minute?

yes

docs/en/concepts-and-designs/event.md
```diff
@@ -55,10 +55,49 @@ The end time of the event. This field may be empty if the event has not ended ye
...
+Event{name=Unhealthy, source={service=A,instance=a}, ...}
+```
+
+will be aggregated into
+
+```
+Event{name=Unhealthy, source={service=A,instance=a}, ...} <count = 6>
+```
+
+so you can configure the following alarm rule to trigger alarm when `Unhealthy` event occurs more than 5 times within 10
```
wu-sheng, line 81: I think you should give a tip here, as people are going to ask, how to identify `unhealthy`?

docs/en/concepts-and-designs/event.md
wu-sheng: Then, is 5 a good number of examples here? I think usually event threshold should be 1?

docs/en/concepts-and-designs/event.md
```diff
@@ -55,10 +55,49 @@ The end time of the event. This field may be empty if the event has not ended ye
...
+Event{name=Unhealthy, source={service=A,instance=a}, ...}
+Event{name=Unhealthy, source={service=A,instance=a}, ...}
+Event{name=Unhealthy, source={service=A,instance=a}, ...}
+Event{name=Unhealthy, source={service=A,instance=a}, ...}
+```
+
+will be aggregated into
+
+```
+Event{name=Unhealthy, source={service=A,instance=a}, ...} <count = 6>
```
wu-sheng, line 78: This count will be confused with the alarm rule's count. Should we consider rename this?

docs/en/concepts-and-designs/event.md
kezhenxu94: > Then, is 5 a good number of examples here? I think usually event threshold should be 1?

Healthiness check is usually a scheduled task, they may be unhealthy for the first few times, and can be unhealthy occasionally due to network jitter, I feel 5 is acceptable as just an example

wu-sheng: OK, then at least, this would be taken as a widely adopted example. So, put some explanation on the document.

docs/en/concepts-and-designs/event.md
```diff
@@ -55,10 +55,56 @@ The end time of the event. This field may be empty if the event has not ended ye
...
+    op: ">"
+    period: 10
+    count: 1
+    message: Service instance has been unhealthy for 10 minutes
+```
+
+For more alarm configuration details, please refer to the [alarm doc](../setup/backend/backend-alarm.md).
+
+**Note** that the `Unhealthy` event above is only for demonstration, they are not detected by default in SkyWalking,
+however, you can use the methods in [How to Report Events](#how-to-report-events) to report this kind of events.
```
wu-sheng, line 102: Does our k8s event detector have the capability to report pod unhealthy already?

kezhenxu94: > Does our k8s event detector have the capability to report pod unhealthy already?

Yes

wu-sheng: But, I can't see anything related to unhealthy here, https://github.com/apache/skywalking-kubernetes-event-exporter/blob/main/assets/default-config.yaml

wu-sheng: A reminder, this k8s-event-exporter repo is not available here, https://skywalking.apache.org/docs/

kezhenxu94: > A reminder, this k8s-event-exporter repo is not available here, https://skywalking.apache.org/docs/

Will update it when 1.0 is ready.

kezhenxu94: > But, I can't see anything related to unhealthy here, https://github.com/apache/skywalking-kubernetes-event-exporter/blob/main/assets/default-config.yaml

Will update in that repo

wu-sheng (approved):

Issue #6897, opened 2021-05-05
Support hot update of LAL and MAL


wu-sheng:
@kezhenxu94 For better 3rd party integration, we still need to evaluate the ways to update LAL and MAL in the runtime. Of course, the OAL should be on the radar, but considering it is heavier than these two, let's focus on LAL and MAL now.
I have talked with people around about the MAL and LAL use case, this is really required because people are keeping add metrics analysis with time.

From a tech perspective, groovy should be easy to support, the biggest challenge is about the streaming process. We need to support uninstall the existing process and tolerate `worker not found` in `RemoteServiceHandler`.

Also, the update should be supported in 2 ways.
1. Dynamic configuration setting, like alarm setting.
2. File change monitoring(`MultipleFilesChangeMonitor`), then, in single-node mode or dev mode, we don't need to use (1), as it is clearly more complex than directly changing file

I know, @kezhenxu94  have concerns about the security issue, we don't recommend definitely expose the script to the user, but use a system to assemble that. In further, we could have a better API set to assemble this script and avoid security issues.

At the last, about the storage format change, it is more about the MAL change, especially when face H2/MySQL/PostgreSQL, which could be a load for the storage, but once this is open, users should take care of it. New metrics installation about these 2, need to check whether the entity(table) exists, if yes, we should follow a new config to determine whether OAP should `delete the entity` or `reject the change and output an error log`.

@apache/skywalking-committers Welcome to discuss and comment. Whoever wants to implement this, please let me know.

liqiangz:
Is there any update？ I am interested in it.

wu-sheng:
@dmsolr told me, he will work on this. Please confirm the status.

wu-sheng:
This is too complex, and too dangerous to cause OOM and GC storm in the load traffic load. 

We wouldn't follow this, unless we have more ideas for a distributed system class level update.

PR #7039, opened 2021-05-30
WIP: Support hot update of MAL

dmsolr:

#6897

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsStreamProcessor.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsStreamProcessor.java
@@ -20,9 +20,10 @@ package org.apache.skywalking.oap.server.core.analysis.worker;
 
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
 import lombok.Getter;
 import lombok.Setter;
 import org.apache.skywalking.oap.server.core.CoreModule;
@@ -43,8 +44,11 @@ import org.apache.skywalking.oap.server.core.storage.annotation.Storage;
 import org.apache.skywalking.oap.server.core.storage.model.Model;
 import org.apache.skywalking.oap.server.core.storage.model.ModelCreator;
 import org.apache.skywalking.oap.server.core.storage.type.StorageBuilder;
+import org.apache.skywalking.oap.server.core.worker.AbstractWorker;
+import org.apache.skywalking.oap.server.core.worker.IWorkerInstanceRemover;
 import org.apache.skywalking.oap.server.core.worker.IWorkerInstanceSetter;
 import org.apache.skywalking.oap.server.library.module.ModuleDefineHolder;
+import org.apache.skywalking.oap.server.library.module.ModuleManager;
 
 /**
  * MetricsStreamProcessor represents the entrance and creator of the metrics streaming aggregation work flow.
@@ -62,7 +66,7 @@ public class MetricsStreamProcessor implements StreamProcessor<Metrics> {
     /**
      * Worker table hosts all entrance workers.
      */
-    private Map<Class<? extends Metrics>, MetricsAggregateWorker> entryWorkers = new HashMap<>();
+    private Map<Class<? extends Metrics>, MetricsAggregateWorker> entryWorkers = new ConcurrentHashMap<>();
 
     /**
      * Worker table hosts all persistent workers.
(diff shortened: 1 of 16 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsStreamProcessor.java
```diff
@@ -199,4 +203,20 @@ private MetricsPersistentWorker downSamplingWorker(ModuleDefineHolder moduleDefi
 
         return persistentWorker;
     }
+
+    // todo javadoc
+    public void remove(ModuleManager manager, String metricName, Class<? extends Metrics> metricsClass) {
```
wu-sheng, line 208: Tip, we need to check whether the worker would be GC successfully to avoid potential OOM risk.

wu-sheng: Question, once this gets removed, we don't touch the storage, right?

wu-sheng:
Let's be clear, in this PR, we focus on MAL update through physical file changing, right? We need to update the doc to indicate this.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsStreamProcessor.java
dmsolr: Yes, it is.

wu-sheng:
@dmsolr Any update about this?

dmsolr:
SkyWalking Metrics data comes from MeterSystem and OALEngine. The source of `MeterSystem` includes Receiver and Fetcher. 

> we had many receivers includes Zabbix, OpenTelmetry, JVM, and Meter. ..

Currently, we have had the Prometheus Fetcher which scraps metrics sampling from the Prometheus exporter. It could convert Prometheus Metrics to MeterSystem's Sample by MAL rules.

image


There are 3 types of operators, ADD, UPDATE and DELETE.
1. For ADD a new metrics, not conflict
2. For DELETE, there are 2 options, 1. None operator, 2. Always drop table.
3. For UPDATE, For now, we provided 3 options(Option c as default).
  1. Always keep old data(table)
  2. Always discard old data(table)
  3. Interrupted when table updated

Above description is not suitation for ElasticSearch. In RMDBS(mysql/postgresql), tidb and influxdb. A metrics represent a model, and a table(measuremnt in influxdb).  And the table name is the mode name. So, for update, 
1. Rename table name after a metrics rule udpated for keeping old data in DB.
2. Drop the table before creating a new metrics writing for discard old data.
3. None operation without the table structure changed. Otherwise, an exception is thrown for interrupting the OAP.

In ElasticSearch(after #6499), have many difference with other storage plugins. SW has put all metrics into one index. And identified metrics data by the function name and mode name. 

1. For ADD a new metrics, not conflict also.
2. For DELETE a metrics, 1. Waiting for TTL to clean up. 2.  Execution a query for delete.
3. For UPDATE, 
  1. Waiting for TTL to sweep.
  2. Execution delete statement.

In ElasticSearch case, there is no error cause by MAL updated.

wu-sheng:
> Interrupted when table updated

Does this mean OAP exiting?

dmsolr:
Yes. If OAP has had the same name rule exist and its expression different, we call it `update`.

wu-sheng:
> Yes. If OAP has had the same name rule exist and its expression different, we call it `update`.

This could happen from time to time in the debug mode, if this keeps exiting, I think this feature is going to useless.

wu-sheng:
> > Yes. If OAP has had the same name rule exist and its expression different, we call it `update`.
> 
> This could happen from time to time in the debug mode, if this keeps exiting, I think this feature is going to useless.

It is better to disable it if you have any concerns. But don't run in exit.

dmsolr:
> > > Yes. If OAP has had the same name rule exist and its expression different, we call it `update`.
> > 
> > 
> > This could happen from time to time in the debug mode, if this keeps exiting, I think this feature is going to useless.
> 
> It is better to disable it if you have any concerns. But don't run in exit.

We talk about non-elasticsearch case, right? (ES should not interrupt because MAL updated)

I got your point and agreed. That means **do not** interrupt OAP server.  So, rename the table. WDYT? 
To rename the table means users have to drop these old tables( because these unmanaged in TTL).

wu-sheng:
> > > > Yes. If OAP has had the same name rule exist and its expression different, we call it `update`.
> 
> > > 
> 
> > > 
> 
> > > This could happen from time to time in the debug mode, if this keeps exiting, I think this feature is going to useless.
> 
> > 
> 
> > It is better to disable it if you have any concerns. But don't run in exit.
> 
> 
> 
> We talk about non-elasticsearch case, right? (ES should not interrupt because MAL updated)
> 
> 
> 
> I got your point and agreed. That means **do not** interrupt OAP server.  So, rename the table. WDYT? 
> 
> To rename the table means users have to drop these old tables( because these unmanaged in TTL).

I think directly dropping table is fine. We could disable hot update in default.

dmsolr:
The PR still takes a long time to complete, so I closed it first before it is ready.

PR #7052, opened 2021-06-01
Fix Jetty HTTP `TRACE` issue, disable HTTP methods except `POST`.

wankai123:
### Fix Jetty HTTP `TRACE` issue, disable HTTP methods except `POST`
- [X] Add a unit test to verify that the fix works.
   
The `TRACE` is enabled in HttpServlet by default, inherited by `JettyHandler`, we need to disable it to avoid security issues.
Since we only use the `POST` method, I disabled others methods such as `GET/PUT/HEAD/OPTIONS/DELETE` to avoid the potential issues.


oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/jetty/JettyHandler.java
```diff
@@ -18,9 +18,64 @@
...
+    protected final void doOptions(HttpServletRequest req,
+                                   HttpServletResponse resp) throws ServletException, IOException {
+        super.doOptions(req, resp);
+    }
+
+    @Override
+    protected final void doTrace(HttpServletRequest req,
+                                 HttpServletResponse resp) throws ServletException, IOException {
+        super.doTrace(req, resp);
+    }
```
kezhenxu94, line 80: Since you have already override `protected final void service`, we don't need to override these methods, right? They are never invoked anymore

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/jetty/JettyServer.java
```diff
@@ -73,6 +73,12 @@ public void initialize() {
         LOGGER.info("http server root context path: {}", jettyServerConfig.getContextPath());
 
         server.setHandler(servletContextHandler);
+
+        JettyDefaultHandler defaultHandler = new JettyDefaultHandler();
+        ServletHolder defaultHolder = new ServletHolder();
+        defaultHolder.setServlet(defaultHandler);
+
+        servletContextHandler.addServlet(defaultHolder, defaultHandler.pathSpec());
```
kezhenxu94, line 81: Can you elaborate more on this, isn't it the default behavior (returning `HttpServletResponse.SC_NOT_FOUND `) when there is no mapped servlet can be found? Why do we need a default handler to do this?

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/jetty/JettyServer.java
```diff
@@ -133,4 +139,12 @@ public boolean equals(Object o) {
...
         return Objects.hash(jettyServerConfig.getHost(), jettyServerConfig.getPort());
     }
+
+    public void  stop() throws ServerException {
+        try {
+            server.stop();
+        } catch (Exception e) {
+            throw new JettyServerException(e.getMessage(), e);
+        }
+    }
```
kezhenxu94, line 149: This method is only for testing, we don't usually add methods only for testing, try to use `((org.eclipse.jetty.server.Server)Whitebox.getInternalState(SERVER, "server")).stop();` to do the same thing.

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/jetty/JettyHandler.java
wankai123: Make these methods final just to avoid other people to override them and then found it doesn't work.

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/jetty/JettyServer.java
wankai123: In `Jetty`, if we don't set a default handler they will set one `Default404Servlet`
image
But this servlet doesn't override `Service`, so the `TRACE` and others methods could get in from here like a back door.

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/jetty/JettyServer.java
wankai123: you are right, done

kezhenxu94 (approved):

PR #7062, opened 2021-06-03
change log REST endpoint with v3 prefixed

Shikugawa:

The previous name of log collection REST endpoint has inconsistency with other REST supported endpoint, such as `/v3/segment`. In this PR, we fix this problem.

### DIFF
--- a/docs/en/protocols/Log-Data-Protocol.md
+++ b/docs/en/protocols/Log-Data-Protocol.md
@@ -45,39 +45,37 @@ Json log record example:
 
 ## HTTP API
 
-Report `json` format logs via HTTP API, the endpoint is http://<oap-address>:12800/logs.
+Report `json` format logs via HTTP API, the endpoint is http://<oap-address>:12800/v3/logs.
 
 Json log record example:
 
 ```json
-[
-  {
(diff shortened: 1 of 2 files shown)


Shikugawa:
cc @kezhenxu94

wu-sheng (commented):
Make sense to me. We missed that before. @kezhenxu94

wu-sheng:
@kezhenxu94 Is this API new in 8.6? Whether we need changelog?

kezhenxu94:
It' OK to me to add that prefix, the reason why I didn't add it is because for logging API, we don't have historical versions, (v1, v2).

This changes don't need change log as logging REST API is not released yet

docs/en/protocols/Log-Data-Protocol.md
```diff
@@ -45,39 +45,37 @@ Json log record example:
...
-        {
-          "key": "logger",
-          "value": "com.example.MyLogger"
-        }
-      ]
-    },
-    "body": {
-      "text": {
-        "text": "log message"
+{
```
kezhenxu94, line 53: It's type should be array

kezhenxu94 (approved):

PR #7069, opened 2021-06-04
Agent supports the collection of JVM arguments and jar dependencies information.

Ax1an:

image

If the frontend is optimized, it will be more friendly.


wu-sheng:
@kezhenxu94 @Ax1an This is a thing I keep waiting for. Once we merge this, `InstanceTraffic` could combine the `jvm arguments` and `dependency libs` changes, which should be considered as a JVM level event. Redeploying with codes/setting upgrading is a very common case for a performance issue, such as `-XMX` change.

wu-sheng:
We should add a config list at the backend, to indicate which instance properties should be identified as an instance event.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
```diff
@@ -0,0 +1,190 @@
...
+
+    private static Set<String> JAR_FILE_LIST = new HashSet<>();
+
+    public static List<String> getJarFileNameList() {
+        List<String> jarFileNameList = new ArrayList<>();
+        for (String jarFile : JAR_FILE_LIST) {
+            String jarFileName = jarFile.substring(jarFile.lastIndexOf("/") + 1);
+            jarFileNameList.add(jarFileName);
+        }
+        return jarFileNameList;
```
kezhenxu94, line 59: For easy comparison lately when we want to detect events, what about sorting the list and save it to a variable?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
```diff
@@ -0,0 +1,190 @@
...
+                    LOGGER.warn("Error msg: {}", e.getMessage());
+                }
+            }
+        }
+        if (springBootJars.size() > 0) {
+            JAR_FILE_LIST.addAll(springBootJars);
+        }
+    }
+
+    private static String getPathSeparator() {
```
kezhenxu94, line 158: This method name doesn't match what it actually does

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
```diff
@@ -0,0 +1,190 @@
...
+        return System.getProperty("java.class.path");
+    }
+
+    public static List<String> getVmArgs() {
+        final List<String> vmArgs = ManagementFactory.getRuntimeMXBean().getInputArguments();
+        if (vmArgs == null) {
+            return Collections.emptyList();
+        }
+        return vmArgs;
```
kezhenxu94, line 167: Sort this for the same reason 👆 with jar file name list

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
```diff
@@ -0,0 +1,190 @@
...
+        loadSpringBootLib();
+    }
+
+    private static void loadSpringBootLib() {
+        List<String> springBootJars = new ArrayList<>();
+        for (String s : JAR_FILE_LIST) {
+            JarFile jar = null;
+            try {
+                jar = new JarFile(s);
+                if (jar.getManifest() == null || !"org.springframework.boot.loader.JarLauncher".equals(jar.getManifest().getMainAttributes().getValue("Main-Class"))) {
```
kezhenxu94, line 129: Can you elaborate more on this?

Ax1an: In order to start with a fat jar, spring boot extends the JarFile URL protocol of JDK, customizes its own classloader and jar file protocol handler, and implements the loading mode of jar in jar and jar in directory.

image

image

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
Ax1an: I'll adjust the name.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
Ax1an: Do you have any suggestions for sorting this list?

Ax1an:
@wu-sheng @kezhenxu94 I'll take a moment to learn about the logic of skywalking's events.

wu-sheng:
> @wu-sheng @kezhenxu94 I'll take a moment to learn about the logic of skywalking's events.

I think you could finish this PR first if you want this included in the 8.6.0 release. It is coming. Event mechanism is a backend feature.

Ax1an:
Sorted results: 

image

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/boot/ServiceManager.java
```diff
@@ -36,11 +38,13 @@
     INSTANCE;
 
     private static final ILog LOGGER = LogManager.getLogger(ServiceManager.class);
+    public static Instrumentation INSTRUMENTATION;
```
wu-sheng, line 41: Why this?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/boot/ServiceManager.java
```diff
@@ -36,11 +38,13 @@
...
     private Map<Class, BootService> bootedServices = Collections.emptyMap();
 
     public void boot() {
         bootedServices = loadAllServices();
 
+        JVMUtil.loadJvmInfo();
```
wu-sheng, line 47: Why called here?

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
```diff
@@ -130,6 +130,7 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
                     .installOn(instrumentation);
 
         try {
+            ServiceManager.INSTRUMENTATION = instrumentation;
```
wu-sheng, line 133: Why this?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/boot/ServiceManager.java
Ax1an: In order to make it more convenient to use instrumentation where necessary.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
Ax1an: In order to make it more convenient to use instrumentation where necessary.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/boot/ServiceManager.java
Ax1an: It seems that this position is not suitable. Do you have any suggestions?

Ax1an:
@wu-sheng According to your question, I made some adjustments.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
```diff
@@ -0,0 +1,193 @@
...
+            if (null != url) {
+                JAR_PATH_MAP.put(url.toString(), "");
+            }
+        } catch (Exception e) {
+            LOGGER.warn("Error msg: {}", e.getMessage());
+        }
+    }
+
+    private static void loadJarFileList() {
+        String[] classPathJar = getClasspath().split(OSUtil.getPathSeparator());
```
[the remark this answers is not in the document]
Ax1an: ignore

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
wu-sheng: This should not expose to only other developers. If anyone wants, build their own agent, it is not SkyWalking responsibility.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/boot/ServiceManager.java
wu-sheng: This is a static method, it could be anywhere you like, what do you mean?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/boot/ServiceManager.java
wu-sheng: This is not recommended way to expose. Metadata like this should be loaded by core directly. Not a service should do.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
```diff
@@ -0,0 +1,190 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+import org.apache.skywalking.apm.agent.core.logging.api.ILog;
+import org.apache.skywalking.apm.agent.core.logging.api.LogManager;
+import org.apache.skywalking.apm.agent.core.os.OSUtil;
+import org.apache.skywalking.apm.network.common.v3.KeyStringValuePair;
+
+public class JVMUtil {
```
wu-sheng, line 44: We need to polish this, comments and explanations are missing.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
```diff
@@ -0,0 +1,190 @@
...
+                classLoaders.add(clzz.getClassLoader());
+            }
+        }
+
+        for (ClassLoader classLoader : classLoaders) {
+            classResCalc(classLoader.getClass());
+        }
+    }
+
+    private static void classResCalc(Class c) {
```
wu-sheng, line 82: What is calc? Don't think this is a term.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
```diff
@@ -0,0 +1,190 @@
...
+        List<String> jarFileNameList = new ArrayList<>();
+        for (String jarFile : JAR_FILE_LIST) {
+            String jarFileName = jarFile.substring(jarFile.lastIndexOf("/") + 1);
+            jarFileNameList.add(jarFileName);
+        }
+        Collections.sort(jarFileNameList);
+        return jarFileNameList;
+    }
+
+    public static void loadJvmInfo(Instrumentation instrumentation) {
```
wu-sheng, line 62: Public methods should be on the top.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
```diff
@@ -0,0 +1,190 @@
...
+        loadJarPath(instrumentation);
+        loadJarFileList();
+    }
+
+    private static void loadJarPath(Instrumentation instrumentation) {
+        Class[] clzzs = instrumentation.getAllLoadedClasses();
```
wu-sheng, line 68: Why getting all loaded classes here? And class and jars may not be loaded here, such as lazy initialization.

Ax1an:
@wu-sheng @kezhenxu94 I've refactored the code and tested it.

wu-sheng (commented):
The key concern is still, the loaded jar set is not dynamically updated.

Ax1an:
@wu-sheng I adjusted the way to get the classloader list. And got the same test results as above.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
```diff
@@ -151,6 +152,7 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
                                                 final TypeDescription typeDescription,
                                                 final ClassLoader classLoader,
                                                 final JavaModule module) {
+            JVMUtil.getCurrentClassloaderList().add(classLoader);
```
wu-sheng, line 155: 1. `JVMUtil` should be renamed to `LoadedLibraryCollector`
2. `getCurrentClassloaderList().add` should be changed as `registerClassLoader`

wu-sheng: getCurrentClassloaderList + add would end in OOM. Keep using `ArrayList#add`? Would not this cause OOM?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * Build the required JVM information to add to the instance properties
+     */
+    public static List<KeyStringValuePair> buildJVMInfo() {
+        List<KeyStringValuePair> jvmInfo = new ArrayList<>();
+        jvmInfo.add(KeyStringValuePair.newBuilder().setKey("Start Time").setValue(getVmStartTime()).build());
+        Gson gson = new GsonBuilder().disableHtmlEscaping().create();
+        jvmInfo.add(KeyStringValuePair.newBuilder().setKey("JVM Arguments").setValue(gson.toJson(getVmArgs())).build());
+        List<String> libJarNames = getLibJarNames();
```
wu-sheng, line 59: You are doing this every time, don't ever consider performance impact. This mechanism would be very dangerous in the prod env.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
```diff
@@ -0,0 +1,158 @@
...
+        if (isLibJarNamesUpdated(libJarNames)) {
```
wu-sheng, line 60: If not updated, then you don't send the `Jar Dependencies`, but I think from backend logic, this would make the whole properties of instance changes, and lose the context. Because the backend doesn't know what do you mean.
Unless we have some specific value to indicate, it means value no change.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
```diff
@@ -0,0 +1,158 @@
...
+            jvmInfo.add(KeyStringValuePair.newBuilder().setKey("Jar Dependencies").setValue(gson.toJson(libJarNames)).build());
+            LAST_LIB_JAR_NAMES.clear();
+            LAST_LIB_JAR_NAMES.addAll(libJarNames);
+        }
+        return jvmInfo;
+    }
+
+    private static String getVmStartTime() {
+        long startTime = ManagementFactory.getRuntimeMXBean().getStartTime();
+        return new SimpleDateFormat("yyyy-MM-dd HH:mm:ss").format(new Date(startTime));
```
wu-sheng, line 70: cc @kezhenxu94 If you accept the start time, should we detect the changes base on the `start time` unchanged?
And another 2 questions
1. Should we have a document to list all property keys?
2. Should we consider a value holder or some mechanism to indicate value no change? Refer https://github.com/apache/skywalking/pull/7069/files#r647909609

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
Ax1an: I will optimize these places.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
Ax1an: You're right. I almost forgot the logic.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
kezhenxu94: > cc @kezhenxu94 If you accept the start time, should we detect the changes base on the `start time` unchanged?

We can detect from start time along with other properties, such as service instance name, this can be discussed in another issue or PR that focuses on event (changes detector).

> And another 2 questions
> 
> 1. Should we have a document to list all property keys?
> 2. Should we consider a value holder or some mechanism to indicate value no change? Refer https://github.com/apache/skywalking/pull/7069/files#r647909609

This PR is related to events, but is not specifically only for events, right? Those unchanged properties are also to be displayed in UI and I feel like they must be reported no matter they changed or not. I don't think we want to do something in OAP like, "if this property (`libs`) is not changed, we query the previous one and copy its value to this record" (otherwise UI could get nothing for this property), no?

wu-sheng: > if this property (libs) is not changed, we query the previous one and copy its value to this record"

This doesn't exist. If something like this, we are doing this in update, rather than query.

wu-sheng:
We have conflicts, please fix.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
kezhenxu94: I meant if you have a value holder (for example, `UNCHANGED`) to indicate the property is not changed, when the OAP receive this it cannot simply save `UNCHANGED` into that property, it should query the previous value and save the previous value into the received record, no?

wu-sheng: Take a look at `InstanceTraffic#combine`, there are some update logic existing.

wu-sheng:
@Ax1an Tests are failing.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMUtil.java
kezhenxu94: > Take a look at `InstanceTraffic#combine`, there are some update logic existing.

Nice! Then having a value holder to indicate "unchanged" looks good

Ax1an: If the change is controlled by the agent. When the specified index deletion TTL is reached, the index is deleted and jar dependencies are not updated. The jar dependencies attribute will always be empty.

Ax1an: I think we should deal with this part of the logic in OAP. If you think the current agent sends jar dependency information too frequently, we can increase the interval, such as one hour, or even longer.

Ax1an: In fact, we currently have a configuration to control this frequency.

image

wu-sheng: Make sense, it seems we have issue in ttl scenario.

wu-sheng: What is the reporting period now?

Ax1an: image

default: 30 * 10 = 300s

test/e2e/e2e-test/src/test/resources/expected/gateway/instances.yml
```diff
@@ -16,12 +16,3 @@
...
     label: not null
-    attributes:
-      - name: OS Name
-        value: not null
-      - name: hostname
-        value: not null
-      - name: Process No.
-        value: gt 0
-      - name: ipv4s
-        value: not null
```
wu-sheng, line 27: Why remove all attributes checking? This is passed, and you only add new attributes rather than deleting.

Ax1an: How to deal with the non java client without newly added attributes?

wu-sheng: These attributes are new, no all language agent reports this. Existing e2e tests have their own logic. I can't say they are correct, but directly removing clearly is wrong.

Ax1an:
@kezhenxu94 I'm a little confused about this. Can I get more detailed information?

image

image

kezhenxu94:
@Ax1an this error typically means the OAP cannot start up successfully in Docker, you can simply build the Docker image locally and `docker run` it, it should crash somehow

wu-sheng:
@kezhenxu94 The feature is good to me. Please check the codes, and especially the tests.

kezhenxu94 (commented):
@Ax1an thanks very much for the patience, I left some comments, otherwise look good to me

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/ServiceManagementClient.java
```diff
@@ -117,6 +118,7 @@ public void run() {
                                                                         .addAllProperties(OSUtil.buildOSInfo(
                                                                             Config.OsInfo.IPV4_LIST_SIZE))
                                                                         .addAllProperties(SERVICE_INSTANCE_PROPERTIES)
+                                                                        .addAllProperties(LoadedLibraryCollector.buildJVMInfo())
```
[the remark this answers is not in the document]
Ax1an: Load jar is a continuous process, it is not a static state. Maybe the result obtained later is different from that before.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/LoadedLibraryCollector.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+
+    /**
+     * Build the required JVM information to add to the instance properties
+     */
+    public static List<KeyStringValuePair> buildJVMInfo() {
+        List<KeyStringValuePair> jvmInfo = new ArrayList<>();
+        jvmInfo.add(KeyStringValuePair.newBuilder().setKey("Start Time").setValue(getVmStartTime()).build());
+        Gson gson = new GsonBuilder().disableHtmlEscaping().create();
```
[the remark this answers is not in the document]
Ax1an: Done.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/LoadedLibraryCollector.java
```diff
@@ -0,0 +1,145 @@
...
+    private final static String PATH_SEPARATOR = "/";
+    private final static String JAR_SEPARATOR = "!";
+    private static Set<ClassLoader> CURRENT_URL_CLASSLOADER_SET = new HashSet<>();
+    /**
+     * Prevent OOM in special scenes
+     */
+    private static int CURRENT_URL_CLASSLOADER_SET_MAX_SIZE = 50;
+
+    public static void registerURLClassLoader(ClassLoader classLoader) {
+        if (CURRENT_URL_CLASSLOADER_SET.size() <= CURRENT_URL_CLASSLOADER_SET_MAX_SIZE && classLoader instanceof URLClassLoader) {
```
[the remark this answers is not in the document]
Ax1an: Done.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/LoadedLibraryCollector.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.Set;
+import org.apache.skywalking.apm.agent.core.logging.api.ILog;
+import org.apache.skywalking.apm.agent.core.logging.api.LogManager;
+import org.apache.skywalking.apm.agent.core.util.CollectionUtil;
+import org.apache.skywalking.apm.network.common.v3.KeyStringValuePair;
+
+public class LoadedLibraryCollector {
+
+    private static final ILog LOGGER = LogManager.getLogger(LoadedLibraryCollector.class);
+    private final static String PATH_SEPARATOR = "/";
```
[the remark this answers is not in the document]
Ax1an: Done.

kezhenxu94 (approved):
LGTM once tests pass, thanks again @Ax1an 🙇🏻

Ax1an:
@kezhenxu94 I found that this test case has failed since the main branch was merged.

image

kezhenxu94:
> @kezhenxu94 I found that this test case has failed since the master branch was merged.
> 
> 
> 
> image

Yes never mind this case. It's not a blocker. Once all others pass we can merge this

PR #7070, opened 2021-06-05
polish debug logging avoid null value when segment be ignored.

JaredTan95:
polish debug logging avoid null value when segment be ignored.

### DIFF
--- a/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/SegmentAnalysisListener.java
+++ b/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/SegmentAnalysisListener.java
@@ -170,12 +170,15 @@ public class SegmentAnalysisListener implements FirstAnalysisListener, EntryAnal
 
     @Override
     public void build() {
-        if (log.isDebugEnabled()) {
-            log.debug("segment listener build, segment id: {}", segment.getSegmentId());
+        if (sampleStatus.equals(SAMPLE_STATUS.IGNORE)) {
+            if (log.isDebugEnabled()) {
+                log.debug("segment ignored, trace id: {}", segment.getTraceId());
+            }
+            return;
         }
 
-        if (sampleStatus.equals(SAMPLE_STATUS.IGNORE)) {
-            return;
+        if (log.isDebugEnabled()) {
+            log.debug("segment listener build, segment id: {}", segment.getSegmentId());
         }
 
         segment.setEndpointId(endpointId);
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -60,6 +60,7 @@ Release Notes.
 * Add HTTP implementation of logs reporting protocol.
 * Make metrics exporter still work even when storage layer failed.
 * Fix Jetty HTTP `TRACE` issue, disable HTTP methods except `POST`.
+* polish debug logging avoids null value when the segment ignored.
 
 #### UI
 * Add logo for kong plugin.


wu-sheng:
Please add this to changelog.

JaredTan95:
> Please add this to changelog.

done~

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/SegmentAnalysisListener.java
```diff
@@ -170,14 +170,17 @@ private void appendSearchableTags(SpanObject span) {
 
     @Override
     public void build() {
-        if (log.isDebugEnabled()) {
-            log.debug("segment listener build, segment id: {}", segment.getSegmentId());
-        }
-
         if (sampleStatus.equals(SAMPLE_STATUS.IGNORE)) {
+            if (log.isDebugEnabled()) {
+                log.debug("segment ignored, trace id: {}", segment.getTraceId());
```
wu-sheng, line 175: Does not-sampled segment have trace id?

JaredTan95: yes, there's always trace id.

wu-sheng: I need to recheck the codes, could you point out how this happens? Having trace id but no segment id.

JaredTan95: In addition to lowering the sample rate, everything else configuration can be configured with the default value in master branch.

wu-sheng: Is this caused by backend sampling? I remember all reported data should all have trace id and segment id. That is what I am not following. What the missing logic here.

JaredTan95: Yes, I forgot to say it was back-end sampling. :-P

wu-sheng: I still need to find out why this happens. Still missing why segment id lost. 😥
Or just make logs more readable.

kezhenxu94: Screen Shot 2021-06-05 at 10 18 19

This is because we defer the value of segment in `parseFirst`.

I think we should review the `parse{First,Entry}` methods and move those span-unrelated logics into `parseSegment`, such as `segment.setSegmentId(segmentObject.getTraceSegmentId()`

wu-sheng: OK, this makes sense more. If we want this log out, I am fine to merge this.

wu-sheng (approved):
Make make logs more readable. It is good to me.
@kezhenxu94 Are you good about this?

kezhenxu94 (approved):
LGTM. If @JaredTan95 you have time, please also consider https://github.com/apache/skywalking/pull/7070#discussion_r645927919

wu-sheng:
> LGTM. If @JaredTan95 you have time, please also consider https://github.com/apache/skywalking/pull/7070#discussion_r645927919

My point is, as segment has been ignored, segment is not meaningful. So outputing trace id may be better. If we want a polish, it is better to have an endpoint name or operation name of first span in log, should be more helpful.

JaredTan95:
Just offline for a while, thanks to the comments, I agreed with @wu-sheng , Make more meaningful logs, such as trace id and endpoint name, enough for debug.

wu-sheng:
> Just offline for a while, thanks to the comments, I agreed with @wu-sheng , Make more meaningful logs, such as trace id and endpoint name, enough for debug. 
> 
> 
> 
> 

Do you want to polish in this PR or later through another?

JaredTan95:
> > Just offline for a while, thanks to the comments, I agreed with @wu-sheng , Make more meaningful logs, such as trace id and endpoint name, enough for debug.
> 
> Do you want to polish in this PR or later through another?

Leave it in another PR.
 If the ignored segment is not meaningful. So, we don't need or could not get the endpoint name(exists in spans) in the segment build stage so far.  Fix it in another PR should be good.

wu-sheng (approved):

PR #7086, opened 2021-06-09
Supports modifying span attributes in async mode

wallezhang:
### Fix span attributes cannot be modified in asynchronous mode

If a span cannot determine some attributes, such as tag, when it is created, and the span needs to be finished in asynchronous mode, the attributes in the span cannot be modified after stopping the current span by invoking `org.apache.skywalking.apm.agent.core.context.ContextManager#stopSpan` method.
Therefore, the solution is to judge the `isInAsyncMode` attribute in the span when modifying the span attribute.


wallezhang:
@wu-sheng Please take a review

CHANGES.md
```diff
@@ -5,7 +5,7 @@ Release Notes.
 8.7.0
 ------------------
 #### Project
-
+* Supports modifying span attributes in async mode.
```
wu-sheng, line 8: Codes are good, this is an agent change, please fix changelog.

wallezhang: Done

wu-sheng (approved):

PR #7091, opened 2021-06-09
Use MAL to calculate JVM metrics, remove OAL dependency.

### LINKED ISSUE #6996: Use MAL to calculate JVM metrics, remove OAL dependency
More metrics analysis should be moved into MAL's responsibility.

liqiangz:
### Use MAL to calculate JVM metrics, remove OAL dependency.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6996.

image


oap-server/server-receiver-plugin/skywalking-jvm-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/jvm/provider/handler/JVMMetricReportServiceHandler.java
```diff
@@ -23,31 +23,35 @@
...
-        this.jvmSourceDispatcher = new JVMSourceDispatcher(moduleManager);
+    private final JVMSourceDispatcher jvmSourceDispatcher;
+
+    public JVMMetricReportServiceHandler(ModuleManager moduleManager, List<Rule> rules) {
+        this.jvmSourceDispatcher = new JVMSourceDispatcher(moduleManager, rules);
         this.namingControl = moduleManager.find(CoreModule.NAME)
-                                          .provider()
-                                          .getService(NamingControl.class);
+                .provider()
+                .getService(NamingControl.class);
```
kezhenxu94, line 45: Please import our code style rules, these lines are not expected to be changed

oap-server/server-bootstrap/src/main/resources/jvm-metrics-rules/jvm.yaml
```diff
@@ -0,0 +1,61 @@
...
+#    "PT20.345S" -- parses as "20.345 seconds"
+#    "PT15M"     -- parses as "15 minutes" (where a minute is 60 seconds)
+#    "PT10H"     -- parses as "10 hours" (where an hour is 3600 seconds)
+#    "P2D"       -- parses as "2 days" (where a day is 24 hours or 86400 seconds)
+#    "P2DT3H4M"  -- parses as "2 days, 3 hours and 4 minutes"
+#    "P-6H3M"    -- parses as "-6 hours and +3 minutes"
+#    "-P6H3M"    -- parses as "-6 hours and -3 minutes"
+#    "-P-6H+3M"  -- parses as "+6 hours and -3 minutes"
+# </pre>
+expSuffix: instance(['service'], ['instance'])
```
wu-sheng, line 31: Why this file doesn't be located in `/meter-analyze-config`? And default activated by `application.yml`?

wu-sheng:
I haven't seen the update of `apm.yml`, which means the UI setting is on your browser only.

liqiangz:
> I haven't seen the update of `apm.yml`, which means the UI setting is on your browser only.

The metric name is the same as before, so the apm file does not need to be updated

wu-sheng:
> before



> > I haven't seen the update of `apm.yml`, which means the UI setting is on your browser only.
> 
> @wu-sheng The metric name is the same as before, so the `apm.yml` file does not need to be updated

The part of your codes, they are, but the MAL should generate `meter_` as entity prefix, doesn't it?

wu-sheng:
You also need to remove all `ServiceInstanceJVM*` sources and related codes.

liqiangz:
> > before
> 
> > > I haven't seen the update of `apm.yml`, which means the UI setting is on your browser only.
> > 
> > 
> > @wu-sheng The metric name is the same as before, so the `apm.yml` file does not need to be updated
> 
> The part of your codes, they are, but the MAL should generate `meter_` as entity prefix, doesn't it?

@wu-sheng  I think `meterAnalyzerActiveFiles` is used for meter system, and data is received through meter-receiver-plugin. While  jvm-receiver-plugin should be like zabbix-receiver and log-receiver, so I used the same approach as them.Is there something wrong with my understanding？

wu-sheng:
> > > before
> 
> > 
> 
> > > > I haven't seen the update of `apm.yml`, which means the UI setting is on your browser only.
> 
> > > 
> 
> > > 
> 
> > > @wu-sheng The metric name is the same as before, so the `apm.yml` file does not need to be updated
> 
> > 
> 
> > The part of your codes, they are, but the MAL should generate `meter_` as entity prefix, doesn't it?
> 
> 
> 
> @wu-sheng  I think `meterAnalyzerActiveFiles` is used for meter system, and data is received through meter-receiver-plugin. While  jvm-receiver-plugin should be like zabbix-receiver and log-receiver, so I used the same approach as them.Is there something wrong with my understanding？

I don't expect any difference among these. Who controls the prefix? I need to recheck the codes.

liqiangz:
Sorry, my answer just now may be misleading.

The following is the correct.
>
> The part of your codes, they are, but the MAL should generate meter_ as entity prefix, doesn't it?
>

At present, some metircs that have been processed by mal use the meter prefix, such as zabbix-rules. Some do not, such as  k8s and log-mal. I feel that only data received through meter-receiver should use the meter prefix. This can be more discriminative.

>
>Why this file doesn't be located in /meter-analyze-config? And default activated by application.yml?
>

 MeterAnalyzerActiveFiles is used for meter system, and data is received through meter-receiver-plugin. While jvm-receiver-plugin should be like zabbix-receiver and log-receiver, so I used the same approach as them.

test/e2e/e2e-data/src/main/java/org/apache/skywalking/e2e/metrics/MetricsQuery.java
```diff
@@ -69,13 +69,31 @@
...
 
+    public static String INSTANCE_JVM_CPU = "instance_jvm_cpu";
+    public static String INSTANCE_JVM_MEMORY_HEAP = "instance_jvm_memory_heap";
+    public static String INSTANCE_JVM_MEMORY_HEAP_MAX = "instance_jvm_memory_heap_max";
+    public static String INSTANCE_JVM_MEMORY_NOHEAP = "instance_jvm_memory_noheap";
+    public static String INSTANCE_JVM_MEMORY_NOHEAP_MAX = "instance_jvm_memory_noheap_max";
+    public static String INSTANCE_JVM_YOUNG_GC_TIME = "instance_jvm_young_gc_time";
+    public static String INSTANCE_JVM_OLD_GC_TIME = "instance_jvm_old_gc_time";
+    public static String INSTANCE_JVM_YOUNG_GC_COUNT = "instance_jvm_young_gc_count";
+    public static String INSTANCE_JVM_OLD_GC_COUNT = "instance_jvm_old_gc_count";
```
wu-sheng, line 80: I think we can't check all of the metrics, because some of them have chance to be none in the e2e test env. Such as CPU and GC.

wu-sheng:
> At present, some metircs that have been processed by mal use the meter prefix, such as zabbix-rules. Some do not, such as k8s , envoy and log-mal. I feel that only data received through meter-receiver should use the meter prefix. This can be more discriminative.

OK, get it.

> MeterAnalyzerActiveFiles in application.yml is used for meter system, and data is received through meter-receiver-plugin. While jvm-receiver-plugin should be like zabbix-receiver and log-receiver, so I used the same approach as them.

I am thinking whether we need another folder(`jvm-metrics-rules`), why can't we share the `meter-analyzer-config`. Slueth is just one of meter analysis.

wu-sheng:
And please check the CI.

liqiangz:
> I am thinking whether we need another folder(jvm-metrics-rules), why can't we share the meter-analyzer-config. Slueth is just one of meter analysis.

Currently each receiver has its own folder, so I also created a folder for jvm-receiver. But what you said makes sense, I will move the file to meter-analyzer-config folder.

> And please check the CI.

I will  check the CI later.

wu-sheng (commented):
Take a reference from `receiver-otel`, you should add `enabledRules` for meter receiver. Because we have to activate meter-receiver in default

test/e2e/e2e-test/docker/base-compose.yml
```diff
@@ -99,6 +99,7 @@ services:
         -javaagent:/jacoco/jacocoagent.jar=classdumpdir=/jacoco/classes/provider,destfile=/jacoco/provider.exec,includes=org.apache.skywalking.*,excludes=org.apache.skywalking.apm.dependencies.*
         -javaagent:/skywalking/agent/skywalking-agent.jar=logging.output=CONSOLE
         -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005
+#        -XX:MaxMetaspaceSize=100m -XX:MaxDirectMemorySize=25m
```
wu-sheng, line 102: Legacy codes, you should remove.

wu-sheng:
@mrproliu slueth meter system should not be activated in default.

liqiangz:
> Take a reference from `receiver-otel`, you should add `enabledRules` for meter receiver. Because we have to activate meter-receiver in default

@wu-sheng 

Do you mean add config like this?

```yaml
receiver-otel:
  selector: ${SW_OTEL_RECEIVER:-}
  default:
    enabledHandlers: ${SW_OTEL_RECEIVER_ENABLED_HANDLERS:"oc"}
    enabledOcRules: ${SW_OTEL_RECEIVER_ENABLED_OC_RULES:"istio-controlplane"}

receiver-jvm:
  selector: ${SW_RECEIVER_JVM:default}
  default:
    enabledRules: ${SW_RECEIVER_JVM_ENABLED_RULES:"jvm"}
```

Now I have hard-coded the folder and file name
```java
 List<Rule> rules = Rules.loadRules("meter-analyzer-config", Collections.singletonList("jvm"));
 this.metricConverts = rules.stream()
                .map(it -> new MetricConvert(it, moduleManager.find(CoreModule.NAME).provider().getService(MeterSystem.class)))
```

oap-server/server-receiver-plugin/skywalking-jvm-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/jvm/provider/handler/JVMMetricReportServiceHandler.java
liqiangz: Updated, thanks.

test/e2e/e2e-test/docker/base-compose.yml
liqiangz: Updated, thanks.

wu-sheng:
All files in `meter-analyzer-config` should be controlled by `receiver-meter`.  If you are hardcoded, let's move `jvm.yaml` back to its own folder, like you used to do.

In the future, I am thinking removing the JVM metrics report grpc service permanently, and reuse the meter protocol directly.

wu-sheng:
`meter-analyzer-config` is managed by `agent-analyzer` module's `meterAnalyzerActiveFiles` config. You should rely on that. And we need to disable `spring-sleuth`, but activate `jvm.

wu-sheng:
> You also need to remove all ServiceInstanceJVM* sources and related codes.

@liqiangz I think you still miss this.

mrproliu:
> `meter-analyzer-config` is managed by `agent-analyzer` module's `meterAnalyzerActiveFiles` config. You should rely on that. And we need to disable `spring-sleuth`, but activate `jvm`. You don't need to hardcode.

Sure, I will create a PR, to disable spring-sleuth by default.

wu-sheng:
@liqiangz We still hardcoded the `jvm.yaml` as a meter rule, right?

oap-server/server-bootstrap/src/main/resources/jvm-metrics-rules/jvm.yaml
```diff
@@ -0,0 +1,61 @@
...
+metricPrefix: instance_jvm
+metricsRules:
+  - name: cpu
+    exp: jvm_cpu
+
+  - name: memory_heap
+    exp: jvm_memory.tagEqual('memory_type', 'used', 'heap_status', 'true')
```
wu-sheng, line 38: Only tagEqual no function? Or because the default function is avg?

liqiangz:
> `meter-analyzer-config` is managed by `agent-analyzer` module's `meterAnalyzerActiveFiles` config. You should rely on that. And we need to disable `spring-sleuth`, but activate `jvm`. You don't need to hardcode.

@wu-sheng  `meterAnalyzerActiveFiles`  is used for MeterReceiver. If we want to use this configuration, we need to use the MeterReceiver protocol to accept jvm data.  Now if we use jvm receiver, we can only load the rules manually, and create metricConverts by rules  to handle it.

```java
 this.metricConverts = rules.stream()
                .map(it -> new MetricConvert(it, moduleManager.find(CoreModule.NAME).provider().getService(MeterSystem.class)))
metricConverts.forEach(metricConvert -> metricConvert.toMeter(jvmSampleFamilies));
```

wu-sheng:
> > `meter-analyzer-config` is managed by `agent-analyzer` module's `meterAnalyzerActiveFiles` config. You should rely on that. And we need to disable `spring-sleuth`, but activate `jvm`. You don't need to hardcode.
> 
> @wu-sheng `meterAnalyzerActiveFiles` is used for MeterReceiver. If we want to use this configuration, we need to use the MeterReceiver protocol to accept jvm data. Now if we use jvm receiver, we can only load the rules manually, and create metricConverts by rules to handle it.
> 
> ```java
>  this.metricConverts = rules.stream()
>                 .map(it -> new MetricConvert(it, moduleManager.find(CoreModule.NAME).provider().getService(MeterSystem.class)))
>                 .collect(Collectors.toList());
> metricConverts.forEach(metricConvert -> metricConvert.toMeter(jvmSampleFamilies));
> ```

This makes sense for now. I think thinking about replacing JVM protocol by meter protocol. But this may require PMC vote on the mail list.

liqiangz:
> This makes sense for now. I think thinking about replacing JVM protocol by meter protocol. But this may require PMC vote on the mail list.

Do you mean replacing JVM protocol by meter protoco in this pr?  If PMC vote passed,  I will modify it.

wu-sheng:
> Do you mean replacing JVM protocol by meter protoco in this pr? If PMC vote passed, I will modify it.

Let's discuss it later. I don't know what PMC prefer.

test/e2e/e2e-data/src/main/java/org/apache/skywalking/e2e/metrics/MetricsQuery.java
liqiangz: Updated,  thanks

oap-server/server-bootstrap/src/main/resources/jvm-metrics-rules/jvm.yaml
liqiangz: It seem not have default function.  When there is not only one sample found according to this tag,  the result may not be so accurate.
I will add it later.

wu-sheng: I mean for downsample default is AVG. If you don't have a function, I think the MAL engine picks the first(`0`).

mrproliu:
I think @liqiangz you could extract the [ProcessService](https://github.com/apache/skywalking/blob/master/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/meter/process/IMeterProcessService.java), For now, just have a meter protocol processor. You could add one method to create another original `SampleFamily` processor. In this way, you could share the `meter-analyzer-config` folder.
@wu-sheng Replacing the JVM protocol to meter protocol, is a good way, but It will break the protocol, also be noticed, you should have a meter name prefix, such as `sw_jvm_`, don't conflict with the spring sleuth original meter name format, or another user could define names because we have the Meter SDK for the users.

wu-sheng:
> I think @liqiangz you could extract the [ProcessService](https://github.com/apache/skywalking/blob/master/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/meter/process/IMeterProcessService.java), For now, just have a meter protocol processor. You could add one method to create another original `SampleFamily` processor. In this way, you could share the `meter-analyzer-config` folder.
> @wu-sheng Replacing the JVM protocol to meter protocol, is a good way, but It will break the protocol, also be noticed, you should have a meter name prefix, such as `sw_jvm_`, don't conflict with the spring sleuth original meter name format, or another user could define names because we have the Meter SDK for the users.

For now, let't not break anything, due to no good feature provided. Replacing OAL to MAL is just for providing consistent metric analysis logic. Other part of OAL is not easy to be replaced, as those are doing metrics aggregation from details, rather than from another format of metrics.

wu-sheng:
@libinglong Make the docs ready, I think this PR should be fine after that.

oap-server/server-bootstrap/src/main/resources/jvm-metrics-rules/jvm.yaml
liqiangz: Updated,  thanks.

liqiangz:
>  also be noticed, you should have a meter name prefix, such as `sw_jvm_`, don't conflict with the spring sleuth original meter name format

@mrproliu Updated, thanks.

wu-sheng:
I don't think this is right review comment to follow.
We should keep metric name unchanged.

wu-sheng:
I think I rejected that comment.
And update the document, please.

liqiangz:
> I don't think this is right review comment to follow.
> We should keep metric name unchanged.

Metric name is not modified,  and the name of sampleFamily is modified

liqiangz:
> OK, then, still require document update.

Could you please check if there are any missing document updates.

wu-sheng (approved):
LGTM. @mrproliu If you want some polish about this and meter system. Please do after we merge this

Issue #7093, opened 2021-06-10
Is this an error example in the document?


xdRight:
**When I follow the example from document to add the OAL to the /config/oal/core.oal**
![image](https://user-images.githubusercontent.com/34833891/121459005-098b1f80-c9dd-11eb-8cab-a54f12cfa6d0.png)
```
2021-06-09 22:32:37,500 - org.apache.skywalking.oal.rt.OALRuntime - 420 [main] ERROR [] - Can't generate method doEndpointAbnormal for EndpointDispatcher.
javassist.CannotCompileException: [source error] syntax error near "rce.get*())

);

org"
	at javassist.CtNewMethod.make(CtNewMethod.java:84) ~[javassist-3.25.0-GA.jar:?]
	at javassist.CtNewMethod.make(CtNewMethod.java:50) ~[javassist-3.25.0-GA.jar:?]
	at org.apache.skywalking.oal.rt.OALRuntime.generateDispatcherClass(OALRuntime.java:418) [oal-rt-8.5.0.jar:8.5.0]
	at org.apache.skywalking.oal.rt.OALRuntime.generateClassAtRuntime(OALRuntime.java:197) [oal-rt-8.5.0.jar:8.5.0]
	at org.apache.skywalking.oal.rt.OALRuntime.start(OALRuntime.java:166) [oal-rt-8.5.0.jar:8.5.0]
	at org.apache.skywalking.oap.server.core.oal.rt.OALEngineLoaderService.load(OALEngineLoaderService.java:65) [server-core-8.5.0.jar:8.5.0]
	at org.apache.skywalking.oap.server.analyzer.provider.AnalyzerModuleProvider.start(AnalyzerModuleProvider.java:116) [agent-analyzer-8.5.0.jar:8.5.0]
	at org.apache.skywalking.oap.server.library.module.BootstrapFlow.start(BootstrapFlow.java:49) [library-module-8.5.0.jar:8.5.0]
	at org.apache.skywalking.oap.server.library.module.ModuleManager.init(ModuleManager.java:60) [library-module-8.5.0.jar:8.5.0]
	at org.apache.skywalking.oap.server.starter.OAPServerBootstrap.start(OAPServerBootstrap.java:43) [server-bootstrap-8.5.0.jar:8.5.0]
	at org.apache.skywalking.oap.server.starter.OAPServerStartUp.main(OAPServerStartUp.java:26) [server-starter-8.5.0.jar:8.5.0]
[...]

PR #7098, opened 2021-06-10
Fix OAL doc example  bug (#7093)

xdRight:
The doc example error on `master` branch
- `sum` should be modified to `count`
- the `endpointName` not in the scope of `Endpoint`, should be modified to `name`

### Fix https://github.com/apache/skywalking/issues/7093


wu-sheng (approved):
If you like, please send a pull request to other release doc branch, and you need to send a pull request to https://github.com/apache/skywalking-website/blob/master/data/docs.yml#L13 to update all commit IDs according to your other PRs.

xdRight:
👌

PR #7106, opened 2021-06-10
Disable spring sleuth analyzer by default

mrproliu:
# Ref #7091 
- Disable spring sleuth analyzer by default
- Add `agent-analyzer.meterAnalyzerActiveFiles` config description in the document.
- Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,10 +9,9 @@ Release Notes.
 
 #### Java Agent
 * Supports modifying span attributes in async mode.
-* Disable Spring sleuth meter analyzer by default.
 
 #### OAP-Backend
-
+* Disable Spring sleuth meter analyzer by default.
 
 #### UI
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ Release Notes.
 
 #### Java Agent
 * Supports modifying span attributes in async mode.
+* Disable Spring sleuth meter analyzer by default.
 
 #### OAP-Backend
 
--- a/docs/en/setup/backend/backend-receivers.md
+++ b/docs/en/setup/backend/backend-receivers.md
@@ -160,10 +160,6 @@ The receiver adds labels with `key = service` and `key = instance` to the collec
 and values from service and service instance name defined in SkyWalking Agent,
 for identification of the metric data.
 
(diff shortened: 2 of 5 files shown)


CHANGES.md
```diff
@@ -9,6 +9,7 @@ Release Notes.
 
 #### Java Agent
 * Supports modifying span attributes in async mode.
+* Disable Spring sleuth meter analyzer by default.
```
wu-sheng, line 12: This is backend change.

docs/en/setup/backend/backend-receivers.md
```diff
@@ -148,6 +148,22 @@ receiver-meter:
...
 
+To activate the meter rule files:
+```yaml
+agent-analyzer:
+  selector: ${SW_AGENT_ANALYZER:default}
+  default:
+    meterAnalyzerActiveFiles: ${SW_METER_ANALYZER_ACTIVE_FILES:} # Which files could be meter analyzed, files split by ","
+```
+
+The receiver adds labels with `key = service` and `key = instance` to the collected data samples,
```
wu-sheng, line 159: Is meter system doing this?

docs/en/setup/backend/backend-receivers.md
```diff
@@ -148,6 +148,22 @@ receiver-meter:
...
+and values from service and service instance name defined in SkyWalking Agent,
+for identification of the metric data.
+
+| Rule Name | Description | Configuration File | Data Source |
```
wu-sheng, line 163: I think this table belongs to `backend-meter.md`? As this config not belonging to the receiver.

docs/en/setup/backend/backend-receivers.md
mrproliu: The agent analyzer analyzes the protocol and adds these two keys as labels. After this, the `MAL` could build the entity from these labels.

docs/en/setup/backend/backend-receivers.md
mrproliu: Changed to the `backend-meter.md`

wu-sheng (approved):

PR #7111, opened 2021-06-11
Only count 5xx as error in Envoy ALS receiver

kezhenxu94:


wu-sheng (dismissed):

wu-sheng:
FYI to other readers, 4xx include 404 and 403, which typically are caused by the client-side issue.

wu-sheng (approved):

Issue #7113, opened 2021-06-12
TomcatInvokeInterceptor doesn't include HTTP method in endpoint name


BFergerson:
Why do you submit this issue?
- [x] Feature or performance improvement

___
### Requirement or improvement

You can use "Group Parameterized Endpoints" to collapse endpoint `/users/10` and `/users/100` but it doesn't differentiate between HTTP method types so doing a `{GET}/users/10` looks the same as `{DELETE}/users/10`.

I've created a patch that adds the HTTP method type before the URL but I'm curious if anyone knows whether the dynamic endpoint can also be determined. That's the `/users/{id}` part that's defined in `@RequestMapping`. My guess is you can't since Tomcat sits a layer above SpringMVC. If that's the case, could you update an entry span's operation name? I ask because I would rather auto-detect this information than create the necessary `endpoint-name-grouping.yml` file. I like that that file exists but I think it's best to consider it as a last result. The idea would be to let Tomcat set the operation name and if a better operation name can be determined by SpringMVC, it would override the original operation name.

If neither method works, is there a way to implement something like `USE_QUALIFIED_NAME_AS_ENDPOINT_NAME` like SpringMVC, or is that again an issue because Tomcat sits above SpringMVC?

wu-sheng:
Hi, we are going to recommend post:/uri globally. Ans support OpenAPI spec to group. @wankai123 is working on this.

wu-sheng:
I would like to say, it should be in defaul as http method + uri, and reuse config globally.

BFergerson:
@wu-sheng, always so quick with the responses, thanks. I'm glad we're moving to standard. I remember bringing up that I wasn't sure which style made the most sense.

What do you think about my comments on updating the operation name though? Even using the OpenAPI spec there will be the issue of not being able to determine the dynamic endpoint part. The `{id}` part, since that's defined in the SpringMVC layer. From my understanding, what you're saying will only result in `post:/users/10` and `delete:/users/100` and I'd like to get it fully to `post:/users/{id}` and `delete:/users/{id}`.

wu-sheng:
The good part of openAPI spec, this kind of id thing is defined there. Once users import their spec files, we could generate grouping rulea accordingly.
But we need to always include http method in a consistent way,meaning remove the <>, but add : between method and uri.

BFergerson:
[...]

PR #7130, opened 2021-06-18
Feature: support endpoint name grouping by OpenAPI definitions.

wankai123:
### Support endpoint name grouping by OpenAPI definitions

Related to #7113
More information about the feature please see `docs/en/setup/backend/endpoint-grouping-rules.md`


oap-server/server-core/pom.xml
```diff
@@ -82,7 +82,11 @@
             <groupId>io.vavr</groupId>
             <artifactId>vavr</artifactId>
         </dependency>
-
+        <dependency>
+            <groupId>io.swagger.parser.v3</groupId>
```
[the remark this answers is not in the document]
wankai123: It's an omission, fixed

docs/en/setup/backend/endpoint-grouping-rules.md
```diff
@@ -1,15 +1,319 @@
...
+          type: array
+          description: List of related products
+          items:
+            $ref: "#/components/schemas/Product"
+
+
+```
+
+Here give some scenario we might use:
+1. Only set the `x-sw-service-name`, `x-sw-endpoint-name-match-rule` and `x-sw-endpoint-name-format` are default:
```
wu-sheng, line 258: I think these 3 scenarios are strange to describe.

docs/en/setup/backend/endpoint-grouping-rules.md
```diff
@@ -1,15 +1,319 @@
...
+  version: v2
+  title: Product API
+  ...
+```
+3. Set the `x-sw-service-name` , `x-sw-endpoint-name-match-rule` and `x-sw-endpoint-name-format` :
+``` yaml
+openapi: 3.0.0
+x-sw-service-name: serviceB
+x-sw-endpoint-name-match-rule: "<${METHOD}>:${PATH}"
+x-sw-endpoint-name-format: "<${METHOD}>:${PATH}"
```
wu-sheng, line 287: Move this as an example under the table should be good enough.

image

docs/en/setup/backend/endpoint-grouping-rules.md
```diff
@@ -1,15 +1,319 @@
...
+   | GET:/products | serviceB | default | default | true | GET:/products |
+   | GET:/products/123 | serviceB | default |default |  true | GET:/products{id} |
+   | GET:/products/asia/cn | serviceB | default | default | true | GET:/products/{region}/{country} |
+   | GET:/products/123/abc/efg | serviceB | default |default |  false | GET:/products/123/abc/efg | 
+   | \<GET\>:/products/123 | serviceB | default | default | false | \<GET\>:/products/123|
+   | GET:/products/123 | serviceC | default | default | false | GET:/products/123 |
+   | \<GET\>:/products/123 | serviceB | \<\${METHOD}\>:\${PATH} | \<\${METHOD}>:\${PATH} | true | \<GET\>:/products/{id} |
+   | GET:/products/123 | serviceB | default | ${PATH}:\<\${METHOD}\> | true | /products/{id}:\<GET\> |
+   | /products/123:\<GET\> | serviceB | ${PATH}:\<\${METHOD}\> | default | true | GET:/products/{id} |
+<br />
```
wu-sheng, line 307:
```suggestion

```

docs/en/setup/backend/endpoint-grouping-rules.md
```diff
@@ -1,15 +1,319 @@
...
+   |-----|-----|-----|-----|-----|-----|
+   | GET:/products | serviceB | default | default | true | GET:/products |
+   | GET:/products/123 | serviceB | default |default |  true | GET:/products{id} |
+   | GET:/products/asia/cn | serviceB | default | default | true | GET:/products/{region}/{country} |
+   | GET:/products/123/abc/efg | serviceB | default |default |  false | GET:/products/123/abc/efg | 
+   | \<GET\>:/products/123 | serviceB | default | default | false | \<GET\>:/products/123|
+   | GET:/products/123 | serviceC | default | default | false | GET:/products/123 |
+   | \<GET\>:/products/123 | serviceB | \<\${METHOD}\>:\${PATH} | \<\${METHOD}>:\${PATH} | true | \<GET\>:/products/{id} |
+   | GET:/products/123 | serviceB | default | ${PATH}:\<\${METHOD}\> | true | /products/{id}:\<GET\> |
+   | /products/123:\<GET\> | serviceB | ${PATH}:\<\${METHOD}\> | default | true | GET:/products/{id} |
```
wu-sheng, line 306: This table is better than the above 3 cases.

kezhenxu94, line 306: I'd rather put these GET:/products, etc. etc. into `` so you don't need to escape the characters, which is error prone

wu-sheng: I am wondering the point of `\` too

wankai123: to escape the markdown keyword...  but if I put them into `` ,  the style (color) would differ from others.

wu-sheng: Could you put a screenshot here to demonstrate what will be seen?

wankai123: image
would like this

wu-sheng: @kezhenxu94 From the preview, it seems better to keep the current way?

docs/en/setup/backend/endpoint-grouping-rules.md
```diff
@@ -1,15 +1,291 @@
...
+   | Incoming Endpiont | Incoming Service | x-sw-endpoint-name-match-rule | x-sw-endpoint-name-format | Matched | Grouping Result |
+   |-----|-----|-----|-----|-----|-----|
+   | GET:/products | serviceB | default | default | true | GET:/products |
+   | GET:/products/123 | serviceB | default |default |  true | GET:/products{id} |
+   | GET:/products/asia/cn | serviceB | default | default | true | GET:/products/{region}/{country} |
+   | GET:/products/123/abc/efg | serviceB | default |default |  false | GET:/products/123/abc/efg | 
+   | \<GET\>:/products/123 | serviceB | default | default | false | \<GET\>:/products/123|
+   | GET:/products/123 | serviceC | default | default | false | GET:/products/123 |
+   | \<GET\>:/products/123 | serviceB | \<\${METHOD}\>:\${PATH} | \<\${METHOD}>:\${PATH} | true | \<GET\>:/products/{id} |
+   | GET:/products/123 | serviceB | default | ${PATH}:\<\${METHOD}\> | true | /products/{id}:\<GET\> |
```
kezhenxu94, line 280: See
```suggestion
   | GET:/products/123 | serviceB | default | \${PATH}:\<\${METHOD}\> | true | /products/{id}:\<GET\> |
```

docs/en/setup/backend/endpoint-grouping-rules.md
```diff
@@ -1,15 +1,291 @@
...
+
+   | Incoming Endpiont | Incoming Service | x-sw-endpoint-name-match-rule | x-sw-endpoint-name-format | Matched | Grouping Result |
+   |-----|-----|-----|-----|-----|-----|
+   | GET:/products | serviceB | default | default | true | GET:/products |
+   | GET:/products/123 | serviceB | default |default |  true | GET:/products{id} |
+   | GET:/products/asia/cn | serviceB | default | default | true | GET:/products/{region}/{country} |
+   | GET:/products/123/abc/efg | serviceB | default |default |  false | GET:/products/123/abc/efg | 
+   | \<GET\>:/products/123 | serviceB | default | default | false | \<GET\>:/products/123|
+   | GET:/products/123 | serviceC | default | default | false | GET:/products/123 |
+   | \<GET\>:/products/123 | serviceB | \<\${METHOD}\>:\${PATH} | \<\${METHOD}>:\${PATH} | true | \<GET\>:/products/{id} |
```
kezhenxu94, line 279: See 
```suggestion
   | \<GET\>:/products/123 | serviceB | \<\${METHOD}\>:\${PATH} | \<\${METHOD}\>:\${PATH} | true | \<GET\>:/products/{id} |
```

docs/en/setup/backend/endpoint-grouping-rules.md
kezhenxu94: - `$`, `<`, and `>` has no special meanings in GitHub markdown, they may have in other extended markdown engine.

- You don't have consistent style in terms of quoting in backticks or not, for example you quote them here 
Screen Shot 2021-06-18 at 17 54 24
but not in this table

- If you choose not to quote in `` but with `\` to escape those characters, see 

- There are cases when users may look into the source codes and copy the example with those `\`, which doesn't work.

kezhenxu94 (dismissed):
LGTM once CI pass. Thanks!!!

docs/en/setup/backend/endpoint-grouping-rules.md
wankai123: make sense, I have put all path/pattern string into ``.

docs/en/setup/backend/endpoint-grouping-rules.md
```diff
@@ -1,15 +1,293 @@
...
+
+
+### How to use
+1. Add some `Specification Extensions` for SkyWalking in the OpenAPI definition documents:<br />
+   `${METHOD}` is a reserved placeholder which represents the HTTP method eg. `POST/GET...` <br />
+   `${PATH}` is a reserved placeholder which represents the path eg. `/products/{id}`.
+
+   | Extension Name | Required | Description | Default Value |
+   |-----|-----|-----|-----|
+   | x-sw-service-name | true | The service name to which these endpoints belong | |
```
wu-sheng, line 28: We should make this optional, and the default should be the folder name. The reason I am asking this is, once all things are not required, `x-sw-endpoint-name-match-rule`'s default value matched our ALS rule(@kezhenxu94 it is matched, right?), we could only ask for standard OpenAPI spec file in the right folder, nothing more.

kezhenxu94: Yes they matched

wankai123: make sense

wu-sheng: > But if the folder is nested,

What do you mean `folder nested`? I think you should require a folder as the service name, but could use `x-sw-service-name` to override.

kezhenxu94: The folder can be any number depth


```
a
 |--b
 |--|--c
 |--|--|--d
```

wankai123: A recommend practice from OAS is `A good rule of thumb is to use the natural hierarchy present in URLs to build your file structure. For example, put all routes starting with /users (like /users and /users/{id}) in the same file (think of it as a “sub-API”).`

wankai123: I think using the folder name that the file belong is enough.

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/ResourceUtils.java
```diff
@@ -58,12 +60,39 @@ public static InputStream readToStream(String fileName) throws FileNotFoundExcep
...
+            new File(url.getPath())
+                .listFiles((dir, name) -> nameSet.contains(name)), "No files in " + parentPath);
 
         if (listFiles.length == 0) {
             throw new FileNotFoundException("files not found:" + nameSet);
         }
         return listFiles;
     }
+
+    public static List<File> getPathFilesRecursive(String path) throws FileNotFoundException {
```
wu-sheng, line 73: Notice, this kind of `Recursive` could easily cause OOM. Like I commented https://github.com/apache/skywalking/pull/7130#discussion_r654422789, please make the structure solid.

docs/en/setup/backend/endpoint-grouping-rules.md
wu-sheng: Read https://github.com/apache/skywalking/pull/7130#discussion_r654422789

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/ResourceUtils.java
wu-sheng: We should have depth as a parameter, as in this OpenAPI case, we only need recursive depth as `1`, which means current folder and one more(service name folder).

docs/en/setup/backend/endpoint-grouping-rules.md
wankai123: done

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/ResourceUtils.java
wankai123: done

docs/en/setup/backend/endpoint-grouping-rules.md
```diff
@@ -1,15 +1,294 @@
...
+   | x-sw-endpoint-name-match-rule | false | The rule used to match the endpoint.| `${METHOD}:${PATH}` |
+   | x-sw-endpoint-name-format | false | The endpoint name after grouping.| `${METHOD}:${PATH}` |
+
+   These extensions are under `OpenAPI Object`. For example, the document below has a full custom config:
+
+``` yaml
+openapi: 3.0.0
+x-sw-service-name: serviceB
+x-sw-endpoint-name-match-rule: "<${METHOD}>:${PATH}"
+x-sw-endpoint-name-format: "<${METHOD}>:${PATH}"
```
wu-sheng, line 38: Let's keep these as same as default value, to avoid any confusion. 

@wankai123 We also need an issue and another PR to fix current operation name in plugins. You should fix those after this gets merged.

docs/en/setup/backend/endpoint-grouping-rules.md
```diff
@@ -1,15 +1,294 @@
...
+1. Put the OpenAPI definition documents into directory `openapi-definitions`, SkyWalking could read all documents or documents in this subDirectorys from it, you can organize these documents by yourself. Recommend using the service name as the subDirectory name then you are not necessary to set `x-sw-service-name`. For example:
+  ```
+├── openapi-definitions
+│   ├── serviceA
+│   │   ├── customerAPI-v1.yaml
+│   │   └── productAPI-v1.yaml
+│   └── serviceB
+│       └── productAPI-v2.yaml
+```
+3. Turn the feature on by setting the `Core Module` configuration `${SW_CORE_ENABLE_ENDPOINT_NAME_GROUPING_BY_OPAENAPI:true}`
```
wu-sheng, line 59: As no side effect, keeping this open in default should be better?

docs/en/setup/backend/endpoint-grouping-rules.md
```diff
@@ -1,15 +1,294 @@
...
+
+
+### How to use
+1. Add some `Specification Extensions` for SkyWalking config in the OpenAPI definition documents, otherwise, all configs are default:<br />
+   `${METHOD}` is a reserved placeholder which represents the HTTP method eg. `POST/GET...` <br />
+   `${PATH}` is a reserved placeholder which represents the path eg. `/products/{id}`.
+
+   | Extension Name | Required | Description | Default Value |
+   |-----|-----|-----|-----|
+   | x-sw-service-name | false | The service name to which these endpoints belong | The directory name which the definition documents belong to|
```
wu-sheng, line 28: What happens in no `x-sw-service-name` and `yaml` file is located in the `openapi-definitions` directory directly.

docs/en/setup/backend/endpoint-grouping-rules.md
```diff
@@ -1,15 +1,294 @@
...
+  description: OpenAPI definition for SkyWalking test.
+  version: v2
+  title: Product API
+  ...
+```
+
+   We highly recommend using the default config, the custom config (`x-sw-endpoint-name-match-rule/x-sw-endpoint-name-format`) would be considered as part of the match rules (regex pattern).
+   We provide some cases in `org.apache.skywalking.oap.server.core.config.group.openapi.EndpointGroupingRuleReader4OpenapiTest`, you could validate your custom config as well.
+
+1. Put the OpenAPI definition documents into directory `openapi-definitions`, SkyWalking could read all documents or documents in this subDirectorys from it, you can organize these documents by yourself. Recommend using the service name as the subDirectory name then you are not necessary to set `x-sw-service-name`. For example:
```
wu-sheng, line 50:
```suggestion
1. All OpenAPI definition documents should be located in the `openapi-definitions` directory, with at most two levels directories. Recommend using the service name as the subDirectory name then you are not necessary to set `x-sw-service-name`. For example:
```

docs/en/setup/backend/endpoint-grouping-rules.md
wankai123: `openapi-definitions` would be the service name, should throw an exception?

docs/en/setup/backend/endpoint-grouping-rules.md
```diff
@@ -1,15 +1,294 @@
...
 
+If the incoming endpoint name hit the rules, SkyWalking will grouping the endpoint by rules.
+
+SkyWalking provides 2 ways to support endpoint grouping:
+1. Endpoint name grouping by OpenAPI definitions.
+2. Endpoint name grouping by custom configuration.
+
+The 2 grouping features can work together in sequence.
+## Endpoint name grouping by OpenAPI definitions
+The OpenAPI definitions are the documents based on The [OpenAPI Specification (OAS)](https://github.com/OAI/OpenAPI-Specification) which used to define a standard, language-agnostic interface for HTTP APIs.
```
wu-sheng, line 17:
```suggestion
The OpenAPI definitions are the documents based on The [OpenAPI Specification (OAS)](https://www.openapis.org/) which used to define a standard, language-agnostic interface for HTTP APIs.
```

Once we say a spec, we should link to their website rather that GitHub repo. GitHub is for developers.

wu-sheng (dismissed):
It is good for me. Thanks.

oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/config/group/openapi/EndpointGroupingBenchmark4Openapi.java
```diff
@@ -0,0 +1,172 @@
...
+
+        @SneakyThrows
+        public FormatClassPaths20() {
+            rule = new EndpointGroupingRule4Openapi();
+            for (int i = 0; i <= 3; i++) {
+                rule.addGroupedRule("serviceA", "GET:/products1/{id}/" + +i, "GET:/products1/([^/]+)/" + i);
+                rule.addGroupedRule("serviceA", "POST:/products1/{id}/" + +i, "POST:/products1/([^/]+)/" + i);
+                rule.addGroupedRule("serviceA", "GET:/products2/{id}/" + +i, "GET:/products2/([^/]+)/" + i);
+                rule.addGroupedRule("serviceA", "POST:/products3/{id}/" + +i, "POST:/products3/([^/]+)/" + i);
+                rule.addGroupedRule("serviceA", "GET:/products3/{id}/" + +i, "GET:/products3/([^/]+)/" + i);
```
kezhenxu94, line 51: Not important though, `"GET:/products1/{id}/" + +i` looks tricky, why not just `GET:/products1/{id}/" + i`?

oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/config/group/openapi/EndpointGroupingBenchmark4Openapi.java
```diff
@@ -0,0 +1,172 @@
...
+
+        @SneakyThrows
+        public FormatClassPaths50() {
+            rule = new EndpointGroupingRule4Openapi();
+            for (int i = 0; i <= 9; i++) {
+                rule.addGroupedRule("serviceA", "GET:/products1/{id}/" + +i, "GET:/products1/([^/]+)/" + i);
+                rule.addGroupedRule("serviceA", "POST:/products1/{id}/" + +i, "POST:/products1/([^/]+)/" + i);
+                rule.addGroupedRule("serviceA", "GET:/products2/{id}/" + +i, "GET:/products2/([^/]+)/" + i);
+                rule.addGroupedRule("serviceA", "POST:/products3/{id}/" + +i, "POST:/products3/([^/]+)/" + i);
+                rule.addGroupedRule("serviceA", "GET:/products3/{id}/" + +i, "GET:/products3/([^/]+)/" + i);
```
kezhenxu94, line 72: Same here

oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/config/group/openapi/EndpointGroupingBenchmark4Openapi.java
```diff
@@ -0,0 +1,172 @@
...
+
+        @SneakyThrows
+        public FormatClassPaths200() {
+            rule = new EndpointGroupingRule4Openapi();
+            for (int i = 0; i <= 39; i++) {
+                rule.addGroupedRule("serviceA", "GET:/products1/{id}/" + +i, "GET:/products1/([^/]+)/" + i);
+                rule.addGroupedRule("serviceA", "POST:/products1/{id}/" + +i, "POST:/products1/([^/]+)/" + i);
+                rule.addGroupedRule("serviceA", "GET:/products2/{id}/" + +i, "GET:/products2/([^/]+)/" + i);
+                rule.addGroupedRule("serviceA", "POST:/products3/{id}/" + +i, "POST:/products3/([^/]+)/" + i);
+                rule.addGroupedRule("serviceA", "GET:/products3/{id}/" + +i, "GET:/products3/([^/]+)/" + i);
```
kezhenxu94, line 93: And here

kezhenxu94 (dismissed):

oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/config/group/openapi/EndpointGroupingBenchmark4Openapi.java
wu-sheng: Haha, interesting, it is better to fix before merge.

wankai123: thanks, @kezhenxu94, I tested `+i` but forgot to remove them.

wu-sheng (approved):

Issue #6664, opened 2021-04-01
Skywalking8.4 cannot upload log after authentication


gtahubb:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [x] Question or discussion
- [ ] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
Skywalking8.4. After setting the authentication, the server log cannot be uploaded to OAP, but the service information can be monitored. After closing the authentication information, the log can be uploaded. Error message of agent log: error 2021-04-01 12:22:46:414 grpc-default-executor-0 grpclogreportserviceclient: try to send 2 log data to collector, with unexpected exception
org.apache.skywalking . apm.dependencies.io . grpc.StatusRuntimeException : PERMISSION_ DENIED

agent:
agent.authentication = ${SW_AGENT_AUTHENTICATION:xxxxxxxx}

plugin.toolkit.log.grpc.reporter.server_host=${SW_GRPC_LOG_SERVER_HOST:192.168.8.121}
plugin.toolkit.log.grpc.reporter.server_port=${SW_GRPC_LOG_SERVER_PORT:11800}
plugin.toolkit.log.grpc.reporter.max_message_size=${SW_GRPC_LOG_MAX_MESSAGE_SIZE:10485760}
plugin.toolkit.log.grpc.reporter.upstream_timeout=${SW_GRPC_LOG_GRPC_UPSTREAM_TIMEOUT:30}

oap:
gRPCHost: ${SW_RECEIVER_GRPC_HOST:0.0.0.0}
gRPCPort: ${SW_RECEIVER_GRPC_PORT:0}
maxConcurrentCallsPerConnection: ${SW_RECEIVER_GRPC_MAX_CONCURRENT_CALL:0}
maxMessageSize: ${SW_RECEIVER_GRPC_MAX_MESSAGE_SIZE:0}
gRPCThreadPoolQueueSize: ${SW_RECEIVER_GRPC_POOL_QUEUE_SIZE:-1}
gRPCThreadPoolSize: ${SW_RECEIVER_GRPC_THREAD_POOL_SIZE:-1}
gRPCSslEnabled: ${SW_RECEIVER_GRPC_SSL_ENABLED:false}
gRPCSslKeyPath: ${SW_RECEIVER_GRPC_SSL_KEY_PATH:""}
gRPCSslCertChainPath: ${SW_RECEIVER_GRPC_SSL_CERT_CHAIN_PATH:""}
[...]

Issue #7040, opened 2021-05-30
Does GRPCLogReportServiceClient support TLS?


BFergerson:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [x] Question or discussion

___
### Question

I'm able to send traces via TLS but when I try to send logs, I get the following exception:
```
ERROR 2021-05-30 22:21:59:387 grpc-default-executor-1 GRPCLogReportServiceClient : Try to send 1 log data to collector, with unexpected exception.
org.apache.skywalking.apm.dependencies.io.grpc.StatusRuntimeException: UNAVAILABLE: Network closed for unknown reason
        at org.apache.skywalking.apm.dependencies.io.grpc.Status.asRuntimeException(Status.java:533)
        at org.apache.skywalking.apm.dependencies.io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:478)
        at org.apache.skywalking.apm.dependencies.io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:413)
        at org.apache.skywalking.apm.dependencies.io.grpc.internal.ClientCallImpl.access$500(ClientCallImpl.java:66)
        at org.apache.skywalking.apm.dependencies.io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInternal(ClientCallImpl.java:742)
        at org.apache.skywalking.apm.dependencies.io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInContext(ClientCallImpl.java:721)
        at org.apache.skywalking.apm.dependencies.io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
        at org.apache.skywalking.apm.dependencies.io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:123)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
        at java.base/java.lang.Thread.run(Unknown Source)
```

I've tried to debug this issue and ran into:
[...]

PR #7131, opened 2021-06-18
Support authentication for log report channel.

zhyyu:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes [#6664 ](https://github.com/apache/skywalking/issues/6664)

The log channel use an independent gRPC channel, and it's not decorated by authentication, I decorate it. So agent can upload log when OAP backend open authentication.

It's my first time to open a pull request. Maybe not very standard,  I will appreciate a lot for advices!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ Release Notes.
 #### Java Agent
 * Supports modifying span attributes in async mode.
 * Agent supports the collection of JVM arguments and jar dependency information.
+* [Temporary] Support authentication for log report channel. This feature and grpc channel is going to be removed after Satellite 0.2.0 release.
 
 #### OAP-Backend
 * Disable Spring sleuth meter analyzer by default.
--- a/apm-sniffer/apm-toolkit-activation/apm-toolkit-logging-common/src/main/java/org/apache/skywalking/apm/toolkit/logging/common/log/GRPCLogReportServiceClient.java
+++ b/apm-sniffer/apm-toolkit-activation/apm-toolkit-logging-common/src/main/java/org/apache/skywalking/apm/toolkit/logging/common/log/GRPCLogReportServiceClient.java
@@ -18,8 +18,16 @@
 
 package org.apache.skywalking.apm.toolkit.logging.common.log;
 
+import io.grpc.CallOptions;
+import io.grpc.Channel;
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.ClientInterceptors;
+import io.grpc.ForwardingClientCall;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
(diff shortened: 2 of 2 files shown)


wu-sheng:
I need to to remind you, this log grpc channel is going to be removed once Satellite project is ready.

zhyyu:
> I need to to remind you, this log grpc channel is going to be removed once Satellite project is ready.

Thanks for remind

> I need to to remind you, this log grpc channel is going to be removed once Satellite project is ready.

Thanks for remind, but before Satellite is ready, our company will still use the SkyWalking project.

wu-sheng:
> > I need to to remind you, this log grpc channel is going to be removed once Satellite project is ready.
> 
> Thanks for remind, but before Satellite is ready, our company will still use the SkyWalking project.

8.7.0 is far from release, so, even we merge this, we wouldn't guarantee this included in the next release. That is the point of my reminder.

wu-sheng (approved):
Temporarily accept. This feature and grpc channel is going to be removed after Satellite 0.2.0 release. cc @EvanLjp @mrproliu

kezhenxu94:
> Temporarily accept. This feature and grpc channel is going to be removed after Satellite 0.2.0 release. cc @EvanLjp @mrproliu 

Do we really want "all in sidecar"? If users don't use Kubernetes etc., they have to deploy another process (satellite) with their application, just for log authentication to work in this case?

wu-sheng:
> > Temporarily accept. This feature and grpc channel is going to be removed after Satellite 0.2.0 release. cc @EvanLjp @mrproliu
> 
> Do we really want "all in sidecar"? If users don't use Kubernetes etc., they have to deploy another process (satellite) with their application, just for log authentication to work in this case?

That is the point. We need a consistent channel, but the current Satellite is not supporting all transport protocols, which is why @EvanLjp set up a new grpc channel for log reporter.

BFergerson:
https://github.com/apache/skywalking/issues/7040 was marked as a duplicate of https://github.com/apache/skywalking/issues/6664 and https://github.com/apache/skywalking/issues/6664 is closed by this. I would think that means this should also fix the issue of the lack of TLS encryption mentioned in https://github.com/apache/skywalking/issues/7040 but this PR doesn't seem to do that.

Also, I'd like to mention that I find it a bit odd that the progress of the primary SkyWalking project is being slowed down by secondary projects like the sidecar. I would think that the primary project sets the pace and the secondary projects would need to keep up to stay functional. Here it seems that there is progress that could be made with SkyWalking but due to issues with a secondary project, that progress is being put on hold.

I believe you know what's best though so feel free to disregard my opinion.

wu-sheng:
@BFergerson I reopened the #7040 as TLS has not been implemented. And if you need this feature, feel free to submit a similar feature like this. 

The satellite project is not moving fast, I guess due to sidecar matters for a large-scale system.

PR #7136, opened 2021-06-20
Extract dependency management to a bom

kezhenxu94:


oap-server/oal-grammar/pom.xml
```diff
@@ -27,6 +27,10 @@
 
     <artifactId>oal-grammar</artifactId>
 
+    <properties>
+        <antlr.version>4.7.1</antlr.version>
+    </properties>
+
```
wu-sheng, line 33: Why this version is managed in specific project?

kezhenxu94: > Why this version is managed in specific project?

This is plugin version, plugin dependency cannot be managed by bom

wu-sheng (approved):

PR #7138, opened 2021-06-20
Add FluentBit integration E2E test and doc

kezhenxu94:


kezhenxu94:
FYI @hailin0

wu-sheng (approved):
FYI @Shikugawa, @kezhenxu94 found a way to use fluentbit directly by using LUA script.

Once e2e passed, it works for me.

Shikugawa:
@wu-sheng Using lua is not enough for the real use-case. It is because it can't modify request headers...

wu-sheng:
> @wu-sheng I'm not sure about how much we affected by performance issue by using lua filter. But it seems enough to have.

Native one is always better to have and have better performance, AFAIK. I just want to inform you, before yours gets merged, we still have an option.

PR #7139, opened 2021-06-20
Add JDK 16 to test matrix, move bom to backend profile

kezhenxu94:


wu-sheng:
I remember there is a legacy issue about this?

kezhenxu94:
> I remember there is a legacy issue about this?

Really? I don't remember that

wu-sheng (approved):

PR #7145, opened 2021-06-21
eliminate `now` error at a fixed rate schedule executor

dmsolr:

to eliminate `System.currentTimeMillis()` error at a fixed rate schedule executor. That will cause `CounterWindow` cannot to remove the element as expected.

### DIFF
--- a/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/counter/CounterWindow.java
+++ b/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/counter/CounterWindow.java
@@ -53,13 +53,20 @@ public class CounterWindow {
         Queue<Tuple2<Long, Double>> window = windows.get(id);
         window.offer(Tuple.of(now, value));
 
-        long waterLevel = now - windowSize + 60;
+        long waterLevel = now - windowSize;
         Tuple2<Long, Double> peek = window.peek();
+        if (peek._1 > waterLevel) {
+            return peek;
+        }
         Tuple2<Long, Double> result = peek;
-        while (peek._1 < waterLevel) {
+        while (peek._1 <= waterLevel) {
             result = window.poll();
             peek = window.element();
         }
+        if (waterLevel - result._1 > peek._1 - waterLevel) {
+            return peek;
+        }
+
         return result;
     }
 
--- a/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/counter/CounterWindow.java
+++ b/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/counter/CounterWindow.java
@@ -42,6 +42,7 @@ public class CounterWindow {
 
     public static final CounterWindow INSTANCE = new CounterWindow();
 
+    private final Map<ID, Tuple2<Long, Double>> lastElementMap = Maps.newHashMap();
     private final Map<ID, Queue<Tuple2<Long, Double>>> windows = Maps.newHashMap();
 
     public Tuple2<Long, Double> increase(String name, ImmutableMap<String, String> labels, Double value, long windowSize, long now) {
@@ -64,11 +65,23 @@ public class CounterWindow {
             peek = window.element();
         }
 
-        if (window.size() < 2 || waterLevel - result._1 <= peek._1 - waterLevel) {
+        if (waterLevel - result._1 <= peek._1 - waterLevel) {
             return result;
         }
 
         return peek;
+    }
+
+    public Tuple2<Long, Double> pop(String name, ImmutableMap<String, String> labels, Double value, long now) {
+        ID id = new ID(name, labels);
+
+        Tuple2<Long, Double> element = Tuple.of(now, value);
+        Tuple2<Long, Double> result = lastElementMap.get(id);
+        lastElementMap.put(id, element);
+        if (result == null) {
+            return element;
+        }
+        return result;
     }
 
     public void reset() {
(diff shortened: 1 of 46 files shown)


hanahmily:
Would you debug some issues? Please elaborate what's these changes' purpose. It looks like this should not be merged until finding the root cause.

wu-sheng:
@dmsolr Could you share an error about `System.currentTimeMillis()`? If this could go wrong, then the agent has a higher risk than backend. Also, we never received a report about this

dmsolr:
Sorry for maybe I use the improper word about `error`.
The period of calling `System.currentTimeMillis()` twice may not equal to `period of fetcher`.

for example, the period of fetching = 10s, we can get the time list as follows,
1624265391580
1624265401585
1624265411581
1624265421582

I set the duration as `PT10S`, and then CounterWindow needs to remove the `1624265401585` after `1624265411581`.
But it is not removed because of `1624265411581 - 1624265401585 = 9996 < 10000`.

wu-sheng:
> I set the duration as PT10S, and then CounterWindow needs to remove the 1624265401585 after 1624265411581.
But it is not removed because of 1624265411581 - 1624265401585 = 9996 < 10000.

Wouldn't remove that in the next iteration?

dmsolr:
It is sure will be removed in the next iteration. But it will lead to aggregation computing improperly and an element(1624265411581) does not remove in this iteration.

https://github.com/apache/skywalking/blob/09944e00a95a61b2a091087ccb49f8b6a66f0604/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/counter/CounterWindow.java#L47-L58

wu-sheng:
>  But it will lead to aggregation computing improperly and an element(1624265411581) does not remove in this iteration.

I can't see this as a removal issue. If you define a window of 5 slots(last 5 minutes), even there are 6 slots kept in short term, there is no point `rate` function that has an issue. If there is an issue, it is a rate function error, rather than a removal bug.

wu-sheng:
@hanahmily An update, this issue is caused by currently, `increase` and `rate` are peeking the first and last of the windows to calculate the value. If the oldest data can't be removed in time, then the result is wrong. 

I have asked @dmsolr to fix in the way, that always use the duration to calculate the number of slot for rate/increase, and don't always peek data for the last element, but `latest` - `N slot`(from rate/increase function parameter)

hanahmily (approved):
> @hanahmily An update, this issue is caused by currently, `increase` and `rate` are peeking the first and last of the windows to calculate the value. If the oldest data can't be removed in time, then the result is wrong.
> 
> I have asked @dmsolr to fix in the way, that always use the duration to calculate the number of slot for rate/increase, and don't always peek data for the last element, but `latest` - `N slot`(from rate/increase function parameter)

Got > @hanahmily An update, this issue is caused by currently, `increase` and `rate` are peeking the first and last of the windows to calculate the value. If the oldest data can't be removed in time, then the result is wrong.
> 
> I have asked @dmsolr to fix in the way, that always use the duration to calculate the number of slot for rate/increase, and don't always peek data for the last element, but `latest` - `N slot`(from rate/increase function parameter)

Thanks for your elaboration, got the point. These changes look great to me.

wu-sheng (changes requested):
@hanahmily What I asked, has not been done. This PR is manipulate time, which is not recommended. Let's wait for slot base mathematics.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/counter/CounterWindow.java
```diff
@@ -47,15 +47,20 @@
...
+            windows.put(id, new PriorityQueue<>());
         }
+
         Queue<Tuple2<Long, Double>> window = windows.get(id);
         window.offer(Tuple.of(now, value));
-        Tuple2<Long, Double> ps = window.element();
-        if ((now - ps._1) >= windowSize) {
-            window.remove();
+
+        long waterLevel = now - windowSize + 60;
```
wu-sheng, line 56: What does `60` mean?

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/counter/CounterWindow.java
```diff
@@ -47,15 +47,27 @@
...
+        long waterLevel = now - windowSize;
+        Tuple2<Long, Double> peek = window.peek();
+        if (peek._1 > waterLevel) {
+            return peek;
+        }
+        Tuple2<Long, Double> result = peek;
+        while (peek._1 <= waterLevel) {
+            result = window.poll();
+            peek = window.element();
         }
```
wu-sheng, line 65: This logic seems correct, have you tried `Collections.binarySearch()`? `window` is an ordered list, right? If so, a binary search would be much faster than a sequential search。

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/counter/CounterWindow.java
```diff
@@ -42,20 +42,46 @@
...
+            return peek;
+        }
+
+        Tuple2<Long, Double> result = peek;
+        while (peek._1 < waterLevel) {
+            result = window.poll();
+            peek = window.element();
+        }
+
+        if (waterLevel - result._1 <= peek._1 - waterLevel) {
```
wu-sheng, line 68:
```suggestion
        // Choose the closed slot to the expected timestamp
        if (waterLevel - result._1 <= peek._1 - waterLevel) {
```

wu-sheng (approved):
LGTM

hanahmily (approved):
LGTM good fix!

PR #7153, opened 2021-06-22
Support prepare and save metrics concurrency

### LINKED ISSUE #7120: Concurrent create PrepareRequest when persist Metrics

- [ * ] Feature or performance improvement

Skywalking v8.4.0
Elasticsearch 7.8.1

___

I watched the Grafana and found persist to ES is too slow that costs 600s. 

51567F30EFE0EBE075423524701AE310
image

I found that there are 2 steps in the PersistenceTimer.extractDataAndSave 

Step 1: Prepare requests which searching ES for merge Metrics. (Need Concurrent)
Step 2: Persist prepare request that we made in `Step 1` to ES. (Already fixed in https://github.com/apache/skywalking/issues/6472)

I fix in my code.  This is the result. 

image

```
List<Future<List<PrepareRequest>>> futures = new ArrayList<>(persistenceWorkers.size());
persistenceWorkers.forEach(worker -> {
    futures.add(prepareExecutorService.submit(() -> {
        if (log.isDebugEnabled()) {
            log.debug("extract {} worker data and save", worker.getClass().getName());
        }
        List<PrepareRequest> innerPrepareRequests = new ArrayList<>(5000);

        worker.buildBatchRequests(innerPrepareRequests);

        worker.endOfRound(System.currentTimeMillis() - lastTime);
        return innerPrepareRequests;
    }));

});
for (Future<List<PrepareRequest>> future : futures) {
    List<PrepareRequest> requests = future.get();
    if(CollectionUtils.isNotEmpty(requests)){
        prepareRequests.addAll(requests);
    }
}
```

Can we change it to concurrent when we prepare the request?

a1vin-tian:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7120 


wu-sheng:
You should update the changelog.

oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimerTest.java
```diff
@@ -0,0 +1,94 @@
...
+            MetricsStreamProcessor.getInstance().getPersistentWorkers().add(genWorkers(i, count));
+        }
+        ModuleManager moduleManager = mock(ModuleManager.class);
+        ModuleServiceHolder moduleServiceHolder = mock(ModuleServiceHolder.class);
+        doReturn((ModuleProviderHolder) () -> moduleServiceHolder).when(moduleManager).find(anyString());
+        doReturn(new MetricsCreatorNoop()).when(moduleServiceHolder).getService(MetricsCreator.class);
+        doReturn(iBatchDAO).when(moduleServiceHolder).getService(IBatchDAO.class);
+
+        PersistenceTimer.INSTANCE.start(moduleManager, moduleConfig);
+        Thread.sleep(7000);
```
wu-sheng, line 77: Using `sleep` in the test case would be unstable in the CI process.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -63,98 +70,149 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+                        }
+                    }
 
-                if (debug) {
-                    log.info("build batch persistence duration: {} ms", System.currentTimeMillis() - startTime);
-                }
-            } finally {
-                timer.finish();
-            }
+                    synchronized (prepareRequests) {
```
wu-sheng, line 156: Where there are 2 `synchronized (prepareRequests)` statements?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -63,98 +70,149 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+                        while (this.prepareRequests.size() < maxSyncoperationNum && !prepareDone) {
```
wu-sheng, line 157: If there is a resource available, why don't run persistence when you have `prepareRequests`?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -63,98 +70,149 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
-                CountDownLatch countDownLatch = new CountDownLatch(partitions.size());
-                for (final List<PrepareRequest> partition : partitions) {
-                    executorService.submit(() -> {
+                        if (CollectionUtils.isEmpty(prepareRequests)) {
+                            continue;
+                        }
+                    }
+
+                    List<PrepareRequest> partition = null;
+                    synchronized (prepareRequests) {
```
wu-sheng, line 170: 3rdt time lock.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -63,98 +70,149 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
-                if (debug) {
-                    log.info("build batch persistence duration: {} ms", System.currentTimeMillis() - startTime);
-                }
-            } finally {
-                timer.finish();
-            }
+                    synchronized (prepareRequests) {
+                        while (this.prepareRequests.size() < maxSyncoperationNum && !prepareDone) {
+                            try {
+                                this.prepareRequests.wait(1000);
```
wu-sheng, line 159: This is a lock wait. Are you going to block globally?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -63,98 +70,149 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+                        if (CollectionUtils.isEmpty(prepareRequests)) {
+                            continue;
+                        }
+                    }
+
+                    List<PrepareRequest> partition = null;
+                    synchronized (prepareRequests) {
+                        List<PrepareRequest> prepareRequestList = this.prepareRequests.subList(0, Math.min(maxSyncoperationNum, this.prepareRequests.size()));
+                        partition = new ArrayList<>(prepareRequestList);
+                        prepareRequestList.clear();
```
wu-sheng, line 173: prepareRequestList is created in L171, and only a reference, why should be clear?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
a1vin-tian: Just distinguish logical code blocks.  Those can merge.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -63,98 +70,149 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+                    }
+                    List<PrepareRequest> finalPartition = partition;
```
wu-sheng, line 175: Why need `finalPartition`? It seems as same as `partition` always.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
wu-sheng: I think we should avoid lock as much as possible. And please use a reentrance lock, because as low concurrency case here, it is better to use a real lock.

wu-sheng:
And CIs fail, please recheck.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -63,98 +70,150 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
-            MetricsTag.EMPTY_KEY, MetricsTag.EMPTY_VALUE
+                "persistence_timer_bulk_prepare_latency", "Latency of the prepare stage in persistence timer",
+                MetricsTag.EMPTY_KEY, MetricsTag.EMPTY_VALUE
         );
         executeLatency = metricsCreator.createHistogramMetric(
-            "persistence_timer_bulk_execute_latency", "Latency of the execute stage in persistence timer",
-            MetricsTag.EMPTY_KEY, MetricsTag.EMPTY_VALUE
+                "persistence_timer_bulk_execute_latency", "Latency of the execute stage in persistence timer",
+                MetricsTag.EMPTY_KEY, MetricsTag.EMPTY_VALUE
+        );
```
kezhenxu94, line 86: Please import our code style settings to avoid these unnecessary changes

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
a1vin-tian: If we execute a very small number at a time, it may cause the previous threads to perform a small number of writes and each thread waits for 10 seconds in the end (wait for the refresh. ). I think the time to fill the prepareRequests faster than wait 10 sec.

wu-sheng: It may execute a small set in the first time, but wouldn't be larger in the next iteration? Because you have very limited threads. If you just wait, and nothing happens. From the whole picture, I think we cost more time?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -63,98 +70,150 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
         if (log.isDebugEnabled()) {
             log.debug("Extract data and save");
         }
 
         long startTime = System.currentTimeMillis();
+        HistogramMetrics.Timer allTimer = allLatency.createTimer();
 
         try {
-            HistogramMetrics.Timer timer = prepareLatency.createTimer();
+            List<PersistenceWorker> persistenceWorkers = new ArrayList<>();
```
kezhenxu94, line 119:
```suggestion
            List<PersistenceWorker<? extends StorageData>> persistenceWorkers = new ArrayList<>();
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -63,98 +70,150 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+            persistenceWorkers.addAll(TopNStreamProcessor.getInstance().getPersistentWorkers());
+            persistenceWorkers.addAll(MetricsStreamProcessor.getInstance().getPersistentWorkers());
+            CountDownLatch countDownLatch = new CountDownLatch(MetricsStreamProcessor.getInstance().getPersistentWorkers().size());
```
kezhenxu94, line 122: I believe it should be

```suggestion
            CountDownLatch countDownLatch = new CountDownLatch(persistenceWorkers.size());
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
kezhenxu94: > Why need `finalPartition`? It seems as same as `partition` always.

You need to make it `final` when passing it into a lambda

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
a1vin-tian: PrepareRequestList clear will remove range from `this.prepareRequests`. So we just need to take between 0 and maxSyncoperationNum from `this.prepareRequests` until `this.prepareRequests` is empty. 


```
this.prepareRequests = [0,1,2,3,4,5,6,7,8,9,10];

this.prepareRequests.subList(0,5).clear();

// after this.prepareRequests = [6,7,8,9,10];

```

kezhenxu94: > prepareRequestList is created in L171, and only a reference, why should be clear?

I think we don't even need `prepareRequestList`, just 

```suggestion
                        partition = this.prepareRequests.subList(0, Math.min(maxSyncoperationNum, this.prepareRequests.size()));
```

because `partition` is not meant to be modified in `batchDAO.synchronous`, no?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
a1vin-tian: Yes. This will wait for all metrics to be refreshed.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
wu-sheng: My point is the same. `partition` is enough. `#clear` is only relating to GC, but in Java, we don't need to do that.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
wu-sheng: Then let's keep a clear final.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
a1vin-tian: Could you tell me where it is?  I can't found it. I use `IntelliJ IDEA`.

wu-sheng: https://github.com/apache/skywalking/blob/master/codeStyle.xml

kezhenxu94: > Could you tell me where it is? I can't found it. I use `IntelliJ IDEA`.

At the root directory of the project https://github.com/apache/skywalking/blob/master/codeStyle.xml

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
a1vin-tian: How about this ?
```
final List<PrepareRequest> finalPartition = partition;
```

kezhenxu94: > How about this ?
> 
> ```
> final List<PrepareRequest> finalPartition = partition;
> ```

Yes, that's what we mean

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
a1vin-tian: If we did not clear the sublist cause always get the same range from `this.prepareRequests` 
https://github.com/apache/skywalking/pull/7153#discussion_r656046059

kezhenxu94: > If we did not clear the sublist cause always get the same range from `this.prepareRequests`
> [#7153 (comment)](https://github.com/apache/skywalking/pull/7153#discussion_r656046059)

Then this makes sense to me. 👍🏻

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,69 +105,111 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
         } finally {
             if (log.isDebugEnabled()) {
                 log.debug("Persistence data save finish");
             }
-
+            allTimer.finish();
             prepareRequests.clear();
+
             lastTime = System.currentTimeMillis();
+            prepareDone = false;
```
wu-sheng, line 212: Why need this?

wu-sheng:
Hi @a1vin-tian, generally the codes are good. Could you add some comments in the codes for helping readers in the future?
It is easier to read currently, but more contributors are going to read this. Such as `prepareDone`, `countDownLatch`, and about many `wait` and `while`.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
a1vin-tian: `prepareDone` means prepare stage is done. 
So we can't batch metrics to storage do not wait for metrics enoughly. And while loop for `batchDAO.synchronous` can break.
We can use local variables instead.

a1vin-tian:
1. The worst case of this change is that the executed `prepare requests` concurrently and then call `batchDAO.synchronous.` But it is also faster than before because prepare requests are executed concurrently.
            2. The reason for accumulating a batch of requests is because of the following reasons:
                1. There will be no small fragments requesting to write to the storage, reducing the pressure of the storage and improving the write performance
                2. In ES, call each `batchDAO.synchronous` takes at least 10 seconds to cause even if one piece of data is submitted, what will wait 10 seconds. Therefore, it is important to increase the number of each request and reduce the number of requests, Can reduce all time.
            
            For example: 
            At `batchDAO.synchronous` stage. 
            `maxSyncoperationNum` = 50000

            Suppose there are 210100 requests.
                    waited:
                    thread 1: [50000,50000,5050] 3time
                    thread 2: [50000,50000,5050] 3time
            
                    not waited:
                    thread 1: [1000,50000,50000,4050] 4 times
                    thread 2: [1000,50000,50000,4050] 4 times
            
            Suppose there are 45000 requests.
                    waited:
                    thread 1: [45000] 1 time
                    thread 2: [45000] 1 time
            
                    not wait:
                    thread 1: [1000,44000] 2 times
                    thread 2: [1000,44000] 2 times

a1vin-tian:
OK

> Hi @a1vin-tian, generally the codes are good. Could you add some comments in the codes for helping readers in the future?
> It is easier to read currently, but more contributors are going to read this. Such as `prepareDone`, `countDownLatch`, and about many `wait` and `while`.

dmsolr:
> Hi @a1vin-tian, generally the codes are good. Could you add some comments in the codes for helping readers in the future?
> It is easier to read currently, but more contributors are going to read this. Such as `prepareDone`, `countDownLatch`, and about many `wait` and `while`.

Yeap. In personally, I am feeling the code is too complex to understand.

wu-sheng:
@a1vin-tian I am thinking about how to polish the codes in an easier way. 2 key points we need to think about.
1. Could you add an internal queue concept to avoid using `prepareRequests` in this very complex way.
2. Hide the lock mechanism inside the queue, and make the codes as clear as possible.

If you need, we could set up a meeting to discuss the details.

EvanLjp:
I also worry about the complexity of the codes. For the provider-consumer model, the queue is an easier way to realize it. How   
about to use LinkedBlockingQueue to transfer the data and use an atomic boolean to control the prepare stage

a1vin-tian:
If use LinkedBlockingQueue:
``` java
class BlockingBatchQueue{

private final LinkedBlockingQueue<E> elementData = new LinkedBlockingQueue<>();

        public void putMany(List<E> elements) {
            elementData.addAll(elements);
        }

        public List<E> popMany() throws InterruptedException {
            List<E> subList = new ArrayList<>();
            do {
                E poll = elementData.poll(1000, TimeUnit.MILLISECONDS);
                if (poll != null) {
                    subList.add(poll);
                }
                if (subList.size() >= maxBatchSize || !needFillFully) {
                    return subList;
                }
            }
            while (true);
        }
}
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,73 +105,165 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+        }
+
+        public void disableNeedFillFully() {
+            synchronized (elementData) {
+                needFillFully = false;
+                elementData.notify();
+            }
+        }
+
+        public void enabledNeedFillFully() {
```
wu-sheng, line 261: This method is unused.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,73 +105,165 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+                }
+                List<E> sublist = this.elementData.subList(
+                    0, Math.min(maxBatchSize, this.elementData.size()));
+                List<E> partition = new ArrayList<>(sublist);
+                sublist.clear();
+                return partition;
+            }
+        }
+
+        public void disableNeedFillFully() {
```
wu-sheng, line 254: This should be renamed as `noFurtherPut`.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,73 +105,165 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+        private int maxBatchSize;
+        private boolean needFillFully = true;
+
+        public BlockingBatchQueue(int maxBatchSize) {
+            this.maxBatchSize = maxBatchSize;
+        }
+
+        private final List<E> elementData = new ArrayList<>(50000);
+
+        public void putMany(List<E> elements) {
```
wu-sheng, line 229:
```suggestion
        public void put(List<E> elements) {
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,73 +105,165 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+            synchronized (elementData) {
+                elementData.addAll(elements);
+                if (elementData.size() >= maxBatchSize) {
+                    elementData.notify();
+                }
+            }
+        }
+
+        public List<E> popMany() throws InterruptedException {
```
wu-sheng, line 238:
```suggestion
        public List<E> blockingPop() throws InterruptedException {
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,73 +105,165 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+    private static class BlockingBatchQueue<E> {
+
+        private int maxBatchSize;
+        private boolean needFillFully = true;
+
+        public BlockingBatchQueue(int maxBatchSize) {
+            this.maxBatchSize = maxBatchSize;
+        }
+
+        private final List<E> elementData = new ArrayList<>(50000);
```
wu-sheng, line 227: Move field at the top of the class.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,73 +105,165 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
         }
 
         if (debug) {
             log.info("Batch persistence duration: {} ms", System.currentTimeMillis() - startTime);
         }
     }
+
+    private static class BlockingBatchQueue<E> {
+
+        private int maxBatchSize;
```
wu-sheng, line 220: This should be final.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,73 +105,165 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+        private boolean needFillFully = true;
+
+        public BlockingBatchQueue(int maxBatchSize) {
+            this.maxBatchSize = maxBatchSize;
+        }
```
wu-sheng, line 225: This could be replaced by lombok annotation.

wu-sheng:
You update `skywalking-ui` submodule incorrectly. Please make sure it is unchanged.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,73 +105,165 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
 
         if (debug) {
             log.info("Batch persistence duration: {} ms", System.currentTimeMillis() - startTime);
         }
     }
+
+    private static class BlockingBatchQueue<E> {
+
+        private int maxBatchSize;
+        private boolean needFillFully = true;
```
wu-sheng, line 221:
```suggestion
        private boolean inAppendingMode = true;
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
wu-sheng, line 254:
```suggestion
        public void noFurtherAppending() {
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,73 +105,165 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+    @VisibleForTesting
+    void extractDataAndSave(IBatchDAO batchDAO) {
+
         if (log.isDebugEnabled()) {
             log.debug("Extract data and save");
         }
 
         long startTime = System.currentTimeMillis();
+        HistogramMetrics.Timer allTimer = allLatency.createTimer();
+        // Use flag entire method is done or exception exit. Mainly used to handle exceptions to let the thread exit.
```
wu-sheng, line 117:
```suggestion
        // Use `stop` as a control signal to make fail-fast in the persistence process.
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -77,9 +83,16 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
             MetricsTag.EMPTY_KEY, MetricsTag.EMPTY_VALUE
         );
+        allLatency = metricsCreator.createHistogramMetric(
+            "persistence_timer_bulk_all_latency", "Latency of the all stage in persistence timer",
+            MetricsTag.EMPTY_KEY, MetricsTag.EMPTY_VALUE
+        );
+
         syncOperationThreadsNum = moduleConfig.getSyncThreads();
         maxSyncoperationNum = moduleConfig.getMaxSyncOperationNum();
+        batchExecutorService = Executors.newSingleThreadExecutor();
```
wu-sheng, line 93: We will use one more thread because of this, besides the new add `moduleConfig#getPrepareThreads()`. Do we have a way to avoid this? Every thread is a very expensive resource in the OAP.

wu-sheng:
@EvanLjp  @dmsolr Could you check the latest codes, it should be easier to read. I have commented about what should be polished.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
a1vin-tian: Yes. We can avoid this thread by just use multi-consumer. It will bring some competition, but it is ok.
```
            List<Future<?>> batchFutures = new ArrayList<>();
            for (int i = 0; i < syncOperationThreadsNum; i++) {
                Future<?> batchFuture = executorService.submit(() -> {
                    // consume the metrics
                    while (!stop.get()) {
                        List<PrepareRequest> partition = prepareQueue.popMany();
                        if (partition.isEmpty()) {
                            break;
                        }
                        HistogramMetrics.Timer executeLatencyTimer = executeLatency.createTimer();
                        try {
                            if (CollectionUtils.isNotEmpty(partition)) {
                                batchDAO.synchronous(partition);
                            }
                        } catch (Throwable e) {
                            log.error(e.getMessage(), e);
                        } finally {
                            executeLatencyTimer.finish();
                        }
                    }
                    return null;
                });
                batchFutures.add(batchFuture);
            }

```


for (Future<?> result : batchFutures) {
      result.get();
 }

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,73 +105,165 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+                });
+            });
 
-            HistogramMetrics.Timer executeLatencyTimer = executeLatency.createTimer();
-            try {
-                List<List<PrepareRequest>> partitions = Lists.partition(prepareRequests, maxSyncoperationNum);
-                CountDownLatch countDownLatch = new CountDownLatch(partitions.size());
-                for (final List<PrepareRequest> partition : partitions) {
-                    executorService.submit(() -> {
+            Future<?> batchFuture = batchExecutorService.submit(() -> {
```
dmsolr, line 161: `BatchExecutorService` doesn't seem to be necessary.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,73 +105,165 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+        private final List<E> elementData = new ArrayList<>(50000);
+
+        public void putMany(List<E> elements) {
+            synchronized (elementData) {
+                elementData.addAll(elements);
+                if (elementData.size() >= maxBatchSize) {
+                    elementData.notify();
+                }
+            }
+        }
```
dmsolr, line 236: should there be a queue full check?
personally, I don't recommend implementing a data structure like this in so important module.

a1vin-tian: The `private final List<PrepareRequest> prepareRequests = new ArrayList<>(50000);` did not check queue full before.
I think that there shouldn't be a queue full check.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
wu-sheng: I asked too. https://github.com/apache/skywalking/pull/7153#discussion_r657171941

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
wu-sheng: This is an Array list, it could resize. What do you mean full check? @dmsolr

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,73 +105,153 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+
+        @Getter
+        private final int maxBatchSize;
+
+        @Getter
+        private boolean inAppendingMode = true;
+
+        private final List<E> elementData = new ArrayList<>(50000);
+
+        public void push(List<E> elements) {
```
kezhenxu94, line 224: I suggest using the same naming family as JDK's classes, `push` / `pop` for Stacks, `offer` / `poll` for Queues, since this is a queue, let's use `offer` / `poll`, WDYT?

a1vin-tian:
JMH result. Please check the code correctly. 
```
Benchmark                                             Mode  Cnt   Score   Error  Units
BlockingBatchQueueBenchmark.testLinkedBlockingQueue  thrpt   10   0.317 ± 0.032  ops/s
BlockingBatchQueueBenchmark.testReentrantLock        thrpt   10  16.018 ± 1.553  ops/s
BlockingBatchQueueBenchmark.testSynchronized         thrpt   10  16.769 ± 0.533  ops/s
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
a1vin-tian: ok. got it

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/BlockingBatchQueue.java
```diff
@@ -0,0 +1,41 @@
...
+package org.apache.skywalking.oap.server.core.storage;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.util.List;
+
+/**
+ A blocking batch queue.
+
+ Use offer a list to the Queue or blocking poll a list from Queue.
+ */
```
wu-sheng, line 28: The style of this comment is very strange, could you fix this?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -43,16 +48,18 @@
 @Slf4j
 public enum PersistenceTimer {
     INSTANCE;
-    private Boolean isStarted = false;
+    @VisibleForTesting
+    Boolean isStarted = false;
```
wu-sheng, line 52: You have `@VisibleForTesting`, why remove private? And why use `Boolean`, rather than `boolean`. I can't see anyone uses it as NULL. I know this is a legacy fault, but please fix it.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,73 +105,165 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+    @RequiredArgsConstructor
+    static class DefaultBlockingBatchQueue<E> implements BlockingBatchQueue<E> {
+
+        @Getter
+        private final int maxBatchSize;
+
+        @Getter
+        private boolean inAppendingMode = true;
+
+        private final List<E> elementData = new ArrayList<>(50000);
```
wu-sheng, line 216: Should we use `maxBatchSize` * 2(or *3) as the initial size? As we only return in `poll` after it reaches `maxBatchSize`. Then we face `resize` every time, which seems not right.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,73 +105,165 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
+
+        @Override
+        public void offer(List<E> elements) {
+            synchronized (elementData) {
```
wu-sheng, line 220: This should work when `inAppendingMode == true`.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -43,16 +48,18 @@
 @Slf4j
 public enum PersistenceTimer {
     INSTANCE;
-    private Boolean isStarted = false;
+    @VisibleForTesting
```
wu-sheng, line 51: > 2021-06-24T13:37:45.9314106Z [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:testCompile (default-testCompile) on project server-core: Compilation failure
2021-06-24T13:37:45.9317124Z [ERROR] /home/runner/work/skywalking/skywalking/oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimerTest.java:[80,34] isStarted has private access in

Is this annotation really working?

a1vin-tian: It just means that I expect him to be used for testing, it has no practical meaning

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -92,73 +105,165 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
     }
+
+    @RequiredArgsConstructor
+    static class DefaultBlockingBatchQueue<E> implements BlockingBatchQueue<E> {
+
+        @Getter
+        private final int maxBatchSize;
+
+        @Getter
+        private boolean inAppendingMode = true;
```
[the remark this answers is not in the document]
a1vin-tian: This should not add `volatile`. This field is operated in the `synchronized `

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
wu-sheng: Then don't remove `private`, you should use `Whitebox` for tests.

wu-sheng (approved):
LGTM. @dmsolr @EvanLjp Please check codes, more feedback is welcome.

EvanLjp (approved):
LGTM, It's more clear than before.

wu-sheng:
@a1vin-tian Before merge, could you run this latest codes on prod or test env, let's see how self observability looks like.

a1vin-tian:
> @a1vin-tian Before merge, could you run this latest codes on prod or test env, let's see how self observability looks like.

Okay, I will do it these two days.

a1vin-tian:
@wu-sheng 

image

After: 
image

Befor:
image
as time goes on. 
image

wu-sheng:
@a1vin-tian Thank you so much for your good polish, patient, and keeping testing in real env. I am going to merge this.

PR #7159, opened 2021-06-22
Python celery plugin

tom-pytel:


tom-pytel:
Are twitter links a good idea to put in code?

kezhenxu94:
> Are twitter links a good idea to put in commits? (not mine, but is what triggered the dead link checker)

That is our official account so I think it's OK, dead link checker is sometimes broken never mind we can rerun it

kezhenxu94 (approved):
Thanks!

PR #7171, opened 2021-06-25
Add an agent plugin for ShardingSphere 5.0.0-beta

TeslaCN:
### Add an agent plugin to support ShardingSphere 5.0.0-beta

- [X] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Related to #7173 .

image


wu-sheng:
Could we consider removing some plugins for old versions? We have hosted many versions for ShardingShpere, I am feeling too many.

TeslaCN:
> Could we consider removing some plugins for old versions? We have hosted many versions for ShardingShpere, I am feeling too many.

Indeed. Maybe we can remove the oldest version. #7173

test/plugin/scenarios/shardingsphere-5.0.0-beta-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,183 @@
...
+      peer: ''
+      skipAnalysis: false
+      tags:
+      - {key: db.statement, value: SELECT * FROM t_order}
+    - {operationName: /ShardingSphere/routeSQL/, operationId: 0, parentSpanId: 2,
+      spanId: 3, spanLayer: Unknown, startTime: not null, endTime: not null,
+      componentId: 60, isError: false, spanType: Local, peer: '', skipAnalysis: false}
+    - {operationName: /ShardingSphere/rewriteSQL/, operationId: 0, parentSpanId: 2,
+      spanId: 4, spanLayer: Unknown, startTime: not null, endTime: not null,
+      componentId: 60, isError: false, spanType: Local, peer: '', skipAnalysis: false}
```
wu-sheng, line 146: These spans are not well formated, and more below, please fix

wu-sheng:
Does this plugin still support ShardingSphere SDK only?

TeslaCN:
> Does this plugin still support ShardingSphere SDK only?

It also supports the ShardingSphere Proxy.

test/plugin/scenarios/shardingsphere-5.0.0-beta-scenario/config/expectedData.yaml
TeslaCN: Fixed.

wu-sheng:
> > Does this plugin still support ShardingSphere SDK only?
> 
> It also supports the ShardingSphere Proxy.

I would like to ask how? I can't see the enhancement to make MySQL/PostgreSQL plugin(JDBC) inject any context in the SQL statement to propagate. How could we link the app to ShardingSphere proxy? This is missed for a long time.

TeslaCN:
> > > Does this plugin still support ShardingSphere SDK only?
> > 
> > 
> > It also supports the ShardingSphere Proxy.
> 
> I would like to ask how? I can't see the enhancement to make MySQL/PostgreSQL plugin(JDBC) inject any context in the SQL statement to propagate. How could we link the app to ShardingSphere proxy? This is missed for a long time.

I'm sorry that I misunderstood the 'support'. I think the context cannot propagate to Proxy for now. There will be a new context when request sent to the Proxy.

wu-sheng:
Another question, do we expect another API changes from beta to GA? I have noticed ShardingSphere keeps changing tracing-related APIs. This is a little frustrating

wu-sheng (commented):
Do you accidentally remove `supported-list.md` change? I remember it was there, but not now.

TeslaCN:
> Do you accidentally remove `supported-list.md` change? I remember it was there, but not now.

I didn't update this file before.

wu-sheng (approved):
Please finish other tasks of #7173 as soon as possible. We need ShardingSphere community's decision. This new plugin is good to me.

TeslaCN:
> Another question, do we expect another API changes from beta to GA? I have noticed ShardingSphere keeps changing tracing-related APIs. This is a little frustrating

There are massive changes from ShardingSphere 4.x to 5.0. From beta to GA the kernel API might not change.

wu-sheng:
I think we are facing a mssql test fails

> manifest for microsoft/mssql-server-linux:latest not found: manifest unknown: manifest unknown

Could you help with checking logs? It may not be caused by your codes. But it keeps failing.

TeslaCN:
> I think we are facing a mssql test fails
> 
> > manifest for microsoft/mssql-server-linux:latest not found: manifest unknown: manifest unknown
> 
> Could you help with checking logs? It may not be caused by your codes. But it keeps failing.

I've checkd the https://github.com/apache/skywalking/actions/workflows/plugins-test.1.yaml and this case also failed in other PRs.
It looks like docker's issue or the network is unstable. Consider creating an issue for this problem if it always failed.

wu-sheng:
> > I think we are facing a mssql test fails
> > > manifest for microsoft/mssql-server-linux:latest not found: manifest unknown: manifest unknown
> > 
> > 
> > Could you help with checking logs? It may not be caused by your codes. But it keeps failing.
> 
> I've checkd the https://github.com/apache/skywalking/actions/workflows/plugins-test.1.yaml and this case also failed in other PRs.
> It looks like docker's issue or the network is unstable. Consider creating an issue for this problem if it always failed.

I can't check the merged PR again. but with this fail, other test cases wouldn't run. I think this is an easy fix issue. Could you take a look at the logs? It seems an image missing.

TeslaCN:
> > > I think we are facing a mssql test fails
> > > > manifest for microsoft/mssql-server-linux:latest not found: manifest unknown: manifest unknown
> > > 
> > > 
> > > Could you help with checking logs? It may not be caused by your codes. But it keeps failing.
> > 
> > 
> > I've checkd the https://github.com/apache/skywalking/actions/workflows/plugins-test.1.yaml and this case also failed in other PRs.
> > It looks like docker's issue or the network is unstable. Consider creating an issue for this problem if it always failed.
> 
> I can't check the merged PR again. but with this fail, other test cases wouldn't run. I think this is an easy fix issue. Could you take a look at the logs? It seems an image missing.

I ran the `mssql-jtds-scenario` locally but didn't reproduce what happened in the GitHub Action.

wu-sheng (approved):
I fixed mssql image issue, which should be very easy to be fixed by yourself. Next time, you should do this on your own.

TeslaCN:
> I fixed mssql image issue, which should be very easy to be fixed by yourself. Next time, you should do this on your own.

Roger that. Thanks!

PR #7177, opened 2021-06-25
fix avg response time of gRPC is not right.

844067874:


apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCallListener.java
```diff
@@ -82,6 +82,12 @@ public void onCancel() {
             throw t;
         } finally {
             ContextManager.stopSpan();
+            try {
+                AbstractSpan abstractSpan = ContextManager.activeSpan();
+                ContextManager.stopSpan(abstractSpan);
```
wu-sheng, line 87: This seems duplicated logic. What do you mean about this?

PR #7179, opened 2021-06-26
fix the response time of gRPC is not right.

844067874:


apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCallListener.java
```diff
@@ -101,4 +96,18 @@ public void onHalfClose() {
             ContextManager.stopSpan();
         }
     }
+
+    @Override
+    public void onComplete() {
+        super.onComplete();
+        ContextManager.stopSpan();
```
wu-sheng, line 103: Is this always in the same thread?

844067874: in the same thread

844067874: give me an email address?i will describe detail  to you what problem we met

wu-sheng: All case should be open and repeated. You could write the case and graph here.

844067874: in the class`org.apache.skywalking.apm.plugin.grpc.v1.server.ServerInterceptor` will create an entry gRPC endpoint,but it always cost a few milliseconds,because it just new a `TracingServerCallListener`.
```
final AbstractSpan span = ContextManager.createEntrySpan(OperationNameFormatUtil.formatOperationName(call.getMethodDescriptor()), contextCarrier);
        span.setLayer(SpanLayer.RPC_FRAMEWORK);
        try {
            return new TracingServerCallListener<>(handler.startCall(new TracingServerCall<>(call, ContextManager.capture()), headers), call.getMethodDescriptor(), ContextManager.capture());
        } finally {
            ContextManager.stopSpan();
        }

```
so ,it cause  the avg response of gRPC is nearly 0 ms.
image

the really cost some time is the endpoint of xx/server/Request/onComplete,

image

`org.apache.skywalking.apm.plugin.grpc.v1.server.TracingServerCallListener`
```
@Override
    public void onHalfClose() {
        final AbstractSpan span = ContextManager.createLocalSpan(operationPrefix + REQUEST_ON_COMPLETE_OPERATION_NAME);
        span.setLayer(SpanLayer.RPC_FRAMEWORK);

        try {
            super.onHalfClose();
        } catch (Throwable t) {
            throw t;
        } finally {
            ContextManager.stopSpan();
        }
    }
```

wu-sheng: I agree that, `onComplete` or `onError` should be the right place to finish span. Only concern is about across threads.

wu-sheng:
You need to update the changes.md in the root folder.

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCallListener.java
844067874: image
I print the thread id in the corresponding location ,thread is all the same
```
@Override
    public void onComplete() {
        String log ="method:%s,threadId:%s,currentTimeMillis:%s";
        System.out.println(String.format(log,"TracingServerCallListener.onComplete", Thread.currentThread().getId(), System.currentTimeMillis()));
        super.onComplete();
    }

    @Override
    public void onReady() {
        String log ="method:%s,threadId:%s,currentTimeMillis:%s";
        System.out.println(String.format(log,"TracingServerCallListener.onReady", Thread.currentThread().getId(), System.currentTimeMillis()));
        super.onReady();
    }
```

Is there anything I haven't considered

wu-sheng: > Is there anything I haven't considered

I am not sure how it looks like in high traffic load. Thread scheduling changes with load.

844067874: so you worry about the onReady method execution thread is not the same as the thread when executing onComplete

`org.apache.skywalking.apm.plugin.grpc.v1.server.TracingServerCallListener`
```
@Override
    public void onComplete() {
        super.onComplete();
        ContextManager.stopSpan();
    }

    @Override
    public void onReady() {
        final AbstractSpan span = ContextManager.createEntrySpan(operation, contextCarrier);
        span.setLayer(SpanLayer.RPC_FRAMEWORK);
        super.onReady();
    }
```

wu-sheng: As you are using `ContextManager#activeSpan`, then once they are not in the same thread, the memory leaks.

If there is a chance they are not(I didn't read gRPC source codes) always in the same thread, you should make the created in async mode, and use `span#asyncFinish` to close it safely. Of course, only need to do so if it crosses threads in some cases.

wu-sheng:
image

Check CI testing, https://github.com/apache/skywalking/pull/7179/checks?check_run_id=2921556341. You are clearly facing bugs in grpc case. Considering an NPE shows, you are facing across threads issue as I asked.

844067874:
I try it in my machine,send much request concurrently,it also throw NPE,i has fixed it,see detail the 
class `org.apache.skywalking.apm.plugin.grpc.v1.server.TracingServerCallListener`
```

@Override
    public void onComplete() {
        super.onComplete();
        asyncSpan.asyncFinish();
    }

    @Override
    public void onReady() {
        final AbstractSpan span = ContextManager.createEntrySpan(operation, contextCarrier);
        span.setLayer(SpanLayer.RPC_FRAMEWORK);
        contextSnapshot = ContextManager.capture();
        asyncSpan = span.prepareForAsync();
        ContextManager.stopSpan(asyncSpan);

        super.onReady();
    }
```

wu-sheng:
> I try it in my machine,send much request concurrently,it also throw NPE,i has fixed it

`#onComplete` causes the NPE, which means, this method is being called in a thread without tracing context. 
You have to get clear about gRPC thread model. Your changes seem to make things worse, from an inaccurate duration to unexpected thread context.

844067874:
the two method may be invoked by different thread,so I use `prepareForAsync` and `asyncFinish`,
the `asyncSpan`is hold by each `org.apache.skywalking.apm.plugin.grpc.v1.server.TracingServerCallListener`.
I test ,it acts normally,just as you say 

> If there is a chance they are not(I didn't read gRPC source codes) always in the same thread, you should make the created in async mode, and use span#asyncFinish to close it safely. Of course, only need to do so if it crosses threads in some cases.

```
@Override
    public void onComplete() {
        super.onComplete();
        asyncSpan.asyncFinish();
    }

    @Override
    public void onReady() {
        final AbstractSpan span = ContextManager.createEntrySpan(operation, contextCarrier);
        span.setLayer(SpanLayer.RPC_FRAMEWORK);
        contextSnapshot = ContextManager.capture();
        asyncSpan = span.prepareForAsync();
        ContextManager.stopSpan(asyncSpan);

        super.onReady();
    }
```

wu-sheng:
Read logs of https://github.com/apache/skywalking/pull/7179/checks?check_run_id=2924255893. You should fix the expected data to match your new interceptor implementations.

wu-sheng:
There is no chance you could make a segment as a string, `not null` is a common case. Could you explain why this literal string could pass the test locally?

844067874:
I see source code in `https://github.com/apache/skywalking-agent-test-tool/tree/master/validator`,it dose not verify the `segmentId`,and the previous version ,it also write 
```
segmentItems:
  - serviceName: correlation-autotag-scenario
    segmentSize: 12
    segments:
      - segmentId: not null
        spans:
          - operationName: /correlation-autotag-scenario/case/healthCheck
            operationId: 0
            parentSpanId: -1
            spanId: 0
            spanLayer: Http
            startTime: gt 0
            endTime: gt 0
            componentId: 1
            isError: false
            spanType: Entry
            peer: ''
            skipAnalysis: false
            tags:
              - {key: url, value: 'http://localhost:8080/correlation-autotag-scenario/case/healthCheck'}
              - {key: http.method, value: HEAD}
      - segmentId: 78a0cdef42a74b4ba80b48bf72b8558d.37.16042096947740000
        spans:
```

wu-sheng:
Grpc cases fail, I believe they are caused by new codes. Please check.

wu-sheng (approved):
LGTM.

PR #7182, opened 2021-06-27
Perf: optimize Envoy ALS analyzer performance in high traffic load scenario (reduce ~1cpu in ~10k RPS)

kezhenxu94:


kezhenxu94:
From what I've tested, `LambdaMetafactory` can nearly reach to the same performance as static call to `setXXX` method, much faster than reflection

kezhenxu94:
I'll check other places to see whether there is critical path that also use reflection

kezhenxu94:
Confirmed, no more place to update in this PR

wu-sheng (approved):
LGTM.

PR #7183, opened 2021-06-27
Logback grpc log report support AsyncAppender

wuwen5:
image

```TraceId``` has been lost when use `AsyncAppender `

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@ Release Notes.
 * Add `Neo4j-4.x` plugin.
 * Correct `profile.duration` to `profile.max_duration` in the default `agent.config` file.
 * Fix the reponse time of gRPC.
+* Support logback `GRPCLogClientAppender` with `AsyncAppender` 
 
 #### OAP-Backend
 * Disable Spring sleuth meter analyzer by default.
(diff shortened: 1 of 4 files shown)


wu-sheng:
You should update the document about how to use this. And update `change.md` in the root folder.

EvanLjp:
I have a question that is the original grpc appender already supports datacarrier component, do we need logback async appender? @wu-sheng

wu-sheng:
> I have a question that is the original grpc appender already supports datacarrier component, do we need logback async appender? @wu-sheng

I have the same question. But due to you have product env experience in this part, I want to hear your opinion.

EvanLjp:
@wu-sheng It works. 
 
image

wu-sheng:
@EvanLjp Thanks. 

@wuwen5 Do you know the point we are discussing? This support means you have an async reporter(SkyWalking) behind an aync logger(logback).

wuwen5:
@wu-sheng @EvanLjp Thinks.

I see. I seen datacarrier component, so we don't need logback async appender

wu-sheng:
> @wu-sheng @EvanLjp Thanks.
> 
> I see. I seen datacarrier component, so we don't need logback async appender

OK, if you can have a specific reason for keeping this, please let us known. Thank you for the contribution

Issue #7170, opened 2021-06-25
missing db.statement tag for jedis blocking api


Yolo-Hao:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Bug
- Which version of SkyWalking, OS, and JRE?
SkyWalking latest

- What happened?
when using jedis blocking api like `blpop`，it won't have a `db.statement` tag, because it only check type `String` and `byte[]` for first argument as below:
```java
if (allArguments.length > 0 && allArguments[0] instanceof String) {
    Tags.DB_STATEMENT.set(span, method.getName() + " " + allArguments[0]);
} else if (allArguments.length > 0 && allArguments[0] instanceof byte[]) {
    Tags.DB_STATEMENT.set(span, method.getName());
}
```
maybe we need to add a check for `Integer`

wu-sheng:
Feel free to enhance plugin, you could send a pull request. Also, you need to enhancr plugin integration test case.

wu-sheng:
Closing due to author's comments, https://github.com/apache/skywalking/pull/7185#issuecomment-870201002

PR #7185, opened 2021-06-28
optimize db.statement of jedis blocking api with timeout

Yolo-Hao:
### Fix miss db.statement of jedis blocking api with a timeout parameter [#7170](https://github.com/apache/skywalking/issues/7170)
`JedisMethodInterceptor` only check `String` and `byte[]` as the first argument type, but some commands like `blpop` use `int` as timeout, so it won't have a `db.statement` tag in this case.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<7170>.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -51,6 +51,9 @@ Release Notes.
 * Add Python celery plugin.
 * Fix default config for metrics.
 
+#### jedis-2.x-plugin
+* add `db.statement` for block command with a timeout parameter like `blpop`
+
 #### Documentation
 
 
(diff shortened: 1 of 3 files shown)


wu-sheng:
Please update the changes.md in the root folder. And integration test hasn't not been updated with expected data. Could you update them too?

wu-sheng:
Still need integration tests.

Yolo-Hao:
> Still need integration tests.

looking~
it's my first PR, so i am not familiar with all processes.

wu-sheng:
I know, GitHub shows that. 
1. This is the case you need to enhance, https://github.com/apache/skywalking/tree/master/test/plugin/scenarios/jedis-scenario
2. http://skywalking.apache.org/docs/main/latest/en/guides/java-plugin-development-guide/#plugin-test-tool This is the doc you need to follow. Tests could be run locally in Linux(maybe Mac?)

Yolo-Hao:
hi, afer test with more versions, i think this PR is worthless.
`2.4.x` and `2.5.x` need to ehance method `blpop (int timeout, ...)` and `blpop(String...keys)`,
but higher version like `2.6.x`, only need to enhance `blpop(String...keys)` because other methods will call it nested.
to solve it, we need to use witness class to distingush these versions, but it will create more instrumentation define class,
and just works for `blpop` and `brpop`, so i think is worthless, though current version will create two spans in one `blpop` call.

for user who just want to support `2.6.x` or higher, simply change `named("blpop")` to `named("blpop").and(takesArgument(0, String[].class))` and deal with it in `JedisMethodInterceptor`, it will works.

Thanks  for your help

PR #7189, opened 2021-06-28
revert #7091

### LINKED ISSUE #7181: Revert 7091:Use MAL to calculate JVM metrics, remove OAL dependency.
@liqiangz We have received confirmation, MAL is not suitable in high load. This change should be reverted. #7091 can't be reverted automatically through GitHub, could you send a pull request to finish this revert?

This performance issue is not caused by MAL, actually, we found it through LAL, which is most likely caused by the groovy engine. Check #7180 for more details.

liqiangz:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7181.

Test cases are retained and other is revert.


wu-sheng (approved):
Very apprieciated. Sorry, we made the wrong call before full evaluation.

liqiangz:
> Very apprieciated. Sorry, we made the wrong call before full evaluation.

Never mind.

PR #7194, opened 2021-06-29
Support configure sampling rate dynamically for service dimension on the backend side

HendSame:
### <Feature description>
- As we know, it supports configure `sampling rate` and `trace latency` on the backend side, but it is global.
- Can I configure sampling rate dynamically follow every service or instance or interface latitude in some cases when I only want to see the exception trace, or  the trace that has cost a very long time? But I do not want to configure the global `sampling rate` and `trace latency`, by reason of the storage layer cannot support large data load.
- We  can configure in a yaml:
```
services:
  - name: serverName # the server name which show skywalking-ui
    instances:
      - name: hostName@127.0.0.1 # the server instance which show skywalking-ui
        sampleRate: 900 # instance latitude
        duration: 10000 # trace latency time
    endpoints:
      - name: '/url' # the endpoint
        sampleRate: 800
        duration: 10000 # trace latency time
    sampleRate: 1000 # endpoint latitude
    duration: 10000 # trace latency time
```


wu-sheng:
2 arguments about this kind of changes.
1. Instance is up and down, with scaling. The name of instance could/should change from time to time, especially in today's k8s deployment. I can't see the point of set sampling for a single instance.
2. About endpoint, a.k.a. your interface. This would be a huge number of sampling, and as this sampler has to keep the seed for every endpoint(URI), if URI is parameterized, your OAP will be OOM eventually.

HendSame:
> 2 arguments about this kind of changes.
> 
> 1. Instance is up and down, with scaling. The name of instance could/should change from time to time, especially in today's k8s deployment. I can't see the point of set sampling for a single instance.
> 2. About endpoint, a.k.a. your interface. This would be a huge number of sampling, and as this sampler has to keep the seed for every endpoint(URI), if URI is parameterized, your OAP will be OOM eventually.

In my opinion, there are some premises of this usage scenario about the tow arguments:

- Maybe in our performance testing environment, it will be of great help to us in troubleshooting performance problems when doing full-link performance tests, that is also a good support  point for full-link monitoring to full-link pressure testing.And at the same time, we usually only configure  endpoint sampling rate for endpoints according to our needs .

- Or it could be that some instance of our production environment has an exception at that point in time. Of course, in most scenarios, the priority is to fix the problem and keep one or two instances in exception for troubleshooting.

- `URI is parameterized`,I  remember that if URI is parameterized, it will be a parameterized URI as the span's `enpointName` such as  : `{POST}/urlpath/{id}`.

image


So I think should pay more attention to usage scenarios.And I think this is a kind of ability we should provide, and how to use it depends on the user.

wu-sheng:
> URI is parameterized,I remember that if URI is parameterized, it will be a parameterized URI as the span's enpointName such as : {POST}/urlpath/{id}.

It only could be detected by SpringMVC spring, nothing more. Once backend supports a mechanism, we need to consider all language agents.

> Or it could be that some instance of our production environment has an exception at that point in time. Of course, in most scenarios, the priority is to fix the problem and keep one or two instances in exception for troubleshooting.

But still, you haven't explained the OOM risk. As the same as the endpoint case.

HendSame:
> > URI is parameterized,I remember that if URI is parameterized, it will be a parameterized URI as the span's enpointName such as : {POST}/urlpath/{id}.
> 
> It only could be detected by SpringMVC spring, nothing more. Once backend supports a mechanism, we need to consider all language agents.
> 
> > Or it could be that some instance of our production environment has an exception at that point in time. Of course, in most scenarios, the priority is to fix the problem and keep one or two instances in exception for troubleshooting.
> 
> But still, you haven't explained the OOM risk. As the same as the endpoint case.

That's it. If there are a lots of configurations. 
Can we limit the number of configurations to avoid OOM?
Or  not  support instances and enpoint dimensions? It is also good to be able to support the service dimension.

wu-sheng:
> Can we limit the number of configurations to avoid OOM?

If it is a dynamic configuration, I can't see how to limit it.

> Or not support instances and enpoint dimensions? It is also good to be able to support the service dimension.

Service level dynamic sampling makes sense to me.

HendSame:
Got it！It has been updated.

HendSame:
> Service level dynamic sampling makes sense to me.

Done.

wu-sheng (commented):
The boundary between `custom-trace-sample-rate` and existing `sampleRate` is not clear. If you want to do service based sample rate setting, the first step should make it supported in static configuration. Sample rate is a static configuration in the `application.yml`, you should make it separate in another `trace-sample-rate-setting.yml`, then, make the dynamic configuration works. Also, this should not be a different configuration than existing logic, otherwise, the users would be very confused, which sample rate would take effect if they set both.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/AnalyzerModuleProvider.java
```diff
@@ -111,18 +117,18 @@ public void prepare() throws ServiceNotProvidedException, ModuleStartException {
     public void start() throws ModuleStartException {
         // load official analysis
         getManager().find(CoreModule.NAME)
-                    .provider()
-                    .getService(OALEngineLoaderService.class)
-                    .load(CoreOALDefine.INSTANCE);
+                .provider()
+                .getService(OALEngineLoaderService.class)
+                .load(CoreOALDefine.INSTANCE);
```
dmsolr, line 122: As far as possible, don't reformat irrelevant code styles. :)

HendSame: ok

HendSame:
> The boundary between `custom-trace-sample-rate` and existing `sampleRate` is not clear. If you want to do service based sample rate setting, the first step should make it supported in static configuration. Sample rate is a static configuration in the `application.yml`, you should make it separate in another `trace-sample-rate-setting.yml`, then, make the dynamic configuration works. Also, this should not be a different configuration than existing logic, otherwise, the users would be very confused, which sample rate would take effect if they set both.

Like that configure sampling rate by the same configuration in the `application.yml`?
```
agent-analyzer:
  selector: ${SW_AGENT_ANALYZER:default}
  default:
    #sampleRate: ${SW_TRACE_SAMPLE_RATE:10000} # The config had moved to 'traceSampleRateSettingFile' and had been replaced by 'global.sampleRate' that the config in 'traceSampleRateSettingFile'.
    traceSampleRateSettingFile: ${SW_TRACE_SAMPLE_RATE_SETTING_FILE:trace-sample-rate-setting.yml}
```

And `trace-sample-rate-setting.yml`:

```
global:
  sampleRate: 10000 # the replace the old config `sampleRate` in  the application.yml.
  duration: 10000 # the replace the old config `slowTraceSegmentThreshold` in  the application.yml
services:
  - name: serverName # the server name which show skywalking-ui
    sampleRate: 1000 # endpoint latitude. It has a higher priority than `global.sampleRate`
    duration: 10000 # trace latency time.  It has a higher priority than `global.duration`
```

wu-sheng:
> Like that configure sampling rate by the same configuration in the `application.yml`?

Yes, something like that. `global` should be renamed as `default`.

wu-sheng:
> # the server name which show skywalking-ui

Notice, this is not a correct comment for the service name. Service has been defined clearly in the document, you don't need to mention it here.

HendSame:
@wu-sheng Updated

GerryYuan:
> > The boundary between `custom-trace-sample-rate` and existing `sampleRate` is not clear. If you want to do service based sample rate setting, the first step should make it supported in static configuration. Sample rate is a static configuration in the `application.yml`, you should make it separate in another `trace-sample-rate-setting.yml`, then, make the dynamic configuration works. Also, this should not be a different configuration than existing logic, otherwise, the users would be very confused, which sample rate would take effect if they set both.
> 
> Like that configure sampling rate by the same configuration in the `application.yml`?
> 
> ```
> agent-analyzer:
>   selector: ${SW_AGENT_ANALYZER:default}
>   default:
>     #sampleRate: ${SW_TRACE_SAMPLE_RATE:10000} # The config had moved to 'traceSampleRateSettingFile' and had been replaced by 'global.sampleRate' that the config in 'traceSampleRateSettingFile'.
>     #slowTraceSegmentThreshold: ${SW_SLOW_TRACE_SEGMENT_THRESHOLD:-1} # The config had moved to 'traceSampleRateSettingFile' and had been replaced by 'global.duration' that the config in 'traceSampleRateSettingFile'.
>     traceSampleRateSettingFile: ${SW_TRACE_SAMPLE_RATE_SETTING_FILE:trace-sample-rate-setting.yml}
> ```
> 
> And `trace-sample-rate-setting.yml`:
> 
> ```
> global:
>   sampleRate: 10000 # the replace the old config `sampleRate` in  the application.yml.
>   duration: 10000 # the replace the old config `slowTraceSegmentThreshold` in  the application.yml
> services:
>   - name: serverName # the server name which show skywalking-ui
>     sampleRate: 1000 # endpoint latitude. It has a higher priority than `global.sampleRate`
>     duration: 10000 # trace latency time.  It has a higher priority than `global.duration`
> ```
Support different endpoint sampling rate settings? In many cases, the call volume of the core interface is relatively large

wu-sheng:
> > > The boundary between `custom-trace-sample-rate` and existing `sampleRate` is not clear. If you want to do service based sample rate setting, the first step should make it supported in static configuration. Sample rate is a static configuration in the `application.yml`, you should make it separate in another `trace-sample-rate-setting.yml`, then, make the dynamic configuration works. Also, this should not be a different configuration than existing logic, otherwise, the users would be very confused, which sample rate would take effect if they set both.
> > 
> > 
> > Like that configure sampling rate by the same configuration in the `application.yml`?
> > ```
> > agent-analyzer:
> >   selector: ${SW_AGENT_ANALYZER:default}
> >   default:
> >     #sampleRate: ${SW_TRACE_SAMPLE_RATE:10000} # The config had moved to 'traceSampleRateSettingFile' and had been replaced by 'global.sampleRate' that the config in 'traceSampleRateSettingFile'.
> >     #slowTraceSegmentThreshold: ${SW_SLOW_TRACE_SEGMENT_THRESHOLD:-1} # The config had moved to 'traceSampleRateSettingFile' and had been replaced by 'global.duration' that the config in 'traceSampleRateSettingFile'.
> >     traceSampleRateSettingFile: ${SW_TRACE_SAMPLE_RATE_SETTING_FILE:trace-sample-rate-setting.yml}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And `trace-sample-rate-setting.yml`:
> > ```
> > global:
> >   sampleRate: 10000 # the replace the old config `sampleRate` in  the application.yml.
> >   duration: 10000 # the replace the old config `slowTraceSegmentThreshold` in  the application.yml
> > services:
> >   - name: serverName # the server name which show skywalking-ui
> >     sampleRate: 1000 # endpoint latitude. It has a higher priority than `global.sampleRate`
> >     duration: 10000 # trace latency time.  It has a higher priority than `global.duration`
> > ```
> 
> Support different endpoint sampling rate settings? In many cases, the call volume of the core interface is relatively large

No, this has been rejected. Only service sampling rate would be accepted if this PR is ready.

HendSame:
> > > > The boundary between `custom-trace-sample-rate` and existing `sampleRate` is not clear. If you want to do service based sample rate setting, the first step should make it supported in static configuration. Sample rate is a static configuration in the `application.yml`, you should make it separate in another `trace-sample-rate-setting.yml`, then, make the dynamic configuration works. Also, this should not be a different configuration than existing logic, otherwise, the users would be very confused, which sample rate would take effect if they set both.
> > > 
> > > 
> > > Like that configure sampling rate by the same configuration in the `application.yml`?
> > > ```
> > > agent-analyzer:
> > >   selector: ${SW_AGENT_ANALYZER:default}
> > >   default:
> > >     #sampleRate: ${SW_TRACE_SAMPLE_RATE:10000} # The config had moved to 'traceSampleRateSettingFile' and had been replaced by 'global.sampleRate' that the config in 'traceSampleRateSettingFile'.
> > >     #slowTraceSegmentThreshold: ${SW_SLOW_TRACE_SEGMENT_THRESHOLD:-1} # The config had moved to 'traceSampleRateSettingFile' and had been replaced by 'global.duration' that the config in 'traceSampleRateSettingFile'.
> > >     traceSampleRateSettingFile: ${SW_TRACE_SAMPLE_RATE_SETTING_FILE:trace-sample-rate-setting.yml}
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > And `trace-sample-rate-setting.yml`:
> > > ```
> > > global:
> > >   sampleRate: 10000 # the replace the old config `sampleRate` in  the application.yml.
> > >   duration: 10000 # the replace the old config `slowTraceSegmentThreshold` in  the application.yml
> > > services:
> > >   - name: serverName # the server name which show skywalking-ui
> > >     sampleRate: 1000 # endpoint latitude. It has a higher priority than `global.sampleRate`
> > >     duration: 10000 # trace latency time.  It has a higher priority than `global.duration`
> > > ```
> > 
> > 
> > Support different endpoint sampling rate settings? In many cases, the call volume of the core interface is relatively large
> 
> No, this has been rejected. Only service sampling rate would be accepted if this PR is ready.

Yes，Only service sampling rate.

HendSame:
No update? 

image

wu-sheng:
@HendSame This PR was in CI broken status. We need you to make all things ready before we do code review. Do you want to do this again? We can open this and continue.

wu-sheng:
Please make the submodule untouched, and back the latest status of master branch. Also, you need to resolve conflicts, and make all CI tasks passed.

HendSame:
> Please make the submodule untouched, and back the latest status of master branch. Also, you need to resolve conflicts, and make all CI tasks passed.

OK

HendSame:
@wu-sheng  I have created a new PR #7554 that is base on 8.8.0.

PR #7201, opened 2021-06-29
Perf: cache metrics id and entity id

kezhenxu94:
There are a lot of calls to `Metrics.id()` and `ISource.getEntityId`, which calculates the id by manipulating strings in every single call, producing many garbage objects.

In this patch, I cache the id and only calculate the id if it's requested for the first time.

### Improve the performance of `Metrics` and `ISource`
- [x] Add a benchmark for the improvement, refer to [the existing ones](https://github.com/apache/skywalking/blob/master/apm-commons/apm-datacarrier/src/test/java/org/apache/skywalking/apm/commons/datacarrier/LinkedArrayBenchmark.java)
    ```java
    AvgFunction function0;
    AvgFunction function1;
    Random random = new Random();
    long timebucket;
    Object name;

    @Setup
    public void setup() {
        timebucket = random.nextLong();
        name = RandomStringUtils.random(5);
        function0 = new AvgFunction() {
            @Override
            public AcceptableValue<Long> createNew() {
                throw new UnexpectedException("createNew should not be called");
            }

            @Override
            public String id0() {
                return timebucket + ID_CONNECTOR + name;
            }
        };
        function1 = new AvgFunction() {
            @Override
            public AcceptableValue<Long> createNew() {
                throw new UnexpectedException("createNew should not be called");
            }

            @Override
            public String id() {
                return timebucket + ID_CONNECTOR + name;
            }
        };
    }

    @Benchmark
    public void id(Blackhole b) {
        for (int i = 0; i < 10; i++) {
            b.consume(function1.id());
        }
    }

    @Benchmark
    public void id0(Blackhole b) {
        for (int i = 0; i < 10; i++) {
            b.consume(function0.id());
        }
    }

    public static void main(String[] args) throws RunnerException {
                .include(MyBenchmark.class.getSimpleName())
                .addProfiler(GCProfiler.class)
                .build();

        new Runner(opt).run();
    }
    ```
- [x] The benchmark result.
  ```text
  Benchmark                                         Mode  Cnt         Score         Error   Units
  MyBenchmark.id                                   thrpt   25   1201376.291 ±  152856.703   ops/s
  MyBenchmark.id:·gc.alloc.rate                    thrpt   25      3683.137 ±     375.101  MB/sec
  MyBenchmark.id:·gc.alloc.rate.norm               thrpt   25      3392.168 ±      73.398    B/op
  MyBenchmark.id:·gc.churn.G1_Eden_Space           thrpt   25      3686.722 ±     376.161  MB/sec
  MyBenchmark.id:·gc.churn.G1_Eden_Space.norm      thrpt   25      3395.417 ±      74.860    B/op
  MyBenchmark.id:·gc.churn.G1_Survivor_Space       thrpt   25         0.005 ±       0.001  MB/sec
  MyBenchmark.id:·gc.churn.G1_Survivor_Space.norm  thrpt   25         0.005 ±       0.001    B/op
  MyBenchmark.id:·gc.count                         thrpt   25      2181.000                counts
  MyBenchmark.id:·gc.time                          thrpt   25      2336.000                    ms
  MyBenchmark.id0                                  thrpt   25  39069507.415 ± 1553343.663   ops/s
  MyBenchmark.id0:·gc.alloc.rate                   thrpt   25        ≈ 10⁻⁴                MB/sec
  MyBenchmark.id0:·gc.alloc.rate.norm              thrpt   25        ≈ 10⁻⁶                  B/op
  MyBenchmark.id0:·gc.count                        thrpt   25           ≈ 0                counts
  ```
- [x] Links/URLs to the theory proof or discussion articles/blogs. <links/URLs here> NO

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/Metrics.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/Metrics.java
@@ -131,4 +131,16 @@ public abstract class Metrics extends StreamData implements StorageData {
     private boolean isDayBucket() {
         return TimeBucket.isDayBucket(timeBucket);
     }
+
+    private volatile String id;
+
+    @Override
+    public String id() {
+        if (id == null) {
+            id = id0();
+        }
+        return id;
+    }
+
+    protected abstract String id0();
 }
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,6 +48,7 @@ Release Notes.
 * Performance: trim useless metadata fields in Envoy ALS metadata to improve performance.
 * Fix: slowDBAccessThreshold dynamic config error when not configured.
 * Performance: cache regex pattern and result, optimize string concatenation in Envy ALS analyzer.
+* Performance: cache metrics id and entity id in `Metrics` and `ISource`.
 
 #### UI
 * Fix the date component for log conditions.
--- a/oap-server/exporter/src/test/java/org/apache/skywalking/oap/server/exporter/provider/grpc/MockMetrics.java
+++ b/oap-server/exporter/src/test/java/org/apache/skywalking/oap/server/exporter/provider/grpc/MockMetrics.java
@@ -24,7 +24,7 @@ import org.apache.skywalking.oap.server.core.remote.grpc.proto.RemoteData;
 public class MockMetrics extends Metrics {
 
     @Override
-    public String id() {
+    protected String id0() {
         return "mock-metrics";
     }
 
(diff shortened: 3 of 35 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/Metrics.java
```diff
@@ -131,4 +132,18 @@ private boolean isHourBucket() {
...
+    public String id() {
+        if (id == null) {
+            id = id0();
+        }
+        return id;
+    }
+
+    protected String id0() {
+        return null;
+    }
```
wu-sheng, line 148: Is better to keep this `abstract`?

kezhenxu94: DONE

wu-sheng (approved):

PR #7204, opened 2021-06-30
Fix the queue initial size is not set by max batch size accordingly

wu-sheng:
- [ ] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md). No changelog, as this fix target(#7153) is never released.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
@@ -212,7 +212,7 @@ public enum PersistenceTimer {
 
         public DefaultBlockingBatchQueue(final int maxBatchSize) {
             this.maxBatchSize = maxBatchSize;
-            // Use the maxBatchSize * 2 as the initial queue size to avoid ArrayList#grow
+            // Use the maxBatchSize * 3 as the initial queue size to avoid ArrayList#grow
             this.elementData = new ArrayList<>(maxBatchSize * 3);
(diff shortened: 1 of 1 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -203,16 +203,18 @@ private void extractDataAndSave(IBatchDAO batchDAO) {
...
-
+        private final List<E> elementData;
         @Getter
         private boolean inAppendingMode = true;
 
-        private final List<E> elementData = new ArrayList<>(50000 * 3);
+        public DefaultBlockingBatchQueue(final int maxBatchSize) {
+            this.maxBatchSize = maxBatchSize;
+            // Use the maxBatchSize * 2 as the initial queue size to avoid ArrayList#grow
+            this.elementData = new ArrayList<>(maxBatchSize * 3);
```
kezhenxu94, line 216: Comment `maxBatchSize * 2` doesn't accord to real code `maxBatchSize * 3`

wu-sheng: Wrong comments, it was `*3`, but hardcoded `maxBatchSize`

wu-sheng: Fixed

kezhenxu94 (approved):

PR #7212, opened 2021-06-30
Doc: update doc to declare that zookeeper 3.5+ is required in cluster coordinator

kezhenxu94:
We previously updated the zookeeper client version to 3.5.x in #7090, which is incompatible with 3.4 server, we need to update the doc accordingly.


kezhenxu94:
Other reference https://stackoverflow.com/a/61630617/5200161

wu-sheng (approved):

PR #7224, opened 2021-07-01
Increase log automatic injection method for log4j2 and logback.

klmzfyf:
### Add an agent plugin to support [log4j2 & logback automatic inject the program]

The main idea comes from "opentelemetry-java-instrumentation".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -47,6 +47,7 @@ Release Notes.
 * Performance: optimize Envoy ALS analyzer performance in high traffic load scenario (reduce ~1cpu in ~10k RPS).
 * Performance: trim useless metadata fields in Envoy ALS metadata to improve performance.
 * Fix: slowDBAccessThreshold dynamic config error when not configured.
+* Performance: cache regex pattern and result, optimize string concatenation in Envy ALS analyzer.
 
 #### UI
 * Fix the date component for log conditions.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,6 +42,7 @@ Release Notes.
 * Upgrade commons-lang3 to avoid potential NPE in some JDK versions.
 * OAL supports generating metrics from events.
 * Support endpoint name grouping by OpenAPI definitions.
+* Concurrent create PrepareRequest when persist Metrics
 * Fix CounterWindow increase computing issue.
 * Performance: optimize Envoy ALS analyzer performance in high traffic load scenario (reduce ~1cpu in ~10k RPS).
 * Performance: trim useless metadata fields in Envoy ALS metadata to improve performance.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -44,6 +44,7 @@ Release Notes.
 * Support endpoint name grouping by OpenAPI definitions.
 * Fix CounterWindow increase computing issue.
 * Performance: optimize Envoy ALS analyzer performance in high traffic load scenario (reduce ~1cpu in ~10k RPS).
+* Performance: trim useless metadata fields in Envoy ALS metadata to improve performance.
 
 #### UI
 * Fix the date component for log conditions.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -46,6 +46,7 @@ Release Notes.
 * Fix CounterWindow increase computing issue.
 * Performance: optimize Envoy ALS analyzer performance in high traffic load scenario (reduce ~1cpu in ~10k RPS).
 * Performance: trim useless metadata fields in Envoy ALS metadata to improve performance.
+* Fix: slowDBAccessThreshold dynamic config error when not configured.
 
 #### UI
 * Fix the date component for log conditions.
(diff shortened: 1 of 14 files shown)


wu-sheng:
Could you give more explanation about what these changes for? What are the benefits for the users?

SkyWalking never does things according to other projects, such as you said `The main idea comes from "opentelemetry-java-instrumentation".`. We focus on the product and tech parts.

klmzfyf:
> Could you give more explanation about what these changes for? What are the benefits for the users?
> 
> SkyWalking never does things according to other projects, such as you said `The main idea comes from "opentelemetry-java-instrumentation".`. We focus on the product and tech parts.


Ok. I have an ambiguity,sorry.

In our company, skywalking is a very new concept, so I want to minimize the cost of access.
I can now allow users to click a button to access apm. This is pretty cool, so I am not going to change this approach.
so the [log4j2&logback-automatic-agent] can do this. only one button. (To be precise, you also need to add %X(traceid), but this will not change the POM)

As for opentelemetry-java-instrumentation, I just borrowed their cut points. (org.apache.logging.log4j.core.impl.ContextDataInjectorFactory.createInjector and ch.qos.logback.classic.Logger)
The framework and main logic are based on the definition of skywalking.

klmzfyf:
> Could you give more explanation about what these changes for? What are the benefits for the users?
> 
> SkyWalking never does things according to other projects, such as you said `The main idea comes from "opentelemetry-java-instrumentation".`. We focus on the product and tech parts.

Just like this:
image
image
image

In addition, I really like skywalking project.

wu-sheng:
> > Could you give more explanation about what these changes for? What are the benefits for the users?
> > SkyWalking never does things according to other projects, such as you said `The main idea comes from "opentelemetry-java-instrumentation".`. We focus on the product and tech parts.
> 
> Ok. I have an ambiguity,sorry.
> 
> In our company, skywalking is a very new concept, so I want to minimize the cost of access.
> I can now allow users to click a button to access apm. This is pretty cool, so I am not going to change this approach.
> so the [log4j2&logback-automatic-agent] can do this. only one button. (To be precise, you also need to add %X(traceid), but this will not change the POM)
> 
> As for opentelemetry-java-instrumentation, I just borrowed their cut points. (org.apache.logging.log4j.core.impl.ContextDataInjectorFactory.createInjector and ch.qos.logback.classic.Logger)
> The framework and main logic are based on the definition of skywalking.

OK, I can get your points now. But we are very careful to provide a plugin like this. Actually we discussed about this from the day at beginning of the toolkit you see right now.
Once we have an auto instrumentation log injection, we face a risk to break user's log analysis system, such as prod env's ELK. Because we inject something breaking the existing pattern, but users would not expect this, as our agent is out of box, people are new to the project.
Also, we will have to inject all the logs, rather than the necessary logs which cause unnecessary costs.

I am not sure whether these make sense to you, but those are the reasons we don't have this today.

klmzfyf:
> > > Could you give more explanation about what these changes for? What are the benefits for the users?
> > > SkyWalking never does things according to other projects, such as you said `The main idea comes from "opentelemetry-java-instrumentation".`. We focus on the product and tech parts.
> > 
> > 
> > Ok. I have an ambiguity,sorry.
> > In our company, skywalking is a very new concept, so I want to minimize the cost of access.
> > I can now allow users to click a button to access apm. This is pretty cool, so I am not going to change this approach.
> > so the [log4j2&logback-automatic-agent] can do this. only one button. (To be precise, you also need to add %X(traceid), but this will not change the POM)
> > As for opentelemetry-java-instrumentation, I just borrowed their cut points. (org.apache.logging.log4j.core.impl.ContextDataInjectorFactory.createInjector and ch.qos.logback.classic.Logger)
> > The framework and main logic are based on the definition of skywalking.
> 
> OK, I can get your points now. But we are very careful to provide a plugin like this. Actually we discussed about this from the day at beginning of the toolkit you see right now.
> Once we have an auto instrumentation log injection, we face a risk to break user's log analysis system, such as prod env's ELK. Because we inject something breaking the existing pattern, but users would not expect this, as our agent is out of box, people are new to the project.
> Also, we will have to inject all the logs, rather than the necessary logs which cause unnecessary costs.
> 
> I am not sure whether these make sense to you, but those are the reasons we don't have this today.

Thanks!! Because I design the logger and the trace(the apm haha), so I want to be able to automate the full link. I will design it carefully as you suggest.

I think if the trace can not be related to the log, the apm is not complete.

wu-sheng:
I know, but that is users call, we even provided log reporter to help log integration.

klmzfyf:
> I know, but that is users call, we even provided log reporter to help log integration.

okay.... Maybe I dote on the user too much.

klmzfyf:
> I know, but that is users call, we even provided log reporter to help log integration.

But what I really want to do is to report all error trace logs to skywalking, not just exception logs.（such as timeout log）

I think the [auto log] can help me.

wu-sheng:
> > I know, but that is users call, we even provided log reporter to help log integration.
> 
> But what I really want to do is to report all logs of error trace to skywalking, not just exception logs.（such as timeout log）
> 
> I think the [auto log] can help me.

If you really want this to be upstream, there is a tradeoff you could do. We could accept this log plugin in the optional plugin list, which means users could activate it by themselves, any side effect becomes users' responsibility rather than ours.

wu-sheng:
Any update? If an optional plugin isn't your want, I am going to close this PR.

PR #7229, opened 2021-07-02
Fix issue that webfluxwebclient plugin async finish repeatedly in multi thread

Lin1997:
     The `ExchangeFunctions$DefaultExchangeFunction` is stateless and  thread-safe, whose `exchange` may be called in different thread concurrently. However, the `spring-webflux-5.x-webclient-plugin` use `setSkyWalkingDynamicField` to  pass the span in `WebFluxWebClientInterceptor`, which broke the stateless design. Thus when the Webclient request concurrently, the execution times of `prepareForAsync` and `asyncFinish()` may mismatch.
     ~~Use `ThreadLocal` to pass the span can fix this bug easily.~~
     Use MIC in v2 API to propagate the span can fix this bug easily.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/spring-plugins/spring-webflux-5.x-webclient-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/webclient/WebFluxWebClientInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/spring-plugins/spring-webflux-5.x-webclient-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/webclient/WebFluxWebClientInterceptor.java
@@ -24,8 +24,8 @@ import org.apache.skywalking.apm.agent.core.context.tag.Tags;
 import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
 import org.apache.skywalking.apm.agent.core.context.trace.SpanLayer;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
-import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
-import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.v2.InstanceMethodsAroundInterceptorV2;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.v2.MethodInvocationContext;
 import org.apache.skywalking.apm.network.trace.component.ComponentsDefine;
 import org.springframework.http.HttpStatus;
 import org.springframework.web.reactive.function.client.ClientRequest;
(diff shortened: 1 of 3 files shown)


apm-sniffer/apm-sdk-plugin/spring-plugins/spring-webflux-5.x-webclient-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/webclient/WebFluxWebClientInterceptor.java
```diff
@@ -78,7 +80,8 @@ public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allA
             return ret;
         }
         Mono<ClientResponse> ret1 = (Mono<ClientResponse>) ret;
-        AbstractSpan span = (AbstractSpan) objInst.getSkyWalkingDynamicField();
+        AbstractSpan span = spanLocal.get();
+        spanLocal.remove();
```
wu-sheng, line 84: We should have a try/finally to make sure this gets removed.

wu-sheng:
I think we have several open issues about this, could you search them, and link from here?

apm-sniffer/apm-sdk-plugin/spring-plugins/spring-webflux-5.x-webclient-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/webclient/WebFluxWebClientInterceptor.java
```diff
@@ -63,12 +65,12 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
...
             ((EnhancedInstance) request).setSkyWalkingDynamicField(contextCarrier);
         }
-        
+
         //user async interface
         span.prepareForAsync();
         ContextManager.stopSpan();
 
-        objInst.setSkyWalkingDynamicField(span);
+        spanLocal.set(span);
```
wu-sheng, line 73: Check plugin API, we have MIC to propagate this reference to after stage, and 100% safe, also have better performance.

Lin1997: Please check the new commit.

Lin1997:
> I think we have several open issues about this, could you search them, and link from here?

I only found this one:
https://github.com/apache/skywalking/issues/5814#issuecomment-767271129

wu-sheng (commented):
And changes.md in the root folder.

wu-sheng:
Running CI now.

wu-sheng (approved):
Glad to see v2 API begins to make things easier.

wu-sheng:
cc @dmsolr Another easy fix after we have v2 APIs.

PR #7243, opened 2021-07-05
Add some new thread metric and class metric to JVMMetric (#7230)

### LINKED ISSUE #7230: Add some new thread metric and class metric to JVMMetric

___
## Requirement or improvement

Add some new thread metric and class metric to JVMMetric.

### Protocal

```protobuf
message JVMMetric {
    int64 time = 1;
    CPU cpu = 2;
    repeated Memory memory = 3;
    repeated MemoryPool memoryPool = 4;
    repeated GC gc = 5;
    Thread thread = 6;
    Class clazz = 7;
}

message Thread {
    int64 liveCount = 1;
    int64 daemonCount = 2;
    int64 peakCount = 3;
    // new metric
    int64 newStateThreadCount = 4;
    int64 runnableStateThreadCount = 5;
    int64 blockedStateThreadCount = 6;
    int64 waitingStateThreadCount = 7;
    int64 timedWaitingStateThreadCount = 8;
    int64 terminatedStateThreadCount = 9;
}

// new metric
message Class {
    int64 loadedClassCount = 1;
    int64 totalUnloadedClassCount = 2;
    int64 totalLoadedClassCount = 3;
}
```

### The result in skywalking ui
image
image

Switch-vov:
### <Feature description>
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7230 .


Switch-vov:
Associate #7230 and https://github.com/apache/skywalking-data-collect-protocol/pull/52

wu-sheng:
`changes.md` in the root should be updated.

wu-sheng:
@mrproliu I think we will need to update satellite and goapi for the latest JVM protocol.

wu-sheng:
At least add one of each kind metrics into e2e testing, we need an automatic process to verify your changes are working.

Switch-vov:
> At least add one of each kind metrics into e2e testing, we need an automatic process to verify your changes are working.

Ok, I try write an e2e testcase.

@wu-sheng Is the tutorial [End to End Tests (E2E)](https://skywalking.apache.org/docs/main/latest/en/guides/readme/#end-to-end-tests-e2e)?

wu-sheng:
> > At least add one of each kind metrics into e2e testing, we need an automatic process to verify your changes are working.
> 
> Ok, I try write some e2e testcase.
> 
> @wu-sheng Is the tutorial [End to End Tests (E2E)](https://skywalking.apache.org/docs/main/latest/en/guides/readme/#end-to-end-tests-e2e)?

You don't have to try that, which is the next round. PMC would take care of this. The current is more Java tech stack.
1. Doc, http://skywalking.apache.org/docs/main/latest/en/guides/readme/#end-to-end-tests-e2e
2. Codebase, https://github.com/apache/skywalking/tree/master/test/e2e
3. How we run e2e through GitHub Action, https://github.com/apache/skywalking/tree/master/.github/workflows

wu-sheng:
The basic processes of e2e are
1. Build skywalking
2. Build demo app
3. Boot (1) and (2)
4. Generate traffic for demo app
5. Verify the metrics through GraphQL.

wu-sheng:
Follow the GitHub Action control file, it should not hard to understand, if you have known maven, unit test, and docker/docker-compose

Switch-vov:
> Follow the GitHub Action control file, it should not hard to understand, if you have known maven, unit test, and docker/docker-compose

I think this isn't hard, but i need some time learning it.
I will write some e2e test case in the week.

wu-sheng:
> > Follow the GitHub Action control file, it should not hard to understand, if you have known maven, unit test, and docker/docker-compose
> 
> I think this isn't hard, but i need some time learning it.
> I will write some e2e test case in the week.

Oh, you don't need to write a new one. There are many existing cases which is verifying JVM-related metrics. You just need to add one metric about thread, the other about class loaded to verify the codes.
Because code reviewers can't find runtime issues.

Switch-vov:
> > > Follow the GitHub Action control file, it should not hard to understand, if you have known maven, unit test, and docker/docker-compose
> > 
> > 
> > I think this isn't hard, but i need some time learning it.
> > I will write some e2e test case in the week.
> 
> Oh, you don't need to write a new one. There are many existing cases which is verifying JVM-related metrics. You just need to add one metric about thread, the other about class loaded to verify the codes.
> Because code reviewers can't find runtime issues.

ok, I get it.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/thread/ThreadProvider.java
```diff
@@ -31,12 +32,58 @@
...
+                            break;
+                        case TERMINATED:
+                            terminatedStateThreadCount++;
+                            break;
+                        default:
+                            break;
+                    }
+                } else {
+                    terminatedStateThreadCount++;
+                }
```
Ax1an, line 70: The code as a whole looks good. However, there may be some inaccuracies in this place. I found that `getAllThreadIds()` and `getThreadInfo` method both return information about the active thread. However, when the thread is in the NEW and TERMINATED state, the `isAlive()` method will return false, that is, they are inactive. Therefore, in this way, we cannot correctly count the number of threads in the NEW and TERMINATED state.

Ax1an: So we need to consider whether we want to remove them.

Switch-vov: > So we need to consider whether we want to remove them.

`New State` is zero always, but `Terminated State` is non-zero sometimes.

So, I should remove them?

wu-sheng: If there is no threadInfo, I think we should not run statistics for it.

Ax1an: If you remove the counting logic from the else code block, you may find that the `Terminated State` is zero always.

Switch-vov: > If you remove the counting logic from the else code block, you may find that the `Terminated State` is zero always.

Of course.

So, I remove `Terminated State` and `New State` later.

Switch-vov:
> > > Follow the GitHub Action control file, it should not hard to understand, if you have known maven, unit test, and docker/docker-compose
> > 
> > 
> > I think this isn't hard, but i need some time learning it.
> > I will write some e2e test case in the week.
> 
> Oh, you don't need to write a new one. There are many existing cases which is verifying JVM-related metrics. You just need to add one metric about thread, the other about class loaded to verify the codes.
> Because code reviewers can't find runtime issues.

I add some metrics in [MetricsQuery.java](https://github.com/apache/skywalking/pull/7243/files#diff-a0a708009817f87fec4c8cc48708ecb6185015cabf9d356e645e96e5fffa7370).

docs/en/concepts-and-designs/scope-definitions.md
```diff
@@ -104,9 +104,25 @@ This calculates the metrics data if the service instance is a JVM and collects t
...
-| peakCount | The current number of peak threads. | | int |
+| liveCount | The current number of live threads. | | long |
+| daemonCount | The current number of daemon threads. | | long |
+| peakCount | The current number of peak threads. | | long |
+| newStateThreadCount | The current number of threads in new state. | | long |
+| runnableStateThreadCount | The current number of threads in runnable state. | | long |
+| blockedStateThreadCount | The current number of threads in blocked state. | | long |
+| waitingStateThreadCount | The current number of threads in waiting state. | | long |
+| timedWaitingStateThreadCount | The current number of threads in time-waiting state. | | long |
+| terminatedStateThreadCount | The current number of threads in terminated state. | | long |
```
Ax1an, line 115:
```suggestion
| runnableStateThreadCount | The current number of threads in runnable state. | | long |
| blockedStateThreadCount | The current number of threads in blocked state. | | long |
| waitingStateThreadCount | The current number of threads in waiting state. | | long |
| timedWaitingStateThreadCount | The current number of threads in time-waiting state. | | long |
```

Ax1an:
Should we remove them from the protocol as well?

wu-sheng:
> Should we remove them from the protocol as well?

Yes, I think so.

Switch-vov:
> > Should we remove them from the protocol as well?
> 
> Yes, I think so.

So, need I remove them?

Ax1an:
> > > Should we remove them from the protocol as well?
> > 
> > 
> > Yes, I think so.
> 
> So, need I remove them?

Yes.

wu-sheng:
> > > Should we remove them from the protocol as well?
> > 
> > 
> > Yes, I think so.
> 
> So, need I remove them?

If you are not going to report them, then we remove them. The protocol should include necessary fields

Switch-vov:
> > > > Should we remove them from the protocol as well?
> > > 
> > > 
> > > Yes, I think so.
> > 
> > 
> > So, need I remove them?
> 
> If you are not going to report them, then we remove them. The protocol should include necessary fields

Ok, I will create PR to remove them.

wu-sheng:
I am not sure why `E2E / Compatibility (pull_request) ` fails, please check the logs.

Ax1an:
@Switch-vov The following information is for your reference.

image

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/jvm/JVMSourceDispatcher.java
```diff
@@ -193,7 +197,29 @@ private void sendToThreadMetricProcess(String service,
         serviceInstanceJVMThread.setLiveCount(thread.getLiveCount());
         serviceInstanceJVMThread.setDaemonCount(thread.getDaemonCount());
         serviceInstanceJVMThread.setPeakCount(thread.getPeakCount());
+        serviceInstanceJVMThread.setRunnableStateThreadCount(thread.getRunnableStateThreadCount());
+        serviceInstanceJVMThread.setBlockedStateThreadCount(thread.getBlockedStateThreadCount());
+        serviceInstanceJVMThread.setWaitingStateThreadCount(thread.getWaitingStateThreadCount());
+        serviceInstanceJVMThread.setTimedWaitingStateThreadCount(thread.getTimedWaitingStateThreadCount());
```
wu-sheng, line 203: I think this would break compatibility for all previous java agent. We need to avoid this.

wu-sheng: You may use `hasField` to determine the version.

Switch-vov: > You may use `hasField` to determine the version.

What means `hasField`?
Read the log, I found the first metrics `Total Class Loader` is zero always.
So, What should I do?

wu-sheng: The compatibility test is not using the latest agent, it checks the compatibility with the previous release. I remember it uses 8.2 or 8.1 release agent. In that agent, you don't have these metrics. You should make that e2e test verifier doesn't check these added metrics.

My point was, you may need to check whether has errors here because the data OAP received in this case, they even don't report all these fields. Now, I think you could ignore my comments about `You may use hasField to determine the version.`, protobuf should have covered this case, and focus on different metrics verification in different test cases.

wu-sheng: @kezhenxu94 Do we have a chance to verify different metrics in the compatible case?

kezhenxu94: Yes. @Switch-vov don’t share compatibility test class with simple case, copy `SimpleE2E` class for compatibility (`ComparE2E`) and replace https://github.com/apache/skywalking/blob/d15b67bd97cfe8386df0c61198d9f1ca2060089b/.github/workflows/e2e.compat.yaml#L61 with the `CompatE2E` class. Then in `CompatE2E` don’t verify those new metrics.

wu-sheng: @Switch-vov Please follow ☝️

Switch-vov: OK

Switch-vov: > @Switch-vov Please follow ☝️

I'm try to refactor SimpleE2E, CompatE2E and KafkaE2E.

wu-sheng:
Before run CI, @kezhenxu94 Could you recheck the CI codes change? Are they expected? I can see inheritance level change happens.

test/e2e/e2e-test/src/test/java/org/apache/skywalking/e2e/simple/SimpleE2EBase.java
```diff
@@ -0,0 +1,202 @@
...
+        );
+    }
+
+    protected void verifyServiceRelationMetrics(final List<Call> calls) throws Exception {
+        verifyRelationMetrics(calls, ALL_SERVICE_RELATION_CLIENT_METRICS, ALL_SERVICE_RELATION_SERVER_METRICS);
+    }
+
+    protected void verifyRelationMetrics(final List<Call> calls,
+                                       final String[] relationClientMetrics,
+                                       final String[] relationServerMetrics) throws Exception {
```
kezhenxu94, line 182:
```suggestion
    protected void verifyRelationMetrics(final List<Call> calls,
                                         final String[] relationClientMetrics,
                                         final String[] relationServerMetrics) throws Exception {
```

kezhenxu94 (commented):
> Before run CI, @kezhenxu94 Could you recheck the CI codes change? Are they expected? I can see inheritance level change happens.

The changes look good to me. Some shared codes are extracted into a class and others extend from it.

wu-sheng:
@dmsolr @EvanLjp @mrproliu @Ax1an Could you review the codes? I can see CIs passed.

wu-sheng (approved):
Generally good to me, wait for others' review. @Switch-vov Could you post a system screenshot to see what does the latest UI look like?

EvanLjp (approved):
LGTM

Switch-vov:
> Generally good to me, wait for others' review. @Switch-vov Could you post a system screenshot to see what does the latest UI look like?

OK，That's all screenshot. @wu-sheng 

image
image

Ax1an (approved):
LGTM.

mrproliu (approved):
LGTM

PR #7249, opened 2021-07-05
Allow configuring max request header size

sergicastro:

This allows configuring the HTTP max request header size from the jetty server.
By default it uses 8192, the same jetty default.

Adding context:
I have a proxy in front of the OAP server that manages authentication.
I'm using OIDC with Azure which provides large tokens.
The proxy forwards requests to the OAP server without removing any header. 
Then I hit HTTP code `431 - Request Header Fields Too Large`.

I found jetty allowing customization for this limit so I brought this customization to the OAP server.

### DIFF
--- a/docs/en/setup/backend/configuration-vocabulary.md
+++ b/docs/en/setup/backend/configuration-vocabulary.md
@@ -12,6 +12,7 @@ core|default|role|Option values, `Mixed/Receiver/Aggregator`. **Receiver** mode 
 | - | - | restIdleTimeOut| Connector idle timeout in milliseconds of restful service| SW_CORE_REST_JETTY_IDLE_TIMEOUT|30000|
 | - | - | restAcceptorPriorityDelta| Thread priority delta to give to acceptor threads of restful service| SW_CORE_REST_JETTY_DELTA|0|
 | - | - | restAcceptQueueSize| ServerSocketChannel backlog  of restful service| SW_CORE_REST_JETTY_QUEUE_SIZE|0|
+| - | - | httpMaxRequestHeaderSize| Maximum request header size accepted| SW_CORE_HTTP_MAX_REQUEST_HEADER_SIZE|8192|
 | - | - | gRPCHost|Binding IP of gRPC service. Services include gRPC data report and internal communication among OAP nodes|SW_CORE_GRPC_HOST|0.0.0.0|
 | - | - | gRPCPort| Binding port of gRPC service | SW_CORE_GRPC_PORT|11800|
 | - | - | gRPCSslEnabled| Activate SSL for gRPC service | SW_CORE_GRPC_SSL_ENABLED|false|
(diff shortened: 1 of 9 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleConfig.java
```diff
@@ -182,6 +182,12 @@
     @Setter
     private boolean enableEndpointNameGroupingByOpenapi = true;
 
+    /**
+     * The maximum size in bytes allowed for request headers.
+     * Use -1 to disable it.
+     */
+    private int httpMaxRequestHeaderSize = 8192;
```
wu-sheng, line 189: Is this the default value of Jetty?

wu-sheng (commented):
This doc should be updated, https://github.com/apache/skywalking/blob/master/docs/en/setup/backend/configuration-vocabulary.md

wu-sheng:
> This doc should be updated, https://github.com/apache/skywalking/blob/master/docs/en/setup/backend/configuration-vocabulary.md

To update this, you also should expose the configurations(core's) and define the system env name here, https://github.com/apache/skywalking/blob/master/oap-server/server-bootstrap/src/main/resources/application.yml.

Also, you should support this config here, https://github.com/apache/skywalking/blob/master/oap-server/server-receiver-plugin/skywalking-sharing-server-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/sharing/server/SharingServerConfig.java. 

SkyWalking has 2 servers, core level and sharing level servers. Sharing server is more widely used to expose for outside.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleConfig.java
sergicastro: yes, from what I saw, it configures it in many places like this:
https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-server/src/main/config/etc/jetty.xml#L65

sergicastro: Also, this is the value the oap-server logs show when I hit the issue, so defaulting to this one should cause no changes when upgrading an environment.

docs/en/setup/backend/configuration-vocabulary.md
```diff
@@ -12,6 +12,7 @@ core|default|role|Option values, `Mixed/Receiver/Aggregator`. **Receiver** mode
 | - | - | restIdleTimeOut| Connector idle timeout in milliseconds of restful service| SW_CORE_REST_JETTY_IDLE_TIMEOUT|30000|
 | - | - | restAcceptorPriorityDelta| Thread priority delta to give to acceptor threads of restful service| SW_CORE_REST_JETTY_DELTA|0|
 | - | - | restAcceptQueueSize| ServerSocketChannel backlog  of restful service| SW_CORE_REST_JETTY_QUEUE_SIZE|0|
+| - | - | httpMaxRequestHeaderSize| Maximum request header size accepted| SW_CORE_HTTP_MAX_REQUEST_HEADER_SIZE|8192|
```
wu-sheng, line 15: You miss the sharing server config in here.

oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -296,6 +297,7 @@ receiver-sharing-server:
     restIdleTimeOut: ${SW_RECEIVER_SHARING_JETTY_IDLE_TIMEOUT:30000}
     restAcceptorPriorityDelta: ${SW_RECEIVER_SHARING_JETTY_DELTA:0}
     restAcceptQueueSize: ${SW_RECEIVER_SHARING_JETTY_QUEUE_SIZE:0}
+    httpMaxRequestHeaderSize: ${SW_CORE_HTTP_MAX_REQUEST_HEADER_SIZE:8192}
```
wu-sheng, line 300:
```suggestion
    httpMaxRequestHeaderSize: ${SW_RECEIVER_SHARING_HTTP_MAX_REQUEST_HEADER_SIZE:8192}
```

sergicastro:
Thanks for the review @wu-sheng, it should be completed now.

kezhenxu94 (approved):
Thanks @sergicastro

wu-sheng (approved):

Issue #6157, opened 2021-01-08
Failed to connect the Agent to the Kafka authenticated by SASL.


JHub-Wei:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [*] Question or discussion
- [ ] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?
When SASL authentication is enabled for Kafka, an error is reported during agent access,here are error messages:
ERROR 2021-01-08 06:30:52:951 main ServiceManager : ServiceManager try to pre-start [org.apache.skywalking.apm.agent.core.kafka.KafkaProducerManager] fail.
org.apache.skywalking.apm.dependencies.org.apache.kafka.common.KafkaException: Failed to create new KafkaAdminClient
        at org.apache.skywalking.apm.dependencies.org.apache.kafka.clients.admin.KafkaAdminClient.createInternal(KafkaAdminClient.java:451)
        at org.apache.skywalking.apm.dependencies.org.apache.kafka.clients.admin.Admin.create(Admin.java:59)
        at org.apache.skywalking.apm.dependencies.org.apache.kafka.clients.admin.AdminClient.create(AdminClient.java:39)
        at org.apache.skywalking.apm.agent.core.kafka.KafkaProducerManager.prepare(KafkaProducerManager.java:58)
        at org.apache.skywalking.apm.agent.core.boot.ServiceManager.prepare(ServiceManager.java:104)
        at org.apache.skywalking.apm.agent.core.boot.ServiceManager.boot(ServiceManager.java:43)
        at org.apache.skywalking.apm.agent.SkyWalkingAgent.premain(SkyWalkingAgent.java:130)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at sun.instrument.InstrumentationImpl.loadClassAndStartAgent(InstrumentationImpl.java:386)
        at sun.instrument.InstrumentationImpl.loadClassAndCallPremain(InstrumentationImpl.java:401)
[...]

PR #7270, opened 2021-07-09
set context class loader to AgentClassLoader

DuanYuePeng:
javax.security.auth.login.LoginContext use context class loader to load the login module, when configuring "sasl.jaas.config", we could set the value as `org.apache.skywalking.apm.dependencies.org.apache.kafka.common.security.scram.ScramLoginModule required username="user" password="pass";`, and the context class loader would load the class correctly, otherwise, NoClassFoundException would be thrown

### Fix <https://github.com/apache/skywalking/issues/6157>


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/loader/AgentClassLoader.java
```diff
@@ -77,6 +77,7 @@ public static void initDefaultLoader() throws AgentPackageNotFoundException {
             synchronized (AgentClassLoader.class) {
                 if (DEFAULT_LOADER == null) {
                     DEFAULT_LOADER = new AgentClassLoader(PluginBootstrap.class.getClassLoader());
+                    Thread.currentThread().setContextClassLoader(DEFAULT_LOADER);
```
wu-sheng, line 80: This is only work for current thread, why this effects the Kafka kernel? How they share the same thread?

DuanYuePeng:
This bug happens in `KafkaProducerManager::prepare` method, which is a `BootService`. It works in the same thread with `SkyWalkingAgent::premain`.
```
AdminClient adminClient = AdminClient.create(properties);
```
Just see the Thread dump.
```
init:278, LoginContext (javax.security.auth.login)
<init>:512, LoginContext (javax.security.auth.login)
login:59, AbstractLogin (org.apache.skywalking.apm.dependencies.org.apache.kafka.common.security.authenticator)
<init>:62, LoginManager (org.apache.skywalking.apm.dependencies.org.apache.kafka.common.security.authenticator)
acquireLoginManager:105, LoginManager (org.apache.skywalking.apm.dependencies.org.apache.kafka.common.security.authenticator)
configure:147, SaslChannelBuilder (org.apache.skywalking.apm.dependencies.org.apache.kafka.common.network)
create:146, ChannelBuilders (org.apache.skywalking.apm.dependencies.org.apache.kafka.common.network)
clientChannelBuilder:67, ChannelBuilders (org.apache.skywalking.apm.dependencies.org.apache.kafka.common.network)
createChannelBuilder:99, ClientUtils (org.apache.skywalking.apm.dependencies.org.apache.kafka.clients)
createInternal:426, KafkaAdminClient (org.apache.skywalking.apm.dependencies.org.apache.kafka.clients.admin)
create:59, Admin (org.apache.skywalking.apm.dependencies.org.apache.kafka.clients.admin)
create:39, AdminClient (org.apache.skywalking.apm.dependencies.org.apache.kafka.clients.admin)
prepare:58, KafkaProducerManager (org.apache.skywalking.apm.agent.core.kafka)
lambda$prepare$1:105, ServiceManager (org.apache.skywalking.apm.agent.core.boot)
accept:-1, 783191662 (org.apache.skywalking.apm.agent.core.boot.ServiceManager$$Lambda$6)
accept:183, ForEachOps$ForEachOp$OfRef (java.util.stream)
end:357, SortedOps$SizedRefSortingSink (java.util.stream)
copyInto:483, AbstractPipeline (java.util.stream)
wrapAndCopyInto:472, AbstractPipeline (java.util.stream)
evaluateSequential:150, ForEachOps$ForEachOp (java.util.stream)
evaluateSequential:173, ForEachOps$ForEachOp$OfRef (java.util.stream)
evaluate:234, AbstractPipeline (java.util.stream)
forEach:485, ReferencePipeline (java.util.stream)
prepare:103, ServiceManager (org.apache.skywalking.apm.agent.core.boot)
boot:44, ServiceManager (org.apache.skywalking.apm.agent.core.boot)
premain:133, SkyWalkingAgent (org.apache.skywalking.apm.agent)
invoke0:-1, NativeMethodAccessorImpl (sun.reflect)
invoke:62, NativeMethodAccessorImpl (sun.reflect)
invoke:43, DelegatingMethodAccessorImpl (sun.reflect)
invoke:498, Method (java.lang.reflect)
loadClassAndStartAgent:386, InstrumentationImpl (sun.instrument)
loadClassAndCallPremain:401, InstrumentationImpl (sun.instrument)
```

wu-sheng:
> This bug happens in KafkaProducerManager::prepare method, which is a BootService. It works in the same thread with SkyWalkingAgent::premain.

Then, I read the codes, then this is not premain thread, this is a thread created through `Executors.newSingleThreadScheduledExecutor`. You should not manipulate the premain thread class loader, instead, please use this inside `KafkaProducerManager#run` when it runs in the first time, which is much safer and only effects a thread we created inside the agent.

Does this make sense for you? Could you try locally?

wu-sheng:
And you should update the `changes.md` in the root.

wu-sheng:
You seems to change a lot of unexpected things. Please recheck your local git.

DuanYuePeng:
The git is broken, I'll start a new PR.

PR #7271, opened 2021-07-09
fix: Fix MultiScopesAnalysisListener#setPublicAttrs log warn frequently

### LINKED ISSUE #6562: MultiScopesAnalysisListener#setPublicAttrs log warn   frequently

- [ ] Question or discussion
- [x] Bug
- [ ] Requirement

___
### Bug
skywalking: 8.3.0
OS: centos 6.7
JRE : 1.8

- What happened?
MultiScopesAnalysisListener#setPublicAttrs log warn   frequently

```
private void setPublicAttrs(SourceBuilder sourceBuilder, SpanObject span) {
       xxxxxxx
        span.getTagsList().forEach(tag -> {
            if (SpanTags.STATUS_CODE.equals(tag.getKey())) {
                try {
                    sourceBuilder.setResponseCode(Integer.parseInt(tag.getValue()));
                } catch (NumberFormatException e) {
                    log.warn("span {} has illegal status code {}", span, tag.getValue());
                }
            }
            sourceBuilder.setTag(tag);
        });
        xxxxxxx
}
```

As above code ,backend only accept status_code as Integer type, but some agent plugins  sent a String .
Because of STATUS_CODE defined  by  Tags#STATUS_CODE in apm agent-core is a StringTag, Most of agent plugin will use Integer.toString to guarantee it correct. but some plugin dosenot obey this rules，such as grpc_v1 、rocketmq_3.x、rocketmq_4.x 。

Suggestion:
create the IntTag class as subclass AbstractTag of  for STATUS_CODE , fix all agent plugins.

wallezhang:
### Fix <#6562>

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6562.

How to fix:

1. Modify `org.apache.skywalking.apm.agent.core.context.tag.Tags#STATUS_CODE` static field type from `StringTag` to `IntegerTag` and all plugins that used the static field.
2. Add `org.apache.skywalking.apm.agent.core.context.tag.Tags#RESPONSE_CODE` static field to hold response code value except http calls. And this tag is only used in grpc-1.x-plugin to replace status code tag for now.
3. Oap server analyzer source add `responseCodeStr` field associated with response code tag. I don't think the same span will have both status code tag and response code tag, so if this span is a http call, `responseCodeStr` field value is same as responseCode field except the field type.


kezhenxu94:
I'm kind of concerned that `RESPONSE_CODE` and `STATUS_CODE` may be somewhat ambiguous as they have very similar names but no indicators to say what scenario they are specifically aimed at, I can only see "http and non-http" in the code comments. Is it a good idea to have these 2 tags simultaneously? @wu-sheng WDYT?

wu-sheng:
A plugin test fails, please recheck, https://github.com/apache/skywalking/pull/7271/checks?check_run_id=3027165338

wallezhang:
> I'm kind of concerned that `RESPONSE_CODE` and `STATUS_CODE` may be somewhat ambiguous as they have very similar names but no indicators to say what scenario they are specifically aimed at, I can only see "http and non-http" in the code comments. Is it a good idea to have these 2 tags simultaneously? @wu-sheng WDYT?

Status code is defined in http protocol, and I think this tag is just for recording http status code. Other non-http requests also have their own response code definitions. So I think reusing status code tag may not be a good idea.

wu-sheng:
Personally, I can feel this confusion. But 2 fields are really necessary due to HTTP code is an integer, could use `>` and `<`. But others are majorly string, `==` and `include` are the options.

@kezhenxu94 What do you recommend to change?

wu-sheng:
@wallezhang Ha, I noticed this is on the summer 2021 task. 

@wallezhang @kezhenxu94 Could we have a decision, and let the student continues from here?

kezhenxu94:
> Personally, I can feel this confusion. But 2 fields are really necessary due to HTTP code is an integer, could use `>` and `<`. But others are majorly string, `==` and `include` are the options.
> 
> @kezhenxu94 What do you recommend to change?

Ok I missed that. I think it makes sense to have these 2 fields now.

kezhenxu94:
> @wallezhang Ha, I noticed this is on the summer 2021 task.
> 
> @wallezhang @kezhenxu94 Could we have a decision, and let the student continues from here?

I’m independent. You can make the call. 😄

wallezhang:
> @wallezhang Ha, I noticed this is on the summer 2021 task.
> 
> @wallezhang @kezhenxu94 Could we have a decision, and let the student continues from here?

It's all right, you can make the decision 😄

wu-sheng:
OK, I prefer to use STATUS_CODE in agent tag and OAL, which refers from gRPC, https://grpc.github.io/grpc/core/md_doc_statuscodes.html

Let's hold this PR for a little longer, to see whether the Summer 2021 student could follow up. @wallezhang Is this working for you? Sorry I didn't notice this at the beginning.

wallezhang:
> OK, I prefer to use STATUS_CODE in agent tag and OAL, which refers from gRPC, https://grpc.github.io/grpc/core/md_doc_statuscodes.html
> 
> Let's hold this PR for a little longer, to see whether the Summer 2021 student could follow up. @wallezhang Is this working for you? Sorry I didn't notice this at the beginning.

It's OK. It doesn't affect me 😄

wu-sheng:
The expect changes are as following
1. Rename current `STATUS_CODE` to `HTTP_RESPONSE_STATUS_CODE`, real tag key should be `http.status_code`.
2. Add a new string tag `RPC_RESPONSE_STATUS_CODE` to hold all string type value tag. The key is `rpc.status_code`.
3. At the backend source level, we should keep `responseCode` for `HTTP_RESPONSE_STATUS_CODE` tag, but add `@Deprecated`. Also, add `httpResponseStatusCode`(a duplicate of `responseCode`) and `rpcStatusCode`(for `RPC_RESPONSE_STATUS_CODE`) fields for OAL.
4. Document should be updated accordingly.

wu-sheng:
@wallezhang Hi, if you are still watching this, let's make this back to work. The student of Summer 2021 seems having no activity. Most likely, I am going to terminate his task.

wallezhang:
> @wallezhang Hi, if you are still watching this, let's make this back to work. The student of Summer 2021 seems having no activity. Most likely, I am going to terminate his task.

OK, I will continue do this work.

wu-sheng:
> > @wallezhang Hi, if you are still watching this, let's make this back to work. The student of Summer 2021 seems having no activity. Most likely, I am going to terminate his task.
> 
> OK, I will continue do this work.

Sorry to change the status around repeatedly.

wu-sheng:
`skywalking-ui` is a mischange, need sync to master.

test/plugin/scenarios/gateway-2.0.x-scenario/config/expectedData.yaml
```diff
@@ -57,7 +57,7 @@ segmentItems:
       tags:
       - {key: url, value: 'http://localhost:8080/provider/b/testcase'}
       - {key: http.method, value: GET}
-      - {key: status_code, value: '200'}
+      - {key: http.status_code, value: '200'}
```
wu-sheng, line 60: FYI @JaredTan95 After this change, Resin plugins in https://github.com/SkyAPM/java-plugin-extensions should break. We need a fix after 8.8.0 release, and release 3.0.0. Could you create an issue there to track this?

wallezhang:
@wu-sheng I think all the works have been completed. Please help to check whether there are any missing modifications.

wu-sheng:
I will check tonight, very appreciated.

dmsolr (dismissed):
LGTM

wu-sheng (changes requested):
Some plugin tests are still failing, please recheck.

wallezhang:
> Some plugin tests are still failing, please recheck.

Sorry, there is a mistake in grpc-scenario expectedData.yaml.

wu-sheng (approved):

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/MultiScopesAnalysisListener.java
```diff
@@ -240,12 +240,15 @@ private void setPublicAttrs(SourceBuilder sourceBuilder, SpanObject span) {
         sourceBuilder.setLatency((int) latency);
         sourceBuilder.setResponseCode(Const.NONE);
         span.getTagsList().forEach(tag -> {
-            if (SpanTags.STATUS_CODE.equals(tag.getKey())) {
+            if (SpanTags.HTTP_RESPONSE_STATUS_CODE.equals(tag.getKey())) {
```
wu-sheng, line 243: I think we should check the old tag too? Unless, OAL engine is not compatible with old agent, right? But we should indicate it is `Deprecated` through comment and make the logic clear through comments.

wu-sheng (changes requested):
One more change required.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/MultiScopesAnalysisListener.java
wallezhang: Yes, it's reasonable. So I should keep the old tag `SpanTags.STATUS_CODE` and explain in comment that it has been deprecated.

wallezhang: I have another question, if application use the new agent, and the tag will be `http.status_code` or `rpc.status_code`, while the old OAL server unrecognized the new tag, so the `responseCode` field will always be 0. Therefore, does agent also need to consider compatibility issues?

wu-sheng: The agent changes should be applied. And OAP would work with both(deprecated and new)

wu-sheng: We don't need `new agent`->`old OAP` works. We never recommended that.

wallezhang: OK, got it

wu-sheng: You could add a change log at the project section to highlight this change, and the new agent plugins' response code would not be compatible with old OAP versions.

wu-sheng (approved):
Thanks.

PR #7274, opened 2021-07-09
Add pagination to event query

kezhenxu94:

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/type/Pagination.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/type/Pagination.java
@@ -18,11 +18,15 @@
 
 package org.apache.skywalking.oap.server.core.query.type;
 
+import lombok.AllArgsConstructor;
 import lombok.Getter;
+import lombok.NoArgsConstructor;
 import lombok.Setter;
 
 @Getter
 @Setter
+@NoArgsConstructor
+@AllArgsConstructor
 public class Pagination {
     private int pageNum;
     private int pageSize;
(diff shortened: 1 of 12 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/type/Pagination.java
```diff
@@ -18,11 +18,15 @@
...
 
+import lombok.AllArgsConstructor;
 import lombok.Getter;
+import lombok.NoArgsConstructor;
 import lombok.Setter;
 
 @Getter
 @Setter
+@NoArgsConstructor
+@AllArgsConstructor
```
wu-sheng, line 29: Who is using `AllArgsConstructor`?

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/AlarmQuery.java
```diff
@@ -88,7 +88,9 @@ public Alarms getAlarm(final Duration duration, final Scope scope, final String
         }
         long startSecondTB = 0;
         long endSecondTB = 0;
-        final EventQueryCondition.EventQueryConditionBuilder conditionPrototype = EventQueryCondition.builder().size(IEventQueryDAO.MAX_SIZE);
+        final EventQueryCondition.EventQueryConditionBuilder conditionPrototype =
+            EventQueryCondition.builder()
+                               .paging(new Pagination(1, IEventQueryDAO.MAX_SIZE, false));
```
kezhenxu94, line 93: > Who is using AllArgsConstructor?

Here

wu-sheng (approved):

wu-sheng:
> I can't see GraphQL submodule update.

e2e still pass because no event related test cases, right?

wu-sheng (changes requested):
I can't see GraphQL submodule update.

wu-sheng (commented):
Do we need e2e case update? I think this is a new condition for client, right?

kezhenxu94:
> Do we need e2e case update? I think this is a new condition for client, right?

E2E client didn't set `size` parameter, it uses the default value of it so we don't need to update.

> Do we need e2e case update? I think this is a new condition for client, right?

Thanks for reminding, updated

wu-sheng (approved):
LGTM, once e2e passed.

wu-sheng (approved):

PR #7286, opened 2021-07-12
Upgrade resteasy version to 4. x, fix resteasy can't be traced in 4. x case

shenzhenwei-hub:


wu-sheng:
Your PR is wrong, I am not sure how you change these so many files.

wu-sheng:
You should check your changelog before submitting a pull request.

PR #7290, opened 2021-07-13
perf: optimize Envoy access log persistence performance

### LINKED ISSUE #7180: Improve LAL performance
We are using LAL to analyze log analysis in the ALS case, which faces a very serious performance impact in only 10k traffic. 
In a very simple case, sampling 5 logs in N seconds if logs show error status.

@kezhenxu94 I remember there is proto to YAML object process, I am not sure whether it causes a big impact. We need to check this later because we also have to resolve this in an OAL style rebuild solution. Also, because it seems we will break the LAL APIs totally, we are better to change in the 8.7.0 rather than in further releases.

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>. Closes #7180 

### DIFF
--- a/oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/filter/FilterSpec.java
+++ b/oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/filter/FilterSpec.java
@@ -26,10 +26,7 @@ import groovy.lang.DelegatesTo;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import lombok.SneakyThrows;
-import org.apache.skywalking.apm.network.logging.v3.JSONLog;
 import org.apache.skywalking.apm.network.logging.v3.LogData;
-import org.apache.skywalking.apm.network.logging.v3.LogDataBody;
 import org.apache.skywalking.oap.log.analyzer.dsl.Binding;
 import org.apache.skywalking.oap.log.analyzer.dsl.spec.AbstractSpec;
 import org.apache.skywalking.oap.log.analyzer.dsl.spec.extractor.ExtractorSpec;
@@ -45,8 +42,6 @@ import org.apache.skywalking.oap.server.library.module.ModuleManager;
 import org.apache.skywalking.oap.server.library.module.ModuleStartException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import static org.apache.skywalking.oap.server.library.util.ProtoBufJsonUtils.toJSON;
 
 public class FilterSpec extends AbstractSpec {
     private static final Logger LOGGER = LoggerFactory.getLogger(FilterSpec.class);
@@ -149,7 +144,6 @@ public class FilterSpec extends AbstractSpec {
         cl.call();
     }
 
-    @SneakyThrows
     @SuppressWarnings("unused")
     public void sink(@DelegatesTo(SinkSpec.class) final Closure<?> cl) {
         if (BINDING.get().shouldAbort()) {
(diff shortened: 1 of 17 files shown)


oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/filter/FilterSpec.java
```diff
@@ -161,9 +169,19 @@ public void sink(@DelegatesTo(SinkSpec.class) final Closure<?> cl) {
...
+        if (extraLog != null) {
+            logData.setBody(
+                LogDataBody.newBuilder()
+                           .setJson(
+                               JSONLog.newBuilder()
+                                      .setJson(toJSON(extraLog))
+                                      .build())
+                           .build()
+            );
+        }
```
wu-sheng, line 181: Is this the last step before persistent?

kezhenxu94: > Is this the last step before persistent?

Yes, it's last step of the LAL, maybe I can move it to right before persistence.

wu-sheng: Could use have multiple LAL files? If so, let's move it out of LAL scope.

kezhenxu94: DONE

oap-server/server-bootstrap/src/main/resources/lal/envoy-als.yaml
```diff
@@ -17,24 +17,22 @@ rules:
   - name: envoy-als
     dsl: |
       filter {
-        json {
-        }
         // only collect abnormal logs (http status code >= 300, or commonProperties?.responseFlags is not empty)
-        if (parsed?.response?.responseCode as Integer < 400 && !parsed?.commonProperties?.responseFlags) {
```
wu-sheng, line 23: Why `!parsed?.commonProperties?.responseFlags` removed?

wu-sheng: Do you mean `< 400` means no internal error?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/log/AbstractLogRecord.java
```diff
@@ -44,6 +44,7 @@
     public static final String SPAN_ID = "span_id";
     public static final String CONTENT_TYPE = "content_type";
     public static final String CONTENT = "content";
+    public static final String CONTENT_TYPE_CLASS = "content_type_class";
```
wu-sheng, line 47: I think we don't need this? You run `#toJSON` already.

oap-server/server-bootstrap/src/main/resources/lal/envoy-als.yaml
kezhenxu94: Added back

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/log/AbstractLogRecord.java
kezhenxu94: Yes, forgot to remove, now removed

wu-sheng (approved):
LGTM. Wait for tests.

PR #7293, opened 2021-07-13
Remove synchronous persistence mechanism in API level and ElasticSearch implementation

wu-sheng:

* Performance: remove the synchronous persistence mechanism from batch ElasticSearch DAO. Because the current enhanced persistent session mechanism, don't require the data queryable immediately after the insert and update anymore.
* Performance: share `flushInterval` setting for both metrics and record data, due
  to `synchronous persistence mechanism` removed. Record flush interval used to be hardcoded as 10s.

### DIFF
--- a/docs/en/FAQ/es-version-conflict.md
+++ b/docs/en/FAQ/es-version-conflict.md
@@ -1,0 +1,36 @@
+# Elasticsearch exception `type=version_conflict_engine_exception` since 8.7.0
+
+Since 8.7.0, we did the following optimization to reduce Elasticsearch load.
+
+```markdown
+Performance: remove the synchronous persistence mechanism from batch ElasticSearch DAO. Because the current enhanced
+persistent session mechanism, don't require the data queryable immediately after the insert and update anymore.
+```
+
+Due to this, we flush the metrics into Elasticsearch without using `WriteRequest.RefreshPolicy.WAIT_UNTIL`. This reduces
+the load of persistent works in OAP server and load of Elasticsearch CPU dramatically.
+
+Meanwhile, there is little chance you could see following **warn**s in your logs.
+
+```
+{
+  "timeMillis": 1626247722647,
+  "thread": "I/O dispatcher 4",
+  "level": "WARN",
+  "loggerName": "org.apache.skywalking.oap.server.library.client.elasticsearch.ElasticSearchClient",
+  "message": "Bulk [70] executed with failures:[failure in bulk execution:\n[18875]: index [sw8_service_relation_client_side-20210714], type [_doc], id [20210714_b3BlcmF0aW9uLXJ1bGUtc2VydmVyQDExNDgx.1-bWFya2V0LXJlZmVycmFsLXNlcnZlckAxMDI1MQ==.1], message [[sw8_service_relation_client_side-20210714/D7qzncbeRq6qh2QF5MogTw][[sw8_service_relation_client_side-20210714][0]] ElasticsearchException[Elasticsearch exception [type=version_conflict_engine_exception, reason=[20210714_b3BlcmF0aW9uLXJ1bGUtc2VydmVyQDExNDgx.1-bWFya2V0LXJlZmVycmFsLXNlcnZlckAxMDI1MQ==.1]: version conflict, required seqNo [14012594], primary term [1]. current document has seqNo [14207928] and primary term [1]]]]]",
+  "endOfBatch": false,
+  "loggerFqcn": "org.apache.logging.slf4j.Log4jLogger",
+  "threadId": 44,
+  "threadPriority": 5,
+  "timestamp": "2021-07-14 15:28:42.647"
+}
+```
+
+This would affect the system much, just a possibility of inaccurate of metrics. If this wouldn't show up in high
+frequency, you could ignore this directly.
+
+In case you could see many logs like this. Then it is a signal, that the flush period of your ElasticSearch template can't
+catch up your setting. Or you set the `persistentPeriod` less than the flush period.
+
+
--- a/docs/en/FAQ/es-version-conflict.md
+++ b/docs/en/FAQ/es-version-conflict.md
@@ -27,10 +27,9 @@ Meanwhile, there is little chance you could see following **warn**s in your logs
 }
 ```
 
-This would affect the system much, just a possibility of inaccurate of metrics. If this wouldn't show up in high
+This would not affect the system much, just a possibility of inaccurate of metrics. If this wouldn't show up in high
 frequency, you could ignore this directly.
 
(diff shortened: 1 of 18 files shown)


dmsolr:
Seem like you have talked to me about this feature serval times, right? @zifeihan

wu-sheng:
> Seem like you have talked to me about this feature serval times, right? @zifeihan

He and @Ax1an both mentioned this several months ago. We can't land this back then, but now with multiple enhancements in 8.7.0, I think we could enhance this with several following others.

kezhenxu94 (dismissed):

wu-sheng:
Add these 2 references. According to these, **(same _index, _type and _id) in same bulk will be in order**
1. https://github.com/elastic/elasticsearch/issues/50199
2. https://discuss.elastic.co/t/order-of--bulk-request-operations/98124

Ax1an (approved):
LGTM.

mrproliu (approved):

wu-sheng:
Notice, the order of different bulks are not guaranteed by ElasticSearch cluster. We are going to have risk of dirty write. But consider we set over 20s period between flush, and index flush period is 10, we should be safe.

@Ax1an will feedback the tests of this feature on his pre prod. We could see the improvement.

wu-sheng:
Because of this optimization, the persistent execution duration has dropped down to `< 1s`.
image

image
23k metrics in just 797ms, the previous versions are 50s

dmsolr (dismissed):

docs/en/FAQ/es-version-conflict.md
```diff
@@ -0,0 +1,36 @@
...
+
+Since 8.7.0, we did the following optimization to reduce Elasticsearch load.
+
+```markdown
+Performance: remove the synchronous persistence mechanism from batch ElasticSearch DAO. Because the current enhanced
+persistent session mechanism, don't require the data queryable immediately after the insert and update anymore.
+```
+
+Due to this, we flush the metrics into Elasticsearch without using `WriteRequest.RefreshPolicy.WAIT_UNTIL`. This reduces
+the load of persistent works in OAP server and load of Elasticsearch CPU dramatically.
```
kezhenxu94, line 11:
```suggestion
the load of persistent works in OAP server and load of Elasticsearch CPU drastically.
```

docs/en/FAQ/es-version-conflict.md
```diff
@@ -0,0 +1,36 @@
...
+  "message": "Bulk [70] executed with failures:[failure in bulk execution:\n[18875]: index [sw8_service_relation_client_side-20210714], type [_doc], id [20210714_b3BlcmF0aW9uLXJ1bGUtc2VydmVyQDExNDgx.1-bWFya2V0LXJlZmVycmFsLXNlcnZlckAxMDI1MQ==.1], message [[sw8_service_relation_client_side-20210714/D7qzncbeRq6qh2QF5MogTw][[sw8_service_relation_client_side-20210714][0]] ElasticsearchException[Elasticsearch exception [type=version_conflict_engine_exception, reason=[20210714_b3BlcmF0aW9uLXJ1bGUtc2VydmVyQDExNDgx.1-bWFya2V0LXJlZmVycmFsLXNlcnZlckAxMDI1MQ==.1]: version conflict, required seqNo [14012594], primary term [1]. current document has seqNo [14207928] and primary term [1]]]]]",
+  "endOfBatch": false,
+  "loggerFqcn": "org.apache.logging.slf4j.Log4jLogger",
+  "threadId": 44,
+  "threadPriority": 5,
+  "timestamp": "2021-07-14 15:28:42.647"
+}
+```
+
+This would affect the system much, just a possibility of inaccurate of metrics. If this wouldn't show up in high
```
kezhenxu94, line 30:
```suggestion
This would not affect the system much, just a possibility of inaccurate of metrics. If this wouldn't show up in high
```

wu-sheng:
> Notice, the order of different bulks are not guaranteed by ElasticSearch cluster. We are going to have risk of dirty write. But consider we set over 20s period between flush, and index flush period is 10, we should be safe.
> 
> @Ax1an will feedback the tests of this feature on his pre prod. We could see the improvement.

We verified, there are only 1 or 2 outputs about version conflict even after we set bulk size to 50k. So, I added a FAQ and recommend 5000 bulk size and 15s flush interval only. The persistent period has been set to 25s.

docs/en/setup/backend/configuration-vocabulary.md
```diff
@@ -20,6 +20,7 @@ core|default|role|Option values, `Mixed/Receiver/Aggregator`. **Receiver** mode
 | - | - | gRPCSslCertChainPath| The file path of gRPC SSL cert chain| SW_CORE_GRPC_SSL_CERT_CHAIN_PATH| - |
 | - | - | gRPCSslTrustedCAPath| The file path of gRPC trusted CA| SW_CORE_GRPC_SSL_TRUSTED_CA_PATH| - |
 | - | - | downsampling| The activated level of down sampling aggregation | | Hour,Day|
+| - | - | persistentPeriod| The execution period of the persistent timer. Unit is second. | | 25s |
```
kezhenxu94, line 23:
```suggestion
| - | - | persistentPeriod| The execution period of the persistent timer. Unit is second. | | 25 |
```

In case users think the unit is required in configuration.

docs/en/FAQ/es-version-conflict.md
```diff
@@ -0,0 +1,36 @@
...
+frequency, you could ignore this directly.
+
+In case you could see many logs like this. Then it is a signal, that the flush period of your ElasticSearch template can't
+catch up your setting. Or you set the `persistentPeriod` less than the flush period.
```
kezhenxu94, line 34: If this is known, can we check the configs at start up and warn the users by a warning log?

kezhenxu94 (commented):
Nice, just some nits

docs/en/FAQ/es-version-conflict.md
wu-sheng: I really mean `dramatically`

docs/en/FAQ/es-version-conflict.md
wu-sheng: @Ax1an I think this is server side warning?

wu-sheng: We can't, this relies on the ElasticSearch performance. I have set the default configurations as well as I can.

kezhenxu94 (approved):

docs/en/FAQ/es-version-conflict.md
wu-sheng: https://github.com/apache/skywalking/pull/7310 this should help.

PR #7294, opened 2021-07-14
Support envoy cluster manager metrics

wankai123:
### Support envoy cluster manager metrics
- [X] If it's UI related, attach the screenshots below.

### Metrics Name
Due to the `envoy gRPC Metric Service `  integrated metrics info and all meta info into the metrics name for now, like 
```
envoy_metrics {
  name: "cluster.outbound|9080||details.default.svc.cluster.local.update_attempt"
  type: COUNTER
  metric {
    counter {
      value: 6.0
    }
    timestamp_ms: 1625104746703
  }
}
```
I used `envoy_cluster_metrics` as the name of all the envoy cluster manager metrics in `OAP`and filter the specific metric with `MAL`.
### Cluster Name(Upstream service) label 
By parsing the `name` like `cluster.outbound|9080||details.default.svc.cluster.local.update_attempt` to get the `ServiceMetaInfo`. The format config is in the `metadata-service-mapping.yaml` which consists with the `node service`. Due to the given info only support
`ServiceName`, `NameSpace`, `Version` for now. 

### Snapshot 
image
image


JaredTan95:
Should `envoy/examples` be updated?

wu-sheng:
> Should [envoy/examples](https://github.com/apache/skywalking/blob/master/docs/en/setup/envoy/examples) be updated?

@wankai123 Please check that.

wu-sheng:
@wankai123 I can see Istio + Metrics Service e2e tests fail, please check the e2e.

wankai123:
> Should [envoy/examples](https://github.com/apache/skywalking/blob/master/docs/en/setup/envoy/examples) be updated?

It seems outdated, I think configuring envoy directly very complex and less practice. Shall we just remove it?

JaredTan95:
> > Should [envoy/examples](https://github.com/apache/skywalking/blob/master/docs/en/setup/envoy/examples) be updated?
> 
> It seems outdated, I think configuring envoy directly very complex and less practice. Shall we just remove it and keep the istio way.

If so, I think We need to say `envoy cluster manager metrics` only support for Istio/Mesh scenario in docs,  and we should adapt to non-mesh/Istio scenario, what do you think?

wu-sheng:
> If so, I think We need to say envoy cluster manager metrics only support for Istio/Mesh scenario in docs

Yes, the doc should be updated.

> and we should adapt to non-mesh/Istio scenario, what do you think?

From what I see today, there is very little case, people expect SkyWalking to accept envoy ALS and metric-service for analysis. Do you face that?

JaredTan95:
> > If so, I think We need to say envoy cluster manager metrics only support for Istio/Mesh scenario in docs
> 
> Yes, the doc should be updated.
> 
> > and we should adapt to non-mesh/Istio scenario, what do you think?
> 
> From what I see today, there is very little case, people expect SkyWalking to accept envoy ALS and metric-service for analysis. Do you face that?

Yes, for our case, we choose envoy to build our API gateway, but non-mesh scenario. and we expect envoy report it's ALS and metric to skywalking for observe envoy runtime status.

kezhenxu94:
> > Should [envoy/examples](https://github.com/apache/skywalking/blob/master/docs/en/setup/envoy/examples) be updated?
> 
> It seems outdated, I think configuring envoy directly very complex and less practice. Shall we just remove it and keep the istio way.

I don’t think it should be considered as “outdate” as it is a demo of direct integration of Envoy which can be absolutely used without istio. But I agree we don’t have much efforts to align the feature parity between mesh and non-mesh scenarios so I agree not to update this directory.

wu-sheng:
> > > If so, I think We need to say envoy cluster manager metrics only support for Istio/Mesh scenario in docs
> > 
> > 
> > Yes, the doc should be updated.
> > > and we should adapt to non-mesh/Istio scenario, what do you think?
> > 
> > 
> > From what I see today, there is very little case, people expect SkyWalking to accept envoy ALS and metric-service for analysis. Do you face that?
> 
> Yes, for our case, we choose envoy to build our API gateway, but non-mesh scenario. and we expect envoy report it's ALS and metric to skywalking for observe envoy runtime status.

In the latest codes, we expect to match the all and metric service as always, so we begin to read identify section. Could you try the codes for your local or use case env, and share with us what is happening? We can't see your case, and can't run this verification for you.

JaredTan95:
> In the latest codes, we expect to match the all and metric service as always

that's could be fine for our case, I'm only afraid there will be Istio's unique metrics. :-p

kezhenxu94:
> In the latest codes, we expect to match the all and metric service as always, so we begin to read identify section.

We had doc for this also if they want to use Envoy without istio https://github.com/apache/skywalking/blob/master/docs/en/setup/envoy/metrics_service_setting.md#configure-envoy-to-send-metrics-to-skywalking-without-istio

JaredTan95:
> > In the latest codes, we expect to match the all and metric service as always, so we begin to read identify section.
> 
> We had doc for this also if they want to use Envoy without istio https://github.com/apache/skywalking/blob/master/docs/en/setup/envoy/metrics_service_setting.md#configure-envoy-to-send-metrics-to-skywalking-without-istio

thx

> Should envoy/examples be updated?
In this comment, I mean that we should update https://github.com/apache/skywalking/blob/master/docs/en/setup/envoy/metrics_service_setting.md#configure-envoy-to-send-metrics-to-skywalking-without-istio.

As the latest PR codes do not affect native envoy metrics, fine to me.

wu-sheng:
> > > In the latest codes, we expect to match the all and metric service as always, so we begin to read identify section.
> > 
> > 
> > We had doc for this also if they want to use Envoy without istio https://github.com/apache/skywalking/blob/master/docs/en/setup/envoy/metrics_service_setting.md#configure-envoy-to-send-metrics-to-skywalking-without-istio
> 
> thx
> 
> > Should envoy/examples be updated?
> 
> In this comment, I mean that if we should update
> https://github.com/apache/skywalking/blob/master/docs/en/setup/envoy/metrics_service_setting.md#configure-envoy-to-send-metrics-to-skywalking-without-istio.
> 
> As the latest PR codes do not affect native envoy metrics, fine to me.

Yes, this PR doesn't change anything related to that. We just add more. @wankai123 You could keep the doc unchanged for this part.

wankai123:
> @wankai123 I can see Istio + Metrics Service e2e tests fail, please check the e2e.

e2e fixed

kezhenxu94 (approved):
LGTM. THANKS

PR #6752, opened 2021-04-14
merge

wingwong-knh:
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>.
- [ ] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md).

PR #7302, opened 2021-07-15
fix hbase intercepting bug seems like this issue: #6572;

wuguangkuo:
improve the interceptors code from using reflect manner to using public methods

### Fix #6752 
- [ x] Explain briefly why the bug exists and how to fix it.
     The bug happened when try to get a private filed relative with remote peer,but this field is null


wu-sheng:
I think you link incorrect issue?

wu-sheng:
And you breaks the plugin test, please verify.

wuguangkuo:
> And you breaks the plugin test, please verify.

The CI log told failure cause as "configuration-nacos" moudle tests，not the code of my committing

wu-sheng:
> > And you breaks the plugin test, please verify.
> 
> The CI log told failure cause as "configuration-nacos" moudle tests，not the code of my committing

Ignore this. I mean this, https://github.com/apache/skywalking/pull/7302/checks?check_run_id=3073114529

PR #7310, opened 2021-07-16
Adjust ElasticSearch index refresh period as INT(flushInterval * 2/3)

wu-sheng:

Adjust index refresh period as INT(flushInterval * 2/3), it used to be as same as bulk flush period. At the edge case, in low traffic(traffic < bulkActions in the whole period), there is a possible case, 2 period bulks are included in one index refresh rebuild operation, which could cause version conflicts. And this case can't be fixed through `core/persistentPeriod` as the bulk fresh is not controlled by the persistent timer anymore.

This PR should avoid the following exception in the low load case, especially when `bulkActions` is set larger than the number of a metric type.
image


docs/en/setup/backend/backend-infrastructure-monitoring.md
```diff
@@ -0,0 +1,119 @@
+# VMs monitoring
```
kezhenxu94, line 1: What's the relationship between this file and `backend-vm-monitoring.md` and `backend-k8s-monitoring.md`, this file seems to be a combination of `backend-k8s-monitoring.md` and `backend-vm-monitoring.md` but the other 2 files are still there, shouldn't them be removed?

wu-sheng: > What's the relationship between this file and `backend-vm-monitoring.md` and `backend-k8s-monitoring.md`, this file seems to be a combination of `backend-k8s-monitoring.md` and `backend-vm-monitoring.md` but the other 2 files are still there, shouldn't them be removed?

I didn't notice this. Let me remove this. This seems a git auto-merge bug.

wu-sheng: Done, revmoed.

kezhenxu94 (approved):

PR #7315, opened 2021-07-16
add http log receiver to support parse to raw log

dmsolr:

This PR is for receiving log which just includes log content. A raw log without `service`, `serviceInstance`, and etc. 
We can get these fields from http request headers.


wu-sheng (commented):
What is the context of this? I think  we should discuss before add new protocol.

dmsolr:
I have raised a discussion mail to dev@s.a.o, https://lists.apache.org/thread.html/r54096403dc5c6ac9b31033293a72997bf50350ff7c15f1d0ab31e7a8%40%3Cdev.skywalking.apache.org%3E

PR #7318, opened 2021-07-16
Logically revert #6642 and partial #7153 to reduce unnecessary thread and concurrency process

wu-sheng:

Logically revert #6642 and partial #7153 to reduce unnecessary thread and concurrency process

* The `core/maxSyncOperationNum` setting(added in 8.5.0) is removed due to metrics persistence is fully asynchronous.
* The `core/syncThreads` setting(added in 8.5.0) is removed due to metrics persistence is fully asynchronous.
* Optimization: Concurrency mode of execution stage for metrics is removed(added in 8.5.0). Only concurrency of prepare
  stage is meaningful and kept.

The key logic behinds all these is, **metrics persistence is fully asynchronous**.


wu-sheng:
FYI @a1vin-tian. We have done a lot of optimization in this 8.7.0 iteration, you could follow my previous PRs. All these optimizations are backed by @Ax1an 's production environment experiences and pre-prod experiments/feedback.

kezhenxu94 (dismissed):

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java
```diff
@@ -104,167 +92,51 @@ public void start(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
...
-                    while (!stop.get()) {
-                        List<PrepareRequest> partition = prepareQueue.poll();
-                        if (partition.isEmpty()) {
-                            break;
-                        }
-                        HistogramMetrics.Timer executeLatencyTimer = executeLatency.createTimer();
-                        try {
-                            if (CollectionUtils.isNotEmpty(partition)) {
-                                batchDAO.flush(partition);
+                        try (HistogramMetrics.Timer executeLatencyTimer = executeLatency.createTimer()) {
```
Ax1an, line 116: The meaning of `prepareLatency` has changed. It will contain `executeLatency`.

wu-sheng: I will fix this.

wu-sheng: @Ax1an I make the time of prepare stage accurate. And remove the outside preparedRequest list initialization, worker instance could always build a suitable size list in the first place (Reduce Array.copy and GC load a little).

Ax1an (approved):

Issue #7173, opened 2021-06-25
Discussion about update agent plugins for ShardingSphere


TeslaCN:
### Discussion

Since the ShardingSphere 5.0.0-beta has been released, we can add an agent plugin for the new version and remove some plugins for old version. I think we can divide this issue into the following steps:
- [ ] Add an agent plugin for ShardingSphere 5.0.0-beta #7171 
- [ ] Remove `sharding-jdbc-1.5.x-plugin`, `sharding-sphere-3.x-plugin`
- [ ] Move plugins for ShardingSphere into a new folder `shardingsphere-plugin`

wu-sheng:
About the removing old version, could you ask for a community vote about which versions should be removed.

TeslaCN:
> About the removing old version, could you ask for a community vote about which versions should be removed.

I will send an email to call for a vote later.

TeslaCN:
Hi community,
I have sent an email to ShardingSphere community to discuss about removing plugins for legacy version.

https://lists.apache.org/thread.html/r762a8235508633a88f950c3ac7cc142d4c81bd588e6964c44dc6f4fe%40%3Cdev.shardingsphere.apache.org%3E

wu-sheng:
@TeslaCN Could you finish the last step of this issue?

TeslaCN:
> @TeslaCN Could you finish the last step of this issue?

Sorry for the delay, I will finish this soon.

PR #7323, opened 2021-07-18
Remove plugins for ShardingSphere legacy versions

TeslaCN:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Related to #7173 .


wu-sheng:
https://github.com/apache/skywalking/blob/master/docs/en/setup/service-agent/java-agent/Supported-list.md This doc should be updated.

wu-sheng:
To people who want to use these removing versions, you still could copy the plugins from SkyWalking's 8.6.0 release to the `/plugins` folder to activate, unless SkyWalking has some core level APIs upgrade. This is very rare to happen.

wu-sheng (approved):
Once tests passed, I am good with this. Thanks.

PR #7332, opened 2021-07-19
Fixed instanceTraffic update problem after enhanced the storage session mechanism.

Ax1an:

image

   After the session mechanism is enhanced, the older `InstanceTraffic` data of `timeBucket` will always be cached in the context.
   
   And because of the logic of `combine` method in `InstanceTraffic`, we will continue to use the older `timeBucket` when updating.

   This causes OAP to update `InstanceTraffic` and report an error after cleaning up the index related to `InstanceTraffic` that has reached the specified TTL. 

   Because we have got the start time of the instance in other ways, I don't think we need to save the older `timeBucket`.


wu-sheng:
Good cache, but we need to discuss how to fix it. If we keep updating the instance, then the instance could be inserted multiple times in the index.

wu-sheng:
Replaced by #7334

PR #7334, opened 2021-07-19
Enhance persistent session timeout mechanism.

wu-sheng:

Enhance persistent session timeout mechanism. Because the enhanced session could cache the metadata metrics forever, a new timeout mechanism is designed for avoiding this specific case. 

This fix should resolve the bug described in ##7332.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
@@ -33,7 +33,6 @@ import org.apache.skywalking.apm.commons.datacarrier.consumer.BulkConsumePool;
 import org.apache.skywalking.apm.commons.datacarrier.consumer.ConsumerPoolFactory;
 import org.apache.skywalking.apm.commons.datacarrier.consumer.IConsumer;
 import org.apache.skywalking.oap.server.core.UnexpectedException;
-import org.apache.skywalking.oap.server.core.analysis.TimeBucket;
 import org.apache.skywalking.oap.server.core.analysis.data.MergableBufferedData;
 import org.apache.skywalking.oap.server.core.analysis.data.ReadWriteSafeCache;
 import org.apache.skywalking.oap.server.core.analysis.metrics.Metrics;
@@ -82,7 +81,7 @@ public class MetricsPersistentWorker extends PersistenceWorker<Metrics> {
     /**
      * @since 8.7.0 TTL settings from {@link org.apache.skywalking.oap.server.core.CoreModuleConfig#getMetricsDataTTL()}
      */
-    private int metricsDataTTLInMillis;
+    private int metricsDataTTL;
 
     MetricsPersistentWorker(ModuleDefineHolder moduleDefineHolder, Model model, IMetricsDAO metricsDAO,
                             AbstractWorker<Metrics> nextAlarmWorker, AbstractWorker<ExportEvent> nextExportWorker,
@@ -100,8 +99,7 @@ public class MetricsPersistentWorker extends PersistenceWorker<Metrics> {
         this.sessionTimeout = storageSessionTimeout;
         this.persistentCounter = 0;
         this.persistentMod = 1;
-        // Convert TTL to millis to make further calculation easier.
-        this.metricsDataTTLInMillis = metricsDataTTL * 24 * 60 * 60 * 1000;
+        this.metricsDataTTL = metricsDataTTL;
 
         String name = "METRICS_L2_AGGREGATION";
         int size = BulkConsumePool.Creator.recommendMaxSize() / 8;
@@ -275,10 +273,8 @@ public class MetricsPersistentWorker extends PersistenceWorker<Metrics> {
                            // database has been removed due to TTL.
                            if (!model.isTimeRelativeID() && supportUpdate) {
                                // Mostly all updatable metadata level metrics are required to do this check.
-                               final long metricsTimestamp = TimeBucket.getTimestamp(
-                                   cachedValue.getTimeBucket(), model.getDownsampling());
-                               // If the cached metric is older than the TTL indicated.
-                               if (currentTimeMillis - metricsTimestamp > metricsDataTTLInMillis) {
+
+                               if (metricsDAO.isExpiredCache(model, cachedValue, currentTimeMillis, metricsDataTTL)) {
                                    // The expired metrics should be tagged `not in cache` directly.
                                    return true;
                                }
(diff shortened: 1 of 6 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
```diff
@@ -246,10 +258,35 @@ private void nextWorker(Metrics metrics) {
...
+                           // Metrics keep coming due to traffic, but the entity in the
+                           // database has been removed due to TTL.
+                           if (!model.isTimeRelativeID() && supportUpdate) {
+                               // Mostly all updatable metadata level metrics are required to do this check.
+                               final long metricsTimestamp = TimeBucket.getTimestamp(
+                                   cachedValue.getTimeBucket(), model.getDownsampling());
+                               // If the cached metric is older than the TTL indicated.
+                               if (currentTimeMillis - metricsTimestamp > metricsDataTTLInMillis) {
+                                   // The expired metrics should be tagged `not in cache` directly.
+                                   return true;
```
wu-sheng, line 283: I don't remove the existing cache, as in this case, the expired key should be overridden directly in the next `put`/`read` stage.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
```diff
@@ -95,6 +100,8 @@
         this.sessionTimeout = storageSessionTimeout;
         this.persistentCounter = 0;
         this.persistentMod = 1;
+        // Convert TTL to millis to make further calculation easier.
+        this.metricsDataTTLInMillis = metricsDataTTL * 24 * 60 * 60 * 1000;
```
kezhenxu94, line 104: Consider using `TimeUnit.DAYS.toMilliSeconds` for better readability.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
kezhenxu94: This is ok for now, but when we support MAL (or OAL) hot reload dynamically, it’s possible users will remove some metrics and this cache will be leaked.

kezhenxu94 (approved):
Basically good to me

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
wu-sheng: Where do we have this method?

Ax1an (dismissed):

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
kezhenxu94: It’s in JDK `TimeUnit.DAYS.toMillis`


https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/TimeUnit.html#toMillis-long-

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
wu-sheng: I think it wouldn't. There are 2 reason
1. Session is built based on hot access. Once there is no traffic, the session gets removed after minutes.
2. Hot reload should remove worker instances too. The session is attached, so should be removed too.

Also, this PR is removing hot data, it will be in the cache one way or another. Such as reloading from database, due to merging.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
```diff
@@ -246,10 +258,35 @@ private void nextWorker(Metrics metrics) {
...
+                           // whether the cache is expired due to TTL.
+                           // This is a cache-DB inconsistent case:
+                           // Metrics keep coming due to traffic, but the entity in the
+                           // database has been removed due to TTL.
+                           if (!model.isTimeRelativeID() && supportUpdate) {
+                               // Mostly all updatable metadata level metrics are required to do this check.
+                               final long metricsTimestamp = TimeBucket.getTimestamp(
+                                   cachedValue.getTimeBucket(), model.getDownsampling());
+                               // If the cached metric is older than the TTL indicated.
+                               if (currentTimeMillis - metricsTimestamp > metricsDataTTLInMillis) {
```
wu-sheng, line 281: This calculation way could cause the removed metadata metrics exist in the cache for hours, due to ElasticSearch implementation manage the index by day, rather than accurate millisecond.
I think I have to move this logic into implementation level to get the current logic.

cc @kezhenxu94 @Ax1an

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
wu-sheng: Done

wu-sheng:
@kezhenxu94 @Ax1an I just submit a new optimization to make the session expired mechanism better.

kezhenxu94 (approved):

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/MetricsEsDAO.java
```diff
@@ -117,4 +121,25 @@ public UpdateRequest prepareBatchUpdate(Model model, Metrics metrics) throws IOE
...
+            cachedValue.getTimeBucket(), model.getDownsampling());
+        // Fast fail check. If the duration is still less than TTL - 1 days(absolute)
+        // the cache should not be expired.
+        if (currentTimeMillis - metricTimestamp < TimeUnit.DAYS.toMillis(ttl - 1)) {
+            return false;
+        }
+        final long deadline = Long.parseLong(new DateTime(currentTimeMillis).plusDays(-ttl).toString("yyyyMMdd"));
+        final long timeBucket = TimeBucket.getTimeBucket(cachedValue.getTimeBucket(), DownSampling.Day);
+        // If time bucket is earlier or equals(mostly) the deadline, then the cached metric is expired.
+        if (timeBucket <= deadline) {
```
wu-sheng, line 140: This keeps consistent with TTL mechanism implementation.

https://github.com/apache/skywalking/blob/4003c4b777d8e4026e87a4474ede939c15282b05/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/HistoryDeleteEsDAO.java#L62-L63

Ax1an (approved):

PR #7340, opened 2021-07-20
Adjust the project structure of the alarm module

JaredTan95:
Adjust the project structure of the alarm module for convenient the user to add their own implementation.


wu-sheng:
What do you mean own implementation? I don't think we would provide 2 kinds of implementations.

Could you provide more context?

JaredTan95:
> What do you mean own implementation? I don't think we would provide 2 kinds of implementations.
> 
> Could you provide more context?

Makes it easy for users to add their own alarm implementations, It's mean that user can use `selector` to extend sw alarm logical, like :

```yaml
alarm:
  selector: ${SW_ALARM:myAlarm}
  default:
  myAlarm:
```

For our case, we may use `myAlarm`  to implement multi-tenant alarm logical, and we can reuse `alarm-default-plugin` core.

kezhenxu94 (approved):
It's OK for me to adjust the structure, so that it respects the same structures as `server-query-plugin`, `server-cluster-plugin`, `server-configuration-plugin`, etc.,

JaredTan95:
> It's OK for me to adjust the structure, so that it respects the same structures as `server-query-plugin`, `server-cluster-plugin`, `server-configuration-plugin`, etc.,

Yes

> It's OK for me to adjust the structure, so that it respects the same structures as `server-query-plugin`, `server-cluster-plugin`, `server-configuration-plugin`, etc.,

Yes

JaredTan95:
> It's OK for me to adjust the structure, so that it respects the same structures as `server-query-plugin`, `server-cluster-plugin`, `server-configuration-plugin`, etc.,

Yes

> It's OK for me to adjust the structure, so that it respects the same structures as `server-query-plugin`, `server-cluster-plugin`, `server-configuration-plugin`, etc.,

Yes

wu-sheng:
Why this merge doesn't show the change in the commit log?

image

Please pay attention next time, this makes readers feeling we merge some dirty commits.

wu-sheng:
> > What do you mean own implementation? I don't think we would provide 2 kinds of implementations.
> > Could you provide more context?
> 
> Makes it easy for users to add their own alarm implementations, It's mean that user can use `selector` to extend sw alarm logical, like :
> 
> ```yaml
> alarm:
>   selector: ${SW_ALARM:myAlarm}
>   default:
>   myAlarm:
> ```
> 
> For our case, we may use `myAlarm` to implement multi-tenant alarm logical, and we can reuse `alarm-default-plugin` core at the same time.

I was objecting to this PR, but the reason for doing this seems not very precise and reasonable. You could do the same thing to extend the implementation, without changing the structure. 
All existing modules with this structure means SkyWalking is going to provide multiple implementation, but we wouldn't. What you are doing in the private fork is working as always. But why it has to be inside SkyWalking code structure?

wu-sheng:
This would become a bad example of upstream has to change code structure for adopting private forks, why?

wu-sheng:
I am submitting a pull request to revert this. If you have a reasonable purpose, we could continue the discussion there.

JaredTan95:
> Why this merge doesn't show the change in the commit log?
> 
> image
> 
> Please pay attention next time, this makes readers feeling we merge some dirty commits.

May caused by my browser networks, it delayed when I `squash and merge` PR, so, I had chosen to merge commit directly by mistake. I will pay attention next time.

JaredTan95:
> I am submitting a pull request to revert this. If you have a reasonable purpose, we could continue the discussion there.

Agree,

PR #7346, opened 2021-07-21
Could org.apache.skywalking:hbase-scenario:5.0.0 drop off redundant dependencies to loose weight?

Celebrate-future:
Hi, I deployed project **_org.apache.skywalking:hbase-scenario:5.0.0_** and found that its pom file introduced **_98_** dependencies. However, among them, **_4_** libraries (**_4%_**) have not been used by your project (the redundant dependencies are listed below). Reduce these useless dependencies can help prevent conflicts between library versions. Meanwhile, it can minimize the total added size to projects. It can also help enable advanced scenarios for this project. 
This PR helps **_org.apache.skywalking:hbase-scenario:5.0.0_** lose weight :) I have tested the revised configuration in my local environment. It is safe to remove the unused libraries.
 
Best regards

### DIFF
--- a/test/plugin/scenarios/hbase-scenario/pom.xml
+++ b/test/plugin/scenarios/hbase-scenario/pom.xml
@@ -51,6 +51,10 @@
                     <groupId>log4j</groupId>
                     <artifactId>*</artifactId>
                 </exclusion>
+                <exclusion>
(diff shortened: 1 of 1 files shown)


wu-sheng:
That is a test project, I don't know why you deploy that.

PR #7347, opened 2021-07-21
Fix the persistent session timeout mechanism bug.

Ax1an:
- [x] Closes #7334.

If we don't remove old data from context after expiration, the old data will continue to be in the context and continue to generate `UpdateRequest`. Finally, the document missing exception continues to be generated.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
@@ -275,7 +275,8 @@ public class MetricsPersistentWorker extends PersistenceWorker<Metrics> {
                                // Mostly all updatable metadata level metrics are required to do this check.
 
                                if (metricsDAO.isExpiredCache(model, cachedValue, currentTimeMillis, metricsDataTTL)) {
-                                   // The expired metrics should be tagged `not in cache` directly.
+                                   // The expired metrics should be removed from the context and tagged `not in cache` directly.
+                                   context.remove(m);
                                    return true;
                                }
                            }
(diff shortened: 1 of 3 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
```diff
@@ -275,7 +275,8 @@ private void loadFromStorage(List<Metrics> metrics) {
                                // Mostly all updatable metadata level metrics are required to do this check.
 
                                if (metricsDAO.isExpiredCache(model, cachedValue, currentTimeMillis, metricsDataTTL)) {
-                                   // The expired metrics should be tagged `not in cache` directly.
+                                   // The expired metrics should be removed from the context and tagged `not in cache` directly.
+                                   context.remove(m);
```
wu-sheng, line 279: Why really need to remove? `return true` would make this in the `notInCacheMetrics`. Then `multiGet` should load and override it.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/MetricsEsDAO.java
```diff
@@ -135,7 +135,7 @@ public boolean isExpiredCache(final Model model,
             return false;
         }
         final long deadline = Long.parseLong(new DateTime(currentTimeMillis).plusDays(-ttl).toString("yyyyMMdd"));
-        final long timeBucket = TimeBucket.getTimeBucket(cachedValue.getTimeBucket(), DownSampling.Day);
+        final long timeBucket = TimeBucket.getTimeBucket(metricTimestamp, DownSampling.Day);
```
wu-sheng, line 138: This is a bug fix.

wu-sheng (approved):

PR #7349, opened 2021-07-21
Fix IllegalArgumentException in PrometheusMetricConverter and NaN value samples in SampleFamily

### LINKED ISSUE #7343: prometheus meter analyzer produce IllegalArgumentException and NaN value

- [ ] Question or discussion
- [x] Bug
- [ ] Requirement

___
### Bug

https://github.com/apache/skywalking/blob/62411964a46aa22d47319ca863feed3a39169548/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/PrometheusMetricConverter.java#L87-L100

I think `Line 91` should be:
```java 
.filter(t -> t != NIL && t._2.samples.length > 0)
```
because `line 98` will throw an `IllegalArgumentException` if `samples.length == 0`.

---

https://github.com/apache/skywalking/blob/62411964a46aa22d47319ca863feed3a39169548/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/SampleFamily.java#L75-L79

I think `line 78` should be:
```java
return new SampleFamily(Arrays.stream(samples).filter(sample -> !Double.isNaN(sample.getValue())).toArray(Sample[]::new), Optional.ofNullable(ctx).orElseGet(RunningContext::instance));
```
because sample's value may be 'NaN'.

Lighfer:
### Fix #7343
 > 1. filtering SampleFamily which samples field is an empty array. 
 > 2. filtering samples which value is NaN

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7343.


wu-sheng:
cc @wankai123 Please review.

wankai123 (commented):
LGTM

wankai123 (approved):
LGTM

wu-sheng (approved):

PR #7354, opened 2021-07-21
Fix possible version_conflict_engine_exception in bulk execution.

Ax1an:
- [x] Closes #7293.

When I adjusted the aggregator from cluster deployment to stand-alone deployment, I found that this exception disappeared, so I concluded that it was an OAP problem. And I found that the exception is always related to the relation's  hour and day dimensionality metrics. Finally, I found that the `remoteHashCode` method of some relation metrics was not set reasonably and related to `timeBucket`. This causes metrics with the same entityId to be sent to multiple aggregators.


wu-sheng (approved):
Great one!!

kezhenxu94 (approved):

PR #7363, opened 2021-07-23
Add an agent plugin to support `spring-rabbit-2.3.x`

### LINKED ISSUE #2719: rabbit in spring boot project, the one total trace will divide two trace

- [ ] Question or discussion
- [x] Bug
- [ ] Requirement

___
### Bug
6.1
- Which company or project?
use rabbit `@RabbitListener` in spring project,  the code like this:

```java
    @RabbitListener(queues = "aaaa")
    public void handlerMessage(String message) {
        //  there will invoke dubbo service
    }
```

It should be one trace like:` rabbitmq-producer -> rabbitmq-consumer -> dubbo`
but in fact, I got two trace: `rabbitmq-producer -> rabbitmq-consumer ` and `dubbo`

image

I have find the reason: 
the rabbit producer InstanceMethodsInterceptPoint was `com.rabbitmq.client.impl.ChannelN#basicPublish` , it will call at `AMQP Connection` Thread, But `@RabbitListener` method will call at other thread, So the trace will divide two trace

qsliu2017:
### Add an agent plugin to support `spring-rabbit-2.3.x`

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #2719 .

As mentioned in #2719 , this pr intends to add a plugin to fix broken traces in rabbitmq. When using with the original `rabbitmq-plugin` it causes redundant traces. So I put it in optional plugins.


wu-sheng:
2 things first. 
1. changes.md in the root should be updated.
2. As you said duplicate case in pure rabbitmq case, it is possible to use this replace the existing plugin rather than adding a new one?

wu-sheng:
Your commit logs seem not clean. Could you make sure it includes your changes only? You should learn how to sync witg upstream.

qsliu2017:
> 2 things first.
> 
> 1. changes.md in the root should be updated.
> 2. As you said duplicate case in pure rabbitmq case, it is possible to use this replace the existing plugin rather than adding a new one?

1. I have updated it
2. The existing plugin works for `rabbitmq java client`, but this one only works for its implementation in spring.

wu-sheng:
This doesn't make sense. You said there is duplicate spans, then these is at least one case both plugins work.

qsliu2017:
The existing plugin enhances rabbitmq java client interface so it works for all rabbitmq client implementations. But this plugin only enhances the implementation of rabbitmq client in spring. When a spring application calling rabbitmq, both plugins can be active. But for clients other than spring, this one will not be active. That's why we can't replace the existing one.

I hope I've made myself clear.:sweat_smile:

wu-sheng:
> The existing plugin enhances rabbitmq java client interface so it works for all rabbitmq client implementations. But this plugin only enhances the implementation of rabbitmq client in spring. When a spring application calling rabbitmq, both plugins can be active. But for clients other than spring, this one will not be active. That's why we can't replace the existing one.
> 
> 
> 
> I hope I've made myself clear.:sweat_smile:

This explained why you need one, but doesn't explain why you need a new one. You said, the existing plugin works for all case, including through Spring, then why need this?

The key point I keep asking is, why this has to be optional? Why can't optimize it to work in default and only necessary when existing plugin doesn't work.

qsliu2017:
> > The existing plugin enhances rabbitmq java client interface so it works for all rabbitmq client implementations. But this plugin only enhances the implementation of rabbitmq client in spring. When a spring application calling rabbitmq, both plugins can be active. But for clients other than spring, this one will not be active. That's why we can't replace the existing one.
> > I hope I've made myself clear.😅
> 
> This explained why you need one, but doesn't explain why you need a new one. You said, the existing plugin works for all case, including through Spring, then why need this?
> 
> The key point I keep asking is, why this has to be optional? Why can't optimize it to work in default and only necessary when existing plugin doesn't work.

As said in #2719, the existing rabbitmq plugin divides a total trace into two parts because of the internals of rabbitmq. I need a new one which can make a trace complete.
This plugin can fix it but **only** in spring. I haven't found a way to disable the existing one automatically when this plugin is active.😅 So I put it in optional plugins, to not make duplicate spans but give users an option to replace the existing plugin manually.

wu-sheng:
You haven't gotten my point. If spans could be duplicate, then both plugins work, then the only possibility a trace breaks, because the context doesn't propagate, you just need a snapshot to propagate a context. Right now, you are creating another span, which makes you facing the duplicate span issue. We have plugins for Spring scope and out of Spring scope, they should not influence each other.

The plugin system is complex, we should never make users to determine how to choose in or out Spring use case, they will be confused about the difference.

qsliu2017:
> You haven't gotten my point. If spans could be duplicate, then both plugins work, then the only possibility a trace breaks, because the context doesn't propagate(across threads mostly), you just need a snapshot to propagate a context. Right now, you are creating another span, which makes you facing the duplicate span issue. We have plugins for Spring scope and out of Spring scope, they should not influence each other.
> 
> The plugin system is complex, we should never make users to determine how to choose in or out Spring use case, they will be confused about the difference.

well, it looks like I misunderstand *optional* plugins.😅 does that mean we need to focus on rabbitmq itself, rather than spring rabbit? And what do you mean by **a snapshot to propagate a context**?

wu-sheng:
I don't say we don't take care of Spring ecosystem. I mean you still don't explain the real reason why trace breaks.
Plugin dev doc explains the snapshot, and capture/continue operation.

When you add a thing, you should make sure it is compatible with other cases, rather than just give users to choose. SkyWalking never asked you to choose plugin, do we.

qsliu2017:
In rabbitmq client, dispatcher handles a message delivery and calls the consumer in a new thread.
```java
package com.rabbitmq.client.impl;
final class ConsumerDispatcher {
    public void handleDelivery(final Consumer delegate,
                               final String consumerTag,
                               final Envelope envelope,
                               final AMQP.BasicProperties properties,
                               final byte[] body) throws IOException {
        executeUnlessShuttingDown(
        new Runnable() {
            @Override
            public void run() {
                try {
                    delegate.handleDelivery(consumerTag,
                            envelope,
                            properties,
                            body);
                } catch (Throwable ex) {
                    connection.getExceptionHandler().handleConsumerException(
                            channel,
                            ex,
                            delegate,
                            consumerTag,
                            "handleDelivery");
                }
            }
        });
    }
}
```

```java
package com.rabbitmq.client;
public interface Consumer {
    void handleDelivery(String consumerTag,
                        Envelope envelope,
                        AMQP.BasicProperties properties,
                        byte[] body)
        throws IOException;
}
```
The existing plugin enhances `Consumer#handleDelivery`, so the consumer span is cross-thread, which causes broken traces. @wu-sheng Does this make sense?

wu-sheng:
If there is only a case of across thread, please capture a snapshot in `ConsumerDispatcher#handleDelivery` method, then propagate the context into `executeUnlessShuttingDown` method with a wrapper runnable.
This is the recommended way to process an across thread case.

qsliu2017:
After trying to propagate a snapshot from `ConsumerDispatcher` to `Consumer`, I've found it doesn't work.😅

Because rabbit java client **does** work and does **not** cause broken traces. In another word, the existing plugin works well with java clients except `spring-rabbit`.(sry I didn't try other java clients before... 😅)

The real reason for broken traces is that `spring-rabbit` using a container to dispatch listeners (in different threads) to consume rabbit messages. So adding a plugin to propagate snapshots from container to listeners should fix this issue and not conflict with the existing one.

wu-sheng:
Thanks for checking and clarification. Are you going to rework on this PR?

qsliu2017:
> Thanks for checking and clarification. Are you going to rework on this PR?

Yes, I'll try to modify this plugin

wu-sheng:
@qsliu2017 Do you still work on this?

PR #7364, opened 2021-07-23
Improve okhttp plugin performance by optimizing Class.getDeclaredField().

ArjenDavid-sjtu:
### Improve the performance of `apm-okhttp-common`
- [ ] Add a benchmark for the improvement: Not necessary
- [ ] The benchmark result: Trivial
- [ ] Links/URLs to the theory proof or discussion articles/blogs.
https://stackoverflow.com/questions/16395881/getdeclaredfields-absolute-performance

- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. No

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,7 +39,6 @@ Release Notes.
 * Remove plugins for ShardingSphere legacy version.
 * Update agent plugin for ElasticJob GA version
 * Remove the logic of generating instance name in `KafkaServiceManagementServiceClient` class.
-* Improve the performance of `Class.getDeclaredField()` in okhttp plugin.
 
 #### OAP-Backend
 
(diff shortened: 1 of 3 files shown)


wu-sheng:
The big performance impact is mostly because reflect execution. But getting field is not very slow. Are you facing real issue?

wu-sheng:
And you should update changes.md in the root.

ArjenDavid-sjtu:
> The big performance impact is mostly because reflect execution. But getting field is not very slow. Are you facing real issue?

I am not facing the real issue yet, maybe later a comparison between the two ways can be made. I was just thought another better way to do getting field.

ArjenDavid-sjtu:
> And you should update changes.md in the root.

I am so sorry for forgetting it! I have updated the changes.md now.

wu-sheng:
> > The big performance impact is mostly because reflect execution. But getting field is not very slow. Are you facing real issue?
> 
> 
> 
> I am not facing the real issue yet, maybe later a comparison between the two ways can be made. I was just thought another better way to do getting field. 

This is better, just want to confirm the real case.

wu-sheng:
Once all tests passed, this at least improves a little, but limited.

wu-sheng (approved):

PR #7370, opened 2021-07-26
When client invoke cancel, it will trigger onCancel, not onComplete

844067874:
Relate to https://github.com/apache/skywalking/pull/7179


wu-sheng:
What is this PR for? Could you explain?

844067874:
https://github.com/apache/skywalking/pull/7179
last month I fix the reponse time of gRPC,and left some problems.

wu-sheng (approved):

PR #7373, opened 2021-07-26
fix:delete accept-encoding

chenhaipeng:
### Fix <zipkin trace accept-encoding handle error>
     The HTTP request header Accept-Encoding informs (to the server) the content encoding method that the client can understand—usually a certain compression algorithm. Through content negotiation, the server will choose a method proposed by the client, use and notify the client of the choice in the response header Content-Encoding. It is to inform the server that I support a certain compression method, and when the web page responds, it uses the corresponding method to compress the data and transmits it to us, rather than specifying what kind of transmission and decoding method the server uses to decompress and accept. @link https://developer.mozilla.org/zh-CN/docs/Web/HTTP/Headers/Accept-Encoding


wu-sheng:
Hi, this is not by our writing. This is originally logic from the Zipkin server.

PR #7377, opened 2021-07-27
Add logging module to load log configuration dynamically

### LINKED ISSUE #7114: Add logging level switching API in OAP backend for debugging purpose
For now, we have some logs that are very helpful in debugging and troubleshooting but too verbose to be printed in production environment, hence we print them in `DEBUG` level.

However, switching the log level requires OAP to restart and that may make the errors recover from a restart, so in this issue, we're going to discuss and implement a new API that allows users to switch a particular logger's level. This is kind of similar to [Spring Boot Actuator loggers endpoint](https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.loggers).

Another doc provided by @hanahmily is [log4j auto-config](https://logging.apache.org/log4j/2.x/manual/customconfig.html)

hanahmily:
### Logging configuration dynamic loading
- [x] Local testing.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7114 .


oap-server/server-logging/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+        <dependency>
+            <groupId>org.apache.skywalking</groupId>
+            <artifactId>server-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-api</artifactId>
+        </dependency>
+        <dependency>
```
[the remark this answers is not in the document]
hanahmily: ignore

wu-sheng:
This doc you should update about how to use this dynamic configuration.
https://github.com/apache/skywalking/blob/master/docs/en/setup/backend/dynamic-config.md

And since you are adding new module, https://github.com/apache/skywalking/blob/master/docs/en/setup/backend/configuration-vocabulary.md has to be updated.

Also, it is better to add a whole new menu in the `advanced setup` section about the dynamic logging setup and usage.
image

wu-sheng:
Before doing all the above doc side changes, I want to ask, why this dynamic log should be a new module rather than a feature inside the core module? It looks like no other modules depend on this one, and there is no chance to switch to a new implementation.

hanahmily:
> Before doing all the above doc side changes, I want to ask, why this dynamic log should be a new module rather than a feature inside the core module? It looks like no other modules depend on this one, and there is no chance to switch to a new implementation.

Dynamic logging depends on the `configuration` module. Can a core component support such a constraint?

After this PR, we need to support dynamic logging without `configuration`, a possible path is to inject log4j configuration through RESTful endpoint/GraphQL mutation, that's a new provider more than this one.

wu-sheng:
Why do you need `configuration`? 

> > Before doing all the above doc side changes, I want to ask, why this dynamic log should be a new module rather than a feature inside the core module? It looks like no other modules depend on this one, and there is no chance to switch to a new implementation.
> 
> Dynamic logging depends on the `configuration` module. Can a core component support such a constraint?
> 
> After this PR, we need to support dynamic logging without `configuration`, a possible path is to inject log4j configuration through RESTful endpoint/GraphQL mutation, that's a new provider more than this one.

Core module level HTTP Rest host:port should be a good way to host this service. The existing Query GraphQL isn't. 
Even you want to GraphQL level, it should not be a query protocol. We need to be careful about this kind of feature. CVE and security alert could be easy injected through this way.

oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -510,3 +510,7 @@ configuration-discovery:
 receiver-event:
   selector: ${SW_RECEIVER_EVENT:default}
   default:
+
+logging:
+  selector: ${SW_LOGGING:-}
+  default:
```
wu-sheng, line 516: Legacy change.

wu-sheng (commented):
Changelog and doc update please.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/logging/LoggingConfigWatcher.java
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * LoggingConfigWatcher watches the change of logging configuration.
+ * Once got the change content, it would apply them to the current logger context.
+ */
+@Slf4j
+public class LoggingConfigWatcher extends ConfigChangeWatcher {
+    private final LoggerContext ctx;
+    private final OapConfiguration originConfiguration;
+    private String content;
```
wu-sheng, line 41:
```suggestion
    private volatile String content;
```
This is being accessed across threads.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/logging/LoggingConfigWatcher.java
```diff
@@ -0,0 +1,89 @@
...
+    public void notify(final ConfigChangeEvent value) {
+        if (EventType.DELETE.equals(value.getEventType())) {
+            this.content = "";
+        } else {
+            this.content = value.getNewValue();
+        }
+        try {
+            boolean applied = updateConfig();
+            if (log.isDebugEnabled() && applied) {
+                log.debug("applied {} B data to logging configuration", this.content.length());
```
wu-sheng, line 59: What does `B` mean?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/logging/LoggingConfigWatcher.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public String value() {
+        return this.content;
+    }
+
+    private boolean reconfigure(final String newValue) {
+        if (Strings.isNullOrEmpty(newValue)) {
+            if (ctx.getConfiguration().equals(originConfiguration)) {
```
wu-sheng, line 78:
```suggestion
            if (!ctx.getConfiguration().equals(originConfiguration)) {
```
Not equal should mean reset?

oap-server/server-configuration/configuration-k8s-configmap/src/main/java/org/apache/skywalking/oap/server/configuration/configmap/ConfigmapConfigurationWatcherRegister.java
```diff
@@ -46,9 +45,7 @@ public ConfigmapConfigurationWatcherRegister(ConfigmapConfigurationSettings sett
             if (log.isDebugEnabled()) {
                 log.debug("read config: name:{} ,value:{}", name, value);
             }
-            if (Objects.nonNull(value)) {
-                configTable.add(new ConfigTable.ConfigItem(name, value));
-            }
+            configTable.add(new ConfigTable.ConfigItem(name, value));
```
wu-sheng, line 48: This should be documented as a bug fix in the changelog.

hanahmily: noted

docs/en/setup/backend/dynamic-config.md
```diff
@@ -11,6 +11,7 @@ Right now, SkyWalking supports following dynamic configurations.
 |alarm.default.alarm-settings| The alarm settings, will override `alarm-settings.yml`. | same as [`alarm-settings.yml`](backend-alarm.md) |
 |core.default.apdexThreshold| The apdex threshold settings, will override `service-apdex-threshold.yml`. | same as [`service-apdex-threshold.yml`](apdex-threshold.md) |
 |core.default.endpoint-name-grouping| The endpoint name grouping setting, will override `endpoint-name-grouping.yml`. | same as [`endpoint-name-grouping.yml`](endpoint-grouping-rules.md) |
+|core.default.log4j-xml| The log4j xml configuration, will override `log4j2.xml`. | same as [`log4j2.xml`](https://logging.apache.org/log4j/2.x/manual/configuration.html#XML) |
```
wu-sheng, line 14: Please link to our internal doc.

wu-sheng:
Please recheck jdk11 whether it is real issue, or network. I can see this task fails repeatedly.

docs/en/setup/backend/dynamic-config.md
hanahmily: done

wu-sheng (approved):

PR #7384, opened 2021-07-28
Log sampler migrate from qps to rpm

kezhenxu94:


oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/sink/sampler/RateLimitingSampler.java
```diff
@@ -34,9 +34,10 @@
 @EqualsAndHashCode(of = {"qps"})
 public class RateLimitingSampler implements Sampler {
     @Getter
-    @Setter
     private volatile int qps;
```
wu-sheng, line 37: Do we still need to keep this?

wu-sheng: You should update the document too.

wu-sheng:
And you should update LAL script, right?

kezhenxu94:
This patch doesn’t change the public API / script, it used qps * 60 as RPM internally.

wu-sheng:
> This patch doesn’t change the public API / script, it used qps * 60 as RPM internally.

I know. I am afraid this could be misleading. Because `30qps * 60` is different with 1800rpm. In the worth case, you could get over 500qps sampled easily in some certain cases.

wu-sheng:
I think we should break the API in this release.

wu-sheng (approved):

PR #7398, opened 2021-08-01
Support Multiple DNS period resolving mechanism

### LINKED ISSUE #6222: dns resolve for skywalking trace receiver service addresses

- [X ] Question or discussion
- [ ] Bug
- [ ] Requirement

___
### Question
- What do you want to know?
I want to add dns resolver for the domain of  skywalking trace receiver service
-  WHY ?
   - It's hard to maintain agent config
`collector.backend_service=${SW_AGENT_COLLECTOR_BACKEND_SERVICES:127.0.0.1:11800}
`
   - If extending OAP service, we must restart all service.
  

- HOW ?
  - In this situation, I want to set domain to this variable.The domain has a number of A records
  - modify the method boot() of GRPCChannelManager, check whether it is domain, if so ,then resolve it ,and add all ip addresses to serverList
```
 public void boot() {
        if (Config.Collector.BACKEND_SERVICE.trim().length() == 0) {
            LOGGER.error("Collector server addresses are not set.");
            LOGGER.error("Agent will not uplink any data.");
            return;
        }
        grpcServers = Arrays.asList(Config.Collector.BACKEND_SERVICE.split(","));
        // resolve all domain to ip and add to server List
        connectCheckFuture = Executors.newSingleThreadScheduledExecutor(
            new DefaultNamedThreadFactory("GRPCChannelManager")
        ).scheduleAtFixedRate(
            new RunnableWithExceptionProtection(
                this,
                t -> LOGGER.error("unexpected exception.", t)
            ), 0, Config.Collector.GRPC_CHANNEL_CHECK_INTERVAL, TimeUnit.SECONDS
        );
    }

```
- I know than nginx can help me .But since skywalking already have client loadbalance, I don't want to use nginx, because it could bring some complexity

ThisSeanZhang:
The previous code only supports DNS reverse check of the first domain config
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #6222.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@ Release Notes.
 #### Project
 
 #### Java Agent
-
+* Support Multiple DNS period resolving mechanism
 #### OAP-Backend
 
 #### UI
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ Release Notes.
 
 #### Java Agent
 * Support Multiple DNS period resolving mechanism
+
 #### OAP-Backend
 
 #### UI
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/GRPCChannelManager.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/GRPCChannelManager.java
@@ -32,6 +32,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.skywalking.apm.agent.core.boot.BootService;
 import org.apache.skywalking.apm.agent.core.boot.DefaultImplementor;
@@ -40,6 +41,7 @@ import org.apache.skywalking.apm.agent.core.conf.Config;
 import org.apache.skywalking.apm.agent.core.logging.api.ILog;
 import org.apache.skywalking.apm.agent.core.logging.api.LogManager;
 import org.apache.skywalking.apm.util.RunnableWithExceptionProtection;
+import org.apache.skywalking.apm.util.StringUtil;
 
 import static org.apache.skywalking.apm.agent.core.conf.Config.Collector.IS_RESOLVE_DNS_PERIODICALLY;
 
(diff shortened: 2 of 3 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/GRPCChannelManager.java
```diff
@@ -99,20 +101,28 @@ public void shutdown() {
...
-                grpcServers = newGrpcServers;
-            } catch (Throwable t) {
-                LOGGER.error(t, "Failed to resolve {} of backend service.", backendService);
-            }
+            grpcServers = Arrays.stream(Config.Collector.BACKEND_SERVICE.split(","))
+                    .filter(StringUtil::isNotBlank)
+                    .map(eachBackendService -> eachBackendService.split(":"))
+                    .filter(domainPortPairs -> {
+                        if (domainPortPairs.length < 2) {
+                            LOGGER.debug(new IllegalArgumentException(), "Service address [{}] format error", domainPortPairs[0]);
```
wu-sheng, line 109: Don't log debug with a new exception, this is pointless.

CHANGES.md
```diff
@@ -8,7 +8,7 @@ Release Notes.
 #### Project
 
 #### Java Agent
-
+* Support Multiple DNS period resolving mechanism
 #### OAP-Backend
```
wu-sheng, line 12:
```suggestion
* Support Multiple DNS period resolving mechanism

#### OAP-Backend
```

wu-sheng (commented):
Besides the code level review, what is the case of multiple DNS addresses?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/GRPCChannelManager.java
wu-sheng:
```suggestion
                            LOGGER.debug("Service address [{}] format error. The expected format is IP:port", domainPortPairs[0]);
```

ThisSeanZhang:
Avoid inertial thinking configuration,When Config DNS. Or Same IP but diff Port.

wu-sheng (approved):

kezhenxu94 (approved):

Issue #7397, opened 2021-07-31
Could not using browser to refresh UI.


mrproliu:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [x] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Bug
- **Which version of SkyWalking, OS, and JRE?**
- Browser: Chrome 92.0.4515.107
- Official demo environment: http://demo.skywalking.apache.org

- **What happened?**
- When using the browser to refresh UI, it returns the `404` error code. If I use UI tabs to switch or refresh it works fine.

wu-sheng:
@JaredTan95 I think you missed this in the webapp tech stack change.

JaredTan95:
I clicked `refresh` button in the chrome browser,  seems no problem.

wu-sheng:
Button works. This means browser level refresh.

wu-sheng:
Such as you are at topology, and use f5 to refresh, 404 found.

JaredTan95:
Okay, I see.

wu-sheng:
We got community report about this too :)

PR #7403, opened 2021-08-03
fix not found error when refresh ui.

JaredTan95:
fix not found error when refresh ui.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/7397.


wu-sheng (approved):

mrproliu (approved):

apm-webapp/src/main/java/org/apache/skywalking/apm/webapp/GlobalErrorWebExceptionHandler.java
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.web.server.ServerWebExchange;
+import reactor.core.publisher.Mono;
+
+@Slf4j
+@Order(-1)
+@Configuration
+public class GlobalErrorWebExceptionHandler implements ErrorWebExceptionHandler {
+
+    @Override
+    public Mono<Void> handle(ServerWebExchange exchange, Throwable ex) {
```
kezhenxu94, line 40: Do we just return the index page for all web exception? Or should we only do that when the exception is 404 not found?

JaredTan95: I think both are fine, it's just a little more complicated to handle the special other exceptions.

kezhenxu94: This may cause infinite refresh. If there is any non-404 error in the front end request (for example 500), it returns the index page and the requests are sent again and the page is infinitely refreshed / redirected.

wu-sheng: @JaredTan95 I think we need a check about the 500 code case, if infinitely refreshing happens, we need to fix it.

PR #7417, opened 2021-08-06
Openapi-definitions file loader fix

### LINKED ISSUE #7414: Information on Openapi based Endpoint Grouping

- [x] Question or discussion
- [  ] Bug
- [  ] Requirement
- [  ] Feature or performance improvement

___
### Question
- What do you want to know?
1 - Are there any limits on openapi yaml files? I have used EndpointGroupingRuleReader4OpenapiTest and that class does not load my openapi definition which is over 3000 lines. I figured checkstyle is limiting the definition file limits. Can i increase these values via ENV variable? 

2 - I couldn't tell if skywalking-server correctly loaded my openapi file or not based on logs. I think there should be logging if no file loaded etc.

3 - getGroupRegex method does not care for statically added server context paths. (For example endpoint is /records but server is setup as /api/records). This information usually goes to servers.url in openapi definition (server.url being '/api' in this case). Should we consider such setup?

onurccn:
- [  ] Add a unit test to verify that the fix works.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7414 .

This fixes openapi-definitions under the server-bootstrap module resources being included in server-boostrap.jar. Because of this issue one can not attach their own openapi-definitions folder as classloader will always return resources under server-bootstrap jar.

### DIFF
--- a/oap-server/server-bootstrap/pom.xml
+++ b/oap-server/server-bootstrap/pom.xml
@@ -280,6 +280,7 @@
                         <exclude>fetcher-prom-rules/</exclude>
                         <exclude>envoy-metrics-rules/</exclude>
                         <exclude>meter-analyzer-config/</exclude>
+                        <exclude>openapi-definitions/</exclude>
                         <exclude>otel-oc-rules/</exclude>
                         <exclude>ui-initialized-templates/</exclude>
                         <exclude>zabbix-rules/</exclude>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@ Release Notes.
 * Add `event` http receiver
 * Support Metric level function `serviceRelation` in `MAL`. 
 * Support envoy metrics binding into the topology.
+* Fix openapi-definitions folder not being read correctly.
 
 #### UI
 
--- a/apm-dist-es7/src/main/assembly/binary-es7.xml
+++ b/apm-dist-es7/src/main/assembly/binary-es7.xml
@@ -63,6 +63,7 @@
                 <include>envoy-metrics-rules/*.yaml</include>
                 <include>meter-analyzer-config/*.yaml</include>
                 <include>zabbix-rules/*.yaml</include>
+                <include>openapi-definitions/*</include>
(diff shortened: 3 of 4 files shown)


oap-server/server-bootstrap/pom.xml
```diff
@@ -280,6 +280,7 @@
                         <exclude>fetcher-prom-rules/</exclude>
                         <exclude>envoy-metrics-rules/</exclude>
                         <exclude>meter-analyzer-config/</exclude>
+                        <exclude>openapi-definitions/</exclude>
```
wu-sheng, line 283: I think `exclude` is not enough.

> org.apache.skywalking.oap.server.library.module.ModuleStartException: path not found: openapi-definitions

This error makes the booting fails. You should modify these two to create an empty folder at least.

https://github.com/apache/skywalking/blob/4e7d80fb88a603e076f6c53ed9817d81dc55da3d/apm-dist/src/main/assembly/binary.xml#L53-L69
https://github.com/apache/skywalking/blob/4e7d80fb88a603e076f6c53ed9817d81dc55da3d/apm-dist-es7/src/main/assembly/binary-es7.xml#L53-L69

onurccn: I have seen these but i wasn't sure if this would create empty folder. 
So do you think `<include>openapi-definitions/*</include>` would fix that? I'm going to check that before pushing.

wu-sheng: I think this could fix. Although the including file(in the bootstrap module) is only a demo

onurccn: I have finally build whole distribution and confirmed this creates folder with demo service. Also i was reluctant if this include affects server-bootstrap jar files but it does not. So thanks for the feedback

wankai123:
wait for e2e

wu-sheng (approved):
LGTM. Wait for CI and @wankai123 's confirmation.

wankai123 (approved):

PR #7428, opened 2021-08-10
Add component id for go-kratos & go-mysql

kagaya85:

part of https://github.com/SkyAPM/go2sky-plugins/pull/28


wu-sheng (commented):
Please update the changes.md in the root, and you should add mapping in this file, to avoid half instruction for client side only.

wu-sheng:
@kagaya85 I updated the changelog for you. Please take a look about how to write a changelog, which should include what you did for this repo, rather than the context. This main repo's maintainers/users don't care of what we extend in another subproject.

wu-sheng (dismissed):

wu-sheng (approved):

Issue #7425, opened 2021-08-09
I configured the  ip , but it didn't work


zympig:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [ ] Feature or performance improvement

___
### Question
- What do you want to know?
![image](https://user-images.githubusercontent.com/26918258/128678459-533d58ac-8378-4264-a1de-bf1ee34db3a2.png)
___
### Bug
- Which version of SkyWalking, OS, and JRE?

- Which company or project?

- What happened?
If possible, provide a way to reproduce the error. e.g. demo application, component version.

___
### Requirement or improvement
- Please describe your requirements or improvement suggestions.

zympig:
8.7.0

wu-sheng:
Don't know what are you saying.

zympig:
ES version： 7.14.0 ，SkyWalking version ：8.7.0-es7

I configured  gRPC reporter like this
> 
![image](https://user-images.githubusercontent.com/26918258/128699279-fc37f5d6-deb3-48f5-b5cb-30ce7f8fa718.png)


SkyWalking  OAP  error :
> `[89]: index [sw_log-20210809], type [_doc], id [18909303f26e43f397511a0d313adfcd], message [ElasticsearchException[Elasticsearch exception [type=mapper_parsing_exception, reason=Cannot copy field [endpoint_name] to fields [endpoint_name_match]. Copy-to currently only works for value-type fields, not objects.]]]
[90]: index [sw_log-20210809], type [_doc], id [8bace155dee445bfa4a06a36a37433c9], message [ElasticsearchException[Elasticsearch exception [type=mapper_parsing_exception, reason=Cannot copy field [endpoint_name] to fields [endpoint_name_match]. Copy-to currently only works for value-type fields, not objects.]]]
[92]: index [sw_log-20210809], type [_doc], id [cc45386c565f49228ebc52a3115c6261], message [ElasticsearchException[Elasticsearch exception [type=mapper_parsing_exception, reason=Cannot copy field [endpoint_name] to fields [endpoint_name_match]. Copy-to currently only works for value-type fields, not objects.]]]
[...]

PR #7429, opened 2021-08-10
Fix issues in ElasticSearch 7.14 and add log cases to ES7.14

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/7425.

### DIFF
--- a/.github/workflows/e2e.log.yaml
+++ b/.github/workflows/e2e.log.yaml
@@ -39,7 +39,7 @@ jobs:
     timeout-minutes: 90
     strategy:
       matrix:
-        storage: ['h2', 'mysql', 'es6', 'es7', 'influxdb']
+        storage: ['h2', 'mysql', 'es6', 'es7', 'es7.14', 'influxdb']
     env:
       SW_STORAGE: ${{ matrix.storage }}
     steps:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,7 @@ Release Notes.
 * Trace segment wouldn't be recognized as a TopN sample service. Add through #4694 experimentally, but it caused
   performance impact.
 * Remove `version` and `endTime` in the segment entity. Reduce indexing payload. 
+* Fix `mapper_parsing_exception` in ElasticSearch 7.14.
 
 #### UI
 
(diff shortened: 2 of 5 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/segment/SegmentRecord.java
```diff
@@ -119,7 +121,9 @@ public String id() {
             map.put(TRACE_ID, storageData.getTraceId());
             map.put(SERVICE_ID, storageData.getServiceId());
             map.put(SERVICE_INSTANCE_ID, storageData.getServiceInstanceId());
-            map.put(ENDPOINT_NAME, storageData.getEndpointName());
+            if (nonNull(storageData.getEndpointName())) {
+                map.put(ENDPOINT_NAME, storageData.getEndpointName());
+            }
             map.put(ENDPOINT_ID, storageData.getEndpointId());
```
wu-sheng, line 127: I think this should not be added. Once this has error, the code already has bug. Trace should not have no endpoint.

oap-server/server-storage-plugin/storage-zipkin-elasticsearch7-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/zipkin/ZipkinSpanRecord.java
```diff
@@ -120,7 +122,9 @@ public String id() {
             map.put(SPAN_ID, storageData.getSpanId());
             map.put(SERVICE_ID, storageData.getServiceId());
             map.put(SERVICE_INSTANCE_ID, storageData.getServiceInstanceId());
-            map.put(ENDPOINT_NAME, storageData.getEndpointName());
+            if (nonNull(storageData.getEndpointName())) {
+                map.put(ENDPOINT_NAME, storageData.getEndpointName());
+            }
```
wu-sheng, line 127: I think this should not be added. Once this has error, the code already has bug. Trace should not have no endpoint.

wu-sheng (commented):
Let's limit the not null in the certain scopes.

kezhenxu94:
> Let's limit the not null in the certain scopes.

The nullability check is added according to whether they are annotated with `org.apache.skywalking.oap.server.core.storage.annotation.Column#matchQuery` that will add a `copy_to` field, which causes the problem. If we skip some of the cases mentioned in the comments, the root cause (`serviceName` being null) is hidden behind the ElasticSearch level message, although `serviceName` is expected to be non null,

wu-sheng:
> > Let's limit the not null in the certain scopes.
> 
> The nullability check is added according to whether they are annotated with `org.apache.skywalking.oap.server.core.storage.annotation.Column#matchQuery` that will add a `copy_to` field, which causes the problem. If we skip some of the cases mentioned in the comments, the root cause (`serviceName` being null) is hidden behind the ElasticSearch level message, although `serviceName` is expected to be non null,

The changes on the log side are reasonable, and should be done. But why we bother to check these `should-not-null` fields? If you want to check and expose the cause, check on the `source#prepare` stage rather than here.

wu-sheng (approved):

PR #7431, opened 2021-08-10
Banyandb integration

### LINKED ISSUE #7412: [Trace] Integration with OAP
This task intends to link OAP to banyandb based on the[ interaction API ](https://github.com/apache/skywalking-banyandb/blob/main/api/proto/banyandb/v1/rpc.proto).

Since the banyandb only supports trace data for now, we should:

 * create a branch, named `banyandb-integration` in main repo.
 * add a new storage module named `banyandb`.
 * send trace data to banyandb in the new module, while sending reset data to h2.

^^^ @wu-sheng anythings else should be included?

lujiajing1126:
### Add an OAP storage plugin to support [Skywalking-BanyanDB](https://github.com/apache/skywalking-banyandb)

- [x] UTs have been added to cover query build and write entity build (IT and E2E may not be included in this preliminary version)

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7412 

### DIFF
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -63,5 +63,6 @@ header:
     - '**/src/main/proto/protoc-gen-swagger/**'
     - '**/src/main/proto/validate/validate.proto'
     - '**/src/main/proto/opencensus/**'
+    - '**/*.textproto'
 
   comment: on-failure
--- a/oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBSchema.java
+++ b/oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBSchema.java
@@ -1,0 +1,101 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.oap.server.storage.plugin.banyandb;
+
+import com.google.common.io.CharStreams;
+import com.google.protobuf.TextFormat;
+import org.apache.skywalking.banyandb.Query;
+import org.apache.skywalking.banyandb.Schema;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.util.HashSet;
+import java.util.Set;
+
+public class BanyanDBSchema {
+    public static final Schema.TraceSeries SCHEMA;
+
+    private static int DurationIndex;
+    private static int TraceStateIndex;
+    private static Schema.FieldSpec.FieldType TraceStateType;
+    private static int valIntError;
+    private static String valStringError;
+
+    private static Set<String> fields = new HashSet<>();
+
+    static {
+        // read schema
+        Schema.TraceSeries _schema;
+        try {
+            InputStream schemaFile = BanyanDBStorageProvider.class.getClassLoader().getResourceAsStream("trace_series.textproto");
+            if (schemaFile != null) {
+                _schema = TextFormat.parse(CharStreams.toString(new InputStreamReader(schemaFile, StandardCharsets.UTF_8)), Schema.TraceSeries.class);
+            } else {
+                _schema = null;
+            }
+        } catch (IOException ignore) {
+            _schema = null;
+        }
+        if (_schema == null) {
+            throw new RuntimeException("cannot find schema");
+        }
+        SCHEMA = _schema;
+        for (int i = 0; i < SCHEMA.getFieldsCount(); i++) {
+            final Schema.FieldSpec spec = SCHEMA.getFields(i);
+            fields.add(spec.getName());
+            if (SCHEMA.getReservedFieldsMap().getState().getField().equals(spec.getName())) {
+                TraceStateIndex = i;
+                TraceStateType = spec.getType();
+            } else if ("duration".equals(spec.getName())) {
+                DurationIndex = i;
+            }
+        }
+        switch (TraceStateType) {
+            case FIELD_TYPE_INT:
+                valIntError = Integer.parseInt(SCHEMA.getReservedFieldsMap().getState().getValError());
+                break;
+            case FIELD_TYPE_STRING:
+                valStringError = SCHEMA.getReservedFieldsMap().getState().getValError();
+                break;
+            default:
+                throw new RuntimeException("invalid traceState type");
+        }
+    }
+
+    public static boolean isErrorEntity(Query.Entity entity) {
+        Query.TypedPair pair = entity.getFields(TraceStateIndex);
+        switch (TraceStateType) {
+            case FIELD_TYPE_INT:
+                return pair.getIntPair().getValues(0) == valIntError;
+            case FIELD_TYPE_STRING:
+            default:
+                return valStringError.equals(pair.getStrPair().getValues(0));
+        }
+    }
+
+    public static int getDuration(Query.Entity entity) {
+        return (int) entity.getFields(DurationIndex).getIntPair().getValues(0);
+    }
+
+    public static Set<String> getFieldNames() {
+        return fields;
+    }
+}
--- a/oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBSchema.java
+++ b/oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBSchema.java
@@ -20,6 +20,7 @@ package org.apache.skywalking.oap.server.storage.plugin.banyandb;
 
 import com.google.common.io.CharStreams;
 import com.google.protobuf.TextFormat;
+import org.apache.skywalking.banyandb.Database;
 import org.apache.skywalking.banyandb.Query;
 import org.apache.skywalking.banyandb.Schema;
 
@@ -32,15 +33,14 @@ import java.util.Set;
 
 public class BanyanDBSchema {
     private final Schema.TraceSeries traceSeries;
+    private final Set<String> fields = new LinkedHashSet<>();
 
     private int durationIndex;
     private int traceStateIndex;
     private Schema.FieldSpec.FieldType traceStateType;
     private int valIntError;
     private String valStringError;
-
-    private Set<String> fields = new LinkedHashSet<>();
-
+    
     public static BanyanDBSchema fromTextProtoResource(String filename) {
         // read schema
         Schema.TraceSeries schema;
@@ -104,5 +104,9 @@ public class BanyanDBSchema {
 
     public Set<String> getFieldNames() {
         return this.fields;
+    }
+
+    public Database.Metadata getMetadata() {
+        return this.traceSeries.getMetadata();
     }
 }
--- a/oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBSchema.java
+++ b/oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBSchema.java
@@ -28,15 +28,18 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
 import java.util.LinkedHashSet;
+import java.util.Map;
 import java.util.Set;
 
 public class BanyanDBSchema {
     private final Schema.TraceSeries traceSeries;
     private final Set<String> fields = new LinkedHashSet<>();
+    private final Map<String, Integer> fieldsIndexMapping = new HashMap<>();
 
-    private int durationIndex;
     private int traceStateIndex;
+    private int traceIdIndex;
     private Schema.FieldSpec.FieldType traceStateType;
     private int valIntError;
     private String valStringError;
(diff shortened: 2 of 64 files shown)


lujiajing1126:
Request review @hanahmily @wu-sheng @kezhenxu94

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBStorageProvider.java
```diff
@@ -0,0 +1,68 @@
...
+    public void prepare() throws ServiceNotProvidedException, ModuleStartException {
+        // first call super to initialize with H2
+        super.prepare();
+        // prepare Trace related modules
+        // 1) IBatchDAO
+        // 2) StorageDao
+        // 3) ITraceQueryDAO
+        this.registerServiceImplementation(IBatchDAO.class, new BanyanDBBatchDAO(this.getH2Client(), this.config));
+        this.registerServiceImplementation(StorageDAO.class, new BanyanDBStorageDAO(getManager(), this.getH2Client(), this.config, this.schema));
+        this.registerServiceImplementation(ITraceQueryDAO.class, new BanyanDBTraceQueryDAO(this.config, this.schema));
```
wu-sheng, line 66: H2 storage is using its own config, are you sure this kind of inheritance works?

wu-sheng (commented):
Please fix CI processes. And you need to
1. Update application.yml to introduce this hybrid storage.
2. Please don't change anything in another storage option, you should copy all from its provider to your provider. Because we will see they are all going to be removed eventually.
3. e2e tests for your new storage is required, when the BanyanDB trace core id testable.
4. Please don't add the whole banyandb codebase into a Java project. It would be chaos. Copy the proto files only. They are in the SkyWalking umbrella, they are safe to copy, until we have BanyanDB proto and Java API repositories one day.

.licenserc.yaml
```diff
@@ -63,5 +63,6 @@ header:
     - '**/src/main/proto/protoc-gen-swagger/**'
     - '**/src/main/proto/validate/validate.proto'
     - '**/src/main/proto/opencensus/**'
+    - '**/*.textproto'
```
wu-sheng, line 66: This is incorrect move. 2 questions
1. Why `textproto` rather than proto? What format are we using?
2. Why don't add comments for ASF LICENSE header?

wu-sheng:
> Please don't change anything in another storage option, you should copy all from its provider to your provider. Because we will see they are all going to be removed eventually.

I think you get me wrong about this point. I said `copy all from its provider to your provider`, not `copy all codes`. It is still OK to depend on the H2 storage in this phrase. My point is, don't change the source codes of H2 storage, because you are doing a POC level testing integration.

wu-sheng:
> e2e tests for your new storage is required, when the BanyanDB trace core id testable.

This is not provided yet.

And your official codes, BanyanDB integration, should have comments.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBSchema.java
```diff
@@ -0,0 +1,154 @@
...
+public class BanyanDBSchema {
+    private final Schema.TraceSeries traceSeries;
+    private final Set<String> fields = new LinkedHashSet<>();
+    private final Map<String, Integer> fieldsIndexMapping = new HashMap<>();
+
+    private int traceStateIndex;
+    private int traceIdIndex;
+    private Schema.FieldSpec.FieldType traceStateType;
+    private int valIntError;
+    private String valStringError;
```
wu-sheng, line 45: I am not sure how you organize the codes, what are the principle, but why this `BanyanDBSchema` directly have so many trace fields? What does it represent? If this is a schema, trace model should be a sub-class, isn't it?

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/dao/BanyanDBBatchDAO.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.skywalking.oap.server.storage.plugin.banyandb.client.BanyanDBInsertRequest;
+import org.apache.skywalking.oap.server.storage.plugin.jdbc.SQLExecutor;
+import org.apache.skywalking.oap.server.storage.plugin.jdbc.h2.dao.H2BatchDAO;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Slf4j
+public class BanyanDBBatchDAO extends H2BatchDAO {
+    private final DataCarrier<PrepareRequest> dataCarrier;
```
wu-sheng, line 41: I don't think you need a new dataCarrier. You have a streaming client. The reason H2 has to add one like this, because we don't want to open transaction for every record, but you don't face the same issue here.

wu-sheng: You are just using H2Dao to cover the blank APIs, so, it wouldn't concern about the performance. Let's remove the queue until we have to add it back. Take a look at the ElasticSearch DAOs implementation. Those are really product level, which process 100+ million traces/logs per seconds.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/dao/BanyanDBRecordDAO.java
```diff
@@ -0,0 +1,60 @@
...
+
+    BanyanDBRecordDAO(ModuleManager manager, JDBCHikariCPClient h2Client, StorageHashMapBuilder<Record> storageBuilder, BanyanDBStorageConfig config, BanyanDBSchema schema) {
+        super(manager, h2Client, storageBuilder, config.getMaxSizeOfArrayColumn(), config.getNumOfSearchableValuesPerTag());
+        this.schema = schema;
+        this.mapper = new BanyanDBSchemaMapper(schema.getFieldNames());
+    }
+
+    @Override
+    public InsertRequest prepareBatchInsert(Model model, Record record) throws IOException {
+        if (record instanceof SegmentRecord) {
```
wu-sheng, line 50: @hanahmily A head up here, in SkyWalking, we don't separate logs, topN sampled data and trace, as they are all in huge size and streaming process without L1/L2 routine concept. Does BanyanDB have to do this through `instanceof`? It wouldn't cause much in JDK8+, but it is causing resources. 
Could you talk about this soon?

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/client/BanyanDBSchemaMapper.java
```diff
@@ -0,0 +1,119 @@
...
+        } else if (value instanceof Integer) {
+            return Write.Field.newBuilder().setInt(Write.Int.newBuilder().setValue((Integer) value).build()).build();
+        } else if (value instanceof Long) {
+            return Write.Field.newBuilder().setInt(Write.Int.newBuilder().setValue((Long) value).build()).build();
+        } else {
+            throw new IllegalStateException("should not reach here");
+        }
+    }
+
+    static Map<String, Object> entity2Storage(SegmentRecord segmentRecord) {
```
wu-sheng, line 100: I don't like this code style. You have all the things after `SegmentRecord#Builder#entity2Storage`. Why build another Map?

wu-sheng:
One important suggestion before I am going deeply in code review.

As a database client(we may need to release it separate one day), we need to organize codes and keep the concept consistent.
- Schema is usually referencing a database.
- Table/Index/Entity is the object we are going to write and read.
- Client is the one controlling the connection. We could have typically SimpleClient or BatchClient to deal with different data model. Or add QueryClient in the future.
- All read/write/query should be organized on the top of different entity, which is the normal case. Take a reference from InfluxDB or ElasticSearch client. Such as, TraceWrite should be an object could host the fields and tags. It also could be accepted by BatchClient only.

All above are not mandatory, but I hope these could inspire you that, this is a very serious set of APIs, we need to make it good.

wu-sheng:
Notice, all protobuf and gRPC definition are network protocol, like MySQL and ElasticSearch JSON format. What we need in SkyWalking are a set of Java BanyanDB Client APIs. 
If needed, we could do review on those APIs first, but change all the codes. Feel free to set up a discussion on GitHub issue, mail list or meeting. 
We need to take this very seriously. @hanahmily Please set this discussion up.

lujiajing1126:
> Notice, all protobuf and gRPC definition are network protocol, like MySQL and ElasticSearch JSON format. What we need in SkyWalking are a set of BanyanDB Java Client APIs.
> If needed, we could do review on those APIs first, rather than change all the codes. Feel free to set up a discussion on GitHub issue, mail list or meeting.
> We need to take this very seriously. @hanahmily Please set this discussion up.

Sure. We will setup a meeting tomorrow morning (11 Aug.) and have a discussion first.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/client/BanyanDBSchemaMapper.java
wu-sheng: If there is really a necessary to override this, you should do this through `StorageDAO#newRecordDao`. Builder has been given as the parameter(`StorageHashMapBuilder`). You could read generic type to get the real metrics/record type, and build a required DAO.

SkyWalking's modularization is powerful to override. Don't do hijack and hardcode.

.licenserc.yaml
lujiajing1126: The `textproto` is one of the `traceSeries` definitions in BanyanDB. This is used to determine the schema at runtime.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBSchema.java
lujiajing1126: Since `traceState` and `traceId` are defined as reserved fields, in particular the `traceState`, it could be either String or Int. I have to parse it while the plugin is starting.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/dao/BanyanDBBatchDAO.java
lujiajing1126: Make sense to me. I was confused about this Batch API...

.licenserc.yaml
wu-sheng: What we did in BanyanDB, we doesn't care here. What is the schema, we doesn't care here. 

SkyWalking is a user for BanyanDB, we only care about the HTTP client APIs, and network protocol. That is all.
Keep this in mind, these are all we cared about. These are low level APIs.
https://github.com/apache/skywalking-banyandb/blob/b7845b0cd9fbebf71d3fc990ad6eae5f3ec16771/api/proto/banyandb/v1/rpc.proto#L29-L30

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/client/BanyanDBSchemaMapper.java
lujiajing1126: The basic ideas here are,

1. We do not need to encode/decode binary in Base64 format.
2. Some of the tags have to be treated as indexed field as they are defined explicity in the schema.

lujiajing1126:
> One important suggestion before I am going deeply in code review.
> 
> As a database client(we may need to release it separate one day), we need to organize codes and keep the concept consistent.
> 
> * Schema is usually referencing a database.
> * Table/Index/Entity is the object we are going to write and read.
> * Client is the one controlling the connection. We could have typically SimpleClient or BatchClient to deal with different data model. Or add QueryClient in the future.
> * All read/write/query should be organized on the top of different entity, which is the normal case. Take a reference from InfluxDB or ElasticSearch client. Such as, TraceWrite should be an object could host the fields and tags. It also could be accepted by BatchClient only.
> 
> All above are not mandatory, but I hope these could inspire you that, this is a very serious set of APIs, we need to make it good.

I completely agree with you that we could have provided a well-encapsulated, clearly modularized, hierarchical Java Client for BanyanBD.

Since I am not familiar with the OAP server and its APIs, I temporarily provide a POC to make it work. We may discuss some more details tomorrow.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/client/BanyanDBSchemaMapper.java
wu-sheng: If you take a look at what I mentioned, all these could be done there.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/dao/BanyanDBBatchDAO.java
wu-sheng: Batch API means streaming write, which is why we designed in proto level.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBSchema.java
wu-sheng: Let's talk tomorrow. I am pretty sure this is not necessary.

wu-sheng: The whole schema is BanyanDB's internal thing. If there is version type convert, it is builder's responsibility. https://github.com/apache/skywalking/pull/7431#discussion_r686052303 is the key.

wu-sheng:
> I completely agree with you that we could have provided a well-encapsulated, clearly modularized, hierarchical Java Client for BanyanBD.
Since I am not familiar with the OAP server and its APIs, I temporarily provide a POC to make it work. We may discuss some more details tomorrow.

The key issue right now, is not familiar with SkyWalking or not. Instead, you are mixing SkyWalking and BanyanDB.

lujiajing1126:
> > Sure. We will setup a meeting tomorrow morning (11 Aug.) and have a discussion first.
> 
> 
> 
> @lujiajing1126 please share the meeting link

https://meeting.tencent.com/dm/9rnGuYvqjTCj

PR #7435, opened 2021-08-10
Fix missing Kratos component id

kezhenxu94:
https://github.com/apache/skywalking/pull/7428 added mappings but no target component id


wu-sheng:
According to https://github.com/apache/skywalking-rocketbot-ui/pull/525#issuecomment-896054274, please fix https://github.com/apache/skywalking/pull/7428.

wu-sheng (approved):

PR #7438, opened 2021-08-11
[Break Change] Remove endpoint name in the trace query condition.

wu-sheng:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,10 +22,12 @@ Release Notes.
 * Fix openapi-definitions folder not being read correctly.
 * Trace segment wouldn't be recognized as a TopN sample service. Add through #4694 experimentally, but it caused
   performance impact.
-* Remove `version` and `endTime` in the segment entity. Reduce indexing payload. 
+* Remove `version` and `endTime` in the segment entity. Reduce indexing payload.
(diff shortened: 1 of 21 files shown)


kezhenxu94 (dismissed):

wankai123 (approved):

wangtr028:
Hello developer, 
when I use skywalking, I added the search term endpointName to the page named “trace”. Then I found that skywalking used to have this function, and it was removed in this version. I would take the liberty to ask, is it because there is a major bug or is it better to deal with it? Because I think searching by endpointName on the page is a good function.

wu-sheng:
> Hello developer, 
> 
> when I use skywalking, I added the search term endpointName to the page named “trace”. Then I found that skywalking used to have this function, and it was removed in this version. I would take the liberty to ask, is it because there is a major bug or is it better to deal with it? Because I think searching by endpointName on the page is a good function.

Read the discussion page, it is there.

wangtr028:
> > Hello developer,
> > when I use skywalking, I added the search term endpointName to the page named “trace”. Then I found that skywalking used to have this function, and it was removed in this version. I would take the liberty to ask, is it because there is a major bug or is it better to deal with it? Because I think searching by endpointName on the page is a good function.
> 
> Read the discussion page, it is there.

I found it! thanks for your reply.

PR #7441, opened 2021-08-11
update jdk 11 in dockerfile and remove unused java_opts.

JaredTan95:
as https://github.com/apache/skywalking-docker/pull/22#discussion_r681776376 mentioned.

### DIFF
--- a/docker/agent/Dockerfile.agent
+++ b/docker/agent/Dockerfile.agent
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG BASE_IMAGE='adoptopenjdk/openjdk11:alpine'
+ARG BASE_IMAGE='adoptopenjdk/openjdk8:alpine'
 
 FROM alpine as build
 
--- a/docker/agent/Dockerfile.agent
+++ b/docker/agent/Dockerfile.agent
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG BASE_IMAGE='adoptopenjdk/openjdk8:alpine'
+ARG BASE_IMAGE='adoptopenjdk/openjdk11:alpine'
 
 FROM alpine as build
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Release Notes.
 ------------------
 
 #### Project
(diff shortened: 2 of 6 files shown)


docker/agent/Dockerfile.agent
```diff
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG BASE_IMAGE='adoptopenjdk/openjdk8:alpine'
+ARG BASE_IMAGE='adoptopenjdk/openjdk11:alpine'
```
wu-sheng, line 17: Do you upgrade the agent's JDK? I am good for moving OAP to JDK11.

kezhenxu94 (dismissed):

docker/agent/Dockerfile.agent
kezhenxu94: > Do you upgrade the agent's JDK? I am good for moving OAP to JDK11.

This is only a default value and can be overrode, it's good to me, we will finally release multiple Docker images based on different JDK version

JaredTan95: agent image's JDK is not used, right?  except users use agent image as their java application runtime base image. @kezhenxu94

kezhenxu94: > agent image's JDK is not used, right? except users use agent image as their java application runtime base image. @kezhenxu94

Yes right

wu-sheng: I am curious, where do we use this agent image? If not, why upgrade? Do we plan to use it somewhere?
JDK8 is the version we announced the minimal supported JVM version, and still in maintain mode by OpenJDK and vendors. Why are we going rid of it now?

JaredTan95: > JDK8 is the version we announced the minimal supported JVM version

If we have this consideration, then we don't need to update it, I will revert this change.

wu-sheng: Thanks.

wu-sheng (approved):

hanahmily (approved):
LGTM.
BTW, Could you use a released CLI to replace the current building from scratch? It will reduce the building time a lot.

wu-sheng:
Please check why CI fails.

JaredTan95:
> Please check why CI fails.

I re-run failed jobs.

PR #7448, opened 2021-08-12
Banyandb integration

### LINKED ISSUE #7412: [Trace] Integration with OAP

lujiajing1126:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7412 

## New Design

After discussion with @hanahmily and @wu-sheng , I present the following new storage plugin,

The banyanDB client sits in the `org.apache.skywalking.banyandb.client` package,

1. Provide general API for query and write, which are not related to the underlying protobuf anymore,
2. Provide high level client which implementes `BanyanDBService` interface,
3. Also the client does not care about the schema definitions

The plugin is now decoupled from H2/JDBC and other storage plugins. For those features unimplemented yet, blank APIs are given as stubs now.


oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/request/TraceSearchRequest.java
```diff
@@ -0,0 +1,88 @@
...
+    @Singular
+    private final List<TraceSearchQuery> queries;
+
+    // paging parameters: limit & offset
+    private final int limit;
+    private final int offset;
+
+    // query order
+    private final String queryOrderField;
+    private final SortOrder queryOrderSort;
```
wu-sheng, line 48: These should be exposed as `OrderBy` object with 2 fields as constructor.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/request/TraceSearchQuery.java
```diff
@@ -0,0 +1,82 @@
...
+                        .setOp(this.op.getOp())
+                        .setCondition(Query.TypedPair.newBuilder().setIntPair(Query.IntPair.newBuilder().setKey(key).addValues((Long) this.value).build()).build())
+                        .build();
+            default:
+                throw new IllegalStateException("should not reach here");
+        }
+    }
+
+    // String
+    public static TraceSearchQuery Eq(String key, String val) {
```
wu-sheng, line 50: The function naming should follow camel. The first level is lower case.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/request/TraceSearchQuery.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.skywalking.banyandb.Query;
+
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class TraceSearchQuery {
+    private final String key;
+    private final TraceSearchRequest.BinaryOperator op;
+    private final Object value;
+    private final Query.TypedPair.TypedCase typedCase;
+
+    public Query.PairQuery toPairQuery() {
```
wu-sheng, line 32: This seems testing only? Please use `default` scope.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/request/HasMetadata.java
```diff
@@ -0,0 +1,31 @@
...
+package org.apache.skywalking.banyandb.client.request;
+
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.experimental.SuperBuilder;
+
+@SuperBuilder
+@Getter
+@EqualsAndHashCode
+public class HasMetadata {
```
wu-sheng, line 28: This is an improper name. The verb. should not lead a class name. Verb is for method name.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/request/TraceFetchRequest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import lombok.Getter;
+import lombok.Singular;
+import lombok.experimental.SuperBuilder;
+
+import java.util.List;
+
+@SuperBuilder
+@Getter
+public class TraceFetchRequest extends HasMetadata {
```
wu-sheng, line 29: Please break the unnecessary inheritance. Metadata should be client's constructor. In the future, metadata is used for database or tenancy isolation, you should not let a client shared.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/impl/BanyanDBGrpcClient.java
```diff
@@ -0,0 +1,173 @@
...
+
+        observer.onCompleted();
+    }
+
+    @VisibleForTesting
+    public static List<Write.Field> buildWriteFields(List<Object> fieldList) {
+        return fieldList.stream().map(BanyanDBGrpcClient::buildField).collect(Collectors.toList());
+    }
+
+    static Write.Field buildField(Object value) {
```
wu-sheng, line 159: Don't use `instance of`, use overload for different types of parameters, which have much better performance and easier to read.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/impl/BanyanDBGrpcClient.java
```diff
@@ -0,0 +1,173 @@
...
+
+                    @Override
+                    public void onCompleted() {
+                    }
+                });
+
+        for (final TraceWriteRequest entity : data) {
+            try {
+                Write.EntityValue entityValue = Write.EntityValue.newBuilder()
+                        .addAllFields(buildWriteFields(entity.getFields()))
```
wu-sheng, line 139: Don't use like this. `get whole list`, then `iterating`, and `instance of` for every element.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/BanyanDBTraceService.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.apache.skywalking.banyandb.client.request.TraceFetchRequest;
+import org.apache.skywalking.banyandb.client.request.TraceSearchRequest;
+import org.apache.skywalking.banyandb.client.request.TraceWriteRequest;
+import org.apache.skywalking.banyandb.client.response.BanyanDBQueryResponse;
+
+import java.util.List;
+
+public interface BanyanDBTraceService {
+    // read/query
```
wu-sheng, line 29: Your comments are incorrect, no matter from format or contents.

1. `/** */` is recommended way to comment methods, and all parameters should have contents.
2. Your comments are pointless.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBClient.java
```diff
@@ -0,0 +1,125 @@
...
+import org.apache.skywalking.oap.server.library.client.Client;
+import org.apache.skywalking.oap.server.library.client.healthcheck.DelegatedHealthChecker;
+import org.apache.skywalking.oap.server.library.client.healthcheck.HealthCheckable;
+import org.apache.skywalking.oap.server.library.util.HealthChecker;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+public class BanyanDBClient implements Client, BanyanDBService, HealthCheckable {
```
wu-sheng, line 41: Please separate this as 2 parts, one is BanyanDB API, the other is OAP DB client. 

You still expose the gRPC here.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/request/TraceSearchRequest.java
lujiajing1126: Done

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/request/TraceSearchQuery.java
lujiajing1126: Done

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/request/TraceSearchQuery.java
lujiajing1126: No. It is not only used in tests

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/request/HasMetadata.java
lujiajing1126: Class removed

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/request/TraceFetchRequest.java
lujiajing1126: Metadata moved to client

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/impl/BanyanDBGrpcClient.java
lujiajing1126: No `instanceof` now.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/impl/BanyanDBGrpcClient.java
lujiajing1126: Done. Use typed API

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/banyandb/client/BanyanDBTraceService.java
lujiajing1126: Docs are added. all `//` simple comments are removed.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/dao/BanyanDBRecordDAO.java
```diff
@@ -0,0 +1,75 @@
...
+public class BanyanDBRecordDAO implements IRecordDAO {
+    private final StorageHashMapBuilder<Record> storageBuilder;
+
+    @Override
+    public InsertRequest prepareBatchInsert(Model model, Record record) throws IOException {
+        if (SegmentRecord.INDEX_NAME.equals(model.getName())) {
+            SegmentRecord segmentRecord = (SegmentRecord) record;
+            TraceWriteRequest request = TraceWriteRequest.builder()
+                    .dataBinary(segmentRecord.getDataBinary())
+                    .timestampSeconds(record.getTimeBucket())
```
wu-sheng, line 47: Time bucket is never being timestamp.

lujiajing1126:
> I put many comments, but not all. @hanahmily I am not sure what code style is required in BanyanDB codebase, but this PR is not ready from main repo's requirements.
> 
> All API level methods should have formal comments to describe usage, and meaning of parameters and return value. Even include how and where should use this APIs.
> 
> My major requirements are for BanyanDB package, I would like to be very strict about this API review. Feel free even to submit a PR to submit this. Also, I recommend to move all BanyanDB client source codes as the root module of `oap-server`. It should be `org.apache.skywalking:banyandb-java-client`.

I have improved this PR following you comments.

1. The client package has been moved as a module in the `oap-server`
2. The typed API are introduced to avoid runtime type checking for performance issues
3. Docs are added to all interfaces and fields in the client package

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBClient.java
lujiajing1126: Done. gRPC is no longer exposed.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/dao/BanyanDBRecordDAO.java
lujiajing1126: I see. Fixed.

wu-sheng:
I will provide a code prototype(most implementation) for you tonight.
You will see how we build official client jar. 
I hope this could inspire you in the future.

wu-sheng:
@lujiajing1126 I open https://github.com/apache/skywalking/pull/7455/ for you.

lujiajing1126:
> @lujiajing1126 I open #7455 for you.

Sure. I will have a check. Thx

PR #7455, opened 2021-08-13
Add banyandb client APIs with most implementation.

wu-sheng:

@lujiajing1126 This is for you. You could add UTs and finish all TODOs. I don't have enough time to finish all types. But most are done.

The different contents for the proto files, I have discussed with @hanahmily. All these should be forwarded in the BanyanDB server side.

Eventually, before merging into master branch. The client lib would be moved into skywalking-banyandb-java-client repo, and run a separate release.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,10 @@ Release Notes.
 ------------------
 
 #### Project
+
 * Upgrade jdk 11 in dockerfile and remove unused java_opts.
+* DataCarrier changes a `#consume` API to add properties as a parameter to initialize consumer when
+  use `Class<? extends IConsumer<T>> consumerClass`.
 
 #### Java Agent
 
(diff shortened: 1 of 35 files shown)


wu-sheng:
@lujiajing1126 This could be merged once we have confirmation from @hanahmily. Then you could fill the missing part and finish the trace write/read DAOs accordingly.

lujiajing1126:
Got it. I will continue working on this PR. The plugin should also be implement in this PR, right?

wu-sheng:
> Got it. I will continue working on this PR. The plugin should also be implement in this PR, right?

We don't need that. We could merge this directly, as no harm. Then you could do through 2 PRs
1. Finish this client API lib
2. Implement storage plugin

wu-sheng:
We just need to wait for @hanahmily confirmation.

lujiajing1126:
> > Got it. I will continue working on this PR. The plugin should also be implement in this PR, right?
> 
> We don't need that. We could merge this directly, as no harm. Then you could do through 2 PRs
> 
> 1. Finish this client API lib
> 2. Implement storage plugin

Oh, I understand....since I am not able to work on main repo's branches...

hanahmily (approved):
LGTM

wu-sheng:
> > > Got it. I will continue working on this PR. The plugin should also be implement in this PR, right?
> > 
> > 
> > We don't need that. We could merge this directly, as no harm. Then you could do through 2 PRs
> > 
> > 1. Finish this client API lib
> > 2. Implement storage plugin
> 
> Oh, I understand....since I am not able to work on main repo's branches...

That is not important, even other committers can, we don't recommend 2 people working on the same branch.

PR #7459, opened 2021-08-14
Update some fields

hanahmily:
* Use general types to replace paired types
 * Update package names according to proto lint rules.

### DIFF
--- a/oap-server/banyandb-java-client/src/main/proto/banyandb/v1/banyandb.proto
+++ b/oap-server/banyandb-java-client/src/main/proto/banyandb/v1/banyandb.proto
@@ -32,44 +32,18 @@ message Metadata {
   string name = 2;
 }
 
-// IntPair in a typed pair with an int64 as values
-message IntPair {
-  string key = 1;
-  int64 value = 2;
-  bool isNull = 3;
-}
-
-// StrPair in a typed pair with a string as values
-message StrPair {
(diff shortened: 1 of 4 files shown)


oap-server/banyandb-java-client/src/main/proto/banyandb/v1/banyandb.proto
```diff
@@ -32,44 +32,18 @@ message Metadata {
...
-  bool isNull = 3;
-}
-
 // Pair is the building block of a record which is equivalent to a key-value pair.
 // In the context of Trace, it could be metadata of a trace such as service_name, service_instance, etc.
 // Besides, other fields/tags are organized in key-value pair in the underlying storage layer.
 // One should notice that the values can be a multi-value.
 message TypedPair {
+  string key = 1;
+  bool  is_null =  2;
```
wu-sheng, line 41: You moved is_null out of time, how does banyandb server reponse when int type is null? What will HasInt include?

hanahmily: has introduced a new null type which contains an underlying field type

wu-sheng (approved):
@lujiajing1126 Please follow this and continue on your PR.

wu-sheng:
I will merge after CI passed. This is caused by server side change.

PR #7462, opened 2021-08-15
Banyandb proto

hanahmily:
Update enums according to https://developers.google.com/protocol-buffers/docs/style#enums


oap-server/banyandb-java-client/src/main/proto/banyandb/v1/banyandb.proto
```diff
@@ -33,10 +33,11 @@ message Metadata {
 }
 
 enum FieldType {
-  FIELD_TYPE_STRING = 0;
-  FIELD_TYPE_INT = 1;
-  FIELD_TYPE_STRING_ARRAY = 2;
-  FIELD_TYPE_INT_ARRAY = 3;
+  FIELD_TYPE_UNSPECIFIED=0;
```
wu-sheng, line 36:
```suggestion
  FIELD_TYPE_UNSPECIFIED = 0;
```

wu-sheng (approved):

PR #7475, opened 2021-08-18
Add a new API to test log analysis language

### LINKED ISSUE #7168: please provide the Log Analysis Language  text  regexp debugger

___
when i use the Log Analysis Language to parser the  unstructured logs. i  have to configure the text regexp .  but I need to change the regular expression many times and then restart the oapsever. It is inefficient to load tests for many times .   the Skywalking can it be provided the regular expression parser debugger  like the kibana grok debugger

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7168 

### DIFF
--- a/oap-server/server-bootstrap/src/main/resources/application.yml
+++ b/oap-server/server-bootstrap/src/main/resources/application.yml
@@ -419,6 +419,11 @@ query:
   selector: ${SW_QUERY:graphql}
   graphql:
     path: ${SW_QUERY_GRAPHQL_PATH:/graphql}
+    # Enable the log testing API to test the LAL.
+    # NOTE: This API evaluates untrusted code on the OAP server.
+    # A malicious script can do significant damage (steal keys and secrets, remove files and directories, install malware, etc).
+    # As such, please enable this API only when you completely trust your users.
+    enableLogTestTool: ${SW_QUERY_GRAPHQL_ENABLE_LOG_TEST_TOOL:false}
 
 alarm:
   selector: ${SW_ALARM:default}
--- a/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/LogTestQuery.java
+++ b/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/LogTestQuery.java
@@ -1,0 +1,128 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.oap.query.graphql.resolver;
+
+import com.coxautodev.graphql.tools.GraphQLQueryResolver;
+import java.util.ArrayList;
+import java.util.Arrays;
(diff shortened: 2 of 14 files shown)


kezhenxu94:
Depends on https://github.com/apache/skywalking-query-protocol/pull/60

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/LogTestQuery.java
```diff
@@ -0,0 +1,128 @@
...
+    private final GraphQLQueryConfig config;
+
+    public LogTestResponse test(LogTestRequest request) throws Exception {
+        if (!config.isEnableLogTestTool()) {
+            throw new IllegalAccessException("LAL debug tool is not enabled");
+        }
+
+        requireNonNull(request, "request");
+        checkArgument(isNotBlank(request.getLog()), "request.log cannot be blank");
+        checkArgument(isNotBlank(request.getDsl()), "request.dsl cannot be blank");
```
wu-sheng, line 61: @Fine0830 When resolving response, please check `errors` in the response JSON.

https://github.com/apache/skywalking/blob/6bb4d35ce80833661d688aa8d4ba4d02e60a5ebc/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/GraphQLQueryHandler.java#L97-L106

Provide some notifications or tips to show error.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/LogTestQuery.java
```diff
@@ -0,0 +1,128 @@
...
+import static java.util.Objects.requireNonNull;
+import static org.apache.skywalking.apm.util.StringUtil.isNotBlank;
+
+@RequiredArgsConstructor
+public class LogTestQuery implements GraphQLQueryResolver {
+    private final ModuleManager moduleManager;
+
+    private final GraphQLQueryConfig config;
+
+    public LogTestResponse test(LogTestRequest request) throws Exception {
```
wu-sheng, line 54: Please add an UT for this method or in the LAL core level. I think we don't need complex e2e for this.

oap-server/server-bootstrap/src/main/resources/application.yml
```diff
@@ -419,6 +419,11 @@ query:
   selector: ${SW_QUERY:graphql}
   graphql:
     path: ${SW_QUERY_GRAPHQL_PATH:/graphql}
+    # Enable the log testing API to test the LAL.
+    # NOTE: This API evaluates untrusted code on the OAP server.
```
wu-sheng, line 423: Could LAL engine disable class call? And limit to the classes we know?

kezhenxu94: Theoretically it can, but there are a lot of workarounds to bypass and we won't be able to confidently enable it by default and this disclaimer is necessary in any point.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/LogTestQuery.java
kezhenxu94: Added

oap-server/server-bootstrap/src/main/resources/application.yml
wu-sheng: We wouldn't enable it in default, but I hope we could provide security as much as possible, because I am planning to enable it on demo env

kezhenxu94: > We wouldn't enable it in default, but I hope we could provide security as much as possible, because I am planning to enable it on demo env

I'd rather track this in a separate issue which might share with all groovy-based DSLs, including MAL.

wu-sheng: It is fine through another PR. I am just concerned before that, we could not use official demo env for @Fine0830 's dev.

kezhenxu94: I'll manage to share a private env for her

wu-sheng (approved):
LGTM

PR #7485, opened 2021-08-18
Harden the security of Groovy-based DSL, MAL and LAL

kezhenxu94:

### DIFF
--- a/oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/DSL.java
+++ b/oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/DSL.java
@@ -18,9 +18,13 @@
 
 package org.apache.skywalking.oap.log.analyzer.dsl;
 
+import com.google.common.collect.ImmutableList;
 import groovy.lang.GroovyShell;
 import groovy.transform.CompileStatic;
 import groovy.util.DelegatingScript;
+import java.lang.reflect.Array;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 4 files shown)


kezhenxu94:
This is not all, I just added some usual cases that I could think of, feel free to add more cases

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/DSL.java
```diff
@@ -53,6 +62,22 @@ public static DSL of(final ModuleManager moduleManager,
                 CompileStatic.class
             );
         cc.addCompilationCustomizers(customizer);
+        final SecureASTCustomizer secureASTCustomizer = new SecureASTCustomizer();
+        secureASTCustomizer.setDisallowedStatements(
+            ImmutableList.<Class<? extends Statement>>builder()
+                         .add(WhileStatement.class)
+                         .add(DoWhileStatement.class)
+                         .add(ForStatement.class)
```
wu-sheng, line 70: Do we need a `If`?

kezhenxu94: > Do we need a `If`?

We need `if` statement, note this is `DisallowedStatements`

wu-sheng: Oh, I see.

wu-sheng (approved):

PR #7487, opened 2021-08-19
Simplify GitHub workflows by using `uses` in `composite` actions

kezhenxu94:
`uses` is recently allowed in `composite` actions, we can simplify our workflows control files further.

- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md). No need


wu-sheng (approved):

kezhenxu94:
Also update `skywalking-eyes` to a recent version

wu-sheng (approved):

PR #7495, opened 2021-08-19
Banyandb storage plugin

lujiajing1126:


wu-sheng:
We have the image, please use infra e2e gha to add a new test.
You could take satellite project e2e as a reference.

wu-sheng:
We only need to verify traces.

.github/workflows/e2e.storages.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
     timeout-minutes: 90
     strategy:
       matrix:
-        storage: ['mysql', 'es6', 'es7.0', 'es7.10', 'es7.14', 'opensearch', 'influxdb', 'tidb', 'postgresql']
+        storage: ['mysql', 'es6', 'es7.0', 'es7.10', 'es7.14', 'opensearch', 'influxdb', 'tidb', 'postgresql', 'banyandb']
```
wu-sheng, line 39: This e2e test would verify all metrics, service register. I think you only make traces working.

lujiajing1126: Sure

.github/workflows/e2e.banyandb.yaml
```diff
@@ -0,0 +1,65 @@
...
+          echo "::group::Copy Dist Package to test/e2e"
+          cp -R dist test/e2e/
+          echo "::endgroup::"
+      - name: Prepare E2E Test
+        shell: bash
+        run: |
+          echo "::group::Install SNAPSHOT apm-application-toolkit"
+          ./mvnw -DskipTests -N install
+          ./mvnw -f apm-application-toolkit -DskipTests -am install
+          echo "::endgroup::"
```
wu-sheng, line 57: You don't need this.

test/e2e/e2e-test/docker/storage/banyandb/expected/traces.yml
```diff
@@ -0,0 +1,24 @@
...
+
+traces:
+  - key: not null
+    endpointNames:
+      - /users
+    duration: ge 0
+    start: gt 0
+    isError: false
+    traceIds:
+      - not null
```
wu-sheng, line 24: You should check all fields of trace to make sure the Database is working as expected.

.github/workflows/e2e.banyandb.yaml
lujiajing1126: I tried to run this locally. But I found that if I need to compile the e2e-provider, I have to install the `apm-application-toolkit` in the local repository.

test/e2e/e2e-test/docker/storage/banyandb/expected/traces.yml
lujiajing1126: Sure. I will complete this later.

.github/workflows/e2e.banyandb.yaml
wu-sheng: @kezhenxu94 Please give some suggestion. I think we need a guide line for new e2e-infra integration, which is already using in this branch.

lujiajing1126:
It seems the workflow I've added did not take effect.

wu-sheng:
> It seems the workflow I've added did not take effect.

I guess we need a separate pull request to inject a no logic GHA control file first, right? @kezhenxu94

kezhenxu94:
> > It seems the workflow I've added did not take effect.
> 
> I guess we need a separate pull request to inject a no logic GHA control file first, right? @kezhenxu94

It did run, only because the file syntax is wrong.

Screen Shot 2021-08-20 at 08 32 50

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/dao/BanyanDBTraceQueryDAO.java
```diff
@@ -0,0 +1,156 @@
...
+            BasicTrace trace = new BasicTrace();
+            trace.setDuration(((Long) entity.getFields().get(3).getValue()).intValue());
+            trace.setStart(String.valueOf(entity.getFields().get(4).getValue()));
+            trace.setSegmentId(entity.getId());
+            trace.setError(((Long) entity.getFields().get(1).getValue()).intValue() == 1);
+            trace.getTraceIds().add((String) entity.getFields().get(0).getValue());
+            trace.getEndpointNames().add(IDManager.EndpointID.analysisId(
+                    (String) entity.getFields().get(2).getValue()
+            ).getEndpointName());
+            return trace;
```
wu-sheng, line 127: Please make sure the codes polished. When you face a 0-N field indices, you should make sure they are ordered.

wu-sheng: And where is the request fields? I am not sure how BanyanDB server implements the logic now when they are loss, but we access them through indices, they must be declared. If not, BanyanDB should response nothing.

lujiajing1126:
@mrproliu 

I am using `${provider_host}` as a placeholder to get IP and port for the services defined in `docker-compose` file. But It seems that all placeholders are returning blank string. Also I've tested on both macOS and Linux.

Is this related to https://github.com/apache/skywalking-infra-e2e/pull/36?

mrproliu:
> @mrproliu
> 
> image
> 
> I am using `${provider_host}` as a placeholder to get IP and port for the services defined in `docker-compose` file. But It seems that all placeholders are returning blank string. Also it can be reproducible on both macOS and Linux.

You need to declare which ports need to export to the environment, please add port declare into `docker-compose.yaml` file, here is a [demo link](https://github.com/apache/skywalking-satellite/blob/main/test/e2e/case/native-protocols/docker-compose.yml#L50-L51). 

> Is this related to [apache/skywalking-infra-e2e#36](https://github.com/apache/skywalking-infra-e2e/pull/36)?

After you change the `docker-compose.yml` port declare, the E2E may still fail, this PR it's to fix this problem, that's two different problems. After this PR has been merge, I will mention you to update the commit id. 

> Another feature request: can we provide an `env` field that allow to pass environments downwards into the container in Github Action environment.

Could you do more explain on this? which environment field do you want?

lujiajing1126:
> > Another feature request: can we provide an `env` field that allow to pass environments downwards into the container in Github Action environment.
> 
> Could you do more explain on this? which environment field do you want?

In the current E2E workflow, `SW_AGENT_JDK_VERSION` is defined to indicate the JDK version we intend to use for the test.

https://github.com/apache/skywalking/blob/master/.github/workflows/e2e.storages.yaml#L24-L25

This variable is used in the `docker-compose.yml` file as a docker-build argument, i.e. `FROM adoptopenjdk/openjdk${SW_AGENT_JDK_VERSION}:alpine`.

https://github.com/apache/skywalking/blob/master/test/e2e/e2e-test/docker/base-compose.yml#L85-L86

and,

https://github.com/apache/skywalking/blob/master/test/e2e/e2e-test/docker/Dockerfile.provider#L16-L19

But this feature is missing in the state-of-art sw-infra-e2e framework according to the `action.yml`.

mrproliu:
> > > Another feature request: can we provide an `env` field that allow to pass environments downwards into the container in Github Action environment.
> > 
> > 
> > Could you do more explain on this? which environment field do you want?
> 
> @mrproliu
> 
> In the current E2E workflow, `SW_AGENT_JDK_VERSION` is defined to indicate the JDK version we intend to use for the test.
> 
> https://github.com/apache/skywalking/blob/cf03304a48a940f5d599b3a2cbd0711b008c17d1/.github/workflows/e2e.storages.yaml#L24-L25
> 
> This variable is used in the `docker-compose.yml` file as a docker-build argument, i.e. `FROM adoptopenjdk/openjdk${SW_AGENT_JDK_VERSION}:alpine`.
> 
> https://github.com/apache/skywalking/blob/cf03304a48a940f5d599b3a2cbd0711b008c17d1/test/e2e/e2e-test/docker/base-compose.yml#L85-L86
> 
> and,
> 
> https://github.com/apache/skywalking/blob/cf03304a48a940f5d599b3a2cbd0711b008c17d1/test/e2e/e2e-test/docker/Dockerfile.provider#L16-L19
> 
> But this feature is missing in the state-of-art sw-infra-e2e framework according to the `action.yml`.
> 
> https://github.com/apache/skywalking-infra-e2e/blob/de5be41596f27867a15ab75dbad97ea538d863b8/action.yaml#L31-L35
> 
> Ping @kezhenxu94 . Do you have other consideration for this scenario? for example this is an intentional design

Thanks for your explain, Do you mean you want to correlate the GitHub Action environment into the Infra-E2E container? I think we missed that, so we can't correlate the GitHub Action Environment into E2E for now. I think we could dump the environment and use `--env-file` to propagate it.

kezhenxu94:
> Ping @kezhenxu94 . Do you have other consideration for this scenario? for example this is an intentional design

I see. All env vars in GitHub actions should be (expected to be) passed downwards to e2e program, but as we use container in GHA the env vars are not implicitly passed so we need to do that manually. @mrproliu Seems more and more problems arise after we use Docker container in GHA.

wu-sheng:
I am feeling we are overheating skywalking-infra-e2e. It is a young and unreleased project, we shouldn't expect its as powerful as others.
If this is not a block, let's create an issue for it to track, and set the env in the docker compose directly.

mrproliu:
>  @mrproliu Seems more and more problems arise after we use Docker container in GHA.

I think it is only unstable now, but as we have more and more cases, we can make it more stable. After we released Infra-E2E and polish the docker image, it was worthwhile because we can speed up E2E testing. WDYT?

lujiajing1126:
> I am feeling we are overheating skywalking-infra-e2e. It is a young and unreleased project, we shouldn't expect its as powerful as others.
> If this is not a block, let's create an issue for it to track, and set the env in the docker compose directly.

Env is not a blocking issue. I've made a workaround.

wu-sheng:
You could submit an issue directly with context, then you or someone else could enhance that young project later.

lujiajing1126:
Another issue report:

```
$ swctl --display yaml --base-url=http://127.0.0.1:49184/graphql trace ls
FATAL graphql: The variables input contains a field name 'endpointName' that is not defined for input object type 'TraceQueryCondition'
```

@wu-sheng is this related to the recent removal of `endpointName` in https://github.com/apache/skywalking/pull/7438

```
$ swctl -v
swctl version 0.7.0
```

wu-sheng:
> Another issue report:
> 
> ```
> $ swctl --display yaml --base-url=http://127.0.0.1:49184/graphql trace ls
> FATAL graphql: The variables input contains a field name 'endpointName' that is not defined for input object type 'TraceQueryCondition'
> ```
> 
> @wu-sheng is this related to the recent removal of `endpointName` in #7438
> 
> ```
> $ swctl -v
> swctl version 0.7.0
> ```

@kezhenxu94 Could you check whether we have removed this from CLI? If not, please do.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/dao/BanyanDBTraceQueryDAO.java
lujiajing1126: > Please make sure the codes polished. When you face a 0-N field indices, you should make sure they are ordered.

Sorted.

lujiajing1126: > And where is the request fields? I am not sure how BanyanDB server implements the logic now when they are loss, but we access them through indices, they must be declared. If not, BanyanDB should response nothing.

What do you mean by request fields here?

test/e2e/e2e-test/docker/storage/banyandb/expected/traces.yml
lujiajing1126: More test cases are added

lujiajing1126:
BanyanDB E2E has passed. Pls have a review again @wu-sheng @hanahmily @kezhenxu94

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBStorageProvider.java
```diff
@@ -0,0 +1,153 @@
...
+        this.registerServiceImplementation(
+                IProfileThreadSnapshotQueryDAO.class, new BanyanDBProfileThreadSnapshotQueryDAO());
+        this.registerServiceImplementation(UITemplateManagementDAO.class, new BanyanDBUITemplateManagementDAO());
+
+        this.registerServiceImplementation(IEventQueryDAO.class, new BanyanDBEventQueryDAO());
+    }
+
+    @Override
+    public void start() throws ServiceNotProvidedException, ModuleStartException {
+        final ConfigService configService = getManager().find(CoreModule.NAME)
```
kezhenxu94, line 127: Unused

kezhenxu94 (commented):
Just some nits

wu-sheng:
@kezhenxu94 Does GitHub provide a snapshot based maven repo? If so, we could keep what we did for agent test, which is, We could create a `banyandb-java-client` repo, and hosts all Java driver related codes, and get a dev maven repo too. In the main repo, we are going to need snapshot dependency only.

After that, we could keep this PR to the master branch(don't need to sync codes anymore), every time this PR gets update, the test could run directly.

@lujiajing1126 @hanahmily What do you think about this, rather than maintaining another branch?

kezhenxu94:
> @kezhenxu94 Does GitHub provide a snapshot based maven repo?

We can do that

hanahmily:
It makes sense to me. Maybe the Banyandb repo is a proper place to hold
client Java codes and build artifacts.

lujiajing1126:
> @kezhenxu94 Does GitHub provide a snapshot based maven repo? If so, we could keep what we did for agent test, which is, We could create a `banyandb-java-client` repo, and hosts all Java driver related codes, and get a dev maven repo too. In the main repo, we are going to need snapshot dependency only.
> 
> After that, we could keep this PR to the master branch(don't need to sync codes anymore), every time this PR gets update, the test could run directly.
> 
> @lujiajing1126 @hanahmily What do you think about this, rather than maintaining another branch?

Make sense to me.

kezhenxu94:
This might be a little late as I didn't review the first few pull requests to this branch `banyandb-integration`, later that I reviewed the merged codes and found that the so-called `banyandb-java-client` is just a simple wrapper of gRPC client, which makes me feel whether we really need to maintain this kind of `java-client`:

- protobuffers itself already serves as a protocol layer / API layer, open source projects usually define protobuffers schema and let users compile and use gRPC to call the services, instead of building another wrapper on top of gRPC.
- when we wrap gRPC in our java-client, it means users have less possibility to customize the gRPC configurations, or, we have to expose gRPC configurations in "our own SDK", which is not a good idea in terms of OOP capsulation. Also, dependency management is what we should consider.
- currently, the `banyandb-java-client` depends on our `apm-datacarrier` and queue the requests ourselves, I think the gRPC itself must have queue internally and know when is the best time to flush the requests.

But, I don't mean the module `banyandb-java-client` is useless, it's useful in oap-server, but I'm kind of concerned whether it's useful for the BanyanDB end users and so is it necessary to maintain and publish it separately .

wu-sheng:
> currently, the banyandb-java-client depends on our apm-datacarrier and queue the requests ourselves, I think the gRPC itself must have queue internally and know when is the best time to flush the requests.

Let me get to this. gRPC and netty have internal queue, but it is for direct memory for sending, which can't avoid OOM :) We have suffered those years ago.

wu-sheng:
> but I'm kind of concerned whether it's useful for the BanyanDB end users and so is it necessary to maintain and publish it separately .

We don't know, but when you dig deeper, you will notice. BanyanDB Java client is not only for providing client implementation, it is scoping the client proto files. All those proto files in the server side are not all cared of by the client. We show how to integrate. If end users will(long future will) run some integration, they could use the proto files in client folder, and go further.

wu-sheng:
Every client is a wrapper for a protocol, like you are doing our new ElasticSearch-protocol client, or what Elastic team did and we are using right now. This is not a condition we should or should not have a client repository. :)

wu-sheng:
> Every client is a wrapper for a protocol, like you are doing our new ElasticSearch-protocol client, or what Elastic team did and we are using right now. This is not a condition we should or should not have a client repository. :)

The value of client repository is making people easier to understand and use the network protocol. Think in this way, our trace protocol is out there for years. But you and only several agents' founders have/had read/understood the protocol field by field. But thousands users out there, they care about language-specific APIs only.

wu-sheng:
> protobuffers itself already serves as a protocol layer / API layer, open source projects usually define protobuffers schema and let users compile and use gRPC to call the services, instead of building another wrapper on top of gRPC.

That is and only for professional cross open source projects integration. Like we analysis Envoy protobuf, and have our complex solution for mesh observability. But that is a very rare case and for a small subset of all open source users.
We need those proto, and we need more.

kezhenxu94:
> We don't know, but when you dig deeper, you will notice. BanyanDB Java client is not only for providing client implementation, it is scoping the client proto files. All those proto files in the server side are not all cared of by the client. We show how to integrate. If end users will(long future will) run some integration, they could use the proto files in client folder, and go further.

IMHO that's our responsibility to make it clear in repo apache/skywalking-banyandb or apache/skywalking-banyandb-api (instead of in banyandb-java-client repo) what are client side `.proto` files and what are server side `.proto` files, even we have Java client, it should not take the responsibility to scope the client `.proto` files, after all, users may use other programming languages.

> That is and only for professional cross open source projects integration. Like we analysis Envoy protobuf, and have our complex solution for mesh observability. But that is a very rare case and for a small subset of all open source users.
We need those proto, and we need more.

That's my point, BanyanDB is an APM-specific database, not general-purpose database, I just thought that there won't be many ordinary users integrate it.

I'm totally OK to have `banyandb-java-client` repo, just wanted to raise the discussion above as soon as I notice it.

lujiajing1126:
> > currently, the banyandb-java-client depends on our apm-datacarrier and queue the requests ourselves, I think the gRPC itself must have queue internally and know when is the best time to flush the requests.
> 
> Let me get to this. gRPC and netty have internal queue, but it is for direct memory for sending, which can't avoid OOM :) We have suffered those years ago.

BTW, I've recently found a project, https://github.com/salesforce/reactive-grpc. 

As I understand, according to the document, this project can properly handle the so-called back-pressure issue which is similar to what we provide thru `apm-data-carrier`. And it seems to be more capable since it is deeply integrated to the transport layer, i.e. HTTP/2 flow control.

wu-sheng:
> IMHO that's our responsibility to make it clear in repo apache/skywalking-banyandb or apache/skywalking-banyandb-api (instead of in banyandb-java-client repo) what are client side .proto files and what are server side .proto files, even we have Java client, it should not take the responsibility to scope the client .proto files, after all, users may use other programming languages.

We should not. But we are for short term. We have not gotten the clear boundary between client/network proto and server internal proto. They are sharing. We will have `apache/skywalking-banyandb-api ` eventually, but not now.

> That's my point, BanyanDB is an APM-specific database, not general-purpose database, I just thought that there won't be many ordinary users integrate it.

I think it wouldn't. But who knows, like many things we did, we didn't expect they are being used widely. But you found yesterday e2e partial code copying case, who knows. SkyWalking is being watched by countless projects' maintainers and users.

wu-sheng:
> BTW, I've recently found a project, https://github.com/salesforce/reactive-grpc.
> 
> As I understand, according to the document, this project can properly handle the so-called back-pressure issue which is similar to what we provide thru `apm-data-carrier`. And it seems to be more capable since it is deeply integrated to the transport layer, i.e. HTTP/2 flow control.

We could try if you want. But you have to resolve the dependency tree, and notice, that is also a vendor-base OSS with very limited activities(https://github.com/salesforce/reactive-grpc/pulse/monthly). Saleforce doesn't have very famous and popular OSS libs, we may face out-of-maintain case.

wu-sheng:
Repo created https://github.com/apache/skywalking-banyandb-java-client

wu-sheng:
@lujiajing1126 I am going to move the client codebase in your fork into that repo. Is that OK? We are still waiting for banyandb server side change right?

lujiajing1126:
> @lujiajing1126 I am going to move the client codebase in your fork into that repo. Is that OK? We are still waiting for banyandb server side change right?

I think it's OK to move now. Yes, we are still waiting for the new index module.

wu-sheng:
@lujiajing1126 All source codes are there. Once @kezhenxu94 have time to set up snapshot maven deploy in the CI process, we could use it directly. 

This PR should be closed. When everything(BanyanDB server) is ready, we should start one targeting to master branch.

kezhenxu94:
We can also deploy those snapshots to maven central via github workflow, which is allowed in infra , if I remember correctly

wu-sheng:
> If that's unacceptable, we can also deploy those snapshots to maven central via github workflow, which is allowed in infra , if I remember correctly

You could do deploy on https://repository.apache.org/, but not Maven central.

kezhenxu94:
> > If that's unacceptable, we can also deploy those snapshots to maven central via github workflow, which is allowed in infra , if I remember correctly
> 
> You could do deploy on https://repository.apache.org/, but not Maven central.

Ticket is created to request infra team to set up the needed token https://issues.apache.org/jira/browse/INFRA-22286

kezhenxu94:
@lujiajing1126 snapshot is deployed, you need to add the following snippet to main repo (`oap-server-bom` / `oap-server-bom-es7`) when you want to reference it

```xml
    <repositories>
        <repository>
            <id>apache-snapshot</id>
            <name>Apache Snapshot</name>
            <url>https://repository.apache.org/content/repositories/snapshots</url>
        </repository>
    </repositories>
```

```xml
        <dependency>
            <artifactId>banyandb-java-client</artifactId>
            <version>0.1.0-SNAPSHOT</version>
        </dependency>
```

lujiajing1126:
> @lujiajing1126 snapshot is deployed, you need to add the following snippet to main repo (`oap-server-bom` / `oap-server-bom-es7`) when you want to reference it
> 
> ```
> <project>
> ...
>     <repositories>
>         <repository>
>             <id>apache-snapshot</id>
>             <name>Apache Snapshot</name>
>             <url>https://repository.apache.org/content/repositories/snapshots</url>
>         </repository>
>     </repositories>
> ```
> 
> and use it in the `pom.xml` where you want to use it
> 
> ```
>         <dependency>
>             <groupId>org.apache.skywalking</groupId>
>             <artifactId>banyandb-java-client</artifactId>
>             <version>0.1.0-SNAPSHOT</version>
>         </dependency>
> ```

Got it. Thanks

PR #7498, opened 2021-08-19
Update license.

totalo:


kezhenxu94 (approved):

.mvn/wrapper/MavenWrapperDownloader.java
```diff
@@ -1,18 +1,21 @@
 /*
- * Copyright 2007-present the original author or authors.
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
wu-sheng, line 2: Why change back to ASF?

wu-sheng (changes requested):

.mvn/wrapper/MavenWrapperDownloader.java
wu-sheng: https://github.com/apache/shardingsphere/blob/c2df0553596ccd24646fafe1ab11b63baa81c8d3/.mvn/wrapper/MavenWrapperDownloader.java#L2-L14

ShardingSphere is using this too.

wu-sheng: At https://github.com/apache/shardingsphere/pull/11883#discussion_r691303345, you said, the `Licensed to the Apache Software Foundation (ASF) under one or more` comes from SkyWalking, but from this change log, SkyWalking has the correct and original header.

Could you share what is missing here?

totalo: Sorry, I think I might understand it wrong.

wu-sheng: OK, get it.

Issue #7081, opened 2021-06-08
Dynamic configuration for a set of configurations.


wu-sheng:
Currently, `ConfigChangeWatcher` only allows implementors to register a key for a dynamic configuration update. With more configurations being dynamic, we may not provide all keys, such as all MAL scripts.

@wankai123 @mrproliu @kezhenxu94 @dmsolr or other @apache/skywalking-committers do we have any ideas about how to add another dynamic configuration API for a set of data?

I don't think we have to keep compatible with the existing one, the keys about this feature are 
1. Watcher could indicate a name/key for multiple items.
2. Every item of the name/key could be changed/updated separately, which means it could be maintained separately.

dmsolr:
Not watcher entirely configuration file anymore? And replacing it with all keys of configuration items.
If so, the result was `alarm-settings.yaml` + `alarm.default.alarm-settings.xxxx` (e.g. rules), righ?

wu-sheng:
Not replace, another new option. I think not all configuration requiring a list.

kezhenxu94:
> With more configurations being dynamic, we may not provide all keys, such as all MAL scripts.

Using MAL as an example, do you mean every single rule 👇 (like `heap_memory_used`) can be updated separately?

https://github.com/apache/skywalking/blob/f54f639c4d9207c20490519121eeabe6acf37374/oap-server/server-bootstrap/src/main/resources/envoy-metrics-rules/envoy.yaml#L34-L59

If this is what you proposed, the config keys would be too many to manage

wu-sheng:
> > With more configurations being dynamic, we may not provide all keys, such as all MAL scripts.
> 
> 
> 
> Using MAL as an example, do you mean every single rule 👇 (like `heap_memory_used`) can be updated separately?
> 
> 
> 
> https://github.com/apache/skywalking/blob/f54f639c4d9207c20490519121eeabe6acf37374/oap-server/server-bootstrap/src/main/resources/envoy-metrics-rules/envoy.yaml#L34-L59
> 
> 
> 
> If this is what you proposed, the config keys would be too many to manage

I don't mean expression. The case is like, MAL script files could be modified and added in the runtime. But we don't know the file name, a.k.a. key name, unless it is added.
So, we have to use a key called OC-MAL-SCRIPTS as a key of the whole collection.

kezhenxu94:
[...]

PR #7509, opened 2021-08-20
Support collection type in dynamic configuration core and add zookeeper implementation 

wankai123:
### Support collection type in dynamic configuration core and add zookeeper implementation

related to #7081

add `GroupConfigTable` and  `GroupConfigChangeWatcher` to watch grouped dynamic configurations

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -44,7 +44,7 @@ Release Notes.
 * Add a new API to test log analysis language.
 * Harden the security of Groovy-based DSL, MAL and LAL.
 * Fix distinct in Service/Instance/Endpoint query is not working.
-* Support grouped dynamic configurations in DCS.
+* Support collection type in dynamic configuration core.
 * Support zookeeper grouped dynamic configurations.
 
 #### UI
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -44,6 +44,8 @@ Release Notes.
 * Add a new API to test log analysis language.
 * Harden the security of Groovy-based DSL, MAL and LAL.
 * Fix distinct in Service/Instance/Endpoint query is not working.
+* Support grouped dynamic configurations in DCS.
+* Support zookeeper grouped dynamic configurations.
 
 #### UI
 
--- a/oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigChangeWatcher.java
+++ b/oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigChangeWatcher.java
@@ -62,15 +62,8 @@ public abstract class ConfigChangeWatcher {
     public static class ConfigChangeEvent {
         private String newValue;
         private EventType eventType;
-        private String groupItemName;
 
         public ConfigChangeEvent(String newValue, EventType eventType) {
-            this.newValue = newValue;
-            this.eventType = eventType;
-        }
-
-        public ConfigChangeEvent(String groupItemName, String newValue, EventType eventType) {
-            this.groupItemName = groupItemName;
             this.newValue = newValue;
             this.eventType = eventType;
         }
--- a/oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigChangeWatcher.java
+++ b/oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigChangeWatcher.java
@@ -37,7 +37,7 @@ public abstract class ConfigChangeWatcher {
         this.module = module;
         this.provider = provider;
         this.itemName = itemName;
-        this.watchType = WatchType.SIMPLE;
+        this.watchType = WatchType.SINGLE;
     }
 
     /**
@@ -81,6 +81,6 @@ public abstract class ConfigChangeWatcher {
     }
 
     public enum WatchType {
-        SIMPLE, GROUP
+        SINGLE, GROUP
     }
 }
--- a/oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigChangeWatcher.java
+++ b/oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigChangeWatcher.java
@@ -31,11 +31,13 @@ public abstract class ConfigChangeWatcher {
     private final String module;
     private final ModuleProvider provider;
     private final String itemName;
+    protected WatchType watchType;
 
     public ConfigChangeWatcher(String module, ModuleProvider provider, String itemName) {
         this.module = module;
         this.provider = provider;
         this.itemName = itemName;
+        this.watchType = WatchType.SIMPLE;
     }
 
     /**
@@ -60,8 +62,15 @@ public abstract class ConfigChangeWatcher {
     public static class ConfigChangeEvent {
         private String newValue;
         private EventType eventType;
+        private String groupItemName;
 
         public ConfigChangeEvent(String newValue, EventType eventType) {
+            this.newValue = newValue;
+            this.eventType = eventType;
+        }
+
+        public ConfigChangeEvent(String groupItemName, String newValue, EventType eventType) {
+            this.groupItemName = groupItemName;
             this.newValue = newValue;
             this.eventType = eventType;
         }
@@ -69,5 +78,9 @@ public abstract class ConfigChangeWatcher {
 
     public enum EventType {
         ADD, MODIFY, DELETE
+    }
+
+    public enum WatchType {
+        SIMPLE, GROUP
     }
 }
--- a/oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigWatcherRegister.java
+++ b/oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigWatcherRegister.java
@@ -25,15 +25,14 @@ import java.util.Set;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.skywalking.apm.util.RunnableWithExceptionProtection;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * The default implementor of Config Watcher register.
  */
+@Slf4j
 public abstract class ConfigWatcherRegister implements DynamicConfigurationService {
-    private static final Logger LOGGER = LoggerFactory.getLogger(ConfigWatcherRegister.class);
     public static final String LINE_SEPARATOR = System.getProperty("line.separator", "\n");
     private Register singleConfigChangeWatcherRegister = new Register();
     @Getter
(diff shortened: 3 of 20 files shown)


oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigChangeWatcher.java
```diff
@@ -60,14 +62,25 @@ public String toString() {
...
+            this.eventType = eventType;
+        }
     }
 
     public enum EventType {
         ADD, MODIFY, DELETE
     }
+
+    public enum WatchType {
+        SIMPLE, GROUP
```
wu-sheng, line 84:
```suggestion
        SINGLE, GROUP
```

I think SINGLE is a better word?

oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigWatcherRegister.java
```diff
@@ -110,6 +110,67 @@ void configSync() {
 
             LOGGER.trace("Current configurations after the sync." + LINE_SEPARATOR + register.toString());
         });
+
+        GroupConfigsSync(configTable);
+    }
+
+    private void GroupConfigsSync(Optional<ConfigTable> configTable) {
+        configTable.ifPresent(config -> {
```
kezhenxu94, line 118: Try to reduce the nested levels

```suggestion
        configTable.ifPresent(this::groupConfigsSync);
    }

    private void groupConfigsSync(ConfigTable config) {
```

Also the first letter should be lowercase

oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigWatcherRegister.java
```diff
@@ -110,6 +110,67 @@ void configSync() {
...
+            config.getGroupItems().forEach(groupConfigItems -> {
+                String groupConfigItemName = groupConfigItems.getName();
+                WatcherHolder holder = register.get(groupConfigItemName);
+                if (holder != null) {
```
kezhenxu94, line 122: Use fast return to reduce the nested level

```suggestion
                if (holder == null) {
                    return;
                }
```

oap-server/server-configuration/configuration-zookeeper/src/main/java/org/apache/skywalking/oap/server/configuration/zookeeper/ZookeeperConfigWatcherRegister.java
```diff
@@ -46,9 +51,28 @@ public ZookeeperConfigWatcherRegister(ZookeeperServerSettings settings) throws E
     @Override
     public Optional<ConfigTable> readConfig(Set<String> keys) {
         ConfigTable table = new ConfigTable();
-        keys.forEach(s -> {
-            ChildData data = this.childrenCache.getCurrentData(this.prefix + s);
-            table.add(new ConfigTable.ConfigItem(s, data == null ? null : new String(data.getData())));
+        keys.forEach(key -> {
+            if (key.startsWith(ConfigChangeWatcher.WatchType.GROUP.name())) {
```
kezhenxu94, line 55: What about adding some delimiter like `GROUP::` or `GROUP.` or `GROUP.`?

wu-sheng: I think we have the enough metadata in the memory, why do we need to check from name's prefix? @wankai123

wankai123: yes, It can get from `WatcherHolder` by changing the class authority. we don't need to check from the name's prefix.
and then should we keep the name's prefix `GROUP` to identify it's a `GROUP key`? @wu-sheng wu

wu-sheng: I think we don't need to add prefix. The one provides the feature based on this, they should show this clearly on the doc. We could add a new column on the table(dynamic configuration table), which should be more clear.

wu-sheng:
I think the title is incorrect? DCS is the gRPC service for dynamic configuration. I don't see any change related to that. Please confirm.

CHANGES.md
```diff
@@ -44,6 +44,8 @@ Release Notes.
 * Add a new API to test log analysis language.
 * Harden the security of Groovy-based DSL, MAL and LAL.
 * Fix distinct in Service/Instance/Endpoint query is not working.
+* Support grouped dynamic configurations in DCS.
```
wu-sheng, line 47:
```suggestion
* Support collection type in dynamic configuration core.
```
I think this is your real meaning.

oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigWatcherRegister.java
```diff
@@ -82,34 +82,96 @@ void configSync() {
...
-                        if (watcher.value() != null) {
-                            // Notify watcher, the new value is null with delete event type.
-                            watcher.notify(
-                                new ConfigChangeWatcher.ConfigChangeEvent(null, ConfigChangeWatcher.EventType.DELETE));
-                        } else {
-                            // Don't need to notify, stay in null.
-                        }
+                if (holder == null) {
+                    LOGGER.warn(
+                        "Config {} from configuration center, doesn't match any WatchType.SINGLE watcher, ignore.",
```
wu-sheng, line 87: Is this going to be group type? Why do we need a warning log here?

oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigWatcherRegister.java
```diff
@@ -82,34 +82,96 @@ void configSync() {
...
+        configTable.ifPresent(this::groupConfigsSync);
+    }
+
+    private void groupConfigsSync(ConfigTable config) {
+        config.getGroupItems().forEach(groupConfigItems -> {
+            String groupConfigItemName = groupConfigItems.getName();
+            WatcherHolder holder = register.get(groupConfigItemName);
+
+            if (holder == null || holder.getWatcher().watchType != ConfigChangeWatcher.WatchType.GROUP) {
+                LOGGER.warn(
```
wu-sheng, line 126: Same question about why warning?

wu-sheng: You are reading according to `config.getGroupItems()`. Why there is a chance falling into this case?

oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigWatcherRegister.java
```diff
@@ -82,34 +82,96 @@ void configSync() {
...
+                    // Notify watcher, the item is deleted with delete event type.
+                    watcher.notify(
+                        new ConfigChangeWatcher.ConfigChangeEvent(oldGroupItemName,
+                                                                  null, ConfigChangeWatcher.EventType.DELETE
+                        ));
+                }
+            });
+        });
+
+        LOGGER.trace("Current configurations after the sync." + LINE_SEPARATOR + register.toString());
```
wu-sheng, line 174: There are 2 `Current configurations after the sync.` logs. Even the old one, it use the `+` without `LOGGER.isTraceEnable`, which has performance impact.(very little, but exists.

Besides performance, the log should be more clear about more clear, what exactly is the final status after sync.

oap-server/server-configuration/configuration-zookeeper/src/main/java/org/apache/skywalking/oap/server/configuration/zookeeper/ZookeeperConfigWatcherRegister.java
```diff
@@ -46,9 +51,28 @@ public ZookeeperConfigWatcherRegister(ZookeeperServerSettings settings) throws E
     @Override
     public Optional<ConfigTable> readConfig(Set<String> keys) {
         ConfigTable table = new ConfigTable();
-        keys.forEach(s -> {
-            ChildData data = this.childrenCache.getCurrentData(this.prefix + s);
-            table.add(new ConfigTable.ConfigItem(s, data == null ? null : new String(data.getData())));
+        keys.forEach(key -> {
+            if (this.getRegister().get(key).getWatcher().getWatchType() == ConfigChangeWatcher.WatchType.GROUP) {
```
wu-sheng, line 55: I think about enum, we should use `ConfigChangeWatcher.WatchType.GROUP.equals`

wu-sheng: Make `this.getRegister().get(key).getWatcher().getWatchType()` a parent method please.

oap-server/server-configuration/configuration-zookeeper/src/main/java/org/apache/skywalking/oap/server/configuration/zookeeper/ZookeeperConfigWatcherRegister.java
```diff
@@ -46,9 +51,28 @@ public ZookeeperConfigWatcherRegister(ZookeeperServerSettings settings) throws E
     @Override
     public Optional<ConfigTable> readConfig(Set<String> keys) {
```
wu-sheng, line 52: Please add the comments about what are the structure of the single and group value in the Zookeeper.

oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigWatcherRegister.java
```diff
@@ -75,45 +88,127 @@ public void start() {
...
+                        watcher.notify(new ConfigChangeWatcher.ConfigChangeEvent(
+                            newItemValue,
+                            ConfigChangeWatcher.EventType.MODIFY
+                        ));
+                    } else {
+                        // Don't need to notify, stay in the same config value.
+                    }
+                }
+            });
+            if (LOGGER.isTraceEnabled()) {
```
wu-sheng, line 131: `LOGGER` is not recommended, please change to `@Slf4j`.

wu-sheng (commented):
General good, just some nits.

wu-sheng (approved):

kezhenxu94 (approved):

oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigWatcherRegister.java
```diff
@@ -55,16 +56,28 @@ synchronized public void registerConfigChangeWatcher(ConfigChangeWatcher watcher
...
+                throw new IllegalArgumentException(
+                    "Unexpected watch type of ConfigChangeWatcher " + watcher.toString());
+        }
     }
 
     public void start() {
         isStarted = true;
 
-        LOGGER.info("Current configurations after the bootstrap sync." + LINE_SEPARATOR + register.toString());
+        LOGGER.info("Current configurations after the bootstrap sync." + LINE_SEPARATOR + singleConfigChangeWatcherRegister.toString());
```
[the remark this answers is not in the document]
wu-sheng: @sonatype-lift ingore

PR #7524, opened 2021-08-21
Use skywalking-eyes GHA instead of manually compile

kezhenxu94:
Also use `quiet` mode for all maven commands in GHA to avoid too MANY logs when we want to spot some outlier cases, we can still see errors when there is failure. And the compile logs never expose real errors

### DIFF
--- a/.github/actions/e2e-test/action.yml
+++ b/.github/actions/e2e-test/action.yml
@@ -44,15 +44,7 @@ runs:
         fi
         echo "::endgroup::"
     - name: Check License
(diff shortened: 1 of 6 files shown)


wu-sheng (approved):

wu-sheng:
Let's see how much time `-q` could help

kezhenxu94:
> Let's see how much time `-q` could help

I don't expect `-q` to save time, but want to get rid of the verbose logs when finding problems.

PR #7526, opened 2021-08-21
Kafka extra producer configuration enhancement

dmsolr:

### DIFF
--- a/apm-sniffer/config/agent.config
+++ b/apm-sniffer/config/agent.config
@@ -216,12 +216,12 @@ plugin.dubbo.consumer_provider_length_threshold=${SW_PLUGIN_DUBBO_CONSUMER_PROVI
 plugin.kafka.bootstrap_servers=${SW_KAFKA_BOOTSTRAP_SERVERS:localhost:9092}
 #  Timeout period of reading topics from the Kafka server, the unit is second. 
 plugin.kafka.get_topic_timeout=${SW_GET_TOPIC_TIMEOUT:10}
-#  Kafka producer configuration.
-plugin.kafka.consumer_config=${SW_PLUGIN_KAFKA_CONSUMER_CONFIG:}
 # Kafka producer configuration. Read [producer configure](http://kafka.apache.org/24/documentation.html#producerconfigs)
 # to get more details. Check [Kafka report doc](How-to-enable-kafka-reporter.md) for more details and examples.
-plugin.kafka.producer_config=${SW_PLUGIN_KAFKA_PRODUCER_CONFIG:}
-#  Specify which Kafka topic name for Meter System data to report to. 
+plugin.kafka.producer_config=${sw_plugin_kafka_producer_config:}
+# Configure Kafka Producer configuration in JSON format. Notice it will be overridden by plugin.kafka.producer_config[key], if the key duplication.
+plugin.kafka.producer_config_json=${sw_plugin_kafka_producer_config_json:}
+#  Specify which Kafka topic name for Meter System data to report to.
 plugin.kafka.topic_meter=${SW_PLUGIN_KAFKA_TOPIC_METER:skywalking-meters}
 #  Specify which Kafka topic name for JVM metrics data to report to. 
 plugin.kafka.topic_metrics=${SW_PLUGIN_KAFKA_TOPIC_METRICS:skywalking-metrics}
--- a/apm-sniffer/config/agent.config
+++ b/apm-sniffer/config/agent.config
@@ -220,7 +220,7 @@ plugin.kafka.get_topic_timeout=${SW_GET_TOPIC_TIMEOUT:10}
 # to get more details. Check [Kafka report doc](How-to-enable-kafka-reporter.md) for more details and examples.
 plugin.kafka.producer_config=${sw_plugin_kafka_producer_config:}
 # Configure Kafka Producer configuration in JSON format. Notice it will be overridden by plugin.kafka.producer_config[key], if the key duplication.
-plugin.kafka.producer_config_json=${sw_plugin_kafka_producer_config_json:}
+plugin.kafka.producer_config_json=${SW_PLUGIN_KAFKA_PRODUCER_CONFIG_JSON:}
 #  Specify which Kafka topic name for Meter System data to report to.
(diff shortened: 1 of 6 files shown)


apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/src/main/java/org/apache/skywalking/apm/agent/core/kafka/KafkaProducerManager.java
```diff
@@ -102,6 +104,10 @@ public void run() {
         Properties properties = new Properties();
         properties.setProperty(
             ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, KafkaReporterPluginConfig.Plugin.Kafka.BOOTSTRAP_SERVERS);
+
+        Map<String, String> producerConfig = (Map<String, String>) new Gson()
```
kezhenxu94, line 108: Extract `new Gson()` as a field

kezhenxu94 (commented):
Documentation and `agent.conf` should be updated, and indicate that if a property is set in both `PRODUCER_CONFIG` and `PRODUCER_CONFIG_JSON`, the later overrides the former

docs/en/setup/service-agent/java-agent/How-to-enable-kafka-reporter.md
```diff
@@ -12,8 +12,10 @@ collector.backend_service=${SW_AGENT_COLLECTOR_BACKEND_SERVICES:127.0.0.1:11800}
 
 # Kafka producer configuration
 plugin.kafka.bootstrap_servers=${SW_KAFKA_BOOTSTRAP_SERVERS:localhost:9092}
-plugin.kafka.producer_config[delivery.timeout.ms]=12000
 plugin.kafka.get_topic_timeout=${SW_GET_TOPIC_TIMEOUT:10}
+
+# Configure extra Kafka Producer configuration
+plugin.kafka.producer_config[delivery.timeout.ms]=12000
```
wu-sheng, line 18: I think comments in the `KafkaReporterPluginConfig` should be a separate section in this doc. 
It is better to remove `plugin.kafka.producer_config[delivery.timeout.ms]=12000` from here.

Adding `Advanced Kafka Configs` maybe better

docs/en/setup/service-agent/java-agent/How-to-enable-kafka-reporter.md
```diff
@@ -12,8 +12,10 @@ collector.backend_service=${SW_AGENT_COLLECTOR_BACKEND_SERVICES:127.0.0.1:11800}
 
 # Kafka producer configuration
```
wu-sheng, line 13: One missing change

> In other words, the following configure cannot be omitted.

These words should be updated. All configurations have been activated in the `agent.config`.

apm-sniffer/config/agent.config
```diff
@@ -216,12 +216,12 @@ plugin.dubbo.consumer_provider_length_threshold=${SW_PLUGIN_DUBBO_CONSUMER_PROVI
...
 plugin.kafka.get_topic_timeout=${SW_GET_TOPIC_TIMEOUT:10}
-#  Kafka producer configuration.
-plugin.kafka.consumer_config=${SW_PLUGIN_KAFKA_CONSUMER_CONFIG:}
 # Kafka producer configuration. Read [producer configure](http://kafka.apache.org/24/documentation.html#producerconfigs)
 # to get more details. Check [Kafka report doc](How-to-enable-kafka-reporter.md) for more details and examples.
-plugin.kafka.producer_config=${SW_PLUGIN_KAFKA_PRODUCER_CONFIG:}
-#  Specify which Kafka topic name for Meter System data to report to. 
+plugin.kafka.producer_config=${sw_plugin_kafka_producer_config:}
+# Configure Kafka Producer configuration in JSON format. Notice it will be overridden by plugin.kafka.producer_config[key], if the key duplication.
+plugin.kafka.producer_config_json=${sw_plugin_kafka_producer_config_json:}
```
wu-sheng, line 223:
```suggestion
plugin.kafka.producer_config_json=${SW_PLUGIN_KAFKA_PRODUCER_CONFIG_JSON:}
```

docs/en/setup/service-agent/java-agent/README.md
```diff
@@ -153,8 +153,8 @@ property key | Description | Default |
 `plugin.dubbo.provider_arguments_length_threshold`| When `plugin.dubbo.collect_provider_arguments` is `true`, Arguments of length from the front will to the OAP backend |`256`| 
 `plugin.kafka.bootstrap_servers`| A list of host/port pairs to use for establishing the initial connection to the Kafka cluster. | `localhost:9092`|
 `plugin.kafka.get_topic_timeout`| Timeout period of reading topics from the Kafka server, the unit is second. |`10`|
-`plugin.kafka.consumer_config`| Kafka producer configuration. ||
```
wu-sheng, line 156: I think this still works, right?

wu-sheng:
@kezhenxu94 I think we should add a system environment variable for the agent setup, as you update `agent.config`.
And we should cut the current `readme.md` into different documents, the page is too long to read.

kezhenxu94:
> @kezhenxu94 I think we should add a system environment variable for the agent setup, as you update `agent.config`.

Yes, it's better to add a column for environment variable name in the agent setup doc

wu-sheng:
> And we should cut the current readme.md into different documents, the page is too long to read.

Create #7532 for further update.

docs/en/setup/service-agent/java-agent/README.md
dmsolr: Not, just forget to remove it from docs. This was originally reserved for `profile task` communication with OAP. But we discard this way. Agent communicate with OAP one way.

dmsolr: Not, just forget to remove it from docs. This was originally reserved for `profile task` communication with OAP. But we discard this way. Agent communicate with OAP one way.

wu-sheng (approved):

kezhenxu94 (approved):
LGTM

PR #7537, opened 2021-08-23
Separate agent docs to avoid long single page.

### LINKED ISSUE #7532: Cut the current java-agent/readme into different documents
We should cut the document into several small documents and update the menu.

wu-sheng:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7532.


wu-sheng:

kezhenxu94 (approved):

PR #7541, opened 2021-08-23
Split CI workflow to fasten the waiting time

kezhenxu94:
Try to split the workflow and speed up CI waiting time


wu-sheng:
I can't see `CI-with-IT` profile. Is this due to draft status?

PR #7549, opened 2021-08-24
Fix browser service(node type)  not queried in getAllServices method.

JaredTan95:
from elasticsearch `service_traffic` index, browser service traffic generated:

```json
{
  "_index": "sw_service_traffic-20210824",
  "_type": "_doc",
  "_id": "dGVzdC11aQ==.1",
  "_score": 1,
  "_source": {
  "node_type": 6,
  "name": "test-ui",
  "service_group": null
}
```

but after browser report traces, we can not query browser service in `Tracing`---> `Service` select


wu-sheng:
I think you missed a key part in browser agent. It would use a new service name for trace, and registration is full completed. No chance needed.

JaredTan95:
> I think you missed a key part in browser agent. It would use a new service name for trace, and registration is full completed. No chance needed.

Okay, I will check it.

PR #7554, opened 2021-08-24
Support configure sampling rate dynamically for service dimension on the backend side

HendSame:
This is a new PR  for  #7194


docs/en/setup/backend/dynamic-config.md
```diff
@@ -12,8 +12,7 @@ Currently, SkyWalking supports the following dynamic configurations.
 |core.default.apdexThreshold| The apdex threshold settings. Overrides `service-apdex-threshold.yml`. | Same as [`service-apdex-threshold.yml`](apdex-threshold.md). |
 |core.default.endpoint-name-grouping| The endpoint name grouping setting. Overrides `endpoint-name-grouping.yml`. | Same as [`endpoint-name-grouping.yml`](endpoint-grouping-rules.md). |
 |core.default.log4j-xml| The log4j xml configuration. Overrides `log4j2.xml`. | Same as [`log4j2.xml`](dynamical-logging.md). |
-|agent-analyzer.default.sampleRate| Trace sampling. Overrides `receiver-trace/default/sampleRate` of `application.yml`. | 10000 |
-|agent-analyzer.default.slowTraceSegmentThreshold| Setting this threshold on latency (in milliseconds) would cause slow trace segments to be sampled if they use up more time, even if the sampling mechanism is activated. The default value is `-1`, which means slow traces will not be sampled. Overrides `receiver-trace/default/slowTraceSegmentThreshold` of `application.yml`. | -1 |
```
wu-sheng, line 16: Why do you remove these two?

wu-sheng: You are removing unrelated things.

oap-server/server-bootstrap/src/main/resources/trace-sample-rate-setting.yml
```diff
@@ -0,0 +1,26 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
wu-sheng, line 1: You add a new file, so you have to change package pom. Otherwise, this file will be in the core jar.

wu-sheng: https://github.com/apache/skywalking/blob/db2bc935737ee24ad79b6f237b0eaa60991434ef/oap-server/server-bootstrap/pom.xml#L267-L291

wu-sheng: https://github.com/apache/skywalking/blob/db2bc935737ee24ad79b6f237b0eaa60991434ef/apm-dist/src/main/assembly/binary.xml#L52-L73

https://github.com/apache/skywalking/blob/db2bc935737ee24ad79b6f237b0eaa60991434ef/apm-dist-es7/src/main/assembly/binary-es7.xml#L52-L73

docs/en/setup/backend/trace-sampling.md
```diff
@@ -8,15 +8,30 @@ segments have been collected and reported by agents, the backend would do their
...
   default:
     ...
-    sampleRate: ${SW_TRACE_SAMPLE_RATE:10000} # The sample rate precision is 1/10000. 10000 means 100% sample in default.
+    # The default sampling rate and the default trace latency time configured by the 'traceSampleRateSettingFile' file.
+    traceSampleRateSettingFile: ${SW_TRACE_SAMPLE_RATE_SETTING_FILE:trace-sample-rate-setting.yml}
     forceSampleErrorSegment: ${SW_FORCE_SAMPLE_ERROR_SEGMENT:true} # When sampling mechanism activated, this config would make the error status segment sampled, ignoring the sampling rate.
-    slowTraceSegmentThreshold: ${SW_SLOW_TRACE_SEGMENT_THRESHOLD:-1} # Setting this threshold about the latency would make the slow trace segments sampled if they cost more time, even the sampling mechanism activated. The default value is `-1`, which means would not sample slow traces. Unit, millisecond.
+```
+
+We can configure sampling rate dynamically not only default globally but for the specified services in `trace-sample-rate-setting.yml`
```
wu-sheng, line 22:
```suggestion
The default `trace-sample-rate-setting.yml` uses the following format.
```

docs/en/setup/backend/dynamic-config.md
```diff
@@ -12,8 +12,7 @@ Currently, SkyWalking supports the following dynamic configurations.
...
+|agent-analyzer.default.traceSampleRateSetting| The sampling rate for default and service dimension, override `trace-sample-rate-setting.yml`. | same as [`trace-sample-rate-setting.yml`](trace-sampling.md) |
```
wu-sheng, line 15: As you are merging configuration, so `traceSampleRateSetting` is a wrong term. Please update it globally to `traceSamplingPolicy`.

docs/en/setup/backend/trace-sampling.md
```diff
@@ -8,15 +8,30 @@ segments have been collected and reported by agents, the backend would do their
...
+default:
+  # Default sampling rate that replaces the 'agent-analyzer.default.sampleRate' since version 8.7.0
+  # The sample rate precision is 1/10000. 10000 means 100% sample in default.
+  sampleRate: 10000
+  # Default trace latency time that replaces the 'agent-analyzer.default.slowTraceSegmentThreshold' since version 8.7.0
+  # Setting this threshold about the latency would make the slow trace segments sampled if they cost more time, even the sampling mechanism activated. The default value is `-1`, which means would not sample slow traces. Unit, millisecond.
+  duration: -1
+services:
+  - name: serverName
+    sampleRate: 1000 # sampling rate of this service which named {name}
```
wu-sheng, line 33:
```suggestion
    rate: 1000 # sampling rate of this service which named {name}
```
You are using `duration` as another threshold name, this should be in the consistent style

docs/en/setup/backend/trace-sampling.md
```diff
@@ -8,15 +8,30 @@ segments have been collected and reported by agents, the backend would do their
...
+    duration: 10000 # trace latency time of this service which named {name}
```
wu-sheng, line 34:
```suggestion
    duration: 10000 # Trace latency threshold for trace sampling for this specific service
```

docs/en/setup/backend/trace-sampling.md
```diff
@@ -8,15 +8,30 @@ segments have been collected and reported by agents, the backend would do their
...
+    # The default sampling rate and the default trace latency time configured by the 'traceSampleRateSettingFile' file.
+    traceSampleRateSettingFile: ${SW_TRACE_SAMPLE_RATE_SETTING_FILE:trace-sample-rate-setting.yml}
     forceSampleErrorSegment: ${SW_FORCE_SAMPLE_ERROR_SEGMENT:true} # When sampling mechanism activated, this config would make the error status segment sampled, ignoring the sampling rate.
-    slowTraceSegmentThreshold: ${SW_SLOW_TRACE_SEGMENT_THRESHOLD:-1} # Setting this threshold about the latency would make the slow trace segments sampled if they cost more time, even the sampling mechanism activated. The default value is `-1`, which means would not sample slow traces. Unit, millisecond.
+```
+
+We can configure sampling rate dynamically not only default globally but for the specified services in `trace-sample-rate-setting.yml`
+```yaml
+default:
+  # Default sampling rate that replaces the 'agent-analyzer.default.sampleRate' since version 8.7.0
```
wu-sheng, line 25: This should be moved into changelog. This is a new file, and not codes, we don't need to remind our history. 
Also, `since 8.7.0` is wrong, 8.7.0 has released, we are running 8.8.0-dev.

docs/en/setup/backend/trace-sampling.md
```diff
@@ -8,15 +8,30 @@ segments have been collected and reported by agents, the backend would do their
...
+  # The sample rate precision is 1/10000. 10000 means 100% sample in default.
+  sampleRate: 10000
+  # Default trace latency time that replaces the 'agent-analyzer.default.slowTraceSegmentThreshold' since version 8.7.0
```
wu-sheng, line 28: Same as https://github.com/apache/skywalking/pull/7554/files#r695368325. This should be moved and corrected.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSampleRateSettingWatcher.java
```diff
@@ -0,0 +1,263 @@
...
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import static java.util.Objects.isNull;
+
+@Slf4j
+public class TraceSampleRateSettingWatcher extends ConfigChangeWatcher {
```
wu-sheng, line 48:
```suggestion
public class TraceSamplePolicyWatcher extends ConfigChangeWatcher {
```

An example of following https://github.com/apache/skywalking/pull/7554/files#r695367288, many others should be changed.

oap-server/analyzer/agent-analyzer/src/test/java/org/apache/skywalking/oap/server/analyzer/provider/trace/ServiceSampleRateSettingWatcherTest.java
```diff
@@ -0,0 +1,140 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import java.util.Optional;
+import java.util.Set;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ServiceSampleRateSettingWatcherTest {
```
wu-sheng, line 36: The watchers have been merged, then please merge the UTs.

wu-sheng: `TraceSamplingPolicyWatcherTest` is required to replace all three

oap-server/analyzer/agent-analyzer/src/test/java/org/apache/skywalking/oap/server/analyzer/provider/trace/UninstrumentedGatewaysConfigTest.java
```diff
@@ -28,6 +28,7 @@
 import org.powermock.reflect.Whitebox;
 
 public class UninstrumentedGatewaysConfigTest {
+
```
wu-sheng, line 31: Unrelated touch.

oap-server/server-bootstrap/src/main/resources/trace-sample-rate-setting.yml
```diff
@@ -0,0 +1,26 @@
...
+  # Default sampling rate that replaces the 'agent-analyzer.default.sampleRate'
+  # The sample rate precision is 1/10000. 10000 means 100% sample in default.
+  sampleRate: 10000
+  # Default trace latency time that replaces the 'agent-analyzer.default.slowTraceSegmentThreshold'
+  # Setting this threshold about the latency would make the slow trace segments sampled if they cost more time, even the sampling mechanism activated. The default value is `-1`, which means would not sample slow traces. Unit, millisecond.
+  duration: -1
+services:
+  - name: serverName
+    rate: 1000 # Sampling rate of this specific service
+    duration: 10000 # Trace latency threshold for trace sampling for this specific service
```
wu-sheng, line 26: For a default configuration file, service specific settings should be kept in commented status.

wu-sheng:
@JaredTan95 Could you run a local test about this? The dynamic configuration seems not having e2e testing. And UTs can't prove all.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,263 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import static java.util.Objects.isNull;
+
+@Slf4j
+public class TraceSamplingPolicyWatcher extends ConfigChangeWatcher {
+    private final AtomicReference<String> settingsString;
+    private AtomicReference<Integer> sampleRate;
+    private AtomicReference<Integer> slowTraceSegmentThreshold;
```
kezhenxu94, line 51:
```suggestion
    private AtomicInteger sampleRate;
    private AtomicInteger slowTraceSegmentThreshold;
```

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,263 @@
...
+        return setting;
+    }
+
+    @Getter
+    @Setter
+    @ToString
+    public static class ServiceSampleConfig {
+        private String name;
+        private AtomicReference<Integer> sampleRate;
+        private AtomicReference<Integer> duration;
```
kezhenxu94, line 233: There is no need to use `AtomicXXXX`, primitives should be enough?

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,263 @@
...
+            return services.iterator();
+        }
+    }
+
+    @Getter
+    @Setter
+    @ToString
+    public static class DefaultSampleRateConfig {
+        private AtomicReference<Integer> sampleRate;
+        private AtomicReference<Integer> duration;
```
kezhenxu94, line 260: No need to use `AtomicXXX`, primitives should be enough

kezhenxu94: And it seems we can make `DefaultSampleRateConfig` immutable, so that no thread safety problem

oap-server/analyzer/agent-analyzer/src/test/resources/trace-sample-rate-setting.yml
```diff
@@ -0,0 +1,25 @@
...
+default:
+  sampleRate: 10000
+  duration: -1
+services:
+  - name: name1
+    rate: 1000
+    duration: 20000
+  - name: name2
+    rate: 2000
+    duration: 30000
```
kezhenxu94, line 25: - Why use `sampleRate` for `default` and `rate` for `services`? I think it better to be consistent.
- I think the following format might be also more consistent.

```yaml
default:
  sampleRate: 10000
  duration: -1
service1:
  sampleRate: 1000
  duration: 20000
name2:
  sampleRate: 2000
  duration: 30000 
```

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/SegmentAnalysisListener.java
```diff
@@ -218,7 +218,7 @@ public AnalysisListener create(ModuleManager moduleManager, AnalyzerModuleConfig
                 namingControl,
                 searchTagKeys,
                 segmentStatusAnalyzer,
-                traceLatencyThresholdsAndWatcher
+                    traceSamplingPolicyWatcher
```
kezhenxu94, line 221:
```suggestion
                traceSamplingPolicyWatcher
```

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
kezhenxu94: This can be immutable class too

wu-sheng:
@HendSame I think you made a mistake in merging. You are adding `skywalking-agent` back, which is legal binary. And you updated UI submodule

wu-sheng:
Please check your changed files.
image

wu-sheng:
UI submodule still change. 
image

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,217 @@
...
+@Slf4j
+public class TraceSamplingPolicyWatcher extends ConfigChangeWatcher {
+
+    private final AtomicReference<String> settingsString = new AtomicReference<>(Const.EMPTY_STRING);
+    private final AtomicReference<SamplePolicySettings> samplePolicySettings = new AtomicReference<>(null);
+    private final AnalyzerModuleConfig moduleConfig;
+
+    public TraceSamplingPolicyWatcher(AnalyzerModuleConfig moduleConfig, ModuleProvider provider) {
+        super(AnalyzerModule.NAME, provider, "traceSamplingPolicy");
+        this.moduleConfig = moduleConfig;
```
wu-sheng, line 53: You don't need to keep `moduleConfig`, just pass the file path into `#loadDefaultPolicy` enough.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/TraceSegmentSampler.java
```diff
@@ -18,20 +18,41 @@
...
+    public boolean shouldSample(SegmentObject segmentObject, int duration) {
+        int sample = Math.abs(segmentObject.getTraceId().hashCode()) % 10000;
+        String serviceName = segmentObject.getService();
+        TraceSamplingPolicyWatcher.SamplePolicy sampleConfig = traceSamplingPolicyWatcher.getSamplePolicy(serviceName);
+        if (sampleConfig != null) {
+            if (shouldSampleService(sampleConfig, sample, duration)) {
+                return true;
+            }
+        }
+        return sample < traceSamplingPolicyWatcher.getSampleRate();
```
wu-sheng, line 44: L39 check and L44 should not exist. You already have fail safe mechanism in the watcher. This is a duplicate mechanism. Also, at the same time, `traceSamplingPolicyWatcher#getSampleRate` is not needed.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,217 @@
...
+
+    public SamplePolicy getSamplePolicy(String service) {
+        return this.samplePolicySettings.get().getPolicy(service);
+    }
+
+    public int getSampleRate() {
+        return this.samplePolicySettings.get().global.rate;
+    }
+
+    public int getSlowTraceSegmentThreshold() {
```
wu-sheng, line 81: This is only used in the test, which is wrong. Only `#getSamplePolicy` should be used in runtime and UTs.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,217 @@
...
+        return this.samplePolicySettings.get().global.duration;
+    }
+
+    public boolean shouldSample(int duration) {
```
wu-sheng, line 85: This method should be removed. `SegmentAnalysisListener` also should sample based on service name. We need consistent logic.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/TraceSegmentSampler.java
```diff
@@ -18,20 +18,41 @@
...
+    }
+
+    private boolean shouldSampleService(TraceSamplingPolicyWatcher.SamplePolicy sampleConfig, int sample, int duration) {
```
wu-sheng, line 47: This method is out of scope. It belongs to `SamplePolicySettings`.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,217 @@
...
+    @NoArgsConstructor
+    public static class SamplePolicy {
+        private Integer rate;
+        private Integer duration;
+    }
+
+    @ToString
+    public static class SamplePolicySettings {
+
+        @Getter
```
wu-sheng, line 192: No Getter

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,217 @@
...
+        // `default` is the keyword ,so named `global`
+        private SamplePolicy global;
+        @Getter
```
wu-sheng, line 195: No Getter

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,217 @@
...
+            SamplePolicy samplePolicy = new SamplePolicy(defaultRate == null ? 10000 : defaultRate, defaultDuration == null ? -1 : defaultDuration);
+            this.global = samplePolicy;
+            this.services = new ConcurrentHashMap<>();
+        }
+
+        public void addPolicy(String service, SamplePolicy samplePolicy) {
+            this.services.put(service, samplePolicy);
+        }
+
+        public SamplePolicy getPolicy(String service) {
```
wu-sheng, line 212: This implementation is incorrect. It should be a simple `get`. You should have `service specific`->`global`->`hardcode default` logic here.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,217 @@
...
+    public int getSlowTraceSegmentThreshold() {
+        return this.samplePolicySettings.get().global.duration;
+    }
+
+    public boolean shouldSample(int duration) {
+        return (this.samplePolicySettings.get().global.duration > -1) && (duration >= this.samplePolicySettings.get().global.duration);
+    }
+
+    private void loadDefaultPolicy() {
+        SamplePolicySettings rateSetting = parseFromFile(moduleConfig.getTraceSamplingPolicySettingsFile());
```
wu-sheng, line 90: Still have wrong name. This is not `rateSetting`.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,217 @@
...
+        // If settingFile has a empty config, use the default settings
```
wu-sheng, line 91:
```suggestion
        // If settingFile has an empty config, use the default settings
```
This comment seem strange, you should check whether having default global rule. Not whether policy setting is empty.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,217 @@
...
+            loadDefaultPolicy();
+        } else {
+            log.info("[trace-sampling-policy] Parse yaml fail, retain last configuration: {}", this.samplePolicySettings);
+        }
+    }
+
+    private SamplePolicySettings parseFromFile(final String file) {
+        try {
+            final Reader reader = ResourceUtils.read(file);
+            Map<String, Object> map = new Yaml(new ClassFilterConstructor(new Class[]{
```
wu-sheng, line 118: This is a wrong usage of API. You should use `SafeConstructor`. Take `RulesReader` as a reference. 
The `ClassFilterConstructor` is designed for requiring initialization from customer Java class.

wu-sheng:
One key principle you are still not 100% following
Only expose necessary methods to usage. All complex things(especially fail back) should be kept in one place and internally. Logic should belong to its owner. API(Watcher's API) users don't care or don't want to care how these threshold(rate/duration) gets initialization.

wu-sheng:
Watcher should expose at most 3 methods
1. `getPolicySetting` -> `PolicySetting`.
2. `PolicySetting#shouldSample(int rate)`
3. `PolicySetting#isOverSlowThreshold(int duration)`.

All other things should be kept internally. And most logic should be written once. Such as `getPolicySetting` -> `PolicySetting` should include all fail back logic.

wu-sheng:
I am so sorry that, I have to be so strict/rigorous  about this PR. Because it is a big change for code structure. New sampling conditions/rules could be added by others in the future, and they will follow your code style. Like you did in other PRs for following older codes.
Once we make this part of codes not clear, it will make further review harder, and logic gets inconsistent. Eventually, we have to face unexpected bugs.

HendSame:
`don't need to keep moduleConfig, just pass the file path into #loadDefaultPolicy enough.`
So we should have another `variable`  to keep `file path`?  Because  it will need load static configs from the file, when `traceSamplingPolicy` is been deleted dynamically.

HendSame:
> I am so sorry that, I have to be so strict/rigorous about this PR. Because it is a big change for code structure. New sampling conditions/rules could be added by others in the future, and they will follow your code style. Like you did in other PRs for following older codes.
> Once we make this part of codes not clear, it will make further review harder, and logic gets inconsistent. Eventually, we have to face unexpected bugs.

Totally understand！I  will update the points  that  you  had proposed.

wu-sheng:
> `don't need to keep moduleConfig, just pass the file path into #loadDefaultPolicy enough.`
> 
> So we should have another `variable`  to keep `file path`?  Because  it will need to load static configs from the file, when `traceSamplingPolicy` been deleted dynamically. 
> 
> 
> 
> 

Don't keep the file path, or read it again. Keeping the content should be enough.

HendSame (commented):
> > `don't need to keep moduleConfig, just pass the file path into #loadDefaultPolicy enough.`
> > So we should have another `variable`  to keep `file path`?  Because  it will need to load static configs from the file, when `traceSamplingPolicy` been deleted dynamically.
> 
> Don't keep the file path, or read it again. Keeping the content should be enough.

I'm still a little confused. How to keep the content?  Keeping the  file content  in  memory? If  the file contains a lot of content that  takes up unnecessary memory.

wu-sheng:
> > > `don't need to keep moduleConfig, just pass the file path into #loadDefaultPolicy enough.`
> > > So we should have another `variable`  to keep `file path`?  Because  it will need to load static configs from the file, when `traceSamplingPolicy` been deleted dynamically.
> > 
> > 
> > Don't keep the file path, or read it again. Keeping the content should be enough.
> 
> I'm still a little confused. How to keep the content? Keeping the file content in memory? If the file contains a lot of content that takes up unnecessary memory.

Yes, keep the content, even the object instance representing the parsed result.
Each service could only take 3 lines of settings, 200-300 services mean a less than 1000 lines file, which means nothing from memory perspective.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,156 @@
...
+            return shouldSampleGlobal(sample, duration);
+        }
+        return shouldSampleService(samplingPolicy, sample, duration);
+    }
+
+    public SamplingPolicy getSamplingPolicy(String service) {
+        return this.samplingPolicySettings.get().get(service);
+    }
+
+    private boolean shouldSampleGlobal(int sample, int duration) {
```
wu-sheng, line 79:
```suggestion
    private boolean shouldSampleByDefault(int sample, int duration) {
```

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/sampling/SamplingPolicySettings.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+@ToString
+public class SamplingPolicySettings {
+
+    @Getter
+    // `default` is the keyword ,so named `global`
+    private SamplingPolicy global;
```
wu-sheng, line 31:
```suggestion
    private SamplingPolicy defaultPolicy;
```

This is easier to read, and no comment is needed.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,156 @@
...
+            activeSetting(value.getNewValue());
+        }
+    }
+
+    @Override
+    public String value() {
+        return this.settingsString.get();
+    }
+
+    public boolean shouldSample(String service, int sample, int duration) {
```
wu-sheng, line 67: This method should have comments as it is a public method.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,156 @@
...
+
+    public SamplingPolicy getSamplingPolicy(String service) {
+        return this.samplingPolicySettings.get().get(service);
+    }
+
+    private boolean shouldSampleGlobal(int sample, int duration) {
+        return isOverSlowThresholdGlobal(duration) || withinRateRangeGlobal(sample);
+    }
+
+    private boolean shouldSampleService(SamplingPolicy samplingPolicy, int sample, int duration) {
```
wu-sheng, line 83: You need to describe the logic, including what 4 conditions mean `should sample`.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,156 @@
...
+        return (samplingPolicy.getDuration() != null && isOverSlowThreshold(duration, samplingPolicy.getDuration()))
+                || (samplingPolicy.getRate() != null && withinRateRange(sample, samplingPolicy.getRate()))
+                // global policy
+                || (samplingPolicy.getDuration() == null && isOverSlowThresholdGlobal(duration))
+                || (samplingPolicy.getRate() == null && withinRateRangeGlobal(sample));
+    }
+
+    private boolean withinRateRangeGlobal(int sample) {
```
wu-sheng, line 91:
```suggestion
    private boolean withinDefaultRateRange(int sample) {
```

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,156 @@
...
+        return withinRateRange(sample, this.samplingPolicySettings.get().getGlobal().getRate());
+    }
+
+    private boolean isOverSlowThresholdGlobal(int duration) {
```
wu-sheng, line 95:
```suggestion
    private boolean isOverDefaultSlowThreshold(int duration) {
```

wu-sheng (commented):
Basically good. Just some nits.

@kezhenxu94 @JaredTan95  Please recheck.

wu-sheng:
`configuration-vocabulary.md` doc should be updated accordingly.

wu-sheng:
@HendSame It seems you changed the source codes, but forget the test cases.

HendSame:
> @HendSame It seems you changed the source codes, but forget the test cases.
Test cases's comments ?

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/sampling/SamplingPolicySettings.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Getter
+    private SamplingPolicy defaultPolicy;
+    private Map<String, SamplingPolicy> services;
+
+    /**
+     * @param defaultRate     The sample rate precision is 1/10000. 10000 means 100% sample in default.
+     * @param defaultDuration Setting this threshold about the latency would make the slow trace segments sampled if they cost more time, even the sampling mechanism activated. The default value is `-1`, which means would not sample slow traces. Unit, millisecond.
+     */
+    public SamplingPolicySettings(Integer defaultRate, Integer defaultDuration) {
```
wu-sheng, line 37: According to codes, there is no point to add `defaultRate` and `defaultDuration`. They are always null.

image

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/AnalyzerModuleProvider.java
```diff
@@ -111,18 +107,17 @@ public void prepare() throws ServiceNotProvidedException, ModuleStartException {
     public void start() throws ModuleStartException {
         // load official analysis
         getManager().find(CoreModule.NAME)
-                    .provider()
-                    .getService(OALEngineLoaderService.class)
-                    .load(CoreOALDefine.INSTANCE);
+                .provider()
+                .getService(OALEngineLoaderService.class)
+                .load(CoreOALDefine.INSTANCE);
```
kezhenxu94, line 112: Please import our code style settings, and be aware of these unrelated changes when submitting an PR

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,187 @@
...
+    /**
+     * Determine whether need to be sampled
+     *
+     * @param service  service's name
+     * @param sample   sample rate of trace segment
+     * @param duration duration of trace segment
+     * @return
+     */
+    public boolean shouldSample(String service, int sample, int duration) {
+        SamplingPolicy samplingPolicy = this.samplingPolicySettings.get().get(service);
```
wu-sheng, line 76: This is `getSamplingPolicy`, which is a one line method. Please remove `getSamplingPolicy` method. When you need `samplingPolicySettings`, please set a method in test case powered by whitebox(powermock)

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/sampling/SamplingPolicySettingsReader.java
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * SamplePolicySettingsReader parses the given `trace-sampling-policy-settings.yml` config file, to the target {@link SamplingPolicySettings}.
+ */
+public class SamplingPolicySettingsReader {
+    private Map yamlData;
+
+    public SamplingPolicySettingsReader(InputStream inputStream) {
+        Yaml yaml = new Yaml(new SafeConstructor());
+        yamlData = (Map) yaml.load(inputStream);
```
kezhenxu94, line 38: Casting `yaml.load` to `Map` is redundant

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/sampling/SamplingPolicySettingsReader.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    public SamplingPolicySettingsReader(Reader io) {
+        Yaml yaml = new Yaml(new SafeConstructor());
+        yamlData = (Map) yaml.load(io);
```
kezhenxu94, line 43: Casting `yaml.load` to `Map` is redundant

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/sampling/SamplingPolicySettingsReader.java
```diff
@@ -0,0 +1,88 @@
...
+        Map<String, Object> objectMap = (Map<String, Object>) yamlData.get("default");
+        if (objectMap == null) {
+            return;
+        }
+        if (objectMap.get("rate") != null) {
+            samplingPolicySettings.getDefaultPolicy().setRate((Integer) objectMap.get("rate"));
+        }
+        if (objectMap.get("duration") != null) {
+            samplingPolicySettings.getDefaultPolicy().setDuration((Integer) objectMap.get("duration"));
+        }
```
kezhenxu94, line 68: Here is one caveat: when the settings is configured via dynamic configuration, and the configured policy is 

```yaml
default:
  rate: 123
  duration: non-a-number
```

The default policy is **partially** updated, i.e., the `rate` is updated while the `duration` is not updated because it has an invalid value, and if I were a user, I'd expect the `rate` and `duration` to be updated both or none

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,187 @@
...
+@Slf4j
+public class TraceSamplingPolicyWatcher extends ConfigChangeWatcher {
+
+    private final AtomicReference<String> settingsString = new AtomicReference<>(Const.EMPTY_STRING);
+    private final AtomicReference<SamplingPolicySettings> samplingPolicySettings = new AtomicReference<>(null);
+    private final SamplingPolicySettings defaultSamplingPolicySettings;
+
+    public TraceSamplingPolicyWatcher(AnalyzerModuleConfig moduleConfig, ModuleProvider provider) {
+        super(AnalyzerModule.NAME, provider, "traceSamplingPolicy");
+        SamplingPolicySettings samplingPolicySettings = parseFromFile(moduleConfig.getTraceSamplingPolicySettingsFile());
```
kezhenxu94, line 46: Local variable 'samplingPolicySettings' is redundant

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/sampling/SamplingPolicySettings.java
```diff
@@ -0,0 +1,50 @@
...
+    @Getter
+    private SamplingPolicy defaultPolicy;
+    private Map<String, SamplingPolicy> services;
+
+    /**
+     * The sample rate precision is 1/10000. 10000 means 100% sample in default.
+     * Setting this threshold about the latency would make the slow trace segments sampled if they cost more time, even the sampling mechanism activated. The default value is `-1`, which means would not sample slow traces. Unit, millisecond.
+     */
+    public SamplingPolicySettings() {
+        SamplingPolicy samplingPolicy = new SamplingPolicy(10000, -1);
```
kezhenxu94, line 38: Local variable 'samplingPolicy' is redundant

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,187 @@
...
+
+    /**
+     * Get sampling policy for specific service
+     *
+     * @param service serviceName
+     * @return
+     */
+    public SamplingPolicy getSamplingPolicy(String service) {
+        return this.samplingPolicySettings.get().get(service);
+    }
```
wu-sheng, line 91: This is a public and unused method. Please remove it

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/TraceSegmentSampler.java
```diff
@@ -18,20 +18,24 @@
...
+import org.apache.skywalking.apm.network.language.agent.v3.SegmentObject;
+import org.apache.skywalking.oap.server.analyzer.provider.trace.TraceSamplingPolicyWatcher;
 
 /**
  * The sampler makes the sampling mechanism works at backend side. Sample result: [0,sampleRate) sampled, (sampleRate,~)
  * ignored
  */
 public class TraceSegmentSampler {
-    private TraceSampleRateWatcher traceSampleRateWatcher;
+    private TraceSamplingPolicyWatcher traceSamplingPolicyWatcher;
```
kezhenxu94, line 29: Field 'traceSamplingPolicyWatcher' may be 'final' and let's use `@RequiredArgsConstructor`

kezhenxu94 (approved):
Generally look good to me, just some nits

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,187 @@
...
+
+import java.io.StringReader;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static java.util.Objects.isNull;
+
+@Slf4j
+public class TraceSamplingPolicyWatcher extends ConfigChangeWatcher {
+
+    private final AtomicReference<String> settingsString = new AtomicReference<>(Const.EMPTY_STRING);
```
wankai123, line 40: suggest: `private final AtomicReference<String> settingsString = new AtomicReference<>(null);`
because in
```
    @Override
    public String value() {
        return this.settingsString.get();
    }
```
`settingsString=""` is not `null` could notify every time when the config is deleted. see `ConfigWatcherRegister.singleConfigsSync 112`

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/TraceSamplingPolicyWatcher.java
```diff
@@ -0,0 +1,187 @@
...
+        super(AnalyzerModule.NAME, provider, "traceSamplingPolicy");
+        SamplingPolicySettings samplingPolicySettings = parseFromFile(moduleConfig.getTraceSamplingPolicySettingsFile());
+        this.defaultSamplingPolicySettings = samplingPolicySettings;
+        loadDefaultPolicySettings();
+    }
+
+    @Override
+    public void notify(ConfigChangeEvent value) {
+        if (EventType.DELETE.equals(value.getEventType()) || StringUtil.isBlank(value.getNewValue())) {
+            this.settingsString.set("");
```
wankai123, line 54: suggest:  `settingsString = null;`
because in

    @Override
    public String value() {
        return this.settingsString.get();
    }
settingsString="" is not null could notify every time when the config is deleted. see ConfigWatcherRegister.singleConfigsSync 112

wankai123 (changes requested):

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/AnalyzerModuleProvider.java
HendSame: Where can I get the `code style settings` ?

wu-sheng: https://github.com/apache/skywalking/blob/master/codeStyle.xml

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/sampling/SamplingPolicySettingsReader.java
HendSame: I had run the Uts for the case ,it is ok.  
image

The same as :
image
image

wankai123 (approved):

wu-sheng (approved):

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/sampling/SamplingPolicySettingsReader.java
wu-sheng: I think you should use `"abc"` in the YAML?

wu-sheng: @HendSame From the code reading, I feel Zhenxu is correct. If `duration` is invalid, then `.setDuration((Integer) objectMap.get("duration"))` would fail. In this case, once `rate` has been updated, then we would face partial-update case.

Could you recheck about this?

kezhenxu94: > @HendSame From the code reading, I feel Zhenxu is correct. If `duration` is invalid, then `.setDuration((Integer) objectMap.get("duration"))` would fail. In this case, once `rate` has been updated, then we would face partial-update case.
> 
> Could you recheck about this?

It's indeed not partially updated, which is kind of tricky because if `org.apache.skywalking.oap.server.analyzer.provider.trace.TraceSamplingPolicyWatcher#parseFromYml` throws exception (`ClassCastException` from `String` `abc` to `Integer`), it returns `null` and thus `onUpdated(parseFromYml(config));` won't be executed, so the settings remain to the last update...

wu-sheng: Oh, I see. This is a tricky flow control. I thought it should be something inside method.
OK, I would not require this has to be changed, @kezhenxu94 your call.

kezhenxu94 (approved):
Thanks for the patience @HendSame !!

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/sampling/SamplingPolicySettingsReader.java
HendSame: @wu-sheng @kezhenxu94 
I think it is ok.Is there anything else I haven't considered? I need your help. 
image
image

PR #7579, opened 2021-08-26
Fix image vulnerability issue

hanahmily:
Related to https://github.com/alpinelinux/docker-alpine/issues/199.

I used a quick way to fix this issue. Once the underlying alpine image fixed this, we should upgrade the base image then remove this.


wu-sheng:
@hanahmily I can see the base image updated, https://hub.docker.com/layers/adoptopenjdk/openjdk11/x86_64-alpine-jdk-11.0.12_7/images/sha256-643cf3ee451a32c96d16a2e5db4869593ae5389002574908b23c30fe22e1e0bc?context=explore.

Please recheck.

hanahmily:
The specific image fixed this, but the base image we're using hasn't been released. Once it's released, we can close this PR.

wu-sheng:
Why so many e2e tests fail? @hanahmily

hanahmily:
The base image has fixed this issue, close this.

PR #7601, opened 2021-08-30
Fix CI job `check-dependency-license` failure

kezhenxu94:
Restore the self modules (oap-servers, server-alarm-plugin, etc.) that are installed and cached in previous jobs and make script `check-LICENSE.sh` succeed


wu-sheng (approved):

wu-sheng:
`@github-actions CI AND IT / check-dependency-license (pull_request) ` still fails. Please recheck. @kezhenxu94

wu-sheng (approved):

PR #7604, opened 2021-08-30
Use new GitHub issue form template

kezhenxu94:
Preview https://github.com/kezhenxu94/skywalking/issues/new/choose


wu-sheng:
`What happened`, `What you expected to happen` and `How to reproduce` should be required in a bug report, right?

wu-sheng (approved):

PR #7611, opened 2021-08-31
Fix `ZookeeperConfigWatcherRegister.readConfig()` could cause `NPE` when `data.getData()` is null

wankai123:
### Fix  `ZookeeperConfigWatcherRegister.readConfig()` could cause `NPE` when `data.getData()` is null

`new String(data.getData())` could cause NPE, when clear data bytes or set data byte[] null.


wu-sheng (approved):
Thanks.

kezhenxu94 (approved):

PR #7636, opened 2021-09-02
Support for filter filtering of int type values

### LINKED ISSUE #7521: Support real equalMatch in OAL core
### Discussed in https://github.com/apache/skywalking/discussions/7517

<div type='discussions-op-text'>

<sup>Originally posted by **807801002** August 20, 2021</sup>
Which field can be used at filter()？I thought all fields in org.apache.skywalking.oap.server.core.source.ServiceRelation can be used to the filter expression for "from(ServiceRelation.*)".But the "componentId == 7" is error. which is a field at ServiceRelation.
Snipaste_2021-08-20_17-20-14
</div>

We have `EqualMatch`(function name, `stringMatch`), but this is for String type only. We should add a real `equalMatch` function in grammar tree and implementation in `org.apache.skywalking.oap.server.core.analysis.metrics.expression`. 
Also, we should rename class `EqualMatch` to `StringMatch`.

YczYanchengzhe:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7521.


YczYanchengzhe:
Already for merge

wu-sheng:
> Already for merge

You should fix CI. Then we will do the review.

wu-sheng (dismissed):
LGTM. Wait for other reviewers

@YczYanchengzhe Have you really added `ClientCpm = from(ServiceInstanceRelation.*).filter(componentId == 7).cpm();` in the OAP for the runtime, is everything working as expected?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/expression/IntMatch.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.skywalking.oap.server.core.analysis.metrics.expression;
+
+import org.apache.skywalking.oap.server.core.analysis.metrics.annotation.FilterMatcher;
+
+@FilterMatcher
+public class IntMatch {
```
kezhenxu94, line 23: Can we use `match(Number left, Number right)` so that they apply to other primitives like long, double, etc.

oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/analysis/metrics/expression/StringMatchTest.java
```diff
@@ -23,44 +23,44 @@
...
 import static org.junit.Assert.assertTrue;
 
-public class EqualMatchTest {
+public class StringMatchTest {
 
     @Test
     public void integerShouldEqualWhenLargerThan128() {
         Integer a = 334;
         Integer b = 334;
-        boolean match = new EqualMatch().match(a, b);
```
kezhenxu94, line 32: From these tests, the original marcher should already support numeric types, just need to wrap primitives such as `int` to boxed type `Integrr` or add method signatures with primitive types

wu-sheng: Same reason as https://github.com/apache/skywalking/pull/7636#discussion_r701062504. And discussed at https://github.com/apache/skywalking/discussions/7517

YczYanchengzhe: Am I trying below whether adding the method signature can be directly supported

YczYanchengzhe: There is a question about what the regression tests need to do every time you modify the oap part of the code

wu-sheng: > Am I trying below whether adding the method signature can be directly supported

What is this? A question or something?

> There is a question about what the regression tests need to do every time you modify the oap part of the code

There is no guidance or standards about this, because this is rarely to change, and has been changed for a long time.

YczYanchengzhe: For this problem kezhenxu94 raised a better solution I m trying

YczYanchengzhe: image
As kezhenxu94 said this can be solved by simply the matching rules support numbers in stringmatch I think this is a better solution and is the class name restored to EqualMatch, I think EqualMatch is more reasonable in design and functionality

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/expression/IntMatch.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public boolean match(int left, int right) {
+        return left == right;
+    }
+
+    public boolean match(Integer left, Integer right) {
```
wu-sheng, line 29: My point is this is not working. Consider to remove this and run test.

YczYanchengzhe: image
The problem is the automatic binning, which is fine under my test

oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/analysis/metrics/expression/StringMatchTest.java
wu-sheng: whether it is working or not, is not important at all. If you want to more method, consider int and long. But you would not compare number, especially double and float. Because simply in Java `1.01 == 1.01` is not guaranteed.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/expression/IntMatch.java
wu-sheng: This could work, but it still requires you to change g4 file, because it was `stringMatch`, which doesn't fit the case.

wu-sheng: My point is NumberEqual and IntEqual both make sense and clear. You could add `match(long left, long right` in case for the future. `StringMatch` should be kept in String only.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/expression/NumberMatch.java
```diff
@@ -0,0 +1,40 @@
...
+        return left == right;
+    }
+
+    public boolean match(float left, float right) {
+        return left == right;
+    }
+
+    public boolean match(Number left, Number right) {
+        return left.equals(right);
+    }
```
wu-sheng, line 39:
```suggestion
```

Number doesn't include float like I mentioned.
```
numberConditionValue
    : NUMBER_LITERAL
    ;

NUMBER_LITERAL :   Digits+;

fragment Digits
    : [0-9] ([0-9_]* [0-9])?
    ;
```

And Number object doesn't make sense too.

wu-sheng (approved):

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/expression/NumberMatch.java
kezhenxu94: @YczYanchengzhe I think supporting `long / Long`, `int / Integer` should be enough in our cases, we only have 1 `double` typed field in `Source` scope and its name is `usePercent`, I don't think it makes sense to compare `==` between `double`s

YczYanchengzhe: I agree

kezhenxu94 (approved):
Thank you

YczYanchengzhe:
Thank you so much for your help, I learned a lot of interesting things, I look to see what more ~ than I can do

PR #7641, opened 2021-09-03
add suffix limit to avoid possible start problems

honganan:
【Improve code robustness】
Here reads all files in the directory with Yaml util, in some situation if the directory contains hidden system file such as `.DS_Store` on Mac OS, and I waste many time on it as the error cannot figure out any clues.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleProvider.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleProvider.java
@@ -376,6 +376,9 @@ public class CoreModuleProvider extends ModuleProvider {
         try {
             final File[] templateFiles = ResourceUtils.getPathFiles("ui-initialized-templates");
             for (final File templateFile : templateFiles) {
+                if (!templateFile.getName().endsWith(".yml") && !templateFile.getName().endsWith(".yaml")) {
+                    continue;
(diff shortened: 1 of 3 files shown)


wu-sheng:
It is fine to add this, please update changes.md in the root.

wu-sheng:
And this doc should be updated, https://skywalking.apache.org/docs/main/latest/en/ui/readme/#custom-dashboard. Because you add new requirements for the file name.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleProvider.java
```diff
@@ -376,6 +376,9 @@ public void notifyAfterCompleted() throws ModuleStartException {
         try {
             final File[] templateFiles = ResourceUtils.getPathFiles("ui-initialized-templates");
             for (final File templateFile : templateFiles) {
+                if (!templateFile.getName().endsWith(".yml")) {
```
kezhenxu94, line 379: Please also take `yaml` into consideration

honganan: ok

wu-sheng (approved):

kezhenxu94 (approved):
Thank you

Issue #7368, opened 2021-07-24
Adopt `skywalking-infra-e2e` in all e2e tests


wu-sheng:
Hi @apache/skywalking-committers and contributors

[SkyWalking INFRA e2e](https://github.com/apache/skywalking-infra-e2e) subproject is almost ready to release because this is for testing scope only, we could adopt it in the main repo based on commit IDs.

We are welcome everyone to join this and replace the existing Java Maven based e2e tests. The latest one should be much more friendly and more powerful.

@mrproliu is leading on the subproject's first release. 

CLI is another key point of the whole replacement. 
@fgksgf @alonelaval Could you help on enhancing the CLI project about the topology part?

mrproliu:
Now I'm working on `satellite` testing base on `infra-e2e`, there are still have some `cli` command need to be provided, such as log, CDS, event, profile. I'm trying to working on the `cli` project too. It does not block the replacement process, just some functions.

wu-sheng:
> Now I'm working on `satellite` testing base on `infra-e2e`, there are still have some `cli` command need to be provided, such as log, CDS, event, profile. I'm trying to working on the `cli` project too. It does not block the replacement process, just some functions.

CLI is a kind of block. Because we have verified the topology already on the current e2e, we at least should keep it.

mrproliu:
> > Now I'm working on `satellite` testing base on `infra-e2e`, there are still have some `cli` command need to be provided, such as log, CDS, event, profile. I'm trying to working on the `cli` project too. It does not block the replacement process, just some functions.
> 
> CLI is a kind of block. Because we have verified the topology already on the current e2e, we at least should keep it.

Sure. Then I think the protocol I mentioned above should also be completed before e2e is implemented.

fgksgf:
I will try my best to help since I'm interning, and working on GSoC simultaneously.

wu-sheng:
I have asked @mrproliu to use this new framework to test satellite w/ or w/o Kafka.

PR #7680, opened 2021-09-08
Use the jre image as the base image

### LINKED ISSUE #7597: Switch to JRE base image from JDK
I think we should do this for dev images and official release. @hanahmily 

@kezhenxu94 I think we should update agent's base image, right?

hanahmily:
Fixed #7597

### DIFF
--- a/docker/oap/Dockerfile.oap
+++ b/docker/oap/Dockerfile.oap
@@ -14,27 +14,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG BASE_IMAGE='adoptopenjdk/openjdk11:alpine'
+ARG BASE_IMAGE='adoptopenjdk/openjdk11:alpine-jre'
 
-FROM golang:1.14 AS cli
-
-ARG COMMIT_HASH=9f267876493943716434fdaa30047a14c0b5b2d9
-ARG CLI_CODE=${COMMIT_HASH}.tar.gz
-ARG CLI_CODE_URL=https://github.com/apache/skywalking-cli/archive/${CLI_CODE}
-
-ENV CGO_ENABLED=0
-ENV GO111MODULE=on
-
-WORKDIR /cli
-
-ADD ${CLI_CODE_URL} .
-RUN tar -xf ${CLI_CODE} --strip 1
-RUN rm ${CLI_CODE}
-
-RUN mkdir -p /skywalking/bin/
-RUN make linux && mv bin/swctl-latest-linux-amd64 /skywalking/bin/swctl
+FROM apache/skywalking-base:8.7.0-es7 AS cli
 
 FROM $BASE_IMAGE
+
+RUN apk add --no-cache openssl
 
 ENV JAVA_OPTS=" -Xms256M " \
     SW_CLUSTER="standalone" \
--- a/docker/oap/Dockerfile.oap
+++ b/docker/oap/Dockerfile.oap
@@ -20,7 +20,7 @@ FROM apache/skywalking-base:8.7.0-es7 AS cli
 
 FROM $BASE_IMAGE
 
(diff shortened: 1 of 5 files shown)


docker/oap/Dockerfile.oap
```diff
@@ -14,28 +14,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG BASE_IMAGE='adoptopenjdk/openjdk11:alpine'
+ARG BASE_IMAGE='adoptopenjdk/openjdk11:alpine-jre'
```
kezhenxu94, line 17: We need to update https://github.com/apache/skywalking/blob/edcf5926a84f4cc31d702f7014b9a6b5769d8b7a/.github/workflows/e2e.jdk-versions.yaml#L40 so that different versions of `jre` can be tested.

docker/oap/Dockerfile.oap
```diff
@@ -14,28 +14,14 @@
...
-
-WORKDIR /cli
-
-ADD ${CLI_CODE_URL} .
-RUN tar -xf ${CLI_CODE} --strip 1
-RUN rm ${CLI_CODE}
-
-RUN mkdir -p /skywalking/bin/
-RUN make linux && mv bin/swctl-latest-linux-amd64 /skywalking/bin/swctl
+FROM apache/skywalking-base:8.7.0-es7 AS cli
```
kezhenxu94, line 19: Although this works for now, but when we begin to adopt infra-e2e (https://github.com/apache/skywalking/issues/7368) I think it's highly possible we will have to enhance CLI and then compile from source codes here, so what about just updating the `COMMIT_HASH` at line 21?

kezhenxu94 (commented):
Also, FYI, I'm merging apache/skywalking-docker into this repo (and the java-agent part into apache/skywalking-java) https://github.com/apache/skywalking-docker/pull/24#discussion_r703498788.

docker/oap/Dockerfile.oap
kezhenxu94: > Although this works for now, but when we begin to adopt infra-e2e (#7368) I think it's highly possible we will have to enhance CLI and then compile from source codes here, so what about just updating the `COMMIT_HASH` at line 21?

Well ignore this, the CLI in infra-e2e should be run outside of OAP container, so the cane here is OK to me

docker/oap/Dockerfile.oap
hanahmily: Gotcha. I also found some other `openjdk` in e2e tests, for example, istio relevant cases.

wu-sheng: Do we have jre agent image already?

docker/oap/Dockerfile.oap
hanahmily: Thanks for sharing this, I will update commit sha instead of switching back to released images to get CLI.

docker/oap/Dockerfile.oap
kezhenxu94: > Do we have jre agent image already?

For Java agent, no, it's still using JDK

wu-sheng: OK, seems we should take care of it later.

wu-sheng:
> javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed: 

@kezhenxu94 This seems definitely showing up in the agent testing. But always be good in other tests. I think something breaking.

wu-sheng (approved):

wu-sheng:
@kezhenxu94 Should we keep merging this? I think we have no clue about the mvnw issue.

kezhenxu94:
> @kezhenxu94 Should we keep merging this? I think we have no clue about the mvnw issue.

Let's merge this

PR #7681, opened 2021-09-08
Merge `Dockerfile`s in apache/skywalking-docker into this codebase

### LINKED ISSUE #7656: [Feature] Update SkyWalking Docker according to latest ElasticSearch client rebuild
### Search before asking

### Description

After https://github.com/apache/skywalking/pull/7634 gets merged, we need to adjust docker files in https://github.com/apache/skywalking-docker for incoming 8.8.0 release.

### Use case

We don't need images for ElasticSearch 6 and 7 separate now.

### Related issues

https://github.com/apache/skywalking/issues/7471

### Code of Conduct

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7656.

### DIFF
--- a/.github/workflows/e2e.jdk-versions.yaml
+++ b/.github/workflows/e2e.jdk-versions.yaml
@@ -33,7 +33,7 @@ jobs:
     timeout-minutes: 90
     strategy:
       matrix:
-        jdk: [ 8, 9, 10, 11, 12, 13, 14, 15 ]
+        jdk: [ 8, 11, 12, 13, 14, 15 ]
     env:
       SW_SIMPLE_CASE: jdk
       SW_AGENT_JDK_VERSION: ${{ matrix.jdk }}
--- a/docker/oap/Dockerfile.oap
+++ b/docker/oap/Dockerfile.oap
@@ -14,36 +14,113 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG BASE_IMAGE='adoptopenjdk/openjdk11:alpine-jre'
+ARG BASE_IMAGE='adoptopenjdk/openjdk11:alpine'
 
-FROM apache/skywalking-base:8.7.0-es7 AS cli
+FROM $BASE_IMAGE as cli
+
+WORKDIR /skywalking
+
+ARG SKYWALKING_CLI_VERSION=0.7.0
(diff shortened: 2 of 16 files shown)


wu-sheng:
let's try to chance jdk versions e2e testing? Remove some version or rename? It is strange they always fail on that mission only.

kezhenxu94:
> Add a doc about how to run through docker and docker compose?
> docker repo should be archived later.

Added

.github/workflows/e2e.jdk-versions.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
     timeout-minutes: 90
     strategy:
       matrix:
-        jdk: [ 8, 9, 10, 11, 12, 13, 14, 15 ]
+        jdk: [ 8 ] # , 9, 10, 11, 12, 13, 14, 15 ] TODO: add it back
```
wu-sheng, line 36: This seems works. Let's try to skip 9 and 10 only?

docs/en/setup/backend/docker.md
```diff
@@ -0,0 +1,25 @@
+# How to use the Docker images
```
wu-sheng, line 1: I think we missed UI part and docker compose

docker/oap/Dockerfile.oap
```diff
@@ -14,36 +14,113 @@
...
+	tar -xvf ${SKYWALKING_CLI_TGZ}; \
+    mkdir "bin/"; \
+	mv skywalking-cli-${SKYWALKING_CLI_VERSION}-bin/bin/swctl-${SKYWALKING_CLI_VERSION}-linux-amd64 bin/swctl; \
+	chmod 755 bin/*; \
+	command -v gpgconf && gpgconf --kill all || :; \
+	ls -la .;
 
 FROM $BASE_IMAGE
 
-RUN apk add --no-cache openssl || true
```
hanahmily, line 23: Could we remove "bash" here? `docker-entrypoint.sh` looks like not needing it anymore.

kezhenxu94: DONE

.github/workflows/e2e.jdk-versions.yaml
kezhenxu94: Let's just drop 9, and 10

wu-sheng (approved):

PR #7698, opened 2021-09-12
Fix that `sortMetrics` should only return `long` not `double`

kezhenxu94:


wu-sheng (dismissed):

wu-sheng (approved):

wu-sheng:
The link to python doc should be updated to website later.

PR #7707, opened 2021-09-14
Add agent compatibility doc for the first time.

### LINKED ISSUE #7652: [Feature] Begin to provide version compatibility doc
### Search before asking

### Description

After we split the Java agent code base out of main repo, end users are going to ask more about which versions of agents we could use for a specific backend version.

We should begin to maintain the document(from ASF's to ecosystem's agents).

@kezhenxu94 Please try to set up a document including a sheet to show that.

The sheet should look like(just for example, the version number doesn't mean anything)

| OAP Server version | Java Agent | Python Agent | Envoy ALS |
|:---:|:---:|:---:|:---:|
| >= 8.4.0| >= 8.0.0 | All | xxx|

### Use case

We should have a menu under `Setup`, and be named as `Version Compatibility`, which targets to the new doc.

### Related issues

_No response_

### Code of Conduct

wu-sheng:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7652.

First edition of agent compatibility. Some may be still missed, such as when CDS and Log-report get supported in agents,  I added this statement

```
All above compatibility are only references, if you face `unimplemented` error, it means you need to upgrade OAP backend
to support newer features in the agents.
```

### DIFF
--- a/docs/en/setup/service-agent/agent-compatibility.md
+++ b/docs/en/setup/service-agent/agent-compatibility.md
@@ -1,0 +1,27 @@
+# Compatibility
+
+SkyWalking 8.0+ uses v3 protocols. Agents don't have to keep the same versions as the OAP backend.
+
+## SkyWalking Native Agents
+
+|OAP Server Version|Java|Python|NodeJS|LUA|Kong|Browser Agent|
+----------- | ---------- | --------- | --------- |--------- |--------- |--------- |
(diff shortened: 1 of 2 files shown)


dmsolr (approved):
lgtm. Lua and Kong are right.

arugal (approved):
LGTM, browser agent and go2sky is ok

PR #7710, opened 2021-09-14
Fix that `Query` cannot be updated after it's set in `SearchBuilder`

kezhenxu94:
### Fix that `Query` cannot be updated after it's set in `SearchBuilder`
When `QueryBuilder` is set in `SearchBuilder`, the following updates to `QueryBuilder` (like `queryBuilder.must(Query.term(GROUP, "abc")` takes no effect, because when `SearchBuilder` accepts an `QueryBuilder`, it builds a `Query` from the builder and not keep the builder itself.
This patch also fixes a serialization error of `MatchQuery`.
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md). NO NEED, not released yet.

### DIFF
--- a/oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/search/MatchQuery.java
+++ b/oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/search/MatchQuery.java
@@ -38,10 +38,14 @@ public final class MatchQuery extends Query {
         public void serialize(final MatchQuery value, final JsonGenerator gen,
                               final SerializerProvider provider)
             throws IOException {
-            gen.writeFieldName("match");
             gen.writeStartObject();
             {
-                gen.writeStringField(value.getName(), value.getText());
+                gen.writeFieldName("match");
+                gen.writeStartObject();
+                {
+                    gen.writeStringField(value.getName(), value.getText());
+                }
+                gen.writeEndObject();
             }
             gen.writeEndObject();
         }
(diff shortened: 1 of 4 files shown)


oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/search/MatchQuery.java
```diff
@@ -38,10 +38,14 @@
         public void serialize(final MatchQuery value, final JsonGenerator gen,
                               final SerializerProvider provider)
             throws IOException {
-            gen.writeFieldName("match");
             gen.writeStartObject();
             {
-                gen.writeStringField(value.getName(), value.getText());
+                gen.writeFieldName("match");
+                gen.writeStartObject();
+                {
```
wu-sheng, line 45: What are these brackets for?

wu-sheng: Just a code style?

wu-sheng (approved):

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/search/MatchQuery.java
kezhenxu94: > Just a code style?

Yes, this is a kind of code style to make it easier to distinguish the written object structures, and help to write the correct `writeStartXxxx` / `writeEndXxx`, otherwise I have to count the number of `writeStartObject` / `writeEndObject`

PR #7731, opened 2021-09-17
Replace `mget` with ids' query as `mget` doesn't apply on aliases

kezhenxu94:
- [X] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md). Not released yet.


kezhenxu94:
When the index is time series and there is multiple indices on the same alias, the query would fail, here is the local verification

Screen Shot 2021-09-17 at 10 35 20

wu-sheng (approved):

PR #7760, opened 2021-09-22
Update ttl.md to add another 2 types of model relating to recordTTL

wu-sheng:

### DIFF
--- a/docs/en/setup/backend/ttl.md
+++ b/docs/en/setup/backend/ttl.md
@@ -1,6 +1,6 @@
 # TTL
 In SkyWalking, besides metadata, there are two types of observability data:
-1. Record: Includes trace and alarm. Log may also be included in the future.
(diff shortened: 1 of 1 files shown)


docs/en/setup/backend/ttl.md
```diff
@@ -1,7 +1,7 @@
 # TTL
-In SkyWalking, besides metadata, there are two types of observability data:
-1. Record: Includes trace and alarm. Log may also be included in the future.
-1. Metric: Includes percentile, heat map, success rate, cpm (rpm), etc.
+In SkyWalking, there are two types of observability data:
```
kezhenxu94, line 2: The summary is "two types of data" but the list items are two types of TTL, I think it's mismatch.

wu-sheng: Reword them.

kezhenxu94 (approved):

PR #7764, opened 2021-09-22
Update README.md

macaty:
update correct cli


kezhenxu94:
`compose` is built into `docker` command already, this is correct command

kezhenxu94:
@macaty please consider upgrade docker CLI version

kezhenxu94:
https://docs.docker.com/engine/reference/commandline/compose/

PR #7766, opened 2021-09-22
IoTDB storage plugin

LIU-WEI-git:
Hi, I am a participant in [Summer 2021 of Open Source Promotion Plan](https://summer.iscas.ac.cn/#/homepage). [Here](https://summer.iscas.ac.cn/#/org/prodetail/210070771) is my project homepage. Now I have finished my project under the guidance of @jixuan1989 from IoTDB and @wu-sheng from Skywalking. However, I still have much work to do. This PR will help me find more bug and deficiencies of my work.

I have written a [blog](https://blog.csdn.net/weixin_44023810/article/details/120555823) to introduce the design of IoTDB storage plugin.

This is a large module for Skywalking. I wrote more than 5 thousand source code lines and comment lines. They contains the core code and unit tests. When I wrote the code, I took the other storage plugin as a reference.

IoTDB doesn't have the index concept. In order to index some fields, I store the value of the index as a node of the storage path. Those nodes in storage path will be stored in the memory, so they serve as index. Eventually the data from the same table is spread across multiple devices. But IoTDB does not support cross-device query well. When I need use aggregation query (e.g. `max`, `avg`, `sum`), sort query (e.g. `order by`) or pagination query (e.g. `limit`, `offset`), I have to use a violence method that query all data meets the condition and then sort, aggregate or paginate them.

Although I have run and tested this IoTDB storage plugin locally, I know it is not enough and it need e2e tests at least. Hope you review my work and give me some suggestions.


wu-sheng:
@dmsolr Please lead on review, due to you lead on InfluxDB storage implementation.

wu-sheng:
@LIU-WEI-git You need to 
1. Make CI passed, including the new added e2e testing. 
2. Remember `rebase` the commit logs, 81 commits seem crazy, even SkyWalking would do `squash and merge`, still, we hope the commit log is clear.
3. Document should be update accordingly.

LIU-WEI-git:
ok, I will deal with them later.

wu-sheng:
Notice this https://github.com/apache/skywalking/tree/master/tools/dependencies. CI will fail due to new dependencies added or changed. This is the way we make sure licenses are checked and managed. You need to update `known-oap-backend-dependencies.txt` with License/Notice by following ASF policy.

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.skywalking</groupId>
+            <artifactId>library-client</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.iotdb</groupId>
+            <artifactId>iotdb-session</artifactId>
+            <version>0.12.2</version>
```
kezhenxu94, line 52: Make it a property and move it to `server-bom` module

wu-sheng:
> Although I have run and tested this IoTDB storage plugin locally, I know it is not enough and it need e2e tests at least. Hope you review my work and give me some suggestions.

The e2e tests could run locally too.

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
LIU-WEI-git: Got it.

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
```diff
@@ -31,8 +31,7 @@
     <packaging>jar</packaging>
 
     <properties>
-        <maven.compiler.source>11</maven.compiler.source>
-        <maven.compiler.target>11</maven.compiler.target>
+        <iotdb-session.version>0.12.2</iotdb-session.version>
```
wu-sheng: IoTDB client doesn't require JDK 11, right? SkyWalking is widely used in JDK8 for now.

LIU-WEI-git: IoTDB client needs JDK>=1.8

wu-sheng:
E2E passed, but you face some UT failure, please fix.

oap-server/server-storage-plugin/storage-iotdb-plugin/README.md
```diff
@@ -0,0 +1,194 @@
+Apache IoTDB - Apache Skywalking适配器
```
wu-sheng, line 1: Chinese document is not recommended to host in SkyWalking codebase.

Our official doc is here, https://github.com/apache/skywalking/blob/master/docs/en/setup/backend/backend-storage.md. And you could read the latest from https://skywalking.apache.org/docs/main/latest/en/setup/backend/backend-storage/

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.skywalking</groupId>
+            <artifactId>library-client</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.iotdb</groupId>
+            <artifactId>iotdb-session</artifactId>
+            <version>${iotdb-session.version}</version>
```
wu-sheng, line 51:
```suggestion
```

iotdb client version has been managed by bom.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/profile/IoTDBProfileThreadSnapshotQueryDAO.java
```diff
@@ -0,0 +1,174 @@
...
+        storageDataList.forEach(storageData -> segmentRecordList.add((SegmentRecord) storageData));
+        // resort by self, because of the select query result order by time.
+        segmentRecordList.sort((SegmentRecord r1, SegmentRecord r2) -> Long.compare(r2.getStartTime(), r1.getStartTime()));
+
+        List<BasicTrace> result = new ArrayList<>(segmentRecordList.size());
+        segmentRecordList.forEach(segmentRecord -> {
+            BasicTrace basicTrace = new BasicTrace();
+            basicTrace.setSegmentId(segmentRecord.getSegmentId());
+            basicTrace.setStart(String.valueOf(segmentRecord.getStartTime()));
+            basicTrace.getEndpointNames().add(String.valueOf(IDManager.EndpointID.analysisId(segmentRecord.getEndpointId())));
```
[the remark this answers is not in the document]
LIU-WEI-git: It has been initialized in constructor and won't be null.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBEventQueryDAO.java
```diff
@@ -0,0 +1,184 @@
...
+                Event event = (Event) storageDataList.get(i);
+                org.apache.skywalking.oap.server.core.query.type.event.Event resultEvent = parseEvent(event);
+                events.getEvents().add(resultEvent);
+            }
+        }
+        events.setTotal(storageDataList.size());
+        // resort by self, because of the select query result order by time.
+        final Order order = Objects.isNull(condition.getOrder()) ? Order.DES : condition.getOrder();
+        if (Order.DES.equals(order)) {
+            events.getEvents().sort(
```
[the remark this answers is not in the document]
LIU-WEI-git: It has been initialized in constructor and won't be null.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBEventQueryDAO.java
```diff
@@ -0,0 +1,184 @@
...
+                    (org.apache.skywalking.oap.server.core.query.type.event.Event e1,
+                     org.apache.skywalking.oap.server.core.query.type.event.Event e2)
+                            -> Long.compare(e2.getStartTime(), e1.getStartTime()));
+        } else {
+            events.getEvents().sort(
```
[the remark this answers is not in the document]
LIU-WEI-git: It has been initialized in constructor and won't be null.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBMetadataQueryDAO.java
```diff
@@ -0,0 +1,215 @@
...
+
+            JsonObject properties = instanceTraffic.getProperties();
+            if (!properties.isJsonNull()) {
+                for (Map.Entry<String, JsonElement> property : properties.entrySet()) {
+                    String key = property.getKey();
+                    String value = property.getValue().getAsString();
+                    if (key.equals(InstanceTraffic.PropertyUtil.LANGUAGE)) {
+                        serviceInstance.setLanguage(Language.value(value));
+                    } else {
+                        serviceInstance.getAttributes().add(new Attribute(key, value));
```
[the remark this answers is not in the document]
LIU-WEI-git: It has been initialized in constructor and won't be null.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/IoTDBClient.java
```diff
@@ -0,0 +1,475 @@
...
+                    String nodeType = (String) map.get(IoTDBClient.NODE_TYPE_IDX);
+                    map.put(IoTDBClient.NODE_TYPE_IDX, Integer.valueOf(nodeType));
+                }
+                if (modelName.equals(BrowserErrorLogRecord.INDEX_NAME) || modelName.equals(LogRecord.INDEX_NAME)) {
+                    map.put(IoTDBClient.TIMESTAMP, map.get("\"" + IoTDBClient.TIMESTAMP + "\""));
+                }
+                for (String key : map.keySet()) {
+                    // remove double quotes
+                    if (key.contains(".")) {
+                        map.put(key.substring(1, key.length() - 1), map.get(key));
```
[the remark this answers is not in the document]
LIU-WEI-git: Got it.

oap-server/server-storage-plugin/storage-iotdb-plugin/README.md
LIU-WEI-git: If IoTDB storage plugin could work well. I will edit official doc later.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBMetricsQueryDAO.java
```diff
@@ -0,0 +1,246 @@
...
+                String[] layerNames = fields.get(0).getStringValue().split("\\" + IoTDBClient.DOT + "\"");
+                String id = client.layerName2IndexValue(layerNames[1]);
+                long value = fields.get(1).getLongV();
+
+                KVInt kv = new KVInt();
+                kv.setId(id);
+                kv.setValue(value);
+                intValues.addKVInt(kv);
+            }
+            sessionPool.closeResultSet(wrapper);
```
kezhenxu94, line 141: You may want to put this in a `finally` block to make sure it's closed anyway

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBTraceQueryDAO.java
```diff
@@ -0,0 +1,156 @@
...
+                        Integer.compare(b2.getDuration(), b1.getDuration()));
+                break;
+        }
+        return traceBrief;
+    }
+
+    @Override
+    public List<SegmentRecord> queryByTraceId(String traceId) throws IOException {
+        StringBuilder query = new StringBuilder();
+        query.append("select * from ").append(client.getStorageGroup()).append(IoTDBClient.DOT).append(SegmentRecord.INDEX_NAME);
```
kezhenxu94, line 140: Consider adding an index formatter instead of concatenating this part over and over, like https://github.com/apache/skywalking/blob/4a344b01e9ad103b1cfda3e56d774dd4778c77de/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchProvider.java#L225-L232

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBAggregationQueryDAO.java
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.skywalking.oap.server.core.query.input.Duration;
+import org.apache.skywalking.oap.server.core.query.input.TopNCondition;
+import org.apache.skywalking.oap.server.core.query.type.KeyValue;
+import org.apache.skywalking.oap.server.core.query.type.SelectedRecord;
+import org.apache.skywalking.oap.server.core.storage.query.IAggregationQueryDAO;
+import org.apache.skywalking.oap.server.storage.plugin.iotdb.IoTDBClient;
+
+@Slf4j
+public class IoTDBAggregationQueryDAO implements IAggregationQueryDAO {
+    private IoTDBClient client;
```
kezhenxu94, line 44: can be final

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBAggregationQueryDAO.java
```diff
@@ -0,0 +1,114 @@
...
+    private static final Comparator<SelectedRecord> ASCENDING =
+            Comparator.comparingLong(a -> Long.parseLong(a.getValue()));
+    private static final Comparator<SelectedRecord> DESCENDING = (a, b) ->
+            Long.compare(Long.parseLong(b.getValue()), Long.parseLong(a.getValue()));
```
kezhenxu94, line 48: Unused

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/IoTDBClient.java
```diff
@@ -0,0 +1,476 @@
...
+    /**
+     * query with "string_contains" function
+     *
+     * @param modelName      model name
+     * @param querySQL       query sql
+     * @param storageBuilder storageBuilder
+     * @return a list of data which has been filtered by "string_contains" function
+     * @throws IOException IoTDBConnectionException or StatementExecutionException
+     */
+    public List<? super StorageData> queryWithContains(String modelName, String querySQL,
```
kezhenxu94, line 267: Unused

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/IoTDBClient.java
```diff
@@ -0,0 +1,476 @@
...
+            sessionPool.closeResultSet(wrapper);
+            healthChecker.health();
+        } catch (IoTDBConnectionException | StatementExecutionException e) {
+            healthChecker.unHealth(e);
+            throw new IOException(e.getMessage() + System.lineSeparator() + "SQL Statement: " + querySQL, e);
+        }
+        return storageDataList;
+    }
+
+    public List<Long> queryWithSelect(String modelName, String querySQL) throws IOException {
```
kezhenxu94, line 356: Unused

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/IoTDBClient.java
```diff
@@ -0,0 +1,476 @@
...
+        return query;
+    }
+
+    public StringBuilder addQueryAsterisk(String modelName, StringBuilder query) {
+        List<String> indexes = IoTDBTableMetaInfo.get(modelName).getIndexes();
+        indexes.forEach(index -> query.append(IoTDBClient.DOT).append("*"));
+        return query;
+    }
+
+    private boolean getStringContainsResult(Field field) {
```
kezhenxu94, line 469: Unused

oap-server/server-storage-plugin/storage-iotdb-plugin/src/test/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/base/IoTDBMetricsDAOTest.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.skywalking.oap.server.core.storage.model.ModelColumn;
+import org.apache.skywalking.oap.server.storage.plugin.iotdb.IoTDBClient;
+import org.apache.skywalking.oap.server.storage.plugin.iotdb.IoTDBStorageConfig;
+import org.apache.skywalking.oap.server.storage.plugin.iotdb.IoTDBTableMetaInfo;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.apache.skywalking.oap.server.storage.plugin.iotdb.IoTDBClientTest.retrieval;
+
+public class IoTDBMetricsDAOTest {
```
kezhenxu94, line 42: This kind of tests belong to integration test, refer to other existing storage tests about how to test.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/test/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/base/IoTDBMetricsDAOTest.java
```diff
@@ -0,0 +1,117 @@
...
+        instanceTraffic2.setName("name_2");
+        instanceTraffic2.setServiceId("service_id_2");
+        metrics.add(instanceTraffic2);
+
+        List<Metrics> newMetrics = metricsDAO.multiGet(IoTDBTableMetaInfo.get(InstanceTraffic.INDEX_NAME).getModel(), metrics);
+        newMetrics.forEach(metrics1 -> {
+            String serviceId = ((InstanceTraffic) metrics1).getServiceId();
+            String name = ((InstanceTraffic) metrics1).getName();
+            assert serviceId.equals("service_id_1") || serviceId.equals("service_id_2");
+            assert name.equals("name_1") || name.equals("name_2");
```
kezhenxu94, line 114: Do not use `assert `, use `org.assertj.core.api.Assertions#assertThat` instead

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+        <groupId>org.apache.skywalking</groupId>
+        <version>8.8.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>storage-iotdb-plugin</artifactId>
+    <packaging>jar</packaging>
+
+    <properties>
+        <iotdb-session.version>0.12.2</iotdb-session.version>
```
kezhenxu94, line 34: Unused

kezhenxu94 (commented):
Just did a quick review, there are some TODOs

LIU-WEI-git:
The docker image is update on Sep 12, but the fuzzy query bug is fixed on [Sep 14](https://github.com/apache/iotdb/pull/3953). So some UTs cannot pass. We have to wait IoTDB version 0.12.3 release.

wu-sheng:
Do you have GitHub host dev image? It is fine to use it for testing. We just need to indicate release version on doc, and update test image version later.
SkyWalking is using this mode, because waiting for another version could be long and face new bugs.

wu-sheng:
Take a reference from here. https://github.com/orgs/apache/packages?repo_name=skywalking

oap-server/server-storage-plugin/storage-iotdb-plugin/src/test/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/base/IoTDBMetricsDAOTest.java
LIU-WEI-git: I need this method to build `org.apache.skywalking.oap.server.core.storage.model`, and add it to `IoTDBTableMetaInfo.TABLE_META_INFOS`. When those query tests run, they need to get information from `IoTDBTableMetaInfo`.

LIU-WEI-git: Other storage tests just test their tool class. However I tests those query methods directly.

wu-sheng: Could we only test these DAOs in e2e rather than in Maven IT? It reduces resource costs.

wu-sheng: SkyWalking doesn't trust maven IT(not personal or relate to this PR, just general idea), SkyWalking is too complex, it is always better to have real OAP up and running, then verify logic.
IT in maven covers the API level check.

oap-server/server-storage-plugin/storage-iotdb-plugin/README.md
wu-sheng: This design part could land in SkyWalking website as a blog, describing the final status, such as how tables and query/index work.

wu-sheng: This kind of readme is hard up to date, so, we prefer don't keep them in official repository. If some design principles are important, such as rules of tables, and keys(indices), we should indicate them in the commenta of related codes.

LIU-WEI-git: Thanks for your suggestion. I will indicate some important design principles in the comments of related codes and delete it from repo in these days. After this project, I will make it as a blog.

LIU-WEI-git:
I will package a IoTDB dev docker image later.

wu-sheng:
> I will package a IoTDB dev docker image later.

SkyWalking is building dev image automatically for every default branch commit, using commitid as tag.

jixuan1989:
> SkyWalking is building dev image automatically for every default branch
commit, using commitid as tag.

Thanks, I will have a try to enable the feature of "Packages" :D

吴晟 Wu Sheng ***@***.***> 于2021年9月23日周四 下午6:58写道：

wu-sheng:
FYI to all, due to the discussion, the IoTDB latest version(0.12.2) can't support fuzzy query as expected(bug). Although it is fixed in their main branch, we don't have an image to test.
The reason of e2e passed, it's we don't test it.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBBrowserLogQueryDAO.java
```diff
@@ -0,0 +1,126 @@
...
+            log.setGrade(browserErrorLog.getGrade());
+            log.setMessage(browserErrorLog.getMessage());
+            log.setLine(browserErrorLog.getLine());
+            log.setCol(browserErrorLog.getCol());
+            log.setStack(browserErrorLog.getStack());
+            log.setErrorUrl(browserErrorLog.getErrorUrl());
+            log.setFirstReportedError(browserErrorLog.getFirstReportedError());
+
+            return log;
+        } catch (InvalidProtocolBufferException e) {
```
wu-sheng, line 122: Is this an IOException? If so, we don't need this try/catch.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBLogQueryDAO.java
```diff
@@ -0,0 +1,149 @@
...
+import org.apache.skywalking.oap.server.core.query.type.Logs;
+import org.apache.skywalking.oap.server.core.storage.StorageData;
+import org.apache.skywalking.oap.server.core.storage.StorageHashMapBuilder;
+import org.apache.skywalking.oap.server.core.storage.query.ILogQueryDAO;
+import org.apache.skywalking.oap.server.library.util.CollectionUtils;
+import org.apache.skywalking.oap.server.storage.plugin.iotdb.IoTDBClient;
+
+@Slf4j
+@RequiredArgsConstructor
+public class IoTDBLogQueryDAO implements ILogQueryDAO {
```
wu-sheng, line 50: @LIU-WEI-git As you haven't override `supportQueryLogsByKeywords`, `keywordsOfContent` and `excludingKeywordsOfContent` wouldn't be supported. Please recheck from UI.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBBrowserLogQueryDAO.java
LIU-WEI-git: No. This is method is refer to `IBrowserLogQueryDAO.parserDataBinary` without Base64 decode, because it has been decoded in `BrowserErrorLogRecord.Builder.storage2Entity`. The InvalidProtocolBufferException is thrown by `org.apache.skywalking.apm.network.language.agent.v3.BrowserErrorLog.parseFrom`

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBLogQueryDAO.java
LIU-WEI-git: I have cancelled fuzzy query in IoTDBLogQueryDAO

tools/dependencies/known-oap-backend-dependencies.txt
```diff
@@ -61,6 +61,7 @@ httpclient-4.5.13.jar
 httpcore-4.4.13.jar
 httpcore-nio-4.4.13.jar
 influxdb-java-2.15.jar
+iotdb-session-0.12.2.jar
```
wu-sheng, line 64: How does IoTDB client implement the network requests? I can see there is no more dependency?

wu-sheng: Dependencies should be updated here, https://github.com/apache/skywalking/blob/fdbbd5a94be8c196fa69056e1bacdd05e1041f31/dist-material/release-docs/LICENSE#L222

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+            <groupId>org.apache.skywalking</groupId>
+            <artifactId>server-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.skywalking</groupId>
+            <artifactId>library-client</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
```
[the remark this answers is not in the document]
wu-sheng: > pkg:maven/org.apache.iotdb/iotdb-session@0.12.2
12 Critical, 2 Severe, 0 Moderate, 0 Unknown vulnerabilities have been found across 1 dependencies

@LIU-WEI-git @jixuan1989 Could you confirm the CVEs? There are 12 critical and 2 severe?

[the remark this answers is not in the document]
LIU-WEI-git: I have set the `tomcat-embed-core` dependency version to 10.0.2

tools/dependencies/known-oap-backend-dependencies.txt
LIU-WEI-git: ok

dist-material/release-docs/LICENSE
```diff
@@ -327,6 +327,7 @@ The text of each license is the standard Apache 2.0 license.
     Armeria 1.11.0, http://github.com/line/armeria, Apache 2.0
     Brotli4j 1.6.0, https://github.com/hyperxpro/Brotli4j, Apache 2.0
     micrometer 1.7.3, https://github.com/micrometer-metrics/micrometer, Apache 2.0
+    iotdb-session 0.12.2, https://github.com/apache/iotdb, Apache 2.0
```
wu-sheng, line 330: All dependency libs should be added here and https://github.com/apache/skywalking/blob/master/tools/dependencies/known-oap-backend-dependencies.txt.

LIU-WEI-git: I've modified `known-oap-backend-dependencies.txt` before

wu-sheng: You just added one line, we check for all real dependencies.

wu-sheng: Read the error logs, they are much more than 2

image

LIU-WEI-git: Got it.

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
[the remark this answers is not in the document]
kezhenxu94: We should be using IoTDB client SDK but `tomcat-embed-core` seems to be a server side dependency, can you confirm we can exclude it? I don't want to include unnecessary dependencies which might bring CVE in the future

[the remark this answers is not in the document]
wu-sheng: I also pinged @jixuan1989 this morning, asked the same question. Wait for confirmation.

wu-sheng:
We have started 8.8.0 release, please sync to master carefully, version in pom.xml changed.

wu-sheng:
You need to use `infra-e2e` tool when you resolve conflicts. The v1 e2e framework is being removed.

oap-server-bom/pom.xml
```diff
@@ -77,6 +77,7 @@
         <awaitility.version>3.0.0</awaitility.version>
         <httpcore.version>4.4.13</httpcore.version>
         <commons-compress.version>1.21</commons-compress.version>
+        <iotdb-session.version>0.13.0-SNAPSHOT</iotdb-session.version>
```
wu-sheng, line 80: FYI, we need to fix this before merging.

test/e2e-v2/cases/alarm/iotdb/docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+version: '2.1'
+
+services:
+  iotdb:
+#    image: apache/iotdb:0.12.2-node
+    image: timcross0/iotdb
```
wu-sheng, line 21: FYI, we need to fix this before merging, including all CI e2e related images.

tools/dependencies/known-oap-backend-dependencies.txt
```diff
@@ -89,12 +98,16 @@ jsr305-3.0.2.jar
...
 listenablefuture-9999.0-empty-to-avoid-conflict-with-guava.jar
 log4j-api-2.14.1.jar
 log4j-core-2.14.1.jar
 log4j-over-slf4j-1.7.30.jar
 log4j-slf4j-impl-2.14.1.jar
+logback-classic-1.2.3.jar
+logback-core-1.2.3.jar
 logging-interceptor-3.13.1.jar
+lz4-1.3.0.jar
 lz4-java-1.6.0.jar
```
wu-sheng, line 111: OMG, these are 2 different libs. :(

- https://github.com/lz4/lz4
- https://github.com/lz4/lz4-java

tools/dependencies/known-oap-backend-dependencies.txt
```diff
@@ -61,17 +61,26 @@ httpclient-4.5.13.jar
 httpcore-4.4.13.jar
 httpcore-nio-4.4.13.jar
 influxdb-java-2.15.jar
+iotdb-session-0.13.0-20211102.115555-1.jar
+iotdb-thrift-0.13.0-20211102.115555-1.jar
```
wu-sheng, line 65: This will be updated after the official release.

wu-sheng:
```
--- /dev/fd/63	2021-11-04 03:22:37.362685756 +0000
+++ /dev/fd/62	2021-11-04 03:22:37.362685756 +0000
@@ -66 +65,0 @@
-istack-commons-runtime-4.0.0.jar
@@ -74,2 +72,0 @@
-jakarta.activation-2.0.0.jar
-jakarta.xml.bind-api-3.0.0.jar
@@ -78 +74,0 @@
-javax.activation-api-1.2.0.jar
@@ -81,3 +76,0 @@
-jaxb-api-2.4.0-b180830.0359.jar
-jaxb-core-3.0.0.jar
-jaxb-runtime-3.0.0.jar
@@ -158 +150,0 @@
-twx2-3.0.0.jar
```

I think there is some missing about dependency libs.

tools/dependencies/known-oap-backend-dependencies.txt
```diff
@@ -61,17 +61,26 @@ httpclient-4.5.13.jar
...
+istack-commons-runtime-4.0.0.jar
 jackson-annotations-2.12.2.jar
 jackson-core-2.12.2.jar
 jackson-databind-2.12.2.jar
 jackson-datatype-guava-2.9.10.jar
 jackson-datatype-jdk8-2.9.10.jar
 jackson-module-afterburner-2.12.2.jar
 jackson-module-kotlin-2.8.8.jar
+jakarta.activation-2.0.0.jar
+jakarta.xml.bind-api-3.0.0.jar
```
wu-sheng, line 75: FYI @kezhenxu94 I think our maven cache may cause some issues when we use `SNAPSHOT` for draft PR.

image

I and @LIU-WEI-git both packaged locally, these jars exist, but CI checks on dependencies say otherwise.
```
@@ -74,2 +72,0 @@
-jakarta.activation-2.0.0.jar
-jakarta.xml.bind-api-3.0.0.jar
```

I don't think we have to fix this, as we never would accept a PR using snapshot dependency. Just find out this curious case.

kezhenxu94: I think it is your local cache that cause this issue, because when maven downloads a SNAPSHOT package it will firstly check whether it exists in your local repository (typically `~/.m2/repository`), and if exists it will be used directly, but SNAPSHOT can be overrode in remote repository so you might be using an old revision, as I check the latest revision of this iotdb-session SNAPSHOT is `iotdb-session-0.13.0-20211102.115555-1.jar Tue Nov 02 14:07:09 UTC 2021  67297` so you might want to update the local SNAPSHOT and repackage to see the changes, generally adding `-U` will do the trick (`./mvnw -U ....`)

wu-sheng: `snapshot` is not always using the local repository, it just has an expired time. 

I did local check again with your recommend `-U`, still get the same result.
image

I checked my iotdb dependencies, they are the latest.
image

My question is not about IoTDB session version, I was asking about `jakata*`, which CI says they don't exist.

kezhenxu94: I build the package and didn't find these 2 jar as well, the same as what we see in GitHub Actions.

```diff
@@ -74,2 +72,0 @@
-jakarta.activation-2.0.0.jar
-jakarta.xml.bind-api-3.0.0.jar
```

kezhenxu94: I also don't find any related jars in the dependency tree of new module storage-iotdb-plugin

wu-sheng: OK, it seems need to delete some local jars. Anyway, this would not be an issue after official release.

wu-sheng:
@kezhenxu94 @JaredTan95 @dmsolr Could you check this PR carefully? I think this would be the biggest PR in this iteration.
We should be careful. Although all E2E tests passed, I would recommend to verify this PR locally and check visualize on UI manually, at least this time.

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+        <groupId>org.apache.skywalking</groupId>
+        <version>8.9.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>storage-iotdb-plugin</artifactId>
+    <packaging>jar</packaging>
+
+    <properties>
+    </properties>
```
kezhenxu94, line 34:
```suggestion
```

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+                    <groupId>net.jpountz.lz4</groupId>
+                    <artifactId>lz4</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.lz4</groupId>
+            <artifactId>lz4-java</artifactId>
+            <version>1.6.0</version>
+        </dependency>
```
kezhenxu94, line 61: I don't see there is `lz4` in `iotdb-session`'s dependency tree, please recheck

dist-material/release-docs/LICENSE
```diff
@@ -394,6 +399,8 @@ The following components are provided under the EPL License. See project link fo
 The text of each license is also included at licenses/LICENSE-[project].txt.
 
     logback 1.1.11: https://github.com/qos-ch/logback: EPL 1.0
+    logback-classic 1.2.3: https://github.com/qos-ch/logback: EPL 1.0
```
kezhenxu94, line 402: Can you please also update the version in the previous line (401) 👆 , it should be 1.2.3 as well

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/IoTDBClient.java
```diff
@@ -0,0 +1,342 @@
...
+
+    /**
+     * Write data to IoTDB
+     *
+     * @param request an IoTDBInsertRequest
+     * @throws IOException IoTDBConnectionException or StatementExecutionException
+     */
+    public void write(IoTDBInsertRequest request) throws IOException {
+        if (log.isDebugEnabled()) {
+            log.debug(request.toString());
```
kezhenxu94, line 117: Can you add some descriptive message to the log so that we can search them when needed, such as

```suggestion
            log.debug("Writing data to IoTDB: {}", request);
```

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/base/IoTDBInsertRequest.java
```diff
@@ -0,0 +1,109 @@
...
+        for (String key : storageMap.keySet()) {
+            if (key.equals(IoTDBClient.TIMESTAMP) || key.contains(".")) {
+                int idx = measurements.indexOf(key);
+                measurements.set(idx, "\"" + key + "\"");
+            }
+        }
+    }
+
+    @Override
+    public String toString() {
```
kezhenxu94, line 98: Use `@lombok.ToString` instead

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/IoTDBStorageProvider.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @Override
+    public void prepare() throws ServiceNotProvidedException {
+        this.registerServiceImplementation(StorageBuilderFactory.class, new StorageBuilderFactory.Default());
+
+        try {
+            client = new IoTDBClient(config);
+        } catch (IOException e) {
+            e.printStackTrace();
```
kezhenxu94, line 108: Never do this in production code, @codeglzhang will you be able to add this prohibition any soon? As we discussed in the other PR

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/IoTDBTableMetaInfo.java
```diff
@@ -0,0 +1,118 @@
...
+            if (columnName.equals(IoTDBClient.ENTITY_ID_IDX) || columnName.equals(IoTDBClient.NODE_TYPE_IDX) ||
+                    columnName.equals(IoTDBClient.SERVICE_ID_IDX) || columnName.equals(IoTDBClient.GROUP_IDX) ||
+                    columnName.equals(IoTDBClient.TRACE_ID_IDX)) {
+                storageAndIndexMap.put(column.getColumnName().getStorageName(), columnName);
+            } else {
+                columnAndTypeMap.put(columnName, typeToTSDataType(column.getType()));
+            }
+        });
+
+        // TODO optimize map to list
```
kezhenxu94, line 64: Seems this is done?

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
wu-sheng: Check the previous exclusive, the tricky thing here is, this repo changed the artifact ID. This is manual override dependency, and keep consistent with our existing dependency.

But the version should be managed in the bom.

kezhenxu94: I mean there is no `lz4`, even `net.jpountz.lz4:lz4`, in `iotdb-session`

wu-sheng: OK, @LIU-WEI-git please recheck, I used to notice this jar in previous dependency check report.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/IoTDBStorageProvider.java
LIU-WEI-git: ok, I have fix it.

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
LIU-WEI-git: image
It still has lz4-1.3.0.

wu-sheng: @kezhenxu94 I can see this now.

image

kezhenxu94: OK, sorry @LIU-WEI-git , it does indeed include an old version of `lz4`, please add the `exclusion` back and move the new `lz4-java` to the `bom` module

LIU-WEI-git: ok

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
```diff
@@ -44,6 +44,16 @@
         <dependency>
             <groupId>org.apache.iotdb</groupId>
             <artifactId>iotdb-session</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>net.jpountz.lz4</groupId>
+                    <artifactId>lz4</artifactId>
+                </exclusion>
```
wu-sheng, line 51: @jixuan1989 We manually replaced this with `1.6.0` to avoid 2 versions existing together in the final binary.

docs/en/setup/backend/backend-storage.md
```diff
@@ -280,6 +280,24 @@ storage:
 All connection-related settings, including URL link, username, and password are found in `application.yml`. 
 Only part of the settings are listed here. Please follow [HikariCP](https://github.com/brettwooldridge/HikariCP) connection pool document for full settings.
 
+## IoTDB
```
wu-sheng, line 283: I think you missed the option list on the top of file.

LIU-WEI-git: Got it, I will fix it in the next commit

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/IoTDBTableInstaller.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package org.apache.skywalking.oap.server.storage.plugin.iotdb;
+
+import org.apache.skywalking.oap.server.core.storage.model.Model;
+import org.apache.skywalking.oap.server.core.storage.model.ModelInstaller;
+import org.apache.skywalking.oap.server.library.client.Client;
+import org.apache.skywalking.oap.server.library.module.ModuleManager;
+
+public class IoTDBTableInstaller extends ModelInstaller {
```
wu-sheng, line 26: Could you consider to add a long comment here about what are typically tables look like? Such as including 5 types
1. Metadata, use `service_traffic` for example.
2. Metrics, use `service_cpm`
3. Trace segment
4. Log
5. Profiling snapshots

With all these, when we face bugs one day, or with IoTDB improvement, we will know the expected table structure.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/base/IoTDBHistoryDeleteDAO.java
```diff
@@ -0,0 +1,44 @@
...
+@Slf4j
+@RequiredArgsConstructor
+public class IoTDBHistoryDeleteDAO implements IHistoryDeleteDAO {
+    private final IoTDBClient client;
+
+    @Override
+    public void deleteHistory(Model model, String timeBucketColumnName, int ttl) throws IOException {
+        if (log.isDebugEnabled()) {
+            log.debug("TTL execution log, model: {}, TTL: {}", model.getName(), ttl);
+        }
```
wu-sheng, line 40: Do we really need logs like this? Even we need, I think we should move this to general DataTTLKeeperTimer, rather than implementation. Feel free to raise another PR to add this if needed.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/base/IoTDBInsertRequest.java
```diff
@@ -0,0 +1,98 @@
...
+
+        measurements = new ArrayList<>(storageMap.keySet());
+        Map<String, TSDataType> columnAndTypeMap = IoTDBTableMetaInfo.get(modelName).getColumnAndTypeMap();
+        measurementTypes = new ArrayList<>(measurements.size());
+        for (String measurement : measurements) {
+            measurementTypes.add(columnAndTypeMap.get(measurement));
+        }
+        measurementValues = new ArrayList<>(storageMap.values());
+
+        // processing illegal measurement
```
wu-sheng, line 90: What do you mean illegal? I think this is a wrong word chosen? Maybe you mean to process several special fields which formats unsupported by IoTDB?

LIU-WEI-git: Ok, I will rewrite it to make it clear.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/IoTDBTableInstaller.java
LIU-WEI-git: I will deal with it these days.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/base/IoTDBHistoryDeleteDAO.java
LIU-WEI-git: Influxdb storage plugin has this log as well. I think we could keep it.

wu-sheng (dismissed):
The general code structure and documents are good. Wait for @dmsolr confirmation.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/IoTDBTableInstaller.java
```diff
@@ -0,0 +1,220 @@
...
+ */
+
+package org.apache.skywalking.oap.server.storage.plugin.iotdb;
+
+import org.apache.skywalking.oap.server.core.storage.model.Model;
+import org.apache.skywalking.oap.server.core.storage.model.ModelInstaller;
+import org.apache.skywalking.oap.server.library.client.Client;
+import org.apache.skywalking.oap.server.library.module.ModuleManager;
+
+/**
```
kezhenxu94, line 26: As you have moved this to the official website, I'd just paste the link here instead of pasting the entire content, the markdown won't be rendered correctly via JavaDoc engine

LIU-WEI-git: ok, I will replace it with blog link in the next commit.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBBrowserLogQueryDAO.java
```diff
@@ -0,0 +1,126 @@
...
+        List<? super StorageData> storageDataList = client.filterQuery(BrowserErrorLogRecord.INDEX_NAME, queryString, storageBuilder);
+        List<BrowserErrorLogRecord> browserErrorLogRecordList = new ArrayList<>(storageDataList.size());
+        storageDataList.forEach(storageData -> browserErrorLogRecordList.add((BrowserErrorLogRecord) storageData));
+        // resort by self, because of the select query result order by time.
+        browserErrorLogRecordList.sort((BrowserErrorLogRecord b1, BrowserErrorLogRecord b2) -> Long.compare(b2.getTimestamp(), b1.getTimestamp()));
+        BrowserErrorLogs logs = new BrowserErrorLogs();
+        int limitCount = 0;
+        for (int i = 0; i < browserErrorLogRecordList.size(); i++) {
+            if (i >= from && limitCount < limit) {
+                limitCount++;
```
dmsolr, line 91: Why do not we start from `start` in the loop?

LIU-WEI-git: Oh, that's my carelessness. I will fix it tomorrow.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/IoTDBTableInstaller.java
wu-sheng: Make sense to me.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBBrowserLogQueryDAO.java
```diff
@@ -0,0 +1,126 @@
...
+        if (StringUtil.isNotEmpty(pagePathId)) {
+            query.append(" and ").append(BrowserErrorLogRecord.PAGE_PATH_ID).append(" = \"").append(pagePathId).append("\"");
+        }
+        if (Objects.nonNull(category)) {
+            query.append(" and ").append(BrowserErrorLogRecord.ERROR_CATEGORY).append(" = ").append(category.getValue());
+        }
+        query.append(IoTDBClient.ALIGN_BY_DEVICE);
+        // IoTDB doesn't support the query contains "1=1" and "*" at the meantime.
+        String queryString = query.toString();
+        queryString = queryString.replace("1=1 and ", "");
```
dmsolr, line 80: Is there a case that does not contain any condition? If so, the result will be `select * from tbl where 1=1`

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBEventQueryDAO.java
```diff
@@ -0,0 +1,184 @@
...
+            }
+            if (time.getEndTimestamp() > 0) {
+                query.append(" and ").append(Event.END_TIME).append(" < ").append(time.getEndTimestamp());
+            }
+        }
+        // IoTDB doesn't support the query contains "1=1" and "*" at the meantime.
+        String queryString = query.toString();
+        queryString = queryString.replace("1=1 and ", "");
+        return new StringBuilder(queryString);
+    }
```
dmsolr, line 157: the same

dmsolr: A suggestion,
we can be the where conditions in a new StringBuilder to avoid operating string frequently. like following:
```java
StringBuilder where = new StringBuilder();
if (!Strigs.isNullOrEmpty(condition.getUuid())) {
    where.append(Event.UUID).append(" = \"").append(condition.getUuid()).append("\"").append(" and ");
}

if (where.length() > 0) {
    int length = where.length();
    where.delete(length - 4, length);
    query.append(" where ").append(where);
}

return query;
```

dmsolr (commented):
I think you could polish the code related to string operation. :)

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBBrowserLogQueryDAO.java
LIU-WEI-git: Got it.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBEventQueryDAO.java
LIU-WEI-git: Thanks for your suggestion. I will deal with it.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBTopologyQueryDAO.java
```diff
@@ -0,0 +1,255 @@
...
+                log.debug("SQL: {}, columnNames: {}", query, wrapper.getColumnNames());
+            }
+            while (wrapper.hasNext()) {
+                RowRecord rowRecord = wrapper.next();
+                List<Field> fields = rowRecord.getFields();
+                Call.CallDetail call = new Call.CallDetail();
+                String[] layerNames = fields.get(0).getStringValue().split("\\" + IoTDBClient.DOT + "\"");
+                String entityId = client.layerName2IndexValue(layerNames[2]);
+                // TODO check DetectPoint
+                call.buildFromEndpointRelation(entityId, DetectPoint.SERVER);
```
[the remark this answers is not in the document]
wu-sheng: image

This is client side detecting, what do you mean exactly?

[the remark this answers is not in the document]
LIU-WEI-git: > This is client side detecting, what do you mean exactly?

No, it's not this method. In line 291 in `TopologyQueryEsDAO` and line 270 in `H2TopologyQueryDAO`, the second parameter of `buildFromEndpointRelation` always is `DetectPoint.SERVER`. But in line 305 of `TopologyQuery`, it can be `DetectPoint.SERVER` or `DetectPoint.CLIENT`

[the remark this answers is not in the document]
wu-sheng: For endpoint, yes. The dependency is only `server` side detection.

dmsolr (dismissed):
LGTM

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/IoTDBClient.java
```diff
@@ -0,0 +1,342 @@
...
+                }
+
+                storageDataList.add(storageBuilder.storage2Entity(map));
+            }
+            healthChecker.health();
+        } catch (IoTDBConnectionException | StatementExecutionException e) {
+            healthChecker.unHealth(e);
+            throw new IOException(e.getMessage() + System.lineSeparator() + "SQL Statement: " + querySQL, e);
+        } finally {
+            sessionPool.closeResultSet(wrapper);
```
kezhenxu94, line 254: Add a null check for `wrapper` in case `wrapper = sessionPool.executeQueryStatement(querySQL);` throws exception and `wrapper` is still null, the same for many other places

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBAggregationQueryDAO.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public List<SelectedRecord> sortMetrics(TopNCondition condition, String valueColumnName, Duration duration,
+                                            List<KeyValue> additionalConditions) throws IOException {
+        // This method maybe have poor efficiency. It queries all data which meets a condition without aggregation function.
+        // https://github.com/apache/iotdb/issues/4006
+        StringBuilder query = new StringBuilder();
+        query.append(String.format("select %s from ", valueColumnName));
+        query = client.addModelPath(query, condition.getName());
+        query.append(" where ").append(IoTDBClient.TIME).append(" >= ").append(duration.getStartTimestamp())
+                .append(" and ").append(IoTDBClient.TIME).append(" <= ").append(duration.getStartTimestamp());
```
kezhenxu94, line 57:
```suggestion
                .append(" and ").append(IoTDBClient.TIME).append(" <= ").append(duration.getEndTimestamp());
```

kezhenxu94 (changes requested):

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBAggregationQueryDAO.java
```diff
@@ -0,0 +1,110 @@
...
+                entityIdAndSumMap.merge(entityId, value, Double::sum);
+                entityIdAndCountMap.merge(entityId, 1, Integer::sum);
+            }
+
+            entityIdAndSumMap.forEach((String entityId, Double sum) -> {
+                double count = entityIdAndCountMap.get(entityId);
+                double avg = sum / count;
+                SelectedRecord topNEntity = new SelectedRecord();
+                topNEntity.setId(entityId);
+                topNEntity.setValue(String.valueOf(avg));
```
kezhenxu94, line 92: Cast this to `long`

```suggestion
                topNEntity.setValue(String.valueOf((long) avg));
```

kezhenxu94: We only have `long` type value in metrics so this double should be cast to `long`

LIU-WEI-git: ok, thanks you for pointing out.

kezhenxu94 (changes requested):
I also found that the query to endpoint metrics is empty, which might be relevant to `org.apache.skywalking.oap.server.storage.plugin.iotdb.query.IoTDBAggregationQueryDAO#sortMetrics`, please fix this

Screen Shot 2021-11-25 at 17 06 57

wu-sheng:
Waiting to TopN verification. @LIU-WEI-git Please confirm when you are done.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/IoTDBClient.java
```diff
@@ -0,0 +1,352 @@
...
+                    String columnName = columnNames.get(i + 2);
+                    Field field = fields.get(i + 1);
+                    if (field.getDataType() == null) {
+                        continue;
+                    }
+                    if (field.getDataType().equals(TSDataType.TEXT)) {
+                        map.put(columnName, field.getStringValue());
+                    } else {
+                        map.put(columnName, field.getObjectValue(field.getDataType()));
+                    }
```
kezhenxu94, line 230: You may need to check whether the `field` is null or not, it seems `field.getStringValue()` returns a string `"null"` if `field` is null

kezhenxu94:
Others LGTM

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/IoTDBClient.java
kezhenxu94: Specifically, this might cause the service group name to be “null” in UI topology.

LIU-WEI-git: `field` won't be null. What's more, if IoTDB stores a null value, `field.getDataType()` will return null and this condition will be deal with in the line 223.

kezhenxu94: Please check locally. The group in topology is "null", which is wrong

kezhenxu94: I didn't dig deeper to see what's the problem but "null" string exists indeed

kezhenxu94: Or maybe there are some places where you use `String.valueOf()` without checking nullability

LIU-WEI-git: When inserting data, the `service_group` will be "null" if user doesn't set since `service_group` is one of the index and its null value is casted to "null". I have located this bug and will fix it.

LIU-WEI-git:
> Waiting to TopN verification. @LIU-WEI-git Please confirm when you are done.

I have confirmed it. It works well.
e91f78ecc9b398395bc4743433b9572

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/base/IoTDBInsertRequest.java
```diff
@@ -0,0 +1,104 @@
...
+        indexes = IoTDBTableMetaInfo.get(modelName).getIndexes();
+        indexValues = new ArrayList<>(indexes.size());
+        Map<String, Object> storageMap = storageBuilder.entity2Storage(storageData);
+
+        indexes.forEach(index -> {
+            if (index.equals(IoTDBIndexes.ID_IDX)) {
+                indexValues.add(storageData.id());
+            } else if (storageMap.containsKey(index)) {
+                // avoid `service_group` be "null" when inserting
+                if (index.equals(IoTDBIndexes.GROUP_IDX) && storageMap.get(index) == null) {
```
kezhenxu94, line 65: Let’s apply this to other fields. I don’t think we want “null” string for null value in all other fields. @wu-sheng correct me if I’m wrong.

LIU-WEI-git: If I don't use the null check for `Group_IDX`,  `String.valueOf(null)` in line 68 will return "null". In production, all other indexes  are xxx_id and they won't be null. But `Group_IDX` maybe null.

LIU-WEI-git: If other non-indexed fields is null, we won't insert it into IoTDB server. Therefore, their query results will be null, won't be "null". Please see line 81.

kezhenxu94 (approved):
LGTM

wu-sheng (approved):

PR #7772, opened 2021-09-23
Only report `bug warning` in precise conditions.

wu-sheng:

User will see the `[Bug warning]` in 2 typical cases
1. If the template exists and index deleted, or in reverse. In this case, this isn't the bug report we expected.
2. OAP shut down over a day and latest index doesn't exist for sure. In this case, we just need to return false, and `create` method will fix this.

### DIFF
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
@@ -70,13 +70,16 @@ public class StorageEsInstaller extends ModelInstaller {
         if (!model.isTimeSeries()) {
             return esClient.isExistsIndex(tableName);
         }
-        boolean exist = esClient.isExistsTemplate(tableName)
-            && esClient.isExistsIndex(TimeSeriesUtils.latestWriteIndexName(model));
+        boolean templateExists = esClient.isExistsTemplate(tableName);
         final Optional<IndexTemplate> template = esClient.getTemplate(tableName);
+        boolean lastIndexExists = esClient.isExistsIndex(TimeSeriesUtils.latestWriteIndexName(model));
 
-        if ((exist && !template.isPresent()) || (!exist && template.isPresent())) {
-            throw new Error("[Bug warning]ElasticSearch client query template result is not consistent. Please file an issue to Apache SkyWalking.(https://github.com/apache/skywalking/issues)");
+        if ((templateExists && !template.isPresent()) || (!templateExists && template.isPresent())) {
+            throw new Error("[Bug warning] ElasticSearch client query template result is not consistent. " +
+                                "Please file an issue to Apache SkyWalking.(https://github.com/apache/skywalking/issues)");
         }
+
+        boolean exist = templateExists && lastIndexExists;
 
         if (exist && IndexController.INSTANCE.isMetricModel(model)) {
             structures.putStructure(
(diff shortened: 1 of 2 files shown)


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -70,14 +70,17 @@ protected boolean isExists(Model model) {
...
         }
-        boolean exist = esClient.isExistsTemplate(tableName)
-            && esClient.isExistsIndex(TimeSeriesUtils.latestWriteIndexName(model));
+        boolean templateExists = esClient.isExistsTemplate(tableName);
         final Optional<IndexTemplate> template = esClient.getTemplate(tableName);
+        boolean lastIndexExists = esClient.isExistsIndex(TimeSeriesUtils.latestWriteIndexName(model));
 
-        if ((exist && !template.isPresent()) || (!exist && template.isPresent())) {
-            throw new Error("[Bug warning]ElasticSearch client query template result is not consistent. Please file an issue to Apache SkyWalking.(https://github.com/apache/skywalking/issues)");
+        if ((templateExists && !template.isPresent()) || (!templateExists && template.isPresent())) {
```
kezhenxu94, line 77: If this is the only condition, I think a basic test should cover it, this is a very basic logic and should be discovered at test phase, not runtime

```java
        assertThat(templateClient.exists(name)).isTrue();
```

https://github.com/apache/skywalking/blob/73b6c67344d83cabe3014fc33898eb15f3bc4063/oap-server/server-library/library-elasticsearch-client/src/test/java/org/apache/skywalking/library/elasticsearch/ITElasticSearchTest.java#L102-L107

wu-sheng: I was proposing this due to we can't test all server versions, and as we are not official, we can't tell when breaking change happens.
This exception codes should never be executed.

wu-sheng: The current conditions checking could fail in daily used, it is not a bug, and we have to suffer endless question and bug report on GitHub.

kezhenxu94: > The current conditions checking could fail in daily used, it is not a bug, and we have to suffer endless question and bug report on GitHub.

I don't mean to keep the current conditions, I mean if the condition is to be `if ((templateExists && !template.isPresent()) || (!templateExists && template.isPresent()))`, then this is a very basic function that we should test in the perspective of ES client.

> I was proposing this due to we can't test all server versions, and as we are not official, we can't tell when breaking change happens.

The supported versions are fixed and the codes won't run on un-tested versions.

https://github.com/apache/skywalking/blob/73b6c67344d83cabe3014fc33898eb15f3bc4063/oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/ElasticSearchVersion.java#L67

We will manually test a new ES version, and add it into our supported list, otherwise if users run on a future ES server for example 8.2, the OAP won't even start

kezhenxu94: It's good to me to keep the new condition, but please also add a test to https://github.com/apache/skywalking/blob/73b6c67344d83cabe3014fc33898eb15f3bc4063/oap-server/server-library/library-elasticsearch-client/src/test/java/org/apache/skywalking/library/elasticsearch/ITElasticSearchTest.java#L102-L107

```java
        assertThat(templateClient.exists(name)).isTrue();
```

wu-sheng: > The supported versions are fixed and the codes won't run on un-tested versions.

I know we have `{"6.3.2"}, {"7.4.2"}, {"7.8.0"}, {"7.10.2"}`, but who knows one day, something breaks unexpected, such as we never would know in the `7.x` there are different cases.

> It's good to me to keep the new condition, but please also add a test to

Added check for template existing check.

kezhenxu94 (approved):

PR #7784, opened 2021-09-24
Fix license of netty-tcnative-boringssl-static

wu-sheng:

There are 2 issues about this license catalog.
1. This is Netty dependency, not Apache's.
2. This library is using standard APLv2, so, we don't need to list it separate.


hanahmily (approved):
May the license analyzer we used to classify licenses automatically cause it. I don't have any idea about this now

PR #7790, opened 2021-09-24
Fix entity naming mechanism failure in MAL

wu-sheng:

Fix entity(service/instance/endpoint) names in the MAL system(prometheus, native meter, open census, envoy metric service) are not controlled by core's naming-control mechanism.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/MeterEntity.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/MeterEntity.java
@@ -25,6 +25,7 @@ import lombok.Setter;
 import lombok.ToString;
 import org.apache.skywalking.oap.server.core.UnexpectedException;
 import org.apache.skywalking.oap.server.core.analysis.IDManager;
+import org.apache.skywalking.oap.server.core.config.NamingControl;
 import org.apache.skywalking.oap.server.core.source.DetectPoint;
 
 /**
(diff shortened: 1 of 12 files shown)


hanahmily (approved):
LGTM

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/MeterEntity.java
```diff
@@ -35,6 +36,8 @@
 @Getter
 @Builder(toBuilder = true)
 public class MeterEntity {
+    private static NamingControl NAMING_CONTROL;
```
kezhenxu94, line 39: Where is it used?

wu-sheng: I forgot to submit the second change. Just did.

kezhenxu94 (changes requested):
This fix is fragile, `MeterEntity` exposes many means to construct its instances, this patch only covers the static methods `newService`, `newServiceInstance`, `newEndpoint`, `newServiceRelation`, future developers who use the uncovered cases to construct the entity might cause the bug again.

The other uncovered cases:

1. `MeterEntity.builder().serviceName("").instanceName("")...`, if it's designed to only use the `newXxxx` methods to construct `MeterEntity`, we should update the `@Builder` annotation to `@Builder(toBuilder = true, access = AccessLevel.PRIVATE)`
2. `meterEntity.toBuilder()` returns another builder, which is the same case with (1.), we'd better also remove `toBuilder = true` in `@Builder`
3. `new` constructor (although it's package-private), we'd better also make it `private`.

wu-sheng:
@kezhenxu94 I removed `@Builder` and fixed the tests.

kezhenxu94 (approved):

PR #7795, opened 2021-09-24
SW_STORAGE default is elasticsearch7

macaty:
SW_STORAGE default is elasticsearch7

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -55,6 +55,6 @@ From the output, we can find out the building tool adopts the files stored in `o
 We can start up backend cluster by docker-compose
 ```
 cd docker
-docker compose up
+docker-compose up -d
 ```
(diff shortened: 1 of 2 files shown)


wu-sheng (approved):

docker/README.md
```diff
@@ -55,6 +55,6 @@ From the output, we can find out the building tool adopts the files stored in `o
 We can start up backend cluster by docker-compose
 ```
 cd docker
-docker compose up
+docker-compose up -d
```
wu-sheng, line 58: This doesn't need to change, I think.

wu-sheng (changes requested):

docker/README.md
wu-sheng: `docker compose` is already a command.

kezhenxu94:
Please check the Chang log , there is no elasticsearch7 in master branch anymore, if you are using old version , check the corresponding tag

PR #7634, opened 2021-09-01
Rebuilt ElasticSearch client on top of their REST API

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7471 , closes #7655.
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md).

PR #7798, opened 2021-09-24
Add `socketTimeout` back to the new implementation

kezhenxu94:
Related to #7634 , which ignored the config `socketTimeout`, now I add it back.


wu-sheng:
Please fix code style

wu-sheng (approved):

PR #7820, opened 2021-09-27
Set the Docker image names the same as the ones in hub.docker.io

kezhenxu94:
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md). NO NEED.


wu-sheng:
It is better to keep our name in GitHub docker repo? This repository name is using in different other repositories.
I asked from https://github.com/apache/skywalking/issues/7819.

wu-sheng:
Update, I also tried this new version of script, it still fails
```
Successfully tagged apache/skywalking-oap-server:8.8.0
real    0m41.949s
user    0m0.108s
sys     0m0.279s
time (docker push apache/oap-server:8.8.0)
The push refers to repository [docker.io/apache/oap-server]
An image does not exist locally with the tag: apache/oap-server
real    0m0.068s
user    0m0.016s
sys     0m0.028s
Makefile:87: recipe for target 'push.docker.oap-server' failed
make: *** [push.docker.oap-server] Error 1
```

wu-sheng (approved):

PR #7826, opened 2021-09-28
test: fix that E2E is not run, also add ElasticSearch 7.15 to test matrix

kezhenxu94:


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/TraceQueryEsDAO.java
```diff
@@ -121,7 +121,7 @@ public TraceBrief queryBasicTraces(long startSecondTB,
         }
         if (CollectionUtils.isNotEmpty(tags)) {
             BoolQueryBuilder tagMatchQuery = Query.bool();
-            tags.forEach(tag -> tagMatchQuery.must(Query.term(SegmentRecord.TAGS, tag)));
+            tags.forEach(tag -> tagMatchQuery.must(Query.term(SegmentRecord.TAGS, tag.toString())));
```
kezhenxu94, line 124: Before: `tag` was serialized as `{"key":"value"}` and was considered as ES query syntax.
Now: `tag` is `key=value`

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/ElasticSearchVersion.java
```diff
@@ -43,7 +43,7 @@ private ElasticSearchVersion(final String distribution, final int major, final i
 
         if (distribution.equalsIgnoreCase("OpenSearch")) {
             requestFactory = new V78RequestFactory(this);
-            codec = V7Codec.INSTANCE;
+            codec = V78Codec.INSTANCE;
```
kezhenxu94, line 46: Forgot to update to `V78Codec` after splitting the de/ser for `[7.0, 7.8)` `[7.8, 8.0)`

wu-sheng:
We will need 8.8.1 to include this fix.

wu-sheng (approved):

wu-sheng (approved):

PR #7849, opened 2021-09-30
Fix wrong rules for checking lists in E2E

kezhenxu94:


kezhenxu94:
Just fix the wrong rules, there are something need to check:

- [ ] why the 2 cases trigger the same endpoint but the expected topology are different

wankai123:
> Just fix the wrong rules, there are something need to check:
> 
> * [ ]  why the 2 cases trigger the same endpoint but the expected topology are different

I think the returned data is not in order, the input `serviceID` here could be different each time？

```
- query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql dependency service $(swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql service ls|grep -B 1 'provider'|yq e '.[0].id' -)
```

test/e2e-v2/cases/simple/auth/e2e.yaml
```diff
@@ -72,4 +72,4 @@ verify:
       expected: ../expected/traces-list.yml
     # trace detail
     - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql trace $(swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql trace ls|grep -A 5 '/info'|tail -n1|awk -F ' ' '{print $2}')
```
wankai123, line 74: Does the trace detail have the same issue?

wu-sheng:
All metadata query results are not in order. You should either do name grep filter or encode name by yourself rather than depending on metadata query.

test/e2e-v2/cases/satellite/native-protocols/expected/dependency-services.yml
```diff
@@ -14,35 +14,18 @@
 # limitations under the License.
 
 nodes:
-{{- range .nodes }}
```
kezhenxu94, line 17: FYI @mrproliu @wankai123 @dmsolr @arugal , `range` + `if` is not designed to verify lists containing specific elements, can you please review other repos E2E and see whether this style is used, if so, please also update them?

For example, the following rules expect there are 4 nodes, but it still passed even if there is only one node (one arbitrary node in the 4 listed):

```yaml
nodes:
{{- range .nodes }}
{{- if eq .name "User" }}
- id: {{ b64enc "User" }}.0
  name: User
  type: USER
  isreal: false
{{- end }}
{{- if eq .name "e2e-service-provider" }}
- id: {{ b64enc "e2e-service-provider"}}.1
  name: e2e-service-provider
  type: Tomcat
  isreal: true
{{- end }}
{{- if eq .name "e2e-service-consumer" }}
- id: {{ b64enc "e2e-service-consumer"}}.1
  name: e2e-service-consumer
  type: Tomcat
  isreal: true
{{- end }}
{{- if eq .name "localhost:-1" }}
- id: {{ b64enc "localhost:-1" }}.0
  name: localhost:-1
  type: H2
  isreal: false
{{- end }}
{{- end }}
```

wu-sheng: What is the recommended way? Seems should separate to verify one node per range?

wu-sheng: What is the recommended way? Seems should separate to verify one node per range?

kezhenxu94: > What is the recommended way? Seems should separate to verify one node per range?

Check what I updated in this patch. `contains` is designed to do this.

wu-sheng: @mrproliu @wankai123 @arugal You should have other cases having the same issue.

test/e2e-v2/cases/satellite/native-protocols/e2e.yaml
```diff
@@ -51,17 +51,17 @@ verify:
     - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql service ls
       expected: expected/service.yml
     # basic check: service metrics
-    - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql metrics linear --name service_sla --service=e2e-service-provider |yq e 'to_entries' -
+    - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql metrics linear --name service_sla --service=e2e-service-provider | yq e 'to_entries' -
       expected: expected/metrics-has-value.yml
     # basic check: service endpoint
-    - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql endpoint list --keyword=info --service-id=$(swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql service ls|grep -B 1 'provider'|yq e '.[0].id' -)
```
kezhenxu94, line 57: @mrproliu avoid using `grep -A` or `grep -B` in the future, which is unstable given the returned order from CLI is unpredictable

wu-sheng:
The go2sky case keeps failure. Strange, we don't use e2e for it yet.

kezhenxu94:
Ready to merge now

PR #7857, opened 2021-10-04
Rename `service` and `instance` labels when origin metrics are contained.

dmsolr:


wu-sheng:
Could you be clear what is the change? This seems a breaking change for existing cases.

wu-sheng:
MAL engine doesn't care what labels are from where.
Such as Prometheus fetcher could add target as labels in default. OpenCensus receiver and Envoy metrics service handler, also LAL generated metrics don't have the idea of sources. 
Please be clear about your cases.

dmsolr:
It is not a good approach to resolve my issue so that I l closed it

PR #7893, opened 2021-10-10
Remove component definitions in Java codes. 

wu-sheng:

Remove component definitions in Java codes. component-libraries.yml controls the ownership now.


wu-sheng:
I noticed this from several PRs about adding component IDs.

PR #7985, opened 2021-10-21
Add MicroBench module to make it easier for developers to write JMH Test

### LINKED ISSUE #7910: [Feature] Add MicroBench module to make it easier for developers to write JMH TEST
### Search before asking

### Description

When I wanted to write JMH Test in SkyWalking, I found that: JMHTest is distributed in various modules, and each class needs to do things like setting basic properties: For example:

```
 public static void main(String[] args) throws RunnerException, FileNotFoundException {

            .include(EndpointGroupingBenchmark4Openapi.class.getName())
            .addProfiler(GCProfiler.class)
            .jvmArgsAppend("-Xmx512m", "-Xms512m")
            .forks(1)
            .build();

        new Runner(opt).run();
    }
```

### Use case

Maybe we can create a MicroBench module to provide basic abstract classes,
The default settings include:

Warmup Iterations: 10
Measure Iterations: 10
Number of Forks: 2

etc...

These settings can be customized through system properties at runtime.

Its name may be like this: AbstractMicrobenchmark, the developer is writing JMH TEST needs to make his JMH Test extend the AbstractMicrobenchmark, which makes sure the test gets run through JUnit and configures some defaults:

another advantage is that in terms of code structure, all JMH Tests are separated from ordinary Tests.

### Related issues

_No response_

### Code of Conduct

CalvinKirs:
- [x] Closes #7910 

Writing the benchmark itself is not hard, but getting it right is. Thankfully, the JMH suite provides helpful annotations and features to mitigate most of them. To get started, you need to make your benchmark extend the AbstractMicrobenchmark, which makes sure the test gets run through JUnit and configures some defaults:
```
@BenchmarkMode({Mode.Throughput})
public class LinkedArrayBenchmark extends AbstractMicrobenchmark {
}
```

For most benchmark tests, you can write it as a normal JUnit.
You can also choose Customizing Runtime Conditions.

At present, you can run all JMH Tests by running the jar directly after packaging.
But running all JMH directly through mvn and showing that the overall report still has some problems, you can first see if it is reasonable.


oap-server/microbench/pom.xml
```diff
@@ -0,0 +1,102 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
+                <version>${exec-maven-plugin.version}</version>
+                <configuration>
+                    <executable>java</executable>
+                    <arguments>
```
kezhenxu94, line 87: Let's add a property to disable this task, as we use the same plugin when checking license dependencies, which will run this unexpectedly 

```xml
<properties>
        <benchmark.skip>false</benchmark.skip>
</properties>
```

```suggestion
                    <skip>${benchmark.skip}</skip>
                    <arguments>
```

and disable it in https://github.com/apache/skywalking/blob/26b81a2e5991ecdd2bde0f7071264cb80bb235cf/tools/dependencies/check-LICENSE.sh#L25

```diff
- ./mvnw --batch-mode -Pbackend -Dexec.executable='echo' -Dexec.args='${project.artifactId}-${project.version}.jar' exec:exec -q > self-modules.txt
+ ./mvnw --batch-mode -Pbackend -Dbenchmark.skip=true -Dexec.executable='echo' -Dexec.args='${project.artifactId}-${project.version}.jar' exec:exec -q > self-modules.txt
```

CalvinKirs: Thanks for the reminder, I'll fix it later ：）

CalvinKirs:
Now, after packaging, we can execute JMHtest through `java -jar benchmark.jar`, and if we want to write a JMHtest, we only need to inherit `AbstractMicrobenchmark` without paying too much attention to the many details of JMH.

kezhenxu94:
> Now, after packaging, we can execute JMHtest through `java -jar benchmark.jar`, and if we want to write a JMHtest, we only need to inherit `AbstractMicrobenchmark` without paying too much attention to the many details of JMH.

If we have multiple benchmark tests in benchmark module (finally `benchmark.jar`), can we choose to run only a specific class?

CalvinKirs:
> > Now, after packaging, we can execute JMHtest through `java -jar benchmark.jar`, and if we want to write a JMHtest, we only need to inherit `AbstractMicrobenchmark` without paying too much attention to the many details of JMH.
> 
> If we have multiple benchmark tests in benchmark module (finally `benchmark.jar`), can we choose to run only a specific class?

Can not do it now :sweat_smile:
If you only want to test a certain test individually, it is more appropriate to do a basic test in the local environment is more appropriate.
You can run it like a normal UT and get the result. At this time, there is usually no need for more warm-up and test iterations. We just want a rough result.

wu-sheng:
For benchmark, I am feeling do we really need to run them in a global command? Or should we only allow them runnable in specific module only?
I can't see who will run all the benchmarks.

CalvinKirs:
> For benchmark, I am feeling do we really need to run them in a global command? Or should we only allow them runnable in specific module only? I can't see who will run all the benchmarks.

What I think is that if we optimize on a certain core link, we can show our results through JMH. Often the local execution of the development environment can only get a rough result. At this time, the result is executed on a specific server in the form of jar. It may be more precise, but I can't customize a certain test, so I can only run all of them.

kezhenxu94:
> It may be more precise, but I can't customize a certain test, so I can only run all of them.

That's because you bundle the jar as an executable binary, if you treat it as a plain jar and run the specific class with that jar as a class path, it might be possible to run a single test, like `java -cp benchmark.jar org.apache.skywalking.SampleDataBenchmarkTest`.

CalvinKirs:
> > It may be more precise, but I can't customize a certain test, so I can only run all of them.
> 
> That's because you bundle the jar as an executable binary, if you treat it as a plain jar and run the specific class with that jar as a class path, it might be possible to run a single test, like `java -cp benchmark.jar org.apache.skywalking.SampleDataBenchmarkTest`.

deeply thanks for your suggestion, let me finish it

kezhenxu94:
Turns out that benchmark already allows specifying the class names/patterns to execute, so we can simply use `java -jar oap-server/microbench/target/benchmarks.jar LinkedArrayBenchmark` to only run `LinkedArrayBenchmark`.

@CalvinKirs we can document this and we are all set 🎉

oap-server/microbench/src/main/java/org/apache/skywalking/oap/server/microbench/base/AbstractMicrobenchmark.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    private ChainedOptionsBuilder newOptionsBuilder() {
+
+        String className = getClass().getSimpleName();
+
+        ChainedOptionsBuilder optBuilder = new OptionsBuilder()
+                // set benchmark class name
+                .include(".*" + className + ".*")
```
kezhenxu94, line 66: This is not needed

```suggestion
```

oap-server/microbench/src/main/java/org/apache/skywalking/oap/server/microbench/base/AbstractMicrobenchmark.java
```diff
@@ -0,0 +1,107 @@
...
+            File file = new File(filePath);
+
+            if (file.exists()) {
+                writeFileStatus = file.delete();
+            } else {
+                writeFileStatus = file.getParentFile().mkdirs();
+                try {
+                    writeFileStatus = file.createNewFile();
+                } catch (IOException e) {
+                    log.warn("jmh test create file error" + e);
```
kezhenxu94, line 87:
```suggestion
                    log.warn("jmh test create file error", e);
```

kezhenxu94 (commented):
Basically LGTM, please add the doc about how to run a single class.

BTW, it seems that you didn't import our code style, please do so and reformat the new files with our code style

oap-server/server-library/library-datacarrier-queue/pom.xml
```diff
@@ -29,8 +29,9 @@
 
     <dependencies>
         <dependency>
-            <groupId>org.openjdk.jmh</groupId>
-            <artifactId>jmh-generator-annprocess</artifactId>
+            <groupId>org.apache.skywalking</groupId>
+            <artifactId>server-testing</artifactId>
+            <version>${project.version}</version>
```
wu-sheng, line 34: I think `LinkedArrayBenchmark` has been moved into `microbench` module already. We don't remove this dependency, and existing test case should be removed too, right?

wu-sheng: Same requests for other repositories. There should be no `pom.xml` change about `jmh`->`server-testing`, right?

CalvinKirs: Yes I have removed this, thank you for checking

oap-server/microbench/src/main/java/org/apache/skywalking/oap/server/microbench/base/AbstractMicrobenchmark.java
CalvinKirs: Thanks for your checking, I have used the plug-in to check it again and fix it.

tools/dependencies/check-LICENSE.sh
```diff
@@ -22,7 +22,7 @@ tar -zxf dist/apache-skywalking-apm-bin.tar.gz -C dist
 
 # List all modules(jars) that belong to the SkyWalking itself, these will be ignored
 # when checking the dependency licenses
-./mvnw --batch-mode -Pbackend -Dexec.executable='echo' -Dexec.args='${project.artifactId}-${project.version}.jar' exec:exec -q > self-modules.txt
+./mvnw --batch-mode -Pbackend -Dbenchmark.skip=true -Dexec.executable='echo' -Dexec.args='${project.artifactId}-${project.version}.jar' exec:exec -q > self-modules.txt
```
kezhenxu94, line 25: We don't need this now, right?

oap-server/microbench/src/main/java/org/apache/skywalking/oap/server/microbench/library/datacarrier/SampleData.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public SampleData setIntValue(int intValue) {
+        this.intValue = intValue;
+        return this;
+    }
+
+    public SampleData setName(String name) {
+        this.name = name;
+        return this;
+    }
```
kezhenxu94, line 42: These methods seem to be useless, let's remove these or if they are needed, let's use lombok `@Getter` / `@Setter` annotation to eliminate the boilerplate codes

kezhenxu94 (commented):
Just some nits, others look good to me, thanks @CalvinKirs !!

tools/dependencies/check-LICENSE.sh
CalvinKirs: yup, thank you.

oap-server/microbench/src/main/java/org/apache/skywalking/oap/server/microbench/library/datacarrier/SampleData.java
CalvinKirs: Yes, I have deleted its get&set. In fact, its parameters are not used, let's keep its parameters for now.

CalvinKirs:
> Just some nits, others look good to me, thanks @CalvinKirs !!

thanks ：）

kezhenxu94 (dismissed):
LGTM

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/config/group/openapi/EndpointGroupingRule4Openapi.java
```diff
@@ -34,7 +34,7 @@ void addDirectLookup(String serviceName, String endpointName, String endpointGro
         endpointNameLookup.put(endpointName, endpointGroupName);
     }
 
-    void addGroupedRule(String serviceName, String endpointGroupName, String ruleRegex) {
+    public void addGroupedRule(String serviceName, String endpointGroupName, String ruleRegex) {
```
wu-sheng, line 37: Exposing new method is not recommended usually. I think this method is only used in benchmark initialization stage.

oap-server/server-receiver-plugin/skywalking-zabbix-receiver-plugin/pom.xml
```diff
@@ -35,14 +35,15 @@
...
             <version>${project.version}</version>
+            <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.openjdk.jmh</groupId>
-            <artifactId>jmh-generator-annprocess</artifactId>
-            <scope>test</scope>
+            <groupId>org.apache.skywalking</groupId>
+            <artifactId>meter-analyzer</artifactId>
+            <version>${project.version}</version>
```
wu-sheng, line 45: Why do we add `server-testing` when remove `jmh-generator-annprocess`?

wu-sheng:
I can't see document update in the https://github.com/apache/skywalking/blob/master/docs/en/guides/README.md#integration-tests. We should add `benchmark` section after `Integration Tests`

wu-sheng:
@CalvinKirs Any feedback?

CalvinKirs:
> @CalvinKirs Any feedback?

Sorry, I will update as soon as possible, thanks for your review

docs/en/guides/README.md
```diff
@@ -44,6 +44,21 @@ and if you would like to run all the ITs, simply run `./mvnw -Pall,CI-with-IT cl
 
 Please be advised that if you're writing integration tests, name it with the pattern `IT*` so they would only run with the `CI-with-IT` profile.
 
+### Java Microbenchmark Harness (JMH)
+JMH is a Java harness for building, running, and analysing nano/micro/milli/macro benchmarks written in Java and other languages targeting the JVM.
+
+We provide the `microbench` module, which provides `org.apache.skywalking.oap.server.microbench.base.AbstractMicrobenchmark`,
+
+all JMH tests need to extend this class to make it easier for you to complete JMHTest, you can also choose to
```
wu-sheng, line 52: What is this blank line for?

docs/en/guides/README.md
```diff
@@ -44,6 +44,21 @@ and if you would like to run all the ITs, simply run `./mvnw -Pall,CI-with-IT cl
...
+customize runtime conditions (Measurement, Fork, Warmup, etc.)
+
+You can run any of the JMH tests as a normal UT, or you can package it and get all the reported results via `java
+-jar benchmark.jar`, or get the results of a particular Test via `java -jar /benchmarks.jar exampleClassName`.
```
wu-sheng, line 56: I think `java -jar` is not enough. Because there are dependencies in other module(depended), we need to package a uber jar for this module, right?

docs/en/guides/README.md
```diff
@@ -44,6 +44,21 @@ and if you would like to run all the ITs, simply run `./mvnw -Pall,CI-with-IT cl
...
+
+if you want to visualize your results, you can choose '-jar benchmark.jar -rf json' to export the report results in json format, there are many websites that publicly provide report analysis, you can find it yourself.
+
```
wu-sheng, line 59: Another blank line.

docs/en/guides/README.md
```diff
@@ -44,6 +44,21 @@ and if you would like to run all the ITs, simply run `./mvnw -Pall,CI-with-IT cl
...
+We provide the `microbench` module, which provides `org.apache.skywalking.oap.server.microbench.base.AbstractMicrobenchmark`,
+
+all JMH tests need to extend this class to make it easier for you to complete JMHTest, you can also choose to
+customize runtime conditions (Measurement, Fork, Warmup, etc.)
+
+You can run any of the JMH tests as a normal UT, or you can package it and get all the reported results via `java
+-jar benchmark.jar`, or get the results of a particular Test via `java -jar /benchmarks.jar exampleClassName`.
+
+if you want to visualize your results, you can choose '-jar benchmark.jar -rf json' to export the report results in json format, there are many websites that publicly provide report analysis, you can find it yourself.
```
wu-sheng, line 58: Please make sure your sentences are complete, and the first letter of sentence should be upper case.

docs/en/guides/README.md
CalvinKirs: Yes, the benchmark is the final uber package.

docs/en/guides/README.md
CalvinKirs: Thank you for your comment, please take a look again

wu-sheng:
I re-write the contributing document for you, generally 2 things
1. Don't use `you could` or `you` repeatedly in the document. The document is more like an instruction, rather than command.
2. Upper case and some other grammar issues still exist, including very long sentences.

Is there anything we should add this PR? If not, please remove the `[WIP]` in the title and update changes.md in the root.

oap-server/microbench/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+                                        implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+                                    <mainClass>org.openjdk.jmh.Main</mainClass>
+                                </transformer>
+                            </transformers>
+                            <filters>
+                                <filter>
+                                    <artifact>*:*</artifact>
+                                    <excludes>
+                                        <exclude>**/Log4j2Plugins.dat</exclude>
+                                        <exclude>**/EndpointGrouping4OpenapiBenchmark.*</exclude>
```
kezhenxu94, line 105: Why exclude this?

oap-server/microbench/src/main/java/org/apache/skywalking/oap/server/microbench/base/AbstractMicrobenchmark.java
```diff
@@ -0,0 +1,116 @@
...
+        return optBuilder;
+    }
+
+    @Test
+    public void run() throws Exception {
+        new Runner(newOptionsBuilder().build()).run();
+    }
+
+    private static String getReportDir() {
+        return System.getProperty("perfReportDir");
```
kezhenxu94, line 113: This is not documented.

oap-server/microbench/pom.xml
CalvinKirs: Resource cannot be obtained when running in jar mode,
Therefore `ResourceUtils.class.getClassLoader().getResource(directoryPath)` will be empty. So I excluded this Test when packaging.

CalvinKirs:
> I re-write the contributing document for you, generally 2 things
> 
> 1. Don't use `you could` or `you` repeatedly in the document. The document is more like an instruction, rather than command.
> 2. Upper case and some other grammar issues still exist, including very long sentences.
> 
> Is there anything we should add this PR? If not, please remove the `[WIP]` in the title and update changes.md in the root.

Thank you very much

oap-server/microbench/src/main/java/org/apache/skywalking/oap/server/microbench/base/AbstractMicrobenchmark.java
CalvinKirs: Thank you for your reminder, I have explained it in the document, thank you again

oap-server/microbench/pom.xml
kezhenxu94: I meant this one `EndpointGrouping4OpenapiBenchmark`, why this benchmark is excluded? If I want to run this benchmark, it throws NCDF


```
java -jar oap-server/microbench/target/benchmarks.jar EndpointGrouping4OpenapiBenchmark                 
# JMH version: 1.25
# VM version: JDK 1.8.0_292, OpenJDK 64-Bit Server VM, 25.292-b10
# VM invoker: /Users/kezhenxu94/.asdf/installs/java/adoptopenjdk-8.0.292+10/jre/bin/java
# VM options: <none>
# Warmup: 10 iterations, 10 s each
# Measurement: 10 iterations, 10 s each
# Threads: 4 threads, will synchronize iterations
# Benchmark: org.apache.skywalking.oap.server.microbench.core.config.group.openapi.EndpointGrouping4OpenapiBenchmark.formatEndpointNameMatchedPaths20

# Run progress: 0.00% complete, ETA 00:20:00
# Fork: 1 of 2
<failure>

java.lang.NoClassDefFoundError: org/apache/skywalking/oap/server/microbench/core/config/group/openapi/EndpointGrouping4OpenapiBenchmark
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:756)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:468)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
	at java.security.AccessController.doPrivileged(Native Method)
```

oap-server/microbench/src/main/java/org/apache/skywalking/oap/server/microbench/base/AbstractMicrobenchmark.java
kezhenxu94: Sorry to find that this is a built-in feature (`-rff`) of JMH so we'd better not to replicate the feature.

```
java -jar oap-server/microbench/target/benchmarks.jar -rf json -rff /tmp/result.json AtomicRangeIntegerBenchmark
```

CalvinKirs: This parameter needs to be added if the user uses IDE to run it.

oap-server/microbench/pom.xml
CalvinKirs: hi, thanks for your review, It needs to read the resource because it needs it, but it cannot get the resource when running through the jar. That is what I mentioned above.

wu-sheng: Files could be loaded through classpath or in the uber jar. You shouldn't exclude meaningful things out of packages.

What is the issue exactly here?

kezhenxu94: The problem is that `EndpointGrouping4OpenapiBenchmark` read files from class path but benchmark tests generate those tests outside of the class path, I'm helping @CalvinKirs to resolve this and fix some other problems in that benchmark

kezhenxu94 (approved):
LGTM

wu-sheng (approved):

CalvinKirs:
Thanks ~

Issue #7988, opened 2021-10-22
[Test] Cluster e2e need verify OAP cluster model data aggregation 


wankai123:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Cluster e2e just verified the integration with coordinators and storage for now. We need to verify the real OAP cluster model data aggregation.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #7989, opened 2021-10-22
e2e-v2: Simplified cluster e2e to reduce failure

wankai123:

Cluster e2e failing frequently, due to the sometimes consumer request can't be dispatched to both provider1 and provider2.
So I removed provider2 since we don't verify the OAP cluster model L2 aggregation for now. 
see: #7988

### DIFF
--- a/test/e2e-v2/cases/cluster/zk/es/docker-compose.yml
+++ b/test/e2e-v2/cases/cluster/zk/es/docker-compose.yml
@@ -93,7 +93,7 @@ services:
       SW_AGENT_COLLECTOR_BACKEND_SERVICES: oap2:11800
       SW_AGENT_INSTANCE_NAME: provider1
     depends_on:
-      oap1:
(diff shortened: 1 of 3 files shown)


kezhenxu94:
Removing provider2 makes it pointless to have this "cluster" e2e.

wu-sheng:
> Removing provider2 makes it pointless to have this "cluster" e2e.

It still works as same as previous, once consumer to oap1 and provider to oap2. As that issue mentioned, we never checked whether L2 aggregation works, only data accurate testing could verify that. 2 providers have no difference with one provider from cluster perspective.

kezhenxu94:
If you record https://github.com/apache/skywalking/issues/7988 and think we should verify L2 aggregation. I'd suggest you try to fix 👇

> sometimes consumer request can't be dispatched to both provider1 and provider2.

That is still an issue when you deal with #7988

PR #8054, opened 2021-11-02
[CI] Update PHP Agent

heyanlong:
- [x] Update the php-agent version of CI to 4.2.0


wu-sheng (approved):
The new PHP e2e passed

PR #8066, opened 2021-11-03
Optimize metrics of minute dimensionality persistence, don't save metric of being default value.

wu-sheng:

Optimize metrics of minute dimensionality persistence.
- The value of metrics, which has declaration of the default value and current value equals the default value logically, the whole row wouldn't be pushed into database.
- Add self-observability `metrics_persistence_skipped` metric to measure the number of skipped metrics.
- SLA(percent function), Apdex, AvgLong(MAL mostly) metrics are mostly optimized because of this. Such as a full-failing error(0% successful rate) and unhealthy(apdex=0) are not costing the storage.
- Hour/Day dimensionalities metrics are not changed, as they cover long duration.
- Fix `max` function in OAL doesn't support negative long.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/HavingDefaultValue.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/HavingDefaultValue.java
@@ -1,0 +1,42 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.oap.server.core.analysis.metrics;
+
+/**
+ * HavingDefaultValue interface defines a capability for the metric implementations, which has the declaration of the
+ * default value.
+ *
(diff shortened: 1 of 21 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/HavingDefaultValue.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * For the declared metrics, the OAP server would skip the persistence of minute dimensionality metrics to reduce
+ * resource costs for the database, when the value of the metric is the default.
+ */
+public interface HavingDefaultValue {
+    /**
+     * @return true if the implementation of this metric has the definition of default value.
+     */
+    default boolean haveDefault() {
+        return false;
```
hanahmily, line 33: It seems all implements return `true`. What's the scenario this has to return `false`?

wu-sheng: Yes

wu-sheng: Many others are not having default value, such as heatmap, labelled value(percentile).

wu-sheng: We have a little chance to optimize the `PercentileFunction` later, but for most labeled value functions, we have to keep them as we need the labels even values are 0.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/MaxLongMetrics.java
```diff
@@ -25,9 +25,6 @@
 import org.apache.skywalking.oap.server.core.analysis.metrics.annotation.SourceFrom;
 import org.apache.skywalking.oap.server.core.storage.annotation.Column;
 
-/**
- *
- **/
 @MetricsFunction(functionName = "max")
 public abstract class MaxLongMetrics extends Metrics implements LongValueHolder {
```
wankai123, line 30: No needs to implement the interface here?

wu-sheng: Yes, I just added and fixed another bug in this class.

wankai123 (approved):
LGTM

PR #8122, opened 2021-11-13
Upgrade Kubernetes Java client to 14.0.0, supports GCP token refreshing and fixes some bugs

kezhenxu94:
The interesting bugfix is https://github.com/kubernetes-client/java/issues/1829 and feature is https://github.com/kubernetes-client/java/pull/1810

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. NO.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -47,6 +47,7 @@ Release Notes.
   value and current value equals the default value logically, the whole row wouldn't be pushed into database.
 * Fix `max` function in OAL doesn't support negative long.
 * Add `MicroBench` module to make it easier for developers to write JMH test.
+* Upgrade Kubernetes Java client to 14.0.0, supports GCP token refreshing and fixes some bugs.
 
 #### UI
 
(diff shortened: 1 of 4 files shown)


wu-sheng (commented):
It seems having more dependencies change.

wu-sheng (approved):

chanjarster:
@kezhenxu94 is it possible migrate this fix to branch 8.8.x so when v8.8.2 released will have this fix too.

wu-sheng:
Typically, I don't think there will be 8.8.2.
8.9.0 should be released soon though.

kezhenxu94:
> @kezhenxu94 is it possible migrate this fix to branch 8.8.x so when v8.8.2 released will have this fix too.

We generally won't back port things to previous versions. Our next release is 8.9.0 not 8.8.2, the reason why we have 8.8.1 is that 8.8.0 didn't work, so as long as 8.8.1 is generally working we won't release 8.8.2 just for fixes, that's not our plan

chanjarster:
Got it

PR #8015, opened 2021-10-27
try to enable cross build for arm64 docker image

mis98zb:
### <Feature description>
- [ ] If this is non-trivial feature, paste the links/URLs to the design doc.
- [ ] Update the documentation to include this new feature.
- [ ] Tests(including UT, IT, E2E) are added to verify the new feature.
- [ ] If it's UI related, attach the screenshots below.

- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>.
- [ ] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md).

PR #8141, opened 2021-11-17
Add Docker images for arm64 architecture

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/pull/8015

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,15 +43,11 @@ OAP_NAME ?= oap
 UI_NAME ?= ui
 TAG ?= latest
 
-.SECONDEXPANSION: #allow $@ to be used in dependency list
-
-.PHONY: docker docker.all docker.oap
+.PHONY: docker docker.all
 
 docker: init build.all docker.all
 
 DOCKER_TARGETS:=docker.oap docker.ui
-
-docker.all: $(DOCKER_TARGETS)
 
 ifneq ($(SW_OAP_BASE_IMAGE),)
   BUILD_ARGS := $(BUILD_ARGS) --build-arg BASE_IMAGE=$(SW_OAP_BASE_IMAGE)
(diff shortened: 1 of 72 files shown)


wu-sheng:
Which platforms are we going to push to docker and ghcr eventually?

kezhenxu94:
> Which platforms are we going to push to docker and ghcr eventually?

linux/amd64 and linux/arm64

Makefile
```diff
@@ -43,64 +43,47 @@ OAP_NAME ?= oap
...
-docker.push: $(DOCKER_PUSH_TARGETS)
-
-
+# 1. Make a directory $(DOCKER_BUILD_TOP)/$(NAME)
+# 2. This rule uses cp to copy all dependency filenames into into $(DOCKER_BUILD_TOP/$(NAME)
+# 3. This rule finally runs docker build passing $(BUILD_ARGS) to docker if they are specified as a dependency variable
+
+define DOCKER_RULE
+	mkdir -p $(DOCKER_BUILD_TOP)/$(NAME)
+	cp -r $^ $(DOCKER_BUILD_TOP)/$(NAME)
```
mis98zb, line 81: makefile variables like $^ cannot be used inside define.

as my experience, the variable will be calculated before it is put to the place where it is called.
so, $^ cannot get the correct dependency list.
that's why I used parameter for the defined function...

Makefile
```diff
@@ -43,64 +43,47 @@ OAP_NAME ?= oap
...
+%.oap: NAME = $(OAP_NAME)
 
-docker.ui: $(CONTEXT)/$(DIST)
-docker.ui: $(SW_ROOT)/docker/ui/Dockerfile.ui
-docker.ui: $(SW_ROOT)/docker/ui/docker-entrypoint.sh
-docker.ui: $(SW_ROOT)/docker/ui/logback.xml
-docker.ui: NAME = $(UI_NAME)
-docker.ui:
+docker.%: PLATFORMS =
+docker.%: LOAD_OR_PUSH = --load
```
mis98zb, line 62: I remember I used to use --load on my pc and got warning "not support yet". Because local docker image cannot support multiple platform with same tag.

Not sure if it works now.

kezhenxu94: > I remember I used to use --load on my pc and got warning "not support yet". Because local docker image cannot support multiple platform with same tag.
> 
> Not sure if it works now.

`--load` is only to build an image just for local machine's use so there is no cross platforms in this mode, take a look at `PLATFORMS = ` (empty)

mis98zb: got it, then it is not a problem here.

wu-sheng:
> > Which platforms are we going to push to docker and ghcr eventually?
> 
> linux/amd64 and linux/arm64

Is this arm64 able to build on the top of a x86 Linux or Mac? (I never had a change to learn this, sorry)

kezhenxu94:
> > > Which platforms are we going to push to docker and ghcr eventually?
> > 
> > 
> > linux/amd64 and linux/arm64
> 
> Is this arm64 able to build on the top of a x86 Linux or Mac? (I never had a change to learn this, sorry)

Yes, that's why we need to leverage BuildKit to build cross platform images

Makefile
mis98zb: build passed... it shocks me much.../(ㄒoㄒ)/~~

I find you use $(DOCKER_RULE) instead of $(call DOCKER_RULE), that may be the reason.
it works, that's fine.

kezhenxu94:
I end up changing the base image from `adoptopenjdk/openjdk:alpine-jre` to `adoptopenjdk/openjdk:jre`, which is closest to the previous one and all tests are passing, also no {high,critical} CVEs.

```
trivy --severity "HIGH,CRITICAL" adoptopenjdk/openjdk11:jre
2021-11-17T19:13:04.132+0800	INFO	Detected OS: ubuntu
2021-11-17T19:13:04.132+0800	INFO	Detecting Ubuntu vulnerabilities...
2021-11-17T19:13:04.134+0800	INFO	Number of PL dependency files: 1

adoptopenjdk/openjdk11:jre (ubuntu 20.04)
=========================================
Total: 0 (HIGH: 0, CRITICAL: 0)
```

wu-sheng:
Once it is JRE, I am fine. Just curious, what are the differences from `openjdk:alpine-jre`.

wu-sheng (approved):

kezhenxu94:
> Once it is JRE, I am fine. Just curious, what are the differences from `openjdk:alpine-jre`.

We previously use `adoptopenjdk/openjdk:alpine-jre`, which only has a `amd64` manifest in Docker image, so when building for `arm64` it cannot find the base image (for `arm64`), the difference is that `adoptopenjdk/openjdk:jre` provides both `amd64` and `arm64` manifest in Docker images so we can pull them as base image to build for both architectures.

wu-sheng:
Got this. My question is more about, what are differences of image size and components built-in inside.

wu-sheng (approved):

kezhenxu94:
Check these 2 images

Screen Shot 2021-11-17 at 20 49 38
>

PR #8142, opened 2021-11-17
Set up QEMU and Buildx in GHA to support cross platform

kezhenxu94:
verified in https://github.com/apache/skywalking/runs/4239030853?check_suite_focus=true

### DIFF
--- a/.github/workflows/publish-docker.yaml
+++ b/.github/workflows/publish-docker.yaml
@@ -52,5 +52,9 @@ jobs:
           registry: ${{ env.HUB }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
       - name: Build and push docker images
         run: make build.all docker.push


.github/workflows/publish-docker.yaml
```diff
@@ -52,5 +52,9 @@ jobs:
           registry: ${{ env.HUB }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set up QEMU
```
wu-sheng, line 55: QEMU, interesting. Anything needs to be added for release document when do container image? I feel there are some extra setups.

kezhenxu94: > QEMU, interesting. Anything needs to be added for release document when do container image? I feel there are some extra setups.

Screen Shot 2021-11-17 at 22 37 14

For maintainers using Mac OS, a Docker Desktop is just enough to release the images, as tested in https://hub.docker.com/repository/docker/kezhenxu94/oap/tags?page=1&ordering=last_updated

wu-sheng: How about Linux(Ubuntu for example)? I was building all release(binary, source tar and image) on GCP Linux

wu-sheng (approved):

Issue #8145, opened 2021-11-18
[Feature] [MAL Core] Add a match pattern mechanism for the MAL core


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

I noticed in the showcase(also demo env), we always face `oap::` services, which mean no actual service name.
I finally got a chance to confirm with @kezhenxu94 why. It turns out `oap`'s self-observability and `vm`'s monitoring have the same metric name.

We need a mechanism to identify the source of a SampleFamily to avoid this case. I would like to propose these kind of mechanism.

We add a field precondition field into all MAL related configuration files, including `oc-receiver`, `envoy-metrics-receiver`, `meter-receiver` etc. This field should be a closure returning true/false. When `false` is returned, this SampleFamily set isn't pushed into rules of all including expressions, but simply skip.

@hanahmily Could you confirm this mechanism makes sense?

### Use case

This feature could avoid empty service issue, and generally it increase the performance of MAL engine, because less calculation are applied by the input metrics set.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wu-sheng:
FYI @wankai123 due to you asked about this in afternoon.

wankai123:
We can get otel `job_name` from `node.getServiceInfo()`, then could add a tag like `job_name=skywalking` to all metrics.
I think we can use this to do the filter.

wu-sheng:
Even we use a tag filter, we need the closure setup to filter irrelative metrics out.

wankai123:
yes, the metrics should filter before into MAL rules

hanahmily:
MAL tends to generate the metrics that are mandatory. If the metrics are invalid, related MALs are redundant, having to be removed from the configuration.

What leads to such issues? Can we avoid them from the source? If their service tags are absent, could we have a chance to provide them? If not, I prefer to drop them.

wu-sheng:
No, they are not redundant. 2 kinds of agent/source(s) are generating the metrics with same name.

**oap's self-observability and vm's monitoring have the same metric name**. More importantly, this could happen again in other monitoring cases.

wu-sheng:
[...]

PR #8157, opened 2021-11-19
Add filter mechanism in MAL core and fix some bugs

kezhenxu94:
Also fixed some minor bugs

* Fix concurrency bug in MAL `increase`-related calculation.
* Fix a null pointer bug when building `SampleFamily`.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/8145.

### DIFF
--- a/docs/en/setup/backend/backend-telemetry.md
+++ b/docs/en/setup/backend/backend-telemetry.md
@@ -128,7 +128,7 @@ Set this up following these steps:
   ```
 2. Set up OpenTelemetry Collector and config a scrape job:
 ``` yaml
-- job_name: 'skywalking'
+- job_name: 'skywalking-so11y' # make sure to use this in the so11y.yaml to filter only so11y metrics
   metrics_path: '/metrics'
   kubernetes_sd_configs:
   - role: pod
--- a/oap-server/server-starter/src/main/resources/otel-oc-rules/istio-controlplane.yaml
+++ b/oap-server/server-starter/src/main/resources/otel-oc-rules/istio-controlplane.yaml
@@ -28,6 +28,7 @@
 #    "-P6H3M"    -- parses as "-6 hours and -3 minutes"
 #    "-P-6H+3M"  -- parses as "+6 hours and -3 minutes"
 # </pre>
+filter: "{ tags -> tags.job_name == 'kubernetes-pods' }" # The OpenTelemetry job name
 expSuffix: tag({tags -> tags.cluster = 'istio-ctrl::' + tags.cluster}).service(['cluster', 'app'])
 metricPrefix: meter_istio
 metricsRules:
--- a/oap-server/server-starter/src/main/resources/otel-oc-rules/k8s-cluster.yaml
+++ b/oap-server/server-starter/src/main/resources/otel-oc-rules/k8s-cluster.yaml
@@ -28,11 +28,10 @@
 #    "-P6H3M"    -- parses as "-6 hours and -3 minutes"
 #    "-P-6H+3M"  -- parses as "+6 hours and -3 minutes"
 # </pre>
+filter: "{ tags -> tags.job_name in [ 'kubernetes-cadvisor', 'kube-state-metrics' ] }" # The OpenTelemetry job name
 expSuffix: tag({tags -> tags.cluster = 'k8s-cluster::' + tags.cluster}).service(['cluster'])
 metricPrefix: k8s_cluster
 metricsRules:
-
-
   - name: cpu_cores
     exp: (kube_node_status_capacity * 1000).tagEqual('resource' , 'cpu').sum(['cluster'])
   - name: cpu_cores_allocatable
--- a/oap-server/server-starter/src/main/resources/otel-oc-rules/vm.yaml
+++ b/oap-server/server-starter/src/main/resources/otel-oc-rules/vm.yaml
@@ -28,6 +28,7 @@
 #    "-P6H3M"    -- parses as "-6 hours and -3 minutes"
 #    "-P-6H+3M"  -- parses as "+6 hours and -3 minutes"
 # </pre>
+filter: "{ tags -> tags.job_name == 'vm-monitoring' }" # The OpenTelemetry job name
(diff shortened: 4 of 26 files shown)


oap-server/server-starter/src/main/resources/otel-oc-rules/vm.yaml
```diff
@@ -28,6 +28,7 @@
 #    "-P6H3M"    -- parses as "-6 hours and -3 minutes"
 #    "-P-6H+3M"  -- parses as "+6 hours and -3 minutes"
 # </pre>
+filter: "{ tags -> tags.job_name == 'vm-monitoring' }"
```
wankai123, line 31: I found `oap.yaml` has the same metrics with `istio-controlplane.yaml`, not `vm`
```
process_cpu_seconds_total
```

wankai123 (commented):
MAL not only supports `OCReceiver`, should we add tag to other receivers too?
Like: meter-receiver.

docs/en/setup/backend/backend-telemetry.md
```diff
@@ -128,7 +128,7 @@ Set this up following these steps:
   ```
 2. Set up OpenTelemetry Collector and config a scrape job:
 ``` yaml
-- job_name: 'skywalking'
+- job_name: 'skywalking-so11y' # make sure to use this in the so11y.yaml to filter only so11y metrics
```
wankai123, line 131: the example file need sync too :)
https://github.com/apache/skywalking/blob/master/docs/en/setup/backend/otel-collector-oap.yaml#L33

kezhenxu94: Updated

oap-server/server-starter/src/main/resources/otel-oc-rules/vm.yaml
kezhenxu94: > I found `oap.yaml` has the same metrics with `istio-controlplane.yaml`, not `vm`
> 
> ```
> process_cpu_seconds_total
> ```

There are metrics name conflicts in `vm` and `so11y` for sure, you cannot only compare our configuration files, the key is what metrics actually OpenTelemetry sends. FYI the conflict between `so11y` and `vm` is `instance_cpu_percentage` from what I can tell.

oap-server/server-starter/src/main/resources/otel-oc-rules/k8s-cluster.yaml
```diff
@@ -31,8 +31,6 @@
 expSuffix: tag({tags -> tags.cluster = 'k8s-cluster::' + tags.cluster}).service(['cluster'])
```
wu-sheng, line 31: Should we add a filter for k8s too?

wu-sheng:
`mal.md` should be updated about filter expression.

oap-server/server-starter/src/main/resources/otel-oc-rules/k8s-cluster.yaml
kezhenxu94: > Should we add a filter for k8s too?

I'm not sure... on the one hand, adding a default filter requires the users to give the same job name when they set up OpenTelemetry job, otherwise there is no data and confused for users, on the other hand, add a filter is absolutely right to narrow down the metrics that need to be processed in this file

oap-server/server-starter/src/main/resources/otel-oc-rules/vm.yaml
wu-sheng: I think we should recommend(if not require) to add filter for all oc-receiver MAL. In the future, OpenTelemetry native metric format(just GA) and meter receiver should recommend this setup too.

oap-server/server-starter/src/main/resources/otel-oc-rules/istio-controlplane.yaml
```diff
@@ -28,6 +28,7 @@
 #    "-P6H3M"    -- parses as "-6 hours and -3 minutes"
 #    "-P-6H+3M"  -- parses as "+6 hours and -3 minutes"
 # </pre>
+filter: "{ tags -> tags.job_name in ['kubernetes-pods', 'kubernetes-cadvisor', 'kube-state-metrics' ] }" # The OpenTelemetry job name
```
wankai123, line 31: i think isito control-plane only required `kubernetes-pods` @hanahmily

oap-server/server-starter/src/main/resources/otel-oc-rules/k8s-cluster.yaml
wu-sheng: Isn't k8s monitoring using OTEL job(s) to forward data?

kezhenxu94: > we can add `kubernetes-cadvisor` and `kube-state-metrics` cause example used these name
> https://github.com/apache/skywalking/blob/master/docs/en/setup/backend/otel-collector-config.yaml#L33

If I have to add a filter, the names we use in the example are first choice for sure, my question is, should we add those if there is no obvious conflicts at the moment?

kezhenxu94:
filter is added to all MAL files now

oap-server/server-starter/src/main/resources/otel-oc-rules/k8s-cluster.yaml
wu-sheng: I would recommend yes. Such as using a `Istio*` to filter all Istio MAL scripts, makes sense always.

kezhenxu94: > I would recommend yes. Such as using a `Istio*` to filter all Istio MAL scripts, makes sense always.

Not sure what you mean. The names of jobs for istio doesn’t necessarily always include `istio*`, the job name is totally human written, it can be arbitrary

wu-sheng (approved):

wankai123 (approved):
LGTM

PR #8161, opened 2021-11-20
Fix the so11y latency of persistence execution latency not correct in ElasticSearch storage

kezhenxu94:
The ES persistence execution is now asynchronous and the execution
latency only counts the time to insert the requests into the bulk
processor, instead of the time after the requests are flushed into the
storage, this patch fixes that issue.

There is the same issue in prepare latency but that needs more changes
so I'll leave it to another pull request.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -54,6 +54,7 @@ Release Notes.
 * Add filter mechanism in MAL core to filter metrics.
 * Fix concurrency bug in MAL `increase`-related calculation.
 * Fix a null pointer bug when building `SampleFamily`.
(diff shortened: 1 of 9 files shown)


kezhenxu94:
before
Screen Shot 2021-11-20 at 15 57 35
 
after
Screen Shot 2021-11-20 at 15 57 56

wu-sheng (approved):
LGTM @LIU-WEI-git

kezhenxu94:
Push another commit to fix the unit test

wu-sheng:
Yes, noticed too.

kezhenxu94:
Should be good now

wu-sheng (approved):

PR #8162, opened 2021-11-21
Remove filter for Istiod monitor to reduce the prerequisites of job name matching

kezhenxu94:
I found out the reason why there is no conflict between Istio and so11y in https://github.com/apache/skywalking/pull/8157#discussion_r753153239 is the `istio-controlplane.yaml` has already filter the metrics by tag `app == istiod`, and since we also have filter for so11y since https://github.com/apache/skywalking/pull/8157, I think we can safely remove the filter in `istio-controlplane.yaml` so that the users don't bother to check the job name in OTEL and the filter name in our file `istio-controlplane.yaml`, which reduce extra work and confusion if users found there is no data on UI, just because of the names' not matching.


wu-sheng (approved):
Ha, I think there two general cases are both working.

PR #8177, opened 2021-11-24
Fix incorrect use of ReentrantLock in GRPCExporter

xuxiawei:
fix that an imse (illegalmonitorstateexception) exception will be thrown when finally releasing the lock when an exception occurs in lock() locking


wu-sheng:
Please take a look at `ReentrantLock` comments in JDK.

```java
 * <p>It is recommended practice to <em>always</em> immediately
 * follow a call to {@code lock} with a {@code try} block, most
 * typically in a before/after construction such as:
 *
 * <pre> {@code
 * class X {
 *   private final ReentrantLock lock = new ReentrantLock();
 *   // ...
 *
 *   public void m() {
 *     lock.lock();  // block until condition holds
 *     try {
 *       // ... method body
 *     } finally {
 *       lock.unlock()
 *     }
 *   }
 * }}</pre>
```

wu-sheng:
Reject this according to JDK recommendation, unless there are other scenarios not covered by JDK comments.

kezhenxu94:
@wu-sheng the patch in this PR exactly changes the codes to what recommended in JDK comment, why close this?

wu-sheng:
Sorry my fault. The cell phone visualization mislead me.

wu-sheng (approved):

PR #8179, opened 2021-11-24
[Feature] Support skip TLS host verify on ElasticSearch

chanjarster:
### <Feature description>

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #8154.

### DIFF
--- a/oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
+++ b/oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
@@ -68,7 +68,7 @@ public class ElasticSearchClient implements Client, HealthCheckable {
     private volatile String trustStorePass;
 
     @Setter
-    private boolean skipHostVerify;
+    private volatile boolean skipHostVerify;
 
     @Setter
     private volatile String user;
--- a/oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
+++ b/oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
@@ -68,6 +68,9 @@ public class ElasticSearchClient implements Client, HealthCheckable {
     private volatile String trustStorePass;
 
     @Setter
+    private boolean skipHostVerify;
+
+    @Setter
     private volatile String user;
 
     @Setter
@@ -133,6 +136,7 @@ public class ElasticSearchClient implements Client, HealthCheckable {
         if (!Strings.isNullOrEmpty(trustStorePass)) {
             cb.trustStorePass(trustStorePass);
         }
+        cb.skipHostVerify(skipHostVerify);
         if (!Strings.isNullOrEmpty(user)) {
             cb.username(user);
         }
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java
@@ -103,7 +103,7 @@ public class StorageModuleElasticsearchConfig extends ModuleConfig {
     private String trustStorePass;
 
     /**
-     * @since 8.8.2 Skip tls host verify on elasticsearch server
+     * @since 8.9.0 Skip tls host verify on ElasticSearch server host
      */
     private boolean skipHostVerify;
 
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java
@@ -101,6 +101,12 @@ public class StorageModuleElasticsearchConfig extends ModuleConfig {
      * @since 7.0.0 This could be managed inside {@link #secretsManagementFile}
      */
     private String trustStorePass;
+
+    /**
+     * @since 8.8.2 Skip tls host verify on elasticsearch server
+     */
+    private boolean skipHostVerify;
+
     private int resultWindowMaxSize = 10000;
     private int metadataQueryMaxSize = 5000;
     private int segmentQueryMaxSize = 200;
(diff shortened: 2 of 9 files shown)


wu-sheng:
OK, now I can understand your proposal. But still not sure why. This kind of TLS ignore seems a security leak.

cc @apache/skywalking-committers

chanjarster:
Sure, it's an insecure way. But skip TLS host verify is always an option for TLS client, Netty, HttpClient, [OpenTelemetry exporters][1] all support it.

In consider the secure issue, so I make this option `false` by default. One should turn it on explicitly if really want to.

[1]: https://opentelemetry.io/docs/collector/configuration/

wu-sheng:
Personally I don't like it, no matter default on or off. If you want to skip the security, HTTP proxy is there for you.(You could set up HTTP proxy between SkyWalking OAP and ElasticSearch server).

A security check is open at the server side, but client skips it because of?

wu-sheng:
I am going to wait for @kezhenxu94 @hanahmily @dmsolr @JaredTan95 or any other committers' idea.

wu-sheng:
> But skip TLS host verify is always an option for TLS client, Netty, HttpClient, OpenTelemetry exporters all support it.

I can represent other open source projects' authors. I believe they have their reasons. I just don't feel right from SkyWalking's perspective. Anyway, let's wait for others' idea.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java
```diff
@@ -101,6 +101,12 @@
      * @since 7.0.0 This could be managed inside {@link #secretsManagementFile}
      */
     private String trustStorePass;
+
+    /**
+     * @since 8.8.2 Skip tls host verify on elasticsearch server
```
kezhenxu94, line 106:
```suggestion
     * @since 8.9.0 Skip tls host verify on ElasticSearch server host
```

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -67,6 +67,9 @@
     @Setter
     private volatile String trustStorePass;
 
+    @Setter
+    private boolean skipHostVerify;
```
kezhenxu94, line 71:
```suggestion
    private volatile boolean skipHostVerify;
```

kezhenxu94 (commented):
It’s ok for me to add the new config option

wu-sheng:
I want to discuss is, why need this configuration. Could you explain the use case?

chanjarster:
I explained before. Below is the use case:

Deploy SkyWalking in K8S by using [chart][1], using an existing ElasticSearch instance which is https enabled and using self signed key.

So how can you give SkyWalking a trust store which containing the ElasticSearch's CA by using environment?

3 Options:

1. Easy way: just tell SkyWalking to skip TLS host verify.
2. Middle way: give SkyWalking a PEM encoded X.509 certificate to trust by using an env variable.
3. Hardest way, and also SkyWalking currently support: make a trust store with the CA certificate, put that trust store into SkyWalking OAP server container. Apparently this way is not suitable for K8S deploy. 

So I implement Option 1. 

Option 2 is good too, I'll implement in another PR.

[1]: https://github.com/apache/skywalking-kubernetes/blob/master/chart/skywalking/README.md

wu-sheng:
Considering you don't provide any case after you submitted the PR, even I asked for several times. 

I have a talk with @hanahmily. 
The only case I got is you are trying to access ES server through unofficial domain name.
But this seems insecurity way, which violates the original purpose of TLS. 

We are a deliverable package, rather than a library. Once we have this, this could be reported as a CVE, even as optional, because it doesn't fit the TLS purpose.

So, I think we need to reject this feature. If you have more cases to provide, we could continue from there.

wu-sheng:
> I explained before. Below is the use case:
> 
> Deploy SkyWalking in K8S by using [chart](https://github.com/apache/skywalking-kubernetes/blob/master/chart/skywalking/README.md), using an existing ElasticSearch instance which is https enabled and using self signed key.
> 
> So how can you give SkyWalking a trust store which containing the ElasticSearch's CA by using environment?
> 
> 3 Options:
> 
> 1. Easy way: just tell SkyWalking to skip TLS host verify.
> 2. Middle way: give SkyWalking a PEM encoded X.509 certificate to trust by using an env variable.
> 3. Hardest way, and also SkyWalking currently support: make a trust store with the CA certificate, put that trust store into SkyWalking OAP server container. Apparently this way is not suitable for K8S deploy.
> 
> So I implement Option 1.
> 
> Option 2 is good too, I'll implement in another PR.

This is not accurate, I am afraid. There are plenty of sidecar tech could inject files into docker inside, such as Vault(just an example). I can't see why this is hard. But you should know, this is correct and real TLS should be done.

chanjarster:
What happened? The use case provided:

> Deploy SkyWalking in K8S by using chart, using an existing ElasticSearch instance which is https enabled and using self signed key.

In this case SkyWalking OAP server CANNOT connect to ElasticSearch and failed to start.

chanjarster:
> The only case I got is you are trying to access ES server through unofficial domain name.

Why I need a OFFICIAL domain name for a ElasticSearch for internal usage?

> There are plenty of sidecar tech could inject files into docker inside, such as Vault(just an example). I can't see why this is hard.

Well, I won't argue about it. 

So, please look at Option 2, is that easier than sidecar injection?

> give SkyWalking a PEM encoded X.509 certificate to trust by using an env variable.

wu-sheng:
> Why I need a OFFICIAL domain name for a ElasticSearch for internal usage?

This was what we are guessing why you need this. You could skip this. We didn't know you think accessing inside k8s is a challenge, all our deployments are in k8s, and nothing blocks us.

wu-sheng:
> So, please look at Option 2, is that easier than sidecar injection?
give SkyWalking a PEM encoded X.509 certificate to trust by using an env variable.

Do you mean put the content of certification as the value of system env?

chanjarster:
> We didn't know you think accessing inside k8s is a challenge, all our deployments are in k8s, and nothing blocks us.

Where is the doc about the situation I've mentioned? I struggled on this for several days.

> Do you mean put the content of certification as the value of system env?

Yes. And let me correct it: put the base64 encoded content of the cert as the value of system env.

wu-sheng:
>> We didn't know you think accessing inside k8s is a challenge, all our deployments are in k8s, and nothing blocks us.
Where is the doc about the situation I've mentioned? I struggled on this for several days.

There isn't. This is not SkyWalking's scope of guiding how to use K8s to manage security. Sorry. Maybe some blogs could be written.

> Yes. And let me correct it: put the base64 encoded content of the cert as the value of system env.

Even with sidecar injection, do you still want this? From my side, this is not a cloud native style solution. I may need to think more about whether this is a security leak. Besides that, I would like to recommend you using sidecar injection. It is really good.

chanjarster:
> Maybe some blogs could be written.
> I would like to recommend you using sidecar injection. It is really good.

I'm looking forward to it.

tirelibirefe:
Hello,
Is there a way using ElasticSearch as backend if a self-signed certificate is used with ES or may we say "Skywalking doesn't support ElasticSearch with self-signed certificates?"

(I don't know there is anyone using ElasticSearch with ssl gotten from a public cert authority in Kubernetes.)

Thanks

PR #8183, opened 2021-11-25
Add OpenSearch 1.2.0 to test and verify it works

kezhenxu94:

### DIFF
--- a/.github/workflows/e2e.storages.yaml
+++ b/.github/workflows/e2e.storages.yaml
@@ -38,12 +38,16 @@ jobs:
       matrix:
         config-file:
           - storage/h2/e2e.yaml
-          - storage/opensearch/e2e.yaml
           - storage/mysql/e2e.yaml
           - storage/tidb/e2e.yaml
           - storage/influxdb/e2e.yaml
           - storage/postgres/e2e.yaml
         include:
+          - opensearch-version: 1.1.0
+            config-file: storage/opensearch/e2e.yaml
+          - opensearch-version: 1.2.0
+            config-file: storage/opensearch/e2e.yaml
+
           - es-version: 6.3.2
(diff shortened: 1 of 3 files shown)


dmsolr (dismissed):

.github/workflows/e2e.storages.yaml
```diff
@@ -38,12 +38,16 @@ jobs:
...
-          - storage/opensearch/e2e.yaml
           - storage/mysql/e2e.yaml
           - storage/tidb/e2e.yaml
           - storage/influxdb/e2e.yaml
           - storage/postgres/e2e.yaml
         include:
+          - os-version: 1.1.0
+            config-file: storage/opensearch/e2e.yaml
+          - os-version: 1.2.0
+            config-file: storage/opensearch/e2e.yaml
```
wu-sheng, line 49: We need a policy for OpenSearch one day, we never tested every version of ElasticSearch.

.github/workflows/e2e.storages.yaml
```diff
@@ -38,12 +38,16 @@ jobs:
       matrix:
         config-file:
           - storage/h2/e2e.yaml
-          - storage/opensearch/e2e.yaml
           - storage/mysql/e2e.yaml
           - storage/tidb/e2e.yaml
           - storage/influxdb/e2e.yaml
           - storage/postgres/e2e.yaml
         include:
+          - os-version: 1.1.0
```
wu-sheng, line 46: os as shortcut of OpenSearch seems a little tricky, os usually is for operation system.

wu-sheng: let's keep opensearch as full word?

wu-sheng (approved):

PR #8186, opened 2021-11-26
[Enhancement] Bump grpc and protobuff version to allow building on Mac M1

### LINKED ISSUE #8184: [Enhancement] Bump grpc and protobuff version to allow building on Mac M1 osx-aarch_64
### Search before asking

### What happened

When you try build skywaling on a Mac M1 some artifacts are missing for the osx-aarch_64 architecture as described in https://github.com/apache/skywalking/blob/master/docs/en/FAQ/How-to-build-with-mac-m1.md#problem.

Build skywalking without using `-Dos.detected.classifier=osx-x86_64`  and instead use the natively available artifacts for Mac M1 osx-aarch_64.

### How to reproduce

Execute `./mvnw clean package -Dmaven.test.skip` to build on a Mac M1 without using `-Dos.detected.classifier=osx-x86_64` as suggested in https://github.com/apache/skywalking/blob/master/docs/en/FAQ/How-to-build-with-mac-m1.md

### Anything else

_No response_

### Code of Conduct

toffentoffen:
Upgrade to the latest grpc-java 1.42.1 which is only compatible and [tested against](https://github.com/grpc/grpc-java/pull/8551#discussion_r714182190) 3.17.2, but we can assume that 3.17.3 (which has osx-aarch_64 compatible maven artifacts) is save to use.

We could have also upgrade to the latest grpc-java 1.42.1 and latest protoc 3.19.1, but it can not be directly done, it needs to tweak the grpcs-java internal dependencies. So we choose the safest option and discarded this option because of the following reasons:
 - Protoc 3.19.1 generates code that relies on methods not available in protobuf-java 3.17.2 ([“issue”](https://github.com/protocolbuffers/protobuf/issues/9236#issue-1059594027)) (which java-grpc relies on). So we need to tell grpc-java via dependency management to use 3.19.1 instead of 3.17.2.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #8184.


kezhenxu94:
Can you also remove the FAQ regarding the M1 compiling question?

wu-sheng:
We should keep as agent side and old releases still have this issue.

kezhenxu94:
> We should keep as agent side and old releases still have this issue.

People using old releases should read the docs in their corresponding tag and agent should have their own doc and FAQ.

wu-sheng:
Make sense for old release. The current cases is that, docs of old releases are unsearchable.
Agent should be updated too.

wu-sheng (approved):

toffentoffen:
@kezhenxu94 should we then delete the [FAQ section for MAC M1](https://github.com/apache/skywalking/blob/master/docs/en/FAQ/How-to-build-with-mac-m1.md)? 
Or should I update it, pointing a comment to the merge commit mentioning that it is not needed for newer releases but still leave the document for older versions?

FOR gRPC java agent I will create another PR.

kezhenxu94:
> @kezhenxu94 should we then delete the [FAQ section for MAC M1](https://github.com/apache/skywalking/blob/master/docs/en/FAQ/How-to-build-with-mac-m1.md)? 
> 

I suggested to remove but it's @wu-sheng 's call.

wu-sheng:
After Java agent is done,  I think it is time to remove.

PR #8190, opened 2021-11-26
Use https for submodules to permit non-authenticated use

ken-duck:
Not a code bug or feature, but a minor tweak to your git modules configuration.

When cloning the code, if the user is unauthenticated they are rejected on one of the submodules with a message requesting your SSH key password, due to the .gitmodules file using `git@github.com` as opposed to `https://github.com/`.

This prevents users without Github accounts from checking out the code (without mucking about), and blocks the use of tools such as Sonatype Lift.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,4 +9,4 @@
 	url = https://github.com/apache/skywalking-rocketbot-ui.git
 [submodule "test/e2e-v2/java-test-service/e2e-protocol/src/main/proto"]
 	path = test/e2e-v2/java-test-service/e2e-protocol/src/main/proto
-	url = git@github.com:apache/skywalking-data-collect-protocol.git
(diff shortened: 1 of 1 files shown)


.gitmodules
```diff
@@ -9,4 +9,4 @@
 	url = https://github.com/apache/skywalking-rocketbot-ui.git
 [submodule "test/e2e-v2/java-test-service/e2e-protocol/src/main/proto"]
 	path = test/e2e-v2/java-test-service/e2e-protocol/src/main/proto
-	url = git@github.com:apache/skywalking-data-collect-protocol.git
+	url = https://github.com/apache/skywalking-data-collect-protocol.git
```
kezhenxu94, line 12: FYI @wankai123

kezhenxu94 (approved):

PR #8193, opened 2021-11-27
Introduce the Customized ALS Protocol for Satellite

mrproliu:
I am working on improving the performance of transparent message transmission for the [skywalking-satellite](https://github.com/apache/skywalking-satellite) project.

Then I realize the  ALS protocol needed to operate the stream on each identity stream message. The satellite needs more resources and time. This is especially true when envoys are high in number. 

Here is a performance test, using `2300` envoy instances are used to transparently transmit ALS messages to OAP through satellite, at a rate of `2.3K+` QPS. Then the satellite memory is full and not stable.

satellite so11y:
image
image

OAP so11y(cluster):
image

I have tried to cache the stream on the satellite side, but more resources are wasted due to resource preemption. So I think we could dependency on the gRPC streaming messages are order feature, using the same stream to send cross identity messages. 

After that, I did a test in this way, and the efficiency of sending ALS messages to OAP was improved, and the memory queue of the satellite was more stable.

This new protocol is copied from ALS and will only be called by the satellite.

Here have some screenshots to see the effect after optimization. Here, `2300` envoy instances are used to transparently transmit ALS messages to OAP through satellite, at a rate of `120K+` QPS. The ALS message tab is read from the `oap`, `envoy_als_in_count` metric. 

satellite so11y:
image
image

oap so11y(cluster):
image


kezhenxu94:
With OAP always parsing the identity node in every message entry, did you figure out how many CPU and memory increase in OAP side?

mrproliu:
> With OAP always parsing the identity node in every message entry, did you figure out how many CPU and memory increase in OAP side?

It only needs to add a bool and null judge every time and as the default, it's off. And I create another protocol as the entry for satellite only. If you want to take a look at oap side, here is the screenshot.
image

kezhenxu94:
For the original protocol it only adds a Boolean check that’s ok, what I meant is the new protocol for Satellite, it parses the node identifier every time, I just don’t want to move the performance cost from Satellite to OAP

mrproliu:
> For the original protocol it only adds a Boolean check that’s ok, what I meant is the new protocol for Satellite, it parses the node identifier every time, I just don’t want to move the performance cost from Satellite to OAP

Ok, understand. But if the satellite also using the original protocol, the OAP also only support one identify for one streaming, the satellite will open, send and close the stream every time. But we can cross the identity in one stream, I think it  good for they both, because we reduce the connect and close time.

kezhenxu94:
> Ok, understand. But if the satellite also using the original protocol, the OAP also only support one identify for one streaming, the satellite will open, send and close the stream every time. But we can cross the identity in one stream

If Satellite can send the messages in batch for even different nodes, why not just send messages in batch grouping by node, example: send messages in batch (say 100 messages) from node 1 in a stream and send 100 messages from node 2 in another stream so that OAP can parse only one node for each stream. Because you had already decided to provide a dedicate protocol for Satellite, you can consider providing the protocol in a more efficient way.

wu-sheng:
Let's back to the original question, where shows OAP will increase load? Why do you say there is a performance impact to OAP? @mrproliu 
From my understanding of this solution, there is only another new grpc service introduced, no performanceimpact for OAP at all.

wu-sheng:
@kezhenxu94 Let me give more clear context. Sending in batch is what we are doing in Satellite already. It definitely works.
In our research, we noticed, sending in bulk costs still too much in opening and closing streams when we face 2k+ envoy instances per satellite. So this is introduced. As it is adding only one flag, so there is no impact for background unpackaging and processing logic.

kezhenxu94:
Just 

> Let's back to the original question, where shows OAP will increase load? Why do you say there is a performance impact to OAP? @mrproliu
> From my understanding of this solution, there is only another new grpc service introduced, no performanceimpact for OAP at all.

The problem is that the new protocol is to replace the old one essentially, think in this way, an ALS message will be processed by and only by one of the protocols, if it’s processed by the old one it is (highly) possible that we don’t need to parse the node identifier but if it’s processed by the new one it always needs to parse the node identifier, do you agree the parsing of node identifier is a waste of resources?

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/AccessLogServiceGRPCHandler.java
```diff
@@ -105,7 +110,7 @@ public AccessLogServiceGRPCHandler(ModuleManager manager,
             public void onNext(StreamAccessLogsMessage message) {
                 HistogramMetrics.Timer timer = histogram.createTimer();
                 try {
-                    if (isFirst) {
+                    if (isFirst || (alwaysAnalyzeIdentity && message.getIdentifier() != null)) {
```
wu-sheng, line 113: Reply to https://github.com/apache/skywalking/pull/8193#issuecomment-980809798

@kezhenxu94 I don't think this will affect the parsing here. The parsing is already done, this if protobuf, not flatbuffer.
So, the only one more cost is reading `message.getIdentifier()`, which is a Java bean getting only.

oap-server/server-receiver-plugin/receiver-proto/src/main/proto/satellite/envoy/accesslog/v3/als.proto
```diff
@@ -0,0 +1,24 @@
...
+import "envoy/service/accesslog/v3/als.proto";
+
+option java_package = "org.apache.skywalking.satellite.envoy.accesslog.v3";
+option java_outer_classname = "SatelliteAlsProto";
+option java_multiple_files = true;
+option java_generic_services = true;
+
+// [#protodoc-title: Satellite gRPC Access Log Service (ALS)]
+
+// Duplicate the Envoy ALS protocol, work for satellite transmit the ALS message to oap.
```
wu-sheng, line 14: This is NOT duplicate, this is a new protocol(new service name changed)

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/AccessLogServiceGRPCHandler.java
kezhenxu94: You overlooked these:

```java
            for (ALSHTTPAnalysis analysis : envoyHTTPAnalysisList) {
                 role = analysis.identify(identifier, role);
            }
```

wu-sheng: How about this? Logic inside has nothing relating to this added flag. I am not following your question.
No matter in which case, this `identifier ` is required. The difference is from `StreamObserver#cache` or next element.

kezhenxu94: The difference to me is 

# old

```java
if (isFirst) { // only true for the first element per stream
            for (ALSHTTPAnalysis analysis : envoyHTTPAnalysisList) {
                 role = analysis.identify(identifier, role);
            }
}
```

# new

```java
if (true) { // it is no 100% always true but is extremely high possibility true from my understanding
            for (ALSHTTPAnalysis analysis : envoyHTTPAnalysisList) {
                 role = analysis.identify(identifier, role);
            }
}
```

does this illustrate my concern? @wu-sheng 

If @mrproliu you had experimented this, can you just post the so11y metrics (especially CPU) of OAP before and after the new protocol?

wu-sheng: If you mean flatbuffer parsing, them this is really having no different than before.
I think your gap is that, you are not considering we merged the stream, currently, one stream is representing the previous multiple streaming, then logically, all those identifiers are pasred too, just through different streamings.

mrproliu: > If @mrproliu you had experimented this, can you just post the so11y metrics (especially CPU) of OAP before and after the new protocol?

I could not happen on this, because it decides by the satellite message package mechanism. Only the first message in the batch could have the identity.

wu-sheng: I am not sure why this possiblily could be high. You are comparing this logic between one stream and 2k+ streaming. Then you have a conclusion, this one streaming has higher possible to parse identify.
More parsing per stream, yes, but no different when condidering this whole 2k streamings(2k envoy instances)

mrproliu: If you want I could do the experiment, we also using the new protocol and the `OAP` side is using current and this one. Then I post the so11y metrics.
```java```
if (isFirst || (alwaysAnalyzeIdentity && message.getIdentifier() != null)) {
                        identifier = message.getIdentifier();
                        isFirst = false;
                        role = Role.NONE;
                    }

                    for (ALSHTTPAnalysis analysis : envoyHTTPAnalysisList) {
                        role = analysis.identify(identifier, role);
                    }
```java```
I think only this one could find the identity analysis whether there is a performance problem.

Comment (kezhenxu94) [reply oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/AccessLogServiceGRPCHandler.java:113]: > If you mean flatbuffer parsing, then this is really having no different than before.

I never mean flatbuffer parsing. I meant this part 

```java
for (ALSHTTPAnalysis analysis : envoyHTTPAnalysisList) {
                 role = analysis.identify(identifier, role);
}
```

> I think your gap is that, you are not considering we merged the stream, currently, one stream is representing the previous multiple streaming, then logically, all those identifiers are pasred too, just through different streamings.

I’m rather aware of that you merged the streams, and because you merge the streams from different Envoy nodes the OAP needs to identify the node in every message

kezhenxu94: @wu-sheng forget the word “parse”, I don’t mean parsing flatbuffer, I mean identifying the role of the node 

```java
for (ALSHTTPAnalysis analysis : envoyHTTPAnalysisList) {
                 role = analysis.identify(identifier, role);
            }
```

kezhenxu94: Let’s have a DM discuss if you still don’t get my point

wu-sheng: I think I know what is the gap here. Let me put in this way

1. `Envoy -> OAP` direct linking

All envoy open unclosed streaming because the connection is always open, then the identifier only gets parsed once. This costs minimal in identifier parsing, but we have been facing load not balanced. Then we were introducing Satellite.

2. `Envoy->Satellite->OAP`, current satellite implementation.

It maintains the context for a bulk of ALS message(such as 100 per bulk), then it created stream every time. With this solution, the load is balanced like you have seen in previous Satellite PR. 
This change doesn't require any OAP side change, but actually, it already introduce the extra identifier parsing logic, you raised here(which is why I can't understand for a while, this is never introduced by this change).
One side, maintaining unclosed streaming in Satellite is impossible(costing more CPUs), also, keeping open/close streaming(current status) is also not good.

Then, we get the idea of this new protocol and new PR.

We are merging these streamings as one new streaming(unclosed this time) but carrying different identifiers.

.licenserc.yaml
```diff
@@ -57,6 +57,7 @@ header:
     - '**/src/main/proto/jaeger/**'
     - '**/src/main/proto/mixer/**'
     - '**/src/main/proto/policy/**'
+    - '**/src/main/proto/satellite/**'
```
kezhenxu94, line 60: I'd say these files are our own work, not third parties', so please add license headers to them instead of excluding them

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/AccessLogServiceGRPCHandler.java
wu-sheng: What you are concerned and asked about actually is already there after local balancer added. So, generally, you need a tradeoff, because trying to balance 2k+ envoy instances, the streaming must be merged one way or another.

But what is cost could be controlled by what is the size of Satellite side bulk setting. Higher is set, less CPU(parsing identifier) used, but more memory is used at Satellite.

wu-sheng: So, generally, your asking about more resource cost at OAP side is true, but not because of this flag, but load balancing mechanism.

.licenserc.yaml
wu-sheng: I was struggling about this, right now, I think yes, this is written and provided by us, let's add license for it.

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/AccessLogServiceGRPCHandler.java
kezhenxu94: > What you are concerned and asked about actually is already there after local balancer added. So, generally, you need a tradeoff, because trying to balance 2k+ envoy instances, the streaming must be merged one way or another.

I agree we need tradeoff and I've also given another merging strategy for discussion, which I think might be better https://github.com/apache/skywalking/pull/8193#issuecomment-980766033

> So, generally, your asking about more resource cost at OAP side is true, but not because of this flag, but load balancing mechanism.

I doubt this and need to confirm with @mrproliu .

Despite of all above, `if (isFirst || (alwaysAnalyzeIdentity && message.getIdentifier() != null))` is buggy because `message.getIdentifier()` is never null, it returns a default instance if `identifier == null`

kezhenxu94:
I'm feeling this is not well tested (because of https://github.com/apache/skywalking/pull/8193#discussion_r757836878) and we don't have a solid result of the new protocol's performance. I'd suggest not to include this in 8.9.0 so hasty or postpone the 8.9.0 release plan if we really need this in it.

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/AccessLogServiceGRPCHandler.java
wu-sheng: >   // Identifier data that will only be sent in the first message on the stream. This is effectively
  // structured metadata and is a performance optimization.
  Identifier identifier = 1;

This is from ALS definition. What do you mean `never null?

kezhenxu94: I said this from code level. `message.getIdentifier()` never returns null, protocol buffer compiler generates codes like `return identifier == null ? defaultInstance : identifier`, please use `message.hasIdentifier()`

wu-sheng: OK, make sense.

wu-sheng:
> I'm feeling this is not well tested (because of https://github.com/apache/skywalking/pull/8193#discussion_r757836878) and we don't have a solid result of the new protocol's performance. I'd suggest not to include this in 8.9.0 so hasty or postpone the 8.9.0 release plan if we really need this in it.

For ALS, yes, we need this. 
About release plan, I think either is working. 😄 We are not urgent in 8.9.0 release.
Having 8.9.0 release in this year should be good enough.

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/AccessLogServiceGRPCHandler.java
mrproliu: Fixed. thanks.

wu-sheng: I can't see the commit.

kezhenxu94 (approved):
LGTM

wu-sheng:
@mrproliu Please attach the screenshots of perf testing results here, which would help more people in the future when reading this.

oap-server/server-receiver-plugin/receiver-proto/src/main/proto/satellite/envoy/accesslog/v2/als.proto
```diff
@@ -0,0 +1,43 @@
...
+
+// The new Envoy ALS protocol, work for satellite transmit the ALS message to oap.
+service SatelliteAccessLogService {
+  // Use the same parameters to transmit access log messages.
+  // The only difference is that the identity information (StreamAccessLogsMessage#identity) may occur on each message.
+  // Rely on the streaming messages are orderly, so there will be no problems with message processing.
+  // Therefore, when the satellite transmits the ALS message, it does not need to open, send and close the stream for each different identity (envoy).
+  // As a result, unnecessary streaming operation requests could be reduced, and the satellite becomes more stable when the satellite sends requests to the upstream.
+  // Especially when the number of envoys increases, the optimization becomes more obvious.
+  rpc StreamAccessLogs(stream .envoy.service.accesslog.v3.StreamAccessLogsMessage) returns (.envoy.service.accesslog.v3.StreamAccessLogsResponse) {
```
kezhenxu94, line 41: This should be `v2`?

kezhenxu94 (changes requested):

oap-server/server-receiver-plugin/receiver-proto/src/main/proto/satellite/envoy/accesslog/v2/als.proto
mrproliu: Currently, the OAP ALS handler has combined the `v2` and `v3` version message and response, and they do not contain conflict. So I think we could just keep it, all they use the `v3` version.

kezhenxu94: For messages they are all v3, but for gRPC service they are different versions, if you want to keep in v3 service between Satellite and OAP, then I don't think you need to copy `v3/als.proto` to `v2/als.proto`, they are exactly the same.

wu-sheng: v3 should be correct. But this reminds me, do we really need v2? Satellite should be 100% good to forward v2 ALS to the v3 endpoint of OAP, right? We could add less codes in OAP for now. v3 is the only main stream.

wu-sheng: @kezhenxu94 That is what you did :) Maybe you forget it. v2 gRPC service with v3 parameter and return values.

https://github.com/apache/skywalking/blob/842b5d927e9c196e14a4dc3f73812d2c4adfebf4/oap-server/server-receiver-plugin/receiver-proto/src/main/proto/envoy/service/accesslog/v2/als.proto

wu-sheng: Anyway, I still prefer to using v3 only.

kezhenxu94: I know it, I remember it...

But what @mrproliu did is provide 2 exactly the same service with same message, what I did is 2 different service with same protocol buffer message, different Envoy version send via different service version.

**HIGHLIGHT**: service version and message version

wu-sheng: The only difference is `package name`, which is the ALS difference. Like I mentioned, this is not important at all. 
Use only `v3/als/proto` please @mrproliu

mrproliu: been removed v2 version protocol.

kezhenxu94 (approved):

PR #8194, opened 2021-11-29
Revert IoTDB as storage option due to negative perf testing result.

wu-sheng:

Revert #7766. I am so sorry @LIU-WEI-git @jixuan1989 I have to initialize this **revert** due to our recent performance testing result.

Here are the graphs of SkyWalking OAP server so11y(self-obsevability) result.
image
image

Before 9:45, the OAP includes the pull request of IoTDB storage implementation, you can see that, it provides much lower performance than it should be(after 9:45). The analysis capability drops from `30+k/s` to `< 10k/s`, which is hard to believe.

Both of these testing are running in new deployed environment, most importantly, IoTDB storage option is even **NEVER ACTIVATED**, which could only provide one possibility, the packages(jars) we introduced have clear conflicts and performance impact to SkyWalking server. 
This is the first time I see this happened, which surprised me too. But tests verified this repeatedly. So I am submitting this pull request, and will ask @mrproliu to verify again according this branch.


hanahmily (approved):
Sorry for seeing this. But it seems that we have to do it.

wu-sheng:
image
image
image

We have the 2nd time verification rechecked. Before 10:48, it was this PR's codebase testing, and after than we reset back to `master branch`(including IoTDB implementation for now).
Same drop happened. 

__

image

According to this dependencies change, we are going to do one more attempting, which is, removing `logback-*`, even I think this has very little chance to affect this, due to we don't output logs many things during testing. But, this is only possible guess, others have to be rechecked by IoTDB team.

jixuan1989:
We are keeping eye on this :D.

wu-sheng:
image
image
image

OK, after removing the logback, the perf seems back to normal. It seems log4j co-existing with logback causing huge performance impact. We are still checking.

wu-sheng:
@mrproliu If you could confirm this removing works, please submit a pull request to remove logback dependency.

@kezhenxu94 Is there a chance we could add a blacklist in the dependency check script to avoid this case in the future?

PR #8202, opened 2021-11-29
Remove `logback` dependencies in IoTDB plugin

mrproliu:
### Restore OAP performance back to IoTDB plugin added before, follow https://github.com/apache/skywalking/pull/8194#issuecomment-981267038 for more details.


kezhenxu94 (commented):
license, known-dependencies.txt should be updated

dist-material/release-docs/LICENSE
```diff
@@ -398,10 +398,6 @@ EPL licenses
 The following components are provided under the EPL License. See project link for details.
 The text of each license is also included at licenses/LICENSE-[project].txt.
```
wu-sheng, line 399: You should remove these licenses.

mrproliu: been removed, forget to update this before.

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
```diff
@@ -49,6 +49,14 @@
                     <groupId>net.jpountz.lz4</groupId>
                     <artifactId>lz4</artifactId>
                 </exclusion>
+                <exclusion>
```
wu-sheng, line 52:
```suggestion
               <!-- logback has runtime negative impact about performance, -->
               <!-- Disable this, and use slf4j-sl4j instead of. -->
                <exclusion>
```

wu-sheng (dismissed):

wu-sheng (dismissed):

dist-material/release-docs/LICENSE
```diff
@@ -391,17 +391,6 @@ CC0-1.0 licenses
...
 reactive-streams 1.0.2:	https://github.com/reactive-streams/reactive-streams-jvm , CC0-1.0
 
-========================================================================
-EPL licenses
-========================================================================
-
-The following components are provided under the EPL License. See project link for details.
-The text of each license is also included at licenses/LICENSE-[project].txt.
-
-    logback 1.2.3: https://github.com/qos-ch/logback: EPL 1.0
```
kezhenxu94, line 401: I'm afraid this needs to be reserved because we still use logback in web app and it's in our distributed tar.

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
```diff
@@ -49,6 +49,16 @@
                     <groupId>net.jpountz.lz4</groupId>
                     <artifactId>lz4</artifactId>
                 </exclusion>
+                <!-- logback has runtime negative impact about performance, -->
```
kezhenxu94, line 52: IMHO this is a kind of arbitrary decision and we didn't even have a solid conclusion about the reason of https://github.com/apache/skywalking/pull/8194 , let's remove this kind of sentence, to me the only reason to exclude this is that we already have slf4j implementation (log4j2) in the classpath

wu-sheng: I added this, not @mrproliu 

I am thinking the performance impact is real, but we don't know why.

dist-material/release-docs/LICENSE
mrproliu: Ok, I see the web app has used `logback-classic` and `logback-core`, I will added back.

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
kezhenxu94: > I added this, not @mrproliu

Don't take it personal, I was just commenting in the perspective of logback's community, no matter who, without benchmark test result / evidence, saying others have pool performance is not cool, I still remember someone said SkyWalking is pool performant in Twitter, but they were not aware of that they misused SkyWalking, I felt the same uncomfortable that time.

> I am thinking the performance impact is real, but we don't know why.

The performance impact is real, but we can only say it exists when logback + log4j2 + slf4j are all in the classpath, if we really want to put this sentence, we at least should test another case where we remove log4j2 and keep logback + slf4j.

wu-sheng: I can understand the statement to co-existing to make more accurate. And ref the PR with screenshots as prove.

oap-server/server-storage-plugin/storage-iotdb-plugin/pom.xml
```diff
@@ -49,6 +49,18 @@
                     <groupId>net.jpountz.lz4</groupId>
                     <artifactId>lz4</artifactId>
                 </exclusion>
+                <!-- When logback + log4j2 + slf4j co-existed in OAP server classpath at the runtime, -->
+                <!-- SkyWalking so11y reported a runtime negative impact about performance. -->
+                <!-- Disable this, and use slf4j-log4j instead of. -->
+                <!-- Ref original discussion, https://github.com/apache/skywalking/pull/8194 -->
```
wu-sheng, line 55: @kezhenxu94 Please recheck.

kezhenxu94 (approved):
LGTM

PR #8215, opened 2021-11-30
Support disables the verification of server's TLS certificate chain for specific hosts

JaredTan95:
###  Support disables the verification of server's TLS certificate chain for specific hosts by `SW_STORAGE_ES_SSL_INSECURE_HOSTS` env.
**NOTE**: You should never use this in production but only for a testing purpose.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -62,6 +62,7 @@ Release Notes.
 * Support using IoTDB as a new storage option.
 * Add customized envoy ALS protocol receiver for satellite transmit batch data.
 * Remove `logback` dependencies in IoTDB plugin.
+* Fix `StorageModuleElasticsearchProvider` doesn't watch on `trustStorePath`.
 
 #### UI
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -63,7 +63,7 @@ Release Notes.
 * Add customized envoy ALS protocol receiver for satellite transmit batch data.
 * Remove `logback` dependencies in IoTDB plugin.
 * Fix `StorageModuleElasticsearchProvider` doesn't watch on `trustStorePath`.
-
+* Support disables the verification of server's TLS certificate chain for specific hosts by `SW_STORAGE_ES_SSL_INSECURE_HOSTS` env.
 #### UI
 
 * Optimize endpoint dependency.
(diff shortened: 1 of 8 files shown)


wu-sheng:
Notice, #8179 got rejected due to skipping the secure checking.
This is an insecure way for general use.
Couldn't you explain more?

JaredTan95:
Okay, I missed this https://github.com/apache/skywalking/pull/8179 issue.
Feedback from some customers, in some user cases, they don't bind domain on elasticseach multi-node , when they set up skywalking in their `dev` or `test` environment, the connection between oap and es could not be established.
So, We could provide this way for ease of debugging and deployment.
And, we've added instructions, don't use it in production, and this feature don't skip any validation by default.

wu-sheng:
No TLS verification should be fine, but don't need through us, actually. They could put a Nginx/Envoy in the front of ElasticSearch server, then OAP server could access HTTP endpoint without TLS.
This is more comfortable for SkyWalking about that, we are not violating TLS which has explicitly activated in configurations.
Is this working for you?

JaredTan95:
> No TLS verification should be fine, but don't need through us, actually. They could put a Nginx/Envoy in the front of ElasticSearch server, then OAP server could access HTTP endpoint without TLS. This is more comfortable for SkyWalking about that, we are not violating TLS which has explicitly activated in configurations. Is this working for you?

The proxy way may be a working way, but it may cause a little change for deployment architecture. No matter, I closed it temporarily. 

And continue to see the community feedback

wu-sheng:
> > No TLS verification should be fine, but don't need through us, actually. They could put a Nginx/Envoy in the front of ElasticSearch server, then OAP server could access HTTP endpoint without TLS. This is more comfortable for SkyWalking about that, we are not violating TLS which has explicitly activated in configurations. Is this working for you?
> 
> The proxy way may be a working way, but it may cause a little change for deployment architecture. I closed it temporarily.
> 
> And continue to see the community feedback

Let's see, I put more thoughts behind this at here, https://twitter.com/wusheng1108/status/1465833723115409410
If others follow this thread, it should be worth to take a look.

JaredTan95:
related discusson https://github.com/apache/skywalking/discussions/8154

PR #8218, opened 2021-12-01
Fix a wrong check about entity if GraphQL at endpoint relation level.

wu-sheng:

A nit, reported in the discussion channel, but the reporter doesn't report the patch in 2 days. So, I fix this in case we forgot.


hanahmily (approved):
LGTM

PR #8234, opened 2021-12-02
Fix potential NPE in OAL string match and a bug when right-hand-side variable includes double quotes

kezhenxu94:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ Release Notes.
 
 #### OAP Server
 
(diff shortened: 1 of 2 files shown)


kezhenxu94:
It's OK for me to (or not to) include this in 8.9.0

wu-sheng:
As it is potential, let's add it to next iteration. :)

wu-sheng:
Version change gets merged. Please resolve conflicts.

wu-sheng (approved):

PR #8247, opened 2021-12-06
OAP-stream integration with Banyandb

### LINKED ISSUE #7732: [Core] Stream data processing
### Search before asking

### Description

In https://github.com/apache/skywalking/issues/6922, we accomplished most features of the trace module and other core components that support processing tracing data. We even did a banyandb java client for OAP. 

Unfortunately, when I was implementing the index module, there was a defect in the original design that blocked us to make the trace feature happen.  Based on that, I set up this new issue to track current tasks.

Now we selected `Stream` as the official notion to ingest trace and log items. The `trace` module will be dropped once this transferring get done.

- [x] Upgrade API to v2. In v2, I introduce `stream` to substitute `trace` and tweak index relevant objects.
- [x] Stream module.
- [x] TSDB module. It's a time-series database engine to handle all of the data models.
- [x] Several index packages. Including inverted index, tree(LSM-tree) index, term dictionary, and query execution tree.
- [x] Prune outdated modules that are `trace`, `database`  and `index`.
- [x] Query module. Adjust logical execution plan accordingly. (@lujiajing1126)
- [x] Liaison. Adapt to the new stream protocols.
- [x] Metadata empowered by embed Etcd. (@lujiajing1126) 
- [x] Java client. Adapt to the new stream protocols. (@lujiajing1126)
- [x] OAP storage layer https://github.com/apache/skywalking/issues/7412.(@lujiajing1126)

### Code of Conduct

lujiajing1126:
### Introduce OAP integration with BanyanDB stream 

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7732

This PR intends to integrate stream related entities to BanyanDB. And this is the last task listed in issue #7732 

I will open another issue to track the remaining features which are not currently supported by BanyanDB but are used in the query protocol of OAP server.

UT and documents will be added later.

@hanahmily


wu-sheng:
I think you may use a wrong milestone? That milestone should track issues for server side release.

wu-sheng:
Could you share what is the pattern of `AbstractBanyanDBDeserializer` about? I am sure this pattern is not being used in other storage implementation.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/stream/AbstractBanyanDBDAO.java
```diff
@@ -0,0 +1,59 @@
...
+    protected AbstractBanyanDBDAO(BanyanDBStorageClient client) {
+        super(client);
+    }
+
+    protected <T> List<T> query(Class<T> clazz, QueryBuilder builder) {
+        return this.query(clazz, builder, 0, 0);
+    }
+
+    protected <T> List<T> query(Class<T> clazz, QueryBuilder builder, long startTimestamp, long endTimestamp) {
+        AbstractBanyanDBDeserializer<T> deserializer = BanyanDBDeserializerFactory.INSTANCE.findDeserializer(clazz);
```
wu-sheng, line 41: This kind of factory seems less performance? Most DAO implementations are simple logic, could we make BanyanDB DAOs more consistent with others?

wu-sheng: The reason behinds the codes are, this kind of logic could be optimized and very custom for every single cases. Adding a deserializer seems moving logic of one method to other places, and only makes things more complex.

lujiajing1126: Make sense to me. I could move the deserializer to each query DAO to optimize this hashmap search away in the sense that almost each kind of queryDAO uses its own deserializer.

lujiajing1126:
> Could you share what is the pattern of `AbstractBanyanDBDeserializer` about? I am sure this pattern is not being used in other storage implementation.

The reason to introduce the `AbstractBanyanDBDeserializer` is that the storage build for BanyanDB is asymmetric, i.e. what we write is several lists of `SerializableTag`s while we read `TagAndValue` from database. Both of these lists are strictly ordered which should comply with the order of schema definition and the order of projection declaration.

Another motivation is that default implementation of `storageBuilder` within the DAO is less performant since binary data is always encoded to the base64 format which is not necessary in the case of BanyanDB. We natively support binary data format.

Besides, I suppose it is not necessary to fetch the full storage entity as an intermediate representation while querying. We could do the projection to reduce bandwidth usage.

wu-sheng:
Writing as async is the common case for all storage implementation. I am not sure what do you mean. All metrics/logs/traces in all storage implementations are async. Please check the exiting codes.

wu-sheng:
> Besides, I suppose it is not necessary to fetch the full storage entity as an intermediate representation while querying. We could do the projection to reduce bandwidth usage.

This is also unnecessary for query. We don't do that in other storage implementation too.

wu-sheng:
> Another motivation is that default implementation of storageBuilder within the DAO is less performant since binary data is always encoded to the base64 format which is not necessary in the case of BanyanDB. We natively support binary data format.

StorageBuilderFactory and StorageBuilder make sense to me to get rid of Map-based mechanism. My question is more about deserializer in query stage.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/stream/BanyanDBStorageDAO.java
```diff
@@ -0,0 +1,140 @@
...
+        } catch (NoSuchMethodException noSuchMethodException) {
+            log.error("cannot find method storage2Entity", noSuchMethodException);
+            throw new RuntimeException("cannot find method storage2Entity");
+        }
+    }
+
+    @Override
+    public INoneStreamDAO newNoneStreamDao(StorageBuilder storageBuilder) {
+        try {
+            Class<?> returnType = storageBuilder.getClass().getMethod("storage2Entity", Map.class).getReturnType();
```
wu-sheng, line 114: Why do we need reflect in this kind of DAO?

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/stream/BanyanDBStorageDAO.java
```diff
@@ -0,0 +1,140 @@
...
+            } else if (InstanceTraffic.class.equals(returnType)) {
+                return new BanyanDBMetricsDAO<>(new Metadata.InstanceTrafficBuilder());
+            } else if (EndpointTraffic.class.equals(returnType)) {
+                return new BanyanDBMetricsDAO<>(new Metadata.EndpointTrafficBuilder());
+            } else if (NetworkAddressAlias.class.equals(returnType)) {
+                return new BanyanDBMetricsDAO<>(new NetworkAddressAliasBuilder());
+            } else {
+                throw new IllegalStateException("record type is not supported");
+            }
+        } catch (NoSuchMethodException noSuchMethodException) {
```
wu-sheng, line 80: MetricsDao with metadata only? Are you sure this is the right way to do this?
`StorageBuilderFactory#Default` bases on the class scan mechanism. If you want your own, you should create another way, such as codes or annotation to declare this. But never do this kind of reflect and `if/else`

wu-sheng:
```java
@Stream(name = ServiceTraffic.INDEX_NAME, scopeId = DefaultScopeDefine.SERVICE,
    builder = ServiceTraffic.Builder.class, processor = MetricsStreamProcessor.class)
@MetricsExtension(supportDownSampling = false, supportUpdate = false)
@EqualsAndHashCode(of = {
    "name",
    "nodeType"
})
public class ServiceTraffic extends Metrics {
```

This is how Builder is declared and mapped to the entity. I think you should have a class or a class with annotations to declare the override? Such as(only an example)
```java
@StreamBuilderOverride(name = ServiceTraffic.INDEX_NAME,  builder = BanyanDBServiceTrafficBuilder.class)
@StreamBuilderOverride(name = InstanceTraffic.INDEX_NAME,  builder = BanyanDBInstanceTrafficBuilder.class)
public class BanyanDBStorageBuilderFactory
```

oap-server/server-storage-plugin/storage-banyandb-plugin/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+        <groupId>org.apache.skywalking</groupId>
+        <version>8.10.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>storage-banyandb-plugin</artifactId>
+    <packaging>jar</packaging>
+
+    <dependencies>
+        <dependency>
```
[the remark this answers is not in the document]
wu-sheng: @kezhenxu94 @hanahmily We should take care of these CVEs by bumping up guava, jetty, etc.

lujiajing1126:
> > Besides, I suppose it is not necessary to fetch the full storage entity as an intermediate representation while querying. We could do the projection to reduce bandwidth usage.
> 
> This is also unnecessary for query. We don't do that in other storage implementation too.

I suppose at least IOTDB implementation is doing this?

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/stream/BanyanDBStorageDAO.java
lujiajing1126: I missed this part. I will check it.

lujiajing1126:
> > Another motivation is that default implementation of storageBuilder within the DAO is less performant since binary data is always encoded to the base64 format which is not necessary in the case of BanyanDB. We natively support binary data format.
> 
> StorageBuilderFactory and StorageBuilder make sense to me to get rid of Map-based mechanism. My question is more about deserializer in query stage.

But the Map-based `storage2Entity`s are also doing base64 decoding, for example builders in `AlarmRecord`, `BrowserErrorLogRecord` etc.

Besides,

wu-sheng:
> But the Map-based storage2Entitys are also doing base64 decoding, for example builders in AlarmRecord, BrowserErrorLogRecord etc.

Only builder did BASE64, I think we are clear builder should be override.

> I suppose at least IOTDB implementation is doing this?

This is the newest implementation. Why do you learn from it? BanyanDB DAO would be closest to core and native concept. You should not learn from IoTDB implementation.

lujiajing1126:
> > But the Map-based storage2Entitys are also doing base64 decoding, for example builders in AlarmRecord, BrowserErrorLogRecord etc.
> 
> Only builder did BASE64, I think we are clear builder should be override.
> 
> > I suppose at least IOTDB implementation is doing this?
> 
> This is the newest implementation. Why do you learn from it? BanyanDB DAO would be closest to core and native concept. You should not learn from IoTDB implementation.

Well. It has been merged. So I guess it could be a referential impl.

I've checked JDBC impl again. I could understand what you meant. I could move the deserialization logic to the query DAO just what we've done in JDBC, influxdb impl. Then we don't need to the deser abstraction.

wu-sheng:
> I've checked JDBC impl again. I could understand what you meant. I could move the deserialization logic to the query DAO just as what we've done in JDBC, influxdb impl. Then we don't need to the deser abstraction.

Most importantly, let's make Builder and BuilderFactory work to override the core. That is the most important.

lujiajing1126:
> > I've checked JDBC impl again. I could understand what you meant. I could move the deserialization logic to the query DAO just as what we've done in JDBC, influxdb impl. Then we don't need to the deser abstraction.
> 
> Most importantly, let's make Builder and BuilderFactory work to override the core. That is the most important.

What about registering customized `StorageBuilderFactory` which can be done in the `prepare` stage of `ModuleProvider`.

By default, `new StorageBuilderFactory.Default()` is used.

wu-sheng:
> What about registering customized StorageBuilderFactory which can be done in the prepare stage of ModuleProvider.
By default, new StorageBuilderFactory.Default() is used.

Yes, we should override it through this way. Notice, this was designed and added by myself in the very beginning, but we never used in upstream in this way. Please let me know if this mechanism has an issue.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/deserializer/SegmentRecordDeserializer.java
```diff
@@ -0,0 +1,53 @@
...
+        record.setSegmentId(row.getId());
+        record.setTraceId((String) searchable.get(0).getValue());
+        record.setIsError(((Number) searchable.get(1).getValue()).intValue());
+        record.setServiceId((String) searchable.get(2).getValue());
+        record.setServiceInstanceId((String) searchable.get(3).getValue());
+        record.setEndpointId((String) searchable.get(4).getValue());
+        record.setLatency(((Number) searchable.get(5).getValue()).intValue());
+        record.setStartTime(((Number) searchable.get(6).getValue()).longValue());
+        final List<TagAndValue<?>> data = row.getTagFamilies().get(1);
+        record.setDataBinary(((ByteString) data.get(0).getValue()).toByteArray());
```
hanahmily, line 50: The module installer you don't create contains the metadata on how the index number maps to the tag. I would like to conceal these numbers under the installer, which is easier to maintain the codes.

hanahmily (commented):
A quick view of the codes

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBStorageConfig.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    private String group = "default";
+    /**
+     * The maximum size of write entities in a single batch write call.
+     */
+    private int maxBulkSize = 5000;
+    /**
+     * Period of flush interval. In the timeunit of seconds.
+     */
+    private int flushInterval = 15;
```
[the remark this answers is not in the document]
KirannKaur: @sonatype-lift help

wu-sheng:
Before you do more integration, the metadata query DAO has re-implemented due to metadata-v2 query API accepted for SkyWalking v9.

Please sync the `banyandan-integration` branch from the master.

PR #8285, opened 2021-12-12
Remove log4j version properties in webapp

wu-sheng:

> Turns out to be the case, the property <log4j.version>2.6.2</log4j.version> is not used, please remove it instead

Quote from https://github.com/apache/skywalking/pull/8285#issuecomment-991826955, webapp doesn't use log4j2.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,8 +18,6 @@ Release Notes.
 
 #### UI
 
-* (UI Web Server)Upgrade log4j2 to 2.15.0 for CVE-2021-44228. This CVE only effects on JDK if JDNI is opened in default.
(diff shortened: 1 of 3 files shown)


fgksgf (approved):

kezhenxu94 (approved):
I thought we use logback in webapp

mrproliu (dismissed):

kezhenxu94 (changes requested):
> I thought we use logback in webapp

Turns out to be the case, the property is not used, please remove it instead

wu-sheng:
OK, let me remove this.

kezhenxu94 (approved):
LGTM

PR #8299, opened 2021-12-15
Upgrade log4j2 to 2.16.0

wu-sheng:

According to [ASF official post](https://blogs.apache.org/foundation/entry/apache-log4j-cves), **fix in Log4j 2.15.0 was incomplete in some non-default configurations and could allow an attacker to execute a denial of service (DoS) attack.**

DoS is not a common case SkyWalking would face, so, I would only put it in the master branch for next(9.0) release.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,8 +7,9 @@ Release Notes.
 
 #### Project
 
-* Upgrade log4j2 to 2.15.0 for CVE-2021-44228. This CVE only effects on JDK if JDNI is opened in default. Notice, using
-  JVM option `-Dlog4j2.formatMsgNoLookups=true` also avoids CVE if your JRE opened JNDI in default.
+* Upgrade log4j2 to 2.16.0 for CVE-2021-44228 and CVE-2021-45046. This CVE only effects on JDK if JDNI is opened in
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -7,8 +7,9 @@ Release Notes.
 
 #### Project
 
-* Upgrade log4j2 to 2.15.0 for CVE-2021-44228. This CVE only effects on JDK if JDNI is opened in default. Notice, using
-  JVM option `-Dlog4j2.formatMsgNoLookups=true` also avoids CVE if your JRE opened JNDI in default.
+* Upgrade log4j2 to 2.16.0 for CVE-2021-44228 and CVE-2021-45046. This CVE only effects on JDK if JDNI is opened in
+  default. Notice, using JVM option `-Dlog4j2.formatMsgNoLookups=true` or setting
```
kezhenxu94, line 11: We are upgrading from 2.9 so the “notice” seems not work at all

wu-sheng: No, we are not. Check this.

https://github.com/apache/skywalking/blob/v8.9.0/tools/dependencies/known-oap-backend-dependencies.txt#L96-L99

kezhenxu94 (approved):

PR #8319, opened 2021-12-21
Change satellite MAL metrics entity to service

mrproliu:
Change the satellite metrics as Service Entity, so the [skywalking-swck HPA](https://github.com/apache/skywalking-swck/blob/master/docs/custom-metrics-adapter.md) could use them.

Because this change is already in the CHANGES.md, so don't need to update it again.


oap-server/server-starter/src/main/resources/meter-analyzer-config/satellite.yaml
```diff
@@ -13,22 +13,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-expSuffix: tag({tags -> tags.service = 'satellite::' + tags.service}).instance(['service'], ['instance'])
+expSuffix: tag({tags -> tags.service = 'satellite::' + tags.service}).service(['service'])
```
wu-sheng, line 16: I think `instance` downsampling function would generate both service and service instance metrics, according to MAL doc.

wu-sheng: I think it is better to just rename the metric(removing `instance_`). Or do you consider instance level metrics are not useful later? I was planning to consider Satellite so11y should be a part of the SkyWalking so11y in the v9 new dashboard.

mrproliu: There is a one-to-one correspondence between the metric names with the entity level, and I see the MAL could only generate the same name metric, why it could generate the service level metric? Do I miss anything?

wu-sheng: Metric name has nothing related to the generated result.

I said so according to the doc, https://skywalking.apache.org/docs/main/v8.9.1/en/concepts-and-designs/mal/#metric-level-function

mrproliu: But when I query the metrics without the instance argument, the data is always zero.
Such as:
```shell
# could generate the result
swctl metrics linear --name satellite_instance_server_cpu_utilization --service-name satellite::satellite-service --instance-name satellite-instance
# no data responded
swctl metrics linear --name satellite_instance_server_cpu_utilization --service-name satellite::satellite-service
```

wu-sheng: @wankai123 Should I consider this is a bug? `SampleFamily#buildMeterEntity` doesn't return a List, such as in instance case. But the document says otherwise.

wu-sheng: OK, I am following up the context. This PR is good, the MAL doc should be updated.

wu-sheng (approved):

PR #8320, opened 2021-12-21
Upgrade maven-wrapper to 3.1.0, maven to 3.8.4 for performance improvements and ARM more native support

kezhenxu94:
maven-wrapper is donated to ASF with a version upgraded, I'm upgrading to the new version 3.1.0 that was released under ASF umbrella.

Also upgrade maven version to 3.8.4 that was said to have performance improvements and better support in ARM architecture.

- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md). Also add missing change log for #8317


wu-sheng (approved):

wu-sheng:
This seems slightly reducing Intel platform compiling time, from 5-6min to  < 5 min.

PR #8327, opened 2021-12-23
Fix `Can't split endpoint id into 2 parts` bug for endpoint ID in the mesh's TCP case.

wu-sheng:


wu-sheng:
Reported by @toffentoffen

kezhenxu94 (approved):

PR #8328, opened 2021-12-23
Upgrade H2 version to 2.0.202 to fix CVE-2021-23463

kezhenxu94:

### DIFF
--- a/oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/h2/dao/H2TopNRecordsQueryDAO.java
+++ b/oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/h2/dao/H2TopNRecordsQueryDAO.java
@@ -24,6 +24,7 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.List;
+import lombok.RequiredArgsConstructor;
 import org.apache.skywalking.oap.server.library.util.StringUtil;
 import org.apache.skywalking.oap.server.core.analysis.IDManager;
 import org.apache.skywalking.oap.server.core.analysis.topn.TopN;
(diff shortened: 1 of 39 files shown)


wu-sheng:
> avoid false negative report from security scanning tools

What do you mean a false report? A CVE can't be produced in our case?

kezhenxu94:
> > avoid false negative report from security scanning tools
> 
> What do you mean a false report? A CVE can't be produced in our case?

I mean the CVE can't actually exist in the user cases (in production env) because H2 is only for local test, WDYT?

wu-sheng:
> > > avoid false negative report from security scanning tools
> > 
> > 
> > What do you mean a false report? A CVE can't be produced in our case?
> 
> I mean the CVE can't actually exist in the user cases (in production env) because H2 is only for local test, in another word, by "false negative report" I mean it's totally OK if we don't fix this, WDYT?

I think, from recommendation perspective, this CVE `shouldn't` rather than `can't`, because it can. H2 server could run in a server mode, which is more powerful and also could support OAP in cluster like MySQL did.

Referring from official doc, https://skywalking.apache.org/docs/main/latest/en/setup/backend/backend-storage/#h2
> Activate H2 as storage, set storage provider to H2 In-Memory Databases. Default in distribution package. Please read Database URL Overview in H2 official document. You can set the target to H2 in Embedded, Server and Mixed modes.

We don't declare this should not be used in prod env.

___
Anyway, the above is just my feeling. I just want to remove ` to avoid false negative report from security scanning tools` from PR title and change log, because this is a real CVE.

___
This change seems breaking the e2e?

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/h2/dao/H2TopNRecordsQueryDAO.java
```diff
@@ -43,8 +45,11 @@ public H2TopNRecordsQueryDAO(JDBCHikariCPClient h2Client) {
 
     @Override
     public List<SelectedRecord> readSampledRecords(final TopNCondition condition,
-                                                   final String valueColumnName,
+                                                   String valueColumnName,
                                                    final Duration duration) throws IOException {
+        if (H2_RESERVED_WORDS.contains(valueColumnName)) {
+            valueColumnName = String.format("`%s`", valueColumnName);
+        }
```
wu-sheng, line 52: These are duplicated.codes, could you move them to a method named formatColumnName?

kezhenxu94:
Here the patch becomes kind of unexpectedly large, because `value` becomes a reserved keyword in h2 and we didn't escape the column names, I slightly refactor the codes to add an escaper for JDBC implementations to escape the column/table names.

wu-sheng:
Noticed this. Do you check the solution of column alias mechanism in MySQL? I remember we have this in MySQL to prevert this conflict.

kezhenxu94:
> Noticed this. Do you check the solution of column alias mechanism in MySQL? I remember we have this in MySQL to prevert this conflict.

That would be preferable, I'll check

PR #8367, opened 2021-12-30
Support OAP server v9 Core concept: Layer, backend side

wankai123:

related to: #8241


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/IDManager.java
```diff
@@ -35,19 +35,19 @@
...
-         */
-        public static String buildId(String name, NodeType type) {
-            return buildId(name, type.equals(NodeType.Normal) || type.equals(NodeType.Browser));
-        }
+//        /**
+//         * @return encoded service id
+//         */
+//        public static String buildId(String name, NodeType type) {
+//            return buildId(name, type.equals(NodeType.Normal) || type.equals(NodeType.Browser));
+//        }
```
wu-sheng, line 43: Tip, remove legacy codes.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/Layer.java
```diff
@@ -0,0 +1,105 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.oap.server.core.analysis;
+
+import org.apache.skywalking.oap.server.core.UnexpectedException;
+
+public enum Layer {
```
wu-sheng, line 23: We need comments on Layer. Such as copy from https://github.com/apache/skywalking-query-protocol/blob/d9039e01767bc79e845bb32dbbeb6cdf74403af5/metadata-v2.graphqls#L40-L54

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/service/ServiceTraffic.java
```diff
@@ -35,40 +35,63 @@
...
+    @Setter
+    @Getter
+    @Column(columnName = LAYER)
+    private Layer layer = Layer.undefined;
+
+    @Setter
+    private boolean isNormal = true;
+
     @Override
     protected String id0() {
```
wu-sheng, line 89: We need comments to explain what `id0` is about, and what is the `#serviceId` rule.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/service/ServiceTraffic.java
```diff
@@ -35,40 +35,63 @@
...
     @Column(columnName = GROUP)
     private String group;
 
+    @Setter
+    @Getter
+    @Column(columnName = LAYER)
+    private Layer layer = Layer.undefined;
+
+    @Setter
+    private boolean isNormal = true;
```
wu-sheng, line 86:
```suggestion
    /**
     * The `normal` status represents this service is detected by an agent.
     *
     * The un-normal service is conjectured by telemetry data collected from agents on/in the `normal` service.
     */
    @Setter
    private boolean isNormal = true;
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/service/ServiceTraffic.java
```diff
@@ -110,16 +136,20 @@ public ServiceTraffic storage2Entity(final Map<String, Object> dbMap) {
         @Override
         public Map<String, Object> entity2Storage(final ServiceTraffic storageData) {
             final String serviceName = storageData.getName();
-            if (NodeType.Normal.equals(storageData.getNodeType())) {
```
wu-sheng, line 113: The normal service condition should be kept. Because no matter what service name of a conjectured service is, it is conjectured by the collected telemetry data. `Group name` should be set by users only.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/service/ServiceTraffic.java
```diff
@@ -98,8 +121,11 @@ public void deserialize(final RemoteData remoteData) {
         public ServiceTraffic storage2Entity(final Map<String, Object> dbMap) {
             ServiceTraffic serviceTraffic = new ServiceTraffic();
             serviceTraffic.setName((String) dbMap.get(NAME));
-            serviceTraffic.setNodeType(NodeType.valueOf(((Number) dbMap.get(NODE_TYPE)).intValue()));
+            serviceTraffic.setShortName((String) dbMap.get(SHORT_NAME));
             serviceTraffic.setGroup((String) dbMap.get(GROUP));
+            if (dbMap.get(LAYER) != null) {
+                serviceTraffic.setLayer(Layer.valueOf(((Number) dbMap.get(LAYER)).intValue()));
+            }
```
wu-sheng, line 128: You should add a `else` here to set Layer as default(`undefined`) for the legacy data.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/MetadataQueryService.java
```diff
@@ -60,49 +58,61 @@ private IMetadataQueryDAO getMetadataQueryDAO() {
...
+//                                    .collect(Collectors.toList());
+//    }
+//
+//    public List<Service> searchBrowserServices(final long startTimestamp, final long endTimestamp,
+//                                               final String keyword) throws IOException {
+//        return getMetadataQueryDAO().searchServices(NodeType.Browser, keyword)
+//                                    .stream()
+//                                    .distinct()
+//                                    .collect(Collectors.toList());
+//    }
```
wu-sheng, line 95: Remove commented codes please.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetadataQuery.java
```diff
@@ -54,68 +51,53 @@ private MetadataQueryService getMetadataQueryService() {
         return metadataQueryService;
     }
 
-    /**
```
wu-sheng, line 57: `@Deprecated` is required on methods and class level.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetadataQueryV2.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.skywalking.oap.server.core.analysis.IDManager;
+import org.apache.skywalking.oap.server.core.query.MetadataQueryService;
+import org.apache.skywalking.oap.server.core.query.input.Duration;
+import org.apache.skywalking.oap.server.core.query.type.Endpoint;
+import org.apache.skywalking.oap.server.core.query.type.EndpointInfo;
+import org.apache.skywalking.oap.server.core.query.type.Service;
+import org.apache.skywalking.oap.server.core.query.type.ServiceInstance;
+import org.apache.skywalking.oap.server.library.module.ModuleManager;
+
+public class MetadataQueryV2 implements GraphQLQueryResolver {
```
wu-sheng, line 37: Comments are required.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetadataQuery.java
wu-sheng: Comments are required, ref to `MetricQuery`.

wu-sheng:
Generally, this PR is good. There are several e2e verifications failed, please recheck.
The following things are required to change
1. The Change Log. This change is adding a new concept, and break some parts from v8, such as MySQL/PostgreSQL/H2(I am not sure for InfluxDB and IoTDB) are not working. We need change log to tell users which tables(of which storages)(`service_traffic`, `instance_traffic` or more?) are required to removed before changing to v9 OAP. Also, we should mention all(?) metrics are compatible with previous data format.
2. `scope-definitions.md` doc should be updated according. Also the change log should include it.
3. `mal.md` doc should be updated for new method parameters of downsampling functions.
4. `lal.md` doc should be updated for new `layer` method in extractor.
5. `overview.md` doc should be updated, **layer** becomes an important concept for the end user.
6. A FAQ should be added to cover details in (1). The change log should not be too long.
7. `Log-Data-Protocol.md` should be updated as new `layer` field is added as optional.

TBD, @mrproliu We need some helps on CLI side once this PR gets merged
1. e2e previous service verification should add `layer` check. We have done compatible upgrade `Server`(GraphQL) object to v2, so, old `getAllServices` or `search*` should support `layer` once the backend version is `v9`.(According to GraphQL service org.apache.skywalking.oap.query.graphql.resolver.Query.version).
2. New metadata query should be added, especially `listServices(layer: String!): [Service!]!`.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/MetadataQueryService.java
```diff
@@ -60,49 +58,61 @@ private IMetadataQueryDAO getMetadataQueryDAO() {
...
-    public Service searchBrowserService(final String serviceCode) throws IOException {
-        return getMetadataQueryDAO().searchService(NodeType.Browser, serviceCode);
-    }
+//    public Service searchService(final String serviceCode) throws IOException {
+//        return getMetadataQueryDAO().searchService(NodeType.Normal, serviceCode);
+//    }
+//
+//    public Service searchBrowserService(final String serviceCode) throws IOException {
+//        return getMetadataQueryDAO().searchService(NodeType.Browser, serviceCode);
+//    }
```
kezhenxu94, line 115: Remove these

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
```diff
@@ -143,55 +128,55 @@ public Service searchService(final NodeType nodeType, final String serviceCode)
...
         final String index =
             IndexController.LogicIndicesRegister.getPhysicalTableName(InstanceTraffic.INDEX_NAME);
-
-        final long minuteTimeBucket = TimeBucket.getMinuteTimeBucket(startTimestamp);
         final BoolQueryBuilder query =
             Query.bool()
-                 .must(Query.range(InstanceTraffic.LAST_PING_TIME_BUCKET).gte(minuteTimeBucket))
-                 .must(Query.term(InstanceTraffic.SERVICE_ID, serviceId));
-        final SearchBuilder search = Search.builder().query(query).size(queryMaxSize);
+                 .must(Query.term("_id", instanceId));
```
kezhenxu94, line 153: How can you find the instance by `_id`? I don't think we ever use `_id` (which is generated by ES) in our logic

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/type/Service.java
```diff
@@ -28,5 +30,7 @@
 public class Service {
     private String id;
     private String name;
+    private String shortName;
     private String group;
+    private List<String> layers = new ArrayList<>();
```
kezhenxu94, line 35: What about using `Set<String>`?

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/h2/dao/H2MetadataQueryDAO.java
```diff
@@ -146,125 +143,84 @@ public H2MetadataQueryDAO(JDBCHikariCPClient h2Client, int metadataQueryMaxSize)
...
-        sql.append("select * from ").append(EndpointTraffic.INDEX_NAME).append(" where ");
-        sql.append(EndpointTraffic.SERVICE_ID).append("=?");
-        condition.add(serviceId);
-        if (!Strings.isNullOrEmpty(keyword)) {
-            sql.append(" and ").append(EndpointTraffic.NAME).append(" like concat('%',?,'%') ");
-            condition.add(keyword);
-        }
-        sql.append(" limit ").append(limit);
+        sql.append("select * from ").append(InstanceTraffic.INDEX_NAME).append(" where ");
+        sql.append(H2TableInstaller.ID_COLUMN).append(" = ?");
```
kezhenxu94, line 168: I don't think we have stored the instance id as column `H2TableInstaller.ID_COLUMN`

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetadataQueryDAO.java
```diff
@@ -20,58 +20,37 @@
...
+    List<ServiceInstance> listInstances(final long startTimestamp, final long endTimestamp,
+                                        final String serviceId) throws IOException;
 
-    /**
-     * @param nodeType    describe which kind of node of Service
-     * @param serviceCode to literal match
-     * @return the service matching the given full name.
-     */
-    Service searchService(final NodeType nodeType, final String serviceCode) throws IOException;
+    ServiceInstance getInstance(final String instanceId) throws IOException;
```
kezhenxu94, line 47: Please add an E2E to verify this new API, from the implementation it seems this is not working

kezhenxu94 (changes requested):
Just did a first-round review

kezhenxu94:
> TBD, @mrproliu We need some helps on CLI side once this PR gets merged
> 
> 1. e2e previous service verification should add `layer` check. We have done compatible upgrade `Server`(GraphQL) object to v2, so, old `getAllServices` or `search*` should support `layer` once the backend version is `v9`.(According to GraphQL service org.apache.skywalking.oap.query.graphql.resolver.Query.version).
> 2. New metadata query should be added, especially `listServices(layer: String!): [Service!]!`.

cc @fgksgf

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/Layer.java
```diff
@@ -0,0 +1,105 @@
...
+
+    undefined(0),
```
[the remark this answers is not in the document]
wu-sheng: This reminds me,  I think should be upper case according to Oracle doc, https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetadataQueryDAO.java
wu-sheng: @wankai123 Please recheck and do manual GraphQL testing. 

@kezhenxu94 e2e is not working at this stage I am afraid. CLI is lack of env to verify its feature. We have to use manual tests for some parts.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/MetadataQueryService.java
```diff
@@ -49,8 +47,8 @@ private IMetadataQueryDAO getMetadataQueryDAO() {
         return metadataQueryDAO;
     }
 
-    public List<Service> getAllServices(final String group) throws IOException {
-        return getMetadataQueryDAO().getAllServices(group).stream()
+    public List<Service> listServices(final String layer, final String group) throws IOException {
+        return getMetadataQueryDAO().listServices(layer, group).stream()
```
wu-sheng, line 51: I know you built a `map`-based distinct/merge on DAO level, which should be moved here. And `distinct` is not working anymore as `layer` list added. You should do `reduce` here to merging according to service name.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
```diff
@@ -61,80 +60,66 @@ public MetadataQueryEsDAO(ElasticSearchClient client, int queryMaxSize) {
...
-    public Service searchService(final NodeType nodeType, final String serviceCode) throws IOException {
+    public Service findService(final String serviceId) throws IOException {
         final String index =
             IndexController.LogicIndicesRegister.getPhysicalTableName(ServiceTraffic.INDEX_NAME);
         final BoolQueryBuilder query =
             Query.bool()
-                 .must(Query.term(ServiceTraffic.NODE_TYPE, nodeType.value()))
-                 .must(Query.term(ServiceTraffic.NAME, serviceCode));
+                 .must(Query.term(ServiceTraffic.SERVICE_ID, serviceId));
         final SearchBuilder search = Search.builder().query(query).size(1);
```
wu-sheng, line 123: I am feeling `size(1)` is not correct here anymore. We could have multiple layers(meaning multiple records per service) now.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
wu-sheng: I think we don't use `ES` generated ID. I haven't verified this method, but AFAIK, `_id` of index is really generally by traffic and metric entities, they have `id()` method for this. Such as

```java
    @Override
    public InsertRequest prepareBatchInsert(Model model, Metrics metrics) {
        Map<String, Object> builder =
            IndexController.INSTANCE.appendMetricTableColumn(model, storageBuilder.entity2Storage(metrics));
        String modelName = TimeSeriesUtils.writeIndexName(model, metrics.getTimeBucket());
        String id = IndexController.INSTANCE.generateDocId(model, metrics.id());
        return getClient().prepareInsert(modelName, id, builder);
    }
```

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/h2/dao/H2MetadataQueryDAO.java
wu-sheng: I guess similar to https://github.com/apache/skywalking/pull/8367#discussion_r776910074.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
kezhenxu94: OK make sense to me, we don't use ID generated by ES, we specify them when inserting documents.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/Layer.java
wankai123: Added

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/service/ServiceTraffic.java
wankai123: Added

wu-sheng:
The CI and e2e passed.Please follow other review comments. Once you are done, please ping me.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetadataQueryDAO.java
wankai123: Manual tested, it works fine. Once CLI supported the new query, I'll add the new verify.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/MetadataQueryService.java
```diff
@@ -117,4 +97,19 @@ public EndpointInfo getEndpointInfo(final String endpointId) throws IOException
...
+                                                        if (service.getGroup() == null) {
+                                                            service.setGroup(Const.EMPTY_STRING);
+                                                        }
+                                                    })
+                                                    .collect(Collectors.toMap(Service::getName, service -> service,
+                                                                              (s1, s2) -> {
+                                                                                  s1.getLayers().addAll(s2.getLayers());
+                                                                                  return s1;
+                                                                              }
+                                                    )).values());
```
wu-sheng, line 113:
```suggestion
        return new ArrayList<>(
            services.stream()
                    .peek(service -> {
                        if (service.getGroup() == null) {
                            service.setGroup(Const.EMPTY_STRING);
                        }
                    })
                    .collect(
                        Collectors.toMap(Service::getName, service -> service,
                                         (s1, s2) -> {
                                             s1.getLayers().addAll(s2.getLayers());
                                             return s1;
                                         }
                        )).values());
    };
```

This format seems more readable, although the current one could pass the style check.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/Layer.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.apache.skywalking.oap.server.core.UnexpectedException;
+
+/**
+ * Layer represents an abstract framework in the computer science, such as operation system(OS_LINUX layer), Kubernetes(k8s layer)
+ */
+public enum Layer {
+
+    UNDEFINED(0),
```
wu-sheng, line 31: Some values of this enum have comments, others are not.
Please add comments to all of them.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/Layer.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    public int value() {
+        return value;
+    }
+
+    public static Layer valueOf(int value) {
+        if (DICTIONARY == null) {
+            DICTIONARY = Arrays.stream(Layer.values()).collect(Collectors.toMap(Layer::value, layer -> layer));
+        }
```
wu-sheng, line 93: This map could be built inside the constructor, which could initial the how things in the booting stage. The current could waste the performance in the concurrency case, even cause NPE as DICTIONARY is not `AtomicRef` and `final`.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/Layer.java
```diff
@@ -0,0 +1,100 @@
...
+     * Database conjectured by client side plugin
+     */
+    VIRTUAL_DATABASE(14),
+    /**
+     * MQ conjectured by client side plugin
+     */
+    VIRTUAL_MQ(15);
+
+    private final int value;
+    private static Map<Integer, Layer> DICTIONARY;
```
wu-sheng, line 80:
```suggestion
    private static final Map<Integer, Layer> DICTIONARY = new HashMap();
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/Layer.java
wankai123: It seems not a good way to use a `dictionary` for mapping. Should we roll back to the switch statement?
@wu-sheng @kezhenxu94 
https://github.com/apache/skywalking/pull/8367#discussion_r776760679

wankai123: It seems not a good way to use `dictionary` for mapping. Should we roll back to the switch statement?
@wu-sheng @kezhenxu94 https://github.com/apache/skywalking/pull/8367#discussion_r776760679

wu-sheng: Why it's not good? I just mean you should not initial it in this way.

wankai123: Enum constructor forbids to access static member, the static block could work, I'll try this way.

wu-sheng:
Codes seem good to me. Tests are passed. Let's update the doc before merging.

wu-sheng:
Zhenxu has some personal stuff for the next 2 weeks. @wankai123 Please open another pull request in order to make sure I can merge this, as I don't want to ping him these days.

wankai123:
Submit a new PR for this

PR #8397, opened 2022-01-09
NPE fix for IndexController#getPhysicalTableName

BFergerson:
`Optional.of().orElse()` doesn't make sense. I got an NPE when I searched for a missing table name. I'm not sure if this is the correct fix. Maybe the NPE is expected? If so, the `orElse` should be removed then as it's pointless.


wu-sheng (approved):
LGTM. Thanks for the fixing.

PR #8398, opened 2022-01-10
`AnalyzerModuleConfig` code improvement.

dkkb:
Change the type of `virtualPeers` to `Set<Integer>` and and `final` to `noUpstreamRealAddressAgents` field.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,7 @@ Release Notes.
 * Upgrade H2 version to 2.0.202 to fix CVE-2021-23463.
 * Extend column name override mechanism working for `ValueColumnMetadata`.
 * Introduce new concept `Layer` and removed `NodeType`. More details refer to [v9-version-upgrade](https://skywalking.apache.org/docs/main/latest/en/faq/v9-version-upgrade/).
+* `AnalyzerModuleConfig` code improvement.
 
 #### UI
 
--- a/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/AnalyzerModuleConfig.java
+++ b/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/AnalyzerModuleConfig.java
@@ -30,8 +30,9 @@ import org.apache.skywalking.oap.server.analyzer.provider.trace.parser.listener.
 import org.apache.skywalking.oap.server.core.Const;
 import org.apache.skywalking.oap.server.library.module.ModuleConfig;
 
-import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import static org.apache.skywalking.oap.server.analyzer.provider.trace.parser.listener.strategy.SegmentStatusStrategy.FROM_SPAN_STATUS;
 
@@ -50,7 +51,7 @@ public class AnalyzerModuleConfig extends ModuleConfig {
      * Read component-libraries.yml for more details.
      */
     @Getter
-    private String noUpstreamRealAddressAgents = Const.EMPTY_STRING;
+    private final String noUpstreamRealAddressAgents = Const.EMPTY_STRING;
     /**
      * The threshold used to check the slow database access. Unit, millisecond.
      */
@@ -111,7 +112,7 @@ public class AnalyzerModuleConfig extends ModuleConfig {
     @Getter
     private String segmentStatusAnalysisStrategy = FROM_SPAN_STATUS.name();
 
-    private List<Integer> virtualPeers;
+    private Set<Integer> virtualPeers;
 
     /**
      * @param componentId of the exit span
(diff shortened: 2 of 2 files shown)


oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/AnalyzerModuleConfig.java
```diff
@@ -50,7 +51,7 @@
      * Read component-libraries.yml for more details.
      */
     @Getter
-    private String noUpstreamRealAddressAgents = Const.EMPTY_STRING;
+    private final String noUpstreamRealAddressAgents = Const.EMPTY_STRING;
```
wu-sheng, line 54: This value is set by config and system env, so, can't be final.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/AnalyzerModuleConfig.java
```diff
@@ -111,15 +112,15 @@
     @Getter
     private String segmentStatusAnalysisStrategy = FROM_SPAN_STATUS.name();
 
-    private List<Integer> virtualPeers;
+    private Set<Integer> virtualPeers;
```
wu-sheng, line 115: This list is very short and nearly never used, as we usually have all agents for all proxies unless it is a private implemented proxy. Why do you think we need to improve this?

CHANGES.md
```diff
@@ -23,6 +23,7 @@ Release Notes.
 * Upgrade H2 version to 2.0.202 to fix CVE-2021-23463.
 * Extend column name override mechanism working for `ValueColumnMetadata`.
 * Introduce new concept `Layer` and removed `NodeType`. More details refer to [v9-version-upgrade](https://skywalking.apache.org/docs/main/latest/en/faq/v9-version-upgrade/).
+* `AnalyzerModuleConfig` code improvement.
```
wu-sheng, line 26: This is not a good changelog, no one can't understand what is a code improvement.

dkkb: Got it.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/AnalyzerModuleConfig.java
dkkb: [[        analysisListeners.forEach(listener -> {](https://github.com/apache/skywalking/blob/f3b567160ce61675cb692c3417101162d67093de/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/TraceAnalyzer.java#L79)](https://github.com/apache/skywalking/blob/f3b567160ce61675cb692c3417101162d67093de/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/TraceAnalyzer.java#L78-L84)

I think it is worthwhile to modify this field to 'Set' in the preceding code because it can be used in a 'foreach' loop. And all we want to do is make sure `componentId` is included in this `Set`, so `Set` would be ideal.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/AnalyzerModuleConfig.java
dkkb: https://github.com/apache/skywalking/blob/1c0fc982b2dc61ca7e452142df7f9846e867d71d/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/AnalyzerModuleConfig.java#L52-L53

🤔 It seems that we don't modify this filed, and IDE warns us about it. Now I'm not sure. 😂

wu-sheng: It is not changed directly through codes, it is changed by reflect API of OAP core.

I am checking why you want to change this.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/AnalyzerModuleConfig.java
wu-sheng: This list in 99% of cases is empty. That is why I am asking. HashSet is not always better than List, especially when we face a small and mostly empty list.
That is why I am asking, we don't encourage you to change codes by following general idea w/o real use cases.
SkyWalking is a very complex project, some codes are tricky, even hijacking due to some reason of history, some are real bugs not found before. 
I want to fully understand the background of changing this.

PR #8419, opened 2022-01-12
Add source layer and dest layer to relation

arugal:

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/ServiceRelation.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/ServiceRelation.java
@@ -20,6 +20,7 @@ package org.apache.skywalking.oap.server.core.source;
 
 import lombok.Getter;
 import lombok.Setter;
+import org.apache.skywalking.oap.server.core.analysis.Layer;
 import org.apache.skywalking.oap.server.library.util.StringUtil;
 import org.apache.skywalking.oap.server.core.analysis.IDManager;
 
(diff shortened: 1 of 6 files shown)


wu-sheng (approved):

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/ServiceRelation.java
```diff
@@ -61,11 +62,17 @@ public String getEntityId() {
...
+    private Layer sourceLayer;
+    @Getter
     private String destServiceId;
     @Getter
     @Setter
     @ScopeDefaultColumn.DefinedByField(columnName = "dest_name", requireDynamicActive = true)
     private String destServiceName;
+    @Getter
+    @Setter
+    private Layer destLayer;
```
wankai123, line 75: Should relation needs layer?  if so the relations here: https://github.com/apache/skywalking/blob/master/oap-server/server-receiver-plugin/skywalking-mesh-receiver-plugin/src/main/java/org/apache/skywalking/aop/server/receiver/mesh/TelemetryDataDispatcher.java
are needs to set to `Layer.mesh` @wu-sheng

wu-sheng: Make sense. Is MAL having required change too?

wu-sheng:
@wankai123 Could you submit another pull request to fix MAL side and ALS side?

wankai123:
> @wankai123 Could you submit another pull request to fix MAL side and ALS side?

:ok_hand:

PR #8487, opened 2022-01-29
add plugin for mysqlclient

katelei6:


wu-sheng:
I think you missed the mapping. You need to map it to mysql server, right? otherwise, you can see MySQL server on the topology.

wu-sheng (approved):
I added the missing pieces.

Issue #8477, opened 2022-01-26
[Feature][Breaking Change] Adopt UI configuration APIs of v9 core


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

We need to adopt the latest protocol update, https://github.com/apache/skywalking-query-protocol/pull/70

1. Remove the existing templates, as this time, we need to set up all templates from the beginning.
1. In the OAP, we need to add a configuration flag(in the YAML), to disable all dashboard update operations(`addTemplate`/`changeTemplate`). Because we don't provide tenancy and authentication, in the demo env or some users' env, we don't want default configurations could be changed unexpectedly.

**From this point, the RocketBot UI is not compatible anymore**. Please relink the submodule to the new UI repository.  https://github.com/apache/skywalking-booster-ui

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wu-sheng:
FYI @Fine0830

wu-sheng:
@mrproliu CLI would be affected, we need an update after the merge. We may need a backend version detecting mechanism this time.

wu-sheng:
Mostly done. Will release soon.

PR #8534, opened 2022-02-11
[OAL] Keep the endpoint avg resp time meter name the same with others scope.

JaredTan95:
This is a minor polish.
So far, we have `cpm` , `sla`, `percentile`, `resp_time` for `service`, `instance`, `database` scope, but for `endpoint`, `endpoint_avg` not keep the same format of `_resp_time`.

And, we also need to update the query parameter in `UI` query/repo.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -56,7 +56,7 @@ Release Notes.
 * Add OpenFunction component ID(5013).
 * Expose configuration `responseTimeout` of ES client.
 * Support datasource metric analysis.
-* Keep the endpoint avg resp time meter name the same with others scope. (This may break 3rd party integration)
+* [Break Change] Keep the endpoint avg resp time meter name the same with others scope. (This may break 3rd party integration and existing alarm rule settings)
 
 #### UI
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -55,6 +55,7 @@ Release Notes.
 * Add Guava EventBus component ID(123) of Java agent.
 * Add OpenFunction component ID(5013).
 * Expose configuration `responseTimeout` of ES client.
+* Keep the endpoint avg resp time meter name the same with others scope.
 
 #### UI
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -55,7 +55,7 @@ Release Notes.
 * Add Guava EventBus component ID(123) of Java agent.
 * Add OpenFunction component ID(5013).
 * Expose configuration `responseTimeout` of ES client.
-* Keep the endpoint avg resp time meter name the same with others scope.
+* Keep the endpoint avg resp time meter name the same with others scope. (This may break 3rd party integration)
 
 #### UI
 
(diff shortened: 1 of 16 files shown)


wu-sheng:
Why do we suddenly change this? I am fine but just curious.

wu-sheng:
This could break 3rd party integration, so, I want to know why.

JaredTan95:
> Why do we suddenly change this? I am fine but just curious.

In order to make the user experience better, for example, when users configure alarm rules, the avg response time of the endpoint can be configured as naturally as the avg response time of the service/instance, rather than making endpoint avg unique. It looks a bit weird, and users will think that there is something specially designed in it?

JaredTan95:
> This could break 3rd party integration, so, I want to know why.

If it involves large changes or coordination troubles, we can still use the original name

wu-sheng:
> > Why do we suddenly change this? I am fine but just curious.
> 
> 
> 
> In order to make the user experience better, for example, when users configure alarm rules, the avg response time of the endpoint can be configured as naturally as the avg response time of the service/instance, rather than making `endpoint_avg` unique. It looks a bit weird, and users will think that there is something specially designed in it?

No specific reason. I believe this is a mistake. If we plan to fix, let us be clear this is a break change in the change log, and search all docs, whether OAL and alarm docs affected by this break.

JaredTan95:
> > > Why do we suddenly change this? I am fine but just curious.
> > 
> > 
> > In order to make the user experience better, for example, when users configure alarm rules, the avg response time of the endpoint can be configured as naturally as the avg response time of the service/instance, rather than making `endpoint_avg` unique. It looks a bit weird, and users will think that there is something specially designed in it?
> 
> No specific reason. I believe this is a mistake. If we plan to fix, let us be clear this is a break change in the change log, and search all docs, whether OAL and alarm docs affected by this break.

Okay

wu-sheng:
Waiting for all e2e passed.

JaredTan95:
> I noticed `endpoint_avg` exists in the GraphQL protocol files(comments), `metrics.graphqls` and `metrics-v2.graphqls`. Could you fix them as well?

ok

wu-sheng:
Please update the submodule and check whether there is UT failing.

JaredTan95:
CI failed looks like timeout, and I tested ok locally.

wu-sheng:
Oh.. [ui-configuration.graphqls](https://github.com/apache/skywalking-query-protocol/compare/d6e294735ab43b2fa9e70b11c91d85268591dec0...3dd9ea164a6699d9e11678c9c039c4d999b6546b#diff-6aff06ba444cee4695a7b03480c317f756fc406afe6beebe4eb4e76cc35a919a) will be merged, I think we can't sync the submodule for now. 

https://github.com/apache/skywalking/issues/8477 is only going to be taken care of by @wankai123 after booster UI is ready.

wu-sheng:
We need a submodule rollback, sorry, I forgot the breaking change.

wu-sheng (approved):

wu-sheng (approved):

PR #8547, opened 2022-02-15
Support all metrics from MAL engine in alarm core

wu-sheng:

All meter objects generated in MeterSystem don't implement `WithMetadata`, which could make the alarm kernel ignore all of them. So, I am now adding `Meter` as the root of all MeterFunction classes, and injecting the default implementations to adopt the alarm kernel.

I am asking a community user to test, and am not sure whether this is good enough.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -58,6 +58,7 @@ Release Notes.
 * Support datasource metric analysis.
 * [Break Change] Keep the endpoint avg resp time meter name the same with others scope. (This may break 3rd party integration and existing alarm rule settings)
 * Add Python FastApi component ID(7014).
+* Support all metrics from MAL engine in alarm core, including Prometheus, OC receiver, meter receiver.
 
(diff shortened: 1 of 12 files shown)


wu-sheng:
@JaredTan95 @mrproliu Could you try to test this locally?

kezhenxu94 (approved):

bbcui:
Do we need to add code to handle the new type here?
https://github.com/apache/skywalking/blob/master/oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/NotifyHandler.java#L72

wu-sheng:
No, we don't. The scope exists in meter system. I have forwarded it in metadata assembling.

mrproliu (approved):
I have checked both meter and agent alarms, they work well.

wankai123 (approved):

PR #8549, opened 2022-02-16
Allow updating non-metrics templates when structure changed

kezhenxu94:
We have template structures changes in #8367 but if we upgrade from an older storage, the non-metrics templates (e.g. `service_traffic`) won't get updated.


wu-sheng (approved):

wu-sheng:
I want to add a notice this although I am totally fine about this change.

This template update was introduced due to metric indices merging, not for forward compatibility. So, even we open this feature to all indices, some break changes are still there, such as [v9 upgrade doc](https://skywalking.apache.org/docs/main/latest/en/faq/v9-version-upgrade/#v9-upgrade) described. 

This is a nice-to-have thing, but legacy data will still be an issue for newer kernal.

PR #8564, opened 2022-02-20
Support ElasticSearch 8 and add it into E2E tests

kezhenxu94:

### DIFF
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
@@ -200,12 +200,12 @@ public class StorageEsInstaller extends ModelInstaller {
     protected Mappings createMapping(Model model) {
         Map<String, Object> properties = new HashMap<>();
         for (ModelColumn columnDefine : model.getColumns()) {
+            final String type = columnTypeEsMapping.transform(columnDefine.getType(), columnDefine.getGenericType());
             if (columnDefine.isMatchQuery()) {
                 String matchCName = MatchCNameBuilder.INSTANCE.build(columnDefine.getColumnName().getName());
 
                 Map<String, Object> originalColumn = new HashMap<>();
-                originalColumn.put(
-                    "type", columnTypeEsMapping.transform(columnDefine.getType(), columnDefine.getGenericType()));
+                originalColumn.put("type", type);
                 originalColumn.put("copy_to", matchCName);
                 properties.put(columnDefine.getColumnName().getName(), originalColumn);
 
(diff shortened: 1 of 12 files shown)


wu-sheng:
`backend-storage.md` should be updated if tests passed.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -215,9 +215,9 @@ protected Mappings createMapping(Model model) {
                 properties.put(matchCName, matchColumn);
             } else {
                 Map<String, Object> column = new HashMap<>();
-                column.put(
-                    "type", columnTypeEsMapping.transform(columnDefine.getType(), columnDefine.getGenericType()));
-                if (columnDefine.isStorageOnly()) {
+                column.put("type", type);
+                // no index parameter is allowed for binary type, since ES 8.0
```
wu-sheng, line 219: Curiously, What binary fields are we setting `parameter`? And which parameters did we set?

kezhenxu94: Parameter name is "index", examples are "tags"

kezhenxu94: https://github.com/apache/skywalking/blob/f7bcb76784f968b865f31c7abfbd7c3a33945986/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/alarm/AlarmRecord.java#L80

wu-sheng: Oh, I see. It actually should add `storageOnly = true`, this is also a code bug.

kezhenxu94: > Oh, I see. It actually should add `storageOnly = true`, this is also a code bug.

I can fix it, but FYI, regardless of that even we set `index=false`, it's still not allowed in ES8 to have parameter `index` in binary type.

kezhenxu94: > Oh, I see. It actually should add `storageOnly = true`, this is also a code bug.

Fixed in this patch.

wu-sheng (approved):

PR #8573, opened 2022-02-22
Remove inapplicable example for enforcer / dropper sampler in LAL doc

kezhenxu94:


wu-sheng (approved):
We may consider 8.8.x 8.9.x doc update as we have the branches. What do you think?

kezhenxu94:
> We may consider 8.8.x 8.9.x doc update as we have the branches. What do you think?

Sure

PR #8585, opened 2022-02-24
Add istio 1.13.1 to test matrix for verification

kezhenxu94:
In case it breaks...


kezhenxu94:
Istio E2E tests don't run for some time and need some adjustments now...

wu-sheng:
Thanks for catching this. Let's fix this

wu-sheng (approved):

PR #8590, opened 2022-02-24
Publish e2e-service-* images to GitHub container registry

### LINKED ISSUE #8582: [Feature] Publish e2e-service provider and consumer to GitHub container registry
### Search before asking

### Description

Publish provider and consumer images (https://github.com/apache/skywalking/tree/master/test/e2e-v2/java-test-service) to support banyandb E2E which we discuss in PR https://github.com/apache/skywalking-banyandb/pull/81

### Use case

Use `https://github.com/apache/skywalking/pkgs/container/skywalking%2Fe2e-service-*` as the prefix.

### Related issues

_No response_

### Code of Conduct

lujiajing1126:
This PR introduces support for `e2e-service-*` images build and publish.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #8582


docker/e2e-service-consumer/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM apache/skywalking-java-agent:8.9.0-java8
```
wu-sheng, line 1: I have concerned about this. e2e of main repo has assigned the commit id of agent.

Could you check it? I think we have this mechanism already.

wu-sheng:
@wankai123 Where do we build the test service images? I think we built them for testing with some scripts.

docker/e2e-service-consumer/Dockerfile
lujiajing1126: Yes. I've noticed this part. But it exists in the `env` file which is read by the `infra-e2e` toolkit.

Shall we introduce a build argument as a `TAG`?

lujiajing1126:
> @wankai123 Where do we build the test service images? I think we built them for testing with some scripts.

I understand we do not build test-service images but only mount the jar files into the java agent image.

wu-sheng:
Then if you like agent installed test services, this build should be in Java agent repository.
I thought you just build the mock service images without agent.

Would that be better? Mounting agent into service when needed(in the e2e) only.

lujiajing1126:
> Then if you like agent installed test services, this build should be in Java agent repository.

I agree. Then we can use the built image with `SW_AGENT_JAVA_COMMIT` in the main repo instead of using a mount strategy. I suppose `e2e-service`s rarely change, right?

> Would that be better? Mounting agent into service when needed(in the e2e) only.

If you meant using in `banyandb` repo, I prefer an all-in-one image 😃

wu-sheng:
What makes all in one necessary? AFAIK, last discussion is about banyandb subproject doesn't want to host Java codes. This PR moves aggressively.

If someone wants to use this pure service, they will face issue again, and we have to deal with similar issues repeatedly. Pure services are common in service mesh.

lujiajing1126:
> What makes all in one necessary? AFAIK, last discussion is about banyandb subproject doesn't want to host Java codes. This PR moves aggressively. 

Sure we can provide pure jre-based images without java agent. But in `banyandb` repo, we have to download a release version of java agent and then mount it as a volume.

> If someone wants to use this pure service, they will face issue again, and we have to deal with similar issues repeatedly. Pure services are common in service mesh.

Well. I don't think it would be a problem if we don't pursue a "minimal" image size. We can always use environment variable (e.g. `JAVA_TOOL_OPTIONS` to control whether or not we enable the java agent.

wu-sheng:
You are asking others could adopt but meanwhile, you could adopt too.
Let's not make things so complex.
Mounting in banyandb e2e seems not an issue. You should know, agent based e2e is only one case only. Banyandb should be verified in the mesh cases too.

lujiajing1126:
> You are asking others could adopt but meanwhile, you could adopt too. Let's not make things so complex. Mounting in banyandb e2e seems not an issue. You should know, agent based e2e is only one case only. Banyandb should be verified in the mesh cases too.

Switch to jre base image now.

Shall we use JDK8 and JDK17 as we did for OAP?

wu-sheng:
@lujiajing1126 Take a look at the release doc, https://skywalking.apache.org/docs/main/latest/en/guides/how-to-release/#publish-the-docker-images. At the latest step, we need to push the image to dockerhub, which is sharing the makefile you changed. I think you will make that step fails as no e2e-service repo there, and also the e2e-service images should not be uploaded. I would recommend you to build a new makefile task to do this separatedly.

@kezhenxu94 Need your suggestions here.

kezhenxu94:
> @lujiajing1126 Take a look at the release doc, https://skywalking.apache.org/docs/main/latest/en/guides/how-to-release/#publish-the-docker-images. At the latest step, we need to push the image to dockerhub, which is sharing the makefile you changed. I think you will make that step fails as no e2e-service repo there, and also the e2e-service images should not be uploaded. I would recommend you to build a new makefile task to do this separatedly.
> 
> @kezhenxu94 Need your suggestions here.

Yes agree, it's better to separate the build scripts for test images and product images. Let's create another Makefile in `test/` and use `make -C test push` to push to images.  (I'd not do that although we can include that `test/Makefile` in our main Makefile).

Also, as the test services are rarely changed, we may not want to trigger the push action in every commit and push so many identical images with different tags, we should consider adding a check in that workflow by using `.github/actions/skip/action.yml` to skip unnecessary builds/pushes.

wu-sheng:
> Also, as the test services are rarely changed, we may not want to trigger the push action in every commit and push so many identical images with different tags

+1, only `test/e2e-v2/java-test-service/e2e-service-provider` and `test/e2e-v2/java-test-service/e2e-service-consumer` changes seem reasonable to push the images.

lujiajing1126:
Thanks to both of you for the information. I will move the `Makefile` to the `test` dir and build a separate workflow to handle this.

.github/workflows/publish-docker-e2e-service.yaml
```diff
@@ -0,0 +1,65 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: publish-docker-e2e-service
+
+on:
+  push:
```
wu-sheng, line 20: `on/push` seems could make image push if someone created a branch in the main repo, like you are doing.

lujiajing1126: > `on/push` seems could make image push if someone created a branch in the main repo, like you are doing.

First, we have a `paths` keyword inside `push` as a filter.

Then, we also have a `if` condition check in the last step to prevent `docker push` from running.

.github/workflows/publish-docker-e2e-service.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - "test/e2e-v2/java-test-service/**"
+      - "test/Makefile"
+
+env:
+  SKIP_TEST: true
+  HUB: ghcr.io/apache/skywalking
+
+jobs:
+  build:
+    if: github.repository == 'apache/skywalking'
```
wu-sheng, line 31: I think it is better to move the branch `if statement` to here directly?

lujiajing1126: Moved. Then we can only verify this PR on the master branch?

wu-sheng: Yes, you could remove these two if and verify on your fork. That is why usually we don't encourage committers to use the main repo's branch, instead, fork repo is always better.

.github/workflows/publish-docker-e2e-service.yaml
kezhenxu94: Add `branches` is a good practice if you plan to run this workflow in master branch, people, if you want to test this PR, add the branch name of this PR to the `branches` and remove it before ready

.github/workflows/publish-docker-e2e-service.yaml
```diff
@@ -0,0 +1,61 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: publish-docker-e2e-service
+
+on:
+  push:
```
kezhenxu94, line 20:
```suggestion
  push:
    branches:
      - master
```

.github/workflows/publish-docker-e2e-service.yaml
```diff
@@ -0,0 +1,61 @@
...
+      - 'test/e2e-v2/java-test-service/**'
+      - 'test/Makefile'
+
+env:
+  SKIP_TEST: true
+  HUB: ghcr.io/apache/skywalking
+
+jobs:
+  build:
+    if: github.repository == 'apache/skywalking' && github.ref == 'refs/heads/master'
```
kezhenxu94, line 31: Why not use this style ?
image

.github/workflows/publish-docker-e2e-service.yaml
lujiajing1126: Fixed

wu-sheng:
Could you create a fork repo to test this CI task running? Then feedback with the uploaded images(fork repo's) and CI logs.

.github/workflows/publish-docker-e2e-service.yaml
```diff
@@ -0,0 +1,63 @@
...
+        with:
+          registry: ${{ env.HUB }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Build and push images
+        run: make -C test build.e2e-service docker.push-e2e-service
```
kezhenxu94, line 63: `make -C test docker.push-e2e-service` is not executable from my side

lujiajing1126: Yes. CI in fork version reported this problem. Trying again

lujiajing1126:
> Could you create a fork repo to test this CI task running? Then feedback with the uploaded images(fork repo's) and CI logs.

CI passed with the following report,

https://github.com/lujiajing1126/skywalking/runs/5331395496?check_suite_focus=true

And images are uploaded to the fork repo.

kezhenxu94 (approved):

wu-sheng:
You could try the image from here, @lujiajing1126 

- https://github.com/apache/skywalking/pkgs/container/skywalking%2Fe2e-service-consumer
- https://github.com/apache/skywalking/pkgs/container/skywalking%2Fe2e-service-provider

PR #8598, opened 2022-02-25
Upgrade Apache Parent version to 25

CalvinKirs:
1: We can see the difference in several versions here. https://maven.apache.org/pom-archives/asf-LATEST/#history
2: The docs plugin is set to default to Java 1.7, so I modified it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@ Release Notes.
 * Add E2E test under Java 17.
 * Upgrade protoc to 3.19.2.
 * Add Istio 1.13.1 to E2E test matrix for verification.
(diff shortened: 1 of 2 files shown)


wu-sheng (commented):
Could you update about what the difference between these two parent poms?

CalvinKirs:
> Could you update about what the difference between these two parent poms?

Thanks for the review, we can see the difference in several versions here. https://maven.apache.org/pom-archives/asf-LATEST/#history

wu-sheng:
Oh, I mean if we bump this up, we need to update changelogs(add section project, take an example of previous release), and links to the ASF change logs, or copy texts.

PR #8615, opened 2022-03-02
Set the length of event#parameters to 2000

### LINKED ISSUE #8608: [Bug] Length of column events.parameters is too small to hold "Start Java Application" event with long JAVA_OPTS
### Search before asking

### What happened

When start a Java application with skywalking javaagent, events are failed to insert into table 'events' because the length of column 'parameters' is 1024 only.
e.g. My application has a very-long "JAVA_OPTS" generated by the IDE, see the attachment

screenshot

1. extend the length of 'events.parameters' or
2. just truncate the input for walk-through

### How to reproduce

Start a Java app. with long JAVA_OPTS

### Anything else

_No response_

### Code of Conduct

mustangxu:
- [ ] Explain briefly why the bug exists and how to fix it.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #8608.


oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/test/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/mysql/MySQLTableInstallerTest.java
```diff
@@ -0,0 +1,40 @@
...
+    public void test() throws Exception {
+        new DefaultScopeDefine.Listener().notify(Event.class);
+        Model m = new StorageModels().add(Event.class, DefaultScopeDefine.EVENT,
+            new Storage(null, false, null), false);
+
+        m.getColumns().stream()
+            .filter(c -> Event.PARAMETERS.equals(c.getColumnName().getName()))
+            .findAny().ifPresent(c -> {
+                String sql = new TestTableInstaller().getColumn(c);
+//                System.out.println(sql);
```
wu-sheng, line 25: Please remove this.

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/test/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/mysql/MySQLTableInstallerTest.java
```diff
@@ -0,0 +1,40 @@
...
+                Assert.assertTrue("Should contains 'TEXT'",
+                    sql.contains("TEXT"));
```
wu-sheng, line 27: I think we have a type mapping method, don't have to test in a separate UT.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ModelInstaller.java
```diff
@@ -62,7 +62,7 @@ public void whenCreating(Model model) throws StorageException {
      * Installer implementation could use this API to request a column name replacement. This method delegates for
      * {@link ModelManipulator}.
      */
-    protected final void overrideColumnName(String columnName, String newName) {
+    protected void overrideColumnName(String columnName, String newName) {
```
wu-sheng, line 65: Don't recommend this change as UT seems not necessary.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/Event.java
```diff
@@ -99,7 +99,7 @@ protected String id0() {
     @Column(columnName = MESSAGE)
     private String message;
 
-    @Column(columnName = PARAMETERS, storageOnly = true, length = 1024)
+    @Column(columnName = PARAMETERS, storageOnly = true, length = 20000)
```
wu-sheng, line 102: I think we need to control length in serialization process too? Otherwise still could have over length.

wu-sheng (commented):
Comments added, and you should update change logs.

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/test/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/mysql/MySQLTableInstallerTest.java
wu-sheng: `MySQLTableInstaller#getColumn` is the source of column type. If you want to add a UT, add it for MySQLTableInstaller.

wu-sheng:
@mustangxu Are you going to follow the comments?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ModelInstaller.java
wu-sheng, line 65:
```suggestion
    protected final void overrideColumnName(String columnName, String newName) {
```

wu-sheng (approved):
Merging for now as the author has no reponse. I will send a pull request to truncate the parameter.

wu-sheng (approved):

PR #8617, opened 2022-03-03
Introduce the entity of Process Type

mrproliu:
### Introduce the entity of Process Type
1. The Process type entity could be created through the EBPF process service. It's at the next level of Service Instance. 
2. There is no complete E2E test yet, I will add it after I complete the Profiling function of the Rover project. 

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/IDManager.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/IDManager.java
@@ -20,6 +20,9 @@ package org.apache.skywalking.oap.server.core.analysis;
 
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
+
+import com.google.common.base.Charsets;
+import com.google.common.hash.Hashing;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
@@ -249,6 +252,21 @@ public class IDManager {
              */
             private final String destServiceId;
             private final String dest;
+        }
+    }
+
+    /**
+     * Process ID related functions.
+     */
+    public static class ProcessID {
+        /**
+         * @param instanceId built by {@link ServiceInstanceID#buildId(String, String)}
+         * @param name process name
+         * @return process id
+         */
+        public static String buildId(String instanceId, String name) {
+            return Hashing.sha256().newHasher().putString(String.format("%s_%s",
+                    name, instanceId), Charsets.UTF_8).hash().toString();
         }
     }
 
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/process/ProcessTraffic.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/process/ProcessTraffic.java
@@ -1,0 +1,226 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.oap.server.core.analysis.manual.process;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+import org.apache.skywalking.oap.server.core.Const;
+import org.apache.skywalking.oap.server.core.analysis.IDManager;
+import org.apache.skywalking.oap.server.core.analysis.Layer;
+import org.apache.skywalking.oap.server.core.analysis.MetricsExtension;
+import org.apache.skywalking.oap.server.core.analysis.Stream;
+import org.apache.skywalking.oap.server.core.analysis.metrics.Metrics;
+import org.apache.skywalking.oap.server.core.analysis.worker.MetricsStreamProcessor;
+import org.apache.skywalking.oap.server.core.remote.grpc.proto.RemoteData;
+import org.apache.skywalking.oap.server.core.storage.StorageHashMapBuilder;
+import org.apache.skywalking.oap.server.core.storage.annotation.Column;
+import org.apache.skywalking.oap.server.library.util.StringUtil;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.skywalking.oap.server.core.source.DefaultScopeDefine.PROCESS;
+
+@Stream(name = ProcessTraffic.INDEX_NAME, scopeId = PROCESS,
+    builder = ProcessTraffic.Builder.class, processor = MetricsStreamProcessor.class)
+@MetricsExtension(supportDownSampling = false, supportUpdate = true)
+@EqualsAndHashCode(of = {
+        "instanceId",
+        "name",
+})
+public class ProcessTraffic extends Metrics {
+    public static final String INDEX_NAME = "process_traffic";
+    public static final String SERVICE_ID = "service_id";
+    public static final String INSTANCE_ID = "instance_id";
+    public static final String NAME = "name";
+    public static final String LAYER = "layer";
+    public static final String AGENT_ID = "agent_id";
+    public static final String PROPERTIES = "properties";
+    public static final String LAST_PING_TIME_BUCKET = "last_ping";
+    public static final String DETECT_TYPE = "detect_type";
+
+    private static final Gson GSON = new Gson();
+
+    @Setter
+    @Getter
+    @Column(columnName = SERVICE_ID)
+    private String serviceId;
+
+    @Setter
+    @Getter
+    @Column(columnName = INSTANCE_ID)
+    private String instanceId;
+
+    @Setter
+    @Getter
+    @Column(columnName = NAME)
+    private String name;
+
+    @Setter
+    @Getter
+    @Column(columnName = LAYER)
+    private int layer = Layer.UNDEFINED.value();
+
+    @Setter
+    @Getter
+    @Column(columnName = LAST_PING_TIME_BUCKET)
+    private long lastPingTimestamp;
+
+    @Setter
+    @Getter
+    @Column(columnName = DETECT_TYPE)
+    private int detectType = ProcessDetectType.UNDEFINED.value();
+
+    @Setter
+    @Getter
+    @Column(columnName = AGENT_ID)
+    private String agentId;
+
+    @Setter
+    @Getter
+    @Column(columnName = PROPERTIES, storageOnly = true, length = 50000)
+    private JsonObject properties;
+
+    @Override
+    public boolean combine(Metrics metrics) {
+        final ProcessTraffic processTraffic = (ProcessTraffic) metrics;
+        this.lastPingTimestamp = processTraffic.getLastPingTimestamp();
+        if (processTraffic.getDetectType() > 0) {
+            this.setDetectType(processTraffic.getDetectType());
+        }
+        if (StringUtil.isNotBlank(processTraffic.getAgentId())) {
+            this.agentId = processTraffic.getAgentId();
+        }
+        if (processTraffic.getProperties() != null && processTraffic.getProperties().size() > 0) {
+            this.properties = processTraffic.getProperties();
+        }
+        if (processTraffic.getDetectType() > 0) {
+            this.detectType = processTraffic.getDetectType();
+        }
+        return true;
+    }
+
+    @Override
+    public int remoteHashCode() {
+        return this.hashCode();
+    }
+
+    @Override
+    public void deserialize(RemoteData remoteData) {
+        setServiceId(remoteData.getDataStrings(0));
+        setInstanceId(remoteData.getDataStrings(1));
+        setName(remoteData.getDataStrings(2));
+        setLayer(remoteData.getDataIntegers(0));
+        setAgentId(remoteData.getDataStrings(3));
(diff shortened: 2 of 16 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/process/ProcessTraffic.java
```diff
@@ -0,0 +1,226 @@
...
+    private long lastPingTimestamp;
+
+    @Setter
+    @Getter
+    @Column(columnName = DETECT_TYPE)
+    private int detectType = ProcessDetectType.UNDEFINED.value();
+
+    @Setter
+    @Getter
+    @Column(columnName = AGENT_ID)
```
wu-sheng, line 95: The default length is only 200, is this enough?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/process/ProcessTraffic.java
```diff
@@ -0,0 +1,226 @@
...
+    private static final Gson GSON = new Gson();
+
+    @Setter
+    @Getter
+    @Column(columnName = SERVICE_ID)
+    private String serviceId;
+
+    @Setter
+    @Getter
+    @Column(columnName = INSTANCE_ID)
```
wu-sheng, line 70: I think even in default, the length of instance ID is longer than 200. It is `base64(service name) + base64(instance name)`

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/process/ProcessTraffic.java
```diff
@@ -0,0 +1,226 @@
...
+    private String instanceId;
+
+    @Setter
+    @Getter
+    @Column(columnName = NAME)
```
wu-sheng, line 75: Same length concern.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/DefaultScopeDefine.java
```diff
@@ -101,6 +103,7 @@
     public static final String SERVICE_RELATION_CATALOG_NAME = "SERVICE_RELATION";
     public static final String SERVICE_INSTANCE_RELATION_CATALOG_NAME = "SERVICE_INSTANCE_RELATION";
     public static final String ENDPOINT_RELATION_CATALOG_NAME = "ENDPOINT_RELATION";
+    public static final String PROCESS_CATALOG_NAME = "PROCESS";
```
wu-sheng, line 106: You should update the process catalog, although we don't set the alert rules of process for now, but most likely we will, and also auto profiling should sent the notification(we need to extend alarm module until then)

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/process/ProcessTraffic.java
mrproliu: Refer to the [default name length in OAP configiguration](https://github.com/apache/skywalking/blob/ff1103ec1d2d7c6a1a896730cfe3024e1e99782f/oap-server/server-starter/src/main/resources/application.yml#L115-L117), It should be less than `200`. 
Maybe I could update the length to the `600`, After I calculate It could handle 200 lengths in service and instance name? I think it's enough. 
Also, I see some other metrics in OAP, they have `instanceId` field and they are using the default length(200), such as `LogRecord`, `NetworkAddressAlias` etc. Should we update too?

wu-sheng: I think we should update all to 600. This was missed for a long time. But, it is better to submit another separate pull request to fix those.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/process/ProcessTraffic.java
mrproliu: update to 500 length.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/process/ProcessTraffic.java
mrproliu: update the `instanceId` in `ProcessTraffic`. I will create another PR to fix other traffic after this PR merge.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/process/ProcessTraffic.java
mrproliu: update to 500.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/DefaultScopeDefine.java
mrproliu: Adapted to the alarm module. But the `ProcessTraffic` only works for the registration mechanism, don't need to implement the `WithMetadata` for the alarm module. Also, because we use a hash algorithm(sha512) to generate the process id, we can't analyze the metadata in the process, so I add a new Entity to appoint them.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/DefaultScopeDefine.java
```diff
@@ -309,6 +316,16 @@ public static boolean inEndpointRelationCatalog(int scopeId) {
         return ENDPOINT_RELATION_CATALOG.containsKey(scopeId);
     }
 
+    /**
+     * Check whether current service belongs process catalog
```
wu-sheng, line 320:
```suggestion
     * Check whether the given scope ID belongs process catalog
```
I think there are several similar typos in `in*Catalog` methods, please fix them too.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/Process.java
```diff
@@ -0,0 +1,88 @@
...
+import lombok.Getter;
+import lombok.Setter;
+import org.apache.skywalking.oap.server.core.analysis.IDManager;
+import org.apache.skywalking.oap.server.core.analysis.Layer;
+import org.apache.skywalking.oap.server.core.analysis.manual.process.ProcessDetectType;
+
+import static org.apache.skywalking.oap.server.core.source.DefaultScopeDefine.PROCESS;
+import static org.apache.skywalking.oap.server.core.source.DefaultScopeDefine.PROCESS_CATALOG_NAME;
+
+@ScopeDeclaration(id = PROCESS, name = "Process", catalog = PROCESS_CATALOG_NAME)
```
wu-sheng, line 31: I think we should not register `Process` in `PROCESS_CATALOG_NAME`. Because this doesn't represent traffic, it is just a registration. This should be more close to `ServiceMeta`?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/Process.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    public String getEntityId() {
+        if (entityId == null) {
+            entityId = IDManager.ProcessID.buildId(instanceId, name);
+        }
+        return entityId;
+    }
+
+    @Getter
+    @ScopeDefaultColumn.DefinedByField(columnName = "instance_id")
```
wu-sheng, line 50: Same as https://github.com/apache/skywalking/pull/8617/files#r818356930, this source is not designed for OAL analysis. So, don't need to set up `ScopeDefaultColumn.DefinedByField`, because you have and only will have a manual dispatcher, which is not sensitive about these annotations.

docs/en/concepts-and-designs/overview.md
```diff
@@ -11,14 +11,17 @@ In multi-language, continuously deployed environments, cloud native infrastructu
...
 Instance and Endpoint are used everywhere today, so it is worth defining their specific meanings in the context of SkyWalking:
 
 - **Service**. Represents a set/group of workloads which provide the same behaviours for incoming requests. You can define the service
   name when you are using instrument agents or SDKs. SkyWalking can also use the name you define in platforms such as Istio.
 - **Service Instance**. Each individual workload in the Service group is known as an instance. Like `pods` in Kubernetes, it 
   doesn't need to be a single OS process, however, if you are using instrument agents, an instance is actually a real OS process.
 - **Endpoint**. A path in a service for incoming requests, such as an HTTP URI path or a gRPC service class + method signature. 
+- **Process**. Each independently running OS process in a Service Instance is a process. In some scenarios, a Service Instance is
+  not a process. such as pods in Kubernetes, each pod could have multiple OS Process, the pod could be a Service Instance, and 
+  the OS Process in each pod is a Process.
```
wu-sheng, line 24:
```suggestion
- **Process**. An operating system process.  In some scenarios, a Service Instance is
  not a process, such as a pod Kubernetes could contain multiple processes.
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/Process.java
mrproliu: Deleted.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/Process.java
mrproliu: Yes, the `Process` is only registration, deleted the catalog declare.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/DefaultScopeDefine.java
mrproliu: Fixed.

wu-sheng (approved):
No config should be added for `configuration-vocabulary.md` doc. But I am wondering we may need to add later. Just a tip.

@kezhenxu94 Could you recheck?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/IDManager.java
```diff
@@ -252,6 +255,21 @@ public static EndpointRelationDefine analysisRelationId(String entityId) {
...
+     * Process ID related functions.
+     */
+    public static class ProcessID {
+        /**
+         * @param instanceId built by {@link ServiceInstanceID#buildId(String, String)}
+         * @param name process name
+         * @return process id
+         */
+        public static String buildId(String instanceId, String name) {
+            return Hashing.sha256().newHasher().putString(String.format("%s_%s",
```
kezhenxu94, line 268: Just curious, why do we choose a different encoding machanism `sha256` than the existing one `base64`?

oap-server/server-receiver-plugin/pom.xml
```diff
@@ -45,6 +45,7 @@
         <module>configuration-discovery-receiver-plugin</module>
         <module>skywalking-event-receiver-plugin</module>
         <module>skywalking-zabbix-receiver-plugin</module>
+        <module>skywalking-ebpf-receiver-plugin</module>
```
kezhenxu94, line 48: The module name looks too general, what about renaming to `skywalking-rover-receiver-plugin`?

oap-server/server-receiver-plugin/skywalking-ebpf-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/ebpf/module/EBPFReceiverModule.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ */
+
+package org.apache.skywalking.oap.server.receiver.ebpf.module;
+
+import org.apache.skywalking.oap.server.library.module.ModuleDefine;
+
+public class EBPFReceiverModule extends ModuleDefine {
+
+    public static final String NAME = "receiver-ebpf";
```
kezhenxu94, line 25: Same naming suggestion here, `receiver-rover`?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/IDManager.java
wu-sheng: Yes, we need to keep the length fixed. The instance ID is closing to 512 which is the limitation for ElasticSearch/OpenSearch. But process belongs to an instance, traditional base64 is not going to work well.

oap-server/server-receiver-plugin/pom.xml
wu-sheng: It is following protocol naming rule, https://github.com/apache/skywalking-data-collect-protocol/tree/master/ebpf/profiling

oap-server/server-receiver-plugin/skywalking-ebpf-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/ebpf/module/EBPFReceiverModule.java
wu-sheng: Same as above.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/process/ProcessTraffic.java
```diff
@@ -0,0 +1,226 @@
...
+        }
+        if (StringUtil.isNotBlank(processTraffic.getAgentId())) {
+            this.agentId = processTraffic.getAgentId();
+        }
+        if (processTraffic.getProperties() != null && processTraffic.getProperties().size() > 0) {
+            this.properties = processTraffic.getProperties();
+        }
+        if (processTraffic.getDetectType() > 0) {
+            this.detectType = processTraffic.getDetectType();
+        }
```
kezhenxu94, line 118: Duplicate code logic

```suggestion
        if (StringUtil.isNotBlank(processTraffic.getAgentId())) {
            this.agentId = processTraffic.getAgentId();
        }
        if (processTraffic.getProperties() != null && processTraffic.getProperties().size() > 0) {
            this.properties = processTraffic.getProperties();
        }
        if (processTraffic.getDetectType() > 0) {
            this.detectType = processTraffic.getDetectType();
        }
```

kezhenxu94 (approved):
Others LGTM

PR #8628, opened 2022-03-07
Support large service list query and replace deprecated API of Armeria

### LINKED ISSUE #8618: [Feature] Support large service list query
### Search before asking

### Description

In some large-scale deployments, the number of services could be over 10k, which could cause `[type=illegal_argument_exception, reason=Result window is too large, from + size must be less than or equal to: [10000] but was [17000]. ` error.  We should consider using `scroll api` to get service list in better performance.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #8618 .

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -82,6 +82,9 @@ Release Notes.
 * Introduce the entity of Process type.
 * Set the length of event#parameters to 2000.
 * Limit the length of Event#parameters.
+* Support large service list query by using ElasticSearch scrolling API, add `metadataQueryBatchSize` to configure scrolling page size.
+* Change default value of `metadataQueryMaxSize` from `5000` to `10000`
+* Replace deprecated Armeria API `BasicToken.of` with `AuthToken.ofBasic`.
 
 #### UI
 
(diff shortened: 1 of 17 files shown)


wu-sheng:
Should we change listInstance too? Just in case we face 5000+ pods per service?(I don't have real scenario).

wu-sheng:
Let's do this too. 10000 IPs seem acommon case as 256*256 IP for a subset, such as 172.0.*.*

CHANGES.md
```diff
@@ -82,6 +82,9 @@ Release Notes.
 * Introduce the entity of Process type.
 * Set the length of event#parameters to 2000.
 * Limit the length of Event#parameters.
+* Support large service list query by using ElasticSearch scrolling API, add `metadataQueryBatchSize` to configure scrolling page size.
```
wu-sheng, line 85:
```suggestion
* Support large service/instance/networkAddressAlias list query by using ElasticSearch scrolling API, add `metadataQueryBatchSize` to configure scrolling page size.
```

wu-sheng (approved):

PR #8649, opened 2022-03-10
Implement process metadata query protocol

mrproliu:
### Implement process metadata query protocol

I'll be submitting a PR for the BPF process profiling protocol recently, and after these are done, add a full E2E test.

GraphQL query:
### listProcesses:
listProcesses

### getProcess:
process-detail

### DIFF
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
@@ -29,6 +29,7 @@ import org.apache.skywalking.library.elasticsearch.requests.search.BoolQueryBuil
 import org.apache.skywalking.library.elasticsearch.requests.search.Query;
 import org.apache.skywalking.library.elasticsearch.requests.search.Search;
 import org.apache.skywalking.library.elasticsearch.requests.search.SearchBuilder;
+import org.apache.skywalking.library.elasticsearch.requests.search.SearchParams;
 import org.apache.skywalking.library.elasticsearch.response.search.SearchHit;
 import org.apache.skywalking.library.elasticsearch.response.search.SearchResponse;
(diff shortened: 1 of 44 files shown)


wu-sheng:
Could you add an issue to track CLI to support this?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
```diff
@@ -191,6 +195,37 @@ public ServiceInstance getInstance(final String instanceId) throws IOException {
...
+                IndexController.LogicIndicesRegister.getPhysicalTableName(ProcessTraffic.INDEX_NAME);
+
+        final BoolQueryBuilder query = Query.bool();
+        final SearchBuilder search = Search.builder().query(query).size(queryMaxSize);
+        if (StringUtil.isNotEmpty(serviceId)) {
+            query.must(Query.term(ProcessTraffic.SERVICE_ID, serviceId));
+        }
+        if (StringUtil.isNotEmpty(instanceId)) {
+            query.must(Query.term(ProcessTraffic.INSTANCE_ID, instanceId));
+        }
```
wu-sheng, line 210: I think serviceId and instanceId are both required.

mrproliu: In the [query comment](https://github.com/apache/skywalking-query-protocol/blob/master/metadata-v2.graphqls#L136-L137), I mentioned they only providing one field is enough. It's not necessary to provide two fields, also the instance id contains the appoint service id. why do we need to provide two?

wu-sheng: I see.

wankai123 (dismissed):
LGTM

wu-sheng (approved):

PR #8656, opened 2022-03-10
[Breaking Change] Deprecate `All` from OAL source.

wu-sheng:

Deprecate `All` from OAL source, according to latest layer concept, `All` as source has lost its meaning. We never would find a way to have scope=all metrics, as each layer is different from others.

@kezhenxu94 Event source is belong to `All`(a.k.a. default`) scope, because we don't set its `ScopeDeclaration#catalog`. Which scope or catalog should it belong?


kezhenxu94:
> @kezhenxu94 Event source is belong to All(a.k.a. default) scope, because we don't set its ScopeDeclaration#catalog`. Which scope or catalog should it belong?

I think it should belong to service scope.

wu-sheng:
@kezhenxu94 Done.

kezhenxu94 (approved):

wankai123 (approved):

wankai123 (approved):

PR #8664, opened 2022-03-11
Enhance DataCarrier#MultipleChannelsConsumer to add priority

wu-sheng:

* Enhance DataCarrier#MultipleChannelsConsumer to add priority for the channels, which makes OAP server has a better
  performance to activate all analyzers on default.
* Activate `receiver-otel#enabledOcRules` receiver with `k8s-cluster,k8s-node,k8s-service,oap,vm` rules on default.
* Activate `satellite,spring-sleuth` for `agent-analyzer#meterAnalyzerActiveFiles`  on default.
* Activate `receiver-zabbix` receiver with `agent` rule on default.

@wankai123 @mrproliu @kezhenxu94 As a result, many rules are being activated on default, also for the latest UI, all metrics would be visible on the configuration page(FYI @Fine0830). 
For the end-users, they don't need to change deployment, reboot the OAP server, or face confused status.

```
* Priority determines the consuming strategy. On default every period consumer thread loops all groups trying
* to fetch the data from queue, if the queue only contains few elements, it is too expensive to consume every
* time.
*
* if 'size of last fetched data' > 0
*
* priority = 'size of last fetched data' * 100 / {@link Channels#size()} * {@link Channels#getChannelSize()}
*
* else
*
* priority = priority / 2
*
* Meaning, priority is the load factor of {@link #channels}
*
* After consuming loop, priority = (priority of current loop + priority of last loop) / 2.
*
* If priority > 50, consuming happens in next loop, otherwise, priority += 10, and wait until priority > 50. In
* worth case, for a low traffic group, consuming happens in 1/10.
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,15 +18,15 @@ Release Notes.
 * Add Istio 1.13.1 to E2E test matrix for verification.
 * Upgrade Apache parent pom version to 25.
 * Use the plugin version defined by the Apache maven parent.
-  * Upgrade maven-dependency-plugin to 3.2.0.
-  * Upgrade  maven-assembly-plugin to 3.3.0.
-  * Upgrade  maven-failsafe-plugin to 2.22.2.
-  * Upgrade  maven-surefire-plugin to 2.22.2.
-  * Upgrade  maven-jar-plugin to 3.2.2.
-  * Upgrade  maven-enforcer-plugin to 3.0.0.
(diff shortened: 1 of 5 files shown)


kezhenxu94 (approved):

mrproliu (approved):

wankai123 (approved):
lgtm

PR #8673, opened 2022-03-14
Migrate the HTTP server from Jetty to Armeria

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number.  related https://github.com/apache/skywalking/discussions/8666.


CHANGES.md
```diff
@@ -105,6 +105,10 @@ Release Notes.
 * Activate `receiver-otel#enabledOcRules` receiver with `k8s-cluster,k8s-node,k8s-service,oap,vm` rules on default.
 * Activate `satellite,spring-sleuth` for `agent-analyzer#meterAnalyzerActiveFiles`  on default.
 * Activate `receiver-zabbix` receiver with `agent` rule on default.
+* Replace Http server (GraphQL, agent http protocol) from Jetty with Armeria.
+* [Breaking Change] Remove configuration `restAcceptorPriorityDelta` (env var: `SW_RECEIVER_SHARING_JETTY_DELTA`, `SW_CORE_REST_JETTY_DELTA`).
+* [Breaking Change] Remove configuration `graphql/path` (env var: `SW_QUERY_GRAPHQL_PATH`).
+* [Breaking Change] Remove configuration `restMinThreads` (env var: `SW_CORE_REST_JETTY_MIN_THREADS`, `SW_RECEIVER_SHARING_JETTY_MIN_THREADS`).
```
wu-sheng, line 111: With removing all these, is there any thread number or other key configurations of Armeria server should be exposed?

wu-sheng: I noticed MAX_THREAD has been kept.

CHANGES.md
```diff
@@ -105,6 +105,10 @@ Release Notes.
 * Activate `receiver-otel#enabledOcRules` receiver with `k8s-cluster,k8s-node,k8s-service,oap,vm` rules on default.
 * Activate `satellite,spring-sleuth` for `agent-analyzer#meterAnalyzerActiveFiles`  on default.
 * Activate `receiver-zabbix` receiver with `agent` rule on default.
+* Replace Http server (GraphQL, agent http protocol) from Jetty with Armeria.
```
wu-sheng, line 108:
```suggestion
* Replace HTTP server (GraphQL, agent HTTP protocol) from Jetty with Armeria.
```

CHANGES.md
wu-sheng: Could you make a comparison about size reducing between two binary tars?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModule.java
```diff
@@ -122,7 +122,7 @@ private void addQueryService(List<Class> classes) {
 
     private void addServerInterface(List<Class> classes) {
         classes.add(GRPCHandlerRegister.class);
-        classes.add(JettyHandlerRegister.class);
+        classes.add(HttpHandlerRegister.class);
```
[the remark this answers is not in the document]
kezhenxu94: Done, there are still some `Http***` from third-party classes.

CHANGES.md
kezhenxu94: There is no much reduction in package size, they are about ~132M the same

wu-sheng (approved):

PR #8679, opened 2022-03-15
Add storage column attribute `indexOnly`

wankai123:
### Add storage column attribute `indexOnly`

Add storage column attribute `indexOnly`, support ElasticSearch only index, and not store some fields.
Add `indexOnly=true` to `SegmentRecord.tags`, `AlarmRecord.tags`, `AbstractLogRecord.tags`, to reduce unnecessary storage.


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -148,7 +149,7 @@ private void createTimeSeriesTable(Model model) throws StorageException {
             if (shouldUpdateTemplate) {
                 structures.putStructure(tableName, mapping);
                 boolean isAcknowledged = esClient.createOrUpdateTemplate(
-                    tableName, settings, structures.getMapping(tableName), config.getIndexTemplateOrder());
+                    tableName, settings, mapping, config.getIndexTemplateOrder());
```
wankai123, line 152: I'm not sure why use `structures.getMapping(tableName)` before. In that logic, the updated template would only add new mapping property but wouldn't update/delete the existing property.

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/response/Mappings.java
```diff
@@ -42,4 +44,9 @@
     @Getter
     @Setter
     private Map<String, Object> properties = new HashMap<>();
+
+    @JsonProperty("_source")
+    @Getter
+    @Setter
+    private Map<String, Object> source = new HashMap<>();
```
kezhenxu94, line 51: This is unused?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/segment/SegmentRecord.java
```diff
@@ -91,7 +90,7 @@
     private byte[] dataBinary;
     @Setter
     @Getter
-    @Column(columnName = TAGS)
+    @Column(columnName = TAGS, indexOnly = true)
```
kezhenxu94, line 93: If the tags are only indexed, how can they be displayed in the UI, or we don't display these ever?

wankai123: All segment info include `tags` are stored in `dataBinary` already.

wankai123:
hold, more codes need to adjust.

wankai123:
@kezhenxu94 @wu-sheng could review now

oap-server/server-library/library-elasticsearch-client/src/test/java/org/apache/skywalking/library/elasticsearch/ITElasticSearchTest.java
```diff
@@ -177,13 +189,18 @@ public void testDoc() {
...
+        sourceConf.getExcludes().add("key3");
         assertTrue(
             client.index().create(
                 index,
                 Mappings.builder()
                         .type("type")
                         .properties(ImmutableMap.of("key1", ImmutableMap.of("type", "keyword")))
-                        .properties(ImmutableMap.of("key2", ImmutableMap.of("type", "keyword")))
+                        .properties(ImmutableMap.of("key2", ImmutableMap.of("type", "keyword"), "key3",
+                                                    ImmutableMap.of("type", "keyword")
```
kezhenxu94, line 201:
```suggestion
                        .properties(ImmutableMap.of("key2", ImmutableMap.of("type", "keyword"),
                                                    "key3", ImmutableMap.of("type", "keyword")
```

oap-server/server-library/library-elasticsearch-client/src/test/java/org/apache/skywalking/library/elasticsearch/ITElasticSearchTest.java
```diff
@@ -209,6 +226,15 @@ public void testSearch() {
...
         });
+        //test indexOnly
+        await().atMost(Duration.ONE_MINUTE).untilAsserted(() -> {
+            SearchResponse response = client.search(
+                Search.builder().query(Query.bool().must(Query.term("key3", "val3"))).build()
+            );
+            assertEquals(1, response.getHits().getTotal());
+            assertEquals("val1", response.getHits().iterator().next().getSource().get("key1"));
+            assertNull(response.getHits().iterator().next().getSource().get("key3"));
+        });
```
kezhenxu94, line 237: Prefer an error message than comments so that when it failed we can see the reason from the logs directly

```suggestion
        await().atMost(Duration.ONE_MINUTE).untilAsserted(() -> {
            SearchResponse response = client.search(
                Search.builder().query(Query.bool().must(Query.term("key3", "val3"))).build()
            );
            assertEquals(1, response.getHits().getTotal());
            assertEquals("val1", response.getHits().iterator().next().getSource().get("key1"));
            assertNull("indexOnly fields should not be stored",
                       response.getHits().iterator().next().getSource().get("key3"));
        });
```

kezhenxu94 (approved):
Generally LGTM, some nits, please check

oap-server/server-library/library-elasticsearch-client/src/test/java/org/apache/skywalking/library/elasticsearch/ITElasticSearchTest.java
wankai123: Good suggestion!

kezhenxu94 (approved):

CHANGES.md
```diff
@@ -109,6 +109,9 @@ Release Notes.
 * [Breaking Change] Remove configuration `restAcceptorPriorityDelta` (env var: `SW_RECEIVER_SHARING_JETTY_DELTA`
   , `SW_CORE_REST_JETTY_DELTA`).
 * [Breaking Change] Remove configuration `graphql/path` (env var: `SW_QUERY_GRAPHQL_PATH`).
+* [Breaking Change] Remove configuration `restMinThreads` (env var: `SW_CORE_REST_JETTY_MIN_THREADS`, `SW_RECEIVER_SHARING_JETTY_MIN_THREADS`).
```
wu-sheng, line 112: Is this missed from the previous PR? @kezhenxu94

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
```diff
@@ -58,6 +58,12 @@
      */
     boolean storageOnly() default false;
 
+    /**
+     * The column(field) is just indexed, never stored. Note: this feature only supported by elasticsearch
+     * and can not support update mappings in current index
```
wu-sheng, line 63:
```suggestion
     * and don't support mappings update due to ElasticSearch server's limitation.
     *
     * NOTICE, metrics should not use this, as the OAP core merges indices of metrics automatically.
```

`can't` usually means we can not originally.

CHANGES.md
kezhenxu94: This was duplicate maybe because of resolving conflict

wu-sheng:
```suggestion
```

wu-sheng (approved):

PR #8680, opened 2022-03-15
Polish Zipkin HTTP receiver with Armeria APIs

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Follow up #8673.


oap-server/server-receiver-plugin/zipkin-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/zipkin/handler/ZipkinSpanHTTPHandler.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    @Post("/api/v1/spans")
+    @ConsumesJson
+    public HttpResponse collectV1JsonSpans(ServiceRequestContext ctx, HttpRequest req) {
+        return doCollectSpans(SpanBytesDecoder.JSON_V1, ctx, req);
+    }
+
+    @Post("/api/v1/spans")
+    @ConsumesThrift
```
wu-sheng, line 101: Could you point out the doc to explain this? I think `x-thrift` is just a decoding process in Zipkin. Why does this relate to HTTP server?

kezhenxu94: You can see there are several methods mapped to the same endpoint `@Post("/api/v1/spans")`, this annotation `@ConsumesThrift` means only those requests with http header `Content-Type: application/x-thrift` will be handled by this method, meaning only thrift stream will be decoded by this method, correctly.

kezhenxu94: Here we are using the annotation `@Consumes**` to route the different protocols to responding decoder.

wu-sheng (approved):

wankai123 (approved):

PR #8685, opened 2022-03-16
Remove hard requirement of BASE64 encoding for binary field

wu-sheng:

Remove hard requirement of BASE64 encoding for binary field, and resolved concerns from https://github.com/apache/skywalking/pull/8684#issuecomment-1068660880


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/type/HashMapConverter.java
```diff
@@ -34,6 +40,25 @@
...
+        @Override
+        public <T, R> R getWith(final String fieldName, final Function<T, R> typeDecoder) {
+            final T value = (T) source.get(fieldName);
+            return typeDecoder.apply(value);
+        }
+
+        /**
+         * Default Base64Decoder supplier
+         */
+        public static class Base64Decoder implements Function<String, byte[]> {
```
kezhenxu94, line 53: This can be a singleton?

wu-sheng: Done.

lujiajing1126 (approved):

kezhenxu94 (dismissed):

wankai123 (approved):

PR #8691, opened 2022-03-16
Improve CI with the GHA new feature

kezhenxu94:

This PR optimize the following things by using the new feature of GitHub Actions "rerun failed jobs"

- Bring all common dependencies (like build tar, build docker, sanity check, etc.) of UT, IT, E2E together and only execute them once.
- Minimize the dependency tree, so when a job failed, we can rerun it and its dependents with minimal resources / time.

The dependency graph looks like this 

image

This PR also fixes a problem that cancelled jobs can pass the "Required" CI check context.


wu-sheng (dismissed):

wu-sheng:
You need to sync the branch manually. It seems you closed the authentication of fork repo update.

wu-sheng (approved):
LGTM. The only side effect of this new GHA version, we can't rerun the failed tasks until the whole process is done.

wu-sheng (approved):

PR #8692, opened 2022-03-17
Remove fake required workflows

kezhenxu94:
Follow up #8691 , remove those fake jobs that was only used to pass the CI

### DIFF
--- a/.github/workflows/skywalking-fake.yaml
+++ b/.github/workflows/skywalking-fake.yaml
@@ -19,6 +19,7 @@ name: CI
 on:
   pull_request:
     paths-ignore:
+      - "**"
       - "!*.md"
       - "!*.txt"
       - "!skywalking-ui"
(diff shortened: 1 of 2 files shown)


wu-sheng:
`Required` is controlled by the GHA file, rather than settings, now?

kezhenxu94:
> `Required` is controlled by the GHA file, rather than settings, now?

I had changed the required context name to `Required` in last PR and now we can use this single name to aggregate all mandatory jobs, so no need to have several fake jobs,

wu-sheng:
I see. Then we need this gets merged before other PRs.

kezhenxu94:
> I see. Then we need this gets merged before other PRs.

Actually not necessary, the fake `Required` workflow is also added in last PR so all new PRs can be merged, this PR is only to remove those unused ones

wu-sheng (approved):

PR #8696, opened 2022-03-17
Implement the eBPF profiling query and data collect protocol

mrproliu:
* Implement the eBPF profiling query and data collect protocol
* Migrate existing profile classes to the profiling package
* I will submit a new PR after rover side finished for E2E testing

### Profiling Data Merging Benchmark
We have three types of data sources, all of the benchmarks are only merging data without query.
|Name|Report Period|Dump Duration|Merge Instance Count|Stack Depth Range|Stack Duplicate Rate(%)|Stack Symbol Duplicate Rate(%)|
|---|----|----|---|---|---|---|
|LowDatasource|5s|60m|10|15,30|5|100,40,35,30,15,10,7,5|
|MedianDatasource|5s|100m|200|15,30|3|50,40,35,30,20,10,7,5,2|
|HighDatasource|5s|2h|2000|15,40|1|30,27,25,20,17,15,10,7,5,2,1|

```
Benchmark                                                Mode  Cnt     Score    Error  Units
EBPFProfilingAnalyzerBenchmark.analyzeLowDataSource     thrpt   20  2134.237 ± 33.583  ops/s
EBPFProfilingAnalyzerBenchmark.analyzeMaxDataSource     thrpt   20     6.128 ±  0.149  ops/s
EBPFProfilingAnalyzerBenchmark.analyzeMedianDataSource  thrpt   20    97.159 ±  3.105  ops/s
```

### GraphQL Screenshot
image
image
image
image


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/ebpf/analyze/EBPFProfilingAnalyzer.java
```diff
@@ -0,0 +1,157 @@
...
+    protected IEBPFProfilingDataDAO getDataDAO() {
+        if (dataDAO == null) {
+            dataDAO = moduleManager.find(StorageModule.NAME)
+                                                         .provider()
+                                                         .getService(IEBPFProfilingDataDAO.class);
+        }
+        return dataDAO;
+    }
+
+    private static class TimeRange {
```
wu-sheng, line 148: What does TimeRange mean?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/ebpf/analyze/EBPFProfilingAnalyzer.java
```diff
@@ -0,0 +1,157 @@
...
+        }
+        while (start < end);
+
+        return timeRanges;
+    }
+
+    /**
+     * Analyze records
+     */
+    protected List<EBPFProfilingTree> analyze(List<EBPFProfilingStack> stacks) {
```
wu-sheng, line 124: This method should have very strong UTs to cover merging logic. This is the key of the whole visualization result.

wu-sheng: Also, we need some benchmarks for this merging, as you provide a list of durations, so we could face a larger merging.

At the same time, we need a `circuit break` to avoid timeout. We should have the capability to provide a half-finished profiling tree with a message of `circuit break`.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/EBPFProfilingTaskEsDAO.java
```diff
@@ -0,0 +1,110 @@
...
+        this.taskMaxSize = config.getProfileTaskQueryMaxSize();
+    }
+
+    @Override
+    public List<EBPFProfilingTask> queryTasks(EBPFProfilingProcessFinder finder, EBPFProfilingTargetType targetType, long taskStartTime, long latestUpdateTime) throws IOException {
+        final String index =
+                IndexController.LogicIndicesRegister.getPhysicalTableName(EBPFProfilingTaskRecord.INDEX_NAME);
+        final BoolQueryBuilder query = Query.bool();
+
+        if (finder != null && finder.getFinderType() != null) {
```
wu-sheng, line 59: Don't check this, we should throw NPE or IllegalArgumentException, rather than doing a full table scan.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/ebpf/storage/EBPFProfilingTaskRecord.java
```diff
@@ -0,0 +1,123 @@
...
+import org.apache.skywalking.oap.server.core.storage.type.Convert2Storage;
+import org.apache.skywalking.oap.server.core.storage.type.StorageBuilder;
+
+import static org.apache.skywalking.oap.server.core.source.DefaultScopeDefine.EBPF_PROFILING_TASK;
+
+@Data
+@ScopeDeclaration(id = EBPF_PROFILING_TASK, name = "EBPFProfilingTask")
+@Stream(name = EBPFProfilingTaskRecord.INDEX_NAME, scopeId = EBPF_PROFILING_TASK,
+        builder = EBPFProfilingTaskRecord.Builder.class, processor = NoneStreamProcessor.class)
+public class EBPFProfilingTaskRecord extends NoneStream {
```
wu-sheng, line 40:
```suggestion
public class EBPFProfilingRecord extends NoneStream {
```

Don't take `task` every place. You have `task`, `schedule`, `record` concepts, right?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/EBPFProfilingTaskEsDAO.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+        if (latestUpdateTime > 0) {
+            query.must(Query.range(EBPFProfilingTaskRecord.LAST_UPDATE_TIME).gt(latestUpdateTime));
+        }
+
+        final SearchBuilder search = Search.builder().query(query)
+                .sort(EBPFProfilingTaskRecord.CREATE_TIME, Sort.Order.DESC)
+                .size(taskMaxSize);
+
+        final SearchResponse response = getClient().search(index, search.build());
```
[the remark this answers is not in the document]
mrproliu: `EBPFProfilingTask` means the task which user created. The `EBPFProfilingData` is the profiling reported data, this one in the ES storage have using the scroll to query.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/ebpf/analyze/EBPFProfilingAnalyzer.java
mrproliu: The `TimeRange` using to split the query, if the analysis time range is bigger than 2 minutes, then split them.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/ebpf/storage/EBPFProfilingTaskRecord.java
mrproliu: Yes, now these three concepts are `EBPFProfilingTaskRecord`, `EBPFProfilingScheduleTraffic`, and `EBPFProfilingDataRecord`.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/EBPFProfilingTaskEsDAO.java
mrproliu: being added in the core level check before the storage query.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/ebpf/analyze/EBPFProfilingAnalyzer.java
mrproliu: I have added a [benchmark](https://github.com/apache/skywalking/pull/8696/files#diff-15d820525f50abd439c83f34d6afd86a01134ed964142dc0fce907b7c1d4cc33) working on checking the analysis time without query data. I think profiling data analysis is not a big problem, we should add the `circuit break` when the query timeout. WDYT?

wu-sheng:
How many threads for these op? Should we set up a recommended max range for UI?

mrproliu:
> How many threads for these op? 

Do you mean Profiling Data Merging benchmark? I have declared the annotation `@Thread(4)` in the benchmark, so I think there using 4 threads.

> Should we set up a recommended max range for UI?

I think maybe two hours should be the max analysis duration. 
1. Long-time range means that a lot of data need be obtained from the storage, which will take a longer time in query and cause a bad user experience
2. Long-time analysis is meaningless, usually a few minutes - one-hour duration is enough to explain the problem.

What do you think?

wu-sheng:
Then let's set a configuration to limit range, how about default in 10 mins(max)?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/EBPFProfilingTaskEsDAO.java
wu-sheng: I mean you should not check parameters not null in DAO, if they are required. We need to avoid in some case, really calling DAO with all NULL, and cause full table scan.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/ebpf/analyze/EBPFProfilingAnalyzer.java
wu-sheng: Make sense.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/ebpf/analyze/EBPFProfilingAnalyzer.java
wu-sheng: Then please add comments. A lot of codes should be explained in comments.

mrproliu:
> Then let's set a configuration to limit range, how about default in 10 mins(max)?

Added.

oap-server/server-receiver-plugin/skywalking-ebpf-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/ebpf/provider/handler/EBPFProfilingServiceHandler.java
```diff
@@ -0,0 +1,180 @@
...
+                data.setUploadTime(task.getCurrentTime());
+                switch (ebpfProfilingData.getProfilingCase()) {
+                    case ONCPU:
+                        try {
+                            processOnCPUProfiling(data, ebpfProfilingData.getOnCPU());
+                        } catch (IOException e) {
+                            log.warn("process ON_CPU profiling data failure", e);
+                        }
+                        break;
+                    case PROFILING_NOT_SET:
```
wu-sheng, line 137: Should be `default`?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/EBPFProfilingScheduleEsDAO.java
```diff
@@ -0,0 +1,93 @@
...
+        if (StringUtil.isNotEmpty(taskId)) {
+            query.must(Query.term(EBPFProfilingScheduleRecord.TASK_ID, taskId));
+        }
+        RangeQueryBuilder timeRangeQuery = null;
+        Supplier<RangeQueryBuilder> timeRangeQuerySupplier = () -> Query.range(EBPFProfilingScheduleRecord.START_TIME);
+        if (startTimeBucket > 0) {
+            timeRangeQuery = Optional.ofNullable(timeRangeQuery).orElseGet(timeRangeQuerySupplier)
+                    .gte(TimeBucket.getTimestamp(startTimeBucket));
+        }
+        if (endTimeBucket > 0) {
```
wu-sheng, line 69: Same as previous review, `startTimeBucket` and `endTimeBucket` should not be null in any case, please throw an illegalArgumentException if so. Recheck for all DAO implementations.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/EBPFProfilingTaskEsDAO.java
mrproliu: I just use one method to cover all the different ways to get the tasks, should I change to the multiple methods to avoid the null check?

wu-sheng: Check is fine, you should add else.
Just a suggestion, not a block.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/ebpf/analyze/EBPFProfilingAnalyzer.java
mrproliu: Add a new configuration item for the query timeout. default `30s`.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/EBPFProfilingTaskEsDAO.java
mrproliu: Ok, I have updated all of the `EBPFProfilingSchedule` and `EBPFProfilignData` unnecessary checks.

oap-server/server-receiver-plugin/skywalking-ebpf-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/ebpf/provider/handler/EBPFProfilingServiceHandler.java
mrproliu: Update to the `default`

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/EBPFProfilingScheduleEsDAO.java
mrproliu: Updated.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/ebpf/analyze/EBPFProfilingAnalyzer.java
mrproliu: Added.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/ebpf/analyze/EBPFProfilingAnalyzer.java
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * eBPF Profiling Analyzer working on data query and combine them for generate the Flame Graph.
+ */
+@Slf4j
+public class EBPFProfilingAnalyzer {
+
+    private static final EBPFProfilingAnalyzeCollector ANALYZE_COLLECTOR = new EBPFProfilingAnalyzeCollector();
+    private static final Long FETCH_DATA_DURATION = TimeUnit.MINUTES.toMillis(2);
+    private static final ExecutorService FETCH_DATA_THREAD_POOL = Executors.newCachedThreadPool();
```
wu-sheng, line 53: The maximumPoolSize of the thread pool should be fixed and configurable. Default, it is Int.MAX, which could easily crash OAP server. We need to reject or timeout when face too many analysis requests

docs/en/setup/backend/configuration-vocabulary.md
```diff
@@ -46,6 +46,8 @@ core|default|role|Option values: `Mixed/Receiver/Aggregator`. **Receiver** mode
 | - | - | maxSizeOfAnalyzeProfileSnapshot| The maximum number of snapshots analyzed by the OAP. | - | 12000 |
 | - | - | prepareThreads| The number of threads used to prepare metrics data to the storage. | SW_CORE_PREPARE_THREADS | 2 |
 | - | - | enableEndpointNameGroupingByOpenapi | Automatically groups endpoints by the given OpenAPI definitions. | SW_CORE_ENABLE_ENDPOINT_NAME_GROUPING_BY_OPAENAPI | true |
+| - | - | maxDurationOfAnalyzeEBPFProfiling| The maximum duration(in minute) of analyze the eBPF profiling data. | - | 10 |
+| - | - | maxDurationOfQueryEBPFProfilingData| The maximum timeout duration(in second) of query the eBPF profiling data when analysis. | - | 30 |
```
wu-sheng, line 50: `maximum` and `timeout` is duplicated :) Choose one. There is not a minimal timeout, right?

wu-sheng: And according to your codes, this is related to time of the database query, right?

apm-protocol/apm-network/src/main/java/org/apache/skywalking/oap/server/network/trace/component/command/EBPFProfilingTaskCommand.java
```diff
@@ -0,0 +1,74 @@
...
+
+package org.apache.skywalking.oap.server.network.trace.component.command;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import org.apache.skywalking.apm.network.common.v3.Command;
+import org.apache.skywalking.apm.network.common.v3.KeyStringValuePair;
+
+/**
+ * eBPF profiling task command use to send task to the ebpf agent side
```
wu-sheng, line 27:
```suggestion
 * eBPF profiling task command, OAP uses this to send a task to the ebpf agent side
```

docs/en/setup/backend/configuration-vocabulary.md
mrproliu: Yes, `maxDurationOfAnalyzeEBPFProfiling` means the user chooses the time range max duration. `maxDurationOfQueryEBPFProfilingData ` means the profiling data query to storage max duration. the second one uses to avoid users waiting for the query taking a long time.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/ebpf/analyze/EBPFProfilingAnalyzer.java
mrproliu: This thread pool size is the same as the java ForkJoin pool size, should I update to `Runtime.getRuntime().availableProcessors()`?

docs/en/setup/backend/configuration-vocabulary.md
wu-sheng:
```suggestion
| - | - | maxDurationOfQueryEBPFProfilingData| The maximum duration(in second) of query the eBPF profiling data from database. | - | 30 |
```

I mean this.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/ebpf/analyze/EBPFProfilingAnalyzer.java
wu-sheng:
```java
    public static ExecutorService newCachedThreadPool() {
        return new ThreadPoolExecutor(0, Integer.MAX_VALUE,
                                      60L, TimeUnit.SECONDS,
                                      new SynchronousQueue<Runnable>());
    }
```

From my local source codes, the 2nd parameter is `maximumPoolSize`, which is `Integer.MAX_VALUE`.

wu-sheng: `Runtime.getRuntime().availableProcessors()` could be default, but it should be configurable for different use cases. Such as some users may want less CPU working for analysis, more for streaming aggregation process.

wu-sheng: Just say, using `Runtime.getRuntime().availableProcessors()` in every time we need a thread pool, globally, the number of threads would also be crazy :) Isn't it.

mrproliu: Yes, may we need two global thread pools for doing that. One for the profiling data analysis, the other one for the wait for the data query(works for the circle break the query when timeout). WDYT?

wu-sheng: Let's consider that later. For now, let's just make sure `Runtime.getRuntime().availableProcessors()` as default and support configuration through application.yml

mrproliu: Added.

wu-sheng (dismissed):
LGTM. @kezhenxu94 Any suggestion?

kezhenxu94 (commented):
Just some comments written yesterday so might be outdated and ignore them if they are not applicable

wu-sheng (approved):

PR #8705, opened 2022-03-18
Add `Column.shardingKeyIdx` for column definition for BanyanDB

wu-sheng:

* Add `Column.shardingKeyIdx` for column definition for BanyanDB.

```
Sharding key is used to group time series data per metric of one entity in one place (same sharding or same 
column for column-oriented database).
For example,
ServiceA's traffic gauge, service call per minute, includes following timestamp values, then it should be sharded by service ID
[ServiceA(encoded ID): 01-28 18:30 values-1, 01-28 18:31 values-2, 01-28 18:32 values-3, 01-28 18:32 values-4]

BanyanDB is the 1st storage implementation supporting this. It would make continuous time series metrics stored closely and compressed better.

NOTICE, this sharding concept is NOT just for splitting data into different database instances or physical files.
```

* Add `JD_PRESERVE_LINE_FEEDS=true` in official code style file. Then our comments don't have to use many blank lines to keep format.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -127,8 +127,8 @@ Release Notes.
 ```
 Sharding key is used to group time series data per metric of one entity.
 For example,
-ServiceA's traffic gauge, service call per minute, includes following timestamp values, then it should be
-[ServiceA: 01-28 18:30 values-1, 01-28 18:31 values-2, 01-28 18:32 values-3, 01-28 18:32 values-4]
+ServiceA's traffic gauge, service call per minute, includes following timestamp values, then it should be shard by service ID
+[ServiceA(encoded ID): 01-28 18:30 values-1, 01-28 18:31 values-2, 01-28 18:32 values-3, 01-28 18:32 values-4]
 
 BanyanDB is the 1st storage implementation supporting this. It would make continuous time series metrics stored closely and compressed better.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,6 +31,7 @@ Release Notes.
 * Update frontend-maven-plugin to 1.12 and npm to 16.14.0 for booster UI build.
 * Improve CI with the GHA new feature "run failed jobs".
 * Fix `./mvnw compile` not work if `./mvnw install` is not executed at least once.
+* Add `JD_PRESERVE_LINE_FEEDS=true` in official code style file.
 
 #### OAP Server
 
@@ -112,7 +113,8 @@ Release Notes.
   , `SW_CORE_REST_JETTY_DELTA`).
 * [Breaking Change] Remove configuration `graphql/path` (env var: `SW_QUERY_GRAPHQL_PATH`).
 * Add storage column attribute `indexOnly`, support ElasticSearch only index and not store some fields.
-* Add `indexOnly=true` to `SegmentRecord.tags`, `AlarmRecord.tags`, `AbstractLogRecord.tags`, to reduce unnecessary storage.
+* Add `indexOnly=true` to `SegmentRecord.tags`, `AlarmRecord.tags`, `AbstractLogRecord.tags`, to reduce unnecessary
+  storage.
 * [Breaking Change] Remove configuration `restMinThreads` (env var: `SW_CORE_REST_JETTY_MIN_THREADS`
   , `SW_RECEIVER_SHARING_JETTY_MIN_THREADS`).
 * Refactor the core Builder mechanism, new storage plugin could implement their own converter and get rid of hard
@@ -120,6 +122,18 @@ Release Notes.
 * [Breaking Change] Break all existing 3rd-party storage extensions.
 * Remove hard requirement of BASE64 encoding for binary field.
 * Add complexity limitation for GraphQL query to avoid malicious query.
+* Add `Column.shardingKey` for column definition for BanyanDB.
+
+```
+Sharding key is used to group time series data per metric of one entity.
+For example,
+ServiceA's traffic gauge, service call per minute, includes following timestamp values, then it should be
+[ServiceA: 01-28 18:30 values-1, 01-28 18:31 values-2, 01-28 18:32 values-3, 01-28 18:32 values-4]
+
+BanyanDB is the 1st storage implementation supporting this. It would make continuous time series metrics stored closely and compressed better.
+
+NOTICE, this sharding concept is NOT for splitting data into different database instances or physical files.
+```
 
 #### UI
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -122,17 +122,18 @@ Release Notes.
 * [Breaking Change] Break all existing 3rd-party storage extensions.
 * Remove hard requirement of BASE64 encoding for binary field.
 * Add complexity limitation for GraphQL query to avoid malicious query.
-* Add `Column.shardingKey` for column definition for BanyanDB.
+* Add `Column.shardingKeyIdx` for column definition for BanyanDB.
 
 ```
-Sharding key is used to group time series data per metric of one entity.
+Sharding key is used to group time series data per metric of one entity in one place (same sharding or same 
+column for column-oriented database).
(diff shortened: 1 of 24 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
```diff
@@ -102,6 +101,28 @@
...
+     * 2. If no column is appointed for this, {@link org.apache.skywalking.oap.server.core.storage.StorageData#id}
+     * would
+     * be used by the storage implementation accordingly.
+     *
+     * NOTICE, this sharding concept is NOT for splitting data into different database instances or physical files.
+     *
+     * @return TRUE if this column could(not must) be used for sharding.
+     * @since 9.0.0
+     */
+    boolean shardingKey() default false;
```
wu-sheng, line 124: @hanahmily @lujiajing1126 Please check this comment carefully. I want to be sure I understand this concept correctly.

wu-sheng:
71743105-449D-41F8-B21D-C8B5C0D57DBA

I have tested this for OAL generated metrics, it works.

lujiajing1126:
I suppose we also need sharding keys for record, e.g. we use `is_error`, `service_id` and `service_instance_id` for `SegmentRecord`

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
```diff
@@ -102,6 +101,28 @@
...
+     * Sharding key is used to group time series data per metric of one entity.
+     * For example,
+     * ServiceA's traffic gauge, service call per minute, includes following timestamp values, then it should be shard
+     * by service ID
+     * [ServiceA(encoded ID): 01-28 18:30 values-1, 01-28 18:31 values-2, 01-28 18:32 values-3, 01-28 18:32 values-4]
+     *
+     * BanyanDB is the 1st storage implementation supporting this. It would make continuous time series metrics stored
+     * closely and compressed better.
+     *
+     * 1. One entity at most has one sharding key
```
lujiajing1126, line 114: No. We support sharding with multiple keys now.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
```diff
@@ -102,6 +101,28 @@
...
 
+    /**
+     * Sharding key is used to group time series data per metric of one entity.
+     * For example,
+     * ServiceA's traffic gauge, service call per minute, includes following timestamp values, then it should be shard
+     * by service ID
+     * [ServiceA(encoded ID): 01-28 18:30 values-1, 01-28 18:31 values-2, 01-28 18:32 values-3, 01-28 18:32 values-4]
+     *
+     * BanyanDB is the 1st storage implementation supporting this. It would make continuous time series metrics stored
+     * closely and compressed better.
```
lujiajing1126, line 112:
```suggestion
     * closely and compressed better. Besides, it can significantly alleviate high-cardinality issue for some indexed tags.
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
```diff
@@ -102,6 +101,28 @@
      */
     AnalyzerType analyzer() default AnalyzerType.OAP_ANALYZER;
 
+    /**
+     * Sharding key is used to group time series data per metric of one entity.
+     * For example,
+     * ServiceA's traffic gauge, service call per minute, includes following timestamp values, then it should be shard
+     * by service ID
+     * [ServiceA(encoded ID): 01-28 18:30 values-1, 01-28 18:31 values-2, 01-28 18:32 values-3, 01-28 18:32 values-4]
```
lujiajing1126, line 109:
```suggestion
     * [ServiceA(encoded via hash function): 01-28 18:30 values-1, 01-28 18:31 values-2, 01-28 18:32 values-3, 01-28 18:32 values-4]
```

lujiajing1126 (changes requested):

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
wu-sheng: @hanahmily What do you prefer? I think you mentioned not to use multiple keys.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
wu-sheng: I think what you add is from database implementation, not has to be. Right?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/Column.java
wu-sheng: encoded means from OAP perspective. Via hash function is another thing from server side implementation. We should not document here.

wu-sheng:
> suppose we also need sharding keys for record, e.g. we use `is_error`, `service_id` and

@hanahmily any suggestion? We didn't discuss for traces and logs.

hanahmily:
> > suppose we also need sharding keys for record, e.g. we use `is_error`, `service_id` and
> 
> @hanahmily any suggestion? We didn't discuss for traces and logs.

After taking a look at segment and abstract log, I recommend the combination of entity pieces:

 * segment: `service_id`, `service_instance_id`, `is_error`
 * log: `service_id`,`service_instance_id`
 * other records: `id`, that's because these records don't need to be grouped by sharding key to get a better compression ratio.

wu-sheng:
@hanahmily @lujiajing1126 After confirmation that multiple columns should be supported, I changed `shardingKey` to `shardingKeyIdx`, which makes more sense as from both data insert/update or database sharding perspective, the sequence matters much. In the core level, I also added a checker to make sure the index is from zero and continuous.

Also, I added multiple-key sharding for trace(native and zipkin) and log(browser and general), using service ID, instance ID, and status.

FYI @mrproliu in the ebpf profiling, as the profiling record could be huge, please consider at least service ID as sharding key.

hanahmily (approved):
LGTM

lujiajing1126 (approved):
LGTM

CHANGES.md
```diff
@@ -112,14 +113,28 @@ Release Notes.
...
+```
+Sharding key is used to group time series data per metric of one entity in one place (same sharding or same 
+column for column-oriented database).
+For example,
+ServiceA's traffic gauge, service call per minute, includes following timestamp values, then it should be sharded by service ID
+[ServiceA(encoded ID): 01-28 18:30 values-1, 01-28 18:31 values-2, 01-28 18:32 values-3, 01-28 18:32 values-4]
+
+BanyanDB is the 1st storage implementation supporting this. It would make continuous time series metrics stored closely and compressed better.
+
+NOTICE, this sharding concept is NOT just for splitting data into different database instances or physical files.
```
kezhenxu94, line 136: You keep using the term `shard` but explaining it's actually for grouping, what's the reason not to just use name like `Column.groupKeyIdx`?

lujiajing1126: From the prospetive of the acutual implementation, the value generated from the ordered combination of all sharkdingKeys here is used for sharding.

wu-sheng: > Sharding key is used to group time-series data per metric of one entity in one place

> this sharding concept is NOT just for splitting data into different database instances or physical files

These 2 sentences are not conflicting. Group the key-equal data, and split the whole data set. If you feel confused, what do you prefer?

PR #8706, opened 2022-03-18
Support ElasticSearch template mappings `properties parameters` and `_source` update

wankai123:
### Support ElasticSearch template mappings `properties parameters` and `_source` update

Previous logic can't update the mappings `properties parameters` and `_source` only support adding a new property.
### Now
1. As the OAP core merges the index of metrics, properties would not be deleted still.
2. Would not update the current index to avoid conflict, a new index follows the new template to create.

### DIFF
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexStructures.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexStructures.java
@@ -21,7 +21,6 @@ package org.apache.skywalking.oap.server.storage.plugin.elasticsearch.base;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.skywalking.library.elasticsearch.response.Mappings;
 import org.apache.skywalking.oap.server.library.client.elasticsearch.ElasticSearchClient;
@@ -67,7 +66,8 @@ public class IndexStructures {
 
     /**
      * Returns mappings with fields that not exist in the input mappings.
-     * do not return _source config to avoid index update conflict.
+     * The input mappings should be history mapping from current index.
+     * Do not return _source config to avoid current index update conflict.
      */
     public Mappings diffStructure(String tableName, Mappings mappings) {
         if (!structures.containsKey(tableName)) {
@@ -102,7 +102,7 @@ public class IndexStructures {
      */
     public static class Fields {
         private final Map<String, Object> properties;
-        Mappings.Source source;
+        private Mappings.Source source;
 
         private Fields(Mappings mapping) {
             this.properties = mapping.getProperties();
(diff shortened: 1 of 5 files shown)


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexStructures.java
```diff
@@ -113,28 +113,39 @@ private Fields(Mappings mapping) {
...
+                                        } else {
+                                            return false;
+                                        }
+                                    });
+            if (!isContains) {
+                return false;
+            }
+
+            if (fields.source != null && this.source != null) {
+                isContains = this.source.getExcludes().toString().equals(fields.source.getExcludes().toString());
```
kezhenxu94, line 134: Consider implement the `equals` method in class `Source`, instead of comparing their `toString()` methods, which is error prone

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexStructures.java
```diff
@@ -113,28 +113,39 @@ private Fields(Mappings mapping) {
...
+                                        Object hisItem = this.properties.get(item.getKey());
+                                        if (hisItem != null) {
+                                            return hisItem.toString().equals(item.getValue().toString());
+                                        } else {
+                                            return false;
+                                        }
+                                    });
+            if (!isContains) {
+                return false;
+            }
```
kezhenxu94, line 131:
```suggestion
            boolean contains =
                fields.properties
                    .entrySet()
                    .stream()
                    .allMatch(
                        item -> Objects.equals(properties.get(item.getKey()), item.getValue()));
            if (!contains) {
                return false;
            }
```

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexStructures.java
```diff
@@ -113,28 +113,39 @@ private Fields(Mappings mapping) {
...
+
+            if (fields.source != null && this.source != null) {
+                isContains = this.source.getExcludes().toString().equals(fields.source.getExcludes().toString());
+            } else {
+                return fields.source == null && this.source == null;
+            }
+            return isContains;
```
kezhenxu94, line 138: After you implement the `equals` method in `Source`, 

```suggestion
            return Objects.equals(this.source, fields.source);
```

kezhenxu94 (approved):

wu-sheng (approved):

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexStructures.java
```diff
@@ -113,28 +113,25 @@ private Fields(Mappings mapping) {
...
-                                     Map.Entry::getKey,
-                                     Map.Entry::getValue
-                                 ));
-            properties.putAll(newFields);
-            if (source != null) {
-                Set<String> exclude = source.getExcludes();
-                Set<String> newExclude = fields.source.getExcludes();
-                exclude.addAll(newExclude);
-            }
+            properties.putAll(fields.properties);
```
wu-sheng, line 133: Just a note, if there are properties conflicts, same key with different values, they would override each other.

wu-sheng: **Warning**, if there are properties conflicts, same key with different values, they would override each other according to booting sequence. Unless we change booting process as merging modules before merging with existing templates and indices, we can't add check mechanism and warnings.

wu-sheng: I have added this in commit logs, in case we need the context tp resolve bug or do enhancement.

PR #8724, opened 2022-03-22
Insure span analyze isNormal and Layer consistent.

wankai123:

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/Layer.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/Layer.java
@@ -131,4 +131,23 @@ public enum Layer {
         }
         return layer;
     }
+
+    /**
+     * The `normal` status represents this service is detected by an agent. The `un-normal` service is conjectured by
+     * telemetry data collected from agents on/in the `normal` service.
+     */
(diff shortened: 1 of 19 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/Layer.java
```diff
@@ -131,4 +131,23 @@ public static Layer valueOf(int value) {
         }
         return layer;
     }
+
+    /**
+     * The `normal` status represents this service is detected by an agent. The `un-normal` service is conjectured by
+     * telemetry data collected from agents on/in the `normal` service.
+     */
+    public static boolean isNormal(Layer layer) {
```
wu-sheng, line 139: Adding `Layer#isNormal` may be easier, but anyway, not requried.

wu-sheng (approved):
Waiting for CI, good to me.

wu-sheng (changes requested):
Some e2e fails, waiting for fix.

wu-sheng (approved):

wu-sheng (changes requested):
I think uninstrumented Gateway is not covered.

wu-sheng (dismissed):

wu-sheng (approved):

PR #8725, opened 2022-03-22
Add data-generator module to run OAP in testing mode, generating mock data for testing

### LINKED ISSUE #8697: [Feature] [OAP + BanyanDB] [Test] Support OAP running in simulation test mode
### Search before asking

### Description

@hanahmily @kezhenxu94 and I had a discussion a week ago, with BanyanDB 0.1 MVP version is going to be ready soon. We need to establish a mode to make OAP could generate simulated traces/metrics/logs without target services and agents installed/deployed

So, I suppose we should create a new maven and OAP module, as well as a new binary(tar and docker image), to provide configurations in `application.yml`. 

### Use case

In this simulation test mode, developers could provide trace/log/metric templates as well as parameters to control data set of simulated telemetry data.

Notice, this module and binary should not be generated in both default package and release package

### Related issues

_No response_

### Code of Conduct

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #8697 .


kezhenxu94:
Some examples

<details>
  <summary>Submit a task to generate segments</summary>

  ```shell

### Submit a task to generate mock segment data, 100 segments per seconds

POST http://localhost:12800/mock-data/segments/tasks?qps=100
Content-Type: application/json

{
    "traceId": {
        "type": "randomString",
        "length": "20",
        "letters": true,
        "numbers": true
    },
    "serviceInstanceId": {
        "type": "randomString",
        "length": "10",
        "letters": true,
        "numbers": true
    },
    "dataBinary": {
        "type": "randomString",
        "length": "10",
        "prefix": "test",
        "letters": true,
        "numbers": true
    },
    "segmentId": {
        "type": "randomString",
        "length": "20",
        "letters": true,
        "numbers": true
    },
    "serviceId": {
        "type": "randomString",
        "length": "10",
        "letters": true,
        "numbers": true
    },
    "endpointId": {
        "type": "randomString",
        "length": "10",
        "prefix": "test_",
        "letters": true,
        "numbers": true,
        "domainSize": 10
    },
    "startTime": {
        "type": "randomInt",
        "min": "1647868381000",
        "max": "1647878402000",
        "domainSize": 10
    },
    "latency": {
        "type": "randomInt",
        "min": 100,
        "max": 1000
    },
    "error": {
        "type": "randomInt",
        "min": "0",
        "max": "1"
    },
    "tags": {
        "type": "randomList",
        "size": 5,
        "item": {
            "key": {
                "type": "randomString",
                "length": "10",
                "prefix": "test",
                "letters": true,
                "numbers": true,
                "domainSize": 10
            },
            "value": {
                "type": "randomString",
                "length": "10",
                "prefix": "test",
                "letters": true,
                "numbers": true
            }
        }
    }
}
  ```

</details>

<details>
  <summary>Cancel the segment generating task</summary>

  ```shell
### Cancel the task, with the requestId return from the previous request
DELETE http://localhost:12800/mock-data/segments/tasks?requestId=8bbdaee6-2193-405d-bc0c-a650cd17597e

  ```

</details>

<details>
  <summary>List all segment tasks</summary>

  ```shell
### List all tasks that are now running
GET http://localhost:12800/mock-data/segments/tasks
  ```

</details>

kezhenxu94:
In this PR I only implement the data generators for Segment and Log, not metrics.

kezhenxu94:
@hanahmily @lujiajing1126 please check whether ^^^ that will be sufficient for our BanyanDB MVP tests.

wu-sheng:
Questions
1. Is the ID generated as a random string only? If so, we may face errors in queries as they are not well encoded. How about generating names?
2. Let's add `UUID` and `Java-Trace-ID` generator for trace ID generating, which could make the ID more reasonable.
3. I can see `domainSize` in endpoint and tags, should we add(or added) for service and instance too?
4. `error` should have a error stack attachment, which should be generated inside span's logs.
5. The number of spans(with tags and logs) seems missing.
6. Recommend to write a README.md for this testing-mode tar, including how to use, and examples of traces generated.

> In this PR I only implement the data generators for Segment and Log

Could you make the log command clear?

kezhenxu94:
@wu-sheng do we need the mock data to be that real and do we need to query the data through OAP? I even wanted to remove the query module from the new starter...

wu-sheng:
Support query is necessary. Because one reason to build this is to verify the query performance in large scale of data.

wu-sheng:
We don't need to make all UI functional working. But GrapgQL should at least be able to work.(considering a lot of logic about data consistent setups in UI and CLI.

oap-server/server-tools/data-generator/src/main/java/org/apache/skywalking/restapi/SegmentGeneratorHandler.java
```diff
@@ -0,0 +1,124 @@
...
+    private final EventLoopGroup eventLoopGroup = EventLoopGroups.newEventLoopGroup(10);
+
+    public SegmentGeneratorHandler(ModuleManager manager) {
+        sourceReceiver = manager.find(CoreModule.NAME).provider().getService(SourceReceiver.class);
+    }
+
+    @Post("/mock-data/segments/tasks")
+    public HttpResponse generateMockSegments(
+        @Default("0") @Param("size") int size,
+        @Default("0") @Param("qps") int qps,
```
hanahmily, line 61: We need a time window param here. The generated segments should go into this window. 
If we pass the time window, the size could be 0.

oap-server/server-tools/data-generator/src/main/java/org/apache/skywalking/restapi/SegmentGeneratorHandler.java
```diff
@@ -0,0 +1,124 @@
...
+                MediaType.PLAIN_TEXT_UTF_8,
+                "No such request: %s", requestId);
+        }
+        log.info("Cancelling request: {}", requestId);
+        future.cancel(true);
+        return HttpResponse.of(HttpStatus.OK);
+    }
+
+    @ProducesJson
+    @Get("/mock-data/segments/tasks")
```
hanahmily, line 120: We should add an endpoint to expose the progress of a request. It's about how many tasks are completed, how many errors, etc.

oap-server/server-tools/data-generator/src/main/java/org/apache/skywalking/restapi/SegmentGeneratorHandler.java
```diff
@@ -0,0 +1,124 @@
...
+            if (f.cause() != null) {
+                log.error(f.cause().getMessage(), f.cause());
+            }
+        });
+
+        return HttpResponse.of(MediaType.PLAIN_TEXT, requestId);
+    }
+
+    @Delete("/mock-data/segments/tasks")
+    public HttpResponse cancelRequest(@Param("requestId") String requestId) {
```
hanahmily, line 106: Can we clear all requests more than sending several requests to this endpoint?

hanahmily (commented):
It's great to see this tool will come. I submit some feats that will give us a better experience in testing.

oap-server/server-tools/data-generator/src/main/java/org/apache/skywalking/restapi/SegmentGeneratorHandler.java
kezhenxu94: > We need a time window param here. The generated segments should go into this window.
> If we pass the time window, the size could be 0.

You should set the time window by specifying the time range (`min`, `max`) in the template. I'm not using the current timestamp as the segment/logs timestamp.

```
  "startTime": {
      "type": "randomInt",
      "min": "1647868381000",
      "max": "1647878402000"
  },
```

wu-sheng: I think documenting this should be done as it is important for time series telemetry data

oap-server/server-tools/data-generator/src/main/java/org/apache/skywalking/restapi/SegmentGeneratorHandler.java
kezhenxu94: > We should add an endpoint to expose the progress of a request. It's about how many tasks are completed, how many errors, etc.

If you specify the `size`, there might be a progress, but if you specify `qps`, there is no progress, it generates the data endlessly until you cancel the task

wu-sheng: endless could have a progress for now, I think?

wu-sheng: Progress is just a statistic of generated data, not has to be a completion rate.

oap-server/server-tools/data-generator/src/main/java/org/apache/skywalking/restapi/SegmentGeneratorHandler.java
hanahmily: I noticed the type is `randomInt`. Does it mean the timestamps are random data points between min and max? Can you support generating monotonic time? The monotonically increasing timestamp is very important to banyandb's data model.

oap-server/server-tools/data-generator/src/main/java/org/apache/skywalking/restapi/SegmentGeneratorHandler.java
hanahmily: `qps` model could show a success/error ratio instead of a simple progress bar.

hanahmily: The response could be 

```json
{
"total":"int"
"running":"int"
"errors": "int"
}
```

oap-server/server-tools/data-generator/src/main/java/org/apache/skywalking/restapi/SegmentGeneratorHandler.java
wu-sheng: monotonic as increasing should be more real. But ramdon should still be used for second(same day, hour, minute),because the timestamp of source services are never 100% sync, and segments finished unmonotonic slightly.

lujiajing1126:
Similar Question like Wu Sheng's 5th one:

Is it possible for us to specify how many spans will share the same `trace_id` since this is specially treated as a global index in BanyanDB.

wu-sheng:
> Is it possible for us to specify how many spans will share the same trace_id since this is specially treated as a global index in BanyanDB.

I think for BanyanDB, we care more about how many segments.

lujiajing1126:
> > Is it possible for us to specify how many spans will share the same trace_id since this is specially treated as a global index in BanyanDB.
> 
> I think for BanyanDB, we care more about how many segments.

Oh, right, a typo. I meant `Segment` since all `Span`s are encoded as binary.

oap-server/server-tools/data-generator/src/main/java/org/apache/skywalking/generator/ListGenerator.java
```diff
@@ -0,0 +1,60 @@
...
+            .mapToObj($ -> item)
+            .collect(Collectors.toList());
+    }
+
+    @Override
+    public void reset() {
+        ((Generator<?>) item).reset();
+    }
+
+    @Data
```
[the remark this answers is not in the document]
kezhenxu94: @sonatype-lift ignore

oap-server/server-tools/data-generator/src/main/java/org/apache/skywalking/restapi/SegmentGeneratorHandler.java
kezhenxu94: > The response could be
> 
> ```json
> {
> "total":"int"
> "running":"int"
> "errors": "int"
> }
> ```

There is no `running` and `errors`, data are generated and directly fed to the receiver (asynchronously), there is no response or callback from the receiver so there is no status (success / failure) of the data, all we have is the total number of generated data

hanahmily: all right. `running` and `error` are out. Let's focus on the `total` now

docs/en/guides/backend-data-generator.md
```diff
@@ -0,0 +1,212 @@
+# Mock data generator for testing
```
wu-sheng, line 1: Consider adding this to the menu?

docs/en/guides/segment-template.json
```diff
@@ -0,0 +1,144 @@
...
+            "error": {
+                "type": "fixedBool",
+                "value": "false"
+            },
+            "tags": {
+                "type": "randomList",
+                "size": 5,
+                "item": {
+                    "key": {
+                        "type": "randomString",
```
wu-sheng, line 126: Could we have `stringCollection[str1, str2, str3]`? Don't have to be in this PR, but would be a good feature for testing.

wu-sheng (approved):
Generally good, wait for @hanahmily. We could use this to run, for now, some more enhancements could be added later.

wu-sheng:
```
FILE: ./docs/en/setup/backend/backend-data-generator.md
  [✖] segment-template.json
  [✖] logs-template.json


  ERROR: 2 dead links found!
  5 links checked.
  [✖] segment-template.json → Status: 400
  [✖] logs-template.json → Status: 400
```

It seems somehow 400 for JSON files.

docs/en/setup/backend/logs-template.json
```diff
@@ -0,0 +1,82 @@
...
+    },
+    "content": {
+        "type": "randomString",
+        "length": "10",
+        "prefix": "test",
+        "letters": true,
+        "numbers": true
+    },
+    "error": {
+        "type": "randomBool"
```
wu-sheng, line 59: `randomBool` is not good enough for `error` in testing. We need specific success rate

```suggestion
        "type": "randomBool"
        "successRate": "90" # 90% is successful
```

docs/en/setup/backend/logs-template.json
```diff
@@ -0,0 +1,82 @@
...
+    },
+    "serviceInstanceName": {
+        "type": "randomString",
+        "length": "20",
+        "prefix": "test_svc_inst_name_",
+        "letters": true,
+        "numbers": true
+    },
+    "endpointName": {
+        "type": "randomString",
```
wu-sheng, line 22: For `randomString` using in service, instance and endpoint, we should have the number of candidates.

kezhenxu94: > For `randomString` using in service, instance and endpoint, we should have the number of candidates.

You can just specify `domainSize` to limit the count of candidates

wu-sheng: FYI @hanahmily Please notice this when you use this tool

wu-sheng (approved):
Merge the current version for now.

PR #8727, opened 2022-03-23
Format change logs in a better way.

wu-sheng:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -76,7 +76,7 @@ Release Notes.
 * Add OpenFunction component ID(5013).
 * Expose configuration `responseTimeout` of ES client.
 * Support datasource metric analysis.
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -99,8 +99,8 @@ Release Notes.
 * Expose more ElasticSearch health check related logs to help to
   diagnose `Health check fails. reason: No healthy endpoint`.
 * Add source `event` generated metrics to SERVICE_CATALOG_NAME catalog.
-* [Breaking Change] Deprecate `All` from OAL source.
-* [Breaking Change] Remove `SRC_ALL: 'All'` from OAL grammar tree.
+* [**Break Change**] Deprecate `All` from OAL source.
```
BFergerson, line 102: @wu-sheng, out of curiosity why change to "Break Change" from "Breaking Change"? The latter made more sense.

wu-sheng: I was trying to change in another way

PR #8739, opened 2022-03-25
Trim tag's key and value in log query.

absorprofess:

### DIFF
--- a/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/LogQuery.java
+++ b/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/LogQuery.java
@@ -19,9 +19,7 @@
 package org.apache.skywalking.oap.query.graphql.resolver;
 
 import graphql.kickstart.tools.GraphQLQueryResolver;
-
 import java.io.IOException;
-
 import org.apache.skywalking.oap.server.core.CoreModule;
 import org.apache.skywalking.oap.server.core.UnexpectedException;
 import org.apache.skywalking.oap.server.core.query.LogQueryService;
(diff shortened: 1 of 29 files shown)


wu-sheng:
This is fine to change, please update changes.md in the root. 
We need that for releasing.

wu-sheng (approved):
@absorprofess I updated the change log for you. Please do that in next time. 

Waiting for CI.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/LogQuery.java
```diff
@@ -61,7 +61,10 @@ public Logs queryLogs(LogQueryCondition condition) throws IOException {
             endSecondTB = condition.getQueryDuration().getEndTimeBucketInSec();
         }
         Order queryOrder = isNull(condition.getQueryOrder()) ? Order.DES : condition.getQueryOrder();
-
+        condition.getTags().forEach(tag -> {
```
wu-sheng, line 64: I think tags could be NULL? All log relative e2e tests are failing.

wu-sheng (changes requested):

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/LogQuery.java
```diff
@@ -62,20 +65,26 @@ public Logs queryLogs(LogQueryCondition condition) throws IOException {
         }
         Order queryOrder = isNull(condition.getQueryOrder()) ? Order.DES : condition.getQueryOrder();
         condition.getTags().forEach(tag -> {
-            tag.setKey(tag.getKey().trim());
-            tag.setValue(tag.getValue().trim());
+            if (tag != null) {
```
wu-sheng, line 68: You should check the error logs, I am not sure it is tags or tag is null. But I prefer tags.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/LogQuery.java
```diff
@@ -62,20 +65,26 @@ public Logs queryLogs(LogQueryCondition condition) throws IOException {
...
+                condition.getServiceId(),
+                condition.getServiceInstanceId(),
+                condition.getEndpointId(),
+                condition.getRelatedTrace(),
+                condition.getPaging(),
+                queryOrder,
+                startSecondTB, endSecondTB,
+                condition.getTags(),
+                condition.getKeywordsOfContent(),
+                condition.getExcludingKeywordsOfContent()
```
wu-sheng, line 87: These should not be changed.

wu-sheng (approved):
Once e2e tests could pass, LGTM

PR #8747, opened 2022-03-25
Enhance endpoint dependency analysis to support cross threads cases.

wu-sheng:

* [**Breaking Change**] Remove **Deprecated responseCode** from sources, including Service, ServiceInstance, Endpoint
* Enhance endpoint dependency analysis to support cross threads cases. Refactor span analysis code structures. 
___

In our showcase and Java agent scenarios, there are several cross-thread tracing cases, we were not generating endpoint dependency if ref exists in local and exit span. In this PR, I am adding this missing part back.
Also, in case the downstream endpoint of the dependency is another local/exit span's operation name, which was not generating endpoint traffic before, I add traffic generation of both sides' endpoints of dependency.
Another important point that drives me to fix this is, in the new Booster UI, metrics of endpoint dependency are shown in a separate dashboard, which requires the endpoints of both sides existing in meta data, otherwise, nothing shows up.

For e2e, I updated the Java agent to the latest.

@kezhenxu94 @wankai123  Please verify this would enhance the endpoint dependency on the showcase.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -139,6 +139,8 @@ NOTICE, this sharding concept is NOT just for splitting data into different data
 
 * Support ElasticSearch template mappings `properties parameters` and `_source` update.
 * Implement the eBPF profiling query and data collect protocol.
+* [**Breaking Change**] Remove **Deprecated responseCode** from sources, including Service, ServiceInstance, Endpoint
+* Enhance endpoint dependency analysis to support cross threads cases. Refactor span analysis code structures. 
 
 #### UI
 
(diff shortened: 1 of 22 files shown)


wankai123 (approved):

wankai123 (changes requested):
Needs more verify

wu-sheng:
* Remove `isNotNormal` service requirement when use alias to merge service topology from client side.

The above bug exists for nearly a year, when we add `isNormal` in v8, we should not add this condition. Peer alias should always be used for building topology from client-side relationship.

wankai123 (approved):

PR #8755, opened 2022-03-27
Refactor IoTDB storage plugin and bump up iotdb-session to 0.12.5

LIU-WEI-git:
Refactor IoTDB storage plugin referring to PR#8684, add IoTDBDataConverter and fix ModifyCollectionInEnhancedForLoop bug. Bump up iotdb-session to 0.12.5.


CHANGES.md
```diff
@@ -124,7 +124,8 @@ Release Notes.
 * Add complexity limitation for GraphQL query to avoid malicious query.
 * Add `Column.shardingKeyIdx` for column definition for BanyanDB.
 * Fix the configuration of `Aggregation` and `GC Count` metrics for oap self observability
-
+* Refactor IoTDB storage plugin, add IoTDBDataConverter and fix ModifyCollectionInEnhancedForLoop bug.
```
wu-sheng, line 127: Please move this to the end of backend change. This is separating  one change log.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBTopologyQueryDAO.java
```diff
@@ -39,6 +39,7 @@
 import org.apache.skywalking.oap.server.core.source.DetectPoint;
 import org.apache.skywalking.oap.server.core.storage.query.ITopologyQueryDAO;
 import org.apache.skywalking.oap.server.storage.plugin.iotdb.IoTDBClient;
+import org.apache.skywalking.oap.server.storage.plugin.iotdb.utils.IoTDBUtils;
 
 @Slf4j
 @RequiredArgsConstructor
```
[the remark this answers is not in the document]
wu-sheng: @jlstephens89 I think you have added `SameNameButDifferent` in ignoreRules, but this is still showing up.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBTopologyQueryDAO.java
```diff
@@ -254,7 +254,7 @@
                 RowRecord rowRecord = wrapper.next();
                 List<Field> fields = rowRecord.getFields();
                 Call.CallDetail call = new Call.CallDetail();
-                List<String> layerNames = Splitter.on(IoTDBClient.DOT)
+                List<String> layerNames = Splitter.on(IoTDBClient.DOT + "\"")
```
wu-sheng, line 257: What is the splitter in the database?

wu-sheng: Could you share some examples? I am a little not following yet.

wu-sheng:
https://github.com/apache/skywalking/pull/8761 is going to be merged, please check its changes and new uses of `MultiGet`.

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBTopologyQueryDAO.java
wu-sheng: Is this index mechanism working everywhere? Be careful for trace/log's tags, or endpoint name. It is common to have `.` and `"` from there. I think IoTDB community should improve this, it is hard to choose and separate fields in multi values index.

wu-sheng (approved):
I am good with this polish. https://github.com/apache/skywalking/pull/8755#discussion_r838313013 is only a left question for IoTDB community. Once this conflict happens, the feature is broken from our user perspective

oap-server/server-storage-plugin/storage-iotdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/iotdb/query/IoTDBTopologyQueryDAO.java
wu-sheng: We did add the endpoint name and instance name because we are concerned about `too many candidates` for IoTDB. But in real cases, we may face performance issues when user wants to find an endpoint from 1m size data set.

wu-sheng: Anyway, we have known, IoTDB as storage is not used in the product mostly for now, because we didn't receive any report about this feature.

PR #8761, opened 2022-03-28
Support eBPF Profiling E2E

mrproliu:
1. Support the eBPF Profiling E2E
2. Fix the query in the storage
3. Combine trace profiling with eBPF profiling E2E under one directory. 
4. Let `multiGet` could query without tag value in the `InfluxDB` storage plugin, it could help on the eBPF Profiling GraphQL query. 


mrproliu:
@dmsolr Please help to verify the `multiGet` in the InfluxDB plugin. 
If we only have the ID and want to use `multiGet` for query traffic, we could degradation to only using field(ID field) without tag(serviceId field).

.github/workflows/skywalking.yaml
```diff
@@ -385,16 +385,27 @@ jobs:
...
+          - name: Trace Profiling ES
+            config: test/e2e-v2/cases/profiling/trace/es/e2e.yaml
+          - name: Trace Profiling MySQL
+            config: test/e2e-v2/cases/profiling/trace/mysql/e2e.yaml
+          - name: Trace Profiling InfluxDB
+            config: test/e2e-v2/cases/profiling/trace/influxdb/e2e.yaml
+          - name: Trace Profiling IoTDB
+            config: test/e2e-v2/cases/profiling/trace/iotdb/e2e.yaml
+
+          - name: EBPF Profiling H2
```
wu-sheng, line 399:
```suggestion
          - name: eBPF Profiling H2
```

Unless in coding, I think we could follow the upper/lower cases according to the official. `eBPF`
See https://ebpf.io/

test/e2e-v2/cases/profiling/ebpf/sqrt.go
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+
+package main
+
+import (
+	"math"
+)
+
+//go:noinline
+func sqrtInGolang() {
```
wu-sheng, line 24: In e2e, we only profile golang app, right?

wu-sheng (approved):
Generally good for others. Wait fo @kezhenxu94 and @dmsolr to confirm

test/e2e-v2/cases/profiling/ebpf/sqrt.go
mrproliu: Yes, In the backend, I think our main focus is on storage is enough. in the rover side, we could profiling with other language.

wu-sheng: Agree, just to recheck.

kezhenxu94 (approved):

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/EBPFProfilingTaskQuery.java
```diff
@@ -122,7 +122,7 @@
             task.setProcessName((String) values.get(6));
             task.setTaskStartTime((long) values.get(7));
             task.setTriggerType(EBPFProfilingTriggerType.valueOf((int) values.get(8)));
-            task.setFixedTriggerDuration((long) values.get(9));
+            task.setFixedTriggerDuration(((Number) values.get(9)).longValue());
             task.setTargetType(EBPFProfilingTargetType.valueOf((int) values.get(10)));
             task.setCreateTime((long) values.get(11));
             task.setLastUpdateTime((long) values.get(12));
```
dmsolr, line 128: I think these types should be unreliable. need to change like Line 125
pls recheck

oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/query/EBPFProfilingDataQuery.java
```diff
@@ -87,7 +87,7 @@
             } else {
                 data.setStacksBinary(new byte[0]);
             }
-            data.setStackDumpCount((Long) values.get(5));
+            data.setStackDumpCount(((Number) values.get(5)).longValue());
             data.setUploadTime((Long) values.get(6));
```
dmsolr, line 91: this line is also fixed.

mrproliu: Updated.

wu-sheng (approved):

dmsolr (approved):

PR #8774, opened 2022-03-31
Add `Error Prone Annotations` dependency to suppress warnings

wu-sheng:

* Add `Error Prone Annotations` dependency to suppress warnings, which are not errors.
* Fix some issues reported by lift, https://lift.sonatype.com/results/github.com/apache/skywalking/01FZET2E227JTF4679DBEFBVBF

### DIFF
--- a/.lift/config.toml
+++ b/.lift/config.toml
@@ -2,7 +2,9 @@ jdk11 = true
 ignoreRules = [ "THREAD_SAFETY_VIOLATION", "MissingSummary", "SameNameButDifferent", "NonAtomicVolatileUpdate", "EqualsGetClass"]
 setup = ".lift/setup.sh"
 ignoreFiles = """
-skywalking-ui/
+skywalking-ui/**
 **/target/**
-/dist-material/release-docs/
+/dist-material/release-docs/**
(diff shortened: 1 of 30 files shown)


kezhenxu94:
> * Add `Error Prone Annotations` dependency to suppress warnings, which are not errors.

Can you elaborate why the dependency can suppress warnings and what warnings it suppresses? I can find related info in the link

wu-sheng:
Dependency can't, I am using some annotations to do.

kezhenxu94 (commented):
- We don't need error prone dependency, a simple built-in annotation `@SuppressWarnings("unused")` can do the trick.
- The error prone dependencies don't need to be runtime dependency, it can be `provided`

wu-sheng:
I just saw it is in the dependency path, what do you prefer? Remove it?

kezhenxu94 (approved):
> I just saw it is in the dependency path, what do you prefer? Remove it?

OK, I thought it was newly introduced as I saw new `<dependency>` in `pom.xml`, since it's been there I'm ok to keep it as is.

Fix conflict and it LGTM

kezhenxu94 (approved):

PR #8776, opened 2022-03-31
Add the Self Observability template of SkyWalking Satellite

mrproliu:

image

### DIFF
--- a/oap-server/server-starter/src/main/resources/ui-initialized-templates/so11y_satellite/so11y-root.json
+++ b/oap-server/server-starter/src/main/resources/ui-initialized-templates/so11y_satellite/so11y-root.json
@@ -22,7 +22,7 @@
       "children":[
         {
           "x":8,
-          "y":0,
+          "y":2,
           "w":8,
           "h":14,
           "i":"1",
@@ -48,7 +48,7 @@
         },
         {
           "x":0,
-          "y":0,
+          "y":2,
           "w":8,
           "h":14,
           "i":"0",
@@ -74,7 +74,7 @@
         },
         {
           "x":16,
-          "y":0,
+          "y":2,
           "w":8,
           "h":14,
           "i":"2",
@@ -104,7 +104,7 @@
         },
         {
           "x":0,
-          "y":14,
+          "y":16,
           "w":12,
           "h":14,
           "i":"3",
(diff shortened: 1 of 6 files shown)


wu-sheng:
You don't require group anymore, as layer is provided. Please adjust MAL script.

oap-server/server-starter/src/main/resources/ui-initialized-templates/so11y_satellite/so11y-root.json
```diff
@@ -0,0 +1,236 @@
...
+          "graph":{
+            "type":"Line",
+            "step":false,
+            "smooth":false,
+            "showSymbol":false,
+            "showXAxis":true,
+            "showYAxis":true
+          },
+          "standard":{
+            "unit":"Count"
```
wankai123, line 25: The `standard` has been abandoned, if not `Card` chart, the unit should type on `title` directly. Like `Connection Count (Count)`.

wu-sheng: @wallezhang Please try to set up a new one? This may be caused by old imported template. 
If not, we need a UI fix.

wankai123: just remove it manually, re-import it and export it could fix

mrproliu: I just make the standard as empty.

> re-import it and export it could fix

I have tried, but the `unit` field still exists.

wu-sheng:
You need a text widget on the the top of the page, links to a doc,which is missing about self observability of satellite.
Could you fix the doc and add a link and text here?

Text could be
> Satellite: an open-source agent designed for the cloud-native infrastructures, which provides a low-cost, high-efficient, and more secure way to collect telemetry data. It is the recommended load balancer for telemetry collecting.

wu-sheng:
Could you update screenshots?

mrproliu:
> You need a text widget on the the top of the page, links to a doc about self observability of satellite. Could you fix the doc and add a link and text here?

Updated and add the screenshot. add the link to: https://skywalking.apache.org/docs/main/latest/en/setup/backend/backend-load-balancer/

wu-sheng:
@wankai123 Please take a look.

wankai123 (approved):
LGTM

Issue #8791, opened 2022-04-02
[Doc] Add agent-Satellite compatibility version matrix


kezhenxu94:
https://github.com/apache/skywalking/blob/f8feaa97b7aba577551e2db092a3c29306127eed/docs/en/setup/service-agent/agent-compatibility.md?plain=1#L7-L13

I think we need another matrix like this to indicate what versions of the language agents can connect to what versions of Satellite

| Satellite |Java|Python|NodeJS|LUA|Kong|Browser Agent|Rust| 
 ----------- | ---------- | --------- | --------- |--------- |--------- |--------- |--------- |
 8.0.1 - 8.1.0 | 8.0.0 - 8.3.0 | < = 0.6.0| < = 0.3.0 | All | All | No | All | 
 8.2.0 - 8.3.0 | 8.0.0 - 8.3.0 | < = 0.6.0| < = 0.3.0 | All | All | All | All |
 8.4.0 - 8.8.1 | \> = 8.0.0 | All | All | All | All | All | All | 
 8.9.0+ | \> = 8.0.0 | All | All | All | All | All | All |
 9.0.0+ | \> = 8.0.0 | All | All | All | All | All | All |

wu-sheng:
@mrproliu I think we at least require a minimal version. We faced NodeJS agent incompatibility in testing today.

mrproliu:
Two things to note:
1. Update document, add a new matric to document which native agent support transport protocol to the satellite. 
2. Update the Satellite, let it could support transport compat protocol from the agent.

wu-sheng:
@kezhenxu94 Older releases should be supported by Satellite(latest) through https://github.com/apache/skywalking-satellite/pull/104 , could you give a try?

PR #8792, opened 2022-04-02
Add documentation of SkyWalking Satellite with native agent compatibility

mrproliu:
Follow https://github.com/apache/skywalking/issues/8791


wu-sheng:
I think this document belongs to the Satellite project? And for supporting more versions, we need another release?

mrproliu:
> I think this document belongs to the Satellite project? 

Sure, I would add it into the Satellite project.

> And for supporting more versions, we need another release?

Yes, we need to do two thing:
1. Support HTTP message receiver for Nginx, Browser, etc.
2. Support Compat protocol for the oldest agent version.

mrproliu:
Closed, add it into the satellite project.

kezhenxu94:
> 1. Support HTTP message receiver for Nginx, Browser, etc.

Do we need this? Satellite is for load balancing gRPC traffic, and http protocol uses a different port so why users will connect to Satellite if they are using http protocol?

wu-sheng:
For now, I would say not, our special HPA/load-balance mechanism is caused by gRPC streaming and connection management mechanism. In HTTP, it is not an issue.

mrproliu:
Sure, I just think we could unify the traffic entrance. Then we could do more things, like sampling, data pre-aggregated, etc.

wu-sheng:
Yes, only we pass the POC pre-aggregation, and could prove pre-aggregation really makes sense. Right now, I think we will not work on that in short term, more works on BanyanDB seem important. One step at a time.

PR #8828, opened 2022-04-06
ADD kafka consumers number default = 1,enhance consumption capacity

chenhaipeng:
### Add kafka consumers number


wu-sheng:
So, generally, you want an OAP server having multiple consumers?

chenhaipeng:
> So, generally, you want an OAP server having multiple consumers?

It depends on the speed of producers and consumers. If MQ accumulates, increase the number of consumers and reduce delay

wu-sheng:
`application.yml`, `configuration-vocabulary.md`, `changes.md` should be updated as well.

dmsolr (dismissed):
LGTM, Thanks.

wu-sheng:
@chenhaipeng A question, how do these consumers avoid consuming duplicated messages? I am away from using Kafka for a very long time.

dmsolr:
@chenhaipeng Hi, I notice we need to configure the `consumePartitions` on cluster mode of OAP. That is very unfriendly. Would you like to improve the feature like this PR?

chenhaipeng:
> @chenhaipeng A question, how do these consumers avoid consuming duplicated messages? I am away from using Kafka for a very long time.

consuming duplicated messages，Depends on different groupId，this feature only add consumer instances,will not cause duplicate messages

chenhaipeng:
> @chenhaipeng Hi, I notice we need to configure the `consumePartitions` on cluster mode of OAP. That is very unfriendly. Would you like to improve the feature like this PR?

consumePartitions  is original feature,this feature only add consumer instances,there is no change in any of the original features

wu-sheng:
> > @chenhaipeng Hi, I notice we need to configure the `consumePartitions` on cluster mode of OAP. That is very unfriendly. Would you like to improve the feature like this PR?
> 
> consumePartitions is original feature,this feature only add consumer instances,there is no change in any of the original features

The point here I think is, this may be not needed anymore due to multiple consumers are available. That thing was added to manually set consumer to avoid duplicated consuming.

chenhaipeng:
> > > @chenhaipeng Hi, I notice we need to configure the `consumePartitions` on cluster mode of OAP. That is very unfriendly. Would you like to improve the feature like this PR?
> > 
> > 
> > consumePartitions is original feature,this feature only add consumer instances,there is no change in any of the original features
> 
> The point here I think is, this may be not needed anymore due to multiple consumers are available. That thing was added to manually set consumer to avoid duplicated consuming.

This is indeed a problem. The two configuration parameters, consumePartitions and consumers, are a bit conflicting. Maybe I didn't enable isSharding and consumePartitions in the production environment, so I didn't find this problem. So how to solve this problem is better

wu-sheng:
I am not familiar with today's Kafka now, please correct me if I am wrong.

When you use multiple OAP servers, and every server has multiple consumers for one topic, do you need any extra configurations?

chenhaipeng:
> 



> I am not familiar with today's Kafka now, please correct me if I am wrong.
> 
> When you use multiple OAP servers, and every server has multiple consumers for one topic, do you need any extra configurations?
> 
> And how the consumers(if multiple) are assigned to different topics?



> I am not familiar with today's Kafka now, please correct me if I am wrong.
> 
> When you use multiple OAP servers, and every server has multiple consumers for one topic, do you need any extra configurations?
> 
> And how the consumers(if multiple) are assigned to different topics?

no need,this feature is added just to enhance the consumption power of a single OAP

wu-sheng:
I am not talking about this PR, I want to make this to make Kafka fetchers work better.

chenhaipeng:
> I am not familiar with today's Kafka now, please correct me if I am wrong.
> 
> When you use multiple OAP servers, and every server has multiple consumers for one topic, do you need any extra configurations?
> 
> And how the consumers(if multiple) are assigned to different topics?
The configuration in kafka-fetcher works on multiple topics. I don't have any special configuration for individual topics online, but there is a configuration enableTopics (not in this PR) to control whether a topic is enabled, and the other is my online The characteristics of the modified kafka-fetcher, the link becomes OAP->MQ->OAP, the purpose is to reduce the configuration of kafka on the agent side, and put the control of kafka on the platform side

chenhaipeng:
> I am not talking about this PR, I want to make this to make Kafka fetchers work better.

The configuration in kafka-fetcher works on multiple topics. I don't have any special configuration for individual topics online, but there is a configuration enableTopics (not in this PR) to control whether a topic is enabled, and the other is my online The characteristics of the modified kafka-fetcher, the link becomes OAP->MQ->OAP, the purpose is to reduce the configuration of kafka on the agent side, and put the control of kafka on the platform side

wu-sheng:
@dmsolr Could you take a step in involved in this? I feel like @chenhaipeng seems to misunderstand my point.

To @dmsolr @chenhaipeng, I don't want simply change this fetcher in this way, I want the consumer could be better organized. Could we have a discussion here about this?

wu-sheng:
Please follow https://github.com/apache/skywalking/pull/8857, I hope that could be landed first.

wu-sheng:
@chenhaipeng I merged that one, if your proposal still makes sense, please update.

chenhaipeng:
> @chenhaipeng I merged that one, if your proposal still makes sense, please update.

I have updated to support multiple instances of consumers

chenhaipeng:
> @chenhaipeng I merged that one, if your proposal still makes sense, please update.

I have updated to support multiple instances of consumers

wu-sheng:
@dmsolr PTAL

wu-sheng:
@chenhaipeng e2e tests for Kafka receivers fail, please check the codes locally, and test.

chenhaipeng:
> @chenhaipeng e2e tests for Kafka receivers fail, please check the codes locally, and test.

I've fixed the bug

dmsolr (approved):
LGTM

docs/en/setup/backend/configuration-vocabulary.md
```diff
@@ -215,6 +215,7 @@ core|default|role|Option values: `Mixed/Receiver/Aggregator`. **Receiver** mode
 | - | - | groupId | A unique string that identifies the consumer group to which this consumer belongs.| - | skywalking-consumer                                                     |
 | - | - | createTopicIfNotExist | If true, this creates Kafka topic (if it does not already exist). | - | true                                                                    |
 | - | - | partitions | The number of partitions for the topic being created. | SW_KAFKA_FETCHER_PARTITIONS | 3                                                                       |
+| - | - | consumers | The number of consumers create.| SW_KAFKA_FETCHER_CONSUMERS | 1                                                                      |
```
Superskyyy, line 218:
```suggestion
| - | - | consumers | The number of consumers to create. | SW_KAFKA_FETCHER_CONSUMERS | 1                                                                      |
```

wu-sheng (approved):

PR #8848, opened 2022-04-11
Fix Zipkin receiver: Decode spans error, missing `Layer` for V9 and wrong time bucket for generate Service and Endpoint.

wankai123:
- [X] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/changelog/docs/en/changes/changes.md).

Upgrade Zipkin to `2.23.16` for `SpanBytesDecoder` support `ByteBuffer`.

### DIFF
--- a/tools/dependencies/known-oap-backend-dependencies.txt
+++ b/tools/dependencies/known-oap-backend-dependencies.txt
@@ -150,7 +150,7 @@ swagger-annotations-1.6.3.jar
 tsfile-0.12.5.jar
 vavr-0.10.3.jar
 vavr-match-0.10.3.jar
-zipkin-2.9.1.jar
+zipkin-2.23.16.jar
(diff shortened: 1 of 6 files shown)


kezhenxu94 (dismissed):

wu-sheng (dismissed):

tools/dependencies/known-oap-backend-dependencies.txt
```diff
@@ -150,7 +150,7 @@ swagger-annotations-1.6.3.jar
 tsfile-0.12.5.jar
 vavr-0.10.3.jar
 vavr-match-0.10.3.jar
-zipkin-2.9.1.jar
+zipkin-2.23.16.jar
```
kezhenxu94, line 153: So please check the version in LICENSE file and update there too

PR #8851, opened 2022-04-11
[CI] set up Java explicitly and use the built-in cache mechanism

kezhenxu94:


wu-sheng (approved):

kezhenxu94:
The failures were so strange that I am even not sure whether this fixes the problem or not. Here are my observations:

- https://github.com/apache/skywalking/runs/5966401737?check_suite_focus=true#step:6:25 in the failure case, I list the files under `~/.m2/repository` and there indeed is no 9.1.0-SNAPSHOT versions.
- Also in the same case 👆 when I manually upload the `~/.m2/repository` folder and download it to my local machine, those `9.1.0-SNAPSHOT` files are there. https://github.com/apache/skywalking/actions/runs/2146307624 see the artifacts `maven-cache`

This PR is just an attempt after I've run out of all my ideas 🤦🏻‍♂️

Issue #8845, opened 2022-04-10
[Bug] Java client failed to handle duplicate data points


hanahmily:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar issues.

### Apache SkyWalking Component

BanyanDB (apache/skywalking-banyandb)

### What happened

BanyanDB identifies unique data points by their metadata(name and group), entity id, and timestamp. For points with the same of them, the database server uses the new value to override the old one. 

Since the java client accepts the milliseconds as the timestamp's unit from the [document](https://github.com/apache/skywalking-banyandb-java-client#stream-1), we should handle the same points on this part. 

The possible solution is the influxdb module: https://github.com/apache/skywalking/blob/master/oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/RecordDAO.java#L53

### What you expected to happen

None

### How to reproduce

None

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wu-sheng:
> BanyanDB identifies unique data points by their metadata(name and group), entity id, and timestamp. For points with the same of them, the database server uses the new value to override the old one.

This is what we expect, and trace segment ID would be duplicated. What is the issue about exactly?

wu-sheng:
> The possible solution is the influxdb module: https://github.com/apache/skywalking/blob/master/oap-server/server-storage-plugin/storage-influxdb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/influxdb/base/RecordDAO.java#L53

This is really not a good example to refer. I am talking with @dmsolr to get this out.

wu-sheng:
I have confirmed this is a bug in InfluxDB storage implementation. I have asked @dmsolr to adjust codes to fix this.
So, what is the case of BanyanDB actually?

hanahmily:
As I mentioned, the data point id is a combination of metadata, entity_id, and timestamp. If an instance sends several segments in one millisecond, the last one that reaches banyandb will win, and each previous data point is dropped. 

>  I have asked @dmsolr to adjust codes to fix this.

I would like to know how to fix this. banyandb could get benefits from the same solution. @dmsolr Would you please elaborate it a bit in this thread?

[...]

PR #8852, opened 2022-04-11
Support BanyanDB global index for entities

wu-sheng:

* [Refactor] Move SQLDatabase(H2/MySQL/PostgreSQL), ElasticSearch and BanyanDB specific configurations out of column.
* Support BanyanDB global index for entities. Log and Segment record entities declare this new feature.
* Remove unnecessary analyzer settings in columns of templates. Many were added due to analyzer's default value.

Refer https://github.com/apache/skywalking/issues/8845#issuecomment-1094453157 FYI @lujiajing1126


hanahmily (approved):
LGTM

PR #8874, opened 2022-04-14
Fix event can't split service ID into 2 parts

### LINKED ISSUE #8864: [Bug] [Event] Can't split service ID into 2 parts
### Search before asking

### What happened

image

Event metrics could cause this error.

### How to reproduce

`oal/event.oal` could make this happens

### Anything else

The event now is treated as a manual metric source(like an updatable record) and a source of OAL dispatcher.

But in the process, it doesn't declare entity ID, so when it is forwarded into alarm kernel, this `Can't split service ID into 2 parts` raises up.
Meanwhile, I am not sure why the timebucket could not be downsampling in OAL-generated metrics. The only thing I could guess is, somehow, it could be not-set.

The `Can't split service ID into 2 parts` is easy to fix. `@ScopeDefaultColumn.VirtualColumnDefinition(fieldName = "entityId", columnName = "entity_id", isID = true, type = String.class)` should be added into event.

@kezhenxu94 Also, we should consider seriously, whether this is a good practice to use one object being both source and manual metrics. Many methods in here could confuse developers and readers, they are responsible to persistent of metrics, but actually, more other metrics are generated through OAL engine in the runtime.

### Code of Conduct

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #8864 .

### DIFF
--- a/oap-server/analyzer/event-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/event/EventAnalyzerServiceImpl.java
+++ b/oap-server/analyzer/event-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/event/EventAnalyzerServiceImpl.java
@@ -39,7 +39,7 @@ public class EventAnalyzerServiceImpl implements EventAnalyzerService, EventAnal
         final Event.Builder eb = event.toBuilder();
         if (event.getStartTime() <= 0 && event.getEndTime() <= 0) {
             log.warn(
-                "Event start time {} and end time are both invalid, they will be set to current time, eventId: {}",
+                "Event start time {} and end time {} are both invalid, they will be set to current time, eventId: {}",
                 event.getStartTime(),
                 event.getEndTime(),
                 event.getUuid());
--- a/oap-server/analyzer/event-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/event/EventAnalyzerServiceImpl.java
+++ b/oap-server/analyzer/event-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/event/EventAnalyzerServiceImpl.java
@@ -21,11 +21,13 @@ package org.apache.skywalking.oap.server.analyzer.event;
 import java.util.ArrayList;
 import java.util.List;
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.skywalking.apm.network.event.v3.Event;
 import org.apache.skywalking.oap.server.analyzer.event.listener.EventAnalyzerListener;
 import org.apache.skywalking.oap.server.library.module.ModuleManager;
 import org.apache.skywalking.oap.server.analyzer.event.listener.EventAnalyzerListenerFactoryManager;
 
+@Slf4j
 @RequiredArgsConstructor
 public class EventAnalyzerServiceImpl implements EventAnalyzerService, EventAnalyzerListenerFactoryManager {
     private final ModuleManager moduleManager;
@@ -34,8 +36,19 @@ public class EventAnalyzerServiceImpl implements EventAnalyzerService, EventAnal
 
     @Override
     public void analyze(final Event event) {
+        final Event.Builder eb = event.toBuilder();
+        if (event.getStartTime() <= 0 && event.getEndTime() <= 0) {
+            log.warn(
+                "Event start time {} and end time are both invalid, they will be set to current time, eventId: {}",
+                event.getStartTime(),
+                event.getEndTime(),
+                event.getUuid());
+            eb.setStartTime(System.currentTimeMillis());
+            eb.setEndTime(System.currentTimeMillis());
+        }
+
         final EventAnalyzer analyzer = new EventAnalyzer(moduleManager, this);
-        analyzer.analyze(event);
+        analyzer.analyze(eb.build());
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


wu-sheng:
Should we close that issue? Is the time bucket issue addressed?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/Metrics.java
```diff
@@ -102,7 +102,7 @@ public long toTimeBucketInHour() {
         if (isMinuteBucket()) {
             return timeBucket / 100;
         } else {
-            throw new IllegalStateException("Current time bucket is not in minute dimensionality");
+            throw new IllegalStateException("Current time bucket is not in minute dimensionality: " + timeBucket);
```
wu-sheng, line 105: I think you could see the time bucket from index-0? It should mean no value of time bucket, right?

wu-sheng: Is this log really helpful?

kezhenxu94:
> Should we close that issue? Is the time bucket issue addressed?

That was caused by kubernetes event exporter, should be fixed in another repo

wu-sheng:
I suggest you add a fail safe of making sure the time bucket not null. If it is not set correctly, use the server time to fill it. WDYT?

kezhenxu94:
> I suggest you add a fail safe of making sure the time bucket not null. If it is not set correctly, use the server time to fill it. WDYT?

Time bucket is always not null, the startTime and endTime of an event are always not null, they are all `long` primitives.

In this case the problem is that event exporter set it wrongly, with a negative value (like -6795364578871345152), do you think I should "correct" this if users deliberately set it wrong? IMO this would hide something that is really wrong, for example, in this case, the event exporter needs a fix. Adding the fail safe hides this problem

wu-sheng:
I think it is worth to avoid this, the mainly concern is in ElasticSearch case, it could create a new and never removed index, and data could keep increasing, which leads crash.

Time bucket before 19700101 is definitely not legal. 😀

If you don't want to hide error, print warning logs in receiver should be fine.

kezhenxu94:
> I think it is worth to avoid this, the mainly concern is in ElasticSearch case, it could create a new and never removed index, and data could keep increasing, which leads crash.

This makes total sense to me!!

oap-server/analyzer/event-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/event/EventAnalyzerServiceImpl.java
```diff
@@ -34,8 +36,20 @@ public class EventAnalyzerServiceImpl implements EventAnalyzerService, EventAnal
...
+            log.warn("Event start time {} is invalid, it will be set to current time, eventId: {}",
+                event.getStartTime(), event.getUuid());
+            eb.setStartTime(System.currentTimeMillis());
+        }
+        if (event.getEndTime() <= 0) {
+            log.warn("Event end time {} is invalid, it will be set to current time, eventId: {}",
+                event.getEndTime(), event.getUuid());
+            eb.setEndTime(System.currentTimeMillis());
+        }
+
```
wu-sheng, line 50: I remember you mentioned, start time and endpoint could be assigned separatedly in two requests?

This warning maybe trigger false warning?

wu-sheng: The think the rule should be either start or end is legal


And these two variables here are timestamps, not time bucket, right?

kezhenxu94: > The think the rule should be either start or end is legal

Ha, right.

> And these two variables here are timestamps, not time bucket, right?

Yes. They are timestamp here, haven't been converted to time bucket yet,

wu-sheng (approved):

oap-server/analyzer/event-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/event/EventAnalyzerServiceImpl.java
```diff
@@ -34,8 +36,19 @@ public class EventAnalyzerServiceImpl implements EventAnalyzerService, EventAnal
 
     @Override
     public void analyze(final Event event) {
+        final Event.Builder eb = event.toBuilder();
+        if (event.getStartTime() <= 0 && event.getEndTime() <= 0) {
+            log.warn(
+                "Event start time {} and end time are both invalid, they will be set to current time, eventId: {}",
```
wu-sheng, line 42:
```suggestion
                "Event start time {} and end time {} are both invalid, they will be set to current time, eventId: {}",
```

wu-sheng: Miss one?

wu-sheng (approved):

wu-sheng:
One tip, Event toBuilder and then build could be a performance impact. I would not concern it as we should not face too many events concurrently.

Issue #8894, opened 2022-04-17
[Bug] [CI] [Backend and Java agent] Dead link check fails


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar issues.

### Apache SkyWalking Component

OAP server (apache/skywalking)

### What happened

SkyWalking Twitter is https://twitter.com/ASFSkyWalking

The readme doc links to there, but CI reports 404, but actually, it exists.

I tried to upgrade to the newer dead-link check tool, but it seems don't support page hook(`/uri#hook`).

### What you expected to happen

Make dead link check back to work

### How to reproduce

Every CI reports this.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wu-sheng:
FYI, I noticed this tool is GitHub action available now. 

https://github.com/marketplace/actions/markdown-link-check

Superskyyy:
This problem happens to all repos with the ASFSkyWalking Twitter link, I tried to resolve it yesterday but the npm package couldn't work anyhow. Instead using the GHA version of the package magically works... `gaurav-nelson/github-action-markdown-link-check@v1`

Here's a link to a successful run https://github.com/Superskyyy/skywalking-python/runs/6042396915?check_suite_focus=true

I'm not sure if the action is Apache approved, other than that it should work.

wu-sheng:
I think once you submit a pull request, you will know.

wu-sheng:
> I tried to resolve it yesterday but the npm package couldn't work anyhow.

I tried locally too, the tool is working, but it reported `page internal hook` as 404. Maybe some parameter should be set?

wu-sheng:
@kezhenxu94 I didn't close this as Java repo still has issues.

kezhenxu94:
OK, I just saw all CI tasks in Java repo main branch passed but didn't notice the link checker task was not running in main branch, let me fix now

PR #8898, opened 2022-04-18
Upgrade dead-link-checker to check anchors and fix CI

kezhenxu94:
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/8894.

### DIFF
--- a/.dlc.json
+++ b/.dlc.json
@@ -16,6 +16,7 @@
   "fallbackRetryDelay": "1000s",
   "aliveStatusCodes": [
     200,
-    401
+    401,
+    403
   ]
 }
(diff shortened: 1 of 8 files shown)


.dlc.json
```diff
@@ -16,6 +16,7 @@
   "fallbackRetryDelay": "1000s",
   "aliveStatusCodes": [
     200,
-    401
+    401,
+    403
```
wu-sheng, line 20: Should we consider 302?

kezhenxu94: I guess the tool will follow the redirect automatically, this configuration section is for "error code" that should be 4xx and 5xx

wu-sheng: Got it. Do you know why `https://twitter.com/ASFSkyWalking` is a dead link? The browser seems to say the other way.

wu-sheng (approved):

PR #8909, opened 2022-04-19
fix ES flush thread will stop work when flush schedule task have ES connection exception

### LINKED ISSUE #8895: [Bug] ES flush thread will stop work when flush schedule task have exception
### Search before asking

### What happened

when i use skywalking in test enviorment,  i find the trace data write to ES have most delay.sometimes data write to ES will cost 30 minute later.

the trace write to ES  within 20 seconds。

### How to reproduce

ES connectTimeout  set 3000 seconds.

### Anything else

the flush task have exception cause the schedule thread all waiting ,didn't work :
6592F812-C690-479B-85EA-E36EE03EA8CA

so  i add a try catch code in native,the exception is :
39920013-C59B-402B-A14E-D37AE76435D2

how to deal:
1.add try cache code in org.apache.skywalking.library.elasticsearch.bulk.BulkProcessor#flush
2.set more time for ES connection time, response time to  avoid exception.

### Code of Conduct

yangyiweigege:
…ption(#8895)

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #8895 .


wu-sheng:
I think you should be clear about which exception could cause this? And you should follow the PR template, linking to issue and update change log.

yangyiweigege:
> I think you should be clear about which exception could cause this? And you should follow the PR template, linking to issue and update change log.

i have changed the title  and linking to linking to issue and update change log.

yangyiweigege:
> I think you should be clear about which exception could cause this? And you should follow the PR template, linking to issue and update change log.

i have changed the title and  linking to issue and update change log.

wu-sheng:
Change log is a file, this link guides you to there

image

wu-sheng:
**I think you should be clear about which exception could cause this? ** This is a key question. @kezhenxu94 showed concerns in the issue, which should be answered before any code change.

yangyiweigege:
> **I think you should be clear about which exception could cause this?** This is a key question. @kezhenxu94 showed concerns in the issue, which should be answered before any code change.



> **I think you should be clear about which exception could cause this?** This is a key question. @kezhenxu94 showed concerns in the issue, which should be answered before any code change.

Ok, I continue to look at this part of the source code and try to find the reason of exception.

wu-sheng:
Thanks, that is important to know why before changing. Your another bug report seems more solid, you could submit another pull request to fix that first.

yangyiweigege:
i have found the reason of the exception:

code in org.apache.skywalking.library.elasticsearch.bulk.BulkProcessor#doFlush:
```java
  if (batch.isEmpty()) {
            return CompletableFuture.completedFuture(null);
        }

        final CompletableFuture<Void> future = es.get().version().thenCompose(v -> {
            try {
                final RequestFactory rf = v.requestFactory();
                final List<byte[]> bs = new ArrayList<>();
                for (final Holder holder : batch) {
                    bs.add(v.codec().encode(holder.request));
                    bs.add("\n".getBytes());
                }
                final ByteBuf content = Unpooled.wrappedBuffer(bs.toArray(new byte[0][]));
                return es.get().client().execute(rf.bulk().bulk(content))
                         .aggregate().thenAccept(response -> {
                        final HttpStatus status = response.status();
                        if (status != HttpStatus.OK) {
                            throw new RuntimeException(response.contentUtf8());
                        }
                    });
            } catch (Exception e) {
                return Exceptions.throwUnsafely(e);
            }
        });
        future.whenComplete((ignored, exception) -> {
            if (exception != null) {
                batch.stream().map(it -> it.future)
                     .forEach(it -> it.completeExceptionally(exception));
                log.error("Failed to execute requests in bulk", exception);
            } else {
                log.debug("Succeeded to execute {} requests in bulk", batch.size());
                batch.stream().map(it -> it.future).forEach(it -> it.complete(null));
            }
        });
```
in this code, when response is not ok,will throw a  runtimeException,then  see  another code:
org.apache.skywalking.library.elasticsearch.bulk.BulkProcessor#flush
```java
        final CompletableFuture<Void> flush = doFlush(batch);
        flush.whenComplete((ignored1, ignored2) -> semaphore.release());
        flush.join();
```

when CompletableFuture have exception,and did't  deal it ,  the join  method  will  throw exception。 so the flush schedule will stoped .
and i test the join method in demo:
image

wu-sheng:
@kezhenxu94 Please confirm.

wu-sheng:
A reminder, you still don't update change log.

docs/en/changes/changes.md
```diff
@@ -20,6 +20,7 @@
 * Set `SW_QUERY_MAX_QUERY_COMPLEXITY` default value to `1000`
 * Webapp module (for UI) enabled compression.
 * [Breaking Change] Add layer field to event, report an event without layer is not allowed.
+* Fix ES flush thread will stop when flush schedule task have exception.
```
wu-sheng, line 23:
```suggestion
* Fix ES flush thread stops when flush schedule task throws exception, such as ElasticSearch flush failed.
```

kezhenxu94 (approved):

PR #8917, opened 2022-04-21
HttpServer support configure the allowed HTTP methods.

wankai123:
- [X] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/changelog/docs/en/changes/changes.md).

### DIFF
--- a/oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServer.java
+++ b/oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServer.java
@@ -27,6 +27,9 @@ import com.linecorp.armeria.server.docs.DocService;
 import com.linecorp.armeria.server.logging.LoggingService;
 import java.net.InetSocketAddress;
 import java.time.Duration;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Locale;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.skywalking.oap.server.library.server.Server;
@@ -55,7 +58,7 @@ public class HTTPServer implements Server {
             .http1MaxHeaderSize(config.getMaxRequestHeaderSize())
             .idleTimeout(Duration.ofMillis(config.getIdleTimeOut()))
             .decorator(Route.ofCatchAll(), (delegate, ctx, req) -> {
-                if (ctx.method() != HttpMethod.POST) {
+                if (!this.isMethodAllowed(ctx.method())) {
                     return HttpResponse.of(HttpStatus.METHOD_NOT_ALLOWED);
                 }
                 return delegate.serve(ctx, req);
@@ -83,5 +86,10 @@ public class HTTPServer implements Server {
     @Override
     public void start() {
         sb.build().start().join();
+    }
+
+    private boolean isMethodAllowed(HttpMethod method) {
+        List<String> allowedMethods = Arrays.asList(this.config.getAllowedHttpMethods().toUpperCase(Locale.ROOT).split(","));
+        return allowedMethods.contains(method.name().toUpperCase(Locale.ROOT));
     }
 }
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -22,6 +22,7 @@
 * [Breaking Change] Add layer field to event, report an event without layer is not allowed.
 * Fix ES flush thread stops when flush schedule task throws exception, such as ElasticSearch flush failed.
 * Fix ES BulkProcessor in BatchProcessEsDAO was initialized multiple times and created multiple ES flush schedule tasks.
+* HttpServer support configure the allowed HTTP methods. 
 
 #### UI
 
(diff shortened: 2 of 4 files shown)


kezhenxu94 (commented):
Do we need this to be configurable? I remember we disable other methods because of Jetty server CVE. Can you take a look at https://github.com/apache/skywalking/pull/8725/files#diff-b034fcbcccfb37d631c81c5c140b518d5b708a395cfa0b24162f492ddb749ddb and see whether that way works too?

wankai123:
> Do we need this to be configurable? I remember we disable other methods because of Jetty server CVE. Can you take a 

I don't want to set the other allowed method as global, since mostly are `POST` for now.
This configuration needs to be explicitly turned on by configuration, I  think it's no different to configure it in java code.

kezhenxu94:
I mean, if you add other methods than POST you need to enable this otherwise the added methods won't take effect. What's the point to let users set it themself if this is going to be always "POST,GET"? Do you think users will change the configuration to something like "POST,GET,DELETE" if they don't modify the codes (add a DELETE method)? I don't think we want to add random configuration that users might never set themselves.

wu-sheng:
From HTTP server perspective, these configurations should only affect to codes interaction, which is fine. We don't expose these to end users.
But for different HTTP server, it could be different, SkyWalking only exposes POST for native protocol, but Zipkin expose Get for query.

kezhenxu94:
Let trace back to when this was added https://github.com/apache/skywalking/pull/7052 I'd just remove the method check as this was to fix Jetty CVE  and now we don't use Jetty we could just allow all methods.

wankai123:
> I mean, if you add other methods than POST you need to enable this otherwise the added methods won't take effect. 

Make sense. So could we just let this configure in the specific module `configure.java` code, like `ZipkinReceiverConfig.java`
not expose it to the config file(yaml) or env. variables.

wu-sheng:
In my mind, I prefer to generate this setting from all declaration of handlers, and set this when server start. This is not a setting, this is a expose limitation, which is good. 
I know this was introduced by jetty, but this kind of CVE happenned from time to time for any HTTP server.
@wankai123 Let's not check it, and accept this as a part of handler registration, and lock it just before server starts.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/server/HTTPHandlerRegister.java
```diff
@@ -18,8 +18,10 @@
 
 package org.apache.skywalking.oap.server.core.server;
 
+import com.linecorp.armeria.common.HttpMethod;
+import java.util.List;
 import org.apache.skywalking.oap.server.library.module.Service;
 
 public interface HTTPHandlerRegister extends Service {
-    void addHandler(Object httpService);
+    void addHandler(Object httpService, List<HttpMethod> allowedMethods);
```
kezhenxu94, line 26: `allowedMethods` should not be handler level, it should be server level, so I don't think it is a good idea to set in `addHandler` method, it should be set in constructing an `HTTPServer`

wankai123: `allowedMethods` of course the server level, but when the handlers share the server, only themselves know provide which methods. So I think it's should send this info to the server when registering.

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServer.java
```diff
@@ -27,14 +27,19 @@
...
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.skywalking.oap.server.library.server.Server;
+import static java.util.Objects.requireNonNull;
 
 @Slf4j
 public class HTTPServer implements Server {
     private final HTTPServerConfig config;
     private ServerBuilder sb;
+    private final Set<HttpMethod> allowedMethods = new HashSet<>();
```
wu-sheng, line 42: Let's rename `allowedMethods` to `httpMethods`, which are more clear.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/server/HTTPHandlerRegisterImpl.java
```diff
@@ -29,7 +31,7 @@ public HTTPHandlerRegisterImpl(HTTPServer server) {
     }
 
     @Override
-    public void addHandler(Object httpService) {
-        server.addHandler(httpService);
+    public void addHandler(final Object httpService, final List<HttpMethod> allowedMethods) {
```
wu-sheng, line 34:
```suggestion
    public void addHandler(final Object httpService, final List<HttpMethod> httpMethods) {
```

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServer.java
wankai123: That's in HTTP server exactly what it means, and have changed others.

wu-sheng (approved):

PR #8923, opened 2022-04-21
[Critical] Fix the priority mechanism of DataCarrier's MultipleChannelsConsumer failure.

wu-sheng:

@kezhenxu94 Reported this bug to us, and noticed https://github.com/apache/skywalking/pull/8664 applied into the mainstream, the OAP server could report gRPC thread pool is full, rejecting the task, even the traffic is very low, and ElasticSearch is healthy, once #8664 is reverted, the error is gone in the same test scenario.

I did some research about my last change, and found out this bug only could occur when traffic is low.
I was using `channels.getChannelSize() * channels.size()`, and I thought it is the size, and actually `channels.size()` is the size already. So with this wrong usage, the priority is always very low, and hard(maybe never?) to reach over 50. Then the queue is full with a very low priority, not consuming anymore.

The fix is easy, make the priority back to expected value.

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -6,6 +6,9 @@
 
 #### OAP Server
 
+* [Critical] Fix the priority mechanism of DataCarrier's MultipleChannelsConsumer failure. This critical bug could stop
+  message consuming if the data are pushed into queue in small pieces but continuously.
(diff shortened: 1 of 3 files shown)


wu-sheng:
FYI @apache/skywalking-committers This is a critical bug, you all should be aware of.

PR #8926, opened 2022-04-22
Fix the kafka consumer configuration override bug.

wyt:

When I try the configuration below to set different consumer group for oap server, it doesn't work.
Still the default group in the program.

```yaml
kafka-fetcher:
  selector: ${SW_KAFKA_FETCHER:default}
  default:
    bootstrapServers: ${SW_KAFKA_FETCHER_SERVERS:192.168.101.1:9092}
    ......
    kafkaConsumerConfig:
      enable.auto.commit: true
      group.id: test-group-mysql
```

The reason is that there is an order problem when setting in the code:

```java
properties = new Properties();
properties.putAll(config.getKafkaConsumerConfig());
properties.setProperty(ConsumerConfig.GROUP_ID_CONFIG, config.getGroupId());
properties.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, config.getBootstrapServers());
```

I don't know if this is by design or a mistake in itself.
Please confirm.


wu-sheng:
image

This fix is good, please follow this link to update change log.

wyt:
> image
> 
> This fix is good, please follow this link to update change log.

Completed.

wu-sheng:
I think you merged from a wrong branch? 4000 LOC changed.

wyt:
> I think you merged from a wrong branch? 4000 LOC changed.
I don't know, which step may be wrong, what to do next? Close the current pull, resubmit?

wu-sheng:
That would be easier. Choose the current upstream branch carefully.

wu-sheng:
I think your codebase still wrong, check the files tab. Many things are not changed by you.

PR #8929, opened 2022-04-22
Fix the kafka consumer configuration override bug.

wyt:

When I try the configuration below to set different consumer group for oap server, it doesn't work.
Still the default group in the program.

```yaml
kafka-fetcher:
  selector: ${SW_KAFKA_FETCHER:default}
  default:
    bootstrapServers: ${SW_KAFKA_FETCHER_SERVERS:192.168.101.1:9092}
    ......
    kafkaConsumerConfig:
      enable.auto.commit: true
      group.id: test-group-mysql
```

The reason is that there is an order problem when setting in the code:

```java
properties = new Properties();
properties.putAll(config.getKafkaConsumerConfig());
properties.setProperty(ConsumerConfig.GROUP_ID_CONFIG, config.getGroupId());
properties.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, config.getBootstrapServers());
```

I don't know if this is by design or a mistake in itself.
Please confirm.

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -25,6 +25,7 @@
 * HTTPServer support the handler register with allowed HTTP methods.
 * [Critical] Revert [**Enhance DataCarrier#MultipleChannelsConsumer to add
   priority**](https://github.com/apache/skywalking/pull/8664) to avoid consuming issues.
+* Fix the problem that some configurations (such as group.id) did not take effect due to the override order when using the kafkaConsumerConfig property to extend the configuration in Kafka Fetcher.
 
 #### UI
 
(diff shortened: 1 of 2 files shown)


wu-sheng (approved):

dmsolr:
In fact, it totally is not a bug. To simplify configuration, we prefect users to configure as follows:
```yaml
kafka-fetcher:
  selector: ${SW_KAFKA_FETCHER:default}
  default:
    bootstrapServers: ${SW_KAFKA_FETCHER_SERVERS:192.168.101.1:9092}
    groupId: ${SW_KAFKA_FETCHER_GROUP_ID: skywalking}
```

Not to configure `kafkaConsumerConfig`, this is for advance configure.

wu-sheng:
@dmsolr I checked the codes, it seems we really provide this `kafkaConsumerConfig`. Once we did, I think the changed behavior is at least more sensible.

dmsolr:
make sense.
Yes, users can configure `BOOTSTRAP_SERVERS_CONFIG` and `GROUP_ID_CONFIG ` by KafkaFetchConfig(base) and KafkaConsumerConfig(advance).

wu-sheng:
Yes, once they could be override, make `kafkaConsumerConfig` to override single configuration key seems a better pratice.

PR #8934, opened 2022-04-22
Fix wrong build time in OAP version

### LINKED ISSUE #8887: [Bug] OAP version generator is not working as expected
### Search before asking

### What happened

image

From version to version, the commit ID in the version is not changing anymore.

Address the reason and fix.

### How to reproduce

Different commits are using a very old commit time as version number suffix

### Anything else

_No response_

### Code of Conduct

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #8887.


dmsolr (approved):

kezhenxu94:
the build time is not working well no matter how I tried as per [their doc](https://github.com/git-commit-id/git-commit-id-maven-plugin/blob/master/docs/using-the-plugin.md) thus I'd just remove the build time to avoid confusion

wu-sheng (approved):

PR #8945, opened 2022-04-25
Attempt to use first NonLoopbackAddress when registering nacos instance.

wyt:

When the oap cluster uses nacos, if `internalComHost` and `internalComPort` are not set, an instance with an IP of 0.0.0.0 is always registered in nacos. Can it be obtained automatically in the program to reduce the difficulty of use?

### DIFF
--- a/oap-server/server-cluster-plugin/cluster-nacos-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/nacos/NacosCoordinator.java
+++ b/oap-server/server-cluster-plugin/cluster-nacos-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/nacos/NacosCoordinator.java
@@ -20,7 +20,13 @@ package org.apache.skywalking.oap.server.cluster.plugin.nacos;
 
 import com.alibaba.nacos.api.naming.NamingService;
 import com.alibaba.nacos.api.naming.pojo.Instance;
+
+import java.net.Inet4Address;
+import java.net.InetAddress;
+import java.net.NetworkInterface;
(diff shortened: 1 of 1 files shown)


wu-sheng:
Ww discussed about this before, 2 things

1. Use this setup means you have known what to do, and what does this setup mean
2. Use the first IP address doesn't change the fact it still could be wrong. A little approvement, but not much

And from the code perspective, if this is going to be accepted, we have to change all cluster manager implementations, besides k8s as it is automatically working.

wyt:
> Ww discussed about this before, 2 things
> 
> 1. Use this setup means you have known what to do, and what does this setup mean
> 2. Use the first IP address doesn't change the fact it still could be wrong. A little approvement, but not much
> 
> And from the code perspective, if this is going to be accepted, we have to change all cluster manager implementations, besides k8s as it is automatically working.

The automatic setting solves the problem in most cases, in a few cases it can be set manually, you are correct, need to change all cluster manager implementations, I didn't realize this at first. It is recommended that the official make a unified improvement on this issue.

wu-sheng:
You should read this 

https://github.com/apache/skywalking/blob/5c247b8d7fa5fa7279f95a5f96d028c0fe0fa3c1/oap-server/server-cluster-plugin/cluster-nacos-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/nacos/NacosCoordinator.java#L71

Actually, SkyWalking is already responding **unhealthy** status.

Check the official doc, https://skywalking.apache.org/docs/main/v9.0.0/en/setup/backend/backend-health-check/#health-check

PR #8959, opened 2022-04-27
Improve the performance of TermsAggregation

chenhaipeng:
### Improve the performance of TermsAggregation

When skywalking queries the topology map, it uses queryTopoInfo to query the corresponding information. When it finds that it takes a long time, the main time-consuming statements are as follows,add ` "execution_hint": "map",
         "collect_mode": "breadth_first",`there is about a 5-10x performance difference.

**My index table sw89_metrics-percent size is about 300GB/day**
```
POST /sw89_metrics-percent/_search
{
  "size": 0,
  "query": {
    "bool": {
      "must": [
        {
          "range": {
            "time_bucket": {
              "gte": 202204212336,
              "lte": 202204212351
            }
          }
        },
        {
          "terms": {
            "entity_id": [
              "xxx1"
            ]
          }
        },
        {
          "term": {
            "metric_table": "service_sla"
          }
        }
      ]
    }
  },
  "aggregations": {
    "entity_id": {
      "terms": {
        "field": "entity_id",
        "size": 1
      },
      "aggregations": {
        "percentage": {
          "avg": {
            "field": "percentage"
          }
        }
      }
    }
  }
}
```


wu-sheng:
Do you check why `breadth_first` is better than `depth_first`?

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/search/aggregation/TermsAggregation.java
```diff
@@ -58,6 +58,8 @@ public void serialize(final TermsAggregation value, final JsonGenerator gen,
                     if (value.getOrder() != null) {
                         writeOrder(value, gen);
                     }
+                    gen.writeStringField("execution_hint", "map");
+                    gen.writeStringField("collect_mode", "breadth_first");
```
wu-sheng, line 62: You put these changes in the `TermsAggregation`, I think this would affect not just topology, it could affect `TopN` query, such as slow instance latency list.

I could find the doc of ElasticSearch, https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html#search-aggregations-bucket-terms-aggregation-collect

So, are you meaning to do level-one aggregation first always provides better performance?

chenhaipeng: please see here https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-aggregations-bucket-terms-aggregation.html#CO114-1


Because the metrics data is relatively small, using breadth_first, and execution_hint map will have a huge query improvement。

I did a simple test
When the index data is empty, "execution_hint" is not added, "map" took 30209ms, and the query timeout

When adding "execution_hint ""map" took 556ms

There are also similar question by other https://stackoverflow.com/questions/50893653/elasticsearch-slow-terms-aggregation/50895792#50895792

chenhaipeng: > 

this elasticsearch version 6.8.0

wu-sheng:
Please update the change log, this improvement looks good to me.

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/search/aggregation/TermsAggregation.java
wu-sheng: Make sense. 2 things

1. Please update `changes.md` in the doc, we have ref in the PR template(but you deleted it)
2. Moving this change into ElasticSearch DAO. `TermsAggregation` is our version of the ElasticSearch client, it should not include OAP's internal logic.

Others are good to me.

kezhenxu94, line 62: Please don't do this in the codec level.

- Add fields `executionHint`, `collectMode` in `TermsAggregation` (you can make them enum type considering their possible values are limited in ES's spec).
- Add fields `executionHint`, `collectMode` in `TermsAggregationBuilder` and corresponding builder methods, pass the two to `TermsAggregation` constructor in `build` method.
- Write the 2 fields here only when they are not empty/blank.

Then you should change set the 2 fields in topology query.

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/search/aggregation/TermsAggregationBuilder.java
```diff
@@ -84,4 +99,34 @@ private ImmutableMap.Builder<String, Aggregation> subAggregations() {
...
+        }
+
+        @Override
+        public String toString() {
+            return value;
+        }
+    }
+
+    public enum ExecutionHint {
+        GLOBAL_ORDINALS("global_ordinals"), map("map");
```
wu-sheng, line 119:
```suggestion
        GLOBAL_ORDINALS("global_ordinals"), MAP("map");
```

I think code style will fail due to lower case.

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/search/CollectMode.java
```diff
@@ -0,0 +1,44 @@
...
+
+package org.apache.skywalking.library.elasticsearch.requests.search;
+
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@Getter
+@RequiredArgsConstructor(access = AccessLevel.PACKAGE)
+public final class CollectMode {
```
wu-sheng, line 26: What is this `CollectMode`?

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/search/aggregation/TermsAggregation.java
```diff
@@ -58,6 +60,12 @@ public void serialize(final TermsAggregation value, final JsonGenerator gen,
                     if (value.getOrder() != null) {
                         writeOrder(value, gen);
                     }
+                    if (value.getCollectMode() != null){
+                        gen.writeStringField("collect_mode", value.getCollectMode().value);
+                    }
+                    if (value.getExecutionHint() != null){
```
kezhenxu94, line 66:
```suggestion
                    if (value.getExecutionHint() != null) {
```

wu-sheng:
@chenhaipeng It seems you still have code style issues.

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/search/aggregation/TermsAggregationBuilder.java
```diff
@@ -31,7 +31,10 @@ public final class TermsAggregationBuilder implements AggregationBuilder {
     private Integer size;
     private ImmutableMap.Builder<String, Aggregation> subAggregations;
 
-    TermsAggregationBuilder(final String name) {
+    private CollectMode collectMode;
+    private ExecutionHint executionHint;
+
+    public TermsAggregationBuilder(final String name) {
```
kezhenxu94, line 37:
```suggestion
    TermsAggregationBuilder(final String name) {
```

kezhenxu94 (dismissed):
Basically good to me

wu-sheng (approved):

docs/en/changes/changes.md
```diff
@@ -39,6 +39,7 @@
 * Add the support eBPF profiling field into the process entity.
 * E2E: fix log test miss verify LAL and metrics.
 * Enhance Converter mechanism in kernel level to make BanyanDB native feature more effective.
+* Add TermsAggregation properties collect_mode and execution_hint.
```
wu-sheng, line 42:
```suggestion
* Add TermsAggregation properties collect_mode and execution_hint.
* add "execution_hint": "map", "collect_mode": "breadth_first" for aggregation and topology query to improve 5-10x performance.
```

wu-sheng (approved):

PR #8960, opened 2022-04-27
Support receive Kubernetes process from the Rover

mrproliu:
### Fix create eBPF profiling task failure
If the start time of the Profiling task is later than the start time of the task to be created, the two-time intervals do not overlap. At present, an error will be prompted, in fact, it should be able to be created normally.

### Support receive Kubernetes process from the Rover

For now, I use YAML config to deploy the rover into the Kubernetes. I think after the Rover has a released version, it could be added to the helm.


docs/en/changes/changes.md
```diff
@@ -28,6 +28,8 @@
 * Fix the problem that some configurations (such as group.id) did not take effect due to the override order when using the kafkaConsumerConfig property to extend the configuration in Kafka Fetcher.
 * Remove build time from the OAP version.
 * Add data-generator module to run OAP in testing mode, generating mock data for testing.
+* Support receive Kubernetes processes from gRPC protocol.
+* Fix create eBPF Profiling task verify failure.
```
wu-sheng, line 32: I think this feature doesn't exist in 9.0, then we don't need this in the change log.

mrproliu: Deleted.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/ebpf/EBPFProfilingMutationService.java
```diff
@@ -150,9 +151,13 @@ private String checkCreateRequest(EBPFProfilingTaskFixedTimeCreationRequest requ
...
         // query exist processes
         final List<EBPFProfilingTask> tasks = getProcessProfilingTaskDAO().queryTasks(
-                Arrays.asList(request.getServiceId()), request.getTargetType(), calculateStartTime(request), 0);
+                Arrays.asList(request.getServiceId()), request.getTargetType(), request.getStartTime(), 0);
         if (CollectionUtils.isNotEmpty(tasks)) {
-            return "already have profiling task at this time";
+            final EBPFProfilingTask mostRecentTask = tasks.stream()
+                    .min(Comparator.comparingLong(EBPFProfilingTask::getTaskStartTime)).get();
+            if (mostRecentTask.getTaskStartTime() < calculateStartTime(request)) {
+                return "already have profiling most recent task at this time";
```
wu-sheng, line 159: Is this checking existing tasks? Or something? `Task's time range overlaps with other tasks` seems a better response to show. And you should take care of grammar, upper case, and complete sentences.

mrproliu: Updated.

wu-sheng (approved):

kezhenxu94 (approved):

PR #8967, opened 2022-04-28
Clean scroll context

kezhenxu94:


docs/en/changes/changes.md
```diff
@@ -30,6 +30,7 @@
 * Add data-generator module to run OAP in testing mode, generating mock data for testing.
 * Support receive Kubernetes processes from gRPC protocol.
 * Fix the problem that es index(TimeSeriesTable, eg. endpoint_traffic, alarm_record) didn't create even after rerun with init-mode. This problem caused the OAP server to fail to start when the OAP server was down for more than a day.
+* Clean up scroll contexts after used.
```
wu-sheng, line 33: Would this cause OOM or something?

kezhenxu94: Usually this is not a critical problem as the context has an expiration time (30s), but in high volume of clusters (millions of service /network aliases that the all results cannot be retrieved within the expiration time) this would cause

```
Trying to create too many scroll contexts. Must be less than or equal to: [500]. This limit can be set by changing the [search.max_open_scroll_context] setting
```

kezhenxu94:
After I verified I'll mark it "Ready for review", before that let's hold this

wu-sheng (approved):

PR #8977, opened 2022-04-30
Add sharding keys to stream records

lujiajing1126:
Signed-off-by: Megrez Lu <lujiajing1126@gmail.com>


wu-sheng (approved):

wu-sheng:
@lujiajing1126 As ShardingKey is hard required for BanyanDB, please add verification in the storage implementation, to check whether shardingKey is missing.

PR #8982, opened 2022-05-04
Enhance CI skipping strategy

Superskyyy:
The current CI skipping strategy relies on a suboptimal mechanism that is hard to tune. 

This PR enhances the process to exclude files and we can add more whenever we need. Please take a look and make sure I did nothing wrong.

I also turned on the dead link checker schedule for the upstream repo only.

BTW, some of the e2e tests are flaky - like fluent-bit, shall we increase the retry times?


.github/workflows/skywalking.yaml
```diff
@@ -568,22 +607,26 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: Merge Requirement
+        # check changes, sanity, dependency license, unit, integration, e2e, e2e-istio and e2e-java-versions, 
+        # if all of them are working as expected then naturally exits else return error code
```
kezhenxu94, line 611: Exit with different codes in different failures looks a good practice but I don’t think it useful here as the failed tasks will be already red in the workflow so we don’t bother clicking into the detail to see the exit code. And it looks the logic becomes kinda verbose?

.github/workflows/skywalking.yaml
```diff
@@ -568,22 +607,26 @@ jobs:
...
-          if [[ ${{ needs.e2e-test.result }} != 'success' ]]; then
-            exit -1
-          fi
-          if [[ ${{ needs.e2e-test-istio.result }} != 'success' ]]; then
-            exit -1
-          fi
-          if [[ ${{ needs.e2e-test-java-versions.result }} != 'success' ]]; then
-            exit -1
-          fi
+          execute=${{ needs.changes.outputs.oap }}
```
kezhenxu94, line 613: I’d just write if not executed exit 0, and check the remaining one by one so you don’t need to check execute there too

Superskyyy: I remember when I first experimented weeks ago, it has to be like this because when there's something wrong before execute, the `changes` job will not run normally.

That said this variable won't be populated but there could be failed jobs before(including changes itself). Then I thought we could solve this by not `need` the sanity check in `changes`, but that in fact also doesn't work because changes -> skip[s] -> required check runs faster than the parallel sanity check, the button will be green before sanity check finishes. so we need to add another condition to check sanity check results to the `if always()`, 

This probably should work, but `changes` itself could potentially fail or suffer from other problems, so I chose to check each `skip` status to ensure the pipeline is doing correctly. Otherwise, it's may become flaky.

.github/workflows/skywalking.yaml
Superskyyy: > Exit with different codes in different failures looks a good practice but I don’t think it useful here as the failed tasks will be already red in the workflow so we don’t bother clicking into the detail to see the exit code. And it looks the logic becomes kinda verbose?

Lemme just change them all into -1, the logic side is explained below.

kezhenxu94:
Seems there is syntax error. Fix them and it looks good to me. If changing exit code back to all -1 doesn’t improve the codes simplicity just leave them as is!

kezhenxu94 (approved):

PR #8986, opened 2022-05-04
Enhance Converter mechanism in kernel level 

wu-sheng:

Enhance Converter mechanism in kernel level to make BanyanDB native feature more effective. Replace https://github.com/apache/skywalking/pull/8983 in a better way.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/trace/ProfileThreadSnapshotRecord.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/trace/ProfileThreadSnapshotRecord.java
@@ -18,7 +18,6 @@
 
 package org.apache.skywalking.oap.server.core.profiling.trace;
 
-import java.util.Base64;
 import lombok.Getter;
 import lombok.Setter;
 import org.apache.skywalking.oap.server.core.Const;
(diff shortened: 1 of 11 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/trace/ProfileThreadSnapshotRecord.java
```diff
@@ -81,11 +80,7 @@ public ProfileThreadSnapshotRecord storage2Entity(final Convert2Entity converter
             snapshot.setDumpTime(((Number) converter.get(DUMP_TIME)).longValue());
             snapshot.setSequence(((Number) converter.get(SEQUENCE)).intValue());
             snapshot.setTimeBucket(((Number) converter.get(TIME_BUCKET)).intValue());
-            if (StringUtil.isEmpty((String) converter.get(STACK_BINARY))) {
-                snapshot.setStackBinary(new byte[] {});
-            } else {
-                snapshot.setStackBinary(Base64.getDecoder().decode((String) converter.get(STACK_BINARY)));
-            }
+            snapshot.setStackBinary(converter.getBytes(STACK_BINARY));
```
lujiajing1126, line 83: For `ProfileThreadSnapshotRecord`,  I suppose the `entity2Storage` also needs some patch,

```diff
-converter.accept(STACK_BINARY, new String(Base64.getEncoder().encode(storageData.getStackBinary())));
+converter.accept(STACK_BINARY, storageData.getStackBinary());
```

lujiajing1126 (changes requested):

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/trace/ProfileThreadSnapshotRecord.java
wu-sheng: Fixed.

lujiajing1126 (approved):
LGTM

PR #8987, opened 2022-05-04
Revert "Enhance CI skipping strategy"

wu-sheng:
Reverts apache/skywalking#8982

Trying to trigger whether CI could be scheduled, https://github.com/apache/skywalking/pull/8986 is not running after I commits a new one


wu-sheng:
I can see this task is running, but https://github.com/apache/skywalking/pull/8986 is not running at all.

PR #9003, opened 2022-05-07
Enhance Deprecated MetricQuery(v1) getValues querying

chenhaipeng:
### Improve the performance of MetricQuery
When querying getValues, metrics.getIds() costs  metrics.size * per request, so I modified it to asynchronous concurrency，It can improve the query efficiency of the topology map。


wu-sheng:
Please update the change log.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetricQuery.java
```diff
@@ -40,6 +44,7 @@
  * @since 8.0.0 This query is replaced by {@link MetricsQuery}
  */
 @Deprecated
+@Slf4j
 public class MetricQuery implements GraphQLQueryResolver {
```
wu-sheng, line 48: Please notice, this class has been `Deprecated`, which means it is a shell only. The real codes are in `org.apache.skywalking.oap.query.graphql.resolver.MetricsQuery`

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetricQuery.java
```diff
@@ -59,17 +64,28 @@ public IntValues getValues(final BatchMetricConditions metrics, final Duration d
...
+            metrics.getIds().stream().forEach(id -> {
 
                 MetricsCondition condition = new MetricsCondition();
                 condition.setName(metrics.getName());
                 condition.setEntity(new MockEntity(id));
-
-                kv.setValue(query.readMetricsValue(condition, duration));
-                values.addKVInt(kv);
-            }
+                CompletableFuture future = CompletableFuture.supplyAsync(() -> {
```
wu-sheng, line 73: The question is, what exactly the case you are facing to make you feel you need query to be async? One way or another, UI has to wait for a database response.

chenhaipeng: This means that the id is queried concurrently, and you don't have to wait for each result before proceeding the next time.

chenhaipeng: I think the best way is to modify MetricsQueryEsDAO to support batch query of ids

wu-sheng: Yes, but v1 metric APIs have been `Deprecated`, we recommend using v2, and making this kind of bulk questions as different GraphQL level queries.

chenhaipeng: When there are many applications, the topology map display is too slow. The reason is that in the previous position, each id triggers a query, and the next query needs to wait for the query result of the previous id.

wu-sheng: Which version's UI are you talking about?

chenhaipeng: v8.9.1

docs/en/changes/changes.md
```diff
@@ -43,6 +43,7 @@
 * Add "execution_hint": "map", "collect_mode": "breadth_first" for aggregation and topology query to improve 5-10x performance.
 * Clean up scroll contexts after used.
 * Support autocomplete tags in logs query.
+* Enhance MetricQuery querying getValues to asynchronous concurrency query
```
wu-sheng, line 46:
```suggestion
* Enhance Deprecated MetricQuery(v1) getValues querying to asynchronous concurrency query
```

wu-sheng:
Please fix warnings.

wu-sheng (dismissed):

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetricQuery.java
```diff
@@ -59,17 +64,28 @@ public IntValues getValues(final BatchMetricConditions metrics, final Duration d
...
+                        return query.readMetricsValue(condition, duration);
+                    } catch (IOException e) {
+                        log.error("query.readMetricsValue error", e);
+                    }
+                    return 0L;
+                }).thenAccept(value -> {
+                    KVInt kv = new KVInt();
+                    kv.setId(id);
+                    kv.setValue(value);
+                    values.addKVInt(kv);
```
kezhenxu94, line 84: It's `ArrayList.add` behind `values.addKVInt(kv);` and `ArrayList` is not thread-safe, you need synchronization here.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetricQuery.java
```diff
@@ -59,17 +64,28 @@ public IntValues getValues(final BatchMetricConditions metrics, final Duration d
             kv.setValue(query.readMetricsValue(condition, duration));
             values.addKVInt(kv);
         } else {
-            for (final String id : metrics.getIds()) {
-                KVInt kv = new KVInt();
-                kv.setId(id);
+            List<CompletableFuture> futureList = new ArrayList<>();
```
kezhenxu94, line 67: You don't need this variable `futureList`, just use `metrics.getIds().stream().map(...).forEach(...)`

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetricQuery.java
```diff
@@ -59,17 +64,28 @@ public IntValues getValues(final BatchMetricConditions metrics, final Duration d
...
                 condition.setEntity(new MockEntity(id));
-
-                kv.setValue(query.readMetricsValue(condition, duration));
-                values.addKVInt(kv);
-            }
+                CompletableFuture future = CompletableFuture.supplyAsync(() -> {
+                    try {
+                        return query.readMetricsValue(condition, duration);
+                    } catch (IOException e) {
+                        log.error("query.readMetricsValue error", e);
```
kezhenxu94, line 77: This somewhat changes the logic, previously it throws and the UI will display the error message, while now you just return all `0` to frontend and there is no clue why all `0`s unless users check OAP logs.

kezhenxu94 (changes requested):
A thread safe issue need to be addressed and some nits.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetricQuery.java
wu-sheng: I think adding `Collections.synchronizedList` wrapper should be enough.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetricQuery.java
chenhaipeng: I think the best way is to modify MetricsQueryEsDAO to support batch query of ids, ignore this pr

chenhaipeng: > 

I change another way to implement it

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetricQuery.java
chenhaipeng: I change another way to implement it, avoid use  lock

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetricQuery.java
```diff
@@ -59,17 +65,26 @@ public IntValues getValues(final BatchMetricConditions metrics, final Duration d
...
-                KVInt kv = new KVInt();
-                kv.setId(id);
+            List<CompletableFuture<Pair<String, Long>>> futureList = metrics.getIds().stream().map(id -> {
 
                 MetricsCondition condition = new MetricsCondition();
                 condition.setName(metrics.getName());
                 condition.setEntity(new MockEntity(id));
-
-                kv.setValue(query.readMetricsValue(condition, duration));
+                CompletableFuture<Pair<String, Long>> future = CompletableFuture.supplyAsync(() -> {
```
wu-sheng, line 73: KVInt is the pair, I think?

chenhaipeng: ok, I remove pair

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetricQuery.java
```diff
@@ -59,17 +63,24 @@ public IntValues getValues(final BatchMetricConditions metrics, final Duration d
...
-                values.addKVInt(kv);
-            }
+                CompletableFuture<KVInt> future = CompletableFuture.supplyAsync(() -> {
+                    try {
+                        KVInt kv = new KVInt();
+                        kv.setId(id);
+                        kv.setValue(query.readMetricsValue(condition, duration));
+                        return kv;
+                    } catch (IOException e) {
+                        throw new RuntimeException(e);
```
wu-sheng, line 78: Once you do this, I think you need to check `CompletableFuture#isCompletedExceptionally` and process in this case?

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetricQuery.java
```diff
@@ -59,17 +63,24 @@ public IntValues getValues(final BatchMetricConditions metrics, final Duration d
...
+                    }
+                });
+                return future;
+            }).collect(Collectors.toList());
+            futureList.stream().map(CompletableFuture::join).forEach(v -> values.addKVInt(v));
```
kezhenxu94, line 83: Actually `CompletableFuture.supplyAsync` can be replaced with `parallelStream` with just a very trivial difference that it doesn't matter in the case here, but is simpler, what do you think?

```java
            metrics.getIds()
                   .parallelStream()
                   .map(id -> {
                        MetricsCondition condition = new MetricsCondition();
                        condition.setName(metrics.getName());
                        condition.setEntity(new MockEntity(id));
                        try {
                            KVInt kv = new KVInt();
                            kv.setId(id);
                            kv.setValue(query.readMetricsValue(condition, duration));
                            return kv;
                        } catch (IOException e) {
                            throw new RuntimeException(e);
                        }
                   })
                   .forEach(v -> values.addKVInt(v));

```

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetricQuery.java
kezhenxu94: The exception will propagate to caller in `.join`

kezhenxu94 (approved):
Generally good to me

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetricQuery.java
wu-sheng: OK, then.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetricQuery.java
wu-sheng: @chenhaipeng What do you think about this?

chenhaipeng: yes! I think so too

wu-sheng: OK, I would wait for your update.

Thanks for bringing this to the community's attention. We missed this when building `v1->v2` bridge.

wu-sheng:
Check the code style please.

kezhenxu94 (dismissed):

wu-sheng (approved):

PR #9008, opened 2022-05-08
Fix the pod match error when the service has multiple selector in kubernates environment.

mrproliu:
When OAP received the metrics from OTEL and bound it to the K8S_SERVICE entity, I found that when the Service has multiple selectors, the matching rules between service and pod are incorrect.

The current match rule is that as long as any selector in the service matches the label in the pod, it is considered a match and should be changed to match all.


wankai123 (approved):
Thanks!

PR #9020, opened 2022-05-10
Add Envoy internal cost metrics

kezhenxu94:

### DIFF
--- a/oap-server/server-starter/src/main/resources/ui-initialized-templates/mesh/mesh-service.json
+++ b/oap-server/server-starter/src/main/resources/ui-initialized-templates/mesh/mesh-service.json
@@ -464,6 +464,61 @@
                     "sortMetrics"
                   ],
                   "moved": false
+                },
+                {
+                  "x": 0,
+                  "y": 54,
+                  "w": 8,
+                  "h": 13,
+                  "i": "23",
+                  "type": "Widget",
+                  "metricTypes": [
+                    "readMetricsValues"
+                  ],
+                  "metrics": [
+                    "service_sidecar_internal_req_latency_nanos"
+                  ],
+                  "moved": false,
+                  "graph": {
+                    "type": "Line",
+                    "step": false,
+                    "smooth": false,
+                    "showSymbol": false,
+                    "showXAxis": true,
+                    "showYAxis": true
+                  },
+                  "widget": {
+                    "title": "Proxy Sidecar Internal Latency in Nanoseconds (Request)",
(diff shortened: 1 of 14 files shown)


oap-server/server-starter/src/main/resources/ui-initialized-templates/mesh/mesh-service.json
```diff
@@ -464,6 +464,65 @@
...
+                    "showXAxis": true,
+                    "showYAxis": true
+                  },
+                  "widget": {
+                    "title": "Proxy Sidecar Internal Latency (Request)",
+                    "tips": "For more details to understand this metrics, refer to https://www.envoyproxy.io/docs/envoy/v1.22.0/api-v3/data/accesslog/v3/accesslog.proto.html#data-accesslog-v3-accesslogcommon"
+                  },
+                  "metricConfig": [
+                    {
+                      "unit": ""
```
wankai123, line 496: Is it better to give a unit in the title? The other places mostly are `ms`.
The empty unit here could remove.

kezhenxu94: > Is it better to give a unit in the title? The other places mostly are `ms`.
> The empty unit here could remove.

I tried to give it a unit, but there is nothing different in UI, where does this unit take effect?

kezhenxu94: > The empty unit here could remove.

It's exported from the UI and I don't want to edit this file, otherwise every time in the future you have to edit it

wu-sheng: This unit seems not expected. @Fine0830 I think this may be dirty data when activating other charts?

@kezhenxu94 In this case, you should add title on the title, as it is `nanosecond`

kezhenxu94: unit added in title

Fine0830: Yep. To avoid similar problems, I will implement functions to remove unexpected data for exporting dashboards.

wu-sheng: @kezhenxu94 Please remove it manually for now. We may need to clean more dashboards once @Fine0830 finished that advanced feature.(@Fine0830 It is better to submit a ticket to track this)

wu-sheng (approved):

PR #9033, opened 2022-05-11
Adjust the UT of dynamic configmap config

dashanji:
Signed-off-by: dashanji <dashanjic@gmail.com>

The correct config key is `receiver-trace.default.uninstrumentedGateways` in the `skywalking-dynamic-configmap.example.yaml`


docs/en/setup/backend/dynamic-config.md
```diff
@@ -32,7 +32,7 @@ Supported configurations are as follows:
 | Config Key | Value Description | Value Format Example |
 |:----:|:----:|:----:|
 |agent-analyzer.default.slowDBAccessThreshold| Thresholds of slow Database statement. Overrides `receiver-trace/default/slowDBAccessThreshold` of `application.yml`. | default:200,mongodb:50|
-|agent-analyzer.default.uninstrumentedGateways| The uninstrumented gateways. Overrides `gateways.yml`. | Same as [`gateways.yml`](uninstrumented-gateways.md#configuration-format). |
+|receiver-trace.default.uninstrumentedGateways| The uninstrumented gateways. Overrides `gateways.yml`. | Same as [`gateways.yml`](uninstrumented-gateways.md#configuration-format). |
```
wu-sheng, line 35: I checked the latest codebase, it seems the configuration is correct.

https://github.com/apache/skywalking/blob/62a67bacb612a19f78b8101de7017a7675673c72/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/UninstrumentedGatewaysConfig.java#L53

https://github.com/apache/skywalking/blob/62a67bacb612a19f78b8101de7017a7675673c72/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/module/AnalyzerModule.java#L26

Anything I missed here?

wu-sheng (changes requested):

docs/en/setup/backend/dynamic-config.md
wu-sheng: `skywalking-dynamic-configmap.example.yaml` file includes the incorrect example. You should fix it there, but it is commented, I am not sure whether it has the real meaning.

dashanji: What I find is the config key is `receiver-trace.default.uninstrumentedGateways`  in https://github.com/apache/skywalking/blob/62a67bacb612a19f78b8101de7017a7675673c72/oap-server/server-configuration/configuration-k8s-configmap/src/test/resources/skywalking-dynamic-configmap.example.yaml#L104-L109

And there are no `uninstrumentedGateways` filed in `application.yml`
https://github.com/apache/skywalking/blob/62a67bacb612a19f78b8101de7017a7675673c72/oap-server/server-starter/src/main/resources/application.yml#L264-L275

But the config key is `agent-analyzer.default.uninstrumentedGateways` in the doc, so I don't know which key should be correct.

wu-sheng: Check the links I posted, that is how the codes control this config. Nothing relates to the application.yml, which is booting intializtion only.

dashanji: Make sense to me, thanks. BTW, should we change the configmap in there and other repo like `skywalking-kubenertes` to keep doc and code consistent?

wu-sheng: If this key exists in other official helm or operator, it should be kept consistent with the codes.

The UT of configmap config center should be changed to avoid other's confusion as you did.

dashanji: Okay, I will change it next.

wu-sheng:
Please fix the title, I think this is only a UT adjustment.

dashanji:
get it.

dashanji:
I find the `agent-analyzer.default.slowDBAccessThreshold` is the same and update.

wu-sheng (approved):
Once tests passed.

PR #9038, opened 2022-05-11
Update the eBPF Profiling text in UI template and link it to the rover

mrproliu:

### DIFF
--- a/oap-server/server-starter/src/main/resources/ui-initialized-templates/general/general-service.json
+++ b/oap-server/server-starter/src/main/resources/ui-initialized-templates/general/general-service.json
@@ -802,7 +802,7 @@
                   "graph": {
                     "fontColor": "white",
                     "backgroundColor": "blue",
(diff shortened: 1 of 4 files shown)


oap-server/server-starter/src/main/resources/ui-initialized-templates/general/general-service.json
```diff
@@ -802,9 +802,10 @@
                   "graph": {
                     "fontColor": "white",
                     "backgroundColor": "blue",
-                    "content": "eBPF Profiling support services written in C, C++, Golang, and Rust.",
+                    "content": "eBPF Profiling support services written in C, C++, Golang, and Rust. Make sure has been deployed SkyWalking Rover. ",
```
wu-sheng, line 805:
```suggestion
                    "content": "eBPF Profiling support services written in C, C++, Golang, and Rust. SkyWalking Rover provides this profiling capability. ",
```

PR #9076, opened 2022-05-17
Storage(ElasticSearch): add search options to tolerate inexisting indices

kezhenxu94:


kezhenxu94:
This should be a left-out in #7634, I added those APIs (`allowNoIndices`, `ignoreUnavailable`, `expandWildcards`) but didn't use them...

wankai123 (approved):

wu-sheng:
Do we consider to do physical index based query later?

kezhenxu94:
> Do we consider to do physical index based query later?

Yes I'll do that later

PR #9107, opened 2022-05-20
Support OFF CPU eBPF Profiling

mrproliu:
Support OFF CPU eBPF Profiling data collection and query protocol. 
Follow https://github.com/apache/skywalking-data-collect-protocol/pull/74, https://github.com/apache/skywalking-query-protocol/pull/89

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/EBPFProfilingData.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/EBPFProfilingData.java
@@ -20,6 +20,7 @@ package org.apache.skywalking.oap.server.core.source;
 
 import lombok.Data;
 import org.apache.skywalking.oap.server.core.Const;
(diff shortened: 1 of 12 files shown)


lujiajing1126 (approved):
For the BanyanDB part, LGTM

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/EBPFProfilingData.java
```diff
@@ -46,7 +47,7 @@ public String getEntityId() {
     private String taskId;
     private long uploadTime;
     private String stackIdList;
+    private EBPFProfilingTargetType targetType;
     private byte[] stacksBinary;
```
wu-sheng, line 51: I think you should change this field name too?

mrproliu: Do you mean `stacksBinary`? updated.

wu-sheng (approved):

PR #9116, opened 2022-05-21
Add shards factor for BanyanDB

lujiajing1126:

This PR also fixes `DataTable` serialization.

Any suggestion for shards number? @hanahmily


oap-server/server-starter/src/main/resources/application.yml
```diff
@@ -239,6 +239,9 @@ storage:
     group: ${SW_STORAGE_BANYANDB_GROUP:default}
     maxBulkSize: ${SW_STORAGE_BANYANDB_MAX_BULK_SIZE:5000}
     flushInterval: ${SW_STORAGE_BANYANDB_FLUSH_INTERVAL:15}
+    metricsShardsFactor: ${SW_STORAGE_BANYANDB_METRICS_SHARDS_FACTOR:2}
```
kezhenxu94, line 242: All these new configurations are not added to https://github.com/apache/skywalking/blob/master/docs/en/setup/backend/configuration-vocabulary.md

oap-server/server-starter/src/main/resources/application.yml
```diff
@@ -239,6 +239,9 @@ storage:
...
+    recordShardsFactor: ${SW_STORAGE_BANYANDB_RECORD_SHARDS_FACTOR:2}
+    superDatasetShardsFactor: ${SW_STORAGE_BANYANDB_SUPERDATASET_SHARDS_FACTOR:2}
```
wu-sheng, line 244: metricsShardsFactor and recordShardsFactor should be 1 ideally. Because we have one measure/stream per metric/record, one shard should be enough.

oap-server/server-starter/src/main/resources/application.yml
wu-sheng: This doc should be updated as well

https://github.com/apache/skywalking/blob/master/docs/en/setup/backend/backend-storage.md

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBStorageConfig.java
```diff
@@ -47,5 +47,17 @@ public class BanyanDBStorageConfig extends ModuleConfig {
...
+     */
+    private int metricsShardsFactor;
+    /**
+     * Shards Factor for a normal record.
+     */
+    private int recordShardsFactor;
+    /**
+     * Shards Factor for a super dataset
+     */
+    private int superDatasetShardsFactor;
```
wu-sheng, line 62: image

Please take another look at how `factor` works in ElasticSearch. The factor is the amplification coefficient. Others are using `ShardsNumber`.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/MetadataRegistry.java
```diff
@@ -258,7 +229,7 @@ private List<TagMetadata> parseExtraTagSpecs(String tags, Schema.SchemaBuilder b
     private TagFamilySpec.TagSpec parseTagSpec(ModelColumn modelColumn) {
         final Class<?> clazz = modelColumn.getType();
         final String colName = modelColumn.getColumnName().getStorageName();
-        if (String.class.equals(clazz) || DataTable.class.equals(clazz) || JsonObject.class.equals(clazz)) {
+        if (String.class.equals(clazz) || StorageDataComplexObject.class.isAssignableFrom(clazz) || JsonObject.class.equals(clazz)) {
             return TagFamilySpec.TagSpec.newStringTag(colName);
```
wu-sheng, line 233: If this is a `StorageDataComplexObject` and `JsonObject`, it should not be indexed as a tag. A field is enough.
The String should follow `storageOnly` annotation attribute.

wu-sheng: We never search on this type of fields as it is meaningless.

lujiajing1126: No. It does not work like this. An index rule determines whether a tag is indexed.

wu-sheng: Then, how should we determine a field or a no-index tag?

lujiajing1126: > Then, how should we determine a field or a no-index tag?

Currently, every entity only have one single field. It is determined by `ValueColumnMetadata.getValueCName`

wu-sheng: So, the field is only for `groupby`'s aggregation function? @hanahmily What is the recommendation from the server-side?
I am feeling we need a doc definition about this, in the cases for further PR review like this, especially for new contributors.

lujiajing1126: > So, the field is only for `groupby`'s aggregation function? @hanahmily What is the recommendation from the server-side? I am feeling we need a doc definition about this, in the cases for further PR review like this, especially for new contributors.

No. It can also be read, for example in `multiGet`

wu-sheng: Yes, it can be read, but as well as tag. The development doc for SkyWalking is too early, we even don't have a conclusion yet. Jaeger is easy, it just needs a span query.
It is better we put recommendation from server's perspective, then we determine which is better for now. We are getting to know more about the progress of this integration. We are very early.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBStorageConfig.java
wu-sheng: @lujiajing1126 Your new commit seems missing this.

lujiajing1126: Fixed

oap-server/server-starter/src/main/resources/application.yml
lujiajing1126: Fixed

wu-sheng (approved):

PR #9117, opened 2022-05-22
Add TiDB, OpenSearch, Postgres storage optional to Trace and eBPF Profiling E2E Testing

mrproliu:
1. Add storage testing into trace and eBPF Profiling E2E Testing.
2. Remove duplicate checking in storage E2E testing. 


wu-sheng:
I am feeling a little curious about what are storage-e2e testing more now?

mrproliu:
The storage E2E tests entity, entity relation, trace, and event between all storage. It seems that only storage E2E tests are verifying these across storage, except for events.

wu-sheng (approved):

PR #9128, opened 2022-05-25
Fix scroll query in Elastic Search storage

mrproliu:
### Fix Scroll query in Elastic Search storage
If the `SW_STORAGE_ES_SCROLLING_BATCH_SIZE` environment config is less them the Service entity count, it would throw `No search context found for id` exception. The root cause is we're going to close the scroll after we've requested it once. 

### Add `profileDataQueryBatchSize` config in Elastic Search storage
EBPF Profiling Data stores a large amount of symbol data. When a large amount of data content is queried at one time, a `ContentTooLargeException` from netty will occur, so I add a new config for the query. 


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/cache/NetworkAddressAliasEsDAO.java
```diff
@@ -62,9 +62,9 @@ public List<NetworkAddressAlias> loadLastUpdate(long timeBucketInMinute) {
 
             SearchResponse results =
                 getClient().search(NetworkAddressAlias.INDEX_NAME, search, params);
-            while (true) {
-                final String scrollId = results.getScrollId();
-                try {
+            final String scrollId = results.getScrollId();
+            try {
+                while (true) {
```
kezhenxu94, line 67: This is wrong, `scrollId` is never updated and you'll be always scrolling to the same context (at line 82)

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java
```diff
@@ -117,6 +117,12 @@ public class StorageModuleElasticsearchConfig extends ModuleConfig {
     private int scrollingBatchSize = 5000;
     private int segmentQueryMaxSize = 200;
     private int profileTaskQueryMaxSize = 200;
+    /**
+     * The batch size that used to scroll on the large eBPF profiling data result.
+     * The profiling data contains full symbol data, that could make ElasticSearch response large content,
+     * so could not same with {@link #scrollingBatchSize}.
+     */
```
wu-sheng, line 124:
```suggestion
    /**
     * The batch size that is used to scroll on the large eBPF profiling data result.
     * The profiling data contains full-stack symbol data, which could make ElasticSearch response large content.
     *  {@link #scrollingBatchSize} would not be used in profiling data query.
     */
```

kezhenxu94 (changes requested):
> If the SW_STORAGE_ES_SCROLLING_BATCH_SIZE environment config is less them the Service entity count, it would throw No search context found for id exception.

Can you check where the exception is thrown, is it `getClient().scroll(SCROLL_CONTEXT_RETENTION, scrollId);` or `getClient().deleteScrollContextQuietly(scrollId);`?

> The root cause is we're going to close the scroll after we've requested it once.

This is by design, read the PR #8967 (https://github.com/apache/skywalking/pull/8967#discussion_r860463732) that introduced this.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/cache/NetworkAddressAliasEsDAO.java
mrproliu: image

> This is wrong, `scrollId` is never updated.

Yes, the `scrollId` is never updated. So we can't delete the `scrollId` after executing the scroll once, so it cannot scroll it again based on the same `scrollId` for the second time.

kezhenxu94 (approved):

PR #9130, opened 2022-05-25
Add APIs to query ondemand Pod log

kezhenxu94:


oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/GraphQLQueryConfig.java
```diff
@@ -31,4 +31,5 @@ public class GraphQLQueryConfig extends ModuleConfig {
     private boolean enableLogTestTool;
     private int maxQueryComplexity = 1000;
     private boolean enableUpdateUITemplate = false;
+    private boolean enableOndemandLog = false;
```
wu-sheng, line 34:
```suggestion
    private boolean enableOnDemandLog = false;
```
The most I saw about this is `on-demand` or `onDemand` as two words.

kezhenxu94:
> image
> 
> We should add `on-demand logs` here.
> 
> And one question, on which dashboards we are going to add this feature? `Service Mesh` and `Kubernetes`?

I discussed with @Fine0830 and want to put it here 
Screen Shot 2022-05-25 at 21 30 38

wu-sheng:
This is good, recommend considering putting into k8s dashboard as well.

wu-sheng (approved):

wu-sheng:
Do we have any chance to have e2e for this? Considering the env limitations, is this practical?

PR #9141, opened 2022-05-27
Upgrade netty to fix bug in some Linux versions

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Related https://github.com/netty/netty/issues/12343.


wu-sheng:
Thanks for fixing it. It is interesting to see netty fails on specific OS.

wu-sheng (approved):

PR #9151, opened 2022-05-31
Add instance properties extractor in MAL

kezhenxu94:

### DIFF
--- a/docs/en/concepts-and-designs/mal.md
+++ b/docs/en/concepts-and-designs/mal.md
@@ -1,6 +1,6 @@
 # Meter Analysis Language
 
-The meter system provides a functional analysis language called MAL (Meter Analysis Language) that lets users analyze and 
+The meter system provides a functional analysis language called MAL (Meter Analysis Language) that lets users analyze and
 aggregate meter data in the OAP streaming system. The result of an expression can either be ingested by the agent analyzer,
 or the OC/Prometheus analyzer.
 
@@ -66,7 +66,7 @@ This feature requires authorizing the OAP Server to access K8s's `API Server`.
 
 ##### retagByK8sMeta
 `retagByK8sMeta(newLabelName, K8sRetagType, existingLabelName, namespaceLabelName)`. Add a new tag to the sample family based on the value of an existing label. Provide several internal converting types, including
-- K8sRetagType.Pod2Service  
+- K8sRetagType.Pod2Service
 
 Add a tag to the sample using `service` as the key, `$serviceName.$namespace` as the value, and according to the given value of the tag key, which represents the name of a pod.
 
@@ -104,13 +104,13 @@ Between a sample family and a scalar, the operator is applied to the value of ev
 
 ```
 instance_trace_count + 2
-``` 
+```
 
-or 
+or
 
 ```
 2 + instance_trace_count
-``` 
+```
 
 results in
 
@@ -120,15 +120,15 @@ instance_trace_count{region="us-east",az="az-3"} 22 // 20 + 2
 instance_trace_count{region="asia-north",az="az-1"} 35 // 33 + 2
 ```
 
-Between two sample families, a binary operator is applied to each sample in the sample family on the left and 
+Between two sample families, a binary operator is applied to each sample in the sample family on the left and
 its matching sample in the sample family on the right. A new sample family with empty name will be generated.
 Only the matched tags will be reserved. Samples with no matching samples in the sample family on the right will not be found in the result.
 
-Another sample family `instance_trace_analysis_error_count` is 
+Another sample family `instance_trace_analysis_error_count` is
 
 ```
 instance_trace_analysis_error_count{region="us-west",az="az-1"} 20
-instance_trace_analysis_error_count{region="asia-north",az="az-1"} 11 
+instance_trace_analysis_error_count{region="asia-north",az="az-1"} 11
 ```
 
 Example expression:
@@ -137,7 +137,7 @@ Example expression:
 instance_trace_analysis_error_count / instance_trace_count
 ```
 
-This returns a resulting sample family containing the error rate of trace analysis. Samples with region us-west and az az-3 
+This returns a resulting sample family containing the error rate of trace analysis. Samples with region us-west and az az-3
 have no match and will not show up in the result:
 
 ```
@@ -154,8 +154,8 @@ resulting in a new sample family having fewer samples (sometimes having just a s
  - min (select minimum over dimensions)
  - max (select maximum over dimensions)
  - avg (calculate the average over dimensions)
- 
-These operations can be used to aggregate overall label dimensions or preserve distinct dimensions by inputting `by` parameter. 
+
+These operations can be used to aggregate overall label dimensions or preserve distinct dimensions by inputting `by` parameter.
 
 ```
 <aggr-op>(by: <tag1, tag2, ...>)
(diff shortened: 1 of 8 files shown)


oap-server/server-starter/src/main/resources/otel-oc-rules/k8s-instance.yaml
```diff
@@ -0,0 +1,23 @@
...
+# limitations under the License.
+
+filter: "{ tags -> tags.job_name in [ 'kubernetes-cadvisor', 'kube-state-metrics' ] }" # The OpenTelemetry job name
+expSuffix: |-
+  service(['cluster' , 'service'], '::', Layer.K8S_SERVICE)
+  .instance(['cluster', 'service'], '::', ['pod'], '', Layer.K8S_SERVICE, { tags -> ['pod': tags.pod, 'namespace': tags.namespace] })
+metricPrefix: k8s_service_instance
+metricsRules:
+  - name: pod_instance_status
+    exp: kube_pod_status_phase.retagByK8sMeta('service' , K8sRetagType.Pod2Service , 'pod' , 'namespace').tagNotEqual('service' , '').sum(['cluster', 'namespace', 'service' , 'pod'])
```
kezhenxu94, line 23: This is a demo of how the extractor works (and will be used in on demand logs), my concern here is that we have to generate an unused instance metrics here, I'm trying to add a drop mechanism in MAL so that we can leverage the "side effect" but not store the metrics, would like to know what do you think @wu-sheng @wankai123

wankai123: A little curious, why don't we get the pod meta info from the K8s API server directly?

wu-sheng: @wankai123 What do you mean? This is for a streaming to add instance to represent a pod.

wankai123: Oh, I see. This is not for mesh, right?

wu-sheng: I think we don't expose enough API for this, no matter what we do to drop, it is a kind of hijacking.

I don't have much concern about this metric actually. This just adds an extra 1% metric :)

docs/en/concepts-and-designs/mal.md
```diff
@@ -239,13 +239,13 @@ last_server_state_sync_time_in_seconds.tagEqual('production', 'catalog').downsam
 They extract level relevant labels from metric labels, then informs the meter-system the level and [layer](../../../oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/Layer.java) to which this metric belongs.
 
  - `service([svc_label1, svc_label2...], Layer)` extracts service level labels from the array argument, extracts layer from `Layer` argument.
- - `instance([svc_label1, svc_label2...], [ins_label1, ins_label2...], Layer)` extracts service level labels from the first array argument, 
-                                                                        extracts instance level labels from the second array argument, extracts layer from `Layer` argument.
- - `endpoint([svc_label1, svc_label2...], [ep_label1, ep_label2...])` extracts service level labels from the first array argument, 
+ - `instance([svc_label1, svc_label2...], [ins_label1, ins_label2...], Layer, Closure<Map<String, String>> propertiesExtractor)` extracts service level labels from the first array argument,
```
wu-sheng, line 242: Is this Closure optional?

kezhenxu94: Yes. It's optional.

oap-server/server-starter/src/main/resources/otel-oc-rules/k8s-instance.yaml
kezhenxu94: > I think we don't expose enough API for this, no matter what we do to drop, it is a kind of hijacking.
> 
> I don't have much concern about this metric actually. This just adds an extra 1% metric :)

Hi, I didn't mean to drop the metrics in the streaming process, I meant to drop the metrics just before MAL sends those metrics, see my changes in https://github.com/apache/skywalking/pull/9151/commits/1db9e71c5401a9a4b91536e7b3b397cd7a6769c8

oap-server/server-starter/src/main/resources/otel-oc-rules/k8s-instance.yaml
```diff
@@ -17,6 +17,7 @@ filter: "{ tags -> tags.job_name in [ 'kubernetes-cadvisor', 'kube-state-metrics
 expSuffix: |-
   service(['cluster' , 'service'], '::', Layer.K8S_SERVICE)
   .instance(['cluster', 'service'], '::', ['pod'], '', Layer.K8S_SERVICE, { tags -> ['pod': tags.pod, 'namespace': tags.namespace] })
+  .drop()
```
wu-sheng, line 20: `drop` is making this strange. 2 perspective
1. User would ask why we generate a metric and then drop
2. Meter system engine is still running for this metric. AKA, this `drop` only means not aggregation.

How about we keep this metric for now?

oap-server/server-starter/src/main/resources/otel-oc-rules/k8s-instance.yaml
wu-sheng: see https://github.com/apache/skywalking/pull/9151#discussion_r886266807 

We could create an instance(pod-oriented) dashboard to make this clear. But for now, even without them, I feel it is fine.

kezhenxu94:
Then this is ready to review

kezhenxu94:
Seems the instance properties are not updated incrementally but overridden instead. That might need more modifications..

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/instance/InstanceTraffic.java
```diff
@@ -79,7 +79,7 @@ public boolean combine(final Metrics metrics) {
         final InstanceTraffic instanceTraffic = (InstanceTraffic) metrics;
         this.lastPingTimestamp = instanceTraffic.getLastPingTimestamp();
         if (instanceTraffic.getProperties() != null && instanceTraffic.getProperties().size() > 0) {
-            this.properties = instanceTraffic.getProperties();
+            instanceTraffic.getProperties().entrySet().forEach(it -> this.properties.add(it.getKey(), it.getValue()));
```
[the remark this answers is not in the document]
wu-sheng: > this.properties = instanceTraffic.getProperties();

The previous version is overriding, how your new codes could make override happens?

wu-sheng:
BTW, could you add e2e verification to do properties check on the mesh case? (We don't have k8s monitoring e2e, I think)
Is current check engine working well for this list check?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/instance/InstanceTraffic.java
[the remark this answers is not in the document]
wu-sheng: In today's k8s monitoring, the instance is pod-oriented. In which case, you could create an instance properties update scenario? Could you share that?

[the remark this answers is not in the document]
wu-sheng: According to `Builder#storage2Entity`, this `properties` could be null, when we need to read an empty properties from the database. Please do null check.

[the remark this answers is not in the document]
kezhenxu94: > In today's k8s monitoring, the instance is pod-oriented. In which case, you could create an instance properties update scenario? Could you share that?

I was no saying my case. I was saying the CI test result, which shows the instance properties are overridden. I'll check more details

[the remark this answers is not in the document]
wu-sheng: Could it be NPE relative? Once the properties list is null, your codes are not able to make it back anymore.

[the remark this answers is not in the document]
kezhenxu94: > Could it be NPE relative? Once the properties list is null, your codes are not able to make it back anymore.

Ok that’s indeed because of NPE

wu-sheng (approved):

PR #9157, opened 2022-06-01
Allow querying logs only by instance to simplify frontend query

kezhenxu94:

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/instance/InstanceTraffic.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/instance/InstanceTraffic.java
@@ -175,6 +175,17 @@ public class InstanceTraffic extends Metrics {
     }
 
     public static class PropertyUtil {
+        /**
+         * `namespace` and `pod` are key properties that help "on demand Pod logs"
+         * to locate the corresponding Pod in Kubernetes, when language agent is
+         * registering a new service instance that is supposed to work in terms of
+         * "on demand Pod logs", the agent should also fill in these 2 properties.
+         *
+         * @since 9.1.0
+         */
+        public static final String NAMESPACE = "namespace";
+        public static final String POD = "pod";
+
         public static final String LANGUAGE = "language";
         public static final String IPV4 = "ipv4";
         public static final String IPV4S = "ipv4s";
--- a/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/OndemandLogQuery.java
+++ b/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/OndemandLogQuery.java
@@ -23,9 +23,9 @@ import static java.util.Objects.isNull;
 import static java.util.Objects.nonNull;
 import java.io.IOException;
 import java.lang.reflect.Type;
+import java.util.Collections;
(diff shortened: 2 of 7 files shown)


oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/OndemandLogQuery.java
```diff
@@ -34,64 +33,86 @@
...
+            metadataQuery.getInstance(condition.getServiceInstanceId());
+        final Map<String, String> attributesMap = ensureInstanceHasPodProperties(instance);
+        final String ns = attributesMap.get(PropertyUtil.NAMESPACE);
+        final String pod = attributesMap.get(PropertyUtil.POD);
+        return ondemandPodLogs(ns, pod, condition);
+    }
 
+    protected Map<String, String> ensureInstanceHasPodProperties(final ServiceInstance instance) {
+        if (instance == null) {
+            throw new IllegalArgumentException("No such instance");
```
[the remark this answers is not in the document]
kezhenxu94: > Should we return an empty container list? rather than an exception?

An exception is more reasonable in my opinion for non-existing instance, I think "no such instance" is totally different from "no logs in this instance"

wu-sheng:
I think generally, we should not report errors if this instance doesn't have on-demand logs. 

If you want this available on the UI, we could add one more field for the response.

kezhenxu94:
> If you want this available on the UI, we could add one more field for the response.

Why? UI already displays the errors reported by backend correctly,
Screen Shot 2022-06-01 at 15 59 26

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/OndemandLogQuery.java
```diff
@@ -34,64 +34,87 @@
...
+                .stream()
+                .collect(Collectors.toMap(Attribute::getName, Attribute::getValue));
+        if (!attributesMap.containsKey(PropertyUtil.NAMESPACE)
+            || !attributesMap.containsKey(PropertyUtil.POD)) {
+            return Collections.emptyMap();
+        }
+        return attributesMap;
+    }
+
+    public List<String> listContainers(
```
wu-sheng, line 107: @Fine0830 Once UI faces `listContainers` returning an empty list, please show a tip on the logs page, **The on-demand logs are not available.**

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/instance/InstanceTraffic.java
```diff
@@ -175,6 +175,8 @@ public Metrics toDay() {
     }
 
     public static class PropertyUtil {
+        public static final String NAMESPACE = "namespace";
+        public static final String POD = "pod";
```
wu-sheng, line 179: Could you add comments for these two to show these are key properties for on-demand logs? And `@since`?

wu-sheng (dismissed):

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/instance/InstanceTraffic.java
kezhenxu94: Added

Issue #9105, opened 2022-05-20
[Bug][BanyanDB] `Having` semantic bug


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar issues.

### Apache SkyWalking Component

BanyanDB (apache/skywalking-banyandb)

### What happened

`Having` for `List<String>` column is `at least one` including one of the collections, this is not correct.

### What you expected to happen

`Having` should be `subset` semantic. 
**columnA** having {value1, value2}, should mean value1 and value2 are both included in this columnA's list values.

### How to reproduce

Do trace query

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wu-sheng:
FYI @lujiajing1126

PR #9168, opened 2022-06-02
Add negative tags search for storage E2E

lujiajing1126:

This PR intends to add negative test for https://github.com/apache/skywalking/issues/9105 and corresponding PR https://github.com/apache/skywalking-banyandb/pull/120

### DIFF
--- a/test/e2e-v2/cases/storage/storage-cases.yaml
+++ b/test/e2e-v2/cases/storage/storage-cases.yaml
@@ -42,8 +42,11 @@ cases:
     expected: expected/metrics-has-value.yml
 
   # trace segment list
-  - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql trace ls  --tags http.method=POST
+  - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql trace ls --tags http.method=POST
     expected: expected/traces-list.yml
+  # negative tags search: relationship should be logical AND instead of logical OR
+  - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql trace ls --tags http.method=POST,http.status_code=200
+    expected: expected/empty-traces-list.yml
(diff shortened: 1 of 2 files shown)


test/e2e-v2/cases/storage/storage-cases.yaml
```diff
@@ -42,8 +42,11 @@ cases:
     expected: expected/metrics-has-value.yml
 
   # trace segment list
-  - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql trace ls  --tags http.method=POST
+  - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql trace ls --tags http.method=POST
     expected: expected/traces-list.yml
+  # negative tags search: relationship should be logical AND instead of logical OR
+  - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql trace ls --tags http.method=POST,http.status_code=200
```
wu-sheng, line 48: Is there no trace segment having both `http.method=POST` and `http.status_code=200`? Because of the test cases?

wu-sheng: `http.status_code >= 400`? This is strange. If there is a HTTP Post, it should be 200.

lujiajing1126: > `http.status_code >= 400`? This is strange. If there is a HTTP Post, it should be 200.

Yes, here it is `200`. I mean `status_code` will be recorded as a tag only if it is `>= 400`.

For example, https://github.com/apache/skywalking-java/blob/3997f0256056788bd054ee37e4603c11c0fd6756/apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-commons/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/commons/interceptor/AbstractMethodInterceptor.java#L204-L207

PR #9171, opened 2022-06-03
Add on demand logs support to service mesh

kezhenxu94:


oap-server/server-receiver-plugin/skywalking-mesh-receiver-plugin/src/main/java/org/apache/skywalking/aop/server/receiver/mesh/TelemetryDataDispatcher.java
```diff
@@ -187,6 +192,23 @@ private static void toServiceInstance(ServiceMeshMetric.Builder metrics, long mi
         SOURCE_RECEIVER.receive(serviceInstance);
     }
 
+    private static void toServiceInstanceTraffic(ServiceMeshMetric.Builder metrics, long minuteTimeBucket) {
```
wu-sheng, line 195: Please use `ServiceInstanceUpdate` as a source to update properties through `SOURCE_RECEIVER#receive`.

Your way is working as well, but usually, `MetricsStreamProcessor` is only called from a SourceDispatcher or Kernel API, such as from MeterSystem or Analyzer in Meter.

wu-sheng (approved):
LGTM, the dashboard updates will come next, right?

kezhenxu94:
> LGTM, the dashboard updates will come next, right?

Yes

PR #9199, opened 2022-06-10
ElasticSearch: scroll id should be updated when scrolling as it may change

kezhenxu94:
This partially reverts #9128 , originally I deleted the scroll id when I scroll one time, later that @mrproliu modified in #9128 to only delete the scroll id after the scroll is done, in #9128 we assumed that the scroll id is not changed but that's not the case, according to https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/search_operations.html 

```
    // When done, get the new scroll_id
    // You must always refresh your _scroll_id!  It can change sometimes
    $scroll_id = $response['_scroll_id'];
```

^^^ found by @wankai123 (thanks 🙇) and it becomes painful due to "It can change sometimes".

SO in this PR I just save the scroll ids and when done delete them all at once.

### DIFF
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/cache/NetworkAddressAliasEsDAO.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/cache/NetworkAddressAliasEsDAO.java
@@ -19,7 +19,9 @@
 package org.apache.skywalking.oap.server.storage.plugin.elasticsearch.cache;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.skywalking.library.elasticsearch.requests.search.Query;
 import org.apache.skywalking.library.elasticsearch.requests.search.Search;
@@ -62,9 +64,11 @@ public class NetworkAddressAliasEsDAO extends EsDAO implements INetworkAddressAl
 
             SearchResponse results =
                 getClient().search(NetworkAddressAlias.INDEX_NAME, search, params);
-            final String scrollId = results.getScrollId();
+            final Set<String> scrollIds = new HashSet<>();
             try {
                 while (true) {
+                    final String scrollId = results.getScrollId();
+                    scrollIds.add(scrollId);
                     if (results.getHits().getTotal() == 0) {
                         break;
                     }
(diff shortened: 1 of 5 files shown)


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/cache/NetworkAddressAliasEsDAO.java
```diff
@@ -62,8 +64,10 @@ public List<NetworkAddressAlias> loadLastUpdate(long timeBucketInMinute) {
 
             SearchResponse results =
                 getClient().search(NetworkAddressAlias.INDEX_NAME, search, params);
-            final String scrollId = results.getScrollId();
+            final Set<String> scrollIds = new HashSet<>();
```
wu-sheng, line 67: As the scroll IDs are changing, is a list enough for this?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/cache/NetworkAddressAliasEsDAO.java
```diff
@@ -62,8 +64,10 @@ public List<NetworkAddressAlias> loadLastUpdate(long timeBucketInMinute) {
...
             try {
+                final String scrollId = results.getScrollId();
+                scrollIds.add(scrollId);
```
wankai123, line 70: should inside 'while' ?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/cache/NetworkAddressAliasEsDAO.java
kezhenxu94: > As the scroll IDs are changing, is a list enough for this?

It's "sometimes changed" 🤦🏻, see the link I posted in comment

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/cache/NetworkAddressAliasEsDAO.java
kezhenxu94: Fixed

wankai123 (approved):

mrproliu (approved):

wu-sheng (approved):

PR #9280, opened 2022-06-28
Add `VIRTUAL` detect type to Process for Network Profiling

mrproliu:
When querying Process, exclude all Virtual type data, this part of the data will only be useful in Network Profiling.

### DIFF
--- a/oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBMetadataQueryDAO.java
+++ b/oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBMetadataQueryDAO.java
@@ -201,7 +201,7 @@ public class BanyanDBMetadataQueryDAO extends AbstractBanyanDBDAO implements IMe
     }
 
     @Override
-    public List<Process> listProcesses(String serviceId, String instanceId, String agentId, ProfilingSupportStatus profilingSupportStatus, long lastPingStartTimeBucket, long lastPingEndTimeBucket) throws IOException {
+    public List<Process> listProcesses(String serviceId, String instanceId, String agentId, ProfilingSupportStatus profilingSupportStatus, boolean excludeVirtual, long lastPingStartTimeBucket, long lastPingEndTimeBucket) throws IOException {
         MeasureQueryResponse resp = query(ProcessTraffic.INDEX_NAME,
                 PROCESS_TRAFFIC_TAGS,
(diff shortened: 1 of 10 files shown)


oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBMetadataQueryDAO.java
```diff
@@ -201,7 +201,7 @@ protected void apply(MeasureQuery query) {
     }
 
     @Override
-    public List<Process> listProcesses(String serviceId, String instanceId, String agentId, ProfilingSupportStatus profilingSupportStatus, long lastPingStartTimeBucket, long lastPingEndTimeBucket) throws IOException {
+    public List<Process> listProcesses(String serviceId, String instanceId, String agentId, ProfilingSupportStatus profilingSupportStatus, boolean excludeVirtual, long lastPingStartTimeBucket, long lastPingEndTimeBucket) throws IOException {
```
wu-sheng, line 204: In any case, would `boolean excludeVirtual` be false?

mrproliu: `excludeVirtual` should be true to exclude all the virtual processes. Do you mean we don't need this parameter, just hard code to exclude all virtual processes in the query?

wu-sheng: Yes, I mean do we need in any case in the future, we would set this to `false`?

mrproliu: No scenario for now. The network topology query should query the processes through ID only.

wu-sheng: Then, let's remove this parameter until we need it. OK?

mrproliu: Sure

wu-sheng (approved):

PR #9299, opened 2022-07-03
Add `forEach`, `processRelation` function to MAL Expression, and add `expPrefix` in MAL config

mrproliu:
I'm preparing network profiling based on eBPF, it needs to complete the process relation through MAL. 
1. Add `forEach`, `processRelation` function to MAL Expression.
2. Add `expPrefix` in MAL config.
3. Add `ProcessRegistry` to generate the process dynamically. 

Here is an example, of the configuration I will add after the network profiling feature is complete. 
```yaml
expSuffix: |-
  processRelation('side', ['service'], ['instance'], 'client_process_id', 'server_process_id')
expPrefix: |-
  forEach(['client', 'server'], { prefix, tags ->
    if (tags[prefix + '_process_id'] != null) {
      return
    }
    if (tags[prefix + '_local'] == 'true') {
      tags[prefix + '_process_id'] = ProcessRegistry.generateVirtualLocalProcess(tags.service, tags.instance)
      return
    }
    tags[prefix + '_process_id'] = ProcessRegistry.generateVirtualRemoteProcess(tags.service, tags.instance, tags[prefix + '_address'])
  })
metricPrefix: network_profiling
metricsRules:
  - name: write_bytes
    exp:  rover_network_monitor_write_bytes_counter.sum(['service', 'instance', 'side', 'client_process_id', 'server_process_id'])
```

### DIFF
--- a/docs/en/concepts-and-designs/mal.md
+++ b/docs/en/concepts-and-designs/mal.md
@@ -211,6 +211,10 @@ Examples:
 #### time
 `time()`: Returns the number of seconds since January 1, 1970 UTC.
 
+#### foreach
+`forEach([string_array], Closure<Void> each)`: Iterates all samples according to the first array argument, and provide two parameters in the second closure argument:
+1. `element`: element in the array.
+2. `tags`: tags in each sample.
 
 ## Down Sampling Operation
 MAL should instruct meter-system on how to downsample for metrics. It doesn't only refer to aggregate raw samples to
@@ -245,6 +249,8 @@ They extract level relevant labels from metric labels, then informs the meter-sy
                                                                       extracts endpoint level labels from the second array argument, extracts layer from `Layer` argument.
  - `serviceRelation(DetectPoint, [source_svc_label1...], [dest_svc_label1...], Layer)` DetectPoint including `DetectPoint.CLIENT` and `DetectPoint.SERVER`,
    extracts `sourceService` labels from the first array argument, extracts `destService` labels from the second array argument, extracts layer from `Layer` argument.
+ - `processRelation(detect_point_lable, [service_label1...], [instance_label1...], source_process_id_label, dest_process_id_label)` extracts `DetectPoint` labels from first argument, the label value should be `client` or `server`.
+   extracts `Service` labels from the first array argument, extracts `Instance` labels from the second array argument, extracts `ProcessID` labels from the fourth and fifth arguments of the source and destination.
 
 ## More Examples
 
--- a/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/meter/config/MeterConfig.java
+++ b/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/meter/config/MeterConfig.java
@@ -29,6 +29,7 @@ import java.util.List;
 public class MeterConfig implements MetricRuleConfig {
     private String metricPrefix;
     private String expSuffix;
+    private String expPrefix;
     private String filter;
     private List<Rule> metricsRules;
 
--- a/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/EntityDescription/ProcessRelationEntityDescription.java
+++ b/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/EntityDescription/ProcessRelationEntityDescription.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.oap.meter.analyzer.dsl.EntityDescription;
+
+import com.google.common.collect.ImmutableList;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.ToString;
+import org.apache.skywalking.oap.server.core.analysis.meter.ScopeType;
+
+import java.util.List;
+
+@Getter
+@RequiredArgsConstructor
+@ToString
+public class ProcessRelationEntityDescription implements EntityDescription {
+    private final ScopeType scopeType = ScopeType.PROCESS_RELATION;
+    private final List<String> serviceKeys;
+    private final List<String> instanceKeys;
+    private final String sourceProcessIdKey;
+    private final String destProcessIdKey;
+    private final String detectPointKey;
+    private final String delimiter;
+
+    @Override
+    public List<String> getLabelKeys() {
+        return ImmutableList.<String>builder()
+                .addAll(serviceKeys)
+                .addAll(instanceKeys)
+                .add(detectPointKey, sourceProcessIdKey, destProcessIdKey).build();
+    }
+}
--- a/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/SampleFamily.java
+++ b/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/SampleFamily.java
@@ -23,9 +23,12 @@ import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.mapping;
 import static java.util.stream.Collectors.toList;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
+
+import org.apache.commons.lang3.StringUtils;
 import org.apache.skywalking.oap.meter.analyzer.dsl.EntityDescription.EndpointEntityDescription;
 import org.apache.skywalking.oap.meter.analyzer.dsl.EntityDescription.EntityDescription;
 import org.apache.skywalking.oap.meter.analyzer.dsl.EntityDescription.InstanceEntityDescription;
+import org.apache.skywalking.oap.meter.analyzer.dsl.EntityDescription.ProcessRelationEntityDescription;
 import org.apache.skywalking.oap.meter.analyzer.dsl.EntityDescription.ServiceEntityDescription;
 import org.apache.skywalking.oap.meter.analyzer.dsl.EntityDescription.ServiceRelationEntityDescription;
 import org.apache.skywalking.oap.meter.analyzer.dsl.tagOpt.K8sRetagType;
@@ -36,6 +39,7 @@ import org.apache.skywalking.oap.server.core.analysis.meter.MeterEntity;
 import org.apache.skywalking.oap.server.core.analysis.meter.ScopeType;
 import org.apache.skywalking.oap.server.core.source.DetectPoint;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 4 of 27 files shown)


oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/k8s/K8sInfoRegistry.java
```diff
@@ -213,12 +233,35 @@ private void recompose() {
                 }
             });
         });
+        nameServiceMap.forEach((serviceName, service) -> {
+            if (isNull(service.getMetadata()) || isNull(service.getStatus()) || isNull(service.getStatus().getLoadBalancer())) {
```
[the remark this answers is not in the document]
mrproliu: @sonatype-lift ignore

docs/en/concepts-and-designs/mal.md
```diff
@@ -245,6 +249,8 @@ They extract level relevant labels from metric labels, then informs the meter-sy
                                                                       extracts endpoint level labels from the second array argument, extracts layer from `Layer` argument.
  - `serviceRelation(DetectPoint, [source_svc_label1...], [dest_svc_label1...], Layer)` DetectPoint including `DetectPoint.CLIENT` and `DetectPoint.SERVER`,
    extracts `sourceService` labels from the first array argument, extracts `destService` labels from the second array argument, extracts layer from `Layer` argument.
+ - `processRelation(detect_point_lable, [service_label1...], [instance_label1...], source_process_id_label, dest_process_id_label)` extracts `DetectPoint` labels from first argument, the label value should be `client` or `server`.
```
wu-sheng, line 252:
```suggestion
 - `processRelation(detect_point_label, [service_label1...], [instance_label1...], source_process_id_label, dest_process_id_label)` extracts `DetectPoint` labels from first argument, the label value should be `client` or `server`.
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/IDManager.java
```diff
@@ -268,6 +268,32 @@ public static String buildId(String instanceId, String name) {
             return Hashing.sha256().newHasher().putString(String.format("%s_%s",
                     name, instanceId), Charsets.UTF_8).hash().toString();
         }
+
+        /**
+         * @return encoded process relation id
+         */
+        public static String buildRelationId(ProcessRelationDefine define) {
+            return define.sourceId + Const.RELATION_ID_CONNECTOR + define.destId;
```
wu-sheng, line 276: Is the length of process relation ID enough for source and dest IDs? What is the max length of process ID?

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/SampleFamily.java
```diff
@@ -497,6 +501,38 @@ public SampleFamily serviceRelation(DetectPoint detectPoint, List<String> source
         return createMeterSamples(new ServiceRelationEntityDescription(sourceServiceKeys, destServiceKeys, detectPoint, layer, Const.POINT));
     }
 
+    public SampleFamily forEach(List<String> array, Closure<Void> each) {
```
wu-sheng, line 504: @hanahmily @kezhenxu94 @wankai123 Could you recheck here?

Array is a static list in MAL.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/IDManager.java
mrproliu: The string fixed length of a single process ID is `64`. The length of two process IDs + `Const.RELATION_ID_CONNECTOR` is `129`, so it's ok to be the entity ID.

wu-sheng (approved):
LGTM

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/registry/ProcessRegistry.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * Generate virtual remote process under the instance
+     * trying to generate the name in the kubernetes environment through the remote address
+     * @return the process id
+     */
+    public static String generateVirtualRemoteProcess(String service, String instance, String remoteAddress) {
+        // trigger K8sInfoRegistry to start
+        K8sInfoRegistry.getInstance().start();
```
wankai123, line 53: I think it should start in the `Analyzer` init. Otherwise, the metadata may be empty in the beginning.

mrproliu: Yes. But there is no way to know whether the Kubernetes has been used during the initialization phase because the method `ProcessRegistry#generateVirtualRemoteProcess` is only called in the `Closure`. 
Can we add a new config in the `MetricRuleConfig` to ensure the Kubernetes metadata is needed? Or do you have other good suggestions for it?

wu-sheng: I think we should add a `init` stage for MAL file?

mrproliu: Yes, we could do that. So we can add the configuration item in the MAL file to start the Kubernetes Registry. WDYT?

wankai123: Make sense. This will explicitly indicate that the K8s API server is needed here.

kezhenxu94:
I have some concerns about the new `forEach` function. See my comments inline. 

> 1. Add `forEach`, `processRelation` function to MAL Expression.

`xxx.forEach(yyy)` usually iterates over `xxx` but in this PR it instead iterates `yyy`, this is confusing for most language programmers in my opinion. 

> 
> 
> Here is an example, I will add the configuration after the network analysis function is complete.
> 
> ```yaml
> 
> expSuffix: |-
> 
>   processRelation('side', ['service'], ['instance'], 'client_process_id', 'server_process_id')
> 
> expPrefix: |-
> 
>   forEach(['client', 'server'], { prefix, tags ->
> 
>     if (tags[prefix + '_process_id'] != null) {
> 
>       return
> 
>     }
> 
>     if (tags[prefix + '_local'] == 'true') {
> 
>       tags[prefix + '_process_id'] = ProcessRegistry.generateVirtualLocalProcess(tags.service, tags.instance)
> 
>       return
> 
>     }
> 
>     tags[prefix + '_process_id'] = ProcessRegistry.generateVirtualRemoteProcess(tags.service, tags.instance, tags[prefix + '_address'])
> 
>   })
> 
> metricPrefix: network_profiling
> 
> metricsRules:
> 
>   - name: write_bytes
> 
>     exp:  rover_network_monitor_write_bytes_counter.sum(['service', 'instance', 'side', 'client_process_id', 'server_process_id'])
> 
> ```

I think we expose too many details/logics in the MAL. I'm wondering why we can't use the similar style as `retagByK8sMeta`.

wu-sheng:
We discussed that two days, that way is much more complex, as many keys have to be exposed.
We have deno codes about what that would look like, it is really worse.
image

There are nothing changed better, as we have to expose process concept as SampleFamily function.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/SampleFamily.java
hanahmily: It's safe to iterate a static array concurrently. But `foreach` is a rare operation in streaming/functional style language. How about picking up [map](https://en.wikipedia.org/wiki/Map_(higher-order_function)) instead? We could pass each sample into the closure to update both values and labels. That seems a more resilient function.

wu-sheng:
When we tried `retag` without closure, the function in SampleFamily would look like this

image

It is even harder to understand what happens after whose methods. And we make the logic boundaries of SampleFamily are not clear. 
So, we got a conclusion to do like this, separate `forEach`, `createVirtualProcess` and `how to analysis the meter/metric data` logic.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/SampleFamily.java
mrproliu: `map` is ok to me. But we need to discuss some details. If we use the closure to update the value, we can add `sample` to the closure as the argument which has `tags` and `value` fields, and we can update them in the closure. for the iterate a static array, we can add it in the closure. here is a demo, WDYT?
```yaml
expPrefix: |-
  map({ sample ->
    // dynamic update sample value
    sample.value = sample.value * 2
    // iterate a static array
    def lst = ['client', 'server']
    lst.each { prefix ->
      // dynamic read or update the tags
      if (sample.tags[prefix + '_process_id'] != null) {
        return
      }
      if (sample.tags[prefix + '_local'] == 'true') {
        sample.tags[prefix + '_process_id'] = ProcessRegistry.generateVirtualLocalProcess(tags.service, tags.instance)
        return
      }
      sample.tags[prefix + '_process_id'] = ProcessRegistry.generateVirtualRemoteProcess(tags.service, tags.instance, tags[prefix + '_address'])
    }
  })
```

wu-sheng: This kind of map seems not better than foreach. 
map usually means iterating elements for the collection calling this method. such as list.map.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/registry/ProcessRegistry.java
mrproliu: I have added the `kubernetesAware` configuration in the MAL file, please recheck.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/meter/config/MeterConfig.java
```diff
@@ -29,8 +30,16 @@
 public class MeterConfig implements MetricRuleConfig {
     private String metricPrefix;
     private String expSuffix;
+    private String expPrefix;
     private String filter;
     private List<Rule> metricsRules;
+    private boolean kubernetesAware;
```
wu-sheng, line 36: How about adding an initExp? And put k8s listening function inside it?

In my understanding, MAL should be not environment sensitive. The functions should do the logic things. MAL is only a skeleton.

oap-server/server-starter/src/main/resources/otel-oc-rules/k8s-cluster.yaml
```diff
@@ -28,6 +28,7 @@
 #    "-P6H3M"    -- parses as "-6 hours and -3 minutes"
 #    "-P-6H+3M"  -- parses as "+6 hours and -3 minutes"
 # </pre>
+initExp: Kubernetes.startMetadataListener()
```
wu-sheng, line 31: @wankai123 We add this for opening listener explictly.

wankai123 (approved):

wu-sheng (approved):
LGTM

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/EntityDescription/ProcessRelationEntityDescription.java
```diff
@@ -0,0 +1,48 @@
...
+import lombok.ToString;
+import org.apache.skywalking.oap.server.core.analysis.meter.ScopeType;
+
+import java.util.List;
+
+@Getter
+@RequiredArgsConstructor
+@ToString
+public class ProcessRelationEntityDescription implements EntityDescription {
+    private final ScopeType scopeType = ScopeType.PROCESS_RELATION;
```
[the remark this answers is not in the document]
mrproliu: @sonatype-lift ignore

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/registry/ProcessRegistry.java
kezhenxu94, line 53: You've added `initExp` but why is this not removed?

mrproliu: Ohh, I forget. Thanks.

kezhenxu94 (approved):

Issue #9297, opened 2022-07-01
[Feature] About support multi threads and auto finalize for skywalking-rust.


jmjoy:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Now the [`skywalking-rust`](https://github.com/apache/skywalking-rust) only support one thread tracing, but usually there are some scenes that require multi threads/coroutines tracing.

### Use case

Rust application requires multi threads and coroutines tracing.

### Related issues

<https://github.com/apache/skywalking-rust/pull/20#issuecomment-1153411401>
<https://github.com/apache/skywalking-rust/pull/26>
<https://github.com/apache/skywalking-rust/pull/28>

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

jmjoy:
I think the new api design is (fake code):

```rust
use skywalking_rust::context::trace_context::TracingContext;
use skywalking_rust::reporter::grpc::Reporter;
use tokio;

static TRACING_MANAGER: Lazy<TracingManager> = Lazy::new(|| TracingManager::default("svc", "ins"));

async fn handle_request(reporter: ContextReporter) {
    let mut ctx = TRACING_MANAGER.create_context();

    {
        // Generate an Entry Span when a request
        // is received. An Entry Span is generated only once per context.
        let span = ctx.create_entry_span("operation1").unwrap();

        // Something...

        {
            // Generates an Exit Span when executing an RPC.
            let span2 = ctx.create_exit_span("operation2").unwrap();
            
            // Something...

            // Drop span2 and finalize to ctx.
        }

        // Drop span and finalize to ctx.
    }

    // Create coroutine to handle async task.
    spawn(async move {
        let mut ctx2 = TRACING_MANAGER.create_context();

        {
            let span3 = ctx2.create_entry_span("operation3").unwrap();

            // Something...

            // Drop span2 and finalize to ctx2.
        }

        // Drop ctx2 and send to TRACING_MANAGER.
    });

    // Drop ctx and send to TRACING_MANAGER.
}

#[tokio::main]
async fn main() {
    let tx = Reporter::start("http://0.0.0.0:11800").await;

    // Start server...

    // Block to report.
    loop {
        let context = TRACING_MANAGER.receive_context();
[...]

PR #9300, opened 2022-07-03
Fix missing sourceEndpointOwnerServiceLayer

lujiajing1126:

This is found in issue https://github.com/apache/skywalking/issues/9297

The missing LOC here will cause NPE and thus failure to save LocalSpan annotated with `MQ` kind.


wu-sheng:
Please update change log, about what kinds if endpoint dependency could be missed.

lujiajing1126:
> Please update change log, about what kinds if endpoint dependency could be missed.

Updated

wu-sheng (approved):

PR #9338, opened 2022-07-13
Skip loading UI templates if folder is empty or doesn't exist

kezhenxu94:
UI templates are configurable, users might add or remove configs
as they want, so we don't need to throw exception for inexistent
paths, this allow users to remove templates they don't want, or
even disable the `ui_template` index.

Also remove util method only used in test codes.


wankai123 (approved):
LGTM

PR #9339, opened 2022-07-13
Optimize elasticsearch query performance by using `_mGet` and physical index name

wankai123:


oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -300,13 +303,33 @@ public boolean existDoc(String indexName, String id) {
...
+        return es.get().documents().mGet(TYPE, map);
+    }
+
+    /**
+     * Query by ids with index alias. When can not locate the physical index.
+     * @param indexAlias Index alias name
+     * @param ids Query ids
+     * @return SearchResponse
+     */
+    public SearchResponse idsWithIndexAlias(String indexAlias, Iterable<String> ids) {
```
kezhenxu94, line 326: What about just keep the method name `ids` and parameter `index`? This method still can query by both physical index or alias. Just update the java doc `@param index Index name or alias name`

kezhenxu94 (commented):
Some nits

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
wankai123: If users can locate the physical index, I prefer to recommend users use `_mGet` way,  I just want to use the method and parameter name to remind users. WDYT @wu-sheng

wu-sheng: Where would we use `idsWithIndexAlias`?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/MetricsEsDAO.java
```diff
@@ -52,50 +52,47 @@ protected MetricsEsDAO(ElasticSearchClient client,
...
         // The groupIndices mostly include one or two group,
         // the current day and the T-1 day(if at the edge between days)
         List<Metrics> result = new ArrayList<>(metrics.size());
-        groupIndices.forEach((tableName, metricList) -> {
+        groupAliasIndices.forEach((tableName, metricList) -> {
             List<String> ids = metricList.stream()
                                          .map(item -> IndexController.INSTANCE.generateDocId(model, item.id()))
                                          .collect(Collectors.toList());
-            final SearchResponse response = getClient().ids(tableName, ids);
+            final SearchResponse response = getClient().idsWithIndexAlias(tableName, ids);
```
wankai123, line 76: When the metric id is not TimeRelative. Could not locate the physical index.

wankai123: @wu-sheng wu

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/MetricsEsDAO.java
```diff
@@ -52,50 +52,47 @@ protected MetricsEsDAO(ElasticSearchClient client,
...
             response.getHits().getHits().forEach(hit -> {
                 Metrics source = storageBuilder.storage2Entity(new HashMapConverter.ToEntity(hit.getSource()));
                 result.add(source);
             });
         });
 
+        Map<String, List<String>> indexIdsGroup = new HashMap<>();
```
kezhenxu94, line 83: I think we can merge `groupAliasIndices` into `groupIndices` and call `getClient().ids` once, then remove `idsWithIndexAlias`.

wankai123: `_mget` doesn't support alias if there multi indexes... Do you have a better idea?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/MetricsEsDAO.java
wu-sheng: I think renaming `idsWithIndexAlias` to `searchIDs` would make things more clear.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/MetricsEsDAO.java
wu-sheng: We don't have to use `mget`. Like https://github.com/apache/skywalking/pull/9339#discussion_r919702835 mentioned, we could keep alias.

kezhenxu94: > `_mget` doesn't support alias if there multi indexes... Do you have a better idea?

Well. Then renaming the method to what @wu-sheng suggest is ok to me

docs/en/changes/changes.md
```diff
@@ -21,6 +21,7 @@
 * Remove legacy OAL `percentile` functions, `p99`, `p95`, `p90`, `p75`, `p50` func(s).
 * Revert [#8066](https://github.com/apache/skywalking/pull/8066). Keep all metrics persistent even it is default value.
 * Skip loading UI templates if folder is empty or doesn't exist.
+* Optimize elasticsearch query performance by using `_mGet` and physical index name.
```
wu-sheng, line 24:
```suggestion
* Optimize ElasticSearch query performance by using `_mGet` and physical index name rather than alias in these scenarios,  (a) Metrics aggregation (b) Zipkin query (c) Metrics query
```

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetricsQueryEsDAO.java
```diff
@@ -77,10 +82,10 @@ public long readMetricsValue(final MetricsCondition condition,
 
         sourceBuilder.aggregation(entityIdAggregation);
 
-        final String index =
-            IndexController.LogicIndicesRegister.getPhysicalTableName(condition.getName());
-
-        final SearchResponse response = getClient().search(index, sourceBuilder.build());
+        final SearchResponse response = getClient().search(new TimeRangeIndexNameGenerator(
```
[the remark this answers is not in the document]
wu-sheng: If this works, then `AggregationQueryEsDAO` should be changed in a similar way too.

Both are aggregation queries.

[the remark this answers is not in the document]
wankai123: yes, it's work. This API supported https://github.com/apache/skywalking/blob/bf74b667639489e51db016aad575d8b8c4247e26/oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java#L251

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/MetricsEsDAO.java
```diff
@@ -52,50 +52,49 @@ protected MetricsEsDAO(ElasticSearchClient client,
...
-                                 return IndexController.INSTANCE.getTableName(model);
+                                 String indexName = IndexController.INSTANCE.getTableName(model);
+                                 groupAliasIndices.computeIfAbsent(indexName, v -> new ArrayList<>()).add(metric);
                              }
-                         })
-                     );
-
+                         });
         // The groupIndices mostly include one or two group,
         // the current day and the T-1 day(if at the edge between days)
```
wu-sheng, line 70: I think this comment is not correct anymore. The alias is only used for `!isTimeRelativeID` entities only, right? It doesn't matter at the day edge or not anymore.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/MetricsEsDAO.java
```diff
@@ -52,50 +52,49 @@ protected MetricsEsDAO(ElasticSearchClient client,
 
     @Override
     public List<Metrics> multiGet(Model model, List<Metrics> metrics) {
-        Map<String, List<Metrics>> groupIndices
-            = metrics.stream()
-                     .collect(
-                         groupingBy(metric -> {
+        Map<String, List<Metrics>> groupAliasIndices = new HashMap<>();
+        Map<String, List<Metrics>> groupIndices = new HashMap<>();
```
wu-sheng, line 56: First of all, we don't need to create two Map(s) always. For a particular model, it either uses `alias` or `mget`.  `model#isTimeRelativeID()` is the condition, so, wouldn't be affected by the metrics list.

Also, I noticed, that for alias usage, there is no point to use a Map, as you are going to have one alias(`IndexController.INSTANCE.getTableName(model)` returns consistent).

wu-sheng: You could use `if/else` or `return` fast in these two cases to avoid unnecessary costs.

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -300,7 +303,28 @@ public boolean existDoc(String indexName, String id) {
...
+        });
+        return es.get().documents().mGet(TYPE, map);
+    }
+
+    /**
+     * Search by ids with index alias.
+     * When can not locate the physical index. Otherwise, recommend use method {@link #ids}
+     * @param indexName Index alias name or physical name
+     * @param ids ids
+     * @return SearchResponse
```
wu-sheng, line 325:
```suggestion
     * Search by ids with index alias, when can not locate the physical index. 
     * Otherwise, recommend use method {@link #ids}
     * @param indexName Index alias name or physical name
     * @param ids ID list
     * @return SearchResponse
     * @since 9.2.0 this method was ids
```

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -300,7 +303,28 @@ public boolean existDoc(String indexName, String id) {
         return es.get().documents().exists(indexName, TYPE, id);
     }
 
-    public SearchResponse ids(String indexName, Iterable<String> ids) {
+
+    /**
+     * @since 9.2.0 Provide to get documents from multi indices by ids.
+     * @param indexIds key: indexName, value: ids list
+     * @return Documents
```
wu-sheng, line 310:
```suggestion
     * Provide to get documents from multi indices by IDs.
     * @param indexIds key: indexName, value: ids list
     * @return Documents
     * @since 9.2.0
```

oap-server/server-library/library-client/src/main/java/org/apache/skywalking/oap/server/library/client/elasticsearch/ElasticSearchClient.java
```diff
@@ -300,7 +303,28 @@ public boolean existDoc(String indexName, String id) {
...
+     */
+    public Optional<Documents> ids(Map<String, List<String>> indexIds) {
```
wu-sheng, line 312: Why do other search APIs return `SearchResponse`, but here returning `Optional<Documents>`. Is this on purpose?

wankai123: The origin ES API returned differently

wu-sheng (approved):
LGTM. Wait for e2e

kezhenxu94 (approved):

PR #9347, opened 2022-07-18
Dashboard: Add metrics association and fix.

wankai123:
- Add metrics associate.
- Fix `FaaS-Root` document link and topology service relation dashboard link.
- Fix `Mesh-Instance` metric `Throughput`.
- Fix `Mesh-Service-Relation` metric `Throughput` and `Proxy Sidecar Internal Latency in Nanoseconds (Client Response)`.
- Fix `Mesh-Instance-Relation` metric `Throughput`.

**Example** 

image

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -41,6 +41,11 @@
 * Fix tag dropdown style.
 * Hide the copy button when db.statement is empty.
 * Fix legend metrics for topology.
+* Dashboard: Add metrics associate.
+* Dashboard: Fix `FaaS-Root` document link and topology service relation dashboard link.
+* Dashboard: Fix `Mesh-Instance` metric `Throughput`.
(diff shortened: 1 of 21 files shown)


docs/en/changes/changes.md
```diff
@@ -41,6 +41,11 @@
 * Fix tag dropdown style.
 * Hide the copy button when db.statement is empty.
 * Fix legend metrics for topology.
+* Dashboard: Add metrics associate.
```
wu-sheng, line 44:
```suggestion
* Dashboard: Add metrics association.
```
I think associate is a verb., right?

wu-sheng: Or do you mean make metrics associate?

wankai123: `association` is better

wu-sheng (approved):
LGTM. @kezhenxu94 It is better you could confirm the added metrics in the mesh dashboard.

PR #9407, opened 2022-07-29
Add a new layer for MYSQL

yswdqz:
Add a new layer for MYSQL.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/Layer.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/Layer.java
@@ -116,7 +116,12 @@ public enum Layer {
     /**
      * Kubernetes service
      */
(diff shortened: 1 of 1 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/Layer.java
```diff
@@ -116,7 +116,12 @@ public enum Layer {
     /**
      * Kubernetes service
      */
-    K8S_SERVICE(17, true);
+    K8S_SERVICE(17, true),
+
+    /**
+     * Telemetry from MYSQL
```
wu-sheng, line 122:
```suggestion
     * mysqld, also known as MySQL Server, is a single multithreaded program that does most of the work in a MySQL installation. It does not spawn additional processes.
```

wu-sheng (approved):

PR #9430, opened 2022-08-05
Support analyze the component in the eBPF Network profiling

mrproliu:


oap-server/server-starter/src/main/resources/component-libraries.yml
```diff
@@ -416,6 +416,12 @@ Apache-ShenYu:
 Hutool:
   id: 128
   languages: Java
+https:
+  id: 129
+  languages: Java,C#,Node.js,Python
+TLS:
+  id: 130
+  languages: Java,C#,Node.js,Python
```
wu-sheng, line 424: In the mesh protocol, TLS has two modes, do we need to have them too at here?

https://github.com/apache/skywalking-data-collect-protocol/blob/906a834dc0c3fb66719b8313f8f9298be3a27349/service-mesh-probe/service-mesh.proto#L54-L55

`http=49` is in lower case, `https=129` is in lower case, why `tls` is upper case?

I think you don't need to list these languages here, right? You should list `ebpf`?

oap-server/server-starter/src/main/resources/meter-analyzer-config/network-profiling.yaml
```diff
@@ -27,35 +27,51 @@ expPrefix: |-
...
+    String protocol = tags['protocol']
+    String ssl = tags['is_ssl']
+    if (protocol == 'http' && ssl == 'true') {
+      result = '129'  // https
+    } else if (protocol == 'http') {
+      result = '49'   // http
+    } else if (ssl == 'true') {
+      result = '130'  // TLS
+    } else {
+      result = '50'   // rpc
```
wu-sheng, line 42: Should this be `tcp=110`? RPC seems not accurate?

oap-server/server-starter/src/main/resources/component-libraries.yml
mrproliu: > In the mesh protocol, TLS has two modes, do we need to have them too at here?

For now, we only could support analyzing the TLS. mTLS is not supported.

oap-server/server-starter/src/main/resources/meter-analyzer-config/network-profiling.yaml
mrproliu: Yes, I think tcp is more accurate. Updated.

oap-server/server-starter/src/main/resources/component-libraries.yml
mrproliu: > `http=49` is in lower case, `https=129` is in lower case, why `tls` is upper case?
> 
> I think you don't need to list these languages here, right? You should list `ebpf`?

Updated.

docs/en/changes/changes.md
```diff
@@ -34,6 +34,7 @@
   In the current one index mode, users still could choose to adjust ElasticSearch's shard number(`SW_STORAGE_ES_INDEX_SHARDS_NUMBER`) to scale out.
 * [Breaking Change] Many columns of metrics and records model names are changed, The H2/Mysql/Tidb/Postgres storage users are required to remove all metrics-related and records-related tables for OAP to re-create or use a new database instance.
 * Fix Zipkin trace query the max size of spans.
+* Adding `tls` and `https` into component for Network Profiling.
```
wu-sheng, line 37:
```suggestion
* Add `tls` and `https` component IDs for Network Profiling.
```

wu-sheng (approved):

wu-sheng (approved):

PR #9437, opened 2022-08-08
[Critical] Fix a low performance issue of metrics persistent in ES

wu-sheng:

[Critical] Fix a low-performance issue of metrics persistent in the ElasticSearch storage implementation. One single metric could have to wait for an unnecessary 7~10s(System Env Variable `SW_STORAGE_ES_FLUSH_INTERVAL`).

This should a solid and critical performance issue, from the self-observability metric, we could find one metric that would need 7-10s to execute the data flush, which is not expected.

image

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -3,7 +3,8 @@
 #### Project
 
 * [Critical] Fix a low performance issue of metrics persistent in the ElasticSearch storage implementation. One single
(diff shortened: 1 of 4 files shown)


wankai123 (approved):

wankai123 (approved):

kezhenxu94 (approved):

mrproliu (approved):

wu-sheng:
image

I have confirmed from demo env. The flush execution perf has been fixed.

PR #9442, opened 2022-08-09
Support Elasticsearch column alias for the compatibility between storage logicSharding model and no-logicSharding model.

wankai123:

### DIFF
--- a/docs/en/FAQ/README.md
+++ b/docs/en/FAQ/README.md
@@ -11,6 +11,7 @@ These are known and frequently asked questions about SkyWalking. We welcome you 
 * [Compiling issues on Mac's M1 chip](How-to-build-with-mac-m1.md)
 
 ## Runtime
+* [New ElasticSearch storage option explanation in 9.2.0](New-ElasticSearch-storage-option-explanation-in-9.2.0.md)
 * [Version 9.x+ upgrade](v9-version-upgrade.md)
 * [Elasticsearch exception `type=version_conflict_engine_exception` since 8.7.0](es-version-conflict.md)
 * [Version 8.x+ upgrade](v8-version-upgrade.md)
@@ -29,7 +30,6 @@ These are known and frequently asked questions about SkyWalking. We welcome you 
 * [Compatibility with other javaagent bytecode processes](Compatible-with-other-javaagent-bytecode-processing.md)
 * [**Java agent memory leak** when enhancing `Worker thread` at Thread Pool](Memory-leak-enhance-Worker-thread.md)
 * [Thrift plugin](thrift-plugin.md)
-* [New ElasticSearch storage option explanation in 9.2.0](New-ElasticSearch-storage-option-explanation-in-9.2.0.md)
 
 ## UI
 * [What is **VNode**? And why does SkyWalking have that?](vnode.md)
--- a/docs/en/FAQ/README.md
+++ b/docs/en/FAQ/README.md
@@ -29,6 +29,7 @@ These are known and frequently asked questions about SkyWalking. We welcome you 
 * [Compatibility with other javaagent bytecode processes](Compatible-with-other-javaagent-bytecode-processing.md)
 * [**Java agent memory leak** when enhancing `Worker thread` at Thread Pool](Memory-leak-enhance-Worker-thread.md)
 * [Thrift plugin](thrift-plugin.md)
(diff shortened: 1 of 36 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/ElasticSearch.java
```diff
@@ -68,4 +68,18 @@ enum AnalyzerType {
...
     }
+
+    @Target({ElementType.FIELD})
+    @Retention(RetentionPolicy.RUNTIME)
+    @interface Column {
+
+        /**
+         * Warning: this is only used to solve the excited column conflict since we need support merge all metrics
+         * in one index or sharding storage type both. If create a new a column, we should avoid the compatibility issue
+         * between these 2 storage type rather than use this alias.
```
wu-sheng, line 79:
```suggestion
         * Warning: this is only used to solve the conflict among the exciting columns since we need support to merge all metrics
         * in one physical index template. When create a new column, we should avoid the compatibility issue
         * between these 2 storage modes rather than use this alias.
```

docs/en/changes/changes.md
```diff
@@ -45,6 +45,7 @@
   instance.
 * Fix Zipkin trace query the max size of spans.
 * Add `tls` and `https` component IDs for Network Profiling.
+* Support Elasticsearch column alias for the compatibility between storage logicSharding model and no-logicSharding model.
```
kezhenxu94, line 48: I think you'd better reword the change log item `* [Breaking Change] Many columns of metrics and records model names are changed, The H2/Mysql/Tidb/Postgres storage
  users are required to remove all metrics-related and records-related tables for OAP to re-create or use a new database
   instance.`

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/ElasticSearch.java
```diff
@@ -68,4 +68,18 @@ enum AnalyzerType {
             }
         }
     }
+
+    @Target({ElementType.FIELD})
+    @Retention(RetentionPolicy.RUNTIME)
+    @interface Column {
+
+        /**
+         * Warning: this is only used to solve the excited column conflict since we need support merge all metrics
```
kezhenxu94, line 77: What do you mean "excited"? Typo?

wankai123: should be `existed`...

docs/en/changes/changes.md
wu-sheng: Besides this, I would recommend writing a `FAQ - new ElasticSearch storage option explanation in 9.2.0` to guide users and link from change log and storage docs.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexController.java
```diff
@@ -161,6 +184,22 @@ public static List<ModelColumn> getPhysicalTableColumns(Model model) {
             return new ArrayList<>(PHYSICAL_INDICES_COLUMNS.get(tableName).values());
         }
 
+        public static String getPhysicalColumnName(String modelName, String columnName) {
```
wu-sheng, line 187:
```suggestion
        @Deprecated
        public static String getPhysicalColumnName(String modelName, String columnName) {
```

Let's declare this method `@Deprecated` too. And put comments about this method should only be used when the context involves the old column names

docs/en/FAQ/README.md
```diff
@@ -29,6 +29,7 @@ These are known and frequently asked questions about SkyWalking. We welcome you
 * [Compatibility with other javaagent bytecode processes](Compatible-with-other-javaagent-bytecode-processing.md)
 * [**Java agent memory leak** when enhancing `Worker thread` at Thread Pool](Memory-leak-enhance-Worker-thread.md)
 * [Thrift plugin](thrift-plugin.md)
+* [New ElasticSearch storage option explanation in 9.2.0](New-ElasticSearch-storage-option-explanation-in-9.2.0.md)
```
wu-sheng, line 32: The latest FAQ should be on top of the list.

wu-sheng (approved):

Issue #9429, opened 2022-08-05
[Feature] Support for Micronaut plugin


pg-yang:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

The first step is support    [ Micronaut Server Application](https://docs.micronaut.io/latest/guide/#creatingServer) and [Micronaut Client](https://docs.micronaut.io/latest/guide/#creatingClient)

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

pg-yang:
```
endTo receive actual data
To validate
[2022-08-08 15:06:30:826] [ERROR] - org.apache.skywalking.plugin.test.agent.tool.Main.verify(Main.java:53) - 
assert failed.
SegmentNotFoundException:
expected: 
  Segment:
  - span[-1, 0] GET:/micronaut/success

actual: 
  Segment[52b9160c868a4b7a95852f01acd16b25.37.16599711821650000] e
  expected:     Span[-1, 0] GET:/micronaut/success
  actual:       span[-1, 0] HEAD:/micronaut/healthCheck
  reason:       [operation name]: expected=>{GET:/micronaut/success}, actual=>{HEAD:/micronaut/healthCheck}

  Segment[52b9160c868a4b7a95852f01acd16b25.41.16599711841430000] e
  expected:     Span[-1, 0] GET:/micronaut/success
  actual:       span[-1, 0] GET:/micronaut/success
  reason:       [tag(url) value]: expected=>{/micronaut/success}, actual=>{http://localhost:8080/micronaut/success}


  Segment[52b9160c868a4b7a95852f01acd16b25.44.16599711842720000] e
  expected:     Span[-1, 0] GET:/micronaut/success
  actual:       span[-1, 0] GET:/micronaut/404
  reason:       [operation name]: expected=>{GET:/micronaut/success}, actual=>{GET:/micronaut/404}

  Segment[52b9160c868a4b7a95852f01acd16b25.48.16599711847720000] e
  expected:     Span[-1, 0] GET:/micronaut/success
  actual:       span[-1, 0] GET:/micronaut/error
  reason:       [operation name]: expected=>{GET:/micronaut/success}, actual=>{GET:/micronaut/error}

```
[...]

PR #9455, opened 2022-08-10
Add component id for Micronaut Http plugin

pg-yang:

Related issue : https://github.com/apache/skywalking/issues/9429
Need attach UI  submodule information. Right?

### DIFF
--- a/oap-server/server-starter/src/main/resources/component-libraries.yml
+++ b/oap-server/server-starter/src/main/resources/component-libraries.yml
@@ -425,6 +425,7 @@ tls:
 Micronaut:
   id: 131
   languages: Java
+
 # .NET/.NET Core components
 # [3000, 4000) for C#/.NET only
 AspNetCore:
--- a/oap-server/server-starter/src/main/resources/component-libraries.yml
+++ b/oap-server/server-starter/src/main/resources/component-libraries.yml
@@ -422,7 +422,9 @@ https:
 tls:
   id: 130
   languages: ebpf
-
+Micronaut:
+  id: 131
+  languages: Java
 # .NET/.NET Core components
 # [3000, 4000) for C#/.NET only
 AspNetCore:
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -81,6 +81,7 @@
 * Add the MYSQL layer and update layer routers.
 * Fix query order for trace list.
 * Add a calculation to convert seconds to days.
+* Add component ID(131) for Java Micronaut plugin
 
 #### Documentation
 


oap-server/server-starter/src/main/resources/component-libraries.yml
```diff
@@ -416,6 +416,12 @@ Apache-ShenYu:
 Hutool:
   id: 128
   languages: Java
+MicronautHttpServer:
+  id: 129
+  languages: Java
+MicronautHttpClient:
+  id: 130
+  languages: Java
```
wu-sheng, line 424: 1. **129** has been assigned, which is why I asked you to register this ID first.
2. If you want to have two IDs, one for the client, and the other for the server, then you need to set up a mapping to make sure if the agent is installed only on Client-server, the virtual node of server could have the correct icon.

pg-yang: The ID only used for mapping icon ?

wu-sheng: Not just that, it also represents the component of spans. If you use the same, the client and server spans are going to show the same component.
If you need two, choose 130 and 131, and set the mapping at the bottom of this file.

pg-yang: OK, I choose 130 and 131

wu-sheng: Sure.

oap-server/server-starter/src/main/resources/component-libraries.yml
```diff
@@ -690,3 +696,5 @@ Component-Server-Mappings:
   ClickHouse-jdbc-driver: ClickHouse
   Apache-Kylin-jdbc-driver: Apache-Kylin
   MysqlClient: Mysql
+  MicronautHttpClient: Micronaut
+  MicronautHttpServer: Micronaut
```
wu-sheng, line 700:
```suggestion
  MicronautHttpClient: MicronautHttpServer
```

You should either set a mapping like this, according to your ID registration, but this would not match your UI side icon name: MICRONAUT.png.

The server side name should match the icon's file name.

wu-sheng: There is only one way to show the icon, the component name is the icon name.

Please update this file according, and post your screenshot after you set everything correctly.

pg-yang: Sorry ,I should be stricter

pg-yang:
Finally , I choose a ID : 131 .

image

wu-sheng:
What are those services' question marks?

wu-sheng:
> Finally , I choose a ID : 131

Then, please remember to update Java agent plugin codes by using one and the new component ID only.

pg-yang:
> What are those services with question marks?

The one is invoke self by localhost, other one is call a common domain

oap-server/server-starter/src/main/resources/component-libraries.yml
```diff
@@ -422,7 +422,9 @@ https:
 tls:
   id: 130
   languages: ebpf
-
+Micronaut:
+  id: 131
+  languages: Java
```
wu-sheng, line 427:
```suggestion
  languages: Java

```

wu-sheng (approved):

PR #9471, opened 2022-08-15
Fix query history process data failure

mrproliu:
### Fix Fix query history process data failure
The process data query is based on the `lastPingTimestamp` of the `ProcessTraffic`.So, if we add the end time in the query condition, then we cannot query the history data. So In this PR, I delete the end time filter of the process query to fix it. 

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -50,6 +50,7 @@
 * Support MySQL monitoring.
 * Fix query services by serviceId error when Elasticsearch storage `SW_STORAGE_ES_QUERY_MAX_SIZE` > 10000.
 * Support sending alarm messages to Discord.
+* Fix query history process data failure.
 
 #### UI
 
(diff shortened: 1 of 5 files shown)


wu-sheng:
I think the current logic should be either
1. start time before last ping time
2. end time after start time.

The instance logic has a similar query, right?

mrproliu:
>The instance logic has a similar query, right?

Yes. After this PR, the process query is similar to the instance query. They are both not using the `end time` in the query. 

>end time after start time.

Since the end time is not used, so there is no such judgment.

wu-sheng:
I mean if you don't use that. Then I could use an illegal time range to get the data, which is also a bug.

mrproliu:
Ok, So I think I can add a judgment to verify that the end time is after the start time. If the validation fails, then just return an empty list.

wu-sheng (approved):

PR #9490, opened 2022-08-19
Add AsyncPG and AIORedis components for Python plugin

tom-pytel:
Component IDs for [#230](https://github.com/apache/skywalking-python/pull/230)


wu-sheng (approved):

wu-sheng:
I think we would need your PR to add new icons for UI?

tom-pytel:
> I think we would need your PR to add new icons for UI?

I didn't find any specific icons for these, they are just Postgres and Redis clients, could reuse those icons?

wu-sheng:
Then you should add mappings in this file, map this client component to server.
Then when exit span accessing the server, the server could show the correct icon.
Check the bottom of the file.

PR #9220, opened 2022-06-14
Adopt license eye to check dependencies' license

kezhenxu94:
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>.
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/changelog/docs/en/changes/changes.md).

This is still not perfect, for example there are some duplicate dependencies that the maven resolver needs more enhancements, but that doesn't matter too much IMO.

PR #9493, opened 2022-08-20
Add Kubernetes support module to share codes between modules and reduce calls to Kubernetes API server

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number> NO.

Previously, we add the Kubernetes watcher in meter-analyzer and receiver-plugin respectively, this causes unnecessary duplicate API calls to Kubernetes API server, in large cluster this would be a problem, this patch introduces a module for Kubernetes support and make the calls unified/unique but interesting parts can register callbacks/listeners and when events occurred the interesting parts can be notified.

This should reduce the calls to Kubernetes API server to ~50%


wu-sheng:
Some UTs seem failing. Please fix.

wu-sheng (approved):
Good to me, once all tests passed.

wu-sheng:
@kezhenxu94 Please read the CVEs, the tool could do false alarm, but we at least should recheck.

kezhenxu94:
> @kezhenxu94 Please read the CVEs, the tool could do false alarm, but we at least should recheck.

Maybe I can bump up the Kubernetes client SDK version after https://github.com/apache/skywalking/pull/9220, so that I don't have to fix the unnecessary conflicts..

wankai123 (approved):

PR #9499, opened 2022-08-22
Enable otlp OC handler by default and polish doc

kezhenxu94:


oap-server/server-starter/src/main/resources/application.yml
```diff
@@ -338,7 +338,7 @@ receiver-meter:
 receiver-otel:
   selector: ${SW_OTEL_RECEIVER:default}
   default:
-    enabledHandlers: ${SW_OTEL_RECEIVER_ENABLED_HANDLERS:"oc"}
+    enabledHandlers: ${SW_OTEL_RECEIVER_ENABLED_HANDLERS:"oc,otlp"}
```
kezhenxu94, line 341: We cannot merge `enabledOcRules` and `enabledOtlpRules` because the rules are read by handler name `oc`/`otlp` like `enabled<handlerName>Rules`

wu-sheng (dismissed):

wankai123 (approved):

oap-server/server-starter/src/main/resources/application.yml
wu-sheng: I think we could make two handlers read the same things?
Do you have concerns?

kezhenxu94: It's hardly possible to do that on the basis of the current code. Read this 

https://github.com/apache/skywalking/blob/dc7c42cd97c37ee377ea144364c15d41c1099864/oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/OtelMetricReceiverConfig.java#L44-L61

The rules are read by handler `getEnabledRulesFrom(String handler)`, so if we want to share the rules for `oc` and `otlp`, we need either:

- hard code the returned rules for `oc` and `otlp` handlers.
- OR, let handler read the rule file on their own, instead of here:

https://github.com/apache/skywalking/blob/dc7c42cd97c37ee377ea144364c15d41c1099864/oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/OtelMetricReceiverProvider.java#L70-L72

Let me try the second way

wu-sheng: If it is too complex, it is fine to keep for now. Just seem a little strange.

wu-sheng (dismissed):

wu-sheng (approved):

wu-sheng (approved):

PR #9508, opened 2022-08-23
Fix JDBC TTL to delete additional tables data.

wankai123:
Previous version the JDBC additional table missed  column `timeBucket`, and the TTL timer couldn't delete data from those 
tables.

### DIFF
--- a/.github/workflows/skywalking.yaml
+++ b/.github/workflows/skywalking.yaml
@@ -603,7 +603,7 @@ jobs:
         with:
           go-version: '1.16'
       - name: ${{ matrix.test.name }}
-        uses: apache/skywalking-infra-e2e@main
+        uses: apache/skywalking-infra-e2e@d71bbba95f95d6629db199e4bc0c0e097ae4f2dc
         env:
           ISTIO_VERSION: ${{ matrix.istio_version }}
           ALS_ANALYZER: ${{ matrix.analyzer }}
@@ -645,7 +645,7 @@ jobs:
         with:
           go-version: '1.16'
       - name: Java version ${{ matrix.java-version }}
-        uses: apache/skywalking-infra-e2e@main
+        uses: apache/skywalking-infra-e2e@d71bbba95f95d6629db199e4bc0c0e097ae4f2dc
         env:
           SW_AGENT_JDK_VERSION: ${{ matrix.java-version }}
         with:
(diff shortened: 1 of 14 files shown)


wu-sheng (dismissed):

wu-sheng (dismissed):

kezhenxu94:
We need to pin the e2e to a fixed version https://github.com/apache/skywalking/blob/77ba2d553d77a4acacd8b2498cb09ea983a9baf4/.github/workflows/skywalking.yaml#L567

.github/workflows/skywalking.yaml
```diff
@@ -564,7 +564,7 @@ jobs:
         run: |
           echo "${{ matrix.test.env }}"  >> $GITHUB_ENV
       - name: ${{ matrix.test.name }}
-        uses: apache/skywalking-infra-e2e@main
+        uses: apache/skywalking-infra-e2e@d71bbba95f95d6629db199e4bc0c0e097ae4f2dc
```
wu-sheng, line 567: FYI @kezhenxu94 Feel free to update the test policy after this PR is merged.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/SQLDatabase.java
```diff
@@ -113,4 +113,26 @@
...
+    @Target({ElementType.TYPE})
+    @Retention(RetentionPolicy.RUNTIME)
+    @Repeatable(MultipleExtraColumn4AdditionalEntity.class)
+    @interface ExtraColumn4AdditionalEntity {
+        String additionalTable();
+        String parentColumn();
+    }
+
+    /**
+     * The support of the multiple {@link ExtraColumn4AdditionalEntity}s on one field.
```
wu-sheng, line 131:
```suggestion
     * The support of the multiple {@link ExtraColumn4AdditionalEntity}s on the class.
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/SQLDatabase.java
```diff
@@ -113,4 +113,26 @@
         String[] additionalTables();
         boolean reserveOriginalColumns() default false;
     }
+
+    /**
+     * Support add an extra column from parent to an additional table.
+     * This column will be created in both the primary and additional tables.
+     * Notice: This annotation should be declared on the subclasses.
```
wu-sheng, line 120:
```suggestion
     * Support add an extra column from the parent classes as a column of the additional table.
     * This column would be created in both the primary and additional tables.
     * Notice: This annotation should be declared on the leaf subclasses.
```

wu-sheng (dismissed):

wu-sheng (approved):

PR #9513, opened 2022-08-23
Fix dependency license's NOTICE and binary jar included issues in the source release.

### LINKED ISSUE #9512: [Bug] Fix dependency license's NOTICE and binary jar included issues in the source release.
### Search before asking

### What happened

We received reports from incubator discussion

```
So taking a quick look at a SkyWalking release [1] I can see a couple of minor issues with LICNSE and NOTICE. You may want to fix these.

Hints:
Is OpenCensus mentioned in LICENSE?
Does prometheus client_model have a NOTICE file?
Does mavin-wrapper have a NOTICE file?

Also there is a compiled code (a jar) in the source release. Source release cannot include compiled code.

Kind Regards,
Justin

1. https://www.apache.org/dyn/closer.cgi/skywalking/9.1.0/apache-skywalking-apm-9.1.0-src.tgz <https://www.apache.org/dyn/closer.cgi/skywalking/9.1.0/apache-skywalking-apm-9.1.0-src.tgz>

```

We should remove the jar file from the source tar.

### How to reproduce

Just repackage the v9.1.0 tag for a source release.

### Anything else

_No response_

### Code of Conduct

CalvinKirs:
I checked: https://github.com/apache/skywalking/blob/77ba2d553d77a4acacd8b2498cb09ea983a9baf4/LICENSE#L215-L225
Only `maven-wrapper` and `Prometheus` need to add notice

this closed #9512

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -217,6 +217,7 @@ The text of each license is the standard Apache 2.0 license.
 
    proto files from cncf/udpa: https://github.com/cncf/udpa Apache 2.0
    proto files from envoyproxy/data-plane-api: https://github.com/envoyproxy/data-plane-api  Apache 2.0
+   proto filrs from opencensus: https://github.com/census-instrumentation/opencensus-proto/tree/master/gen-go
    proto files from prometheus/client_model: https://github.com/prometheus/client_model Apache 2.0
    proto files from opentelemetry: https://github.com/open-telemetry/opentelemetry-proto/tree/main/opentelemetry/proto Apache 2.0
    flatbuffers files from istio/proxy: https://github.com/istio/proxy Apache 2.0
--- a/LICENSE
+++ b/LICENSE
@@ -217,7 +217,6 @@ The text of each license is the standard Apache 2.0 license.
 
    proto files from cncf/udpa: https://github.com/cncf/udpa Apache 2.0
    proto files from envoyproxy/data-plane-api: https://github.com/envoyproxy/data-plane-api  Apache 2.0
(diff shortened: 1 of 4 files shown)


wu-sheng:
I think OpenCensus was being asked too. Could you recheck that?

CalvinKirs:
> I think OpenCensus was being asked too. Could you recheck that?

Hi, I checked, PTAL, thx

CalvinKirs:
But about this, `oap-server/server-receiver-plugin/receiver-proto/src/main/proto/satellite/envoy/accesslog/v3/als.proto` I can't be sure, could you help me take a look?

wu-sheng:
als.proto is from envoy project.

CalvinKirs:
> als.proto is from envoy project.

Thanks, done

LICENSE
```diff
@@ -217,6 +217,8 @@ The text of each license is the standard Apache 2.0 license.
 
    proto files from cncf/udpa: https://github.com/cncf/udpa Apache 2.0
    proto files from envoyproxy/data-plane-api: https://github.com/envoyproxy/data-plane-api  Apache 2.0
+   proto files from satellite/envoy/: https://github.com/envoyproxy/envoy  Apache 2.0
```
wu-sheng, line 220: I think this is skywalking's als fork? @mrproliu 
If so, this is not from envoy. And I can see we have envoy data panel proto licences already.

wu-sheng: I checked, this is a shell of envoy ALS, and it belongs to SkyWalking.

wu-sheng:
```suggestion
```

PR #9529, opened 2022-08-26
Bump up infra e2e version and fix istio cases

kezhenxu94:
Also serve as an example that cases use different Kubernetes versions


wu-sheng (approved):
Once e2e passed.

wu-sheng:
This seems not working.

wu-sheng (dismissed):

wu-sheng:
Strange, TLS case fails 3 times(after I retried).

kezhenxu94:
> Strange, TLS case fails 3 times(after I retried).

The mTLS cert is expired... I have extended the expiration date to 100 years

wu-sheng (approved):

Issue #9487, opened 2022-08-19
[Feature] Support for NATS Java client Plugin


pg-yang:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

We have supported Micronaut . And I discover Micronaut support [NATS](https://docs.nats.io/) as message system . 
So , I want to try implementing [NATS Java Client ](https://github.com/nats-io/nats.java)agent plugin   . And I will put the plugin to `optional-plugins` module.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wu-sheng:
> I will put the plugin to optional-plugins module.

Could you explain why this should be in `optional`? Is that able to hurt the application somehow? Or trigger some conflicts?

pg-yang:
> > I will put the plugin to optional-plugins module.
> 
> Could you explain why this should be in `optional`? Is that able to hurt the application somehow? Or trigger some conflicts?

I think it's not commonly used .Your suggestion ?

wu-sheng:
If this is only targeting NATS client's classes(such as under `io.nats.client.*` only), the agent would not make it work automatically.
About the optional plugins, https://skywalking.apache.org/docs/skywalking-java/latest/en/setup/service-agent/java-agent/optional-plugins/, you could notice all plugins have their own reasons to be optional only, as they may have a side-effect.

pg-yang:
Got it . I will choose appropriate  module

PR #9533, opened 2022-08-27
Add component ID(132) for NAST java client plugin

pg-yang:

Related issue : https://github.com/apache/skywalking/issues/9487

image

* nats-8080 is springboot application , include nats-java library
* nats://127.0.0.1:4222 is nats server address

I changed `changes.md` , because the original comment from sync ui . It's not necessary，I think.

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -62,6 +62,9 @@
 * [Breaking Change] Fix JDBC TTL to delete additional tables data. 
   SQL Database requires removing `segment`,`segment_tag`, `logs`, `logs_tag`, `alarms`, `alarms_tag`, `zipkin_span`, `zipkin_query` before OAP starts.
 * SQL Database: add `@SQLDatabase.ExtraColumn4AdditionalEntity` to support add an extra column from parent to an additional table.
+* Add component ID(131) for Java Micronaut plugin
+* Add component ID(132) for Nats java client plugin
+
 
 #### UI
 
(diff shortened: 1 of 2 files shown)


docs/en/changes/changes.md
```diff
@@ -102,7 +102,7 @@
 * Fix picking calendar with a wrong time range and setting a unique value for dashboard grid key.
 * Add PostgreSQL to Database sub-menu.
 * Implement the network profiling widget.
-* Add Nats icon for Java plugin.
```
wu-sheng, line 105: This should be kept, as UI is released in main repo.

docs/en/changes/changes.md
```diff
@@ -102,7 +102,7 @@
...
+* Add component ID(132) for NAST java client plugin
```
wu-sheng, line 105: Move to backend part of changelog.

wu-sheng (changes requested):

docs/en/changes/changes.md
pg-yang: `Add component ID(131) for Java Micronaut plugin` is also in UI part . 
The previous format of  micronaut plugin was also wrong , I will fix it  together  .
Finally , add mictonaut、nats icon should be in UI part , add component id for micronaut、nats in backend part

wu-sheng: Yes, icon is UI part. ID is backend part.

wu-sheng (approved):

PR #9535, opened 2022-08-29
Add eBPF CPU Profiling documentation

mrproliu:

### DIFF
--- a/docs/menu.yml
+++ b/docs/menu.yml
@@ -38,6 +38,8 @@ catalog:
             path: "/en/papers/stam"
           - name: "SDK Profiling to Fix the Blind Spot"
             path: "/en/concepts-and-designs/sdk-profiling"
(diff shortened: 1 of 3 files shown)


docs/menu.yml
```diff
@@ -38,6 +38,8 @@ catalog:
             path: "/en/papers/stam"
           - name: "SDK Profiling to Fix the Blind Spot"
             path: "/en/concepts-and-designs/sdk-profiling"
+          - name: "EBPF CPU Profiling to pinpoint Service Mesh Critical Performance Impact"
```
wu-sheng, line 41:
```suggestion
          - name: "eBPF CPU Profiling to pinpoint Service Mesh Critical Performance Impact"
```

Is lower case e better?

wu-sheng (approved):

PR #9551, opened 2022-09-02
Fix the wrong watch semantics in Kubernetes watchers

kezhenxu94:

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -24,6 +24,9 @@
   downsampling.
 * Do sorting `readLabeledMetricsValues` result forcedly in case the storage(database) doesn't return data consistent
   with the parameter list.
+* Fix the wrong watch semantics in Kubernetes watchers, which causes heavy traffic to API server in some Kubernetes clusters,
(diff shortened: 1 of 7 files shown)


kezhenxu94:
Adding `don't merge` until I verified this fixes the cluster in problem (it's not easy to reproduce 😢 )

docs/en/changes/changes.md
```diff
@@ -5,7 +5,9 @@
 
 #### OAP Server
 
-
+* Fix the wrong watch semantics in Kubernetes watchers, which causes heavy traffic to API server in some Kubernetes cluster,
+  we should use `Get State and Start at Most Recent` semantic instead of `Start at Exact`
+  because we don't need the changing history events, see https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-watch.
```
wu-sheng, line 10:
```suggestion
  because we don't need the changing history events, see https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-watch.

```

kezhenxu94:
The PR is good to go now.

wu-sheng (approved):

Issue #9587, opened 2022-09-09
Elasticsearch8x: no handler found for uri [<index>/_doc/<document_id>/_update] and method [POST][Bug] 


dingzx3:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar issues.

### Apache SkyWalking Component

OAP server (apache/skywalking)

### What happened

![image](https://user-images.githubusercontent.com/111035198/189258875-1612300d-99d1-4e4a-a4e0-ead7fe41c4de.png)
![image](https://user-images.githubusercontent.com/111035198/189259102-97a2c5e8-ce68-47bf-ba3a-d8c9d622fee6.png)

Error when saving tab item

### What you expected to happen

Support Elasticsearch8x update api

### How to reproduce

The api used in V7DocumentFactory is deprecated and removed in 8.1. 
7.x and 8.x supports the following api
![image](https://user-images.githubusercontent.com/111035198/189260147-b4075ffd-9e33-43db-be50-1e32c3f0b4e7.png)

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #9589, opened 2022-09-09
fix Elasticsearch8x: no handler found for uri [<index>/_doc/<document…

dingzx3:
…_id>/_update] and method [POST][Bug] (#9587)

### DIFF
--- a/oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/factory/v7/V78RequestFactory.java
+++ b/oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/factory/v7/V78RequestFactory.java
@@ -45,7 +45,7 @@ public final class V78RequestFactory implements RequestFactory {
         template = new V78TemplateFactory(version);
         index = new V7IndexFactory(version);
         alias = new CommonAliasFactory(version);
-        document = new V7DocumentFactory(version);
+        document = new V8DocumentFactory(version);
         search = new CommonSearchFactory(version);
         bulk = new CommonBulkFactory(version);
     }
--- a/oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/factory/v7/V7DocumentFactory.java
+++ b/oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/factory/v7/V7DocumentFactory.java
@@ -164,7 +164,7 @@ final class V7DocumentFactory implements DocumentFactory {
         }
         final byte[] content = version.codec().encode(ImmutableMap.of("doc", doc));
 
-        builder.post("/{index}/_doc/{id}/_update")
+        builder.post("/{index}/_update/{id}")
                .pathParam("index", index)
                .pathParam("id", id)
                .content(MediaType.JSON, content);
--- a/oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/factory/v7/V8DocumentFactory.java
+++ b/oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/factory/v7/V8DocumentFactory.java
@@ -164,7 +164,7 @@ final class V7DocumentFactory implements DocumentFactory {
         }
         final byte[] content = version.codec().encode(ImmutableMap.of("doc", doc));
 
-        builder.post("/{index}/_doc/{id}/_update")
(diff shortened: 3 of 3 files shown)


oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/factory/v7/V7DocumentFactory.java
```diff
@@ -164,7 +164,7 @@ public HttpRequest update(UpdateRequest request, Map<String, ?> params) {
         }
         final byte[] content = version.codec().encode(ImmutableMap.of("doc", doc));
 
-        builder.post("/{index}/_doc/{id}/_update")
+        builder.post("/{index}/_update/{id}")
```
wu-sheng, line 167: This is a v7 factory, why change this? You mentioned v8.1.

kezhenxu94: > This is a v7 factory, why change this? You mentioned v8.1.

I think the related issue mentioned that both 7.x and 8.x support this, the previous one is deprecated, right @dingzx3 ? I'll start the CI to see

wu-sheng: Let's confirm from ElasticSearch API docs as well.

dingzx3: The last one supported since 7.3，the previous one is  deprecated and removed in 8.1.

wu-sheng: > The last one supported since 7.3

Which one? I think we need to support the whole 7.x ElasticSearch?

kezhenxu94: > The last one supported since 7.3，the previous one is  deprecated and removed in 8.1.

Can you please add a new codec for 8.x and keep the original one as is?

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/factory/v7/V78RequestFactory.java
```diff
@@ -45,7 +45,7 @@ public V78RequestFactory(final ElasticSearchVersion version) {
         template = new V78TemplateFactory(version);
         index = new V7IndexFactory(version);
         alias = new CommonAliasFactory(version);
-        document = new V7DocumentFactory(version);
+        document = new V8DocumentFactory(version);
```
wu-sheng, line 48: Is this a correct change?

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/factory/v7/V8DocumentFactory.java
```diff
@@ -0,0 +1,195 @@
...
+import org.apache.skywalking.library.elasticsearch.requests.search.Query;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Strings.isNullOrEmpty;
+import static com.google.common.collect.Iterables.isEmpty;
+import static java.util.Objects.requireNonNull;
+
+@Slf4j
+@RequiredArgsConstructor
+final class V7DocumentFactory implements DocumentFactory {
```
wu-sheng, line 44: The file name is v8, the class is v7?

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/factory/v7/V78RequestFactory.java
wu-sheng: I think @zhenxu mean we should identify v8 ElasticSearch, and use a new v8 factory.

wu-sheng:
Also, we should update IT(ElasticSearch version matrix) to verify the 8.1 case after new v8 factory added.

wu-sheng:
Any update?

kezhenxu94:
Hey @dingzx3 , I hope we could fix this in 9.3.0, if you need any help, please ping us.

PR #9592, opened 2022-09-10
webapp: migrate from Spring Web to Armeria

kezhenxu94:
Webapp module is a very simple web service serves static files and a server proxy to OAP, it contributes 40+ dependencies due to the Spring ecosystem, this PR replace the Spring web container with Armeria, which we've been already using, and it reduce our maintaining efforts, such as CVEs, licensing issues, etc.


wu-sheng:
@wankai123 See, I felt this would happen by someone :P Those dependencies and CVEs are really annoying.

docs/en/setup/backend/ui-setup.md
```diff
@@ -46,4 +31,4 @@ We could set up environment variables to configure this image.
 
 ### SW_OAP_ADDRESS
 
-The address of your OAP server. The default value is `http://127.0.0.1:12800`.
\ No newline at end of file
+The address of your OAP server. The default value is `127.0.0.1:12800`.
```
wu-sheng, line 34: Some ppl would be surprised we don't need `http://` again. Because we just changed from this to `http://xx` when move from zuul to Spring Gateway last year?

JaredTan95 (approved):
It's great to see this change and more lighter UI.

docs/en/setup/backend/ui-setup.md
wu-sheng: The things like docs on docker hub, we can't versioning it, https://hub.docker.com/r/apache/skywalking-ui.

Is there any chance we could still keep `http://ip:port` for the Armeria web container?

docs/en/changes/changes.md
```diff
@@ -12,9 +12,12 @@
 
 #### UI
 
-* Fix: tab active incorrectly, when click tab space 
+* Fix: tab active incorrectly, when click tab space
 * Add impala icon for impala JDBC Java agent plugin.
 * (Webapp)Bump up snakeyaml to 1.31 for fixing CVE-2022-25857
+* [Breaking Change]: migrate from Spring Web to Armeria, now you should use the environment variable name `SW_OAP_ADDRESS`
```
wu-sheng, line 18: How about we use codes to check and remove http:// prefix if exist?

This breaking change seems not necessary.

kezhenxu94: This is how I did currently. This line didn't mention the http prefix.

wu-sheng:
BTW deadlinks due to we change latest to next for the dev version.
It has redirect, but deadlink tool still reporta it.

wu-sheng (dismissed):

kezhenxu94:
> BTW deadlinks due to we change latest to next for the dev version.
> It has redirect, but deadlink tool still reporta it.

The change doesn't look a correct redirect, from what I observed, it's displaying a custom 404 page and then reload the page by modifying the URL location, this is not a canonical way to redirect. If  you `curl` the link with flag to follow the redirect `-l`, it doesn't find redirect headers.

wu-sheng:
One question, I can't find manual http:// removing, is it processed automatically by Armeria?

apm-webapp/src/main/java/org/apache/skywalking/oap/server/webapp/OapProxyService.java
```diff
@@ -0,0 +1,79 @@
...
+
+public final class OapProxyService extends AbstractHttpService {
+    private final WebClient loadBalancingClient;
+
+    public OapProxyService(String[] oapServices) throws Exception {
+        final List<Endpoint> endpoints =
+            Stream
+                .of(oapServices)
+                .map(URI::create)
+                .map(URI::getAuthority)
```
kezhenxu94, line 48: > One question, I can't find manual http:// removing, is it processed automatically by Armeria?

`URI::create` and `URI::getAuthority` removes the http scheme, `http://httpbin.org:80` will become `httpbin.org:80`

wu-sheng: Oh  I see.

wu-sheng (dismissed):

wu-sheng (approved):

PR #9593, opened 2022-09-10
Add layer property to LAL script

yswdqz:


oap-server/server-starter/src/main/resources/lal/default.yaml
```diff
@@ -16,6 +16,7 @@
 # The default LAL script to save all logs, behaving like the versions before 8.5.0.
 rules:
   - name: default
+    layer: UNDEFINED
```
wu-sheng, line 19: This should be general service layer

docs/en/concepts-and-designs/lal.md
```diff
@@ -7,6 +7,9 @@ segment ID and span ID) and metrics (by generating metrics from the logs and sen
 The LAL config files are in YAML format, and are located under directory `lal`. You can
 set `log-analyzer/default/lalFiles` in the `application.yml` file or set environment variable `SW_LOG_LAL_FILES` to
 activate specific LAL config files.
+## Layer
+Every LAL file is a list of LAL rule, and every LAL rule has a layer property.If a log sent to OAP has the same layer value, this rule will be enable.
+NOTICE: Every layer should only has one rule.
```
wu-sheng, line 12:
```suggestion

## Layer
Layer should be declared in the LAL script to represent the analysis scope of the logs.
```

wu-sheng: Notice the way of describing things in docs. It is not about the code logic, instead it should decribe how end users would use it.

oap-server/server-starter/src/main/resources/lal/default.yaml
wu-sheng: I feel we may not need a default LAL file. You could change this LAL for general service only.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
```diff
@@ -19,55 +19,65 @@
...
-                dsl.evaluate();
-            } catch (final Exception e) {
-                log.warn("Failed to evaluate dsl: {}", dsl, e);
+        DSL dsl = dsls.get(logData.getLayer());
+        try {
+            if (dsl == null) {
+                if (StringUtil.isEmpty(logData.getLayer())) {
+                    log.debug("The layer is empty, will use default rules");
+                }
+                dsl = dsls.get(Layer.UNDEFINED.name());
```
wu-sheng, line 52: I think if we can't find suitable layer, we could simply skip the log process.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/LogAnalyzerModuleConfig.java
```diff
@@ -51,7 +51,7 @@ public class LogAnalyzerModuleConfig extends ModuleConfig {
     private List<Rule> meterConfigs;
 
     public List<String> lalFiles() {
-        return Splitter.on(",").omitEmptyStrings().splitToList(Strings.nullToEmpty(getLalFiles()));
+        return Splitter.on(",").omitEmptyStrings().trimResults().splitToList(Strings.nullToEmpty(getLalFiles()));
```
wu-sheng, line 54: Pleasw verify whether layer is legal and not empty.

wu-sheng (commented):
e2e seems broken, please fix.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
```diff
@@ -19,55 +19,65 @@
...
-        dsls.forEach(dsl -> dsl.bind(new Binding().log(logData.build())
+        dsls.forEach((layer, dsl) -> dsl.bind(new Binding().log(logData.build())
                                                   .extraLog(extraLog)));
+        this.logData = logData;
         return this;
     }
 
     public static class Factory implements LogAnalysisListenerFactory {
-        private final List<DSL> dsls;
+        private final Map<String, DSL> dsls;
```
wu-sheng, line 70: I checked how this Factory is used, I feel you don't need this Map in the LogFilterListener. You could use `layer` as a parameter in `#create` method to build the suitable listener directly.

yswdqz: I have thought about it. But I think if I add a param to Factory#create method will affect other two listeners(RecordAnalysisListener and TrafficAnalysisListener).
Is that OK? They seems don't need layer param.

yswdqz:
The deadlink check always fail and seems not cause by me.

wu-sheng:
Don't worry about that. You just need to focus your changed parts.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
```diff
@@ -19,61 +19,76 @@
...
 import org.apache.skywalking.oap.log.analyzer.provider.LogAnalyzerModuleConfig;
+import org.apache.skywalking.oap.server.core.UnexpectedException;
+import org.apache.skywalking.oap.server.core.analysis.Layer;
 import org.apache.skywalking.oap.server.library.module.ModuleManager;
 
 @Slf4j
 @RequiredArgsConstructor
 public class LogFilterListener implements LogAnalysisListener {
-    private final List<DSL> dsls;
+    private final DSL dsl;
```
kezhenxu94, line 41: `dsl` should be non null, otherwise you should not create this listener.

```suggestion
    @lombok.NonNull
    private final DSL dsl;
```

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
```diff
@@ -19,61 +19,76 @@
 package org.apache.skywalking.oap.log.analyzer.provider.log.listener;
 
 import com.google.protobuf.Message;
-import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.stream.Collectors;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.collections4.map.HashedMap;
```
kezhenxu94, line 27: Prefer `java.util.HashMap`

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
```diff
@@ -19,61 +19,76 @@
...
                                                          .stream()
                                                          .flatMap(it -> it.getRules().stream())
                                                          .collect(Collectors.toList());
             for (final LALConfig c : configList) {
-                dsls.add(DSL.of(moduleManager, config, c.getDsl()));
+                try {
+                    Layer layer = Layer.nameOf(c.getLayer());
+                    dsls.put(layer, DSL.of(moduleManager, config, c.getDsl()));
+                } catch (UnexpectedException e) {
+                    log.warn("layer not found, will ignore this rule:" + c.getName());
```
kezhenxu94, line 81:
```suggestion
                    log.warn("layer {} not found, will ignore this rule.", c.getName());
```

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
```diff
@@ -19,61 +19,76 @@
...
         }
 
         @Override
-        public LogAnalysisListener create() {
-            return new LogFilterListener(dsls);
+        public LogAnalysisListener create(Layer layer) {
+            if (layer == null) {
+                return null;
+            }
+            return new LogFilterListener(dsls.get(layer));
```
kezhenxu94, line 91:
```suggestion
            final DSL dsl = dsls.get(layer);
            if (dsl == null) {
                return null;
            }
            return new LogFilterListener(dsl);
```

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
```diff
@@ -19,61 +19,76 @@
...
+                }
             }
         }
 
         @Override
-        public LogAnalysisListener create() {
-            return new LogFilterListener(dsls);
+        public LogAnalysisListener create(Layer layer) {
+            if (layer == null) {
+                return null;
```
kezhenxu94, line 89: Check the places where this method `create` is invoked, if you return null here, `NullPointerException` might be thrown, you have to check the nullability in the callers.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
```diff
@@ -19,61 +19,76 @@
...
 
             final List<LALConfig> configList = LALConfigs.load(config.getLalPath(), config.lalFiles())
                                                          .stream()
                                                          .flatMap(it -> it.getRules().stream())
                                                          .collect(Collectors.toList());
             for (final LALConfig c : configList) {
-                dsls.add(DSL.of(moduleManager, config, c.getDsl()));
+                try {
+                    Layer layer = Layer.nameOf(c.getLayer());
+                    dsls.put(layer, DSL.of(moduleManager, config, c.getDsl()));
```
kezhenxu94, line 79: Now there is a limitation that only one LAL script can be configured for a layer, you might need to check here and throw exception if multiple DSLs are configured for a same layer

kezhenxu94 (changes requested):

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
yswdqz: I have checked it. No other invoker use the return value.

kezhenxu94: > I have checked it. No other invoker use the return value.

Hi, `listeners` is an `ArrayList`, it doesn't allow null, so if `create` returns `null`, `NullPointerException` will be thrown in the following line

https://github.com/apache/skywalking/blob/81878331e59c435e011eb8c73fe5617201708968/oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/LogAnalyzer.java#L79-L80

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
```diff
@@ -19,61 +19,86 @@
...
+                Layer layer = null;
+                try {
+                    layer = Layer.nameOf(c.getLayer());
+                    if (dsls.containsKey(layer)) {
+                        log.warn("layer {} has set a rule, old one will be ignore.", c.getName());
+                    }
+                    dsls.put(layer, DSL.of(moduleManager, config, c.getDsl()));
+                } catch (UnexpectedException e) {
+                    log.warn("layer {} not found, will ignore this rule.", c.getName());
+                }
```
kezhenxu94, line 88:
```suggestion
                try {
                    Layer layer = Layer.nameOf(c.getLayer());
                    if (dsls.put(layer, DSL.of(moduleManager, config, c.getDsl())) != null) {
                        log.warn("layer {} has already set a rule {}, the old one will be ignored.", layer, c.getName());
                    }
                } catch (UnexpectedException e) {
                    log.warn("layer {} not found, will ignore this rule.", c.getName(), e);
                }
```

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
```diff
@@ -19,61 +19,86 @@
...
-        dsls.forEach(dsl -> {
-            try {
-                dsl.evaluate();
-            } catch (final Exception e) {
-                log.warn("Failed to evaluate dsl: {}", dsl, e);
+        try {
+            if (dsl == null) {
+                log.warn("No rules match, the process will skip.");
+                return;
             }
```
kezhenxu94, line 51: `dsl` is already non null.

```suggestion
```

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
```diff
@@ -19,61 +19,86 @@
...
     }
 
     @Override
     public LogAnalysisListener parse(final LogData.Builder logData,
                                      final Message extraLog) {
-        dsls.forEach(dsl -> dsl.bind(new Binding().log(logData.build())
-                                                  .extraLog(extraLog)));
+        if (dsl == null) {
+            return null;
+        }
```
kezhenxu94, line 63: `dsl` is non null.
```suggestion
```

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
yswdqz: @kezhenxu94 I have tested a null value situation. It will not throw a Exception.
An ArrayList can store more than one null.

wu-sheng:
Please check why log e2e are all failure.

yswdqz:
> Please check why log e2e are all failure.

I get it.

yswdqz:
Maybe because of the layer check I have set in this PR.
```
try {
            layer = Layer.nameOf(builder.getLayer());
        } catch (UnexpectedException e) {
            log.warn("layer not found, will skip this process.");
            return;
        }
```
I think that I can set `GENERAL`as default layer.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/filter/FilterSpec.java
```diff
@@ -169,7 +169,7 @@ public void sink(@DelegatesTo(SinkSpec.class) final Closure<?> cl) {
         final Optional<AtomicReference<Log>> container = BINDING.get().logContainer();
         if (container.isPresent()) {
             factories.stream()
-                     .map(LogAnalysisListenerFactory::create)
+                     .map(it -> it.create(null))
```
wu-sheng, line 172: I am wondering, why we use `null` here?

yswdqz: These two listeners needn't layer param.

yswdqz: This listener needn't layer param.

wu-sheng: @kezhenxu94 Should we separate `LogFilterListener.Factory` and `TrafficAnalysisListener.Factory`/`RecordAnalysisListener.Factory`?

`TrafficAnalysisListener.Factory`/`RecordAnalysisListener.Factory` are more likely `LogSinkListener`.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
kezhenxu94: > @kezhenxu94 I have tested a null value situation. It will not throw a Exception.
> An ArrayList can store more than one null.

Hey, hosting a `null` value in the list is not a good idea. What's more, did you test the case where the logs will be persisted? The following lines when iterating the array list will throw exception. `listeners.forEach(LogAnalysisListener::build);` Element in `listeners` might be null and calling `null.build()` should throw.

https://github.com/apache/skywalking/blob/166a62612d20e403241aec24b95ceef1c64d898e/oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/LogAnalyzer.java#L73-L81

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/LogAnalyzer.java
```diff
@@ -64,8 +79,8 @@ private void notifyListenerToBuild() {
         listeners.forEach(LogAnalysisListener::build);
     }
 
-    private void createListeners() {
+    private void createListeners(Layer layer) {
         factoryManager.getLogAnalysisListenerFactories()
-                      .forEach(factory -> listeners.add(factory.create()));
+                      .forEach(factory -> listeners.add(factory.create(layer)));
```
kezhenxu94, line 84: Maybe change to 

```java
        factoryManager.getLogAnalysisListenerFactories()
                      .stream()
                      .map(factory -> factory.create(layer))
                      .filter(Objects::nonNull)
                      .forEach(listeners::add);
```

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/filter/FilterSpec.java
kezhenxu94: > @kezhenxu94 Should we separate `LogFilterListener.Factory` and `TrafficAnalysisListener.Factory`/`RecordAnalysisListener.Factory`?
> 
> `TrafficAnalysisListener.Factory`/`RecordAnalysisListener.Factory` are more likely `LogSinkListener`.

Hi, I'm not sure, these factories share the same interface and they are used basically in the core log analyzer, so splitting these might bring huge changes.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
yswdqz: I understand , it's my fault. I will change it soon.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/filter/FilterSpec.java
wu-sheng: The issue is what I proposed. The create  method is adding layer, but those two for sink are not relative. 

@yswdqz put a null layer for create method. I feel we should change it.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/LogAnalyzer.java
```diff
@@ -46,7 +50,20 @@ public void doAnalysis(LogData.Builder builder, Message extraLog) {
...
-        createListeners();
+        Layer layer;
+        if ("".equals(builder.getLayer())) {
+            layer = Layer.GENERAL;
+        }
+        else {
+            try {
+                layer = Layer.nameOf(builder.getLayer());
+            } catch (UnexpectedException e) {
+                log.warn("layer not found, will skip this process.");
```
wu-sheng, line 61:
```suggestion
                log.warn(The Layer {} is not found, abandon the log.", builder.getLayer());
```

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
```diff
@@ -19,61 +19,77 @@
...
                                                          .stream()
                                                          .flatMap(it -> it.getRules().stream())
                                                          .collect(Collectors.toList());
             for (final LALConfig c : configList) {
-                dsls.add(DSL.of(moduleManager, config, c.getDsl()));
+                try {
+                    Layer layer = Layer.nameOf(c.getLayer());
+                    if (dsls.put(layer, DSL.of(moduleManager, config, c.getDsl())) != null) {
+                        log.warn("layer {} has already set a rule {}, the old one will be ignored.", layer, c.getName());
+                    }
```
wu-sheng, line 76: Layer conflicts should cause booting failure. Please throw the exception with clear information about which rules(files) are using the same layer.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
```diff
@@ -19,61 +19,77 @@
...
+                } catch (UnexpectedException e) {
+                    log.warn("layer {} not found, will ignore this rule.", c.getName(), e);
```
wu-sheng, line 78: Layer should break the booting process too.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
yswdqz: DSL has not stored the name of rule. So it is difficult to log which rules(files) are using the same layer.

kezhenxu94: > DSL has not stored the name of rule. So it is difficult to log which rules(files) are using the same layer.

Simply log the layer should be enough IMO

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/LogFilterListener.java
```diff
@@ -19,61 +19,73 @@
...
 
             final List<LALConfig> configList = LALConfigs.load(config.getLalPath(), config.lalFiles())
                                                          .stream()
                                                          .flatMap(it -> it.getRules().stream())
                                                          .collect(Collectors.toList());
             for (final LALConfig c : configList) {
-                dsls.add(DSL.of(moduleManager, config, c.getDsl()));
+                Layer layer = Layer.nameOf(c.getLayer());
+                if (dsls.put(layer, DSL.of(moduleManager, config, c.getDsl())) != null) {
+                    throw new UnexpectedException("Layer "+ layer.name() +" has already set a rule.");
```
kezhenxu94, line 74:
```suggestion
                    throw new ModuleStartException("Layer " + layer.name() + " has already set a rule.");
```

wu-sheng:
I think in today's e2e log cases don't report layer as it was not required. It is better if we could add default=general layer to the log receiver before it is forwarded into LAL engine to keep forward compatibility.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/LogAnalyzer.java
```diff
@@ -46,7 +50,20 @@ public void doAnalysis(LogData.Builder builder, Message extraLog) {
...
+            layer = Layer.GENERAL;
+        }
+        else {
+            try {
+                layer = Layer.nameOf(builder.getLayer());
+            } catch (UnexpectedException e) {
+                log.warn("The Layer {} is not found, abandon the log.", builder.getLayer());
+                return;
+            }
+        }
```
wu-sheng, line 64:
```suggestion
        Layer layer;
        if ("".equals(builder.getLayer())) {
            layer = Layer.GENERAL;
        }  else {
            try {
                layer = Layer.nameOf(builder.getLayer());
            } catch (UnexpectedException e) {
                log.warn("The Layer {} is not found, abandon the log.", builder.getLayer());
                return;
            }
        }
```

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/filter/FilterSpec.java
wu-sheng: @yswdqz Generally tests passed. Please follow this to separate `LogSinkListener` out of `LogAnalysisListener` implementations. And `LogSinkListener#Factory#create` doesn't require layer as it is generally usable.

kezhenxu94: > @yswdqz Generally tests passed. Please follow this to separate `LogSinkListener` out of `LogAnalysisListener` implementations. And `LogSinkListener#Factory#create` doesn't require layer as it is generally usable.

Hi, I’m confused. Currently the reason why there is a `null` in the `create` method is that the 3 factories share the same interface and the interface is used in the core module of log analysis, if we could separate the sink factory we could just remove the null parameter from the create method. Adding a new sink listener factory doesn’t help from what I can tell. Or I misunderstood your request here.

wu-sheng: @kezhenxu94 The reason is `LogFilterListener`, it is required to have a `layer` as parameter to create `listener`, which we changed in this PR.

But once we add them, we have to change all others due to this hierarchy tree.

image

The only reason I asked for this, is to make `create(null)` disappear.

wu-sheng: > Currently the reason why there is a null in the create method is that the 3 factories share the same interface and the interface is used in the core module of log analysis, if we could separate the sink factory we could just remove the null parameter from the create method.

I feel you are asking the same thing. 3 implementations for one interface. Add one for <2> and <3>, and add one for <1>? I just don't think they are doing the same things, as one for core, the other two for sink. And `null` as a `create` method's parameter seems not right.
I didn't insist on anything relative to which interface and the names. If you want any preference, feel free to propose.

kezhenxu94: > @kezhenxu94 The reason is `LogFilterListener`, it is required to have a `layer` as parameter to create `listener`, which we changed in this PR.
> 
> But once we add them, we have to change all others due to this hierarchy tree.
> 
> image
> 
> The only reason I asked for this, is to make `create(null)` disappear.

That’s my point. How to separate the 3 factories?  Add a new sink factory and the add the sink factory here?

https://github.com/apache/skywalking/blob/5ba874a49946571506b6cb3c1abbf0d4ce1863bf/oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/ILogAnalysisListenerManager.java#L23

```yaml
void addSinkListenerFactory(LogSinkListenerFactory factory);
```

wu-sheng: TrafficAnalysisListener#Factory and RecordAnalysisListener#Factory implement from a new interface, that is all.

wu-sheng: LogAnalysisListenerFactory interface is changed for core to add a new layer in create method.

wu-sheng: > > @kezhenxu94 The reason is `LogFilterListener`, it is required to have a `layer` as parameter to create `listener`, which we changed in this PR.
> 
> > 
> 
> > But once we add them, we have to change all others due to this hierarchy tree.
> 
> > 
> 
> > image
> 
> > 
> 
> > The only reason I asked for this, is to make `create(null)` disappear.
> 
> 
> 
> That’s my point. How to separate the 3 factories?  Add a new sink factory and then add the sink factory here?
> 
> 
> 
> https://github.com/apache/skywalking/blob/5ba874a49946571506b6cb3c1abbf0d4ce1863bf/oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/ILogAnalysisListenerManager.java#L23
> 
> 
> 
> ```yaml
> 
> void addSinkListenerFactory(LogSinkListenerFactory factory);
> 
> ```
> 
> 
> 
> 

I am not sure whether we need. I remember those two factories are hard coded created.

yswdqz:
LogSinkListener.Factory's constructor 's LogAnalyzerModuleConfig param maybe also need to remove.

https://github.com/yswdqz/skywalking/blob/lalchange/oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/RecordSinkListener.java#L151

https://github.com/yswdqz/skywalking/blob/lalchange/oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/TrafficSinkListener.java#L104

wu-sheng (approved):

kezhenxu94 (approved):

PR #9610, opened 2022-09-15
UI-template: Fix metrics calculation in `general-service/mesh-service/faas-function` top-list dashboard.

wankai123:
- [X] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #9607.

### DIFF
--- a/oap-server/server-starter/src/main/resources/ui-initialized-templates/faas/faas-function.json
+++ b/oap-server/server-starter/src/main/resources/ui-initialized-templates/faas/faas-function.json
@@ -323,12 +323,6 @@
                   "metricTypes": [
                     "sortMetrics"
                   ],
-                  "metricConfig": [
-                    {
-                      "sortOrder": "DES",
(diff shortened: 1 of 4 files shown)


wu-sheng (approved):

oap-server/server-starter/src/main/resources/ui-initialized-templates/faas/faas-function.json
```diff
@@ -323,12 +323,6 @@
                   "metricTypes": [
                     "sortMetrics"
                   ],
-                  "metricConfig": [
-                    {
-                      "sortOrder": "DES",
```
wu-sheng, line 328: Should we keep DES? As this is a metric of traffic? More traffic more important?

wankai123: `DES` is the default order

Issue #9556, opened 2022-09-03
[Feature] Add metric for analyzing  cache-related  span like Virtual Database


pg-yang:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Virtual  Cache will analysis  span  which  is  cache-related . It will generate SLA, response time or some else metric
In a word  , it's   similar to Virtual Database
Reference [virtual-database](https://skywalking.apache.org/docs/main/latest/en/setup/service-agent/virtual-database/)  [skywalking demo ](http://demo.skywalking.apache.org/dashboard/VIRTUAL_DATABASE/Service/bG9jYWxob3N0Oi0x.0/Virtual-Database-Service)

Works should  to  be  done :

#### Agent (focus on Java agent) 
- [x]   Update  cache plugins for  uniform `Span’s layer ` , `tag `  if  necessary .  
    The known cache-related plugin(need suggestion : which plugin is cache-related):
    1. jedis-plugins
    2. redisson-plugin
    3. xmemcached-plugin
    4. zookeeper-plugin
    5. spymemcached-plugin

#### Backend 
- [x]  develop  Virtual  Cache metric
    1. Add  RequestType : Cache 
    1. ServiceMeta for registering  cache  service 
    9.  Cache-source for OAL
    10. Slow query Top N metric 
- [x]  Update [core oal ]( https://github.com/apache/skywalking/blob/master/oap-server/server-starter/src/main/resources/oal/core.oal) for metric (reponse time  or  some  else )
- [x] Update  Doc  for  Virtual  Cache 

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

pg-yang:
Maybe , we should expand  the meaning  of  `Database ` . `Virtual Database`  should  contains  cache  instead of creating `Virtual Cache`

wu-sheng:
I think they are different, basically. Cache is much faster than database.

pg-yang:
They  are used for saving  data , and contain statement  too .
We provide slow threshold config for various DB
```
agent-analyzer:
  selector: ${SW_AGENT_ANALYZER:default}
  default:
    slowDBAccessThreshold: ${SW_SLOW_DB_THRESHOLD:default:200,mongodb:100,redis:20} 
```
[...]

Issue #258, opened 2017-07-01
Failed to access Injectable class


ascrutae:
`org.skywalking.apm.plugin.dubbo.DubboInterceptor$1` is anonymous inner class for `Injectable`. Is that byte-buddy support access anonymous inner class in the interceptor method ?
Here is the error log
```
java.lang.IllegalAccessError: tried to access class org.skywalking.apm.plugin.dubbo.DubboInterceptor$1 from class org.skywalking.apm.plugin.dubbo.DubboInterceptor
	at org.skywalking.apm.plugin.dubbo.DubboInterceptor.beforeMethod(DubboInterceptor.java:60)
	at org.skywalking.apm.agent.core.plugin.interceptor.enhance.InstMethodsInter.intercept(InstMethodsInter.java:62)
	at com.alibaba.dubbo.monitor.support.MonitorFilter.invoke(MonitorFilter.java)
	at com.alibaba.dubbo.rpc.protocol.ProtocolFilterWrapper$1.invoke(ProtocolFilterWrapper.java:91)
	at com.alibaba.dubbo.rpc.protocol.dubbo.filter.FutureFilter.invoke(FutureFilter.java:53)
	at com.alibaba.dubbo.rpc.protocol.ProtocolFilterWrapper$1.invoke(ProtocolFilterWrapper.java:91)
	at com.alibaba.dubbo.rpc.filter.ConsumerContextFilter.invoke(ConsumerContextFilter.java:48)
```

wu-sheng:
It is triggered by ClassLoader scope. Fixed by new API style.

PR #9622, opened 2022-09-16
Support analyzing cache related spans to provide metrics and slow commands for cache services from client side.

pg-yang:

Related issue  https://github.com/apache/skywalking/issues/9556 。
I‘m not sure whether CICD could pass . In my repo , CICD always need rerun one  time .


docs/menu.yml
```diff
@@ -111,6 +111,8 @@ catalog:
                 path: "/en/setup/backend/trace-sampling"
               - name: "Detect Slow Database Statement"
                 path: "/en/setup/backend/slow-db-statement"
+              - name: "Detect Slow Database Statement"
```
wu-sheng, line 114: You forgot to update the menu name.

wu-sheng (commented):
Are you keeping UI update(for root menu) and dashboard settings for another PR?

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/AnalyzerModuleConfig.java
```diff
@@ -60,6 +62,23 @@ public class AnalyzerModuleConfig extends ModuleConfig {
...
+    @Getter
+    private CacheWriteLatencyThresholdsAndWatcher cacheWriteLatencyThresholdsAndWatcher;
+
+    @Setter
+    @Getter
+    private String slowCacheReadThreshold = "default:20";
+
+    @Setter
+    @Getter
+    private CacheReadLatencyThresholdsAndWatcher cacheReadLatencyThresholdsAndWatcher;
```
wu-sheng, line 80: With all these new configs and dynamic configuration watchers added, the relative docs should be updated too.

oap-server/oal-grammar/src/main/antlr4/org/apache/skywalking/oal/rt/grammar/OALLexer.g4
```diff
@@ -45,6 +45,9 @@ SRC_SERVICE_INSTANCE_CLR_GC: 'ServiceInstanceCLRGC';
 SRC_SERVICE_INSTANCE_CLR_THREAD: 'ServiceInstanceCLRThread';
 SRC_ENVOY_INSTANCE_METRIC: 'EnvoyInstanceMetric';
 SRC_EVENT: 'Event';
+SRC_CACHE_READ: 'VirtualCacheRead';
+SRC_CACHE_WRITE: 'VirtualCacheWrite';
```
wu-sheng, line 49: Virtual Cache should be one source. And it should include one field named `op`.

oap-server/server-starter/src/main/resources/ui-initialized-templates/virtual_cache/virtual-cache-root.json
```diff
@@ -0,0 +1,122 @@
...
+            "url": "https://skywalking.apache.org/docs/main/latest/en/setup/service-agent/virtual-cache/"
+          },
+          "moved": false
+        }
+      ],
+      "id": "Virtual-Cache-Root",
+      "layer": "VIRTUAL_CACHE",
+      "entity": "All",
+      "name": "Virtual-Cache-Root",
+      "isRoot": true
```
wu-sheng, line 119: I don't think UI supports this layer as root. Could you provide your local screenshot?

test/e2e-v2/java-test-service/e2e-service-provider/src/main/java/org/apache/skywalking/e2e/controller/UserController.java
```diff
@@ -68,4 +80,39 @@ private long randomSleepLong(int min, int max) {
         int randomNumber = rand.nextInt((max - min) + 1) + min;
         return randomNumber;
     }
+
+    private void testCacheService() {
```
wu-sheng, line 84: @kezhenxu94 Do you have recommendation about the test service to include virtual cache testing?

test/e2e-v2/script/env
```diff
@@ -25,4 +25,4 @@ SW_KUBERNETES_COMMIT_SHA=0f3ec68e5a7e1608cec8688716b848ed15e971e5
 SW_ROVER_COMMIT=d956eaede57b62108b78bca48045bd09ba88e653
 SW_BANYANDB_COMMIT=5a326d7e36a008c5ea10e3ae506309cb29733c53
 
-SW_CTL_COMMIT=a7c4ccae106e4a5a0b6296f860396f289444058d
+SW_CTL_COMMIT=fd9b93348f51655ec58046aa645cc127c68bbdd2
```
wu-sheng, line 28: Why do you revert CLI commit? `fd9b` seems even not existing. What happens?

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/AnalyzerModuleProvider.java
```diff
@@ -85,15 +90,19 @@ public ModuleConfig createConfigBeanIfAbsent() {
 
     @Override
     public void prepare() throws ServiceNotProvidedException, ModuleStartException {
-        thresholds = new DBLatencyThresholdsAndWatcher(moduleConfig.getSlowDBAccessThreshold(), this);
+        dbLatencyThresholdsAndWatcher = new DBLatencyThresholdsAndWatcher(moduleConfig.getSlowDBAccessThreshold(), this);
+        cacheReadLatencyThresholdsAndWatcher = new CacheReadLatencyThresholdsAndWatcher(moduleConfig.getSlowCacheReadThreshold(), this);
+        cacheWriteLatencyThresholdsAndWatcher = new CacheWriteLatencyThresholdsAndWatcher(moduleConfig.getSlowCacheWriteThreshold(), this);
```
wu-sheng, line 95: Don't put new codes in the first lines. This doesn't have to be here. Please add after `traceSamplingPolicyWatcher`.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/vservice/VirtualCacheProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+        String peer = span.getPeer();
+        // peer is blank if it's a local span
+        if (StringUtil.isBlank(peer)) {
+            peer = tags.get(SpanTags.CACHE_TYPE) + "-local";
+        }
+        long timeBucket = TimeBucket.getMinuteTimeBucket(span.getStartTime());
+        String serviceName = namingControl.formatServiceName(peer);
+        int latency = (int) (span.getEndTime() - span.getStartTime());
+        if (latency > 0) {
+            log.info("latency data : {}", new Gson().toJson(span));
```
wu-sheng, line 78: If you call this, you have to use `log.isInfoEnable` as pre condition, otherwise, `new Gson().toJson(span)` would be called every time, which is very expensive.
Meanwhile, `new Gson` should only be created once as it is a utility.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/vservice/VirtualCacheProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+            if (latency > config.getCacheWriteLatencyThresholdsAndWatcher().getThreshold(cacheType)) {
+                sourceList.add(parseSlowWrite(segmentObject, latency, serviceName, span, tags));
+            }
+        } else if ("read".equals(op)) {
+            sourceList.add(parseReadMetrics(span, latency, serviceName, timeBucket));
+            if (latency > config.getCacheReadLatencyThresholdsAndWatcher().getThreshold(cacheType)) {
+                sourceList.add(parseSlowRead(segmentObject, latency, serviceName, span, tags));
+            }
+        } else {
+            log.warn("Illegal tag value :{} for 'cache.op' , only write/read would be accepted", op);
```
wu-sheng, line 93: Aren't you having `command` as `op`? And we don't put this warning because there is nothing to warn.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/vservice/VirtualCacheProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+    private final AnalyzerModuleConfig config;
+
+    private final List<Source> sourceList = new ArrayList<>();
+
+    @Override
+    public void prepareVSIfNecessary(SpanObject span, SegmentObject segmentObject) {
+        if (span.getSpanLayer() != SpanLayer.Cache) {
+            return;
+        }
+        Map<String, String> tags = span.getTagsList().stream()
```
wu-sheng, line 63: I can get the idea why you created `VirtualServiceProcessor`. But there is a side effect, you make every processor implementations iterate span's tags by themselves. This is a performance impact.
Because SkyWalking could process 100b segments, which means 1000-2000b spans per day. The increment of iteration could be a disaster.

test/e2e-v2/cases/storage/storage-cases.yaml
```diff
@@ -47,7 +47,7 @@ cases:
   - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql trace ls
     expected: expected/traces-list.yml
   # negative tags search: relationship should be logical AND instead of logical OR
-  - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql trace ls --tags http.method=POST,http.status_code=200
+  - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql trace ls --tags http.method=POST,http.status_code=200,iserror=true
```
wu-sheng, line 50: Why do you change this case?

test/e2e-v2/cases/meter/meter-cases.yaml
```diff
@@ -31,4 +31,50 @@
...
+      expected: expected/metrics-has-value0.yml
+    - query: |
+        swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql metrics linear --name=cache_read_sla  --service-id=R3VhdmFDYWNoZS1sb2NhbA==.0  | yq e 'to_entries' -
+      expected: expected/metrics-has-value.yml
+    - query: |
+        swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql metrics linear --name=cache_read_cpm  --service-id=R3VhdmFDYWNoZS1sb2NhbA==.0  | yq e 'to_entries' -
+      expected: expected/metrics-has-value.yml
+    - query: |
+        swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql metrics multiple-linear --name=cache_read_percentile  --service-id=R3VhdmFDYWNoZS1sb2NhbA==.0  | yq e 'to_entries | with(.[] ; .value=(.value | to_entries))' -
+      expected: expected/metrics-has-value-percentile.yml
```
wu-sheng, line 47: To save cost and time in the CI, I think we don't have to test all metrics of the virtual cache.

You could choose one metric and sampled record.

test/e2e-v2/java-test-service/e2e-service-provider/src/main/java/org/apache/skywalking/e2e/controller/UserController.java
wu-sheng: @pg-yang Could you share why do we need so many put/get in this method?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/DefaultScopeDefine.java
```diff
@@ -106,6 +107,10 @@ public class DefaultScopeDefine {
     public static final int ZIPKIN_SERVICE_SPAN = 52;
     public static final int ZIPKIN_SERVICE_RELATION = 53;
     public static final int PROCESS_RELATION = 54;
+    public static final int CACHE_READ_ACCESS = 55;
+    public static final int CACHE_WRITE_ACCESS = 56;
+    public static final int CACHE_SLOW_WRITE = 57;
+    public static final int CACHE_SLOW_READ = 58;
```
wu-sheng, line 113: These four should be merged into two. The source should include read/write as an attribute rather than using two sources.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/DefaultScopeDefine.java
```diff
@@ -163,18 +168,18 @@ private static final void addNewScope(ScopeDeclaration declaration, Class origin
         int id = declaration.id();
         if (ID_2_NAME.containsKey(id)) {
             throw new UnexpectedException(
-                "ScopeDeclaration id=" + id + " at " + originalClass.getName() + " has conflict with another named " + ID_2_NAME
-                    .get(id));
+                    "ScopeDeclaration id=" + id + " at " + originalClass.getName() + " has conflict with another named " + ID_2_NAME
+                            .get(id));
```
wu-sheng, line 172: All following changes in this file from here, are not necessary. You seem to have different format settings compared with others. Maybe it still could pass format check, but these should not exist in the git changes.

docs/en/changes/changes.md
```diff
@@ -12,6 +12,7 @@
 * Enhance LAL to support converting LogData to DatabaseSlowStatement.
 * [**Breaking Change**] Change the LAL script format(Add layer property).
 * Adapt ElasticSearch 8.1+, migrate from removed APIs to recommended APIs.
+* Add virtual cache to support analyze cache related span
```
wu-sheng, line 15:
```suggestion
* Support analyzing cache related spans to provide metrics and slow commands for cache services from client side.
```

pg-yang:
Couldn't you visit my comments? My comments are marked `pending`

image

wu-sheng:
You need to submit your comments from this

image

`Pending` means only visible for yourself.

pg-yang (commented):
finish my review

test/e2e-v2/cases/storage/storage-cases.yaml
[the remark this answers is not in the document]
wu-sheng: What do you mean `negative`?

wu-sheng:
image

You should reply on the same context. Starting a new comment doesn't help reviewer to follow.

wu-sheng:
> The UI code I updated:

Why don't you submit a pull request to update UI right? I can't see a local change would help in any case.

test/e2e-v2/cases/meter/meter-cases.yaml
wu-sheng: No, that is not testing about. What an e2e test is for checking the workflow of the target service. It costs time and money if you add all cases to verify. 
It is easy to say, let's add all cases, then one day, e2e would end in hours even over a day.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/DefaultScopeDefine.java
wu-sheng: There is a setting file in the root. Have you imported it?

test/e2e-v2/cases/storage/storage-cases.yaml
wu-sheng: So? What is wrong? e2e locked the agent version, and the PR you posted has been merged months ago.
e2e is not for every released version, that is not the purpose.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/vservice/VirtualCacheProcessor.java
wu-sheng: OK, make sense.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/vservice/VirtualCacheProcessor.java
wu-sheng: You should consider a metric call `cache_access_*` to cover all cases(read/write/others). This would show the access payload more accurate.

test/e2e-v2/cases/storage/storage-cases.yaml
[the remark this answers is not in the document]
pg-yang: According the  comment in this test case and  the expected data  , `negative ` means query nothing .
The previous java agent  collect the status_code which max  400 , so  the test case could match   expected data .
But the latest java agent collect all http code (include 200), In order to match expected data , I added the param  

You could find this PR https://github.com/apache/skywalking-java/pull/258

BTW , I don't know the meaning of this cache , I'm just keeping this case logically correct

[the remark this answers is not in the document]
wu-sheng: statu_code=200, isn't `iserror` should be `false`? 
And `is_error` should be a parameter, rather than a tag.

wu-sheng:
@pg-yang I am going to remove all your duplicated reply. It is really becoming hard to follow the discussions.

pg-yang:
> Why don't you submit a pull request to update UI right? I can't see a local change would help in any case.

Because  I'm not sure I'm allowed to submit PR  to UI , If i'm allowed , I could do it .

>  You should reply on the same context. Starting a new comment doesn't help reviewer to follow.

Very sorry , I added comments to explain why I update  `storage-cases` after submitting  this PR . And it's  marked as `pending ` .
After I  fix `pending ` , I make this PR chaotic . Very sorry .

wu-sheng:
> Because I'm not sure I'm allowed to submit PR to UI , If i'm allowed , I could do it .

There is nothing allowed in an open-source community. If there is only something that may not be correct, then reviewers would guide you out.

test/e2e-v2/cases/storage/storage-cases.yaml
pg-yang: Before  #258 PR , All http plugins only  collect http code which max 400 , so   the case was meet .
But  since #285 , http plugins collect all  http code , and  I updated `SW_AGENT_JAVA_COMMIT` to latest , so  the the case couldn't meet . In  order keep it correct , I added the param 


BTW  the expected data of this case if  `traces: []` .

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/DefaultScopeDefine.java
pg-yang: I  don't know `setting file in the root` you said

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/vservice/VirtualCacheProcessor.java
pg-yang: make sense

test/e2e-v2/cases/storage/storage-cases.yaml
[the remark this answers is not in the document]
pg-yang: This case means getting  empty result , I think .
I add the param to keep the result  empty .

wu-sheng: Then, why do you use a meaningless tag? Such as `randomKey=randomValue`? `isError` is a segment field, rather than a tag. Using this similar but not the same tag key would make ppl confused.

https://github.com/apache/skywalking-query-protocol/blob/2c541beae657306127f07af127d179492a1d7cf9/trace.graphqls#L28

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/DefaultScopeDefine.java
wu-sheng: https://github.com/apache/skywalking/blob/master/codeStyle.xml This is for IntelliJ

test/e2e-v2/cases/storage/storage-cases.yaml
pg-yang: OK, I would choose an appropriate tag

pg-yang:
Could you help me about CICD error  : https://github.com/apache/skywalking/actions/runs/3073171067/jobs/4965295825
I  don't  update the case

wu-sheng:
If you didn't touch relative codes, don't worry. We could recheck/rerun the wrong cases when you have fixed everything. Currently, the CI only matters for yourself to make sure codes are working.

oap-server/server-starter/src/main/resources/ui-initialized-templates/virtual_cache/virtual-cache-root.json
pg-yang: image
image


I would submit PR to UI repo later

pg-yang: image

image


I  would submit PR to UI repo later

test/e2e-v2/java-test-service/e2e-service-provider/src/main/java/org/apache/skywalking/e2e/controller/UserController.java
pg-yang: I have removed some put/get  method .

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/cache/CacheSlowAccessDispatcher.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.apache.skywalking.oap.server.core.analysis.SourceDispatcher;
+import org.apache.skywalking.oap.server.core.analysis.worker.TopNStreamProcessor;
+import org.apache.skywalking.oap.server.core.source.VirtualCacheSlowAccess;
+
+public class CacheSlowAccessDispatcher implements SourceDispatcher<VirtualCacheSlowAccess> {
+
+    @Override
+    public void dispatch(VirtualCacheSlowAccess source) {
+        if (source.getOp().equals("read")) {
```
wu-sheng, line 29:
```suggestion
        if ("read".equals(source.getOp())) {
```

This is a good practice to avoid NPE.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/cache/CacheSlowAccessDispatcher.java
```diff
@@ -0,0 +1,49 @@
...
+            TopNCacheReadCommand readCommand = new TopNCacheReadCommand();
+            readCommand.setId(source.getId());
+            readCommand.setCommand(source.getCommand() + " " + source.getKey());
+            readCommand.setLatency(source.getLatency());
+            readCommand.setTraceId(source.getTraceId());
+            readCommand.setServiceId(source.getCacheServiceId());
+            readCommand.setTimeBucket(source.getTimeBucket());
+            TopNStreamProcessor.getInstance().in(readCommand);
+        } else {
+            TopNCacheWriteCommand writeCommand = new TopNCacheWriteCommand();
```
wu-sheng, line 39: Do we have other commands? Should we use `else if` here?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/VirtualCacheAccess.java
```diff
@@ -0,0 +1,58 @@
...
+    private int cacheTypeId;
+    @Getter
+    @Setter
+    private int latency;
+    @Getter
+    @Setter
+    private boolean status;
+    @Getter
+    @Setter
+    private String op; // operation
```
wu-sheng, line 57:
```suggestion
    /**
     * op represents an operation of this cache access. 
     */
    @Getter
    @Setter
    private String op;
```

If you want to provide a comment, please provide details and for the field.

Also, I recommend using an enum rather than a string, it has higher performance and is harder to mis-set.  Refer `RequestType` which is in many sources.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/VirtualCacheSlowAccess.java
```diff
@@ -0,0 +1,66 @@
...
+    @Setter
+    private String traceId;
+
+    @Getter
+    @Setter
+    private boolean status;
+
+    @Getter
+    @Setter
+    private String op;
```
wu-sheng, line 54: Same recommendation. Please use the enum.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/CacheReadLatencyThresholdsAndWatcher.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public CacheReadLatencyThresholdsAndWatcher(String config, ModuleProvider provider) {
+        super(AnalyzerModule.NAME, provider, "slowCacheReadThreshold");
+        thresholds = new AtomicReference<>(new HashMap<>());
+        initialSettingsString = config;
+
+        activeSetting(config);
+    }
+
+    private void activeSetting(String config) {
```
wu-sheng, line 43: I noticed a potential bug in here, as same as `DBLatencyThresholdsAndWatcher#activeSetting`.

This method doesn't set `default` if it doesn't exist in the provided value. So, in `getThreshold`, a NPE would through as we try to use `thresholds.get().get("default")` for an `int.

We need to fix the existing(also update the change log to add a bug fix), also the new added two watchers.

If there is no default, we should add a default value, as same as we set in the `application.yml`. NOTICE, `initialSettingsString` could be overridden by system env or manual changing. We need to hard code a default value inside this class.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/cache/CacheSlowAccessDispatcher.java
pg-yang: We create `VirtualCacheSlowAccess` only  when the op is write/read

wu-sheng: OK, but when you change it to an enum, I think doing a `else if` check should be super fast, and make codes better.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/CacheReadLatencyThresholdsAndWatcher.java
pg-yang: Got it . I would fix these bug together .

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/vservice/VirtualCacheProcessor.java
```diff
@@ -0,0 +1,112 @@
...
+                .collect(Collectors.toMap(KeyStringValuePair::getKey, KeyStringValuePair::getValue));
+        if (!(tags.containsKey(SpanTags.CACHE_KEY) && tags.containsKey(SpanTags.CACHE_OP) && tags.containsKey(SpanTags.CACHE_CMD) && tags.containsKey(SpanTags.CACHE_TYPE))) {
```
pg-yang, line 61: I could change this condition to ` if (!tags.containsKey(SpanTags.CACHE_TYPE))`,
Because if Op ,Command ,key not exist , the `VirtualCacheAccess ` also work .

wu-sheng: Yes, you could. But we still need to check others to process slow commands when needed.

Also, one nit suggestion, you could use `get(key)` and `not null` instead of `containsKey` to get better performance. `containsKey` + `get` mean you have done `hashcode(key)` and `key match in same hash ID` twice.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/CacheReadLatencyThresholdsAndWatcher.java
pg-yang: I  have inspect the classes which extends `ConfigChangeWatcher` .  We should update `CacheReadLatencyThresholdsAndWatcher` , `CacheWriteLatencyThresholdsAndWatcher` , `DBLatencyThresholdsAndWatcher`. 
I  think we have two strategy to choose default value  when `get("default")` is Null :

1. Same as `application.yml` 
2. Choose a maximum,  e. g `DBLatencyThresholdsAndWatcher#getThreshold`, we choose `Interger.MAX_VALUE ` as default value  .

Strategy 1 ,  means we use default value from `application.yaml ` if user set default as null .

Strategy 2,  means all condition that depends the  value would be mismatched   (e.g. `if((latency > threshold)` always return false )

pg-yang: I would like Strategy 2 .  Because  If user set property null , it's better that we  could skip  some feature(function) ,
and If user want  some feature(function)  , they should specify appropriate property clearly .

pg-yang: If  we choose Strategy 2 , developers needn't keep the default value consistent  between hard code  and  `application.yaml`.

wu-sheng: Make sense.

wu-sheng:
You seem to have forgotten to update UTs accordingly. UTs fail in compiling.

wu-sheng:
image

According to the screenshots provided in [the UI PR](https://github.com/apache/skywalking-booster-ui/pull/159)
1. Please move access metrics on the top of read and write, and access is more general
2. Have you set an association for metrics among inside access(also read and write). https://skywalking.apache.org/docs/main/next/en/ui/readme/#association-options

wu-sheng:
UI PR merged, please update the submodule and changelog as well.

pg-yang:
> 2\. Have you set an association for metrics among inside access(also read and write). https://skywalking.apache.org/docs/main/next/en/ui/readme/#association-options

I  have set association , but  couldn't take affect .

image

wu-sheng:
You need to click the dot on the line to see association.

pg-yang:
image

image

image

Done

wu-sheng:
Please consider make access on the top, then read on the left(2 lines, 2 metrics per line) and slow SQLs, write on the right?

pg-yang:
image


right?

wu-sheng:
Yes, this should be good.

wu-sheng:
I can't see UI submodule update

pg-yang:
> I can't see UI submodule update
Here

image

wu-sheng (approved):
You should be able to prepare the showcase update to bring this new feature into the demo env.

pg-yang:
I  have prepared show case ,  after this PR merged , would test on My PC .

docs/en/concepts-and-designs/scope-definitions.md
```diff
@@ -286,3 +286,54 @@ This calculates the metrics data from [events](event.md).
...
+
+This calculates the metrics data from each request of cache system.
+
+| Name | Remarks | Group Key | Type | 
+|---|---|---|---|
+| name | The service name of virtual cache service. |  | string |
+| cacheTypeId | The ID of the component used in this call. | | int |
+| latency | The time taken by each request. | | int(in ms)|
+| status | Indicates the success or failure of the request.| | boolean |
+| op | Indicates this access is used for `write` or `read` | | string |
```
kezhenxu94, line 324: This is `operation` in code

```suggestion
| operation | Indicates this access is used for `write` or `read` | | string |
```

docs/en/concepts-and-designs/scope-definitions.md
```diff
@@ -286,3 +286,54 @@ This calculates the metrics data from [events](event.md).
...
+
+| Name | Remarks | Group Key | Type | 
+|---|---|---|---|
+| cacheServiceId | The service id of virtual cache service. |  | string |
+| command | The cache command . | | string |
+| key | The cache command key. | | string|
+| latency | The time taken by each request. | | int(in ms)|
+| traceId | The traceId of this slow access| | string|
+| status | Indicates the success or failure of the request.| | boolean |
+| op | Indicates this access is used for `write` or `read` | | string |
```
kezhenxu94, line 339:
```suggestion
| operation | Indicates this access is used for `write` or `read` | | string |
```

kezhenxu94 (commented):
@pg-yang can you update the doc in a new PR?

pg-yang:
OK

wu-sheng:
One more thing, the scope should are not as the same as the codes. `VirtualCacheAccess`, `VirtualCacheSlowAccess`
If you want to keep the cache consistent with the database, please remove `virtual` from source names.

PR #9632, opened 2022-09-19
[Critical] Remove physical index existing check and fix TTL activation bug

wu-sheng:

I removed this only to avoid blocking booting, but it seems TTL timer fails to create no data index somehow. @wankai123

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -2,7 +2,6 @@
 
 #### Project
 
-
 #### OAP Server
 
 * Add component ID(133) for impala JDBC Java agent plugin and component ID(134) for impala server.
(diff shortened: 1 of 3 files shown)


kezhenxu94 (dismissed):

wankai123 (approved):

wankai123 (approved):

wu-sheng:
> but it seems TTL timer fails to create no data index somehow. 

It has been confirmed, the TTL selector mechanism could be failure due to OAP instance list not ordered. Then in a multi nodes env, there could be no TTL running.

wankai123 (approved):

wu-sheng:
I posted a discussion for people easier to understand and follow this critical bug, https://github.com/apache/skywalking/discussions/9634

PR #9641, opened 2022-09-20
Fix dashboard link for virtual cache

wu-sheng:

@pg-yang We have 302 for `*/latest/*`, but the official should be `*/next/*`. Ref https://skywalking.apache.org/docs/main/next/en/setup/service-agent/virtual-cache/


kezhenxu94 (approved):

pg-yang:
Got it

PR #9663, opened 2022-09-22
Fix inconsistent virtual cache describe between code and scope definitions doc

pg-yang:

### DIFF
--- a/docs/en/concepts-and-designs/scope-definitions.md
+++ b/docs/en/concepts-and-designs/scope-definitions.md
@@ -315,7 +315,7 @@ This calculates the metrics data from each request of cache system.
 | cacheTypeId | The ID of the component used in this call. | | int |
 | latency | The time taken by each request. | | int(in ms)|
 | status | Indicates the success or failure of the request.| | boolean |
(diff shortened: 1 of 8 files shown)


wu-sheng:
One more, https://github.com/apache/skywalking/pull/9622#issuecomment-1254442278

pg-yang:
It's  necessary to update change.md . Right?

wu-sheng (commented):
Once you changed the source codes, OAL has to change according.

wu-sheng:
Also, antlr grammar file. 
Changlog is not relative breaking, you don't need to change, because this feature is not released.

pg-yang:
Got it , I should  run test case on my PC firstly .

pg-yang:
Local test have passed

wu-sheng (approved):
Once e2e passed, it should be good.

pg-yang:
Somethings throw error  after reruning

wu-sheng:
What things? Something env relative?

pg-yang:
Done , a full test take half of day

wu-sheng:
In the afternoon, we would face resource competition with all other projects mostly working at Euro.

PR #9681, opened 2022-09-25
Add property mutation mechanism to simplify dynamic config watcher in…

pg-yang:

This PR for preview  . I would list the main point of modification 
 I  would polish up it  if  you agree


oap-server/server-configuration/configuration-api/src/main/java/org/apache/skywalking/oap/server/configuration/api/ConfigWatcherMutator.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.skywalking.oap.server.library.module.ModuleProvider;
+
+public interface ConfigWatcherMutator extends ModuleConfigMutator {
+    /**
+     * mutate the propertyValue to Dynamic Watcher which is marked with @AsWatcher
+     * The filed type class must have appropriate constructor method , which signature is
+     * "(String itermName , String config, ModuleProvider provider)"
+     */
+    @Override
+    default Object mutate(final ModuleDefine currentModule,
```
pg-yang, line 34: Convert property value to watcher  if it's marked witch @AsWatcher

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/AnalyzerModuleConfig.java
```diff
@@ -19,34 +19,29 @@
...
 
 @Slf4j
-public class AnalyzerModuleConfig extends ModuleConfig {
-    /**
-     * The sample policy setting file
-     */
-    @Setter
-    @Getter
-    private String traceSamplingPolicySettingsFile;
+public class AnalyzerModuleConfig extends ModuleConfig implements ConfigWatcherMutator {
```
pg-yang, line 40: implements ConfigWatcherMutator

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/AnalyzerModuleProvider.java
```diff
@@ -123,12 +97,7 @@ public void start() throws ModuleStartException {
                                                                               .provider()
                                                                               .getService(
                                                                                   DynamicConfigurationService.class);
-        dynamicConfigurationService.registerConfigChangeWatcher(dbLatencyThresholdsAndWatcher);
-        dynamicConfigurationService.registerConfigChangeWatcher(uninstrumentedGatewaysConfig);
-        dynamicConfigurationService.registerConfigChangeWatcher(traceSamplingPolicyWatcher);
-        dynamicConfigurationService.registerConfigChangeWatcher(cacheReadLatencyThresholdsAndWatcher);
-        dynamicConfigurationService.registerConfigChangeWatcher(cacheWriteLatencyThresholdsAndWatcher);
-
+        moduleConfig.registerMarkedWatcherTo(dynamicConfigurationService);
```
pg-yang, line 100: Register watcher filed which is marked with @AsWatcher

wu-sheng:
Currently, the boundaries of APIs are not clear. As I mentioned in the discussion. the module manager should never include the watcher mechanism into the core, no matter in any case. 
And the `ModuleConfigMutator` is not necessary. Also in the previous discussion, you proposed an interface to config the initializer to create an instance. But you seem to change all the things from the discussion. 
The previous discussion lost the whole point.

wu-sheng:
You were proposing to remove string config, but keep watcher instance to be created automatically through module manager. But now, you are adding more things, you want to remove the manual watcher registration mechanism, which is never discussed, and AFAIK, this is not recommended.

wu-sheng:
One of the SkyWalking APIs and integration design principles is, doesn't add many implicit mechanisms as Spring did.
We don't say Spring is bad, we just want to lower the bar of ppl who want to contribute to SkyWalking. Today, if you want to learn a project deeply involved in Spring, pure knowing Java is not enough.

I think you were new to SkyWalking codebases, you could try to recall your old days, and compare them with what you are trying to do right now. The new ppl in the future would be very confused about how a field would work magically.

The OAL engine(a compiling language) and MAL/LAL engines(groovy based) have already put a very high bar when ppl want to change SkyWalking OAP server. But at least, right now, the module management mechanism is still simple. Let's keep this as much as possible.

Issue #9477, opened 2022-08-17
[Feature] Support multiple MySQL instances as a sharding cluster powered by ShardingSphere.


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

As ShardingSphere provides a new DistSQL concept recently, it gives us an opportunity to evaluate whether it is good enough to create dynamic routing rules automatically.

SkyWalking creates new tables(in MySQL and PostgreSQL) according to metrics definition based on MAL and OAL, the traditional configuration file based sharding is not suitable for us.

Also, we need to use `hint` mechanism to try to implement TTL deleting through `truncate`, rather than the heavy `delete`.

### Use case

Support Sharding and Time-series based rolling for metrics/logs/traces/events/profiling telemetry data in SkyWalking

### Related issues

Doc of ShardingSphere, https://shardingsphere.apache.org/document/5.1.2/en/reference/distsql/syntax/rdl/rule-definition/sharding/create-default-sharding-strategy/

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

tristaZero:
Hello, it looks cool. ;-) 
Sure, [ShardingSphere](https://shardingsphere.apache.org/) not only provides the DistSQL(SQL-Like) to manage your distributed or sharding DB and table but makes itself more cloud-native on K8S by providing an [operator](https://medium.com/codex/database-plus-embracing-the-cloud-shardingsphere-on-cloud-solution-released-29916290ad06).

wu-sheng:
MySQL-Sharding is supported since 9.3.0, ref https://skywalking.apache.org/docs/main/next/en/setup/backend/backend-storage/#mysql-sharding

We are working with ShardingSphere community to expand better compatibility, because right now, this feature is only working with 5.1.2 release, but fails in 5.2.0 and 5.2.1

A blog would be posted to provide more introduction after SkyWalking's 9.3.0 release.

PR #9689, opened 2022-09-26
Support sharding MySQL database instances and tables.

wankai123:
- Support sharding MySQL database instances and tables by [Shardingsphere-Proxy](https://shardingsphere.apache.org/document/current/en/overview/#shardingsphere-proxy).
relate to https://github.com/apache/skywalking/issues/9477

### Support sharding MySQL database instances and tables


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/input/Duration.java
```diff
@@ -27,45 +27,50 @@
 
 @Getter
 @Setter
+/**
+ * To optimize the query, the range of start and end times will be trimmed to [TTL_deadLine < time <= CurrentTime].
+ */
```
wu-sheng, line 32:
```suggestion
/**
 * @since 9.3.0 To optimize the query, the range of start and end times will be trimmed to [TTL_deadLine < time <= CurrentTime].
 */
@Getter
@Setter
```

This seems a better comment style?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/ShardingAlgorithm.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ */
+
+package org.apache.skywalking.oap.server.core.storage;
+
+public enum ShardingAlgorithm {
+    TIME_RELATIVE_ID_SHARDING_ALGORITHM,
+    TIME_RANGE_SHARDING_ALGORITHM,
+    TIME_BUCKET_SHARDING_ALGORITHM
+}
```
wu-sheng, line 25: Let's add a full description of these three algorithms, as they are very important from a design perspective.
Put some demo layouts of the target data.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/SQLDatabaseModelExtension.java
```diff
@@ -37,6 +39,12 @@ public class SQLDatabaseModelExtension {
     //exclude the columns from the main table
     private final List<ModelColumn> excludeColumns = new ArrayList<>(5);
 
+    @Setter
+    private boolean isShardingTable = false;
+    @Setter
+    @Getter
+    private Sharding sharding;
```
wu-sheng, line 46: Merge these two as `Optional<Sharding>`?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/zipkin/ZipkinSpanRecord.java
```diff
@@ -71,6 +75,7 @@ public class ZipkinSpanRecord extends Record {
     @Setter
     @Getter
     @Column(columnName = TRACE_ID)
+    @SQLDatabase.AdditionalEntity(additionalTables = {ADDITIONAL_QUERY_TABLE}, reserveOriginalColumns = true)
     private String traceId;
```
wu-sheng, line 79: Is this a bug fix? If so, we need a change log about this.

wankai123: No, just for additional table data source sharding.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/input/Duration.java
```diff
@@ -27,45 +27,50 @@
...
     private String end;
     private Step step;
 
     /**
-     * See {@link DurationUtils#convertToTimeBucket(Step, String)}
+     * See {@link DurationUtils#trimToStartTimeBucket(Step, String, boolean)}
      */
-    public long getStartTimeBucket() {
-        return DurationUtils.INSTANCE.convertToTimeBucket(step, start);
+    public long getStartTimeBucket(boolean isRecord) {
```
wu-sheng, line 41: When I review and read this method, I am wondering, whether we separate this method as two, is that better for reading?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/input/Duration.java
```diff
@@ -27,45 +27,50 @@
...
+            step, DurationUtils.INSTANCE.trimToEndTimeBucket(step, end));
     }
 
     /**
      * Assemble time point based on {@link #step} and {@link #start} / {@link #end}
      */
-    public List<PointOfTime> assembleDurationPoints() {
-        return DurationUtils.INSTANCE.getDurationPoints(step, getStartTimeBucket(), getEndTimeBucket());
+    public List<PointOfTime> assembleDurationPoints(boolean isRecord) {
+        return DurationUtils.INSTANCE.getDurationPoints(step, getStartTimeBucket(isRecord), getEndTimeBucket());
```
wu-sheng, line 74: This change is going to break UI if the query happens on the edge. Because the `PointOfTime` is shorter than expected.
From my understanding, there are not enough elements in the list.

test/e2e-v2/cases/log/mysql/sharding/docker-compose.yml
```diff
@@ -0,0 +1,101 @@
...
+    environment:
+      MYSQL_ROOT_PASSWORD: "root@1234"
+      MYSQL_DATABASE: "swtest"
+      MYSQL_ROOT_HOST: "%"
+    healthcheck:
+      test: ["CMD", "bash", "-c", "cat < /dev/null > /dev/tcp/127.0.0.1/3306"]
+      interval: 5s
+      timeout: 60s
+      retries: 120
+
```
kezhenxu94, line 64: Try to use yaml anchor to reduce duplications

```suggestion

  mysql-1: &mysql
    image: mysql/mysql-server:8.0.13
    networks:
      - e2e
    ports:
      - 3306
    environment:
      MYSQL_ROOT_PASSWORD: "root@1234"
      MYSQL_DATABASE: "swtest"
      MYSQL_ROOT_HOST: "%"
    healthcheck:
      test: ["CMD", "bash", "-c", "cat < /dev/null > /dev/tcp/127.0.0.1/3306"]
      interval: 5s
      timeout: 60s
      retries: 120

  mysql-2:
    <<: *mysql
```

test/e2e-v2/cases/storage/mysql/sharding/docker-compose.yml
```diff
@@ -0,0 +1,114 @@
...
+    environment:
+      MYSQL_ROOT_PASSWORD: "root@1234"
+      MYSQL_DATABASE: "swtest"
+      MYSQL_ROOT_HOST: "%"
+    healthcheck:
+      test: ["CMD", "bash", "-c", "cat < /dev/null > /dev/tcp/127.0.0.1/3306"]
+      interval: 5s
+      timeout: 60s
+      retries: 120
+
```
kezhenxu94, line 64:
```suggestion
  mysql-1: &mysql
    image: mysql/mysql-server:8.0.13
    networks:
      - e2e
    ports:
      - 3306
    environment:
      MYSQL_ROOT_PASSWORD: "root@1234"
      MYSQL_DATABASE: "swtest"
      MYSQL_ROOT_HOST: "%"
    healthcheck:
      test: ["CMD", "bash", "-c", "cat < /dev/null > /dev/tcp/127.0.0.1/3306"]
      interval: 5s
      timeout: 60s
      retries: 120

  mysql-2:
    <<: *mysql
```

test/e2e-v2/cases/zipkin/mysql/sharding/docker-compose.yml
```diff
@@ -0,0 +1,113 @@
...
+      MYSQL_ROOT_PASSWORD: "root@1234"
+      MYSQL_DATABASE: "swtest"
+      MYSQL_ROOT_HOST: "%"
+    healthcheck:
+      test: ["CMD", "bash", "-c", "cat < /dev/null > /dev/tcp/127.0.0.1/3306"]
+      interval: 5s
+      timeout: 60s
+      retries: 120
+
+  mysql-2:
```
kezhenxu94, line 49: Same here

test/e2e-v2/script/docker-compose/base-compose.yml
```diff
@@ -114,5 +114,21 @@ services:
       timeout: 60s
       retries: 120
 
+  sharding-proxy:
```
kezhenxu94, line 117: This base docker is for some very basic components that all cases should use, I'd rather not to add this here, otherwise future developers tend to add all other components here and this file will lose its point

test/e2e-v2/cases/browser/mysql/sharding/docker-compose.yml
```diff
@@ -0,0 +1,176 @@
...
+      MYSQL_ROOT_PASSWORD: "root@1234"
+      MYSQL_DATABASE: "swtest"
+      MYSQL_ROOT_HOST: "%"
+    healthcheck:
+      test: ["CMD", "bash", "-c", "cat < /dev/null > /dev/tcp/127.0.0.1/3306"]
+      interval: 5s
+      timeout: 60s
+      retries: 120
+
+  mysql-2:
```
kezhenxu94, line 49: And here

test/e2e-v2/script/prepare/setup-ss-proxy/download-mysql.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
kezhenxu94, line 1: Consider reusing the existing `download-mysql.sh` instead of duplicating it

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/input/Duration.java
kezhenxu94: > When I review and read this method, I am wondering, whether we separate this method as two, is that better for reading?

Aggree, when adding a `boolean` argument to a method, it's usually better in terms of readability to split the method into two, like `getRecordStartTimeBucket` and `getXXXStartTimeBucket`

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/SQLDatabase.java
```diff
@@ -135,4 +136,12 @@
     @interface MultipleExtraColumn4AdditionalEntity {
         ExtraColumn4AdditionalEntity[] value();
     }
+
+    @Target({ElementType.TYPE})
+    @Retention(RetentionPolicy.RUNTIME)
+    @interface Sharding {
+        ShardingAlgorithm shardingAlgorithm();
+        String dsShardingColumn();
```
kezhenxu94, line 144: In such case I'd rather write the full name, `dataSourceShardingColumn`, with IDE autocompletion it's the same key strokes but it is more clear

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/HistogramFunction.java
```diff
@@ -31,19 +31,25 @@
...
+import static org.apache.skywalking.oap.server.core.analysis.metrics.Metrics.ID;
+
 /**
  * Histogram includes data range buckets and the amount matched/grouped in the buckets. This is for original histogram
  * graph visualization
  */
 @MeterFunction(functionName = "sumHistogram")
 @Slf4j
 @ToString
+@SQLDatabase.Sharding(shardingAlgorithm = ShardingAlgorithm.TIME_RELATIVE_ID_SHARDING_ALGORITHM, tableShardingColumn = ID, dsShardingColumn = ENTITY_ID)
```
wu-sheng, line 52: I think we should move this into `Meter`? As it has `getEntityId` method, all implementations should have this field?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/MaxLongMetrics.java
```diff
@@ -23,12 +23,18 @@
...
+import org.apache.skywalking.oap.server.core.storage.ShardingAlgorithm;
 import org.apache.skywalking.oap.server.core.storage.annotation.Column;
+import org.apache.skywalking.oap.server.core.storage.annotation.SQLDatabase;
+
+import static org.apache.skywalking.oap.server.core.analysis.metrics.Metrics.ENTITY_ID;
+import static org.apache.skywalking.oap.server.core.analysis.metrics.Metrics.ID;
 
 /**
  *
  **/
```
wu-sheng, line 35: Remove empty comments(legacy codes).

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/StorageModels.java
```diff
@@ -250,6 +250,14 @@ private void retrieval(final Class<?> clazz,
             }
         }
 
+        if (clazz.isAnnotationPresent(SQLDatabase.Sharding.class)) {
+            SQLDatabase.Sharding sharding = clazz.getAnnotation(SQLDatabase.Sharding.class);
```
wu-sheng, line 254: `SQLDatabase.Sharding.class` is present in the parent class. `FunctionCategory#doGetAnnotation` use a recursion to read **MeterFunction** annotation from parent class.

docs/en/setup/backend/configuration-vocabulary.md
```diff
@@ -130,6 +130,14 @@ The Configuration Vocabulary lists all available configurations provided by `app
...
+| -                       | -             | numOfSearchableValuesPerTag                                                                                                                                              | In a trace segment, this includes multiple spans with multiple tags. Different spans may have same tag key, e.g. multiple HTTP exit spans all have their own `http.method` tags. This configuration sets the limit on the maximum number of values for the same tag key.                                                                                                                                                                                        | SW_STORAGE_NUM_OF_SEARCHABLE_VALUES_PER_TAG       | 2                                                                       |
```
wu-sheng, line 137: These seem expired configurations from MySQL and PostgreSQL configurations.

docs/en/setup/backend/backend-storage.md
```diff
@@ -211,6 +212,34 @@ All connection-related settings, including URL link, username, and password, are
 Only part of the settings is listed here. See the [HikariCP](https://github.com/brettwooldridge/HikariCP) connection pool document for full settings.
 To understand the function of the parameter `rewriteBatchedStatements=true` in MySQL, see the [MySQL official document](https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-connp-props-performance-extensions.html#cj-conn-prop_rewriteBatchedStatements).
 
+## MySQL-Sharding
```
kezhenxu94, line 215: Is this only applicable to MySQL? Or is it applicable to other JDBC protocol?

wankai123: This PR only applicable to MySQL, postgre needs some adaptation and verification.

wu-sheng: We are not certain about ShardingSphere compatible capability crossing MySQL and PostgreSQL.
Kai has noticed and reported to their community, we could only work with ShardingSphere 5.1.2, but 5.2.0 doesn't work.

PR #9710, opened 2022-09-29
Remove Prometheuse fetcher, fix MAL Histogram type bug, and reorganize the doc

### LINKED ISSUE #110: Support transaction sampling
If agent send all TraceSegment to collector-cluster, it costs so much network bandwidth.

We provide a config set the percentage of sampling. This minimum unit is 1/10000, which means send one TraceSegment to the collector every 10000 times.

When we face this scenario: A -> B, if TraceSegment is sampled in A, then the related TraceSegment in B must be sampled, no matter you sampling rate.

If we want to support this, we will add a new field(`sampling`) in TraceSegmentRef.

cc @ascrutae.

And @pengys5 , I think this do not influence the Collector? Confirm?

### LINKED ISSUE #9691: [Bug] OpenTelemetryMetricHandler Histogram buildBucket
### Search before asking

### What happened

When using OpenTelemetry Histogram metric it not working correctly

The metrics generate from opentelemetry client will be same result.

### How to reproduce

Use default application.yaml 

the opentelemetry client (java) create DoubleHistogram, and record 1 ~ 1000.
total count is 1000, and sum is 500500
```
Metric #0
Descriptor:
     -> Name: test
     -> Description: 
     -> Unit: 
     -> DataType: Histogram
     -> AggregationTemporality: AGGREGATION_TEMPORALITY_CUMULATIVE
HistogramDataPoints #0
StartTimestamp: 2022-09-27 04:41:01.914 +0000 UTC
Timestamp: 2022-09-27 04:42:01.921 +0000 UTC
Count: 1000
Sum: 500500.000000
ExplicitBounds #0: 5.000000
ExplicitBounds #1: 10.000000
ExplicitBounds #2: 25.000000
ExplicitBounds #3: 50.000000
ExplicitBounds #4: 75.000000
ExplicitBounds #5: 100.000000
ExplicitBounds #6: 250.000000
ExplicitBounds #7: 500.000000
ExplicitBounds #8: 750.000000
ExplicitBounds #9: 1000.000000
ExplicitBounds #10: 2500.000000
ExplicitBounds #11: 5000.000000
ExplicitBounds #12: 7500.000000
ExplicitBounds #13: 10000.000000
Buckets #0, Count: 5
Buckets #1, Count: 5
Buckets #2, Count: 15
Buckets #3, Count: 25
Buckets #4, Count: 25
Buckets #5, Count: 25
Buckets #6, Count: 150
Buckets #7, Count: 250
Buckets #8, Count: 250
Buckets #9, Count: 250
Buckets #10, Count: 0
Buckets #11, Count: 0
Buckets #12, Count: 0
Buckets #13, Count: 0
Buckets #14, Count: 0
```

### Anything else

I think [OpenTelemetry Metric Handler buildBucket](https://github.com/apache/skywalking/blob/master/oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricHandler.java#L155) function is bug or change it.

The process of [OpenTelemetry Metric Handler](https://github.com/apache/skywalking/blob/master/oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricHandler.java) is
1. get Metrics from opentelemetry metric export
2. change to Prometheus metric format
3. do analyze

In step#2 send to [PrometheusConverter](https://github.com/apache/skywalking/blob/bf4cf9d8a026b82693066719c1607eb79d35a844/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/PrometheusMetricConverter.java#L57)

Prometheus Histogram Metric using 'le', but opentelemetry histogram has no 'le' using [explicit_bounds](https://github.com/apache/skywalking/blob/master/oap-server/server-receiver-plugin/receiver-proto/src/main/proto/opentelemetry/proto/metrics/v1/metrics.proto#L441), so In PrometheusConverter create 'le' through explicit bounds

In opentelemetry-proto file the histogram is [boundary](https://github.com/apache/skywalking/blob/master/oap-server/server-receiver-plugin/receiver-proto/src/main/proto/opentelemetry/proto/metrics/v1/metrics.proto#L433),

opentelemetry like
0 ~ 100 : 20
100 ~ 200 : 30
200 ~ 300 : 10
300 ~ +Inf : 0

but prometheus like
le 100 : 20
le 200 : 50
le 300 : 60

The problem occurs when [SampleFamily histogram method](https://github.com/apache/skywalking/blob/bf4cf9d8a026b82693066719c1607eb79d35a844/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/SampleFamily.java#L408)

I think it will be change to opentelemetry histogram build bucket method to cumulative.

Am I misunderstood opentelemetry-proto file?

### Code of Conduct

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #9691 .


docs/en/setup/backend/prometheus-metrics.md
```diff
@@ -1,84 +1,4 @@
...
-`prometheus-fetcher` as follows:
-
-```yaml
-prometheus-fetcher:
-  selector: ${SW_PROMETHEUS_FETCHER:default}
-  default:
-    enabledRules: ${SW_PROMETHEUS_FETCHER_ENABLED_RULES:"self,metric-custom"}
-```
+The Prometheus fetcher is deprecated and remove since 9.3.0,
+please use [OpenTelemetry receiver](opentelemetry-receiver.md) instead.
```
wu-sheng, line 4: Are you avoiding 404 page to keep this? I prefer to remove this from docs and menu.

kezhenxu94: > Are you avoiding 404 page to keep this? I prefer to remove this from docs and menu.

Yes, I can remove this if desired

wu-sheng: If we want to avoid 404, you could keep markdown file, but remove menu. Then it would work as protection for the expired link.

wu-sheng:
There seems one UT failing, it seems relative.

docs/en/setup/backend/backend-telemetry.md
```diff
@@ -33,63 +33,13 @@ telemetry:
     port: 1543
 ```
 
-2. Set up Prometheus fetcher.
+2. Set up OpenTelemetry to scrape the metrics from OAP telemetry.
```
wu-sheng, line 36: We should link to e2e setup as a good example to setup otel collector.

kezhenxu94: I think so, but we don't have such E2E using otel to collect so11y metrics

wu-sheng: Oh, I see. Let's add a TODO issue to track e2e missing for this. 

Right now, we could link to the showcase?

kezhenxu94: Now this is added

wu-sheng (approved):
LGTM once e2e passed.

wankai123 (approved):
LGTM

wu-sheng (dismissed):

docs/en/concepts-and-designs/mal.md
```diff
@@ -252,6 +252,45 @@ They extract level relevant labels from metric labels, then informs the meter-sy
...
    extracts `Service` labels from the first array argument, extracts `Instance` labels from the second array argument, extracts `ProcessID` labels from the fourth and fifth arguments of the source and destination.
 
+## Configuration file
+
+The OAP can load the configuration at bootstrap. If the new configuration is not well-formed, the OAP fails to start up. The files
+are located at `$CLASSPATH/otel-rules`.
+
+The file is written in YAML format, defined by the scheme described below. Brackets indicate that a parameter is optional.
+
+A full example can be found [here](../../../oap-server/server-starter/src/main/resources/otel-rules/oap.yaml)
```
wu-sheng, line 262: I think this is not very accurate. Not only OTEL files are using the MAL engine, but meter analysis is also using this. too.

kezhenxu94: This is an example. What do you expect then?

wu-sheng: I mean we should list MAL file folders.

wu-sheng (approved):

PR #9711, opened 2022-09-29
(Summer 2022)Feat add golang runtime metrics

easonyipj:
### <Feature description>

# Feature Description
This feature is about the golang runtime metrics collcection, while using go2Sky gRPCReporter, it can collect golang runtime metrics automatically, and sending the metrics to OAP Server where you can monitor these metrics in general-Instance dashboard.

Supported metrics:

- instance_golang_heap_alloc
bytes of allocated heap objects.
- instance_golang_stack_used
bytes in stack spans.
- instance_golang_gc_count
the gc count in one minute
- instance_golang_gc_pause_time
the sum of all gc pause time in one minute
- instance_golang_live_goroutines_count
the number of goroutines that currently exist.
- instance_golang_os_threads_count
the number of records in the thread creation profile
- instance_host_cpu_used_rate
the percentage of cpu used combined.
- instance_host_mem_used_rate
the memory used rate

# How it works
1. The skywalking-meter-receiver-plugin will receive the meter data sending by go2sky
2. The meter sysytem will process these meters according to the meter-analyzer-config/golang.yaml
3. The meter panel defined in general-instance.json, the ui system will load there panels when you open the General-Instance dashboard and choose the Golang tap

# How to use
1. Update application.yml to add golang in meterAnalyzerActiveFiles
`meterAnalyzerActiveFiles: ${SW_METER_ANALYZER_ACTIVE_FILES:datasource,threadpool,satellite,golang}`
3. Start your go2sky agent and oap server
4. Open the  General-Instance dashboard and choose Golang tap

# Example
example image


oap-server/server-starter/src/main/resources/meter-analyzer-config/golang.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+expSuffix: instance(['service'], ['instance'], Layer.GENERAL)
+metricPrefix: meter
+metricsRules:
+  - name: instance_golang_heap
+    exp: instance_golang_heap
+  - name: instance_golang_stack
+    exp: instance_golang_stack
+  - name: instance_golang_gc_time
+    exp: instance_golang_gc_time
```
wu-sheng, line 24: Is this a gauge?

wu-sheng:
Also, if you want to activate this, you should add this file to the activated in the application.yml.

easonyipj:
> Also, if you want to activate this, you should add this file to the activated in the application.yml.

Yes, I just added that

oap-server/server-starter/src/main/resources/ui-initialized-templates/general/general-instance.json
```diff
@@ -1012,6 +1012,230 @@
...
+                  },
+                  "metrics": [
+                    "meter_instance_golang_gc_time"
+                  ],
+                  "metricTypes": [
+                    "readMetricsValues"
+                  ],
+                  "metricConfig": [
+                    {
+                      "calculation": "nanoToMill"
```
wu-sheng, line 1107: Pending on https://github.com/apache/skywalking-booster-ui/pull/164

wu-sheng:
> > Also, if you want to activate this, you should add this file to the activated in the application.yml.
> 
> Yes, I just added that

Locally? I can't see it from this PR.

easonyipj:
> > > Also, if you want to activate this, you should add this file to the activated in the application.yml.
> > 
> > 
> > Yes, I just added that
> 
> Locally? I can't see it from this PR.

Sorry, I didn't make it clear, I mean I have mentioned the user who wants to see golang meter should active the golang.yaml in application.yml on the document. Should I add golang to meterAnalyzerActiveFiles in this pr directly neither than let user to active?

wu-sheng:
> I mean I have mentioned the user who wants to see golang meter should active the golang.yaml in application.yml on the document.

Which document? I didn't notice that too.

> Should I add golang to meterAnalyzerActiveFiles in this pr directly neither than let user to active?

Yes, you should.

easonyipj:
> > I mean I have mentioned the user who wants to see golang meter should active the golang.yaml in application.yml on the document.
> 
> Which document? I didn't notice that too.
> 
> > Should I add golang to meterAnalyzerActiveFiles in this pr directly neither than let user to active?
> 
> Yes, you should.

The pr docs, I mentioned it at the first step on "How to use" module, I will directly add golang to meterAnalyzerActiveFiles in application.yml, thanks for reminding

wu-sheng:
> The pr docs, I mentioned it at the first step on "How to use" module

Could you share the link here? I can't follow your point.

wu-sheng:
image

This is all I can see about the changes.

easonyipj:
> Don't remove the template, which indicates you should update `changes.md`.

OK， I will change it

easonyipj:
> image
> 
> This is all I can see about the changes.

Sorry, I mean this, maybe it should not be called "pr docs"? And I will change there words follow the pr template
example image

wu-sheng:
No, that is not docs. And there is no PR's docs thing. 

This is the [official doc](https://skywalking.apache.org/docs/#SkyWalking), which is built from [here](https://github.com/apache/skywalking/blob/master/docs/README.md) from the latest to every selected version.

What you wrote here is called the content or description of this PR. The purpose of this is to introduce what you changed and why you changed in the way. You should write what you had written here, that is good practice. Just it isn't being called a doc.

easonyipj:
> No, that is not docs. And there is no PR's docs thing.
> 
> This is the [official doc](https://skywalking.apache.org/docs/#SkyWalking), which is built from [here](https://github.com/apache/skywalking/blob/master/docs/README.md) from the latest to every selected version.
> 
> What you wrote here is called the content or description of this PR. The purpose of this is to introduce what you changed and why you changed in the way. You should write what you had written here, that is good practice. Just it isn't being called a doc.

Thank you for your correction！

oap-server/server-starter/src/main/resources/ui-initialized-templates/general/general-instance.json
wu-sheng: According to UI side PR, this should be udpated.

oap-server/server-starter/src/main/resources/meter-analyzer-config/golang.yaml
easonyipj: instance_golang_stack represents the bytes in stack spans right now
instance_golang_gc_time represents  the most recent gc pause time
those are MeterSingleValue?

wu-sheng: My point is about GC time. How a metric could be called recent? 
What is the time duration about the recent?

easonyipj: > 
time duration is decided by user input, if user input is 15 seconds, than every 15 seconds will collect the latest gc's puase time using

wu-sheng: Then, what is your set? As the GC time is splitted, you should use sum function to merge, and use sum downsampling for hour and day level metric aggregation.

wu-sheng: The default downsampling is an avg, check hour level GC time you set now, you would see a hour GC time is close to a minute GC time.

easonyipj: OK, Should I change to "instance_golang_gc_time.downsampling(SUM)" ?

easonyipj: OK, thanks, I will test this method

oap-server/server-starter/src/main/resources/application.yml
```diff
@@ -234,7 +234,7 @@ agent-analyzer:
     # Nginx and Envoy agents can't get the real remote address.
     # Exit spans with the component in the list would not generate the client-side instance relation metrics.
     noUpstreamRealAddressAgents: ${SW_NO_UPSTREAM_REAL_ADDRESS:6000,9000}
-    meterAnalyzerActiveFiles: ${SW_METER_ANALYZER_ACTIVE_FILES:datasource,threadpool,satellite} # Which files could be meter analyzed, files split by ","
+    meterAnalyzerActiveFiles: ${SW_METER_ANALYZER_ACTIVE_FILES:datasource,threadpool,satellite,golang} # Which files could be meter analyzed, files split by ","
```
wu-sheng, line 237:
```suggestion
    meterAnalyzerActiveFiles: ${SW_METER_ANALYZER_ACTIVE_FILES:datasource,threadpool,satellite,go-runtime} # Which files could be meter analyzed, files split by ","
```

Please update the MAL file accordingly.

docs/en/changes/changes.md
```diff
@@ -28,6 +28,7 @@
   we should use `Get State and Start at Most Recent` semantic instead of `Start at Exact`
   because we don't need the changing history events, see https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-watch.
 * Unify query services and DAOs codes time range condition to `Duration`.
+* Support Golang VM metrics analysis.
```
wu-sheng, line 31:
```suggestion
* Support Golang runtime metrics analysis.
```

oap-server/server-starter/src/main/resources/application.yml
easonyipj: Ok, done.

wu-sheng:
Pending your metrics verification. https://github.com/apache/skywalking/pull/9711 should get approval first.

oap-server/server-starter/src/main/resources/meter-analyzer-config/go-runtime.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - name: instance_golang_gc_count
+    exp: instance_golang_gc_count.increase('PT1M')
+  - name: instance_golang_thread_num
+    exp: instance_golang_thread_num
+  - name: instance_golang_goroutine_num
+    exp: instance_golang_goroutine_num
+  - name: instance_golang_cpu_used_rate
+    exp: instance_golang_cpu_used_rate
+  - name: instance_golang_mem_used_rate
+    exp: instance_golang_mem_used_rate
```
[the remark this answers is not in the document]
easonyipj: Actually, these are host meters, I will change these.

oap-server/server-starter/src/main/resources/meter-analyzer-config/go-runtime.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+expSuffix: instance(['service'], ['instance'], Layer.GENERAL)
+metricPrefix: meter
+metricsRules:
+  - name: instance_golang_heap
+    exp: instance_golang_heap
+  - name: instance_golang_stack
+    exp: instance_golang_stack
+  - name: instance_golang_gc_time
+    exp: instance_golang_gc_time.downsampling(SUM)
```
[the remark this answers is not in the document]
easonyipj: I had tested, and the new instance_golang_gc_time  meter in UI represents the sum of all gc puase time that collected in a one minute. For example, if the meter interval is 15s, then go2sky will collect gc_time four times in a minute, and UI will show the sum of four collected gc_pause_time, but gc will happened more than 4 time in one minute(around 500 times one minute),
so these sum  of incomplete gc pause time may confuse users?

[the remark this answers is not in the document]
wu-sheng: What do you mean? I am not following. The sum of time cost of GC should be reasonable, no matter how many time you collected and reported.

oap-server/server-starter/src/main/resources/ui-initialized-templates/general/general-instance.json
```diff
@@ -1012,6 +1012,230 @@
...
+                },
+                {
+                  "x": 0,
+                  "y": 13,
+                  "w": 8,
+                  "h": 13,
+                  "i": "18",
+                  "type": "Widget",
+                  "widget": {
+                    "title": "Golang GC Count"
```
wu-sheng, line 1120:
```suggestion
                    "title": "Golang GC Count Per Minute"
```

This seems actually the real meaning of this metric if you report this per minute.
In hour/day level, the gc_count(increase) would still get `avg` downsampling. We always get a `count/min`

wu-sheng: Report period matters in the MAL analysis case, as you are reporting statistical data. So, I recommend you either report this per minute or not reported it.

oap-server/server-starter/src/main/resources/meter-analyzer-config/go-runtime.yaml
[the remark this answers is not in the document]
easonyipj: I mean the ui only shows the sum of time cost of **collected** GC (only four times in one minute if the interval is 15s) not the **all** GC(around 500 times in one minute) in one minute, so if ui shows the gc_cost_time is 0.5ms at 20:01, actually, the total gc_pause_time would more than 0.5ms at 20:01.  I just confuse about that, thanks for helping me.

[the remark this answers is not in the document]
wu-sheng: Which type actually this metric is? Is it a counter or gauge? If it is a gauge, then it should always report value from last time, the report period should not matter.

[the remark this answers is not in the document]
easonyipj: The current collection method is a gauge, Because it just collect the latest single gc's pause time, I think I should change the collection method to collect the total gc_pause_time in the collect period, thanks.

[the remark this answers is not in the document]
wu-sheng: Yes, so, you could check whether there is a hook/callback to measure this. Reading GC pause time periodly seems a back practice.

oap-server/server-starter/src/main/resources/meter-analyzer-config/go-runtime.yaml
```diff
@@ -0,0 +1,34 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+expSuffix: instance(['service'], ['instance'], Layer.GENERAL)
+metricPrefix: meter
+metricsRules:
+  - name: instance_golang_heap
```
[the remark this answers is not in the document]
easonyipj: > I think we should add an e2e to test these metrics?

Ok，should the e2e be put in go2sky?

oap-server/server-starter/src/main/resources/meter-analyzer-config/go-runtime.yaml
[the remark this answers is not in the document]
easonyipj: I use stats.PauseTotal to collect the total gc time which represents the sum of gc time after the instance started. So it is a counter meter. And I use increase('PT1M') to calculate the sum of one minute. So the metrics is instance_golang_gc_time.increase('PT1M'). I have tested that and it worked, is this reasonable?

oap-server/server-starter/src/main/resources/meter-analyzer-config/go-runtime.yaml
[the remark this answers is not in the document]
wankai123: yes, you can put it in go2sky.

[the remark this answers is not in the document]
easonyipj: > yes, you can put it in go2sky.

OK

oap-server/server-starter/src/main/resources/meter-analyzer-config/go-runtime.yaml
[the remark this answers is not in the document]
wu-sheng: Yes, then this is the GC pause time per minute metric.

Also, please try to re-word all the metrics names, they should be more accurate, formal and clear.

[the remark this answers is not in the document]
easonyipj: > Yes, then this is the GC pause time per minute metric.
> 
> Also, please try to re-word all the metrics names, they should be more accurate, formal and clear.

Ok, thanks, I will do this work.

oap-server/server-starter/src/main/resources/meter-analyzer-config/go-runtime.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - name: instance_golang_gc_count
+    exp: instance_golang_gc_count.increase('PT1M')
+  - name: instance_golang_thread_num
+    exp: instance_golang_thread_num
+  - name: instance_golang_goroutine_num
+    exp: instance_golang_goroutine_num
+  - name: instance_cpu_used_rate
+    exp: instance_cpu_used_rate
+  - name: instance_mem_used_rate
+    exp: instance_mem_used_rate
```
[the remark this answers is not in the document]
easonyipj: Yes, I am doing some changes on go2sky, and I will change these metrics names later, and pushed them after tested.

oap-server/server-starter/src/main/resources/meter-analyzer-config/go-runtime.yaml
```diff
@@ -16,19 +16,19 @@
...
-    exp: instance_golang_goroutine_num
-  - name: instance_cpu_used_rate
-    exp: instance_cpu_used_rate
-  - name: instance_mem_used_rate
-    exp: instance_mem_used_rate
\ No newline at end of file
+  - name: instance_golang_os_threads_count
+    exp: instance_golang_os_threads_count
+  - name: instance_golang_live_goroutines_count
+    exp: instance_golang_live_goroutines_count
```
wu-sheng, line 30: I think `thread_num` and `goroutine_num` are good, count usually indicates it is a counter to do `rate` or `increase`.

easonyipj: OK, i will change these to _num_

wu-sheng:
Also, you need to sync with the master branch, and resolve conflicts.(just a changelog doc)

wu-sheng:
I think this PR generally is good, we just wait for your agent side ready.

easonyipj:
> I think this PR generally is good, we just wait for your agent side ready.

Ok, thanks for your guidance, I will add e2e test on go2sky, and then test it before the agent side ready

wu-sheng:
You have to test manually as if this PR doesn't get merged, you can't find the image to test.

@arugal Do we use agent test tool in go2sky repo?

arugal:
> 

Yes, just like our plugin e2e test https://github.com/SkyAPM/go2sky/blob/master/plugins/http/test/docker-compose.yml

wu-sheng:
OK, @easonyipj then agent side could be tested without backend OAP in the e2e actually.

The server side would not be able to test.

@arugal Do you prefer to add e2e to skywalking main repo, which could be a little too heavy? Or we should have a real OAP placed e2e for this case? Or we could leave backend OAP manually tested only?

easonyipj:
> OK, @easonyipj then agent side could be tested without backend OAP in the e2e actually.
> 
> The server side would not be able to test.
> 
> @arugal Do you prefer to add e2e to skywalking main repo, which could be a little too heavy? Or we should have a real OAP placed e2e for this case? Or we could leave backend OAP manually tested only?

OK, I will do this on agent side only. Thanks for you all.

wu-sheng:
@easonyipj Let me explain a little more about this

> Yes, just like our plugin e2e test https://github.com/SkyAPM/go2sky/blob/master/plugins/http/test/docker-compose.yml

The docker compose is using `ghcr.io/apache/skywalking-agent-test-tool/mock-collector` as the backend rather than OAP server to test. As a result, all MAL scripts are not going to be tested if you ref this case to test.

Mock collector focuses on testing the data agent reported, whether it matches the expectation, but not how they would be analyzed.

easonyipj:
> 

OK, thanks very much, I will learn how to use these to build an e2e test.

arugal:
@wu-sheng https://github.com/SkyAPM/go2sky/pull/168 is merged

easonyipj:
> @wu-sheng [SkyAPM/go2sky#168](https://github.com/SkyAPM/go2sky/pull/168) is merged
> 
> Hi @easonyipj After this pr merged, please use oap as the server to add e2e test in go2sky

OK, got it, thanks

wu-sheng:
Thanks for the update, so, this PR would still be hold still e2e added, right?

arugal:
> Thanks for the update, so, this PR would still be hold still e2e added, right?

need to add the e2e test in the main repo？

wu-sheng:
No, we don't require that. I mean should I need the go2sky side e2e added? If you have tested this and feel OK to merge, I am good.

arugal:
> If you have tested this and feel OK to merge, I am good.

Yes, we encountered some problems when using[ skywalk-agent-test-tool](https://github.com/apache/skywalking-agent-test-tool) for e2e test, we will add go2sky e2e testing after this pr merged

wu-sheng (approved):

wu-sheng:
Merged.

PR #9735, opened 2022-10-07
Support skywalking-client-js report empty serviceVersion and `pagePath

### LINKED ISSUE #9734: [Bug] Malicious / misconfigured client can cause inaccessible trace view
### Search before asking

UI (apache/skywalking-booster-ui)

### What happened

When opening the "browser" dashboard:

```
Exception while fetching data (/browser_app_page_pv0) : Can't split endpoint id into 2 parts, Base64AppName.1_
```

The Trace-View cannot be used any more.

The trace-View should not be unusable, an empty page path should be handled correctly.

### How to reproduce

Insert a trace that has an empty pagePath.

Instantiate a JS client with:

```
const getCollectorProps = () => ({
  collector: 'https://your-collector/'
  service: 'xxx',
  serviceVersion: '1.0',
  pagePath: '' // < this is the important part
});
```

The endpoint to add this trace is usually not protected by authentication (would not make sense, since then clients could not send their traces). So any JS page using skywalking can be affected by this, since requests can be crafted by attackers.

### Anything else

This problem is reproducible easily. Version 9.2.0. The Skywalking team does not consider this a security issue.

#9269 has a similar error message, but does not seem to be related.

### Code of Conduct

wu-sheng:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #9734.

cc @Fine0830

### DIFF
--- a/oap-server/server-receiver-plugin/skywalking-browser-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/browser/provider/parser/performance/BrowserPerfDataDecorator.java
+++ b/oap-server/server-receiver-plugin/skywalking-browser-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/browser/provider/parser/performance/BrowserPerfDataDecorator.java
@@ -17,98 +17,96 @@
 
 package org.apache.skywalking.oap.server.receiver.browser.provider.parser.performance;
 
-import lombok.RequiredArgsConstructor;
 import org.apache.skywalking.apm.network.language.agent.v3.BrowserPerfData;
 
-@RequiredArgsConstructor
 public class BrowserPerfDataDecorator {
-    private boolean isOrigin = true;
-    private final BrowserPerfData browserPerfData;
     private BrowserPerfData.Builder builder;
 
+    public BrowserPerfDataDecorator(final BrowserPerfData browserPerfData) {
+        this.builder = browserPerfData.toBuilder();
+    }
+
     public String getService() {
-        return isOrigin ? browserPerfData.getService() : builder.getService();
+        return builder.getService();
     }
 
     public String getServiceVersion() {
-        return isOrigin ? browserPerfData.getServiceVersion() : builder.getServiceVersion();
+        return builder.getServiceVersion();
     }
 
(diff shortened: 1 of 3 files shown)


arugal (approved):

oap-server/server-receiver-plugin/skywalking-browser-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/browser/provider/parser/performance/BrowserPerfDataDecorator.java
```diff
@@ -17,98 +17,96 @@
 
 package org.apache.skywalking.oap.server.receiver.browser.provider.parser.performance;
 
-import lombok.RequiredArgsConstructor;
 import org.apache.skywalking.apm.network.language.agent.v3.BrowserPerfData;
 
-@RequiredArgsConstructor
 public class BrowserPerfDataDecorator {
```
kezhenxu94, line 22: `BrowserPerfDataDecorator` is redundant now after your changes in this PR

wu-sheng: Do you ask to change all codes to use builder directly? The builder is always built even before because of `decorator.setTime(nowMillis);` called always.

wu-sheng: https://github.com/apache/skywalking/blob/8aa046b8d722f0dc40dd6afdf3ce8f1b67894717/oap-server/server-receiver-plugin/skywalking-browser-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/browser/provider/parser/performance/PerfDataAnalyzer.java#L41-L44

kezhenxu94 (approved):

Issue #9000, opened 2022-05-06
[Feature] Set TagController for tags of traces and logs


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Tag-based log/trace search and auto complete(https://github.com/apache/skywalking/issues/8933) were/are being added.
One thing should be added to limit the max length of tag's `key=value` to avoid candidate overload and/or over column's length.

Trace and Log records, with their auto-complete should follow this new controller mechanism.

### Use case

This is not urgent, once after Zipkin alternative(https://github.com/apache/skywalking/issues/8932) finished, this should be done, as Zipkin span's tags/annotations could be searched, which are very long.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #9741, opened 2022-10-08
Limit the max length of trace and log tag's `key=value`

wankai123:
- [X] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/9000.


wu-sheng (approved):

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/SegmentAnalysisListener.java
```diff
@@ -152,6 +152,12 @@ private void appendSearchableTags(SpanObject span) {
         span.getTagsList().forEach(tag -> {
             if (searchableTagKeys.contains(tag.getKey())) {
                 final Tag spanTag = new Tag(tag.getKey(), tag.getValue());
+                if (tag.getValue().length()  > Tag.TAG_LENGTH || spanTag.toString().length() > Tag.TAG_LENGTH) {
+                    if (log.isDebugEnabled()) {
+                        log.debug("Segment tag : {} length > : {}, dropped", spanTag, Tag.TAG_LENGTH);
```
kezhenxu94, line 157: I think we should `warn` this, if users can't find search results for those tags, they should be notified that those are dropped instead of thinking there is no such tag at all, otherwise it's hard to diagnose why there is no result in the UI.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/RecordSinkListener.java
```diff
@@ -53,6 +55,7 @@
  */
 @RequiredArgsConstructor
 public class RecordSinkListener implements LogSinkListener {
+    private static final Logger LOGGER = LoggerFactory.getLogger(RecordSinkListener.class);
```
kezhenxu94, line 58: Prefer `@Slf4j` annotation

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/SegmentAnalysisListener.java
wu-sheng: Warning logs seem a little tricky. Because this kind of log would not disappear unless the user changes business codes or plugin codes. And the warning would not stop and then they would crash the log collecting system. 
I think that is not what we expected.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/provider/log/listener/RecordSinkListener.java
wankai123: Class field `private final Log log = new Log()` conflicts with the `@Slf4j` generated field name `log`, should I change all `log` to another name?

kezhenxu94: > Class field `private final Log log = new Log()` conflicts with the `@Slf4j` generated field name `log`, should I change all `log` to another name?

OK forgot this, let's keep this now

kezhenxu94 (approved):

Issue #9742, opened 2022-10-09
[Bug] In this case, profiling tasks should not be created


xzyJavaX:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar issues.

### Apache SkyWalking Component

OAP server (apache/skywalking)

### What happened

During the creation of profiling task, a field called time_bucket will be generated,time_bucket is converted from the end time of task monitoring.For example, for Task A, the monitoring start time is 12:00:00, the monitoring time is 5 minutes, and the monitoring end time is 12:05:00, time_bucket is 12:05.Then create Task B, with the start time of 12:04:00 and the end time of 12:09:00, will go to the database to check whether there is a time_ Buckets in this range. If yes, throw "current service already has monitor task execute at this time".For example, start is 12:04:00, end is 12.09:00, and there is a time_bucket is 12:05, so Task B failed to be created.
But if you create a task with a monitoring time of 15 minutes at 12:00:00,  time_bucket is 12:15. At this time, a task with start is 12:04:00 and end is 12:09:00 can be created successfully. This should not be as expected.
There is also a UI problem. When the "current service already has monitor task execute at this time" is returned, the pop-up window will show that the creation was successful, but it was not actually created.
![518a51ea-b99f-4c57-af35-6af68dd0657b](https://user-images.githubusercontent.com/49760141/194735415-6d71be0d-b0b6-4b28-85b6-2e99cc94f626.png)

### What you expected to happen

In this case, profiling tasks should not be created

### How to reproduce

First create a profiling task with a monitoring duration of 15 minutes, and then create a profiling task with a monitoring duration of 5 minutes

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wu-sheng:
This seems a bug. Are you going to submit a pull request to fix?

xzyJavaX:
Yes I am willing to submit a PR

wu-sheng:
Assigned to you. cc owner @mrproliu

PR #9744, opened 2022-10-09
optimize the creation conditions of profiling task

xzyJavaX:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #[9742](https://github.com/apache/skywalking/issues/9742)

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/trace/ProfileTaskMutationService.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/trace/ProfileTaskMutationService.java
@@ -139,15 +139,17 @@ public class ProfileTaskMutationService implements Service {
         }
 
         // Each service can monitor up to 1 endpoints during the execution of tasks
-        long startTimeBucket = TimeBucket.getMinuteTimeBucket(monitorStartTime);
         long endTimeBucket = TimeBucket.getMinuteTimeBucket(monitorEndTime);
         final List<ProfileTask> alreadyHaveTaskList = getProfileTaskDAO().getTaskList(
-            serviceId, null, startTimeBucket, endTimeBucket, 1);
+            serviceId, null, null, endTimeBucket, 1);
         if (CollectionUtils.isNotEmpty(alreadyHaveTaskList)) {
-            // if any task time bucket in this range, means already have task, because time bucket is base on task end time
-            return "current service already has monitor task execute at this time";
+            for (ProfileTask profileTask : alreadyHaveTaskList) {
+                if (profileTask.getStartTime() + TimeUnit.MINUTES.toMillis(profileTask.getDuration()) >= monitorStartTime) {
+                    // if the endTime is greater or equal than the startTime of the newly created task, i.e. there is overlap between two tasks, it is an invalid case
+                    return "current service already has monitor task execute at this time";
+                }
+            }
         }
-
         return null;
     }
 
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -49,6 +49,7 @@
 * [**Breaking Change**] Limit the max length of trace/log/alarm tag's `key=value`, set the max length of column `tags`
   in tables`log_tag/segment_tag/alarm_record_tag` and column `query` in `zipkin_query` and column `tag_value` in `tag_autocomplete` to 256.
   SQL-Database requires altering these columns' length or removing these tables before OAP starts, if bump up from previous releases.
+* Optimize the creation conditions of profiling task
 
 #### UI
 


wu-sheng:
Could you explain why removing the end-time condition helps?

mrproliu:
Thanks for your contribution, I have two questions:
1. I don't think deleting the condition is a good idea, because we have other queries that need to use the end time bucket filter(such as cache timer). If you don't need the end time bucket filter, just set it as `null`.
2. If you ignore the end time bucket when creating a task, then the profiling will generally be created successfully(if the start time not be set), It's not right.

wu-sheng:
I think we should build a strict condition to avoid overlap.

mrproliu:
I think a good way to avoid overlap need following these steps:
1. find the recent profiling task from DB. An example SQL: `startBucket <= ${creating task end time}`.
2. Check the creating task start/end time is in the recent profiling task monitoring range.

xzyJavaX:
I think this should be feasible:
Continue to judge after the alreadyHaveTaskList is found,if alreadyHaveTaskList isEmpty,continue to query DB,query condition is 
query.must(Query.range(ProfileTaskRecord.START_TIME).lte(endTimeBucket));
query.must(Query.range(ProfileTaskRecord.TIME_BUCKET).gte(endTimeBucket));
If the data can be found, it indicates that an overlay has occurred, and the task should not be created at this time.

wu-sheng:
> Continue to judge after the alreadyHaveTaskList is found,if alreadyHaveTaskList isEmpty

This seems wrong from a grammar perspective. If a method named `alreadyHaveTaskList` return null, it should have no further query, right?

wu-sheng:
I think you could `getProfileTaskDAO().getTaskList()`  to get a basic list, which should not be long, and just filter it in the memory.

mrproliu:
> query.must(Query.range(ProfileTaskRecord.TIME_BUCKET).gte(endTimeBucket));

The `TIME_BUCKET` is usually the same as `START_TIME `, we should ignore it. 

> If the data can be found, it indicates that an overlay has occurred, and the task should not be created at this time.

You need to recheck again. If the previous task is already finished, the creation process should continue.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/trace/ProfileTaskMutationService.java
```diff
@@ -140,14 +140,16 @@ private String checkDataSuccess(final String serviceId,
...
-        long endTimeBucket = TimeBucket.getMinuteTimeBucket(monitorEndTime);
         final List<ProfileTask> alreadyHaveTaskList = getProfileTaskDAO().getTaskList(
-            serviceId, null, startTimeBucket, endTimeBucket, 1);
+            serviceId, null, startTimeBucket, null, null);
         if (CollectionUtils.isNotEmpty(alreadyHaveTaskList)) {
-            // if any task time bucket in this range, means already have task, because time bucket is base on task end time
-            return "current service already has monitor task execute at this time";
+            for (ProfileTask profileTask : alreadyHaveTaskList) {
+                if (profileTask.getStartTime() <= monitorEndTime) {
+                    // if any task time bucket in this range, means already have task, because time bucket is base on task end time
```
lujiajing1126, line 148:
```suggestion
                    // if the startTime is less or equal than the endTime of the newly created task, i.e. there is overlap between two tasks, it is an invalid case
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/trace/ProfileTaskMutationService.java
```diff
@@ -140,14 +140,16 @@ private String checkDataSuccess(final String serviceId,
 
         // Each service can monitor up to 1 endpoints during the execution of tasks
         long startTimeBucket = TimeBucket.getMinuteTimeBucket(monitorStartTime);
-        long endTimeBucket = TimeBucket.getMinuteTimeBucket(monitorEndTime);
         final List<ProfileTask> alreadyHaveTaskList = getProfileTaskDAO().getTaskList(
-            serviceId, null, startTimeBucket, endTimeBucket, 1);
+            serviceId, null, startTimeBucket, null, null);
```
mrproliu, line 144: 1. If you use the start time to filter, then you may not get any data. For example, you create a 10 min task at 10:00, then when you create a new task at 10:05, then you cannot get the existing task. Am I right?
2. You should limit the profiling task count.

xzyJavaX: sorry, I used to think of TIME_BUCKE as END_TIME.
What do you think the "limit" should be set to.

xzyJavaX: sorry, I used to think of TIME_BUCKE as END_TIME.
What do you think the "limit" should be set to.

mrproliu: I think `1` is enough. Usually, you just need to check if the most recent task has overlap or not.

mrproliu (approved):
Thanks.

wu-sheng:
3 ES relative profiling e2e fail. I just rerun to see whether they could pass, or there are some issues of this change.

PR #9747, opened 2022-10-09
Lazy load the Kubernetes metadata and switch from event-driven to polling

kezhenxu94:
This is for ALS k8s analyzer.

Previously we set up watchers to watch the Kubernetes metadata changes, this is perfect when there are deployments changes and SkyWalking can react to the changes in real time. However when the cluster has many events (such as in large cluster or some special Kubernetes engine like OpenShift), SkyWalking might send massive requests to Kubernetes API server, causing heavy load to the API server.

This PR switches from the watcher mechanism to polling mechanism, SkyWalking polls the metadata in a specified interval, so that the requests sent to API server is predictable (~10 requests every `interval`), and the requests count is constant regardless of the cluster's changes. However with this change SkyWalking can't react to the cluster changes in time, but the delay is acceptable in our case.


wu-sheng:
Let's mention this in the Changelog when this is ready.

oap-server/server-library/library-kubernetes-support/src/main/java/org/apache/skywalking/library/kubernetes/KubernetesServices.java
```diff
@@ -0,0 +1,66 @@
...
+import io.kubernetes.client.openapi.models.V1Service;
+import lombok.SneakyThrows;
+
+public enum KubernetesServices {
+    INSTANCE;
+
+    private final LoadingCache<KubernetesServices, List<V1Service>> services;
+
+    @SneakyThrows
+    private KubernetesServices() {
```
wu-sheng, line 40: As a constructor of an enum, how do we control this initialization? I think this should not run if we are not in a k8s env.
The initialization should only run when it is used in the first time.

wu-sheng: Or whether we could check whether there is an API server to fetch metadata? If not, close this automatically.
I just have concerns about the error log rolling endlessly.

kezhenxu94: > As a constructor of an enum, how do we control this initialization? I think this should not run if we are not in a k8s env.

We don't because we don't need.

> The initialization should only run when it is used in the first time.

The enum being instantiated doesn't mean it will fetch the metadata immediately, it fetches the metadata ONLY when the `list` method is called, and the `list` method is only called in `k8s-mesh` analyzer, or Kubernetes metrics comes to MAL.

In short, SkyWalking only tries to fetch the metadata when requested, if it throws error, it means someone tries to fetch the metadata in the wrong place (i.e. open `k8s-mesh` analyzer in non-Kubernetes cluster)

wu-sheng: Oh, I didn't realize this behavior of the cache loader.

kezhenxu94:
Added the change log by simply copying the PR comment and title, please take a look

docs/en/changes/changes.md
```diff
@@ -50,7 +50,17 @@
 * [**Breaking Change**] Limit the max length of trace/log/alarm tag's `key=value`, set the max length of column `tags`
   in tables`log_tag/segment_tag/alarm_record_tag` and column `query` in `zipkin_query` and column `tag_value` in `tag_autocomplete` to 256.
   SQL-Database requires altering these columns' length or removing these tables before OAP starts, if bump up from previous releases.
-* Optimize the creation conditions of profiling task
+* Optimize the creation conditions of profiling task.
+* Lazy load the Kubernetes metadata and switch from event-driven to polling.
+  This is for ALS k8s analyzer.
```
wu-sheng, line 55:
```suggestion
```

Right now, this is for many things, right?

PR #9789, opened 2022-10-15
Fix metrics was put into wrong slot of the window in the alarting kernel

### LINKED ISSUE #9788: [Bug] Alarm Rule Warn is not  accurate
### Search before asking

### What happened

when i use the Alarm Rule to monitor  metrics,i find the alarm rule is not accurately.

hope zhe alarm rule is  accurately.

### How to reproduce

all 

### Anything else

when i find the alarm rule is not accurately, i read the source code and add logs in org.apache.skywalking.oap.server.core.alarm.provider.RunningRule.Window#isMatch like this
image

 i find the same bucket metrics in the windows.

so i analysis  the code of org.apache.skywalking.oap.server.core.alarm.provider.RunningRule.Window#add,
image

the endtime in this code  will have seconds data, such as:20221014174301,Assuming that：

endtime  is 20221014174301
two metrics  bucket  is    [202210141743, 202210141744]
the  code minutes will  all return   0, and set the tow metrics  in the same index.
```
 int minutes = Minutes.minutesBetween(timeBucket, this.endTime).getMinutes();
 this.values.set(values.size() - minutes - 1, metrics);
```
so  it will cause the windows data error.

### Code of Conduct

yangyiweigege:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #9788.


docs/en/changes/changes.md
```diff
@@ -61,6 +61,7 @@
   regardless of the cluster's changes. However with this change SkyWalking can't react to the cluster changes in time, but the delay
   is acceptable in our case.
 * Optimize the query time of tasks in ProfileTaskCache.
+* Fix the windows in Alarm Rule hava wrong data,such as have same bucket metrics.
```
wu-sheng, line 64:
```suggestion
* Fix metrics was put into wrong slot of the window in the alarting kernel.
```

wu-sheng (approved):

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/RunningRule.java
```diff
@@ -222,7 +222,9 @@ private boolean validate(String target, List<String> includeList, List<String> e
      * @param targetTime of moving target
      */
     public void moveTo(LocalDateTime targetTime) {
-        windows.values().forEach(window -> window.moveTo(targetTime));
+        //format targetTime to minute ,make sure the second is `00` and millisSeconds is `00` such as: 18:30:00.000
```
kezhenxu94, line 225:
```suggestion
        // Truncate targetTime to minute, make sure the second is `00` and milliseconds is `00` such as: 18:30:00.000
```

kezhenxu94 (approved):
A nit

kezhenxu94 (approved):

PR #9795, opened 2022-10-17
Improve Kubernetes Pod lister with field selector

kezhenxu94:
I think I had been trapped in the thoughts that we need to list all resources (Pods, services, etc.) and recompose them in memory, recently when I am reviewing the codes, I found that we can simply list/fetch the exact resource by just namespace/name.

In this patch I improve the logic to fetch the Pod by IP with field selector, this prevents us to listing all Pods and then filter in memory.

The usage of other resources can not be improved that much as they don't have suitable filed selector supported for now.

### DIFF
--- a/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/k8s/K8sInfoRegistry.java
+++ b/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/k8s/K8sInfoRegistry.java
@@ -26,34 +26,31 @@ import java.util.Objects;
 import java.util.Optional;
 import org.apache.skywalking.library.kubernetes.KubernetesPods;
 import org.apache.skywalking.library.kubernetes.KubernetesServices;
+import org.apache.skywalking.library.kubernetes.ObjectID;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
-import io.kubernetes.client.openapi.models.V1ObjectMeta;
 import io.kubernetes.client.openapi.models.V1Pod;
 import io.kubernetes.client.openapi.models.V1Service;
 import lombok.SneakyThrows;
 
 public class K8sInfoRegistry {
-
     private final static K8sInfoRegistry INSTANCE = new K8sInfoRegistry();
-    private final LoadingCache<String/* podName.namespace */, String /* serviceName.namespace */> podServiceMap;
-    private final LoadingCache<String/* podIP */, String /* podName.namespace */> ipPodMap;
-    private final LoadingCache<String/* serviceIP */, String /* serviceName.namespace */> ipServiceMap;
-    private static final String SEPARATOR = ".";
+    private final LoadingCache<ObjectID /* Pod */, ObjectID /* Service */> podServiceMap;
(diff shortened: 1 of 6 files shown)


wu-sheng:
Filtering pods through IPs mean we have to wait for traffic of pods, and wait for next round fetching?
I think this would have pro and con effects.

kezhenxu94:
> Filtering pods through IPs mean we have to wait for traffic of pods, and wait for next round fetching?
> I think this would have pro and con effects.

Not the next round fetching. It will fetch (and wait) when the first traffic request comes.

wu-sheng:
So, we must lose the first seconds  data, even the pod is up for a while. Considering business booting  takes 20-30s at least, we nearly don't lose data before.

kezhenxu94:
> So, we must lose the first seconds  data, even the pod is up for a while. Considering business booting  takes 20-30s at least, we nearly don't lose data before.

Sounds like you miss something. Even if the app container takes some time to boot, users should have configured healthiness check and readiness check so if the app containers is not ready no requests will be routed to the pod

wu-sheng:
> > So, we must lose the first seconds  data, even the pod is up for a while. Considering business booting  takes 20-30s at least, we nearly don't lose data before.
> 
> Sounds like you miss something. Even if the app container takes some time to boot, users should have configured healthiness check and readiness check so if the app containers is not ready no requests will be routed to the pod

My question is mainly about when pod1(IP-A) boots and is standing by, but no request, would OAP get the IP-pod-deployment mapping?

kezhenxu94:
> > > So, we must lose the first seconds  data, even the pod is up for a while. Considering business booting  takes 20-30s at least, we nearly don't lose data before.
> > 
> > 
> > Sounds like you miss something. Even if the app container takes some time to boot, users should have configured healthiness check and readiness check so if the app containers is not ready no requests will be routed to the pod
> 
> My question is mainly about when pod1(IP-A) boots and is standing by, but no request, would OAP get the IP-pod-deployment mapping?

No. this case is what I want to improved, it’s unnecessary to fetch this Pod mapping right?

wu-sheng:
> > > > So, we must lose the first seconds  data, even the pod is up for a while. Considering business booting  takes 20-30s at least, we nearly don't lose data before.
> > > 
> > > 
> > > Sounds like you miss something. Even if the app container takes some time to boot, users should have configured healthiness check and readiness check so if the app containers is not ready no requests will be routed to the pod
> > 
> > 
> > My question is mainly about when pod1(IP-A) boots and is standing by, but no request, would OAP get the IP-pod-deployment mapping?
> 
> No. this case is what I want to improved, it’s unnecessary to fetch this Pod mapping right?

That is why I mentioned pro and con.  The pro is, you could read less data, the con is the first few traffic has to be missed.

I don't have a preference for now, but I suppose this is worth thinking.

kezhenxu94:
> That is why I mentioned pro and con. The pro is, you could read less data, the con is the first few traffic has to be missed.
> 

Hi. When the first traffic request access log comes to skywalking, skywalking reads the Pods by IP and do the analysis, why do you say the first few traffic will be missed? Did I miss anything?

wu-sheng:
I thought you would read pod IP in async way.

kezhenxu94:
> I thought you would read pod IP in async way.

No, not read the Pods asynchronously, just like before: lazy load the Pod info only when necessary, but with field selector to only fetch needed Pod.

wu-sheng:
Got it, make sense.

wu-sheng (approved):

PR #9799, opened 2022-10-17
Bump up nacos to 2.1.2

kezhenxu94:


kezhenxu94:
This is ridiculous, upgrading the nacos version causes the packaging timeout without any reason, in CI, in my local, and in my codespace machine.

wu-sheng:
The new pom is confusing the maven?

kezhenxu94:
> The new pom is confusing the maven?

I have no idea because the maven command just gets stuck without any useful message, even I turn on maven debug logs.

wu-sheng:
😂😂 I never used this kind of pom file. Let's feedback to Nacos team, this at least doesn't work for us.

Issue #9802, opened 2022-10-18
[Feature] Support HTTP metrics and req/resp collecting in Network profiling


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

After 9.2.0, we are doing research to analysis more information from network through ebpf. 
The next move of us is going to decode HTTP req/resp with metrics of each HTTP 1.x request during the profiling process.
Then we would add more metrics for network profiling when we detect the HTTP protocol is being used automatically.

Rover agent would implement most of the network analysis, and statistics. And the OAP server could add those metrics on the dashboard.

### Related issues

A new query protocol is added, https://github.com/apache/skywalking-query-protocol/pull/100. This would replace the existing sampled record, and provide a more general record query to cover existing sampled record query and more other collected records, including the HTTP req/resp mentioned in this new feature.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #9804, opened 2022-10-18
Support `sumPerMinLabeled` in `MAL`

mrproliu:
### Support `sumPerMinLabeled` in `MAL`
When implementing the https://github.com/apache/skywalking/issues/9802, we can generate the metrics that calculate the count of response status code in minutes, so the `sumPerMinLabeled` needs to be added. 

### DIFF
--- a/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/Analyzer.java
+++ b/oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/Analyzer.java
@@ -34,6 +34,7 @@ import lombok.RequiredArgsConstructor;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.text.CaseUtils;
 import org.apache.skywalking.oap.meter.analyzer.dsl.DSL;
 import org.apache.skywalking.oap.meter.analyzer.dsl.DownsamplingType;
 import org.apache.skywalking.oap.meter.analyzer.dsl.Expression;
@@ -247,8 +248,8 @@ public class Analyzer {
     private void createMetric(final ScopeType scopeType,
                               final String dataType,
                               final DownsamplingType downsamplingType) {
-        String functionName = String.format(
-            "%s%s", downsamplingType.toString().toLowerCase(), StringUtils.capitalize(dataType));
+        String downSamplingStr = CaseUtils.toCamelCase(downsamplingType.toString().toLowerCase(), false, '_');
+        String functionName = String.format("%s%s", downSamplingStr, StringUtils.capitalize(dataType));
         meterSystem.create(metricName, functionName, scopeType);
     }
 
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/sumpermin/SumPerMinLabeledFunction.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/sumpermin/SumPerMinLabeledFunction.java
@@ -20,7 +20,6 @@ package org.apache.skywalking.oap.server.core.analysis.meter.function.sumpermin;
 
 import lombok.Getter;
 import lombok.Setter;
-import lombok.ToString;
 import org.apache.skywalking.oap.server.core.Const;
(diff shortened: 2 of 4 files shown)


oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/Analyzer.java
```diff
@@ -247,8 +248,8 @@ private void init() {
     private void createMetric(final ScopeType scopeType,
                               final String dataType,
                               final DownsamplingType downsamplingType) {
-        String functionName = String.format(
-            "%s%s", downsamplingType.toString().toLowerCase(), StringUtils.capitalize(dataType));
+        String downSamplingStr = CaseUtils.toCamelCase(downsamplingType.toString().toLowerCase(), false, '_');
+        String functionName = String.format("%s%s", downSamplingStr, StringUtils.capitalize(dataType));
```
wu-sheng, line 252: I think this changed the format. Could you explain why?

mrproliu: If we combine the downsampling(`SUM_PER_MIN`) and dataType(`labeled`), the function name would be `sum_per_minLabeled`. The format of the name is not consistent, so I change the downsampling to camel format. 
After the change, the name is `sumPerMinLabeled`.

wu-sheng: I am feeling this change impacts the existing function names? Most of e2e are failing.

mrproliu: It only affects the `SUM_PER_MIN` downsampling. The e2e are failing because I cancel it. The `Build Tar` CI is running timeout, so I want to re-run it.

mrproliu: I have clicked the `cancel workflow` button, but It's doesn't work. Maybe I need to wait a while.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/meter/function/sumpermin/SumPerMinLabeledFunction.java
```diff
@@ -0,0 +1,188 @@
...
+    @Getter
+    @Setter
+    @Column(columnName = VALUE, dataType = Column.ValueDataType.COMMON_VALUE, function = Function.Avg)
+    @ElasticSearch.Column(columnAlias = "datatable_value")
+    private DataTable value = new DataTable(30);
+
+    @Getter
+    @Setter
+    @Column(columnName = TOTAL, storageOnly = true)
+    @ElasticSearch.Column(columnAlias = "datatable_total")
```
wankai123, line 72: Since `columnAlias` marked as `@deprecate` and only used for the previous function column name.
I think we should better use the column name `datatable_total` directly rather than use alias. WDYT @wu-sheng

mrproliu: Do you mean we only need to use the `@Column(columnName)` to replace the `columnAlias`? We also have a lot of metrics using this annotation too.

wu-sheng: No, that attribute was added as we have merged all metric indices into one. So, during that process, we can't change the column name of index if the no-merging option is opened. That is why we have **A NEW ADDED BUT Deprecated** annotation.
For the new metric function(OAL and meter), we should avoid this column name conflicts from the beginning. And this attribute is not recommended to use. The column name across functions is still sharable once they don't have data type conflict.

mrproliu: Ok, I have updated the column name.

wu-sheng (approved):

wankai123 (approved):

Issue #9676, opened 2022-09-23
[Feature] Add virtual MQ analysis for native traces.


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

As we have the virtual database and cache analysis in 9.3.0, let's bring the last missing core analysis, virtual MQ, into the backend analysis.

As we designed years ago, span layer includes the following options

```java
public enum SpanLayer {
    DB(1), RPC_FRAMEWORK(2), HTTP(3), MQ(4), CACHE(5);
```

MQ span layer represents a queue server with consumer and producer sides, such as, Kafka, RocketMQ, and Pulsar. SkyWalking Java agent has had plugins for these typical and widely used MQ's Java clients for years too.

Now, let's analyze their access load(producing/consuming load), such as X messages per second, and other typical metrics
There is one special about time, MQ is able to consume and produce messages in bulk mode, and some queue clients are using pulling mode(typically Kafka), so the span could only include pulling time, but, meanwhile, some pushing mode client span could including all further processes' time.

Also, `endpoint_mq_consume_count` and `endpoint_mq_consume_latency` are purely analyzed through OAL, even having tag match expression, we should consider merging it in a consistent way like existing `VirtualServiceProcessor` implementations.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

pg-yang:
Please assign to me

pg-yang:
Don't OAL support labeled AVG metrics ? 

I'm  analyzing MQ metrics , I think it's better to provide transmission latency metrics labeled by `queue` or `topic` , as  following :
```
mq_transmission_latency = from(MessageQueueAccess.transmissionLatency).filter(operation == MessageQueueOperation.Consume).avgLabeled(MessageQueueAccess.queue);
```
But  `AvgLabeledFunction` is in meter package , and  annotated by `@MeterFunction`. 

`MetricsHolder` only initialize Class annotated by `@MetricsFunction`
https://github.com/apache/skywalking/blob/5abe6ceb1ff0938be917f8b440c5eb3590cf31db/oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/parser/MetricsHolder.java#L41-L45

[...]

PR #9813, opened 2022-10-19
labeled avg in OAL

pg-yang:


oap-server/oal-grammar/src/main/antlr4/org/apache/skywalking/oal/rt/grammar/OALParser.g4
```diff
@@ -87,7 +87,9 @@ functionName
 funcParamExpression
     : expression
     ;
-
+arrayFuncParamExpression
+    : LS_BRACKET ((funcParamExpression|literalExpression|attributeExpression) (COMMA (funcParamExpression|literalExpression|attributeExpression))*) RS_BRACKET
```
pg-yang, line 91: Support array argument  , such as  `labeledLongAvg([topic,queue,queue,"1"]);`

oap-server/server-starter/src/main/resources/oal/core.oal
```diff
@@ -15,74 +15,4 @@
...
-cache_write_sla = from(CacheAccess.*).filter(operation == VirtualCacheOperation.Write).percent(status == true);
-cache_write_cpm = from(CacheAccess.*).filter(operation == VirtualCacheOperation.Write).cpm();
-cache_write_percentile = from(CacheAccess.latency).filter(operation == VirtualCacheOperation.Write).percentile(10);
-
-cache_access_resp_time = from(CacheAccess.latency).longAvg();
-cache_access_sla = from(CacheAccess.*).percent(status == true);
-cache_access_cpm = from(CacheAccess.*).cpm();
-cache_access_percentile = from(CacheAccess.latency).percentile(10);
\ No newline at end of file
+mq_transmission_latency = from(MessageQueueAccess.transmissionLatency).filter(operation == MessageQueueOperation.Consume).labeledLongAvg([topic,queue,queue,"1"]);
```
pg-yang, line 18: The  demo  labeled avg  metrics

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/LabeledLongAvgMetrics.java
```diff
@@ -0,0 +1,95 @@
...
+    @ElasticSearch.Column(columnAlias = "datatable_count")
+    protected DataTable count = new DataTable(30);
+    @Getter
+    @Setter
+    @Column(columnName = VALUE, dataType = Column.ValueDataType.LABELED_VALUE, storageOnly = true)
+    @ElasticSearch.Column(columnAlias = "datatable_value")
+    private DataTable value = new DataTable(30);
+
+    @Entrance
+    public final void combine(@SourceFrom long value, @ConstOne long count, @Arg String[] labels) {
```
pg-yang, line 59: Entrance

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/parser/DeepAnalysis.java
```diff
@@ -81,9 +82,10 @@ public AnalysisResult analysis(AnalysisResult result) {
         EntryMethod entryMethod = new EntryMethod();
         result.setEntryMethod(entryMethod);
         entryMethod.setMethodName(entranceMethod.getName());
-
+        final Parameter[] parameters = entranceMethod.getParameters();
         // 4. Use parameter's annotation of entrance method to generate aggregation entrance.
-        for (Parameter parameter : entranceMethod.getParameters()) {
+        for (int i = 0; i < parameters.length; i++) {
+            Parameter parameter = parameters[i];
```
wu-sheng, line 88: Why change this?

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/OALRuntime.java
```diff
@@ -421,13 +421,13 @@ private Class generateDispatcherClass(String scopeName,
...
                     }
                 );
-
             dispatcherClass.setGenericSignature(dispatcherSignature.encode());
         } catch (NotFoundException e) {
             log.error("Can't find Dispatcher interface for " + className + ".", e);
             throw new OALCompileException(e.getMessage(), e);
         }
 
+
```
wu-sheng, line 430: This should be reverted.

oap-server/oal-grammar/src/main/antlr4/org/apache/skywalking/oal/rt/grammar/OALParser.g4
wu-sheng, line 91: The element of an array could be `funcParamExpression`? Could you demo how it would be?

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/parser/AggregationFuncStmt.java
```diff
@@ -49,22 +50,23 @@ public ConditionExpression getNextFuncConditionExpression() {
         return funcConditionExpressions.get(funcConditionExpressionGetIdx++);
     }
 
-    public void addFuncArg(Argument argument) {
+    public void addFuncArg(Argument... argument) {
```
wu-sheng, line 53: An argument is an Argument Array. I think this is wrong from grammar level. We should discuss how to host an Array inside the `Argument`, or create a sub-type of `Argument`, or create an interface with two implementations??

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/OALRuntime.java
pg-yang: Sorry , This PR  just for preview or  discussion  . You could ignore it

oap-server/oal-grammar/src/main/antlr4/org/apache/skywalking/oal/rt/grammar/OALParser.g4
pg-yang:
```
mq_transmission_latency = from(MessageQueueAccess.transmissionLatency).filter(operation == MessageQueueOperation.Consume).labeledLongAvg([topic,queue,queue,"1"]);

```

Like  this , we  could  use  `[a]` to represent an arg

pg-yang:
```
@Entrance
    public final void combine(@SourceFrom long value, @ConstOne long count, @Arg String[] labels)
```

pg-yang: I  recked `funcParamExpression ` , you  are right  . I should  remove  `funcParamExpression `

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/parser/AggregationFuncStmt.java
pg-yang: Yes ,  Do you  agree  with adding `arrayFuncParamExpression `  to `grammar` ?
The  demo  such as 
```
mq_transmission_latency = from(MessageQueueAccess.transmissionLatency).filter(operation == MessageQueueOperation.Consume).labeledLongAvg([topic,queue,queue,"1"]);

```
Let's  talk  all here  .

oap-server/oal-grammar/src/main/antlr4/org/apache/skywalking/oal/rt/grammar/OALParser.g4
wu-sheng: My question is about `funcParamExpression`, what you mentioned is `attributeExpression`(topic,queue,queue).

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/parser/AggregationFuncStmt.java
wu-sheng: Please explain this `topic,queue,queue,"1"` first. Is this meaning, first three elements of the array are from source attributes, and the last is a literal string?

pg-yang: queue , topic is from  source  attributes , and  "1" is demo  just  for demonstrating  literal string

wu-sheng: Then, you still don't have funcParamExpression. Could you confirm whether you need this?

About the array type, do we need to make it type-sensitive? Such as building StringArray, rather than a general Array?

pg-yang: >  you still don't have funcParamExpression

Yes , I would  confirm it .

> About the array type

Actually , I want to build the type of array  based on the parameter type , which will support base-type , such as  StringArray  , intArray，doubleArray， etc

wu-sheng: But, there is an issue about this. `SourceFrom` should be for `getting field value from a source's attribute`, and `Arg` should be a kind of literal declared from the OAL.

Now, my question would be, can't we build a field in the source including key-value pairs directly? And use `SourceFrom` rather than adding this complexity for the OAL? We are not really building an array mixed with attribute values and literal strings, right?

wu-sheng: I commented another concern https://github.com/apache/skywalking/issues/9676#issuecomment-1283735427 about the original proposal.

Issue #9762, opened 2022-10-11
MySQL/PostgreSQL monitor should be able to support cluster mode scenario


kezhenxu94:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

In recent pull requests from @yswdqz we can monitor metrics from MySQL and PostgreSQL, but currently we assume that every MySQL/PostgreSQL server only has one instance, and we use that instance as the service name, (refer to our demo environment http://demo.skywalking.apache.org/mySQL)

<img width="1182" alt="image" src="https://user-images.githubusercontent.com/15965696/195023145-784519cb-af23-4736-b9dc-cdc87862f11d.png">

However, a MySQL/PostgreSQL server should be able to have multiple instances and form a cluster, in that case, we should have aggregated metrics for those instances.

### Use case

When there are multiple instances/nodes for one MySQL/PostgreSQL server, users can have an aggregated overview of the database server.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

kezhenxu94:
@yswdqz I can't assign this to you YET, are you willing to continue this?

wu-sheng:
@yswdqz I think we could have perf of both k8s service and per pod(as skywalking instance level).

We had a similar discussion with @pg-yang from https://github.com/apache/skywalking-showcase/pull/83#issuecomment-1272504487. Generally about the same preferences.

yswdqz:
Yes, I will.
But I am doing another PR now, might be doing this later.

wu-sheng:
If you mean the SQL-sign, I think you don't need to hurry. After all, I could image that is a long-term feature, we may face forward and backward several times, also be struggling due to the tradeoff :)

yswdqz:
> If you mean the SQL-sign, I think you don't need to hurry. After all, I could image that is a long-term feature, we may face forward and backward several times, also be struggling due to the tradeoff :)

Got it.

wu-sheng:
@yswdqz How is this going?

yswdqz:
> @yswdqz How is this going?

I haven't started it yet ( because of something about my university ). I will start it tomorrow.

By the way, Does this mean that I should set what I previously set as `service` to `instance` and then set the cluster's metrics to `service`?

[...]

PR #9822, opened 2022-10-20
Change the way of loading MAL rules

yswdqz:
When setting the level for MAL metrics, using the service and instance methods for each metric will generate a lot of redundant code.
We can use MAL's "expSuffix" to avoid it, but because the "expSuffix" required by the instance level metric and the service level metric is different, it will inevitably lead to file splitting (mysql.yaml -> mysql-instance.yaml + mysql-service.yaml )
So I changed the loading method of MAL files to be folder based to facilitate future expansion.

I haven't updated the doc yet because I think maybe somewhere needs to be modified.

### <Feature description>


wu-sheng:
@kezhenxu94 @wankai123 What do you think to do this breaking change?

wankai123:
One thing I think we should be concerned about is: such as k8s, this way will enable all metrics at the same time and no other choice. Some metrics need access to the API server. If that can be accepted, I'm OK with this.

wu-sheng:
@yswdqz Do you think is it better to support the path pattern as well as use a file name? Such as `k8s/*.yaml` or `apisix/apisix.yaml`

yswdqz:
> One more compatibility we could add is, support ignore suffix(`.yaml`). Then a folder should be `/folder/*`, a file(`folder/mal.yaml`) should be either `folder/mal` or `folder/mal.yaml`. By this, we could support users' local configuration files placed and config in the old way.

Maybe we should replace `/folder/*` as 'folder' so that old configuration can be support.
And we can judge file/folder by if there is a '/' in the string

wu-sheng:
> > One more compatibility we could add is, support ignore suffix(`.yaml`). Then a folder should be `/folder/*`, a file(`folder/mal.yaml`) should be either `folder/mal` or `folder/mal.yaml`. By this, we could support users' local configuration files placed and config in the old way.
> 
> Maybe we should replace `/folder/*` as 'folder' so that old configuration can be support. And we can judge file/folder by if there is a '/' in the string

That is not good practice. In typical setup, files in the folder should include `*` for files in the folder, and `**` represents files in the current and sub-folders.

yswdqz:
> > > One more compatibility we could add is, support ignore suffix(`.yaml`). Then a folder should be `/folder/*`, a file(`folder/mal.yaml`) should be either `folder/mal` or `folder/mal.yaml`. By this, we could support users' local configuration files placed and config in the old way.
> > 
> > 
> > Maybe we should replace `/folder/*` as 'folder' so that old configuration can be support. And we can judge file/folder by if there is a '/' in the string
> 
> That is not good practice. In typical setup, files in the folder should include `*` for files in the folder, and `**` represents files in the current and sub-folders.

I understand. Should we support other pattern? Such as mysql/mysql*.yaml

wu-sheng:
* `/folder/*` should be from the root.
* `folder/*` represents `folder` in the current context
* `folder/file` represents a file or a folder. In our case, `folder` is illegal for our case. So, we treat this as a file, just add a default suffix(`.yaml`) by following we are doing for now.

yswdqz:
> > I understand. Should we support other pattern? Such as mysql/mysql*.yaml
> 
> If you want, we could add a utility in SkyWalking for further general use, but I think we don't have to. Your call.

I agree.

wu-sheng:
> > * `/folder/*` should be from the root.
> 
> Is the `root` means MAL 's root folder?

Actually, we don't have the root. Because we have declared all files should be `otel-rules`, the root and current context are the same.

kezhenxu94:
Hi. What's the idea / use case here to move the yaml files into a folder, I'm thinking this just leads to "one file per folder" eventually and doesn't make much sense at all.

kezhenxu94:
As I can see in this pull request, every folder only has a single file, even the exception one (k8s folder) we are considering to support activating one of them independently.

wu-sheng:
> As I can see in this pull request, every folder only has a single file, even the exception one (k8s folder) we are considering to support activating one of them independently.

Actually, there was an off-list discussion from https://github.com/apache/skywalking/issues/9762
MySQL would become multiple files as instance files would be added from MAL analysis. 
I think the main reason for this, we are setting entity statements in expSuffix. So, once we need service and instance levels metrics at the same time, we usually use two files.

> expSuffix: tag({tags -> tags.cluster = 'istio-ctrl::' + tags.cluster}).service(['cluster', 'app'], Layer.MESH_CP)

kezhenxu94:
@yswdqz you edited my comment and that makes me (and others) confused because I didn’t post that comment content and mine is gone.

yswdqz:
> @yswdqz you edited my comment and that makes me (and others) confused because I didn’t post that comment content and mine is gone.

Sorry that is what I want to say ,I made a mistake

yswdqz:
> As I can see in this pull request, every folder only has a single file, even the exception one (k8s folder) we are considering to support activating one of them independently.

I need to change mysql.yaml to mysql-service.yaml and mysql-instance.yaml
Postgresql.yaml also has the same situation. So I send this pr.

The reason of these changes is about half of metrics should be instance level, and other metrics should be service level.
Without this change, we can only add level for each metric.
such as:
https://github.com/apache/skywalking/blob/master/oap-server/server-starter/src/main/resources/otel-rules/apisix.yaml

wu-sheng:
What is the question? My point is `/folder` and `folder` have no difference. The we only have one concept, root and current path are the same thing for MAL initialization, no matter in `otel-rules` or others

yswdqz:
> What is the question? My point is `/folder` and `folder` have no difference. The we only have one concept, root and current path are the same thing for MAL initialization, no matter in `otel-rules` or others

I understand now. 
So we should support:
- `mysql/mysql-instance.yaml` : for file
- `mysql/mysql-instance` : the same as `mysql/mysql-instance.yaml`
- `/mysql/mysql-instance` : the same as `mysql/mysql-instance.yaml`
- `/mysql/mysql-instance.yaml` : the same as `mysql/mysql-instance.yaml`

- `/mysql/*.yaml` : for folder
- `/mysql/*` : the same as `/mysql/*.yaml`
- `mysql/*.yaml` : the same as `/mysql/*.yaml`
- `mysql/*` : the same as `/mysql/*.yaml`

Is it right? And maybe we should support `**` , too?

wu-sheng:
Do we really have case for deeper folder? I doubt that.

yswdqz:
> Do we really have case for deeper folder? I doubt that.

For now we don't have it.
I understand. We can support it when we really need it.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -47,29 +49,70 @@ public static List<Rule> loadRules(final String path) throws ModuleStartExceptio
...
+                            }
+                            Rule rule = new Yaml().loadAs(r, Rule.class);
+                            if (rule == null) {
+                                return null;
+                            }
+                            rule.setName(ruleName);
+                            return Stream.of(rule);
+                        } catch (IOException e) {
+                            LOG.debug("Reading file {} failed", f, e);
+                        }
```
kezhenxu94, line 111: Can you extract this to a method and share it with the lines above in the `if` statement?

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -47,29 +49,70 @@ public static List<Rule> loadRules(final String path) throws ModuleStartExceptio
...
         try {
-            rules = ResourceUtils.getPathFiles(path);
+            rules = ResourceUtils.getAllPathFiles(path);
         } catch (FileNotFoundException e) {
             throw new ModuleStartException("Load fetcher rules failed", e);
         }
+        final List<String> formedEnabledRules =
+                enabledRules
+                        .stream()
+                        .map(rule -> rule.endsWith(".yaml") ? rule : rule + ".yaml")
```
kezhenxu94, line 59: let's also take `.yml` into consideration, it's also valid extension name of yaml

kezhenxu94 (commented):
`apisix/*` and `apisix` are the same in our case, can we make it simpler to just allow `apisix`, and `apisix/some-file.yaml`?

Basically we just need to support scenarios:

- Loading the entire folder, using `apisix`
- Loading some files under a folder, using `apisix/file1.yaml`, `apisix/file2.yaml`.

I don't think the wildcard patterns like `xx/*` is necessary in our use cases.

yswdqz:
> > One more compatibility we could add is, support ignore suffix(`.yaml`). Then a folder should be `/folder/*`, a file(`folder/mal.yaml`) should be either `folder/mal` or `folder/mal.yaml`. By this, we could support users' local configuration files placed and config in the old way.
> 
> Maybe we should replace `/folder/*` as 'folder' so that old configuration can be support. And we can judge file/folder by if there is a '/' in the string

@kezhenxu94 
Please see this. I thought about that too

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
yswdqz: Got it.

kezhenxu94:
> Maybe we should replace `/folder/*` as 'folder' so that old configuration can be support. And we can judge file/folder by if there is a '/' in the string

@yswdqz "old configuration can be support", do you mean the current implementation can support old configurations file structure?

yswdqz:
> > > One more compatibility we could add is, support ignore suffix(`.yaml`). Then a folder should be `/folder/*`, a file(`folder/mal.yaml`) should be either `folder/mal` or `folder/mal.yaml`. By this, we could support users' local configuration files placed and config in the old way.
> > 
> > 
> > Maybe we should replace `/folder/*` as 'folder' so that old configuration can be support. And we can judge file/folder by if there is a '/' in the string
> 
> That is not good practice. In typical setup, files in the folder should include `*` for files in the folder, and `**` represents files in the current and sub-folders.

@kezhenxu94 
Sorry I have not said clear. 
My idea was rejected here.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -47,29 +50,72 @@ public static List<Rule> loadRules(final String path) throws ModuleStartExceptio
     public static List<Rule> loadRules(final String path, List<String> enabledRules) throws ModuleStartException {
         File[] rules;
         try {
-            rules = ResourceUtils.getPathFiles(path);
+            rules = ResourceUtils.getAllPathFiles(path);
         } catch (FileNotFoundException e) {
             throw new ModuleStartException("Load fetcher rules failed", e);
         }
+        final List<String> formedEnabledRules =
```
wu-sheng, line 57: Could you build UTs for the new logic? Such as through `loadRules` method?

wu-sheng:
Could you recheck the document? I think we need some adjustment about the documents of metrics relative configurations

- OTEL https://skywalking.apache.org/docs/main/next/en/setup/backend/opentelemetry-receiver/
- Meter https://skywalking.apache.org/docs/main/next/en/setup/backend/backend-meter/#configuration-file
- Zabbix https://skywalking.apache.org/docs/main/next/en/setup/backend/backend-zabbix/
- Things to review on PR for telegraf https://github.com/apache/skywalking/pull/9620

We are mentioning these as rule names, but it seems they are not anymore. At least they should be `rule pattern`, rather than `rule name`. @kezhenxu94 WDYT?
image

kezhenxu94:
> Could you recheck the document? I think we need some adjustment about the documents of metrics relative configurations
> 
> 
> 
> - OTEL https://skywalking.apache.org/docs/main/next/en/setup/backend/opentelemetry-receiver/
> 
> - Meter https://skywalking.apache.org/docs/main/next/en/setup/backend/backend-meter/#configuration-file
> 
> - Zabbix https://skywalking.apache.org/docs/main/next/en/setup/backend/backend-zabbix/
> 
> - Things to review on PR for telegraf https://github.com/apache/skywalking/pull/9620
> 
> 
> 
> We are mentioning these as rule names, but it seems they are not anymore. At least they should be `rule pattern`, rather than `rule name`. @kezhenxu94 WDYT?
> 
> image
> 
> 

I think we can simply remove the Rile Name column because there is an existing column named Configuration File and users can compose a pattern to include that file.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
yswdqz: Got it.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -47,29 +50,72 @@ public static List<Rule> loadRules(final String path) throws ModuleStartExceptio
...
+            if (rule == null) {
                 return null;
-            })
-            .filter(Objects::nonNull)
-            .collect(Collectors.toList());
+            }
+            rule.setName(ruleName);
+            return rule;
+        } catch (IOException e) {
+            LOG.debug("Reading file {} failed", file, e);
```
kezhenxu94, line 117: This is a fatal exception and we should propagate to the caller, and stop the OAP from starting

yswdqz: Got it.

kezhenxu94 (commented):
Please add a case to test the `enabledOtelRules=k8s/*` and the configuration file is `k8s/abc.yml`, from the codes I think the configuration file is not loaded, which is expected.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
yswdqz: > Could you build UTs for the new logic? Such as through `loadRules` method?

What does `build UTs` mean? Is it extract a method?

wu-sheng: Unit Tests. The path/file pattern mechanism is added newly, and it is better if you could test through correct input and wrong put with predictable results(correct file list or exception or empty list)

yswdqz: I understand.

yswdqz:
About the doc. I only found three places to use the loadRules method (otel-rules, log-mal-rules, envoy-metric-rules). So I think some docs needn't be modified.

wu-sheng:
I think we don't have to move so many real otel YAML files into the tests. The test files should be as simple as possible. Unless you are going to verify some expressions, your tests don't like so

oap-server/analyzer/meter-analyzer/src/test/java/org/apache/skywalking/oap/meter/analyzer/dsl/RuleLoaderTest.java
```diff
@@ -0,0 +1,91 @@
...
+    public void test() {
+        List<Rule> rules = null;
+        try {
+            rules = Rules.loadRules("otel-rules", enabledRule);
+        } catch (Exception e) {
+            if (isThrow) {
+                return;
+            }
+            fail("load rules failed");
+        }
```
wu-sheng, line 87: Learn how to wrong predicable exceptions in UT

```java
@Test(expected = IllegalArgumentException.class)
```

oap-server/analyzer/meter-analyzer/src/test/java/org/apache/skywalking/oap/meter/analyzer/dsl/RuleLoaderTest.java
```diff
@@ -0,0 +1,91 @@
...
+    public int rulesNumber;
+
+    @Parameterized.Parameter(2)
+    public boolean isThrow;
+
+    @Parameterized.Parameters(name = "{index}: {0}")
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][] {
+            // abc/abc.yml is a not well-formed yml file.
+            {Arrays.asList("abc/*"), 0, true},
```
wu-sheng, line 55: The good test mock samples should be 
- /illegal-yaml/test.yml
- /test-folder/case1.yml
- single-file-case.yml

oap-server/analyzer/meter-analyzer/src/test/java/org/apache/skywalking/oap/meter/analyzer/dsl/RuleLoaderTest.java
```diff
@@ -0,0 +1,91 @@
...
+            {Arrays.asList("k8s/k8s-cluster.yaml"), 1, false},
+            {Arrays.asList("k8s/k8s-cluster"), 1, false},
+            {Arrays.asList("oap.yaml"), 1, false},
+            {Arrays.asList("oap.yml"), 1, false},
+            {Arrays.asList("oap"), 1, false},
+            {Arrays.asList("/oap.yaml"), 1, false},
+            {Arrays.asList("/oap.yml"), 1, false},
+            {Arrays.asList("/oap.yml", "/k8s/*"), 5, false},
+            //if enabledRule not found, will not fail but will not load any rules
+            {Arrays.asList(UUID.randomUUID().toString()), 0, false},
```
wu-sheng, line 73: I think this should through an exception rather than a `false` return.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -55,24 +57,24 @@ public static List<Rule> loadRules(final String path, List<String> enabledRules)
...
-                            if (rule.endsWith(".yaml")) {
-                                return rule;
-                            } else if (rule.endsWith(".yml")) {
-                                return rule.substring(0, rule.length() - 2) + "aml";
-                            } else {
-                                return rule + ".yaml";
-                            }
-                        })
-                        .collect(Collectors.toList());
+        Map<String, Boolean> formedEnabledRules = enabledRules
```
wu-sheng, line 60: Map<String, Boolean> means a Set<String>?

wu-sheng: What does this value true/false mean?

yswdqz: This is used to determine whether an enabled rule does not have a corresponding configuration file

docs/en/changes/changes.md
```diff
@@ -68,6 +68,7 @@
 * Add new config initialization mechanism of module provider. This is a ModuleManager lib kernel level change.
 * [**Breaking Change**] Support new records query protocol, rename the column named `service_id` to `entity_id` for support difference entity.
   Please re-create `top_n_database_statement` index/table.
+* [**Breaking Change**] Change the way of loading MAL rules(support pattern).
```
wu-sheng, line 71:
```suggestion
* [**Breaking Change**] Change the way of loading MAL rules(support pattern).
* Move k8s relative MAL files into `/otel-rules/k8s`.
```

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
yswdqz: At first they were false.
If a rule found its configuration file, it change to true. 
So that I can judge which enabled rule's file is not exist by map's false value

docs/en/changes/changes.md
yswdqz: Got it.

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/ResourceUtils.java
```diff
@@ -90,4 +90,13 @@ private static List<File> getDirectoryFilesRecursive(String directoryPath, List<
         }
         return fileList;
     }
+
+    public static File[] getAllPathFiles(String path) throws FileNotFoundException {
```
wu-sheng, line 94: What is the difference from existing `getPathFiles` method?

yswdqz: Got it.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -47,29 +53,79 @@ public static List<Rule> loadRules(final String path) throws ModuleStartExceptio
...
+        try (Reader r = new FileReader(file)) {
+            String fileName = file.getName();
+            int dotIndex = fileName.lastIndexOf('.');
+            if (fileName.endsWith(".yml")) {
+                fileName = fileName.substring(0, fileName.length() - 2) + "aml";
+            }
+            if (dotIndex == -1 || !"yaml".equals(fileName.substring(dotIndex + 1))) {
+                return null;
+            }
+            String ruleName = fileName.substring(0, dotIndex);
```
wu-sheng, line 105: What would happens if illegal files and legal files are mixed together? And I think there are two cases
1. illegal.yaml is mixed with other normal files, then we should fail in booting. Right?
2. Another case is, we have a hidden file, such as `.abc` is a hidden file in the folder, but it is not a YAML file.

Such as in a macOS, you could have several hidden files in every folder,  **.DS_Store**.
image

yswdqz: I have concern about it.
For files like `abc` which  without `.yaml/.yml`. Method will return null.
image
For files which with `.yaml/.yml`. If it is a empty file.  Method will return null.
image

For files with `.yaml/.yml`, and it is not a empty file. Method will throw YAML Exception. I have not found good way to resolve it. 
Because of we don't support getting files recursively, so if temp files is in a deeper folder, we should not concern.

Tomorrow I will ignore files start with ".".

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -47,29 +53,79 @@ public static List<Rule> loadRules(final String path) throws ModuleStartExceptio
...
+        if (formedEnabledRules.containsValue(false)) {
+            throw new UnexpectedException("Some configuration files of enabled rules are not found, enabled rules: " + formedEnabledRules.keySet());
+        }
+        return result;
+    }
+
+    private static Rule getRulesFromFile(Map<String, Boolean> formedEnabledRules, File directory, File file) {
+        try (Reader r = new FileReader(file)) {
+            String fileName = file.getName();
+            int dotIndex = fileName.lastIndexOf('.');
```
wu-sheng, line 98: You are using `lastIndexOf`, so `.asf.yaml` would only get the index of the dot in `.yaml`, isn't it?

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -47,29 +53,79 @@ public static List<Rule> loadRules(final String path) throws ModuleStartExceptio
...
+            if (fileName.endsWith(".yml")) {
+                fileName = fileName.substring(0, fileName.length() - 2) + "aml";
+            }
+            if (dotIndex == -1 || !"yaml".equals(fileName.substring(dotIndex + 1))) {
+                return null;
+            }
```
wu-sheng, line 104: Why do we change yml to yaml? These lines seems meaning we support file with `.yml` and `.yaml` suffixes? Why don't we put the codes in that way?

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
yswdqz: My fault. I forget it.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
yswdqz: image
For this, we support .yml.
And this will cause a problem : enabledRule: mysql means mysql.yml or mysql.yaml ? So I just change yml to yaml to support both situation. The only problem with such code is that if there are both mysql.yml and mysql.yaml, there is no guarantee which one will be loaded.

wu-sheng: I know we should support both. I mean the way you put in codes is strange. Taking two suffixes should do directly, rather than a rename. In your way, we could(maybe not now) output the wrong file causing confusing.

yswdqz: I understand, I will try it later.

kezhenxu94 (commented):
To be honest, I think the related codes (manipulating the file names) are hard to maintain / review for me. @yswdqz I suggest you take a look at the JDK nio SDK, I have wrote a quick demo to show how to solve the pattern matching neatly that you can reference.

```java
    public static void main(String[] args) throws IOException {
        final String path = "otel-rules";
        final Path root =
            new File(ResourceUtils.class.getClassLoader().getResource(path).getPath()).toPath();
        List<String> enabledRules = Arrays.asList(
            "apisix.yaml", // Exact match
            "oap", // Without file extension
            "k8s/*" // Glob match
        )
            .stream()
            .map(rule -> {
                // File name without extension, add {.yaml,.yml} to match both .yml and .yaml
                if (!rule.endsWith(".yaml") && !rule.endsWith(".yml") && !rule.endsWith("*")) {
                    return rule + "{.yaml,.yml}";
                }
                return rule;
            })
            .collect(toList());

        final List<Path> rules;
        try (Stream<Path> stream = Files.walk(root)) {
            rules = stream
                .filter(it -> enabledRules.stream()
                    .anyMatch(rule -> FileSystems.getDefault().getPathMatcher("glob:" + rule)
                        .matches(root.relativize(it))))
                .collect(toList());
        }

        // Rule files that need to be loaded

        rules.forEach(System.out::println);
    }
```

yswdqz:
Thank you！ I'm looking at other open source code similar to this situation and haven't found a suitable one.
Thank you for your help.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -40,36 +49,71 @@
...
+        final Path root = ResourceUtils.getPath(path);
+
+        Map<String, Boolean> formedEnabledRules = enabledRules
+                .stream()
+                .map(rule -> {
+                    rule = rule.trim();
+                    if (rule.startsWith("/")) {
+                        rule = rule.substring(1);
+                    }
+                    if (!rule.endsWith(".yaml") && !rule.endsWith(".yml") && !rule.endsWith("*")) {
```
kezhenxu94, line 67: We can use this to ensure that all listed files are `.yaml` or `.yml`

```suggestion
                    if (!rule.endsWith(".yaml") && !rule.endsWith(".yml")) {
```

So (2)

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -40,36 +49,71 @@
...
+        }
+        try (Reader r = new FileReader(file)) {
+            String fileName = file.getName();
+            int dotIndex = fileName.lastIndexOf('.');
+            if (fileName.startsWith(".") || dotIndex == -1 || (!fileName.endsWith(".yml") && !fileName.endsWith(".yaml"))) {
                 return null;
-            })
-            .filter(Objects::nonNull)
-            .collect(Collectors.toList());
+            }
```
kezhenxu94, line 106: (2) ... we don't need to check these

yswdqz: Got it.

oap-server/server-library/library-util/src/main/java/org/apache/skywalking/oap/server/library/util/ResourceUtils.java
```diff
@@ -90,4 +91,8 @@ private static List<File> getDirectoryFilesRecursive(String directoryPath, List<
         }
         return fileList;
     }
+
+    public static Path getPath(String path) {
+        return new File(Objects.requireNonNull(ResourceUtils.class.getClassLoader().getResource(path)).getPath()).toPath();
```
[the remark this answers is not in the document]
wu-sheng: @sonatype-lift ignore

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -40,36 +49,71 @@
...
+                                        }
+                                        return matches;
+                                    }))
+                    .map(path1 -> getRulesFromFile(root.relativize(path1), path1))
+                    .filter(Objects::nonNull)
+                    .collect(Collectors.toList()) ;
+        }
+
+        if (formedEnabledRules.containsValue(false)) {
+            throw new UnexpectedException("Some configuration files of enabled rules are not found, enabled rules: " + formedEnabledRules.keySet());
```
wu-sheng, line 91: I feel you should only put keys with `formedEnabledRules#value == true` here?

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -40,36 +49,71 @@
...
+                    .filter(it -> formedEnabledRules.keySet().stream()
+                                    .anyMatch(rule -> {
+                                        boolean matches = FileSystems.getDefault().getPathMatcher("glob:" + rule)
+                                                .matches(root.relativize(it));
+                                        if (matches) {
+                                            formedEnabledRules.put(rule, true);
+                                        }
+                                        return matches;
+                                    }))
+                    .map(path1 -> getRulesFromFile(root.relativize(path1), path1))
```
wu-sheng, line 85:
```suggestion
                    .map(pathPointer -> getRulesFromFile(root.relativize(pathPointer), pathPointer))
```

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -40,36 +49,71 @@
...
+                    .collect(Collectors.toList()) ;
+        }
+
+        if (formedEnabledRules.containsValue(false)) {
+            throw new UnexpectedException("Some configuration files of enabled rules are not found, enabled rules: " + formedEnabledRules.keySet());
+        }
+        return rules;
+    }
+
+    private static Rule getRulesFromFile(Path relative, Path path) {
```
wu-sheng, line 96:
```suggestion
    private static Rule getRulesFromFile(String ruleName, Path path) {
```
I think the caller should build the ruleName, rather than hide the logic inside this method with two Paths as parameters.

wu-sheng:
Some nits, generally this is good.

yswdqz:
Got it , I will fix it soon.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -40,36 +49,76 @@
...
+                                        if (matches) {
+                                            formedEnabledRules.put(rule, true);
+                                        }
+                                        return matches;
+                                    }))
+                    .map(pathPointer -> {
+                        Path relativize = root.relativize(pathPointer);
+                        String relativizePath = relativize.toString();
+                        relativizePath = relativizePath.substring(0, relativizePath.lastIndexOf("."));
+                        return getRulesFromFile(relativizePath, pathPointer);
```
wu-sheng, line 89:
```suggestion
                        // Use relativized file path without suffix as the rule name. 
                        String ruleName = root.relativize(pathPointer).toString().substring(0, relativizePath.lastIndexOf("."));
                        return getRulesFromFile(ruleName, pathPointer);
```

I think this is better?

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -40,36 +49,76 @@
...
+            return null;
+        }
+        try (Reader r = new FileReader(file)) {
+            String fileName = file.getName();
+            if (fileName.startsWith(".")) {
                 return null;
-            })
-            .filter(Objects::nonNull)
-            .collect(Collectors.toList());
+            }
```
kezhenxu94, line 113: Not sure why we need to skip hidden files, but if you insist, please use this 

```java
         if (!file.isFile() || file.isHidden()) {
             return null;
         }
```

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -40,36 +49,76 @@
...
+                    .filter(it -> formedEnabledRules.keySet().stream()
+                                    .anyMatch(rule -> {
+                                        boolean matches = FileSystems.getDefault().getPathMatcher("glob:" + rule)
+                                                .matches(root.relativize(it));
+                                        if (matches) {
+                                            formedEnabledRules.put(rule, true);
+                                        }
+                                        return matches;
+                                    }))
+                    .map(pathPointer -> {
```
kezhenxu94, line 85: Why `pathPointer`? What does `pointer` mean?

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
```diff
@@ -40,36 +49,76 @@
...
                 return null;
-            })
-            .filter(Objects::nonNull)
-            .collect(Collectors.toList());
+            }
+            Rule rule = new Yaml().loadAs(r, Rule.class);
+            if (rule == null) {
+                return null;
+            }
+            rule.setName(ruleName);
```
kezhenxu94, line 118: @yswdqz can you confirm whether rule.name is used anywhere? I locally search and find nowhere use rule.name. If you can confirm rule.name is never used, we may consider remove it

kezhenxu94:
Codes look good to me overall, just some comments above.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
wu-sheng: The hidden file would exist when we set the rule as `/folder/*`.(Such as .DS_store in MacOS) That was the first request.
But with we added `.yml` and `.yaml` to the match pattern, I think hidden file is not an issue.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
wu-sheng: The `path` variable is used, it was named as `path`. So, I suggest adding a point suffix just to separate the name in closure.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
yswdqz: So should we skip hidden files for now?

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
yswdqz: For now it is not be used.
But maybe we will use it in the future? Such as LAL.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/rule/Rules.java
wu-sheng: We could use Zhenxu's way to do so.

wu-sheng:
@yswdqz CI fails, please fix them.

yswdqz:
> @yswdqz CI fails, please fix them.

I don't know why the dead link checker fail, and maybe it cause oap failed to start?

kezhenxu94:
> > @yswdqz CI fails, please fix them.
> 
> 
> 
> I don't know why the dead link checker fail, and maybe it cause oap failed to start?

Dead link checker won't block OAP from starting.

kezhenxu94:
Chances might be you mess up the ui submodule. It was pointed to an earlier version and the OAP failed to start up because of that (protocol mismatch)

yswdqz:
> Chances might be you mess up the ui submodule. It is pointed to an earlier version and the OAP failed to start up because of that (protocol mismatch)

Got it.

kezhenxu94:
> Chances might be you mess up the ui submodule. It is pointed to an earlier version and the OAP failed to start up because of that (protocol mismatch)

Uh sorry. I don't mean UI submodule.  What I wanted to say is the query protocol module 🤣

yswdqz:
> Uh sorry. I don't mean UI submodule. What I wanted to say is the query protocol module 🤣

Got it.🤣

wu-sheng:
If you have synced UI, you need to update change logs accordingly.

yswdqz:
> If you have synced UI, you need to update change logs accordingly.
How can I update logs about?
By the way, I have not change codes about ui, should I sync UI ?

wu-sheng:
You don't have to. If you do, update the UI part changelog, according to the commits between before and after commits.

wu-sheng:
image

The query proto is reverting not updating actually. Your changes should not touch these protocols, as irrelevant

wu-sheng (approved):

PR #9827, opened 2022-10-21
Support new records query protocol

mrproliu:
1. Following https://github.com/apache/skywalking-query-protocol/pull/100, adapt the new records query protocol. Rename the column name `service_id` to `entity_id` for support difference entity.
2. Update the `ProcessRegistry#generateVirtualProcess` as public, It helps the network profiling not working in the Kubernetes environment. 


wu-sheng (approved):
LGTM

wu-sheng (approved):

PR #9828, opened 2022-10-21
Add new config initialization mechanism of module provider to replace existing

wu-sheng:

The root cause of this change is when I want to extend an existing module provider to another, but need to add new configurations. In the previous implementations, most providers have to hold a field/pointer to the configuration, in order to use it in `init`/`start`/... stages. This should be fine until we want to extend and add a new configuration, a new configuration class could inherit from the existing config, but the new provider can't inherit the existing provider, because you can't inject the extended config to replace the config instance created and hold in the old provider.
Through this new creator interface, the new provider could intercept the config type, and return a new one, and notify the old provider with the new initialized config.


kezhenxu94:
> The root cause of this change is when I want to extend an existing module provider to another, but need to add new configurations. In the previous implementations, most providers have to hold a field/pointer to the configuration, in order to use it in `init`/`start`/... stages. This should be fine until we want to extend and add a new configuration, a new configuration class could inherit from the existing config, but the new provider can't inherit the existing provider, because you can't inject the extended config to replace the config instance created and hold in the old provider.

Hi, this is how I extend the provider AND the configuration. Let's take `AnalyzerModuleProvider` as an example.

- Extend the configuration

```java
public class MyAnalyzerModuleConfig extends AnalyzerModuleConfig {
  // This is my new configuration
  @Getter
  @Setter
  private String myNewConfiguration;
}
```

- Extend the provider

```java

public class MyAnalyzerModuleProvider extends AnalyzerModuleProvider {
  private final MyAnalyzerModuleConfig newModuleConfig;

  public MyAnalyzerModuleProvider(MyAnalyzerModuleConfig newModuleConfig) {
    this.newModuleConfig = newModuleConfig;
  }

  @Override
  public ModuleConfig createConfigBeanIfAbsent() {
      return newModuleConfig;
  }

  @Override
  public void prepare() throws ServiceNotProvidedException, ModuleStartException {
      this.registerServiceImplementation(MyService.class, new MyService(newModuleConfig));
  }
}
```

- And the new service

```java

public class MyAnalyzerModuleConfig extends AnalyzerModuleConfig {
  // This is my new configuration
  @Getter
  @Setter
  private String myNewConfiguration;
}
```


Did I understand correctly? Do we have anything new in this PR that we can't do in the previous implementation?

wu-sheng:
If AnalyzerModuleProvider needs the values in its configuration, how would you do?

kezhenxu94:
> If AnalyzerModuleProvider needs the values in its configuration, how would you do?

You extend the `AnalyzerModuleProvider` (as `MyAnalyzerModuleProvider`) and you want the original `AnalyzerModuleProvider` to access the extended configuration?

wu-sheng:
The codes in AnalyzerModuleProvider are accessing, we have to copy all codes in AnalyzerModuleProvider to the new provider, which loses the point of inheritance. I know we have to do this before, I want to change the suitiation.

wankai123:
> > If AnalyzerModuleProvider needs the values in its configuration, how would you do?
> 
> 
> 
> You extend the `AnalyzerModuleProvider` (as `MyAnalyzerModuleProvider`) and you want the original `AnalyzerModuleProvider` to access the extended configuration? 

yes，when we need to invoke some super methods

kezhenxu94:
OK, I get the point. Changing the config field to `protected` should also do the trick but that's not strictly enforced for new codes. The idea in this PR sounds good to me now.

wu-sheng:
> OK, I get the point. Changing the config field to `protected` should also do the trick but that's not strictly enforced for new codes. The idea in this PR sounds good to me now.

Yes, that works. But this can't be forced when we implement a new module and forget in the review process. So, I changed a lot to make we don't care `protected` or not.

kezhenxu94:
Wait, I don't see this patch fixes the problem we want to solve in the discussion, and the tests don't reflect that too.

> Through this new creator interface, the new provider could intercept the config type, and return a new one, and notify the old provider with the new initialized config.

Can you share a code snippet demonstrating how this can be done?

wu-sheng:
This is how we are going to extend AnalyzerModuleProvider with a new configuration class without copy codes. All methods in AnalyzerModuleProvider still work, of course, you could override them and call `super.xxx` then add new logic.

```java
public class AnalyzerModuleProviderExt extends AnalyzerModuleProvider{
    private NewConfig config;

    @Override
    public ConfigCreator newConfigCreator() {
        final ConfigCreator parentConfigCreator = super.newConfigCreator();
        return new ConfigCreator<NewConfig>() {
            @Override
            public Class type() {
                return NewConfig.class;
            }

            @Override
            public void onInitialized(final NewConfig initialized) {
                config = initialized;
                parentConfigCreator.onInitialized(initialized);
            }
        };
    }

    public static class NewConfig extends AnalyzerModuleConfig {
        private String newField;
    }
}
```

oap-server/server-library/library-module/src/main/java/org/apache/skywalking/oap/server/library/module/ModuleProvider.java
```diff
@@ -52,9 +52,28 @@ protected final ModuleManager getManager() {
...
+         * Declare the type of the config class
+         * @return class type
+         */
+        Class<T> type();
+
+        /**
+         * Callback when the ModuleManager kernel has initialized this configuration.
+         * @param initialized instance of the given {@link #type()}
+         */
+        void onInitialized(T initialized);
```
kezhenxu94, line 75: I'm wondering whether we can lift this method up the the class `ModuleProvider`, because IMO `onInitialized` is a lifecycle method just similar to `ModuleProvider.prepare`, `ModuleProvider.start`, `ModuleProvider.notifyAfterCompleted`, etc., maybe we can unify the lifecycle methods together. And the changes should become very minimal

wu-sheng: I am doing this on purpose, even though this makes more changes in this PR actually. Module config is just a small thing for a provider, and it actually should not be heavily used by a provider. 
Do you think which scenarios this method/stage would be helpful?

wu-sheng: > why do you say "it actually should not be heavily used by a provider."?

I mean to indicate config initialized as a stage like prepare and start.

>I doubt that they need to be "helpful", all these are lifecycle methods and they should be invoked by the module manager only once.

My point is, I want to show config has its own initialization process, which doesn't related to prepare, start and notifyAfterCompleted
In the current design, the boundaries of these three are 
- prepare, initial things inside provider.
- start, booting with required modules initialized(prepared) services
- notifyAfterCompleted, process logic after all modules started.

If we add onConfigInitialized here, it means we should put some logic in it, which leads to the confusion about, which logic should be in it, and why others should be in prepare? They seem the same.

So, I added this config creator to set the boundaries that, only config relative things should be placed in here. Such as we need to initial some fields based on file configurations, because those fields are complex class, can't be simply just set a value.

I hope this explanation could make the reason clear why I even did more changes, rather than much simpler to add a new method on provider.

wankai123 (dismissed):
Once test passed

oap-server/server-library/library-module/src/main/java/org/apache/skywalking/oap/server/library/module/ModuleProvider.java
```diff
@@ -52,9 +52,28 @@ protected final ModuleManager getManager() {
     public abstract Class<? extends ModuleDefine> module();
 
     /**
-     *
+     * Create a config creator to initialize this configuration of this module provider
+     * @return creator instance to initialize the configuration with callback. Or return null if no config is required.
      */
-    public abstract ModuleConfig createConfigBeanIfAbsent();
+    public abstract ConfigCreator newConfigCreator();
```
kezhenxu94, line 58: There are tons of warnings (all implementations) because of the raw types, maybe add the generic type to eliminate them?

```suggestion
    public abstract ConfigCreator<? extends ModuleConfig> newConfigCreator();
```

kezhenxu94 (approved):
LGTM

wu-sheng:
I need to resolve the conflicts when I am back to my laptop

wankai123 (approved):

PR #9834, opened 2022-10-24
Fix typo in UTs

liuzc9:
fix misspelled words.


wankai123 (approved):

wu-sheng:
Are those all you planned to fix? Please submit all in once, rather than one by one.

liuzc9:
@wu-sheng  Ok

wu-sheng (approved):

PR #9839, opened 2022-10-24
Update BanyanDB storage module to adapt BanyanDB 0.2

### LINKED ISSUE #9805: [Feature] Upgrade BanyanDB storage module
### Search before asking

### Description

Since #9761 merged, we could kick off upgrading the OAP BanyanDB storage module.

The client lib introduces several improvements:

* `exist` endpoints of metadata registry. stream and measure `exist` methods' [return value](https://github.com/apache/skywalking-banyandb-java-client/blob/main/src/main/java/org/apache/skywalking/banyandb/v1/client/metadata/ResourceExist.java#L25
) has `hasGroup`, which will simplify the interaction of creating groups.
* `AbstractQuery`'s [`and` method](https://github.com/apache/skywalking-banyandb-java-client/blob/main/README.md#example-of-and) has the same semantic as before. We could leave it alone.
* `AbstractQuery`'s [`or` method](https://github.com/apache/skywalking-banyandb-java-client/blob/main/README.md#example-of-or) could fit the profile query scenario. 
* `Property`'s `apply`, `get` and `delete` operations support selecting tags. We could disable a UI template by updating a tag instead of a whole property.
* Tag specification supports the `index_only` flag. We could set trace segment's `searchable` tags to `index_only=true`.
* `WriteProcessor`'s `add` method could return a `CompletableFuture<Void>`. We could return it to the OAP's core module.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

lujiajing1126:
Signed-off-by: Megrez Lu <lujiajing1126@gmail.com>

Pending for client PR https://github.com/apache/skywalking-banyandb-java-client/pull/21

1. Adopt new `OR` logical operator for,
   - [x] `MeasureIDs` query
   - [x] `BanyanDBProfileThreadSnapshotQueryDAO` query
   - [x] Multiple Event conditions

2. Simplify Group check and creation
3. Partially apply `UITemplate` changes
4. Support `index_only`
5. Return `CompletableFuture<Void>` directly from BanyanDB client
6. Optimize data binary parse methods in `*LogQueryDAO`
7. Support indexType
8. Support configuration for TTL and (block|segment) intervals

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #9805 .


oap-server-bom/pom.xml
```diff
@@ -73,7 +73,7 @@
         <awaitility.version>3.0.0</awaitility.version>
         <httpcore.version>4.4.13</httpcore.version>
         <commons-compress.version>1.21</commons-compress.version>
-        <banyandb-java-client.version>0.1.0</banyandb-java-client.version>
+        <banyandb-java-client.version>0.2.0-SNAPSHOT</banyandb-java-client.version>
```
wu-sheng, line 76: You need a repository to get this version, I think?

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/MetadataRegistry.java
```diff
@@ -322,9 +323,9 @@ public Group getOrCreateGroup(BanyanDBClient client) throws BanyanDBException {
             }
             switch (kind) {
                 case STREAM:
-                    return client.define(Group.create(this.group, Catalog.STREAM, this.shard, 0, Duration.ofDays(7)));
+                    return client.define(Group.create(this.group, Catalog.STREAM, this.shard, IntervalRule.create(IntervalRule.Unit.HOUR, 4), IntervalRule.create(IntervalRule.Unit.DAY, 1), IntervalRule.create(IntervalRule.Unit.DAY, 7)));
                 case MEASURE:
-                    return client.define(Group.create(this.group, Catalog.MEASURE, this.shard, 12, Duration.ofDays(7)));
+                    return client.define(Group.create(this.group, Catalog.MEASURE, this.shard, IntervalRule.create(IntervalRule.Unit.HOUR, 4), IntervalRule.create(IntervalRule.Unit.HOUR, 24), IntervalRule.create(IntervalRule.Unit.DAY, 7)));
```
wu-sheng, line 328: I think we should provide these as storage settings. These are polices controlling segment and blocks creating, right?

hanahmily:
> Besides the last one(Zipkin), we should be able to fix those TODOs at this time, right? 0.2 BanyanDB server should have provided full features for OAP already.

Besides the double field, others could be stripped out.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/MetadataRegistry.java
lujiajing1126: Added

oap-server-bom/pom.xml
lujiajing1126: Yes. Pending util we publish java client `0.2.0`.

wu-sheng:
Is this ready? We only leave double and new top N to add in next version? If so, could you create an issue for 9.4.0 to track this?

lujiajing1126:
> Is this ready? We only leave double and new top N to add in next version? If so, could you create an issue for 9.4.0 to track this?

Yes. It is generally ready for review. I will fix CI and E2E.

Created #9864

hanahmily:
@kezhenxu94 skywalking-eye fails to handle some repos' license:

```
ERROR failed to identify the licenses of following packages (2):
build.buf.protoc-gen-validate:pgv-java-stub
build.buf.protoc-gen-validate:protoc-gen-validate 
Error: Process completed with exit code 1.
```

https://github.com/apache/skywalking/actions/runs/3342606435/jobs/5537555471

kezhenxu94:
> @kezhenxu94 skywalking-eye fails to handle some repos' license:
> 
> ```
> ERROR failed to identify the licenses of following packages (2):
> build.buf.protoc-gen-validate:pgv-java-stub
> build.buf.protoc-gen-validate:protoc-gen-validate 
> Error: Process completed with exit code 1.
> ```
> 
> https://github.com/apache/skywalking/actions/runs/3342606435/jobs/5537555471

There is no licensing information from the perspective of maven dependency, so for now I'm afraid you have to apply the patch:

```diff
diff --git a/.licenserc.yaml b/.licenserc.yaml
index e7a33ff930..ec0af00052 100644
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -119,3 +119,9 @@ dependency:
     - name: com.google.flatbuffers:flatbuffers-java
       version: 1.12.0
       license: Apache-2.0
+    - name: build.buf.protoc-gen-validate:pgv-java-stub
+      version: 0.6.13
+      license: Apache-2.0
+    - name: build.buf.protoc-gen-validate:protoc-gen-validate
+      version: 0.6.13
+      license: Apache-2.0
```

.licenserc.yaml
```diff
@@ -119,3 +119,9 @@ dependency:
     - name: com.google.flatbuffers:flatbuffers-java
       version: 1.12.0
       license: Apache-2.0
+    - name: build.buf.protoc-gen-validate:pgv-java-stub
+      version: 0.6.13
+      license: Apache-2.0
+    - name: build.buf.protoc-gen-validate:protoc-gen-validate
+      version: 0.6.13
+      license: Apache-2.0
```
wu-sheng, line 127: As these are added, you should update license files accordingly.

docs/en/changes/changes.md
```diff
@@ -3,6 +3,7 @@
 #### Project
 
 * Bump up the embedded `swctl` version in OAP Docker image.
+* Migrate to BanyanDB v0.2.0.
```
wu-sheng, line 6: I think this should be in `backend` only?

wu-sheng: Also, please make this a breaking change. BanyanDB server breaks the protocol, and we change the ways to store data.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/BanyanDB.java
```diff
@@ -91,4 +91,19 @@
...
+     *
+     * @since 9.3.0
+     */
+    @Target({ElementType.FIELD})
+    @Retention(RetentionPolicy.RUNTIME)
+    @interface IndexRule {
+        IndexType indexType() default IndexType.INVERTED;
+
+        enum IndexType {
+            INVERTED, TREE;
```
wu-sheng, line 106: Please provide a simple explanation for those two index types.

lujiajing1126: Added

docs/en/changes/changes.md
lujiajing1126: Added

.licenserc.yaml
lujiajing1126: LICENSE regenerated

wu-sheng:
According to the discussion between @hanahmily and me, we missed downsampling based index mechanism from v0.1.
We should fix this to make sure time dimension wouldn't be mixed.For example, 5pm has 3 values in three dimensions to represent  precisions. They are mixed for now.

@lujiajing1126 please correct me if I am wrong.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/MetadataRegistry.java
```diff
@@ -293,21 +335,41 @@ public SchemaMetadata parseMetadata(Model model, BanyanDBStorageConfig config) {
...
+         *
+         * @param modelName    name of the model
+         * @param downSampling not used if it is {@link DownSampling#None}
+         * @return entity (e.g. measure, stream) name
+         */
+        static String formatName(String modelName, DownSampling downSampling) {
+            if (downSampling == null || downSampling == DownSampling.None) {
+                return modelName;
+            }
+            return modelName + ":" + downSampling.getName();
```
hanahmily, line 363: Underscore "_"  is better and compatible with the metric name's convention.

lujiajing1126: Fixed

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBMetricsQueryDAO.java
```diff
@@ -205,19 +199,38 @@ public HeatMap readHeatMap(MetricsCondition condition, String valueColumnName, D
...
+
+        Map<String, DataPoint> map = new HashMap<>();
+        MeasureQueryResponse resp = query(schema, Collections.emptySet(), ImmutableSet.of(valueColumnName), timestampRange, new QueryBuilder<MeasureQuery>() {
+            @Override
+            protected void apply(MeasureQuery query) {
+                query.and(eq(Metrics.ENTITY_ID, entityID));
             }
+        });
+        for (final DataPoint dp : resp.getDataPoints()) {
+            map.putIfAbsent(dp.getId(), dp);
```
hanahmily, line 231: If a duplicated data point is here, please raise warnings to make some noise.

oap-server-bom/pom.xml
wu-sheng, line 76: TODO: Once PR is approved, we would wait for the BanyanDB client 0.2.0 released.

hanahmily (changes requested):

oap-server/server-starter/src/main/resources/application.yml
```diff
@@ -236,6 +236,12 @@ storage:
     superDatasetShardsFactor: ${SW_STORAGE_BANYANDB_SUPERDATASET_SHARDS_FACTOR:2}
     concurrentWriteThreads: ${SW_STORAGE_BANYANDB_CONCURRENT_WRITE_THREADS:15}
     profileTaskQueryMaxSize: ${SW_STORAGE_BANYANDB_PROFILE_TASK_QUERY_MAX_SIZE:200} # the max number of fetch task in a request
+    streamBlockInterval: ${SW_STORAGE_BANYANDB_STREAM_BLOCK_INTERVAL:4} # Unit is Hours
+    streamSegmentInterval: ${SW_STORAGE_BANYANDB_STREAM_SEGMENT_INTERVAL:24} # Unit is Hours
+    streamTTL: ${SW_STORAGE_BANYANDB_STREAM_TTL:24} # Unit is Days
```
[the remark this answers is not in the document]
wu-sheng: Yes, `ConfigService` is exposed from the Core module.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBConverter.java
```diff
@@ -215,11 +218,19 @@ public static class StorageToMeasure implements Convert2Entity {
...
+            this.schema = schema;
             this.dataPoint = dataPoint;
         }
 
+        /**
+         * Currently, only used by {@link org.apache.skywalking.oap.server.storage.plugin.banyandb.measure.BanyanDBNetworkAddressAliasDAO}.
+         * The default DownSampling strategy, i.e. {@link DownSampling#Minute} is assumed in this case.
+         */
+        public StorageToMeasure(String measureModelName, DataPoint dataPoint) {
+            this(checkNotNull(MetadataRegistry.INSTANCE.findMetadata(measureModelName, DownSampling.Minute), "measure schema"), dataPoint);
```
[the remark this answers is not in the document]
wu-sheng: What assumption? Could you be more specific?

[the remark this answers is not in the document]
wu-sheng: I read the comments, why don't we move this special case into the BanyanDBNetworkAddressAliasDAO? This seems a thing not belonging to the converter.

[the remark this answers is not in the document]
wu-sheng: > What assumption? Could you be more specific?

Does this mean `DownSampling == Minute`? If so, for this specific case, yes, it is correct. `NetworkAddressAlias` has this annotation `@MetricsExtension(supportDownSampling = false, supportUpdate = true)`.

`*Traffic` has the same annotation for no downsampling too. This is a very special annotation, for now, it is used for models having their own DAOs.

oap-server/server-starter/src/main/resources/application.yml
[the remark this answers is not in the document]
lujiajing1126: Fixed

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/stream/BanyanDBProfileThreadSnapshotQueryDAO.java
```diff
@@ -105,43 +103,40 @@ public void apply(StreamQuery query) {
...
-                .collect(Collectors.toList());
+        final StreamQueryResponse segmentRecordResp = query(SegmentRecord.INDEX_NAME,
+                TAGS_TRACE,
+                new QueryBuilder<StreamQuery>() {
+                    @Override
+                    public void apply(StreamQuery traceQuery) {
+                        for (final String segmentID : segmentIds) {
+                            traceQuery.or(eq(SegmentRecord.SEGMENT_ID, segmentID));
+                            traceQuery.setLimit(segmentIds.size());
+                            traceQuery.setOrderBy(desc(SegmentRecord.START_TIME));
```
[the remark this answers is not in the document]
lujiajing1126: Fixed

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBConverter.java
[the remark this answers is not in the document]
lujiajing1126: Moved to `BanyanDBNetworkAddressAliasDAO`

docs/en/changes/changes.md
```diff
@@ -93,7 +105,7 @@
   to change the OAP backend service addresses, like `SW_OAP_ADDRESS=localhost:12800,localhost:12801`, and use
   environment
   variable `SW_SERVER_PORT` to change the port. Other Spring-related configurations don't take effect anymore.
-* Polish the endpoint list graph.
+* Polish the endpoint list graph.g
```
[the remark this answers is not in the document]
lujiajing1126: Oh, deleted

oap-server/server-starter/src/main/resources/application.yml
```diff
@@ -236,6 +236,10 @@ storage:
     superDatasetShardsFactor: ${SW_STORAGE_BANYANDB_SUPERDATASET_SHARDS_FACTOR:2}
     concurrentWriteThreads: ${SW_STORAGE_BANYANDB_CONCURRENT_WRITE_THREADS:15}
     profileTaskQueryMaxSize: ${SW_STORAGE_BANYANDB_PROFILE_TASK_QUERY_MAX_SIZE:200} # the max number of fetch task in a request
+    streamBlockInterval: ${SW_STORAGE_BANYANDB_STREAM_BLOCK_INTERVAL:4} # Unit is Hours
+    streamSegmentInterval: ${SW_STORAGE_BANYANDB_STREAM_SEGMENT_INTERVAL:24} # Unit is Hours
+    measureBlockInterval: ${SW_STORAGE_BANYANDB_MEASURE_BLOCK_INTERVAL:4} # Unit is Hours
+    measureSegmentInterval: ${SW_STORAGE_BANYANDB_MEASURE_SEGMENT_INTERVAL:24} # Unit is Hours
```
wu-sheng, line 242: backend-storage.md should be updated as well.

wu-sheng:
```suggestion
    streamBlockInterval: ${SW_STORAGE_BANYANDB_STREAM_BLOCK_INTERVAL:4} # Unit is hour
    streamSegmentInterval: ${SW_STORAGE_BANYANDB_STREAM_SEGMENT_INTERVAL:24} # Unit is hour
    measureBlockInterval: ${SW_STORAGE_BANYANDB_MEASURE_BLOCK_INTERVAL:4} # Unit is hour
    measureSegmentInterval: ${SW_STORAGE_BANYANDB_MEASURE_SEGMENT_INTERVAL:24} # Unit is hour
```

H should not be in upper case, and don't need `s`?

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBIndexInstaller.java
```diff
@@ -35,28 +36,32 @@
...
+            final boolean resourceExist = metadata.checkResourceExistence(c);
+            if (!resourceExist) {
+                return false;
             }
-            log.info("group {} created", g.name());
+
             // then check entity schema
             if (metadata.findRemoteSchema(c).isPresent()) {
-                MetadataRegistry.INSTANCE.registerModel(model, config);
+                MetadataRegistry.INSTANCE.registerModel(model, config, configService);
```
wu-sheng, line 60: I think we should keep existence checking here only. Otherwise, what is the point of `createTable`?

wu-sheng: Also, you have this 

```
// normally, we should not reach here
return false;
```

If it can't be reached, we should face an IllegalStateException

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBIndexInstaller.java
```diff
@@ -67,13 +72,13 @@ protected boolean isExists(Model model) throws StorageException {
...
             if (model.isTimeSeries() && model.isRecord()) { // stream
-                Stream stream = (Stream) MetadataRegistry.INSTANCE.registerModel(model, config);
+                Stream stream = (Stream) MetadataRegistry.INSTANCE.registerModel(model, config, configService);
                 if (stream != null) {
                     log.info("install stream schema {}", model.getName());
                     ((BanyanDBStorageClient) client).define(stream);
                 }
             } else if (model.isTimeSeries() && !model.isRecord()) { // measure
-                Measure measure = (Measure) MetadataRegistry.INSTANCE.registerModel(model, config);
+                Measure measure = (Measure) MetadataRegistry.INSTANCE.registerModel(model, config, configService);
```
wu-sheng, line 81: I think we should separate `registerStreamModel` and `registerMeasureModel` to keep API clear, especially for the `return type`.

And where would we create the object when `!model.isTimeSeries()`? Or it is not required to be pre-created?

oap-server/server-starter/src/main/resources/application.yml
wu-sheng: > backend-storage.md should be updated as well.

There is a TODO doc fix

lujiajing1126: Fixed

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBIndexInstaller.java
lujiajing1126: Since I've removed this `registerModel` action here, it is now simplified

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBMetricsQueryDAO.java
lujiajing1126: Done

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBIndexInstaller.java
wu-sheng: If it isn't required, please add comments here to make this clear.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBIndexInstaller.java
lujiajing1126: This line cannot be removed. Since even if schema is already registered in the BanyanDB, we still have to register it locally for query and write.

So these lines are restored.

wu-sheng: Do we have a better to make codes more clear? Such as, when we do `isPresent`, could we use the metadata of existing measure/stream to build this local registration automatically?

lujiajing1126: > Do we have a better to make codes more clear? Such as, when we do `isPresent`, could we use the metadata of existing measure/stream to build this local registration automatically?

I don't think it is enough to extract all necessary metadata with the schema in the BanyanDB. For example, 

```java
    public static class StreamToStorage implements Convert2Storage<StreamWrite> {
...

        @Override
        public void accept(String fieldName, Object fieldValue) {
            MetadataRegistry.ColumnSpec columnSpec = this.schema.getSpec(fieldName);
            if (columnSpec == null) {
                throw new IllegalArgumentException("fail to find field[" + fieldName + "]");
            }
            try {
                this.streamWrite.tag(fieldName, buildTag(fieldValue, columnSpec.getColumnClass()));
            } catch (BanyanDBException ex) {
                log.error("fail to add tag", ex);
            }
        }
    }

    private static Serializable<BanyandbModel.TagValue> buildTag(Object value, final Class<?> clazz) {
        if (int.class.equals(clazz) || Integer.class.equals(clazz)) {
            return TagAndValue.longTagValue(((Number) value).longValue());
        } else if (Long.class.equals(clazz) || long.class.equals(clazz)) {
            return TagAndValue.longTagValue((Long) value);
        } else if (String.class.equals(clazz)) {
            return TagAndValue.stringTagValue((String) value);
        } else if (Double.class.equals(clazz) || double.class.equals(clazz)) {
            return TagAndValue.binaryTagValue(ByteUtil.double2Bytes((double) value));
        } else if (StorageDataComplexObject.class.isAssignableFrom(clazz)) {
            return TagAndValue.stringTagValue(((StorageDataComplexObject<?>) value).toStorageData());
        } else if (clazz.isEnum()) {
            return TagAndValue.longTagValue((int) value);
        } else if (JsonObject.class.equals(clazz)) {
            return TagAndValue.stringTagValue((String) value);
        } else if (byte[].class.equals(clazz)) {
            return TagAndValue.stringTagValue((String) value);
        }
        throw new IllegalStateException(clazz.getSimpleName() + " is not supported");
    }
```

The mapping between data types in BanyanDB and OAP is not a surjection.

wu-sheng: I see, then how about we separating local registration logic as new methods?
Remote registration could call local registration automatically for sure.

lujiajing1126: If we want to do so, I would suggest we add more overloading functions to the `Convert2Storage` interface. For example,

```java
public interface Convert2Storage<R> {
    void accept(String fieldName, String fieldValue);
    void accept(String fieldName, int fieldValue);
    void accept(String fieldName, long fieldValue);
    void accept(String fieldName, double fieldValue);
// etc.
}
```

One for each column type in OAP, so the type can be determined at compile time.

lujiajing1126: > I see, then how about we separating local registration logic as new methods? Remote registration could call local registration automatically for sure.

`registerMeasureModel` and `registerStreamModel` are only responsible for registering local cache.

The `define` method is for remote registration.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBIndexInstaller.java
hanahmily: We will create a validation to check it as https://github.com/apache/skywalking/issues/9902. Please add comments here.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBIndexInstaller.java
wu-sheng: Got it. My fault. we should be good now.

wu-sheng (approved):
Generally, I am good for this PR.

If there is no more issue to fix, I think it is time to release the BanyanDB server and Java client. Then we should be able to merge this one.

Congrats @lujiajing1126 @hanahmily

wu-sheng:
FYI @wankai123 This would help a lot to finish BanyanDB storage for Zipkin part(for query)

hanahmily (approved):

PR #9843, opened 2022-10-25
Do not log grpc envoy receiver client-side streaming cancellation as errors

### LINKED ISSUE #9840: [Bug] io.grpc.StatusRuntimeException: CANCELLED: client cancelled
### Search before asking

### What happened

When AccessLogServiceGRPC is used as a receiver, and a pod is terminated (either killed, or because of a rolling update), the sidecar that generates the ALS entries and sends them to OAP cancels the client-side grpc stream.
On the server side, when it receives the client cancelled status, `AccessLogServiceGRPC` handles the Throwable and logs it at ERROR level:
https://github.com/apache/skywalking/blob/9d3982db74dbe7ba9c028aef674ace13fb74cc18/oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/AccessLogServiceGRPCHandler.java#L173

In environments with Continuous deployment and high commit rate, starting new pods and terminating other can be quite frequent causing very noise logs with error messages  than can be cause confusions. 

When the client cancelled the stream, log is a debug.
When any other error happens, log is as an error as before.

### How to reproduce

Use the skywalking  showcase with mesh enable. Kill a pod and you will see this error message.

### Anything else

_No response_

### Code of Conduct

toffentoffen:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #9840.

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -69,7 +69,7 @@
 * [**Breaking Change**] Support new records query protocol, rename the column named `service_id` to `entity_id` for support difference entity.
   Please re-create `top_n_database_statement` index/table.
 * Remove improper self-obs metrics in JvmMetricsHandler(for Kafka channel).
-* Client-side grpc stream envoy metrics receiver is not logged as an error when client cancels the stream. Client 
+* gRPC stream canceling code is not logged as an error when the client cancels the stream. The client 
   cancels the stream when the pod is terminated.
 
 #### UI
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -69,6 +69,8 @@
 * [**Breaking Change**] Support new records query protocol, rename the column named `service_id` to `entity_id` for support difference entity.
   Please re-create `top_n_database_statement` index/table.
 * Remove improper self-obs metrics in JvmMetricsHandler(for Kafka channel).
+* Client-side grpc stream envoy metrics receiver is not logged as an error when client cancels the stream. Client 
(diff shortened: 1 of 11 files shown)


wu-sheng:
I think there are many other gRPC handlers that should be fixed as well. Could you fix them in this PR once?

kezhenxu94 (dismissed):
LGTM

toffentoffen:
> I think there are many other gRPC handlers that should be fixed as well. Could you fix them in this PR once?

Sure @wu-sheng Will fix them as well.

docs/en/changes/changes.md
```diff
@@ -69,6 +69,8 @@
 * [**Breaking Change**] Support new records query protocol, rename the column named `service_id` to `entity_id` for support difference entity.
   Please re-create `top_n_database_statement` index/table.
 * Remove improper self-obs metrics in JvmMetricsHandler(for Kafka channel).
+* Client-side grpc stream envoy metrics receiver is not logged as an error when client cancels the stream. Client 
+  cancels the stream when the pod is terminated.
```
wu-sheng, line 73:
```suggestion
* gRPC stream canceling code is not logged as an error when the client cancels the stream. The client 
  cancels the stream when the pod is terminated.
```

wu-sheng (approved):
LGTM. I just updated the change log for you @toffentoffen

wu-sheng:
@kezhenxu94 I think commit log is forgotten to update accordingly.

PR #9851, opened 2022-10-26
Split TCP related service mesh metrics protobuffer definition

kezhenxu94:


wu-sheng:
As upstream is not merged yet, do we have connection ID logic in this PR?

kezhenxu94:
> As upstream is not merged yet, do we have connection ID logic in this PR?

This PR doesn't include the fix, it's essentially a refactor and all works just like before.

When the upstream Envoy fix is merged, we can start to tackle the connection ID and that see what else is needed for that case.

I'm also OK to postpone this PR until when upstream Envoy fix gets merged, just wanted to minimize the changes in a single PR and separate these two things.

wu-sheng:
Let's merge it first. Separating sources could avoid a lot of noises for agent and HTTP analysis in the future. I just want to confirm it.

oap-server/server-starter/src/main/resources/oal/core.oal
```diff
@@ -16,9 +16,6 @@
  *
  */
 
-// For services using protocols HTTP 1/2, gRPC, RPC, etc., the cpm metrics means "calls per minute",
-// for services that are built on top of TCP, the cpm means "packages per minute".
-
```
wu-sheng, line 21: I remember we had some UI dashboards mentioning these too. Could you recheck?

wu-sheng: They should be on Mesh dashboard

wu-sheng:
Are you updating UI submodule by mistake?

kezhenxu94:
> Are you updating UI submodule by mistake?

Yes reverted.

wu-sheng (approved):

wu-sheng:
It seems you need to resolve conflicts manually.

wu-sheng (approved):

PR #9855, opened 2022-10-27
Add virtual MQ analysis for native traces

pg-yang:

Closes https://github.com/apache/skywalking/issues/9676

This PR contains 3 modification as following :
### Remove legacy relative mq analysis from OAL , ui  template 

image

image

### Add virtual MQ analysis

image

image

image


oap-server/server-starter/src/main/resources/oal/core.oal
```diff
@@ -25,8 +25,6 @@ service_sla = from(Service.*).percent(status == true);
 service_cpm = from(Service.*).cpm();
 service_percentile = from(Service.latency).percentile(10); // Multiple values including p50, p75, p90, p95, p99
 service_apdex = from(Service.latency).apdex(name, status);
-service_mq_consume_count = from(Service.*).filter(type == RequestType.MQ).count();
```
wu-sheng, line 28: I think these should be kept. As this is still meaningful from a service perspective.

pg-yang: I  have restored it
image

wu-sheng: I am afraid you didn't revert `general-endpoint` dashboard?

pg-yang: endpoint_mq_consume_latency , endpoint_mq_consume_count should be demonstrated in  virtual mq  and  endpoint dashboard  at the same time ？

wu-sheng: The endpoint of the service is not the endpoint of the virtual queue service. This is not showing one thing in two places. There are two dimensions statistics, right?

pg-yang: We provide virtual service to demonstrate special statistics , and should  keep `General-Endpoint` common (`General-Service` also is )

The following screenshots from my local env 

image

int  another aspect  , analyzing `endpoint of the service `  and  `endpoint of the virtual queue service` would produce repeat logic/calculation

wu-sheng: > int another aspect , analyzing endpoint of the service and endpoint of the virtual queue service would produce repeat logic/calculation

What do you mean by repeat?
Let's put it this way. Queue-A is consumed by serv1 and serv2 at the same time.
1. From the service perspective, serv1./queue-a is the entity of the metrics.
2. From the virtual MQ perspective, queue-a is an endpoint of the MQ service(IP based).

Those are different. I don't think this is a repeated calculation. The case it is duplicated only when a MQ only have one consumer service.

pg-yang: Sorry,  It's two dimensional .
And  I  have another  question .

image

As  above screenshot , there is an endpoint named `Kafka/topic/Consumer/a` , in this scenario , `Endpoint Load (calls / min)
` is  similar with `Message Queue Consuming Count ` except  one is `CPM` , another is `Count ` .   Right ?

wu-sheng: I think those are really some kinds of duplicates. I didn't notice this.

It should be fine to remove MQ's consuming count, meanwhile, we should update the title and tips for Endpoint CPM metric to mention it has various meanings for different cases.

pg-yang: And do we have another kind of `transmission.latency` ? 
I  think we should change
 `endpoint_mq_consume_latency = from((str->long)Endpoint.tag["transmission.latency"]).filter(type == RequestType.MQ).filter(tag["transmission.latency"] != null).longAvg();`    

to  

`endpoint_transmission_latency = from((str->long)Endpoint.tag["transmission.latency"]).filter(tag["transmission.latency"] != null).longAvg();`

wu-sheng: I think both are fine. The only difference is `.filter(type == RequestType.MQ)`, which is faster than not null. As in OAL compiling, `.filter(tag["transmission.latency"] != null)` means to read a value from a map, but the `type` is a simple `getFromField`.

pg-yang: The  main purpose is to make the metrics as  `transmission.latency metrics ` rather than `mq  transmission.latency metrics ` , like above example  the endpoint name should distinguish whether it's  related with MQ

wu-sheng: Then I think endpoint_mq_consume_latency is better. Because it indicates MQ endpoint only.

pg-yang: Yes , It's better if we only have MQ transmission latency , if we would provide other kind of `transmission latency` such as  `HTTP transmission latency`  in  future , `endpoint_transmission_latency ` should be  better . in order to make `transmission latency` more  flexible , I  remove `.filter(type == RequestType.MQ)`

wu-sheng: HTTP transmission latency means the client side has to inject the timestamp of the client side and calculate the duration on the server side. But different from MQ, HTTP latency is usually less than 5ms, then time sync between client and server matters a lot, in the real prod env, I don't believe any infra team could set this up.

So, this is why the `transmission latency` is specifically for MQ only.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/vservice/VirtualMQProcessor.java
```diff
@@ -0,0 +1,139 @@
...
+        endpointAccess.setEndpoint(endpointName);
+        sourceList.add(endpointAccess);
+    }
+
+    private String buildEndpointName(String topic, String queue) {
+        return Arrays.stream(new String[] {
+                         topic,
+                         queue
+                     }).filter(StringUtil::isNotBlank)
+                     .reduce((a, b) -> a + "/" + b).orElse("");
```
kezhenxu94, line 90:
```suggestion
        return Stream.of(topic, queue)
                     .filter(StringUtil::isNotBlank)
                     .reduce((a, b) -> a + "/" + b).orElse("");
```

kezhenxu94 (commented):
Generally LGTM, a nit, can you please also update `docs/en/concepts-and-designs/scope-definitions.md`?

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/vservice/VirtualMQProcessor.java
pg-yang: Yes,   I forgot it .

oap-server/server-starter/src/main/resources/ui-initialized-templates/general/general-endpoint.json
```diff
@@ -273,7 +227,8 @@
                   "type": "Widget",
                   "widget": {
                     "title": "Endpoint Load (calls / min)",
-                    "name": "Endpoint_Load"
+                    "name": "Endpoint_Load",
+                    "tips": "This metrics means incoming requests load (e.g. HTTP Request, MQ message consume, etc), "
```
wu-sheng, line 231:
```suggestion
                    "tips": "requests(HTTP / RPC) / min, consuming messages(MQ) / min"
```

wu-sheng:
@pg-yang We need some fixes for this PR.

pg-yang:
> @pg-yang We need some fixes for this PR.

Got ,  I think  I could finish it at this  weekend.

wu-sheng:
Could you update the PR descriptions to match the latest changes?

pg-yang:
1. update scope definitions doc 
2. update endpoint list  in  ` virtual-mq-service.json ` to fix  incorrect metrics name 

image



3. restore `endpoint_mq_consume_latency` metrics in core.oal  and `general-endpoint`

image

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/MQEndpointAccess.java
```diff
@@ -0,0 +1,72 @@
...
+@ScopeDefaultColumn.VirtualColumnDefinition(fieldName = "entityId", columnName = "entity_id", isID = true, type = String.class)
+public class MQEndpointAccess extends Source {
+    @Override
+    public int scope() {
+        return MESSAGE_QUEUE_ENDPOINT_ACCESS;
+    }
+
+    @Override
+    public String getEntityId() {
+        return IDManager.EndpointID.buildId(serviceId, endpoint);
```
wu-sheng, line 37: You could refer to Endpoint#entityId, a field should be used to cache this value in case of encoding the entity ID repeatedly.

I noticed `DatabaseAccess` and `CacheAccess` don't have it either, please add them. 

In the OAL compiling, the `getEntityId` method would be called repeatedly in every new generated dispatcher, so an extra field helps.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/source/MQEndpointAccess.java
```diff
@@ -47,10 +52,6 @@ public String getEntityId() {
     @ScopeDefaultColumn.DefinedByField(columnName = "endpoint", requireDynamicActive = true)
     private String endpoint;
 
-    @Getter
-    @ScopeDefaultColumn.DefinedByField(columnName = "service_id")
-    private String serviceId;
```
wu-sheng: I remember we need this field in final entity to support endpoint top N query with parent service ID, don't we?

pg-yang: Sorry , what's  `final entity  ` ? , I couldn't  understand it   ,  I don't provide top N metrics  at least now .  and  I remove `serviceId` because it  dose not affect virtual mq metrics

wu-sheng: You don't provide, but you should make sure it is supported, as it should from protocol perspective.
All endpoint and instance metrics should be.

Please recheck all recent added sources.

pg-yang: Yes,  I have checked other virtual service , only mq have endpoint/instance

wu-sheng (approved):
LGTM @kezhenxu94 Please recheck

docs/en/setup/service-agent/virtual-mq.md
```diff
@@ -0,0 +1,15 @@
+# Virtual Message Queue (MQ)
+
+Virtual MQ represent the MQ nodes detected by [server agents' plugins](server-agents.md). The performance
+metrics of the MQ are also from the MQ client-side perspective.
+
+For example, Kafka plugins in the Java agent could detect the transformation latency of message
```
kezhenxu94, line 6: Let's use consistent words with other places `transmissionLatency` like this

```suggestion
For example, Kafka plugins in the Java agent could detect the transmission latency of message
```

docs/en/setup/service-agent/virtual-mq.md
```diff
@@ -0,0 +1,15 @@
...
+As a result, SkyWalking would show message count , transformation latency, success rate powered by backend analysis capabilities in this dashboard.
```
kezhenxu94, line 7:
```suggestion
As a result, SkyWalking would show message count, transmission latency, success rate powered by backend analysis capabilities in this dashboard.
```

kezhenxu94 (dismissed):
Thanks for the high quality PR!!

kezhenxu94 (approved):

wu-sheng:
@pg-yang We need another to manual sync again, as https://github.com/apache/skywalking/pull/9867 merged.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/vservice/VirtualMQProcessor.java
```diff
@@ -0,0 +1,137 @@
...
+    public void prepareVSIfNecessary(final SpanObject span, final SegmentObject segmentObject) {
+        if (span.getSpanLayer() != SpanLayer.MQ) {
+            return;
+        }
+        MQTags mqTags = collectTags(span.getTagsList());
+        if (StringUtil.isBlank(mqTags.broker)) {
+            return;
+        }
+        long timeBucket = TimeBucket.getMinuteTimeBucket(span.getStartTime());
+        String serviceName = namingControl.formatServiceName(mqTags.broker);
```
wu-sheng, line 58: This may relate to https://github.com/apache/skywalking-java/pull/369#discussion_r1009015114 too.

Is broker address or peer a better value for service name? Ideally, we expect a list to represent multiple nodes.

wu-sheng:
@pg-yang Could you check, are peer and broker addr of Kafka and ActiveMQ same? And use the cluster addr list?

wu-sheng:
I left you some messages on slack.

pg-yang:
> @pg-yang Could you check, are peer and broker addr of Kafka and ActiveMQ same? And use the cluster addr list?

Kafka uses addr  list  as broker tag value , which reads from configuration properties `bootstrap.servers` , it's correct .

AcviteMQ is complex ,  ActiveMQ  plugin read addr through `session.getConnection().getTransport().getRemoteAddress()` .
 The method  `getRemoteAddress` in  `FailoverTransport` (an  implementation of `Transport`) means get current server address  . In a word , AcviteMQ plugin couldn't get cluster addr  list in some case

wu-sheng:
> Kafka uses addr list as broker tag value , which reads from configuration properties bootstrap.servers , it's correct .

This is a good practice. Could you try to consistent these plugins? Read the broker addr list from the initialization configuration?

pg-yang:
Yes , I would try to fix it .

wu-sheng:
`peer` is used for Topology build, so, how does Kafka's span peer look like? How about we use `peer` as a service like how we do for the virtual database? And leave broker_addr as they are for now?

pg-yang:
> `peer` is used for Topology build, so, how does Kafka's span peer look like? How about we use `peer` as a service like how we do for the virtual database? And leave broker_addr as they are for now?

Both of peer (at producer side ) and broker tag are same  , for example kafka , nats ,rabbit ,
I choose broker tag as service name  because EntrySpan (at  consumer side ) don't have peer  ,  and  I could read server address through consistent way `broker tag`  , whether at consumer side or producer side

pg-yang:
OK , I would check the peer at consumer side , and use peer as service  name  if it's suitable .

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/vservice/VirtualMQProcessor.java
```diff
@@ -51,17 +52,25 @@ public void prepareVSIfNecessary(final SpanObject span, final SegmentObject segm
...
+        final String serviceName;
+        if (span.getSpanType() == SpanType.Entry) {
+            mqOperation = MQOperation.Consume;
+            final String peer = span.getRefsList()
+                                    .stream()
+                                    .findFirst()
+                                    .map(SegmentReference::getNetworkAddressUsedAtPeer)
+                                    .orElse(null);
+            serviceName = namingControl.formatServiceName(peer);
+        } else {
```
wu-sheng, line 65: Should we have an `else if` here?

pg-yang: Yes ， some plugin would create localSpan .  we only need exit entry span .

wu-sheng (dismissed):

docs/en/setup/service-agent/virtual-mq.md
```diff
@@ -0,0 +1,15 @@
...
+Virtual MQ represent the MQ nodes detected by [server agents' plugins](server-agents.md). The performance
+metrics of the MQ are also from the MQ client-side perspective.
+
+For example, Kafka plugins in the Java agent could detect the transmission latency of message
+As a result, SkyWalking would show message count, transmission latency, success rate powered by backend analysis capabilities in this dashboard.
+
+The MQ operation span should have
+- It is an **Exit**(at producer side) or **Entry**(at consumer side) span
+- **Span's layer == MQ**
+- Tag key = `mq.broker`, value = MQ addresses, it must contain complete server address, and keep consistent at consumer, producer side.
```
pg-yang, line 12: The doc needs to update , I  would remove  this line

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/SpanTags.java
```diff
@@ -45,6 +45,14 @@ public class SpanTags {
 
     public static final String CACHE_KEY = "cache.key";
 
+    public static final String MQ_BROKER = "mq.broker";
```
wu-sheng, line 48: You remind me, this should be remove as well.

wu-sheng (dismissed):

kezhenxu94 (approved):

wu-sheng:
@kezhenxu94 @pg-yang Do we have cases in the showcase about Kafka?

kezhenxu94:
> @kezhenxu94 @pg-yang Do we have cases in the showcase about Kafka? 

Not now

wu-sheng:
Should we consider adding it? Is the resource cost reasonable?

kezhenxu94:
> Should we consider adding it? Is the resource cost reasonable?

It's not worthwhile to waste too many resource to fill this dashboard, but if we can manage to only use say at most 2CPU and 2G to set this up, I think we can do this.

Not sure whether 2CPU and 2G is enough to run the message queue, maybe we can try Pulsar standalone mode.

pg-yang:
Kafka would take more resource , and  I didn't  know Pulsar is MQ , so maybe I should confirm Pulsar plugin whether  follow virtual MQ  specification

wu-sheng:
Yes, pulsar is a MQ. 

If Kafka costs too much, maybe activeMQ? This is not a hard requirement though.

wu-sheng:
> Although sometimes it couldn't record complete server addresses 

What do you mean by `sometimes`? I think the agent's behavior is predictable. `Sometimes` seems scary.

pg-yang:
> AcviteMQ is complex , ActiveMQ plugin read addr through session.getConnection().getTransport().getRemoteAddress() .
The method getRemoteAddress in FailoverTransport (an implementation of Transport) means get current server address . In a word , AcviteMQ plugin couldn't get cluster addr list in some case

Here  . Our test case only include  common case  , if the test  scenario involve   `FailoverTransport` , the agent couldn't  record complete addresses

wu-sheng:
I see. You mean that plugin is robust. That is fine, once you could confirm this would not cause
1. Topology shows some strange nodes, such as a node named `null` or a random IP
2. Cause tracing context leak, and we would not have trace and metric to show anymore.

PR #9857, opened 2022-10-28
Support `sampledTrace` in LAL

mrproliu:
* Support `sampledTrace` statement in LAL and add some configuration for the network profiling analysis. Since the network profiling in Rover may analysis processes from different layers, multiple LAL configurations need to be added. 
* Support multiple rules with different names under the same layer of LAL script. 

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Part of https://github.com/apache/skywalking/issues/9802.


wu-sheng:
How about `sampleTrace` rather than `slowTrace`? SlowStatement is a thing we built for SQL database, as we only would collect that.
But in ebpf, we may sample all requests with trace headers, rather than only slow, right?
I prefer we could provide a more general use name.
And in the record, it would be SampledTraceRecord, with a column of sampling reason is `slow`. WDYT?

mrproliu:
> But in ebpf, we may sample all requests with trace headers, rather than only slow, right?

Yes, In the eBPF we can sample all kinds of requests with trace headers, for now, we only have the `slow` cases. There may be more in the future.

 > And in the record, it would be SampledTraceRecord, with a column of sampling reason is slow. WDYT?

It's good to be, when the rover agent uploads the logs, they need to add a `reason` field to confirm the sampling reason. But for the GraphQL query, for now, we only could be ordering the record. We need a filter mechanism to filter the `reason` filed.

wu-sheng:
We could build a format of sampled trace for this kind of data yo include reason. More fields seem hard to maintain.

mrproliu:
But we can't control the table/index size of record, right? When querying the sampling trace, we can't be knowing which data is displayed first. Unless we query a group field list, then try to query each top N with the group name? 
Otherwise, we can only specify several cause types and then create the corresponding tables/index.

wu-sheng:
https://github.com/apache/skywalking-query-protocol/blob/master/record.graphqls

If you mean Order in the condition, we could consider it as optional? As we could have slow trace or other reason(another metric) to sample?
What do you think?

mrproliu:
I think the order field is necessary. The record tables only insert the sampled traces, but maybe not ordering. 
Such as we have the following records:
```
{"reason": "slow", "name": "slow-/requestURI", "latency": 90}
{"reason": "slow", "name": "slow-/requestURI", "latency": 80}
{"reason": "slow", "name": "slow-/requestURI", "latency": 100}
{"reason": "error", "name": "error-/requestURI", "latency": 70}
```

Then, If we want to query the sampled records with top2. we may face the following issues:
1.  If not ordering, the top <N> records are not correct. Because we just insert the record in the OAP and order the records when query.
2. If ordering, we may only see a part of the data or missing data(in this case, the error reason records are missing). So I recommend we query the reason list first(in this case, it will be "slow" and "error"), then query each top N record with the same reason.

wu-sheng:
`caused_by` could be considered as a metric name, so we could have the slow trace sampled record or others. We could indicate this as a parameter of `sampledTrace(sampleReason)` function in LAL.

A question, the collecting HTTP body would not be reported this way, right? It goes through `SpanAttachedEventReportService` directly, rover just reports trace ID here.

mrproliu:
> `caused_by` could be considered as a metric name, so we could have the slow trace sampled record or others. We could indicate this as a parameter of `sampledTrace(sampleReason)` function in LAL.

Do you mean to fix the reason name in LAL? such as `sampledTrace("slow") {}`? For now, the LAL haven't the pre-load ability to read the method value for creating an index/table when OAP starts. 

> A question, the collecting HTTP body would not be reported this way, right? It goes through `SpanAttachedEventReportService` directly, rover just reports trace ID here.

Yes, the log only reporters the trace ID, and reports the HTTP body through `SpanAttachedEventReportService`.

wu-sheng:
> > `caused_by` could be considered as a metric name, so we could have the slow trace sampled record or others. We could indicate this as a parameter of `sampledTrace(sampleReason)` function in LAL.
> 
> Do you mean to fix the reason name in LAL? such as `sampledTrace("slow") {}`? For now, the LAL haven't the pre-load ability to read the method value for creating an index/table when OAP starts.

I don't have to be that complex. I just mean `slow` could be an enum or string, we just need an IF/ELSE in the LAL kernel. 
The point of proposing this is that, we may need to extend later, but we may don't want to keep adding a new LAL method to cause confusing. The record means physical tables in SQL database, and an index in un-merging mode(before 9.2.0, and still supported now). This could not be a light concept, and loaded through LAL. So, don't worry for that.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/Binding.java
```diff
@@ -107,6 +110,15 @@ public Binding databaseSlowStatement(DatabaseSlowStatementBuilder databaseSlowSt
         return this;
     }
 
+    public SampleTraceBuilder sampleTraceBuilder() {
+        return (SampleTraceBuilder) getProperty(KEY_SAMPLE_TRACE);
+    }
+
+    public Binding sampleTrace(SampleTraceBuilder sampleTraceBuilder) {
```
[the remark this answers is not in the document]
mrproliu: @sonatype-lift ignoreall

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/trace/SampleTraceSlowRecord.java
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.skywalking.oap.server.core.storage.type.Convert2Storage;
+import org.apache.skywalking.oap.server.core.storage.type.StorageBuilder;
+
+import static org.apache.skywalking.oap.server.core.source.DefaultScopeDefine.SAMPLE_TRACE_SLOW;
+
+@Setter
+@Getter
+@ScopeDeclaration(id = SAMPLE_TRACE_SLOW, name = "SampleTraceSlowRecord")
+@Stream(name = SampleTraceSlowRecord.INDEX_NAME, scopeId = SAMPLE_TRACE_SLOW, builder = SampleTraceSlowRecord.Builder.class, processor = RecordStreamProcessor.class)
+public class SampleTraceSlowRecord extends Record {
```
wu-sheng, line 41:
```suggestion
public class SampledSlowTraceRecord extends Record {
```

Please adjust the class, index name, and document accordingly. **Sampled slow trace record**s match the purpose more.

mrproliu: Updated.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/Binding.java
```diff
@@ -107,6 +110,15 @@ public Binding databaseSlowStatement(DatabaseSlowStatementBuilder databaseSlowSt
         return this;
     }
 
+    public SampledTraceBuilder sampledTraceBuilder() {
+        return (SampledTraceBuilder) getProperty(KEY_SAMPLED_TRACE);
+    }
+
+    public Binding sampledTrace(SampledTraceBuilder sampledTraceBuilder) {
```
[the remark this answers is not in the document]
mrproliu: @sonatype-lift ignoreall

wu-sheng (dismissed):
Good for me. @kezhenxu94 Please take a look.

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/SampledTraceBuilder.java
```diff
@@ -0,0 +1,160 @@
...
+        return null;
+    }
+
+    private String validateBiggerZero(String error, String name, long val) {
+        if (error != null) {
+            return error;
+        }
+        if (val < 0) {
+            return name + " must bigger zero.";
+        }
```
kezhenxu94, line 150: This doesn't check `0`, `val == 0` is not `BiggerZero` but no error is reported

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/SampledTraceBuilder.java
```diff
@@ -0,0 +1,160 @@
...
+    @Getter
+    private DetectPoint detectPoint;
+
+    @Setter
+    @Getter
+    private long timestamp;
+
+    public String validate() {
+        String result = null;
+        result = validateNotEmpty(result, "traceId", traceId);
```
kezhenxu94, line 82: Can we just use Guava `Preconditions`?

```java
Preconditions.checkArgument(!Strings.isNullOrEmpty(traceId), "traceId can't be empty");
```

mrproliu: Updated, please help to re-check.

kezhenxu94 (approved):
Thanks

wu-sheng (approved):

PR #9823, opened 2022-10-21
[feat]Add Golang metrics panels in FaaS-Instance dashboard

easonyipj:
- [ ] If this is non-trivial feature, paste the links/URLs to the design doc.
- [ ] Update the documentation to include this new feature.
- [ ] Tests(including UT, IT, E2E) are added to verify the new feature.
- [x] If it's UI related, attach the screenshots below.

- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>.
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/docs/en/changes/changes.md).

### <Feature description>
Add Golang runtime metrics and cpu/memory used rate panels in FaaS-Instance dashboard, the more detail about the Golang metrics can be seen in [(Summer 2022)Feat add golang runtime metrics #9711](https://github.com/apache/skywalking/pull/9711).
image

PR #9863, opened 2022-10-28
feat: support Python runtime metrics

jiang1997:
This PR is the last part of work to complete https://github.com/apache/skywalking-python/pull/238. After the docker image of OAP updates, files as followed need to be removed in Python Agent.
```
skywalking/config/meter-analyzer-config/python.yaml
skywalking/config/application.yml
```

# Looks like
Screenshot_2022-10-28_17-04-08

# References
https://github.com/apache/skywalking/pull/9823
https://github.com/apache/skywalking/pull/9711


oap-server/server-starter/src/main/resources/meter-analyzer-config/python-runtime.yaml
```diff
@@ -0,0 +1,36 @@
...
+  - name: instance_pvm_total_cpu_utilization
+    exp: instance_pvm_total_cpu_utilization
+  - name: instance_pvm_total_mem_utilization
+    exp: instance_pvm_total_mem_utilization
+  - name: instance_pvm_gc_g0
+    exp: instance_pvm_gc_g0
+  - name: instance_pvm_gc_g1
+    exp: instance_pvm_gc_g1
+  - name: instance_pvm_gc_g2
+    exp: instance_pvm_gc_g2
```
wu-sheng, line 32: Are these GC counters? It feels increasing always like so.

oap-server/server-starter/src/main/resources/meter-analyzer-config/python-runtime.yaml
```diff
@@ -0,0 +1,36 @@
...
+  - name: instance_pvm_gc_time
+    exp: instance_pvm_gc_time.increase("PT1M")
```
jiang1997, line 34: Considering the current implementation of `MeterService`, It's more reasonable to report `gc_time` this way.
I will create a new PR for Python Agent to adapt this modification later.

wu-sheng: What do you mean `more reasonable`? `increase` is usually used for a counter which keeps growing.

wu-sheng:
As you have marked this PR ready to review. Please reply or resolve the PR's comments from reviewers.

jiang1997:
> As you have marked this PR ready to review. Please reply or resolve the PR's comments from reviewers.

I'm so sorry. 
I thought I have answered and didn't notice the "pending" tag.
image

jiang1997:
> What do you mean `more reasonable`? `increase` is usually used for a counter which keeps growing.

Because so far gc_time is reported by MeterService every 20s but not reported right after gc.
I'm goint to make gc_time accumulatable to get average time consumption for gc within 1 minute.
Is this okay?

jiang1997:
> Are these GC counters? It feels increasing always like so.

Nope. It's the number of objects for each generation.
Like Java, the objects from older generations are more infrequently to be checked if it can be deleted.

wu-sheng:
> > As you have marked this PR ready to review. Please reply or resolve the PR's comments from reviewers.
> 
> I'm so sorry. I thought I have answered and didn't notice the "pending" tag. image

Pending means you should submit it through `comment` once of all
image

oap-server/server-starter/src/main/resources/meter-analyzer-config/python-runtime.yaml
wu-sheng: > Nope. It's the number of objects for each generation.
Like Java, the objects from older generations are more infrequently to be checked if it can be deleted.

Oh, I don't ask about what `g1` and `g2` mean. I mean is the number a counter or a gauge? What does it mean in every report period? Do we need increase for them too?

oap-server/server-starter/src/main/resources/meter-analyzer-config/python-runtime.yaml
wu-sheng: > Because so far gc_time is reported by MeterService every 20s but not reported right after gc.
I'm goint to make gc_time accumulatable to get average time consumption for gc within 1 minute.
Is this okay?

This is not an issue.

oap-server/server-starter/src/main/resources/meter-analyzer-config/python-runtime.yaml
jiang1997: Gauge.
It's the number of objects that each generation keeps at the moment of reporting. We don't need to increase.

jiang1997:
> > > As you have marked this PR ready to review. Please reply or resolve the PR's comments from reviewers.
> > 
> > 
> > I'm so sorry. I thought I have answered and didn't notice the "pending" tag. image
> 
> Pending means you should submit it through `comment` once of all image

Thanks!

oap-server/server-starter/src/main/resources/meter-analyzer-config/python-runtime.yaml
wu-sheng: Then it reports every 20s, but it is downsampling to minute.
So, we are calculating avg object number in the duration.

jiang1997: > Then it reports every 20s, but it is downsampling to minute. So, we are calculating avg object number in the duration.

Yes. Is this okay?

wu-sheng: @arugal Could you confirm this is from design? 
There are two options, one is using the current way, the other is using the Max as downsampling policy, which shows the worst case.

wu-sheng: Sorry, should ping @Superskyyy @kezhenxu94 Who is your mentor?

jiang1997: > Sorry, should ping @Superskyyy @kezhenxu94 Who is your mentor?

@kezhenxu94

kezhenxu94: Yes, this is reasonable to me.

kezhenxu94 (approved):

wu-sheng (approved):
LGTM, as the mentor has approved.

Superskyyy:
One E2E test case seems to be stuck in a cache retrieval step :0

https://github.com/apache/skywalking/actions/runs/3368331192/jobs/5588961894

wu-sheng:
> One E2E test case seems to be stuck in a cache retrieval step :0 https://github.com/apache/skywalking/actions/runs/3368331192/jobs/5588961894
> 
> Maybe we should tune the segment download timeout down a bit https://github.com/actions/cache/blob/main/workarounds.md#cache-segment-restore-timeout

It makes sense. FYI @kezhenxu94

kezhenxu94:
> Default value of this timeout is 60 minutes and can be customized by specifying an environment variable named SEGMENT_DOWNLOAD_TIMEOUT_MINS with timeout value in minutes.

60 minutes is too long in my opinion. Usually if it takes that long it means something wrong in GitHub Actions and I would expect it to fail as soon as possible.

Superskyyy:
> > Default value of this timeout is 60 minutes and can be customized by specifying an environment variable named SEGMENT_DOWNLOAD_TIMEOUT_MINS with timeout value in minutes.
> 
> 60 minutes is too long in my opinion. Usually if it takes that long it means something wrong in GitHub Actions and I would expect it to fail as soon as possible.

I guess somewhere around 5 minutes should be sufficient for our case? I can make a pr later.

PR #9897, opened 2022-11-02
Reduce the buffer size(queue) of MAL(only) metric streams

wu-sheng:

(Optimization) Reduce the buffer size(queue) of MAL(only) metric streams. Set L1 queue size as 1/20, L2 queue size as 1/2.

In MAL meter streaming, the load of data flow is much less as they are statistics already,
but in OAL sources, they are raw data.Set the buffer(size of queue) as 1/2 to reduce unnecessary resource costs.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsAggregateWorker.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsAggregateWorker.java
@@ -50,13 +50,26 @@ public class MetricsAggregateWorker extends AbstractWorker<Metrics> {
     private CounterMetrics aggregationCounter;
     private long lastSendTime = 0;
 
-    MetricsAggregateWorker(ModuleDefineHolder moduleDefineHolder, AbstractWorker<Metrics> nextWorker,
-                           String modelName, long l1FlushPeriod) {
+    MetricsAggregateWorker(ModuleDefineHolder moduleDefineHolder,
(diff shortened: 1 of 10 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsAggregateWorker.java
```diff
@@ -50,13 +50,26 @@ public class MetricsAggregateWorker extends AbstractWorker<Metrics> {
...
         this.mergeDataCache = new MergableBufferedData();
         String name = "METRICS_L1_AGGREGATION";
-        this.dataCarrier = new DataCarrier<>("MetricsAggregateWorker." + modelName, name, 2, 10000);
+        int queueChannelSize = 2;
+        int queueBufferSize = 10_000;
+        if (MetricStreamKind.MAL == kind) {
+            // In MAL meter streaming, the load of data flow is much less as they are statistics already,
+            // but in OAL sources, they are raw data.
+            // Set the buffer(size of queue) as 1/20 to reduce unnecessary resource costs.
+            queueChannelSize = 1;
```
wu-sheng, line 68: This could help to load consumer thread load too. In the runtime, BulkConsumePool would scan much fewer queues created for MAL. :)
Currently, we have much more metrics from MAL than OAL, I believe.

kezhenxu94 (approved):

wankai123 (approved):

mrproliu (approved):

PR #9904, opened 2022-11-04
Refactor jdbc storage implementations to reuse logics

kezhenxu94:


kezhenxu94:
A brief change list:

- Rename all general `H2XXXX` to `JDBCXXXX` and move them to `jdbc/common` package for general uses of JDBC-compatible storages.
- Unify the configurations of JDBC storages by adding `JDBCStorageConfig`, if specific storage needs new configurations, they can extend `JDBCStorageConfig`, example like `MySQLShardingStorageConfig`.
- Remove mounting config files in `test/e2e-v2/script/shardingsphere-proxy/base-compose.yml`, this is general base docker-compose.yaml and the config files should be mounted in where it's used, like when testing mysql we should mount mysql config files in the mysql case folder, and mount postgresql config files in postgresql folder.
- Remove configuratioin `h2.driver: ${SW_STORAGE_H2_DRIVER:org.h2.jdbcx.JdbcDataSource}`, this is not needed as the driver name can be inferred from the URL `jdbc:h2`, also use the same `properties` configuration style as other JDBC storages used.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ModelInstaller.java
```diff
@@ -58,6 +57,9 @@ public void whenCreating(Model model) throws StorageException {
         }
     }
 
+    public void start() {
```
wu-sheng, line 60: What does `ModelInstaller#start` mean?

kezhenxu94: > What does `ModelInstaller#start` mean?

This is to provide a hook for child-implementations of model installer to do initializations, for now, we do column overrides (e.g. `overrideColumnName("value", "value_");`) in this method.

Previously we do `overrideColumnName("value", "value_");` in the constructor of model installer, this is not an appropriate phase because `overrideColumnName("value", "value_");` requires services that must be already registered to module manager and it might raise `Still in preparing stage` exception if the construction of model installer is too early.

oap-server/server-storage-plugin/storage-shardingsphere-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/shardingsphere/ShardingSphereTableInstaller.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public void createTableIndexes(
+        JDBCHikariCPClient client,
+        Connection connection,
+        String tableName,
+        List<ModelColumn> columns,
+        boolean additionalTable) throws JDBCClientException {
+        delegatee.createTable(client, connection, tableName, columns, additionalTable);
```
wankai123, line 90: Should invoke `createTableIndexes`?
If you want to reuse the super method, should better test the index creation by using the `sharding-proxy`, 
when I try to reuse it before, I got many issues, such as the index name overlength, and can not create index success in this way.

wu-sheng (approved):
Overall are good to me. Pending on @wankai123 to recheck.

oap-server/server-storage-plugin/storage-shardingsphere-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/shardingsphere/ShardingSphereTableInstaller.java
wu-sheng: The index seems can not be covered by the e2e.

kezhenxu94: This is what I changed:

1. We previously create table and then create table indices separately for MySQL storage, and
2. use create table with indices (`create table xxx (yyy int, index key0...)`), for MySQL ShardingSphere storage.

The method (2) applies to both MySQL and ShardingSphere, but method (1) only applies to MySQL storage, to unify the two implementations, I changed (1) to (2) as well so we don't need different table installer for these two implementations.

wankai123 (approved):
LGTM

PR #9909, opened 2022-11-04
Silent NPE fix

BFergerson:
I don't think this change has much affect but noticed that a NPE was being thrown when https://github.com/apache/skywalking/blob/d32a318637bc506364928456aff0f5b02f9bc9c9/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java#L129-L131 is called on the first pass. Since `bulkProcessor` isn't set till https://github.com/apache/skywalking/blob/d32a318637bc506364928456aff0f5b02f9bc9c9/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/PersistenceTimer.java#L135 so calling `endOfFlush` throws silent NPE.


wu-sheng (approved):
Good catch. This nit is hard to catch.

PR #9911, opened 2022-11-06
Optimize MQ Topology analysis. Use entry span's peer from the consumer side as source service.

pg-yang:

As discussion in  https://github.com/apache/skywalking-java/pull/369#discussion_r1011774721 

If there is no producer side instrumentation, we organize `peer` (in consumer EntrySpan ) as source name , virtual mq service name 

image

The data flow in above screenshot  is  `127.0.0.1:5672 (mq  server )-> Your_ApplicationName (as  consumer )`

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -96,7 +96,7 @@
   * Optimize data binary parse methods in *LogQueryDAO
   * Support different indexType
   * Support configuration for TTL and (block|segment) intervals
-* Optimize MQ Topology to organize peer (in consumer EntrySpan) as source service when no producer instrumentation 
+* Optimize MQ Topology analysis. Use entry span's peer from the consumer side as source service when no producer instrumentation(no cross-process reference). 
 
 #### UI
 
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -96,6 +96,7 @@
   * Optimize data binary parse methods in *LogQueryDAO
   * Support different indexType
   * Support configuration for TTL and (block|segment) intervals
+* Optimize MQ Topology to organize peer (in consumer EntrySpan) as source service when no producer instrumentation 
 
 #### UI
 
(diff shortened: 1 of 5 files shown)


docs/en/setup/service-agent/virtual-mq.md
```diff
@@ -12,4 +12,5 @@ The MQ operation span should have
 - **Span's layer == MQ**
 - Tag key = `mq.queue`, value = MQ queue name
 - Tag key = `mq.topic`, value = MQ queue topic , it's optional as some MQ don't have topic concept.
-- Tag key = `transmission.latency`, value = Transmission latency from consumer to producer
\ No newline at end of file
+- Tag key = `transmission.latency`, value = Transmission latency from consumer to producer
+- Set `peer` for EntrySpan (at consumer side) ,because producer might not be instrumented
```
wu-sheng, line 16:
```suggestion
- Set `peer` at both sides(producer and consumer). And the value of peer should represent the MQ server cluster.
```

docs/en/changes/changes.md
```diff
@@ -96,6 +96,7 @@
   * Optimize data binary parse methods in *LogQueryDAO
   * Support different indexType
   * Support configuration for TTL and (block|segment) intervals
+* Optimize MQ Topology to organize peer (in consumer EntrySpan) as source service when no producer instrumentation
```
wu-sheng, line 99:
```suggestion
* Optimize MQ Topology analysis. Use entry span's peer from the consumer side as source service when no producer instrumentation(no cross-process reference). 
```

oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/trace/parser/listener/RPCAnalysisListener.java
```diff
@@ -125,6 +125,21 @@ public void parseEntry(SpanObject span, SegmentObject segmentObject) {
                 setPublicAttrs(sourceBuilder, span);
                 callingInTraffic.add(sourceBuilder);
             }
+        } else if (span.getSpanLayer() == SpanLayer.MQ && StringUtil.isNotBlank(span.getPeer())) {
+            // For MQ , if there is no producer side instrumentation , and consumer has peer , we set the peer as source service name
```
wu-sheng, line 129:
```suggestion
            // For MQ, if there is no producer-side instrumentation, we set the existing peer as the source service name.
```

wu-sheng (commented):
LGTM, some nits for docs.

wu-sheng (approved):

PR #9912, opened 2022-11-06
fix typo that invalids python-runtime.yaml

jiang1997:
### Fix: remove extra space that invalid python-runtime.yaml.

Tested with ghcr.io/apache/skywalking/oap:d32a318637bc506364928456aff0f5b02f9bc9c9.


kezhenxu94 (approved):

wu-sheng:
We seem should trim to avoid rhis kind of issue? @yswdqz

yswdqz:
> We seem should trim to avoid rhis kind of issue? @yswdqz

Got it.

Issue #9849, opened 2022-10-26
[Feature] Allow per index shard and replicas number overrides.


toffentoffen:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Currently, when Elasticsearch storage option is used you can specify the number of shard and replicas **all** the indices should use. Each index which backs a concrete metric, (or all metrics), record, ..., has different growing rates. Therefore, we would like to have an option to override shard/replica setting per index. Specially, this feature would be really helpful as since 9.2.0 
a new storage option was introduced https://skywalking.apache.org/docs/main/v9.2.0/en/faq/new-elasticsearch-storage-option-explanation-in-9.2.0/ that magnifies the different growing rate between indices.

### Use case

We would like to have an option to override shard/replica setting per index as each index might have different growing rates.
Specially, this feature would be really helpful as since 9.2.0 
a new storage option was introduced https://skywalking.apache.org/docs/main/v9.2.0/en/faq/new-elasticsearch-storage-option-explanation-in-9.2.0/ that magnifies the different growing rate between indices.

- We need a way to set the default shard numbers and default replicas number using `SW_STORAGE_ES_INDEX_SHARDS_NUMBER` and `SW_STORAGE_ES_INDEX_REPLICAS_NUMBER`.
- We should be able to override the default number of shards or replicas targeting concrete Indices irrespectively of `SW_STORAGE_ES_LOGIC_SHARDING`. Which means if `SW_STORAGE_ES_LOGIC_SHARDING` is false only 4 indices can be overriden. If `SW_STORAGE_ES_LOGIC_SHARDING` is false, ~40 indices can be overriden.
- Instead of having 1 property name per index, we could have 1 new property that contains a JSON to override per index number of shard of replicas. The new property could be `SW_STORAGE_ES_INDICES_SETTINGS` . But any other name is fine as well.
- If overriding property is define, it will be used. If an index does not have an override, the default values will be used.
- About the JSON format I don’t have a strong opinion, whatever is easier I am fine with. However, It could be something like, if `SW_STORAGE_ES_LOGIC_SHARDING` is set to false for instance:
```json
{
  "metrics-all": {
    "shards_number": 4,
    "replicas_number": 2
  },
  "zipkin_span": { // This will override discard the `SW_STORAGE_ES_SUPER_DATASET_INDEX_SHARDS_FACTOR`
    "shards_number": 2,
[...]

PR #9914, opened 2022-11-07
Elasticsearch storage: support specify the settings `(number_of_shards/number_of_replicas)` for each index individually.

wankai123:
- Elasticsearch storage: Provide system environment variable(`SW_STORAGE_ES_SPECIFIC_INDEX_SETTINGS`) and support specify the settings `(number_of_shards/number_of_replicas)` for each index individually.
- Elasticsearch storage: Support update index settings `(number_of_shards/number_of_replicas)` for the index template after re-configured.

- [X] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes (https://github.com/apache/skywalking/issues/9849).


kezhenxu94 (dismissed):
If CI passed

docs/en/setup/backend/backend-storage.md
```diff
@@ -163,6 +166,37 @@ storage:
     advanced: ${SW_STORAGE_ES_ADVANCED:"{\"index.translog.durability\":\"request\",\"index.translog.sync_interval\":\"5s\"}"}
 ```
 
+### Specify Settings For Each Elasticsearch Index Individually
```
wu-sheng, line 169: So, we support other configurations too? Besides shards and replications. Because the original #9849 mentioned these two configurations only.

docs/en/changes/changes.md
```diff
@@ -97,6 +97,8 @@
   * Support different indexType
   * Support configuration for TTL and (block|segment) intervals
 * Optimize MQ Topology analysis. Use entry span's peer from the consumer side as source service when no producer instrumentation(no cross-process reference). 
+* Elasticsearch storage: Provide system environment variable(`SW_STORAGE_ES_SPECIFIC_INDEX_SETTINGS`) and support specify the settings `(number_of_shards/number_of_replicas)` for each index individually.
+* Elasticsearch storage: Support update index settings `(number_of_shards/number_of_replicas)` for the index template after re-configured.
```
wu-sheng, line 101:
```suggestion
* Elasticsearch storage: Support update index settings `(number_of_shards/number_of_replicas)` for the index template after rebooting.
```

I think we need to make this and doc consistent. The doc doesn't mention rebooting and override existing index settings.

docs/en/setup/backend/backend-storage.md
```diff
@@ -151,7 +154,21 @@ Once it is changed manually or through a 3rd party tool, such as [Vault](https:/
 the storage provider will use the new username, password, and JKS password to establish the connection and close the old one. If the information exists in the file,
 the `user/password` will be overridden.
 
-### Advanced Configurations For Elasticsearch Index
+
+### Each Elasticsearch Index Settings
+You can set and update the index settings `(number_of_shards/number_of_replicas)` by the following configuration items for the index template after rebooting:
```
wu-sheng, line 159:
```suggestion
### Index Settings

The following settings control the number of shards and replicas for new and existing index templates. The update only got applied after OAP reboots.
```

docs/en/setup/backend/backend-storage.md
```diff
@@ -163,6 +180,39 @@ storage:
...
+#### Specify Settings For Each Elasticsearch Index Individually
+You can specify the settings for one or more indexes individually by using `SW_STORAGE_ES_SPECIFIC_INDEX_SETTINGS`.
+
+**NOTE:**
+Supported settings:
+- number_of_shards
+- number_of_replicas
+
+**NOTE:** If configured, the settings in this configuration have the highest priority and will override the existing 
+generic settings which configured by the others configuration items.
```
wu-sheng, line 192:
```suggestion
**NOTE:** These settings have the highest priority and will override the existing 
generic settings mentioned in [index settings doc](#index-settings).
```

Please check whether the new link is correct.

docs/en/setup/backend/backend-storage.md
```diff
@@ -163,6 +180,39 @@ storage:
...
+
+The settings in `JSON` format and the index name in the config should exclude the `${SW_NAMESPACE}` which is `sw` by default, e.g.
```
wu-sheng, line 194:
```suggestion
The settings are in `JSON` format. The index name here is logic entity name, which should exclude the `${SW_NAMESPACE}` which is `sw` by default, e.g.
```

docs/en/setup/backend/backend-storage.md
```diff
@@ -163,6 +180,39 @@ storage:
...
+    "number_of_replicas":"2"
+  },
+  "segment":{
+    "number_of_shards":"6",
+    "number_of_replicas":"1"
+  }
+}
+```
+
+And put it into the following configure item:
```
wu-sheng, line 208:
```suggestion
This configuration in the YAML file is like this,
```

docs/en/setup/backend/backend-storage.md
```diff
@@ -151,7 +154,21 @@ Once it is changed manually or through a 3rd party tool, such as [Vault](https:/
...
+You can set and update the index settings `(number_of_shards/number_of_replicas)` by the following configuration items for the index template after rebooting:
+```yaml
+storage:
+  elasticsearch:
+    # ......
+    indexShardsNumber: ${SW_STORAGE_ES_INDEX_SHARDS_NUMBER:1} 
+    indexReplicasNumber: ${SW_STORAGE_ES_INDEX_REPLICAS_NUMBER:1} 
+    specificIndexSettings: ${SW_STORAGE_ES_SPECIFIC_INDEX_SETTINGS:""}
+    superDatasetIndexShardsFactor: ${SW_STORAGE_ES_SUPER_DATASET_INDEX_SHARDS_FACTOR:5} 
+    superDatasetIndexReplicasNumber: ${SW_STORAGE_ES_SUPER_DATASET_INDEX_REPLICAS_NUMBER:0}
```
wu-sheng, line 168: I think as we use merged index policy by default, we could document which indicates affected by which configurations clearly.

wu-sheng (approved):

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexStructures.java
```diff
@@ -146,4 +172,18 @@ private Map<String, Object> diffFields(Fields fields) {
...
+     * The index settings structure which only include needs to compare and update fields
+     */
+    @EqualsAndHashCode
+    public static class UpdatableIndexSettings {
+        private final int replicas;
+        private final int shards;
+
+        public UpdatableIndexSettings(Map<String, Object> indexSettings) {
+            this.replicas = Integer.parseInt((String) indexSettings.getOrDefault("number_of_replicas", "0"));
+            this.shards = Integer.parseInt((String) indexSettings.getOrDefault("number_of_shards", "0"));
```
toffentoffen, line 186: I am fully aware how it works globally, but I have some doubts.

1. Is it mandatory to set always both properties for a given index in the json to specify the concrete number of replicas and shards?
2. If only replicas is provided for instance, what will be the number of shards 0? Or the value specified by `SW_STORAGE_ES_INDEX_SHARDS_NUMBER` if provided or the default value of `SW_STORAGE_ES_INDEX_SHARDS_NUMBER` if absent?

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -183,13 +214,14 @@ private void createTimeSeriesTable(Model model) throws StorageException {
...
-        setting.put("index.number_of_shards", model.isSuperDataset()
-            ? config.getIndexShardsNumber() * config.getSuperDatasetIndexShardsFactor()
-            : config.getIndexShardsNumber());
+        Map<String, Object> indexSettings = new HashMap<>();
+        setting.put("index", indexSettings);
+        indexSettings.put("number_of_replicas", model.isSuperDataset()
+            ? Integer.toString(config.getSuperDatasetIndexReplicasNumber())
+            : Integer.toString(config.getIndexReplicasNumber()));
+        indexSettings.put("number_of_shards", model.isSuperDataset()
+            ? Integer.toString(config.getIndexShardsNumber() * config.getSuperDatasetIndexShardsFactor())
```
toffentoffen, line 223: I would suggest that we should not multiply the number of shards by the shards factor if a concrete number of shards is configure for a given index. Otherwise, the user might be confused. If the user says I want X number of shards for zipkin index, X number of shards should be used. Otherwise, when the users checks the number of shards it will see X*Shards_Factor.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -111,28 +123,36 @@ private void createNormalTable(Model model) throws StorageException {
...
+                    throw new StorageException("update " + tableName + " index mapping failure");
                 }
             }
+            //needs to update settings
+            if (!structures.compareIndexSetting(tableName, settings)) {
+                log.warn(
+                    "index {} settings configuration has been updated to {}, please remove it before OAP starts",
+                    tableName, settings
+                );
+            }
```
toffentoffen, line 155: Under which scenario, which type of setting, causes that the index needs to be removed? I guess that if specific number of shards or replicas are specified for a concrete index, this warning will be displayed isn't it?
I not sure if we should display it when only shards or replicas have changed:

1. User could be confused, as the shards and replica changes only apply for new indices after the changes are applied to the index template as per https://github.com/apache/skywalking/pull/9914/files#diff-7c42c99e33896647d67385263076ec5178df8e57d072e760895ed7ca9f5d1c20R199
2. Maybe we should display an info message instead if only shards and replica number have changed, as those changes can never be applied because ES doesn't allow to change the number of shards and replicas for and existing index. Moreover, I think those changes are never applied to an existing index, that's the reason why I would suggest to not show a warning message.

toffentoffen (commented):
I have mainly some doubts and suggestions. 
OverAll looks good to me. Well done, thanks for implementing it!!!! Many user can benefit from this new feature.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
wu-sheng: Factor should only be used in default settings, not a part of override mechanism.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexStructures.java
wankai123: Thanks, I'll check.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
wankai123: Hi, this warning is only for the `Normal index` which would not generate new indexes with time series by index template. So the new settings will not be applied until the index is recreated. 

For the `Time series index` I log the info:
https://github.com/apache/skywalking/blob/48a0985345bc65a63b17d4539f80de334383dedb/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java#L199

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/IndexStructures.java
wankai123: > 1. Is it mandatory to set always both properties for a given index in the json to specify the concrete number of replicas and shards?

No, we can set one or both, the absent one will use the generic settings.
see: https://skywalking.apache.org/docs/main/next/en/setup/backend/backend-storage/#index-settings

> 2. If only replicas is provided for instance, what will be the number of shards 0? 

No, `UpdatableIndexSettings` structure only used for comparing old and new settings, so the the number of shards will be the the generic settings which I mentioned above.
If 
```java
Integer.parseInt((String) indexSettings.getOrDefault("number_of_replicas", "0"));
```
 cause some confusion, I'll submit a PR, and change it to 
```java
String.valueOf(indexSettings.getOrDefault("number_of_replicas", Const.EMPTY_STRING));
```

>Or the value specified by `SW_STORAGE_ES_INDEX_SHARDS_NUMBER` if provided or the default value of `SW_STORAGE_ES_INDEX_SHARDS_NUMBER` if absent?

yes, they have a default value
see: https://github.com/apache/skywalking/blob/48a0985345bc65a63b17d4539f80de334383dedb/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java#L60

PR #9915, opened 2022-11-07
add component id for eventMesh java plugin.

heihei180:
add component id for pr: https://github.com/apache/skywalking-java/pull/376

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -101,6 +101,9 @@
 * Optimize MQ Topology analysis. Use entry span's peer from the consumer side as source service when no producer instrumentation(no cross-process reference).
 * Refactor JDBC storage implementations to reuse logics.
 * Fix `ClassCastException` in `LoggingConfigWatcher`.
+* Support span attached event concept in Zipkin and SkyWalking trace query.
+* Support span attached events on Zipkin lens UI.
+* Force UTF-8 encoding in `JsonLogHandler` of `kafka-fetcher-plugin`.
 
 #### UI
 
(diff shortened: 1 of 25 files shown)


oap-server/server-starter/src/main/resources/component-libraries.yml
```diff
@@ -435,6 +435,10 @@ Impala:
   id: 134
   languages: Java
 
+EventMesh-producer:
```
wu-sheng, line 438: Could you explain why only producer? Where is consumer and server?
And why do you need producer rather than client?

wu-sheng: As you are adding a tracing plugin, you have to do both sides, otherwise, the tracing context breaks, and ppl would feel confused.

oap-server/server-starter/src/main/resources/component-libraries.yml
```diff
@@ -435,6 +435,18 @@ Impala:
...
+  id: 135
+  languages: Java
+
+eventmesh-producer:
+  id: 136
+  languages: Java
+
+eventmesh-consumer:
+  id: 137
+  languages: Java
```
wu-sheng, line 448: Please follow the formats. No empty lines.

wu-sheng:
The style is still wrong.

oap-server/server-starter/src/main/resources/component-libraries.yml
```diff
@@ -434,6 +434,15 @@ Impala-jdbc-driver:
...
   languages: Java
+EventMesh:
+  id: 135
+  languages: Java
+eventmesh-producer:
+  id: 136
+  languages: Java
+eventmesh-consumer:
+  id: 137
+  languages: Java
```
wu-sheng, line 445: Would you keep the upper case and lower case consistent? These names seem very strange.

wu-sheng:
Any update?

oap-server/server-starter/src/main/resources/component-libraries.yml
wu-sheng: OK, if that is not a mistake, I am fine.

wu-sheng (approved):

wu-sheng:
Oops, you missed updating changes.md in the /docs/en/changes. Please fix it.

Issue #9803, opened 2022-10-18
[Feature] Support SpanAttachedEvent Concept


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

This is a new extension and the first one after [Trace Data Protocol v3](https://skywalking.apache.org/docs/main/next/en/protocols/trace-data-protocol-v3/). So I marked this as v3.1, the main document would be updated once this is really implemented.

## Case Scenario
When an RPC is being traced by the in-process language agent, a span would be reported by the client-side agent.
And the rover would be aware of this RPC due to the existing tracing header.
Then, the rover agent collects extra information from the OS level to provide assistance information to diagnose network performance.

## Who is going to implement this
The Rover(ebpf) agent would implement this. This relies on the rover agent having the capability to decode HTTP(at least) header and body on the fly.

## Query Protocol 
```
# Since 9.3.0
# Record is a general and abstract type for collected raw data.
# In the observability, traces and logs have specific and well-defined meanings, and the general records represent other
# collected records. Such as sampled slow SQL statement, HTTP request raw data(request/response header/body)
extend type Query {
    # Query collected records with given metric name and parent entity conditions, and return in the requested order.
    readRecords(condition: RecordCondition!, duration: Duration!): [Record!]!
}
```

A new query protocol is being added, to replace the `readSampledRecords`. This would be a compatible upgrade to support more use cases for sampled records(such as in instance level) and incoming HTTP header/body collecting(through ebpf agent).

### Use case

In ebpf case, we would be able to sample HTTP request/response according to latency threshold, even 100% sampling. And link those records with trace ID/span ID decoded from the sw8(skywalking)/b3(zipkin) header. Then we could have this diagnose workflow, `slow HTTP request` -> `TRACE view` -> `RPC span with attached event`. The event includes the HTTP request/response in step <1>, as well as statistics of network events(ref https://github.com/apache/skywalking-data-collect-protocol/pull/77)

### Related issues

Another doc should be added later after this implementation to introduce what it is the SpanAttachedEvent Concept - Owner @wu-sheng 

We have introduced this concept in the protocol definition, we also should update the docs of protocol and chapters in the concepts part to explain this new concept.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

[...]

PR #9916, opened 2022-11-07
Support span attached event concept

mrproliu:
Support receiving, storing, and querying the span attached event in tracing.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Part of https://github.com/apache/skywalking/issues/9803.


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/spanattach/SpanAttachedEventRecord.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.skywalking.oap.server.core.storage.type.Convert2Storage;
+import org.apache.skywalking.oap.server.core.storage.type.StorageBuilder;
+
+import static org.apache.skywalking.oap.server.core.source.DefaultScopeDefine.SPAN_ATTACHED_EVENT;
+
+@Setter
+@Getter
+@ScopeDeclaration(id = SPAN_ATTACHED_EVENT, name = "SpanAttachedEvent")
+@Stream(name = SpanAttachedEventRecord.INDEX_NAME, scopeId = SPAN_ATTACHED_EVENT, builder = SpanAttachedEventRecord.Builder.class, processor = RecordStreamProcessor.class)
+public class SpanAttachedEventRecord extends Record {
```
wu-sheng, line 40: Are you going to use 100% sampling? This means sampling all events matching the trace header. If so, consider `@SuperDataset`.

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/handler/v8/grpc/SpanAttachedEventReportServiceHandler.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    public StreamObserver<SpanAttachedEvent> collect(StreamObserver<Commands> responseObserver) {
+        return new StreamObserver<SpanAttachedEvent>() {
+            @Override
+            public void onNext(SpanAttachedEvent event) {
+                if (log.isDebugEnabled()) {
+                    log.debug("receive span attached event is streaming");
+                }
+
+                final SpanAttachedEventRecord record = new SpanAttachedEventRecord();
```
wu-sheng, line 49: Should we consider HTTP endpoint?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/spanattach/SpanAttachedEventRecord.java
```diff
@@ -0,0 +1,112 @@
...
+    @Column(columnName = START_TIME_NANOS)
+    private int startTimeNanos;
+    @Column(columnName = EVENT)
+    private String event;
+    @Column(columnName = END_TIME_SECOND)
+    private long endTimeSecond;
+    @Column(columnName = END_TIME_NANOS)
+    private int endTimeNanos;
+    @Column(columnName = TRACE_REF_TYPE)
+    private int traceRefType;
```
wu-sheng, line 65: Should we add a `SpanReferenceType` type field(not a column` in the Record?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/spanattach/SpanAttachedEventTraceType.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * The {@link SpanAttachedEventRecord} tracing context reference type.
+ */
+public enum SpanAttachedEventTraceType {
+
+    SKYWALKING(0),
+
+    ZIPKIN(1),
+    ;
```
wu-sheng, line 32:
```suggestion
    ZIPKIN(1);
```

Should we use this format?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/TraceQueryService.java
```diff
@@ -244,4 +268,61 @@ private void findChildren(List<Span> spans, Span parentSpan, List<Span> children
...
+            return;
+        }
+
+        for (SpanAttachedEventRecord record : events) {
+            if (!StringUtils.isNumeric(record.getTraceSpanId())) {
+                continue;
+            }
+
+            // find matches span
+            final int eventSpanId = Integer.parseInt(record.getTraceSpanId());
```
wu-sheng, line 283: I think we should check SpanReferenceType first, then do casting. Otherwise, we may face NPE or cast error, if trace ID actually exists on both sides. Such as Lua agent uses UUID as well.

wu-sheng: According to this, https://github.com/apache/skywalking-query-protocol/pull/101, we should update the Zipkin query as well. And verify whether we could put this in tags or some places for lens UI?

mrproliu: Updated. I have filtered the skywalking events in the query and checked the parent span ID is numeric or not.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/spanattach/SpanAttachedEventRecord.java
mrproliu: Updated it as an enum.

oap-server/server-receiver-plugin/skywalking-trace-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/trace/provider/handler/v8/grpc/SpanAttachedEventReportServiceHandler.java
mrproliu: Because there is no need now, so ignore it.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/spanattach/SpanAttachedEventRecord.java
mrproliu: Updated it as super data set.

docs/en/changes/changes.md
```diff
@@ -101,6 +101,7 @@
 * Optimize MQ Topology analysis. Use entry span's peer from the consumer side as source service when no producer instrumentation(no cross-process reference).
 * Refactor JDBC storage implementations to reuse logics.
 * Fix `ClassCastException` in `LoggingConfigWatcher`.
+* Support span attached event concept.
```
wu-sheng, line 104:
```suggestion
* Support span attached event concept in Zipkin and SkyWalking trace query.
* Support span attached events on Zipkin lens UI.
```

This doc should be updated as well, https://skywalking.apache.org/docs/main/next/en/setup/backend/zipkin-trace/

Also, we should consider adding a menu here to introduce ebpf besides tracing/logging/metrics.
image

wu-sheng (dismissed):

docs/en/changes/changes.md
wu-sheng: Let's update the docs for profiling in the next PR.

wu-sheng (approved):

PR #9928, opened 2022-11-09
Polish Elasticsearch IndexStructures

wankai123:
relate to: https://github.com/apache/skywalking/pull/9914#pullrequestreview-1171156294

@toffentoffen


toffentoffen:
Thanks @wankai123!
Does this PR fixes this comment https://github.com/apache/skywalking/pull/9914#discussion_r1015993463?

wu-sheng:
@toffentoffen I think that comment doesn't need to fix. Isn't it? This override directly uses the value, rather than the factor parameter.

toffentoffen:
@wu-sheng Probably I am misinterpreting the code. And it is already working correctly.
I thought that https://github.com/apache/skywalking/pull/9914/files#diff-7c42c99e33896647d67385263076ec5178df8e57d072e760895ed7ca9f5d1c20R219-R223
is always applied, and that `config.getIndexShardsNumber()` already contains either the default value or the override.

wankai123:
> @wu-sheng Probably I am misinterpreting the code. And it is already working correctly. I thought that https://github.com/apache/skywalking/pull/9914/files#diff-7c42c99e33896647d67385263076ec5178df8e57d072e760895ed7ca9f5d1c20R219-R223 is always applied, and that `config.getIndexShardsNumber()` already contains either the default value or the override.

Hi @toffentoffen, these codes just `rebuild the Map structure for the override`, they don't change the original logic. You can check the code history.

PR #9936, opened 2022-11-09
Optimize document statements

zhangrm:

### DIFF
--- a/docs/en/guides/How-to-release.md
+++ b/docs/en/guides/How-to-release.md
@@ -37,7 +37,7 @@ If you are a committer, use your Apache ID and password to log in this svn, and 
 Apache maven staging repository checklist.
 
 ## Test your settings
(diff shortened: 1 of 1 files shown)


docs/en/guides/How-to-release.md
```diff
@@ -37,7 +37,7 @@ If you are a committer, use your Apache ID and password to log in this svn, and
 Apache maven staging repository checklist.
 
 ## Test your settings
-This step is only for testing purpose. If your env is correctly set, you don't need to check every time.
+This step is only for testing purpose. If your env is correctly set, you needn't check it every time.
```
wu-sheng, line 40: According to http://www.dualtexts.com/grammar-bank/388-needn-t-vs-don-t-need-to#:~:text=Needn't%2Bverb%20and%20don,when%20talking%20about%20the%20past.

`don't need to` matches the point better.

PR #9948, opened 2022-11-11
modify doc changes.md

heihei180:
modify doc changes.md， add `eventmesh-1.x` plugin , link: https://github.com/apache/skywalking/pull/9915


wu-sheng:
Please make your commit log clear. This is so long. And the change log should be only about, your added component ID(s). There is nothing about the Java plugin

wu-sheng:
What's done? Did you forget or fail to push the update?

docs/en/changes/changes.md
```diff
@@ -4,6 +4,9 @@
 
 * Bump up the embedded `swctl` version in OAP Docker image.
 
+#### Java Agent
+* Add `eventmesh-1.x` plugin
+
```
wu-sheng, line 9:
```suggestion
```

This isn't main repo change.

wu-sheng (approved):
I made the change log for you. Please be more careful to process documents.
The codes are not the only thing that matter.

PR #9954, opened 2022-11-12
Fix Elasticsearch storage installer: ignore check index settings for updating  in no-init running mode

### LINKED ISSUE #9953: [Bug] OAP server is blocked from booting in no-init mode(9.3 dev only) in certain cases.
### Search before asking

### What happened

OAP server is blocked from booting in no-init mode. This only happens in 9.3 dev after https://github.com/apache/skywalking/pull/9914 got merged.

In this PR, OAP began to check index settings(shard and replica) in both init and no-init mode. But if init mode OAP and no-init mode OAP don't share the same configuration for those settings, the no-init OAPs are failure to boot.

We should consider polishing this mechanism, and no-init mode OAP should have no responsibility to check the index setting(shard and replica). Because they don't control this part. Only things they care are whether the index existing as well as relative columns which should be in the same type.

### How to reproduce

When init mode OAP and no-init mode OAP don't share the same configuration for those settings

### Anything else

[NOTICE] This was noticed by @kezhenxu94 when he redeployed the latest 9.3-dev on demo env. This issue should be able to be fixed before the 9.3.0 release, so, no end user should be able to be concerned by users.

### Code of Conduct

wankai123:
- [X] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #9953 

### DIFF
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
@@ -30,6 +30,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.skywalking.library.elasticsearch.response.Index;
 import org.apache.skywalking.library.elasticsearch.response.IndexTemplate;
 import org.apache.skywalking.library.elasticsearch.response.Mappings;
+import org.apache.skywalking.oap.server.core.RunningMode;
 import org.apache.skywalking.oap.server.core.storage.StorageException;
 import org.apache.skywalking.oap.server.core.storage.model.Model;
 import org.apache.skywalking.oap.server.core.storage.model.ModelColumn;
@@ -84,8 +85,15 @@ public class StorageEsInstaller extends ModelInstaller {
                 Optional<Index> index = esClient.getIndex(tableName);
                 Mappings historyMapping = index.map(Index::getMappings).orElseGet(Mappings::new);
                 structures.putStructure(tableName, historyMapping, index.map(Index::getSettings).orElseGet(HashMap::new));
-                exist = structures.containsMapping(tableName, createMapping(model))
-                    && structures.compareIndexSetting(tableName, createSetting(model));
+                boolean containsMapping = structures.containsMapping(tableName, createMapping(model));
+                boolean compareIndexSetting = structures.compareIndexSetting(tableName, createSetting(model));
+                // "no-init mode" no needs to check index settings for updating,
+                // to avoid conflicts between "init mode and no-init mode" index settings configurations
+                if (RunningMode.isNoInitMode()) {
+                    exist = containsMapping;
+                } else {
+                    exist = containsMapping && compareIndexSetting;
+                }
             }
             return exist;
         }
@@ -104,8 +112,15 @@ public class StorageEsInstaller extends ModelInstaller {
             structures.putStructure(
                 tableName, template.get().getMappings(), template.get().getSettings()
             );
-            exist = structures.containsMapping(tableName, createMapping(model))
-                && structures.compareIndexSetting(tableName, createSetting(model));
+            boolean containsMapping = structures.containsMapping(tableName, createMapping(model));
+            boolean compareIndexSetting = structures.compareIndexSetting(tableName, createSetting(model));
+            // "no-init mode" no needs to check index settings for updating,
+            // to avoid conflicts between "init mode and no-init mode" index settings configurations
+            if (RunningMode.isNoInitMode()) {
+                exist = containsMapping;
+            } else {
+                exist = containsMapping && compareIndexSetting;
+            }
         }
         return exist;
     }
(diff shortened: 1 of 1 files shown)


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -84,8 +85,15 @@ public boolean isExists(Model model) throws StorageException {
...
-                    && structures.compareIndexSetting(tableName, createSetting(model));
+                boolean containsMapping = structures.containsMapping(tableName, createMapping(model));
+                boolean compareIndexSetting = structures.compareIndexSetting(tableName, createSetting(model));
+                // "no-init mode" no needs to check index settings for updating,
+                // to avoid conflicts between "init mode and no-init mode" index settings configurations
+                if (RunningMode.isNoInitMode()) {
+                    exist = containsMapping;
+                } else {
+                    exist = containsMapping && compareIndexSetting;
+                }
```
wu-sheng, line 96:
```suggestion
                boolean containsMapping = structures.containsMapping(tableName, createMapping(model));
                // Do not check index settings in the "no-init mode",
                // because the no-init mode OAP server doesn't take responsibility for index settings.
                if (RunningMode.isNoInitMode()) {
                    exist = containsMapping;
                } else {
                    exist = containsMapping && structures.compareIndexSetting(tableName, createSetting(model));
                }
```

I think this should be better in perf.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -104,8 +112,15 @@ public boolean isExists(Model model) throws StorageException {
...
-                && structures.compareIndexSetting(tableName, createSetting(model));
+            boolean containsMapping = structures.containsMapping(tableName, createMapping(model));
+            boolean compareIndexSetting = structures.compareIndexSetting(tableName, createSetting(model));
+            // "no-init mode" no needs to check index settings for updating,
+            // to avoid conflicts between "init mode and no-init mode" index settings configurations
+            if (RunningMode.isNoInitMode()) {
+                exist = containsMapping;
+            } else {
+                exist = containsMapping && compareIndexSetting;
+            }
```
wu-sheng, line 123: A similar suggestion as above.

wu-sheng:
@wankai123 I didn't change it locally, so don't know whether it is 100% correct.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -104,8 +111,14 @@ public boolean isExists(Model model) throws StorageException {
             structures.putStructure(
                 tableName, template.get().getMappings(), template.get().getSettings()
             );
-            exist = structures.containsMapping(tableName, createMapping(model))
-                && structures.compareIndexSetting(tableName, createSetting(model));
+            boolean containsMapping = structures.containsMapping(tableName, createMapping(model));
+            // "no-init mode" no needs to check index settings for updating,
+            // to avoid conflicts between "init mode and no-init mode" index settings configurations
```
wu-sheng, line 116:
```suggestion
            // Do not check index settings in the "no-init mode",
            // because the no-init mode OAP server doesn't take responsibility for index settings.
```

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -84,8 +85,14 @@ public boolean isExists(Model model) throws StorageException {
                 Optional<Index> index = esClient.getIndex(tableName);
                 Mappings historyMapping = index.map(Index::getMappings).orElseGet(Mappings::new);
                 structures.putStructure(tableName, historyMapping, index.map(Index::getSettings).orElseGet(HashMap::new));
-                exist = structures.containsMapping(tableName, createMapping(model))
-                    && structures.compareIndexSetting(tableName, createSetting(model));
+                boolean containsMapping = structures.containsMapping(tableName, createMapping(model));
+                // Do not check index settings in the "no-init mode",
+                // because the no-init mode OAP server doesn't take responsibility for index settings.
```
kezhenxu94, line 90: Does no init mode take responsibility of index mappings?

wankai123: no-init mode needs check and wait init mappings

wu-sheng: Kind of. As OAP should wait for the index mappings contain the required.

kezhenxu94:
The changes LGTM

wu-sheng (approved):

PR #9955, opened 2022-11-13
Nodejs plugin AWSDynamoDB, AWSSNS,  AWSSQS component IDs

tom-pytel:
For: https://github.com/apache/skywalking-nodejs/pull/100

Also includes some previous component server mappings which I guess I forgot to push.

### DIFF
--- a/oap-server/server-starter/src/main/resources/component-libraries.yml
+++ b/oap-server/server-starter/src/main/resources/component-libraries.yml
@@ -443,6 +443,15 @@ eventmesh-producer:
 eventmesh-consumer:
   id: 137
   languages: Java
+AWSDynamoDB:
+  id: 138
+  languages: Java,C#,Node.js,Python
+AWSSNS:
+  id: 139
+  languages: Java,C#,Node.js,Python
+AWSSQS:
+  id: 140
+  languages: Java,C#,Node.js,Python
 
 # .NET/.NET Core components
 # [3000, 4000) for C#/.NET only


wu-sheng (approved):
We don't need extra logos for those services, right?

oap-server/server-starter/src/main/resources/component-libraries.yml
```diff
@@ -425,6 +425,15 @@ tls:
...
   languages: Java
+AWSDynamoDB:
+  id: 132
+  languages: Java,C#,Node.js,Python
+AWSSNS:
+  id: 133
+  languages: Java,C#,Node.js,Python
+AWSSQS:
+  id: 134
+  languages: Java,C#,Node.js,Python
```
wu-sheng, line 436: 132-134 have been registered. Please register to new IDs.

wu-sheng (changes requested):

tom-pytel:
> We don't need extra logos for those services, right?

They would be logos from Amazon itself, I am guessing copyrighted?

wu-sheng (approved):

wu-sheng:
One more, changes should be uodated.

tom-pytel:
> One more, changes should be uodated.

Yes, will update in a bit.

oap-server/server-starter/src/main/resources/component-libraries.yml
tom-pytel: Fixed.

wu-sheng:
Merged first, in case of conflicts.

PR #9964, opened 2022-11-16
Add `LongText` to support longer logs persistent as a text type 

### LINKED ISSUE #9961: log data is too large, the trace log can't recorded.
### Search before asking

### What happened

when the log data is too large,  the trace log can't  recorded.   
because the elasticsearch keywork type max lenth is  UTF8 encoding is longer than the max length 32766
sugguest  keep a record and ignore any unnecessary information

After setting ignore_above, data beyond the specified length is not indexed and cannot be returned through term exact matching retrieval
https://www.elastic.co/guide/en/elasticsearch/reference/5.4/ignore-above.html#ignore-above%3E

trace log insert elastic error

"caused_by": {
            "type": "illegal_argument_exception",
            "reason": "Document contains at least one immense term in field=\"content\" (whose UTF8 encoding is longer than the max length 32766), all of which were skipped. Please correct the analyzer to not produce such terms. The prefix of the first immense term is: '[49, 53, 55, 56, 52, 54, 51, 51, 51, 118, 99, 111, 109, 115, 97, 100, 112, 97, 110, 100, 97, 117, 115, 115, 97, 102, 101, 114, 118, 112]...'"
        }

### How to reproduce

GRPC-LOG 

record  more than  32766 [whose UTF8 encoding is longer than the max length 32766]  length word

### Anything else

_No response_

### Code of Conduct

wu-sheng:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #9961.

Add `LongText` to support longer logs persistent as a text type in ElasticSearch, instead of a keyword, to avoid length limitation.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/record/LongText.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/record/LongText.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.oap.server.core.analysis.record;
+
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.ToString;
+import org.apache.skywalking.oap.server.core.storage.type.StorageDataComplexObject;
+
+/**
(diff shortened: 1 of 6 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/record/LongText.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * LongText represents a string field, but the length is not predictable and could be longer than 1000.
+ * This is a wrapper of Java String only, which provides an indicator of long text field.
+ */
+@Getter
+@Setter
+@ToString
+@EqualsAndHashCode
+public class LongText implements StorageDataComplexObject<LongText> {
```
wu-sheng, line 35: I am adding this to adopt this

https://github.com/apache/skywalking/blob/07bfde6e254e364b6ab31324c9e342a5f7e64653/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/ColumnTypeEsMapping.java#L39-L42

Move from `keyword` to `text`.

@kezhenxu94 Could you confirm this way is good to keep the log text is still the raw text, rather than JSON wrapped? I am not sure about `@JsonSerialize(using = StorageDataComplexObject.Serializer.class)` in its parent class.

kezhenxu94: Yes. The serializer uses `toStorageData` as the sterilized string and you just return the raw text in that method so it’s not wrapped to JSON.

kezhenxu94 (approved):
LGTM

kezhenxu94 (approved):

kezhenxu94 (approved):

PR #9988, opened 2022-11-18
Modify the format of timebucket in ExtractorSpec

yswdqz:


oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/extractor/ExtractorSpec.java
```diff
@@ -269,13 +274,12 @@ public void slowSql(@DelegatesTo(SlowSqlSpec.class) final Closure<?> cl) {
         DatabaseSlowStatementBuilder builder = new DatabaseSlowStatementBuilder(namingControl);
         builder.setLayer(Layer.nameOf(log.getLayer()));
 
-        LocalDateTime localDateTime = Instant.ofEpochSecond(log.getTimestamp()).atZone(ZoneId.systemDefault()).toLocalDateTime();
-        String timeBucket = DTF.format(localDateTime);
-        builder.setTimeBucket(Long.parseLong(timeBucket));
+        long timeBucket = TimeBucket.getTimeBucket(log.getTimestamp() * 1000, DownSampling.Minute);
```
wu-sheng, line 277: Can we be sure, the timestamp is in second?

yswdqz: No, but for more time user will use lua(os.time() to get current timestamp) in fluent-bit. So I use it.
I also feel that this way maybe not good.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/extractor/ExtractorSpec.java
```diff
@@ -290,13 +294,34 @@ public void slowSql(@DelegatesTo(SlowSqlSpec.class) final Closure<?> cl) {
...
+        }
+        // to endpoint traffic
+        if (StringUtil.isNotEmpty(log.getEndpoint())) {
+            EndpointMeta endpointMeta = new EndpointMeta();
+            endpointMeta.setServiceName(serviceName);
+            endpointMeta.setServiceNormal(true);
+            endpointMeta.setEndpoint(namingControl.formatEndpointName(serviceName, log.getEndpoint()));
+            endpointMeta.setTimeBucket(timeBucket);
+            sourceReceiver.receive(endpointMeta);
+        }
```
wu-sheng, line 324: I think we don't have a plan to lock slowSql for instance and endpoint level? Because the query doesn't support to do so, right?

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/extractor/ExtractorSpec.java
```diff
@@ -290,13 +294,34 @@ public void slowSql(@DelegatesTo(SlowSqlSpec.class) final Closure<?> cl) {
             return;
         }
 
+        LocalDateTime localDateTime = Instant.ofEpochSecond(log.getTimestamp()).atZone(ZoneId.systemDefault()).toLocalDateTime();
+        String timeBucketForDB = DTF.format(localDateTime);
+        builder.setTimeBucket(Long.parseLong(timeBucketForDB));
```
wu-sheng, line 299: Relate to https://github.com/apache/skywalking/pull/9988/files#r1026572253. The two ways seem very different, but for nearly the same thing?

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/extractor/ExtractorSpec.java
yswdqz: Got it.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/extractor/ExtractorSpec.java
yswdqz: I forgot Timebucket class also support second unit. I will fix it later.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/extractor/ExtractorSpec.java
wu-sheng: In the log definition, the report format is a timestamp. ref 

https://github.com/apache/skywalking-data-collect-protocol/blob/c2c9bbbff43f0ad9f917a0a55285538a6b45739e/logging/Logging.proto#L44

So, from my understanding, you should move this `times 1000` into the fluentbit side.

Ref `RecordSinkListener#L79`

wu-sheng:
@hanahmily This should be able to fix your issue.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/extractor/ExtractorSpec.java
yswdqz: I understand.

wu-sheng (approved):

PR #10004, opened 2022-11-22
[Feature] Zipkin module support BanyanDB storage.

### LINKED ISSUE #9183: Add zipkin storage BanyanDB  implementation
### Search before asking

### Description

https://github.com/apache/skywalking/pull/9177 doesn't support BanyanDB. 
Since BanyanDB has been released, we need to add the implementation.

### Use case

_No response_

### Related issues

https://github.com/apache/skywalking/pull/9177

### Code of Conduct

wankai123:
* Zipkin traces query API, sort the result set by start time by default.

- [X] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #9183 .


oap-server-bom/pom.xml
```diff
@@ -73,7 +73,7 @@
         <awaitility.version>3.0.0</awaitility.version>
         <httpcore.version>4.4.13</httpcore.version>
         <commons-compress.version>1.21</commons-compress.version>
-        <banyandb-java-client.version>0.2.0</banyandb-java-client.version>
+        <banyandb-java-client.version>0.3.0-SNAPSHOT</banyandb-java-client.version>
```
wu-sheng, line 76: @lujiajing1126 I think you are able to initialize the 0.2.1 release.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBZipkinQueryDAO.java
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    @Override
+    public List<List<Span>> getTraces(final QueryRequest request, Duration duration) throws IOException {
+        final int tracesLimit = request.limit();
+        int scrollLimit = 1000;
+        int scrollFrom = 0;
+        Set<String> traceIds = new HashSet<>();
+        while (traceIds.size() < tracesLimit) {
+            Set<String> resp = getTraceIds(request, duration, scrollFrom, scrollLimit);
```
wu-sheng, line 184: @hanahmily @lujiajing1126 Please confirm this is good enough. 

Zipkin provides trace IDs query to the Span table, and also requires distinct trace IDs according to the `limit` condition. It is using `aggregation` or `distinct .. group by` in other storage options, but BanyanDB doesn't provide this.
So, we have to do a rolling query, but as BanyanDB can't provide a rolling query(no rolling ID concept), we have to use `from... to...`

wu-sheng (approved):
If tests are passed, I am good.

wu-sheng (approved):

oap-server-bom/pom.xml
lujiajing1126: Sure. We could cherry-pick the last commit and release a patch version.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBZipkinQueryDAO.java
lujiajing1126: Shall we consider using an extra condition, e.g. `ZipkinSpanRecord.TIMESTAMP_MILLIS` instead of using `limit` and `offset`?
WDYT @hanahmily

wu-sheng: Time range is user condition, not ours. Unless we use some kind of order by and rolling according to last read record. But is the timestamp accurate enough?
 Is it millisecond? Or it is just second but formated in millisecond format? @wankai123

wankai123: The milliseconds in the span and storage are formatted from nanoseconds.
The milliseconds in the query condition are formatted from seconds, as a user can not select a milliseconds time.

Issue #10008, opened 2022-11-23
[Bug] Incorrect Timestamp used in BanyanDB


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar issues.

### Apache SkyWalking Component

BanyanDB (apache/skywalking-banyandb)

### What happened

BanyanDB used the measure/stream receiving timestamp as its timestamp.
When in the query stage, there are several cases, in which you can't read the data correctly.
1. OAP has 20-30s delay to build bulk data, so when do T1-T10 metrics query, you always can't T10 metrics correctly. Because the latest T10 value is updated in T10 + N seconds
2. Similar to <1>, the traces and logs have a delay in persistent. So, you always can't get the result correctly. And more importantly, the UI is showing the start time of the traces/logs, you would see traces and logs not in the query condition. This would be treated as bugs.

### What you expected to happen

We should use OAP persistent kernel to assign the timestamp for persistent, rather than using a timestamp by default.

### How to reproduce

Generate a trace/metric/log, and put them into the BanyanDB several minutes later, but query in the trace running duration.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

hanahmily:
BanyanDB leverage the below timestamp fields defined in the writing spec to identify the data.

1. The metrics use a "time bucket" to set this field
2. Trace segment's timestamp should be picked up.

We need more investigations on this issue.

wu-sheng:
There are at least two conclusions based on today's discussion
1. Timestamp should be added as a field of Record in the OAP. All record implementation should provide an accurate timestamp. @wankai123 
2. BanyanDB server should maintain a snowflake-style ID generator. A record ID should constitute by `record name` + `timestamp`(not time bucket) + server-generated suffix. The server-generated suffix should be held for every sharding ID in every several seconds/minutes(time window). 

The reason for setting the rolling time bucket is to avoid conflict when a new record carries an old timestamp, and the suffix ID just rolls back to 0.

- Trace ID-1, Service A, Timestamp-1, suffix-0
- Trace ID-2, Service A, Timestamp-2, suffix-1
- Trace ID-3, Service A, Timestamp-3, suffix-2
...
- Trace ID-1000, Service A, Timestamp-1000, suffix-1000
[...]

PR #10019, opened 2022-11-24
Add `@BanyanDB.TimestampColumn` to identify `which column in Record` is providing the timestamp(milliseconds) for BanyanDB

wankai123:
part of https://github.com/apache/skywalking/issues/10008

And polish banyandb Zipkin traces query.
relate to https://github.com/apache/skywalking/pull/10004#discussion_r1029510526

### DIFF
--- a/oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBConverter.java
+++ b/oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBConverter.java
@@ -67,11 +67,10 @@ public class BanyanDBConverter {
     public static class StreamToStorage implements Convert2Storage<StreamWrite> {
         private final MetadataRegistry.Schema schema;
         private final StreamWrite streamWrite;
-        private final String timestampColumn;
 
         @Override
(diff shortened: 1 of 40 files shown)


mrproliu (approved):

wu-sheng (dismissed):
LGTM. Pending on e2e checking.

wankai123:
Do not merge now, I'm checking.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBConverter.java
```diff
@@ -67,9 +67,13 @@ public byte[] getBytes(String fieldName) {
     public static class StreamToStorage implements Convert2Storage<StreamWrite> {
         private final MetadataRegistry.Schema schema;
         private final StreamWrite streamWrite;
+        private final String timestampColumn;
```
lujiajing1126, line 70: I suppose this field can be incorporated into the `MetadataRegistry.Schema` during the model parse, so we don't need this field here.

wankai123: make sense 👍

lujiajing1126 (approved):
LGTM

PR #10021, opened 2022-11-24
Enhance cache mechanism in the metric persistent process

wu-sheng:

* Enhance the cache mechanism in the metric persistent process.
  * This cache only worked when the metric is accessible(readable) from the database. Once the insert execution is delayed due to the scale, the cache loses efficacy. It only works for the last time update per minute, considering our 25s period.
  * Fix ID conflicts for all JDBC storage implementations. Due to the insert delay, the JDBC storage implementation would still generate another new insert statement.

This is a big implementation update, we need to be very careful about this.


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/SessionCacheCallback.java
```diff
@@ -0,0 +1,50 @@
...
+    public void onInsertCompleted() {
+        if (isFailed) {
+            return;
+        }
+        sessionCache.put(metrics, metrics);
+    }
+
+    public void onUpdateFailure() {
+        sessionCache.remove(metrics);
+        isFailed = true;
```
kezhenxu94, line 48: I think you need to flip these two lines

```suggestion
        isFailed = true;
        sessionCache.remove(metrics);
```

kezhenxu94 (dismissed):

kezhenxu94 (approved):

kezhenxu94 (dismissed):

wankai123 (approved):

wankai123 (approved):

PR #10026, opened 2022-11-25
Fix Elasticsearch storage: In `No-Sharding Mode`, add specific analyzer to the template before index creation to avoid update index error.

wankai123:
In the `No-Sharding Mode`:
 https://skywalking.apache.org/docs/main/next/en/faq/new-elasticsearch-storage-option-explanation-in-9.2.0/
The merged models required analyzer or not are shared the same index.
When adding a new model(with an analyzer) into an existed index by update will be failed, if the index is without analyzer settings.
To avoid this, we should add the analyzer settings to the template before index creation.

### DIFF
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
@@ -266,9 +266,10 @@ public class StorageEsInstaller extends ModelInstaller {
 
     //In the `No-Sharding Mode`:
     //https://skywalking.apache.org/docs/main/next/en/faq/new-elasticsearch-storage-option-explanation-in-9.2.0/
-    //The merged models required analyzer or not are shared the same index.
+    //Some of models require a analyzer to run match query, some others are not.
+    //They are merged into the one physical index(metrics-all or record-all)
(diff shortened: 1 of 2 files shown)


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/StorageEsInstaller.java
```diff
@@ -264,8 +264,25 @@ protected Map<String, Object> createSetting(Model model) throws StorageException
         return setting;
     }
 
-    private Map getAnalyzerSetting(List<ModelColumn> analyzerTypes) throws StorageException {
+    //In the `No-Sharding Mode`:
+    //https://skywalking.apache.org/docs/main/next/en/faq/new-elasticsearch-storage-option-explanation-in-9.2.0/
+    //The merged models required analyzer or not are shared the same index.
+    //When adding a new model(with an analyzer) into an existed index by update will be failed, if the index is without analyzer settings.
+    //To avoid this, we should add the analyzer settings to the template before index creation.
```
wu-sheng, line 271:
```suggestion
    //https://skywalking.apache.org/docs/main/next/en/faq/new-elasticsearch-storage-option-explanation-in-9.2.0/
    //Some of models require a analyzer to run match query, some others are not.
    //They are merged into the one physical index(metrics-all or record-all)
    //When adding a new model(with an analyzer) into an existed index by update will be failed, if the index is without analyzer settings.
    //To avoid this, add the analyzer settings to the template before index creation.
```

wu-sheng (dismissed):
nits for comments.

wu-sheng (approved):

PR #10032, opened 2022-11-25
Fix `BanyanDB.ShardingKey` annotation missed in the generated OAL metrics classes.

wu-sheng:


hanahmily (approved):

lujiajing1126 (approved):

wu-sheng:
503 from ghcr.io/apache/skywalking-java/skywalking-java:3f88d735ba2bfd1196aff946502447d4b14450c8-java8

I think ghcr is down.

PR #10035, opened 2022-11-25
Use column value name for apdex topn sorted queries

### LINKED ISSUE #10034: [Bug] AggregationQueryEsDAO sortMetrics NullPointerException for instanceScalarApdex metrics when logical shard is disabled
### Search before asking

### What happened

[armeria-eventloop-epoll-4-4] WARN  [] - Exception while fetching data (/instanceScalarApdex_1) : null
java.lang.NullPointerException: null
        at org.apache.skywalking.oap.server.storage.plugin.elasticsearch.query.AggregationQueryEsDAO.sortMetrics(AggregationQueryEsDAO.java:126) 

No error.

### How to reproduce

Use a topN query for a Apdex metric.

### Anything else

_No response_

### Code of Conduct

toffentoffen:
### Fix #10034 AggregationQueryEsDAO sortMetrics NullPointerException for instanceScalarApdex metrics when logical shard is disabled
- [ ] The column name to retrieve the aggregation value was hardcoded to `value`. However, when logical sharding is disabled some metrics uses other column names to store their values..

Additional debug info:
```
Metrics Name service_instance_sla terms bucket {key=redacted, doc_count=15, percentage={value=0.0}}
valueColumnName percentage
valueColumnName value null
Metrics Name service_instance_apdex terms bucket {key=redacted, doc_count=15, value={value=null}}
valueColumnName value
valueColumnName value null
```
`valueColumnName value null` this causes the NullPointerException.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #10034

### DIFF
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/AggregationQueryEsDAO.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/AggregationQueryEsDAO.java
@@ -120,11 +120,13 @@ public class AggregationQueryEsDAO extends EsDAO implements IAggregationQueryDAO
         final List<Map<String, Object>> buckets =
             (List<Map<String, Object>>) idTerms.get("buckets");
         for (Map<String, Object> termsBucket : buckets) {
-            SelectedRecord record = new SelectedRecord();
-            record.setId((String) termsBucket.get("key"));
             Map<String, Object> value = (Map<String, Object>) termsBucket.get(valueColumnName);
-            record.setValue(String.valueOf(((Number) value.get(valueColumnName)).longValue()));
-            topNList.add(record);
+            if (value.get(valueColumnName) != null) {
+                SelectedRecord record = new SelectedRecord();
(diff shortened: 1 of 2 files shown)


oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/AggregationQueryEsDAO.java
```diff
@@ -120,11 +120,13 @@ public List<SelectedRecord> sortMetrics(final TopNCondition condition,
         final List<Map<String, Object>> buckets =
             (List<Map<String, Object>>) idTerms.get("buckets");
         for (Map<String, Object> termsBucket : buckets) {
-            SelectedRecord record = new SelectedRecord();
-            record.setId((String) termsBucket.get("key"));
             Map<String, Object> value = (Map<String, Object>) termsBucket.get(valueColumnName);
-            record.setValue(String.valueOf(((Number) value.get("value")).longValue()));
-            topNList.add(record);
+            if (value.get("value") != null) {
```
wu-sheng, line 124: This could only avoid NPE. I think @wankai123 has found the root cause, should be able to fix soon.

PR #10040, opened 2022-11-27
Add OpenSearch 1.3.6 and 2.4.0 to test matrix

kezhenxu94:

### DIFF
--- a/oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/ElasticSearchVersion.java
+++ b/oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/ElasticSearchVersion.java
@@ -43,7 +43,7 @@ public final class ElasticSearchVersion {
         this.minor = minor;
 
         if (distribution.equalsIgnoreCase("OpenSearch")) {
-            requestFactory = new V78RequestFactory(this);
+            requestFactory = new V81RequestFactory(this);
             codec = V78Codec.INSTANCE;
             return;
         }
--- a/.github/workflows/skywalking.yaml
+++ b/.github/workflows/skywalking.yaml
@@ -29,7 +29,7 @@ env:
   SW_AGENT_JDK_VERSION: 8
   MAVEN_OPTS: -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
   SEGMENT_DOWNLOAD_TIMEOUT_MINS: 5 # Cache restore timeout
-  
+
 jobs:
   license-header:
     if: (github.event_name == 'schedule' && github.repository == 'apache/skywalking') || (github.event_name != 'schedule')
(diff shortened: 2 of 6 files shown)


wu-sheng (approved):
If e2e passed, I am good.

oap-server/server-library/library-elasticsearch-client/src/test/java/org/apache/skywalking/library/elasticsearch/ITElasticSearchTest.java
```diff
@@ -104,6 +104,16 @@ public static Collection<Object[]> es() {
                                        "docker.elastic.co/elasticsearch/elasticsearch-oss"))
                     .withEnv("plugins.security.disabled", "true")
                     .withStartupTimeout(java.time.Duration.ofMinutes(5))
+            },
+            {
+                "OpenSearch 2.4.0",
```
wu-sheng, line 109: @wankai123 mentioned, we should consider to separating ElasticSearch IT from general ITs, and try to only booting ElasticSearch once per version.

wankai123: Yes, I found our ElasticSearch IT frequently fails recently and costs a lot of time.

wu-sheng: And IT had become the longest CI task in our e2e task list.

wu-sheng:
> Error:  testDocUpdate[version: OpenSearch 2.4.0](org.apache.skywalking.library.elasticsearch.ITElasticSearchTest)  Time elapsed: 14.576 s  <<< ERROR!
java.util.concurrent.CompletionException: java.lang.RuntimeException: {"error":"no handler found for uri [/test-index-update/_doc/an id/_update] and method [POST]"}
Caused by: java.lang.RuntimeException: {"error":"no handler found for uri [/test-index-update/_doc/an id/_update] and method [POST]"}

We seem to have an incompatible issue.

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/ElasticSearchVersion.java
```diff
@@ -43,7 +43,7 @@ private ElasticSearchVersion(final String distribution, final int major, final i
         this.minor = minor;
 
         if (distribution.equalsIgnoreCase("OpenSearch")) {
-            requestFactory = new V78RequestFactory(this);
+            requestFactory = new V81RequestFactory(this);
             codec = V78Codec.INSTANCE;
```
wu-sheng, line 47: V81 Factory with V78 code?

wu-sheng:
BTW, let's put support versions of OpenSearch on the doc. https://skywalking.apache.org/docs/main/next/en/setup/backend/backend-storage/#opensearch

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/ElasticSearchVersion.java
kezhenxu94: > V81 Factory with V78 code?

`RequestFactory` is to compose the API endpoint, such as `_doc/_update/<id>`, `<index>/_update/<id>`, etc., while `Codec` is to (de)serialize the request body, so they don't have to match in version. For example, they may change the endpoint from `_doc/_update/<id>` to `/_update/<id>` but the (de)serialization of the request body is unchanged.

wankai123 (approved):

PR #10044, opened 2022-11-29
Adding HTTP/1.x metrics tab on network profiling widget

mrproliu:
image
image
image
image
image
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>.
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/docs/en/changes/changes.md).

PR #10045, opened 2022-11-29
Fix missing HTTP/1.x metrics in network profiling

mrproliu:
Following https://github.com/apache/skywalking/pull/10044, This PR is fixed missing metrics for analysis HTTP/1.x metrics.


wu-sheng (approved):
Let's add e2e for this if possible.

PR #10049, opened 2022-11-29
Fix wrong time bucket on attached event

mrproliu:
Fix the wrong time bucket on the attached event, otherwise, the event will write to the monthly index in ES storage.

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -137,6 +137,7 @@
 * Rename `BanyanDB.ShardingKey` to `BanyanDB.SeriesID`.
 * Self-Observability: Add counters for metrics reading from DB or cached. Dashboard:`Metrics Persistent Cache Count`.
 * Self-Observability: Fix `GC Time` calculation.
+* Update the `trace_id` field as storage only(cannot be queried) in `top_n_database_statement`, `top_n_cache_read_command`, `top_n_cache_read_command` index.
 
 #### UI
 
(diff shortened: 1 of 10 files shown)


wu-sheng (approved):

wu-sheng:
This seems a thing could be covered by e2e.

wankai123 (approved):

mrproliu:
> This seems a thing could be covered by e2e.

Do you mean using e2e to check the index name or time bucket value?

wu-sheng:
I think this can't be read if the time bucket is wrong, right?

mrproliu:
> I think this can't be read if the time bucket is wrong, right?

No, I have done the code research. When we query the records, we just use the alias name, such as `records-all`, the wrong index name will be `records-all-YYYYMM`. But for the long term, the data cannot be queried, because we have the `TTL` mechanism to delete the wrong and older indexes. 
In the E2E, unless we query it for a long time, so we cannot find this kind of problem.

wankai123:
> > I think this can't be read if the time bucket is wrong, right?
> 
> No, I have done the code research. When we query the records, we just use the alias name, such as `records-all`, the wrong index name will be `records-all-YYYYMM`. But for the long term, the data cannot be queried, because we have the `TTL` mechanism to delete the wrong and older indexes. In the E2E, unless we query it for a long time, so we cannot find this kind of problem.

I think we should query like this:
```java
        final SearchResponse response = getClient().search(
            new TimeRangeIndexNameGenerator(
                IndexController.LogicIndicesRegister.getPhysicalTableName(SegmentRecord.INDEX_NAME),
                startSecondTB,
                endSecondTB
            ), search.build());
```

wu-sheng:
Yes, we should use physical indices to query, it has much better performance.

mrproliu:
Sure, but for the attached event query, we need to calculate all of the spans to get the start/end time. If the event timestamp does not contain in the span timestamp range(such as has offset between span and event), It may not be queried in some extreme situations.

wu-sheng:
You reminded me, I think there are some things wrong with the SpanAttachedEventRecord annotation about BanyanDB.

Please use `@BanyanDB.GlobalIndex` for the trace ID. And we need to think about the `@SeriesID`

wu-sheng:
For `SampledStatus4xxTraceRecord` and others, please move `@BanyanDB.SeriesID(index = 0)` to entity ID.

wu-sheng (dismissed):

wu-sheng (approved):

Issue #10051, opened 2022-11-29
[Feature] [OAP] Consider removing metrics reload in persistent


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

In 9.3.0, we enhanced the cache mechanism in #10021 and verified through #10025. The `ID` read by the persistent worker at the minute dimensionality now has 60% cached. Only the first bulk of every minute would require `ID` read.

When we look deeper into the logic, we would find out actually that, **from the minute dimensionality, this `ID` read is not required**.
We are concerned that no metric in the cache doesn't mean there is no metric in the database. Especially when 
1. The timestamp is not synced in the cluster, so, timestamps of telemetry are not ordered by the time series.
2. OAP is booting/rebooting, and the cache is cold.

About <1>, we don't expect this anymore. Our TTL and metric/topo analysis are all relying on timestamps generally synced. It doesn't have to be synced in the `ms` level, but at least with only a 3-5s gap.

About <2>, we only should try to load metrics from DB in the 1 minute after rebooting, considering the assumption about time synced in the <1>. So, the metrics would overlap existing metrics generated in one booting period. 
There is little chance we faced data conflicts, even if we faced them, we just generate metrics at the booting minute inaccurate. In the best practice, we could keep loading metrics from the database when metrics timestamps are before the **OAP started timestamp** as a fail-safe.

Regarding hour and day dimensionalities, there is nothing different. We just should keep `loading metrics from database` when the hour/day time bucket before **OAP started hour / day**.

@hanahmily @kezhenxu94 @wankai123 PTAL. Considering 9.3.0 releasing soon, I don't want to take a risk to change for now.
But the theory should be correct, please help on rechecking.

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wu-sheng:
FYI @apache/skywalking-committers if you have interests in SkyWalking OAP kernel logic.

wu-sheng:
Ideally, due to this enhancement, the cached metrics should go closing to 95%+, and highly reduce the IOPS of the database(read part). 
The OAP would only read from database when CLI/UI applies for query metrics/traces/logs.

hanahmily:
[...]

PR #10083, opened 2022-12-03
Add `ServerStatusService` in the core module

wu-sheng:

Add `ServerStatusService` in the core module to provide a new way to expose the booting status to other modules. This helps on #10051


oap-server/server-starter/src/main/java/org/apache/skywalking/oap/server/starter/OAPServerBootstrap.java
```diff
@@ -43,12 +42,10 @@ public static void start() {
...
 
-            manager.find(TelemetryModule.NAME)
+            manager.find(CoreModule.NAME)
                    .provider()
-                   .getService(MetricsCreator.class)
-                   .createGauge("uptime", "oap server start up time", MetricsTag.EMPTY_KEY, MetricsTag.EMPTY_VALUE)
-                   // Set uptime to second
-                   .setValue(System.currentTimeMillis() / 1000d);
+                   .getService(ServerStatusService.class)
+                   .bootedNow(System.currentTimeMillis());
```
kezhenxu94, line 48: It doesn't mean the OAP starts up **successfully** when this line is executed, no?

kezhenxu94 (commented):
I think we already have the mechanism to get the server status. https://github.com/apache/skywalking/blob/18008aa6b24ada6602d9c73e1a9d19b6105d8b59/oap-server/server-health-checker/src/main/java/org/apache/skywalking/oap/server/health/checker/provider/HealthQueryService.java#L28. I prefer to move that health checker module into core if you really need that in core module.

oap-server/server-starter/src/main/java/org/apache/skywalking/oap/server/starter/OAPServerBootstrap.java
wu-sheng: It is. I did this because, in https://github.com/apache/skywalking/issues/10051, you would see the booting uptime matter. So I moved this setup into the core as a unique mechanism.
I don't intend to replace the existing checker.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/status/ServerStatusService.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public void bootedNow(long uptime) {
+        bootingStatus.setBooted(true);
+        bootingStatus.setUptime(uptime);
+        manager.find(TelemetryModule.NAME)
+               .provider()
+               .getService(MetricsCreator.class)
+               .createGauge("uptime", "oap server start up time", MetricsTag.EMPTY_KEY, MetricsTag.EMPTY_VALUE)
+               // Set uptime to second
+               .setValue(uptime / 1000d);
```
wu-sheng, line 48: I moved the logic here to keep consistent. @kezhenxu94

wu-sheng: The next step we should need a registration and notification mechanism in this service for other modules, which could sense cluster module update events or latest status.

kezhenxu94: This doesn't make sense to me. You are essentially proposing to replicate the health checker module in the core module again. What's the boundary to do or not to do this kind of stuff in your mind? You said you don't want to replace the health checker module but it can not persuade me and we have no reason to stop future contributors from doing similar things. The new server status service and the old health checker are the same to me except for the fact that they are written by different authors.

wu-sheng: In next PR, I need to read boot time, I don't think this makes sense to read data from health checker. 

Thr boundaries are health check is for exposing true false status and error message for out of the process. The core service is for modules in the OAP to interact or read status.

wu-sheng: At least, when @hanahmily added health checker, it only cares the OAP is running or not, rather than inside details. 
For example, we are not going to expose cluster information to checker, or out of process. Right?

From data point, uptime is already  both in health check and telemetry, two components. .

wu-sheng:
If I follow what you are asking, do you mean other modules should react based on the telemetry module's data? That is not correct from my understanding.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/status/ServerStatusService.java
wu-sheng: > You are essentially proposing to replicate the health checker module in the core module again

I am not, I am not sure why you think this way. I just add the boot status, because I just need it for the next step. But in https://github.com/apache/skywalking/issues/10051, more status relative mechanism is going to be added.

kezhenxu94:
> If I follow what you are asking, do you mean other modules should react based on the telemetry module's data? That is not correct from my understanding.

I don't mean that. Let's say in another way, if you insist we should add the new service in core module, I hope finally we can rewrite the telemetry module to reuse the server status logics in core module, that should make sense right?

wu-sheng:
> I don't mean that. Let's say in another way, if you insist we should add the new service in core module, I hope finally we can rewrite the telemetry module to reuse the server status logics in core module, that should make sense right?

I am afraid I can from tech perspective. The Telemetry is the leaf of the module dependencies. That is why I am adding a layer in the core as a delegation. 

https://github.com/apache/skywalking/blob/18008aa6b24ada6602d9c73e1a9d19b6105d8b59/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleProvider.java#L401-L407

kezhenxu94 (approved):

wankai123 (approved):

PR #10084, opened 2022-12-04
Refactor session cache in MetricsPersistentWorker.

wu-sheng:

This is another preparation for #10051. After this, I would be more confident to move the booting timestamp as a time point to control whether metrics should be loaded from the database or return not existing if it can't be found in the cache.
Also, this makes the cache controller out of the worker process.


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsSessionCache.java
```diff
@@ -0,0 +1,81 @@
...
+import lombok.Setter;
+import org.apache.skywalking.oap.server.core.analysis.metrics.Metrics;
+import org.apache.skywalking.oap.server.library.client.request.InsertRequest;
+import org.apache.skywalking.oap.server.library.client.request.UpdateRequest;
+
+/**
+ * MetricsSessionCache is a key-value cache to hold hot metric in-memory to reduce payload to pre-read.
+ *
+ * There are two ways to make sure metrics in-cache,
+ * 1. Metrics is read from the Database through {@link MetricsPersistentWorker}.loadFromStorage
```
kezhenxu94, line 34: 👇 suggestion so we can directly jump to the method?

```suggestion
 * 1. Metrics is read from the Database through {@link MetricsPersistentWorker#loadFromStorage}
```

wankai123 (approved):

wankai123 (approved):

kezhenxu94 (approved):
One nit

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsSessionCache.java
wu-sheng: Tried and failed :P This is a private method in the worker

image

kezhenxu94: 🥹 ok then. LGTM

Issue #10076, opened 2022-12-02
[Feature] Support listeners for the `RemoteClients` change.


wankai123:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

    In the cluster mode, OAP instances send metrics to each other according to the `remote-hash code` to do L2 aggregation and alarm calculations for balancing​. Some of the calculations leverage the previously received metrics in the memory cache and incoming metrics. So when the OAP cluster scales,  the metrics might lose their cache and impact the calculations.
   To avoid this situation, I think we should support listeners for the `RemoteClients` change, then the stream workers can be notified and reload previous metrics from storage for calculation.

### Use case

1. Provide listeners' interface for notification and a registration mechanism.
2. When the `RemoteClients` changes, notice all listeners.

### Related issues

https://github.com/apache/skywalking/issues/10051

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

kezhenxu94:
This is duplicate of https://github.com/apache/skywalking/issues/9814 and I think @fgksgf is already working on this

wu-sheng:
Let's see how @fgksgf would do. I think we need more work after this mechanism is ready.
Is there an ETA for review?

kezhenxu94:
> Let's see how @fgksgf would do. I think we need more work after this mechanism is ready.
> Is there an ETA for review?

Before https://github.com/apache/skywalking/issues/10051 this issue is a "nice to have" enhancement so we don't set an ETA, but if it impacts the metrics in https://github.com/apache/skywalking/issues/10051, we could prioritize this.

wu-sheng:
#10051 doesn't rely on this as a hard requirement. Alerting enhancement requires this more.

wankai123:
I think #9814  is not exactly the same. #9814 is trying to fix the delay between the  OAP fetched `RemoteClients` and the real instance status.  But this issue wants to resolve the other OAP module doesn't know the `RemoteClients` is changed.

kezhenxu94:
> I think #9814  is not exactly the same. #9814 is trying to fix the delay between the  OAP fetched `RemoteClients` and the real instance status.  But this issue wants to resolve the other OAP module services doesn't know the `RemoteClients` is changed.

[...]

PR #10085, opened 2022-12-05
Cache enhancement - don't read new metrics from database in minute 

### LINKED ISSUE #10051: [Feature] [OAP] Consider removing metrics reload in persistent

wu-sheng:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #10051.

Through this PR, the metrics in minutes would mostly not read metrics from the database anymore. This should be able to be observed in by so11y's `metrics_persistent_cache`.

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -7,6 +7,20 @@
 * Add `ServerStatusService` in the core module to provide a new way to expose booting status to other modules.
 * Adds Micrometer as a new component.(ID=141)
 * Refactor session cache in MetricsPersistentWorker.
+* Cache enhancement - don't read new metrics from database in minute dimensionality.
+```
+    // When
+    // (1) the time bucket of the server's latest stability status is provided
+    //     1.1 the OAP has booted successfully
+    //     1.2 the current dimensionality is in minute.
+    // (2) the metrics are from the time after the timeOfLatestStabilitySts
+    // (3) the metrics don't exist in the cache
+    // the kernel should NOT try to load it from the database.
+    //
+    // Notice, about condition (2),
+    // for the specific minute of booted successfully, the metrics are expected to load from database when
+    // it doesn't exist in the cache.
+```
 
 #### UI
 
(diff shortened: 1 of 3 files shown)


docs/en/changes/changes.md
```diff
@@ -7,6 +7,20 @@
 * Add `ServerStatusService` in the core module to provide a new way to expose booting status to other modules.
 * Adds Micrometer as a new component.(ID=141)
 * Refactor session cache in MetricsPersistentWorker.
+* Cache enhancement - don't read new metrics from database in minute dimensionality.
+```
+    // When
+    // (1) the time bucket of the server's latest stability status is provided
+    //     1.1 the OAP has booted successfully
+    //     1.2 the current dimensionality is in minute.
```
wu-sheng, line 15: @wankai123 1.3 should be related to #10076 

```
1.3 the OAP cluster is rebalanced due to scaleing
```

wankai123 (approved):
LGTM

PR #10098, opened 2022-12-05
Remove timeout offset and correct `MetricsExtension` declarations

wu-sheng:

* Remove the offset of metric session timeout according to worker creation sequence.
* Correct `MetricsExtension` annotations declarations in manual entities.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/relation/process/ProcessRelationClientSideMetrics.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/relation/process/ProcessRelationClientSideMetrics.java
@@ -39,7 +39,8 @@ import org.apache.skywalking.oap.server.core.storage.type.StorageBuilder;
         builder = ProcessRelationClientSideMetrics.Builder.class, processor = MetricsStreamProcessor.class)
 @MetricsExtension(supportDownSampling = false, supportUpdate = false, timeRelativeID = true)
 @EqualsAndHashCode(of = {
-        "entityId"
+        "entityId",
+        "component_id"
 }, callSuper = true)
 @SQLDatabase.Sharding(shardingAlgorithm = ShardingAlgorithm.NO_SHARDING)
 public class ProcessRelationClientSideMetrics extends Metrics {
@@ -73,7 +74,7 @@ public class ProcessRelationClientSideMetrics extends Metrics {
 
     @Override
     protected String id0() {
-        return getTimeBucket() + Const.ID_CONNECTOR + entityId;
+        return getTimeBucket() + Const.ID_CONNECTOR + entityId + Const.ID_CONNECTOR + componentId;
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/relation/process/ProcessRelationClientSideMetrics.java
```diff
@@ -37,7 +37,7 @@
 
 @Stream(name = ProcessRelationClientSideMetrics.INDEX_NAME, scopeId = DefaultScopeDefine.PROCESS_RELATION,
         builder = ProcessRelationClientSideMetrics.Builder.class, processor = MetricsStreamProcessor.class)
-@MetricsExtension(supportDownSampling = false, supportUpdate = true, timeRelativeID = true)
+@MetricsExtension(supportDownSampling = false, supportUpdate = false, timeRelativeID = true)
```
wu-sheng, line 40: @mrproliu Please check this, this metric seems having no point to update.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/relation/process/ProcessRelationServerSideMetrics.java
```diff
@@ -37,7 +37,7 @@
 
 @Stream(name = ProcessRelationServerSideMetrics.INDEX_NAME, scopeId = DefaultScopeDefine.PROCESS_RELATION,
         builder = ProcessRelationServerSideMetrics.Builder.class, processor = MetricsStreamProcessor.class)
-@MetricsExtension(supportDownSampling = false, supportUpdate = true, timeRelativeID = true)
+@MetricsExtension(supportDownSampling = false, supportUpdate = false, timeRelativeID = true)
```
wu-sheng, line 40: @mrproliu Please check this, this metric seems having no point to update.

wankai123 (dismissed):

mrproliu (dismissed):

wu-sheng:
```
- 		map[interface{}]interface{}{
 - 			string("detectpoints"):     []interface{}{string("CLIENT")},
 - 			string("id"):               string("a6998b690466d1d80fa9379dafd9145d6a6d8b45ab49949c9405ddecd169960d-b456c20e471dc1ca4dfa06529b6e3cf320916c49bd12dcd847de8856ab321fd3"),
 - 			string("source"):           string("a6998b690466d1d80fa9379dafd9145d6a6d8b45ab49949c9405ddecd169960d"),
 - 			string("sourcecomponents"): []interface{}{string("tcp")},
 - 			string("target"):           string("b456c20e471dc1ca4dfa06529b6e3cf320916c49bd12dcd847de8856ab321fd3"),
 - 			string("targetcomponents"): []interface{}{},
 - 		},
 + 		map[interface{}]interface{}{
 + 			string("detectpoints"):     []interface{}{string("CLIENT"), string("SERVER")},
 + 			string("id"):               string("a6998b690466d1d80fa9379dafd9145d6a6d8b45ab49949c9405ddecd169960d-e68bf4959a6e1d10f6d9032662f1187f0325[445](https://github.com/apache/skywalking/actions/runs/3620492727/jobs/6104321005#step:7:452)a348bd5708a4f87e71420d3bb"),
 + 			string("source"):           string("a6998b690466d1d80fa9379dafd9145d6a6d8b45ab49949c9405ddecd169960d"),
 + 			string("sourcecomponents"): []interface{}{string("http")},
 + 			string("target"):           string("e68bf4959a6e1d10f6d9032662f1187f0325445a348bd5708a4f87e71420d3bb"),
 + 			string("targetcomponents"): []interface{}{string("http")},
```

It seems this relationship query logic is not correct. You never put the component ID in the ID.
Because I remove metrics persistent offsite,  there is no guarantee there is TCP or HTTP component ID in the relationship. I think you should either 
1. Support component IDs as a list, and support update when new component ID is added
2. Put the component ID in the ID.

wu-sheng:
I am trying option <2>, let's see.

One question, is the aggregation still good when `storageOnly = true` is set now?
```
    @Setter
    @Getter
    @Column(columnName = COMPONENT_ID, storageOnly = true)
    private int componentId;
```

mrproliu:
Got it. I will fix this PR.

mrproliu (approved):

PR #10127, opened 2022-12-07
use `core/persistentPeriod` + 5s as `refresh_interval` for all indices

wu-sheng:

I deeply run some research on ElasticSearch and asked for help from the Elastic China team. 
From ElasticSearch 6.0, the operation for the doc has seg to guarantee the order.

Ref https://www.elastic.co/cn/blog/elasticsearch-sequence-ids-6-0

From what the seg was introduced, we should be confident that `reflush period` doesn't have to be short. 

Ref from official doc,  https://www.elastic.co/guide/en/elasticsearch/reference/current/tune-for-indexing-speed.html#_unset_or_increase_the_refresh_interval 
Ref from an old blog, https://sematext.com/blog/elasticsearch-refresh-interval-vs-indexing-performance/

They all show a longer refresh period could bring higher performance.

Finally, in our current case, 30s latency to show metrics should be a good balance point. As usual, the persistent period is 25s. @kezhenxu94 @wankai123 Does this make sense?


wankai123 (dismissed):

wu-sheng:
@kezhenxu94 Considering we have a certain index `refresh_period`, do you think it is reasonable to shorten the time of BulkProcess' flushInterval? It is 15s for now, plus the persistent period(25s), the worth case is 25s + 15s + 30s(refresh_interval) to make the data visible.

Of course, this only happens in a small dataset. We could keep the current setting for now.

wankai123 (approved):
make sense to me

kezhenxu94 (approved):

Issue #10104, opened 2022-12-06
[BanyanDB][OAP] Move sampled metrics from measure-default to measure-sampled 


hanahmily:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

The downsampled metrics in minute dimension, their name has a suffix `_minute`, should go to a separate measure group called `measure-sampled`. 

The new `measure-sampled` has a 1-day/24-hour segment/block interval rule as the global settings. 

`measure-default` should use 10-day/240-hour strategy, which is more than the TTL, 7-day. Therefore, `measure-default` will open 2 blocks at most. On the 8th day, 80% of the block’s interval, the next block will be created. The data will keep in last 19 days at most. 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #10128, opened 2022-12-07
Optimize BanyanDB interval rules

### LINKED ISSUE #10102: [BanyanDB][OAP] Optimize interval rule of segment and block.
### Search before asking

### Description

There are some flaws in segment/block interval rules configurations that should be twisted:

  * The unit of segment interval should be *DAY*. The number defaults to 1 day.
  * The default block interval should be 24 hours.
  * Support setting the interval rule based on the group name more than global defaults.

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

lujiajing1126:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #10102 and #10104.

In order to address #10104, both interval of block/segment are determined by the `MeasureTTL` now. So configurations for Measure's intervals are no longer needed.

Then for #10102, the Unit of `streamSegmentInterval` is changed to days.

Besides, fix groups of metadata, e.g. `*Traffic`, `Labels`, etc.


oap-server/server-starter/src/main/resources/application.yml
```diff
@@ -236,10 +236,8 @@ storage:
     superDatasetShardsFactor: ${SW_STORAGE_BANYANDB_SUPERDATASET_SHARDS_FACTOR:2}
     concurrentWriteThreads: ${SW_STORAGE_BANYANDB_CONCURRENT_WRITE_THREADS:15}
     profileTaskQueryMaxSize: ${SW_STORAGE_BANYANDB_PROFILE_TASK_QUERY_MAX_SIZE:200} # the max number of fetch task in a request
-    streamBlockInterval: ${SW_STORAGE_BANYANDB_STREAM_BLOCK_INTERVAL:4} # Unit is hour
-    streamSegmentInterval: ${SW_STORAGE_BANYANDB_STREAM_SEGMENT_INTERVAL:24} # Unit is hour
-    measureBlockInterval: ${SW_STORAGE_BANYANDB_MEASURE_BLOCK_INTERVAL:4} # Unit is hour
-    measureSegmentInterval: ${SW_STORAGE_BANYANDB_MEASURE_SEGMENT_INTERVAL:24} # Unit is hour
+    streamBlockInterval: ${SW_STORAGE_BANYANDB_STREAM_BLOCK_INTERVAL:24} # Unit is hour
```
hanahmily, line 239: The key should be `blockIntervalHours`, which defaults to `4` hours as before.

oap-server/server-starter/src/main/resources/application.yml
```diff
@@ -236,10 +236,8 @@ storage:
...
+    streamSegmentInterval: ${SW_STORAGE_BANYANDB_STREAM_SEGMENT_INTERVAL:1} # Unit is day
```
hanahmily, line 240: `segmentIntervalDays` is the key.

hanahmily (changes requested):

oap-server/server-starter/src/main/resources/application.yml
lujiajing1126: Done

oap-server/server-starter/src/main/resources/application.yml
lujiajing1126: Done

lujiajing1126:
CI is fine now. PTAL @hanahmily @wu-sheng

oap-server/server-starter/src/main/resources/application.yml
```diff
@@ -235,10 +235,9 @@ storage:
     superDatasetShardsFactor: ${SW_STORAGE_BANYANDB_SUPERDATASET_SHARDS_FACTOR:2}
     concurrentWriteThreads: ${SW_STORAGE_BANYANDB_CONCURRENT_WRITE_THREADS:15}
     profileTaskQueryMaxSize: ${SW_STORAGE_BANYANDB_PROFILE_TASK_QUERY_MAX_SIZE:200} # the max number of fetch task in a request
-    streamBlockInterval: ${SW_STORAGE_BANYANDB_STREAM_BLOCK_INTERVAL:4} # Unit is hour
-    streamSegmentInterval: ${SW_STORAGE_BANYANDB_STREAM_SEGMENT_INTERVAL:24} # Unit is hour
-    measureBlockInterval: ${SW_STORAGE_BANYANDB_MEASURE_BLOCK_INTERVAL:4} # Unit is hour
-    measureSegmentInterval: ${SW_STORAGE_BANYANDB_MEASURE_SEGMENT_INTERVAL:24} # Unit is hour
+    blockIntervalHours: ${SW_STORAGE_BANYANDB_BLOCK_INTERVAL_HOURS:24} # Unit is hour
+    segmentIntervalDays: ${SW_STORAGE_BANYANDB_SEGMENT_INTERVAL_DAYS:1} # Unit is day
+    overrideGroupIntervals: ${SW_STORAGE_BANYANDB_OVERRIDE_GROUP_INTERVALS:""} # For example, group1,4,1;group2,2,2
```
wu-sheng, line 240: What is `overrideGroupIntervals`?

lujiajing1126: > What is `overrideGroupIntervals`?

Mentioned in issue https://github.com/apache/skywalking/issues/10102, we have to Support setting the interval rule based on the group name more than global defaults.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBStorageConfig.java
```diff
@@ -57,31 +57,24 @@ public class BanyanDBStorageConfig extends ModuleConfig {
...
-     * @since 9.3.0
+     * @since 9.4.0
      */
-    private int streamSegmentInterval;
+    private int segmentIntervalDays;
     /**
-     * block interval for Measure group.
-     * Unit is hours.
+     * Override the default global configurations.
+     * It works for all groups except `measure-default`.
```
wu-sheng, line 75: What do you mean `override` default? Who set the default?

lujiajing1126: > What do you mean `override` default? Who set the default?

`blockIntervalHours` and `segmentIntervalDays` are now global interval rules

oap-server/server-starter/src/main/resources/application.yml
wu-sheng: Do you mean a thing similar with index specific settings in elasticsearch?

```
    # Specify the settings for each index individually.
    # If configured, this setting has the highest priority and overrides the generic settings.
    specificIndexSettings: ${SW_STORAGE_ES_SPECIFIC_INDEX_SETTINGS:""}
```

lujiajing1126: > Do you mean a thing similar with index specific settings in elasticsearch?
> 
> ```
>     # Specify the settings for each index individually.
>     # If configured, this setting has the highest priority and overrides the generic settings.
>     specificIndexSettings: ${SW_STORAGE_ES_SPECIFIC_INDEX_SETTINGS:""}
> ```

I think so

wu-sheng: Then we should rename it to `specificGroupSettings`. And the value format should be in JSON for better extension in the future. What do you think?

lujiajing1126: > Then we should rename it to `specificGroupSettings`. And the value format should be in JSON for better extension in the future. What do you think?

Make sense to me

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBIndexInstaller.java
```diff
@@ -60,9 +61,9 @@ public boolean isExists(Model model) throws StorageException {
             // then check entity schema
             if (metadata.findRemoteSchema(c).isPresent()) {
                 // register models only locally but not remotely
-                if (model.isTimeSeries() && model.isRecord()) { // stream
+                if (model.isRecord()) { // stream
                     MetadataRegistry.INSTANCE.registerStreamModel(model, config, configService);
-                } else if (model.isTimeSeries() && !model.isRecord()) { // measure
```
wu-sheng, line 65: I don't think this could be removed. Not `model.isTimeSeries()` model exists.

wu-sheng: I don't think we should register UI_TEMPLATE as a measure, right?

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBIndexInstaller.java
```diff
@@ -78,20 +79,18 @@ public boolean isExists(Model model) throws StorageException {
...
         try {
             ConfigService configService = moduleManager.find(CoreModule.NAME).provider().getService(ConfigService.class);
-            if (model.isTimeSeries() && model.isRecord()) { // stream
+            if (model.isRecord()) { // stream
                 Stream stream = MetadataRegistry.INSTANCE.registerStreamModel(model, config, configService);
                 if (stream != null) {
                     log.info("install stream schema {}", model.getName());
                     ((BanyanDBStorageClient) client).define(stream);
                 }
-            } else if (model.isTimeSeries() && !model.isRecord()) { // measure
```
wu-sheng, line 87: Same as https://github.com/apache/skywalking/pull/10128/files#r1045088460

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBIndexInstaller.java
lujiajing1126: We've already checked at the beginning of `isExists`,

```java
@Override
    public boolean isExists(Model model) throws StorageException {
        if (!model.isTimeSeries()) {
            return true;
        }
        ...
}
```

So I think we can safely ignore other conditions?

oap-server/server-starter/src/main/resources/application.yml
```diff
@@ -235,10 +235,9 @@ storage:
     superDatasetShardsFactor: ${SW_STORAGE_BANYANDB_SUPERDATASET_SHARDS_FACTOR:2}
     concurrentWriteThreads: ${SW_STORAGE_BANYANDB_CONCURRENT_WRITE_THREADS:15}
     profileTaskQueryMaxSize: ${SW_STORAGE_BANYANDB_PROFILE_TASK_QUERY_MAX_SIZE:200} # the max number of fetch task in a request
-    streamBlockInterval: ${SW_STORAGE_BANYANDB_STREAM_BLOCK_INTERVAL:4} # Unit is hour
-    streamSegmentInterval: ${SW_STORAGE_BANYANDB_STREAM_SEGMENT_INTERVAL:24} # Unit is hour
-    measureBlockInterval: ${SW_STORAGE_BANYANDB_MEASURE_BLOCK_INTERVAL:4} # Unit is hour
-    measureSegmentInterval: ${SW_STORAGE_BANYANDB_MEASURE_SEGMENT_INTERVAL:24} # Unit is hour
+    blockIntervalHours: ${SW_STORAGE_BANYANDB_BLOCK_INTERVAL_HOURS:24} # Unit is hour
+    segmentIntervalDays: ${SW_STORAGE_BANYANDB_SEGMENT_INTERVAL_DAYS:1} # Unit is day
```
wu-sheng, line 239: A question for @hanahmily, do we need a hard-coded specific setting for superset records? We had this for ElasticSearch.

oap-server/server-starter/src/main/resources/application.yml
```diff
@@ -235,10 +235,9 @@ storage:
...
+    specificGroupSettings: ${SW_STORAGE_BANYANDB_SPECIFIC_GROUP_SETTINGS:""} # For example, {"group1": {"blockIntervalHours": 4, "segmentIntervalDays": 1}}
```
wu-sheng, line 240: @lujiajing1126 Do we have the static group list? If so, we could add the group list here, and in the `field specificGroupSettings`'s comment.

lujiajing1126: We have several groups now: `measure-default`, `measure-sampled`, `stream-default` and `stream-{superdataset}`. We can add some comments here

lujiajing1126: Done

wu-sheng (approved):
Pending on https://github.com/apache/skywalking/pull/10128/files#r1045093697 answer, others are good to me.

oap-server/server-starter/src/main/resources/application.yml
hanahmily: Yes, we should add a superset dedicated settings here. We can override settings through the newly added `overrideGroupInterval`, but the group name will be changed from time to time. A compiling time flag, superset, might apply these settings to renamed groups. @lujiajing1126

oap-server/server-starter/src/main/resources/application.yml
hanahmily: Could you refer to `https://skywalking.apache.org/docs/skywalking-banyandb/next/crud/group/#list-operation` in the comment? Users could use `bydbctl` to get all groups in the server.

oap-server/server-starter/src/main/resources/application.yml
wu-sheng: `overrideGroupInterval` had been changed to `specificGroupSettings`.

lujiajing1126:
@hanahmily PTAL

wu-sheng:
> Yes, we should add a superset dedicated settings here.

@lujiajing1126 I think you missed this one?

lujiajing1126:
> > Yes, we should add a superset dedicated settings here.
> 
> @lujiajing1126 I think you missed this one?

There are `superDatasetSegmentIntervalDays` and `superDatasetBlockIntervalHours` now

wu-sheng (approved):
LGTM

hanahmily (approved):

PR #10135, opened 2022-12-08
Optimize `flushInterval` of ElasticSearch BulkProcessor 

wu-sheng:

* Optimize `flushInterval` of ElasticSearch BulkProcessor to avoid extra periodical flush in the continuous bulk streams. 

Run periodical flush if there is no `flushIfNeeded` executed in the second half of the flush period.
Otherwise, wait for the next round. By default, the last 2 seconds of the 5s period.
This could avoid periodical flush running among bulks(controlled by bulkActions).

@kezhenxu94 This is following my #10129, also from my review about `BulkProcessor` logic. Please check whether this makes sense.


wankai123 (approved):
LGTM

kezhenxu94 (approved):

PR #10142, opened 2022-12-11
Fix MAL exp combine error when only contains one segment and expPrefix not blank

pg-yang:
In MAL  config, if `expPrefix` is not blank and `exp` only contains one segment (doesn't contain any dot), The combined exp would end with a dot, and encounter `org.codehaus.groovy.control.MultipleCompilationErrorsException`

For example :

```yaml 

expPrefix: tag({tags -> tags.service_name = 'APISIX::'+(tags['skywalking_service']?.trim()?:'APISIX')})
metricsRules:
  # Service
  - name: sv_http_connections
    exp: apisix_nginx_http_current_connections
```

The combined exp would be like  `(apisix_nginx_http_current_connections.tag({tags -> tags.service_name = 'APISIX::'+(tags['skywalking_service']?.trim()?:'APISIX')})).`

### DIFF
--- a/oap-server/analyzer/meter-analyzer/src/test/java/org/apache/skywalking/oap/meter/analyzer/MetricConvertTest.java
+++ b/oap-server/analyzer/meter-analyzer/src/test/java/org/apache/skywalking/oap/meter/analyzer/MetricConvertTest.java
@@ -1,0 +1,236 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


oap-server/analyzer/meter-analyzer/src/test/java/org/apache/skywalking/oap/meter/analyzer/MetricConvertTest.java
```diff
@@ -0,0 +1,236 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import org.apache.skywalking.oap.server.core.analysis.meter.MeterSystem;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class MetricConvertTest {
```
wu-sheng, line 32: Could you share in which following cases would fail before the fix? And how the wrong expression looks like.

wu-sheng:
I want to know does this PR block your AWS monitoring. Han and Kai faced some COVID positive issues.
I am not familiar with this part, but I could try my best to check your chances.

wu-sheng (approved):
LGTM.

OK, I think I have followed your case.
A more accurate description should be 
**An unexpected dot is added when exp is a pure metric name and expPrefix != null`.**

The final expression was (`metric_name`(exp).`expPrefix`)`.`(this is the extra dot)

pg-yang:
Thanks for your help.

Issue #10033, opened 2022-11-25
[Feature] Discard `id()` method in the BanyanDB Measure process


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

All metrics from OAL/MAL/Manual have an `id` method to generate a unique ID for this metric for persistence. 
So, in the current implementation, there is an extra tag(Measure.ID) created by BanyanDB storage implementation.
But, in the BanyanDB kernel level, ID is just a simple tag, which relies on an index structure, but not a time-series concept.
Meanwhile, the current ShardingKey would be set to ID when not set.

Considering all the above, we would face the low efficiency `BanyanDBMetricsDAO#multiGet` which relies on ID.at

### Use case

To improve the circumstance, we should 
1. Hard require for ShardingKey. No default sharding key, and no ID tag.
2. Add ShardingKey to all metadata and manual metrics implementations.
3. Rename ShardingKey to **SeriesID**, which indicate this is not just for sharding, but also used as a ID with timestamp.

`BanyanDBMetricsDAO#multiGet` should only use the current time bucket and IDs extracted from a new `BanyanDBConverter#MeasureToStorage` implementation(only read SeriesIDs) to build `multiGet` query. 

In the BanyanDB implementation, the `ID` would never be used in the query.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wu-sheng:
FYI @lujiajing1126 Please take a look at this proposal. This is from the discussion between @hanahmily and me, when we try to locate https://github.com/apache/skywalking/pull/10032

wu-sheng:
Fix <3> through https://github.com/apache/skywalking/pull/10037

PR #10157, opened 2022-12-14
Refactor `StorageData#id` to the new StorageID object from a String type.

wu-sheng:

The last preparation for #10033. @hanahmily @lujiajing1126 This makes your BanyanDB optimization possible.
Please take a look. 

@kezhenxu94 @wankai123 @mrproliu  This change would impact a lot of parts of the codes. Please do a recheck.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/StorageID.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/StorageID.java
@@ -1,0 +1,159 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.oap.server.core.storage;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import org.apache.skywalking.oap.server.library.util.StringUtil;
+
+/**
+ * StorageID represents an identification for the metric or the record.
+ * Typically, an ID is composited by two parts
+ * 1. Time bucket based on downsampling.
+ * 2. The encoded entity ID, such as Service ID.
+ *
+ * In the SQL database and ElasticSearch, the string ID is preferred.
+ * In the BanyanDB, time series and entity ID(series ID) would be treated separately.
+ *
+ * @since 9.4.0 StorageID replaced the `string id()` method in the StorageData. An object-oriented ID provides a more
+ * friendly interface for various database implementation.
+ */
+@EqualsAndHashCode(of = {
+    "fragments"
+})
+public class StorageID {
+    private final List<Fragment> fragments;
+    /**
+     * Once the storage ID was {@link #build()} or {@link #read()},
+     * this object would switch to the sealed status, no more append is allowed.
+     */
+    private boolean sealed = false;
+
+    public StorageID() {
+        fragments = new ArrayList<>(2);
+    }
+
+    public StorageID append(String name, String value) {
+        if (StringUtil.isBlank(name)) {
+            throw new IllegalArgumentException("The name of storage ID should not be null or empty.");
+        }
+        if (sealed) {
+            throw new IllegalStateException("The storage ID is sealed. Can't append a new fragment, name=" + name);
+        }
+        fragments.add(new Fragment(name, String.class, false, value));
+        return this;
+    }
+
+    public StorageID append(String name, long value) {
+        if (StringUtil.isBlank(name)) {
+            throw new IllegalArgumentException("The name of storage ID should not be null or empty.");
+        }
+        if (sealed) {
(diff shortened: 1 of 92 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/StorageID.java
```diff
@@ -0,0 +1,159 @@
...
+ * In the SQL database and ElasticSearch, the string ID is preferred.
+ * In the BanyanDB, time series and entity ID(series ID) would be treated separately.
+ *
+ * @since 9.4.0 StorageID replaced the `string id()` method in the StorageData. An object-oriented ID provides a more
+ * friendly interface for various database implementation.
+ */
+@EqualsAndHashCode(of = {
+    "fragments"
+})
+public class StorageID {
```
wu-sheng, line 45: To all reviewers, this is the key point of all changes. This StorageID would represent the ID to replace the original string ID.

wu-sheng:
**Notice, after this change, should be no feature change.**

_I noticed, during the changes, some codes are using an object rather than an explicit String type, which could be missed from compiling level check._ 
This is the risk I will bring bugs due to this change.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/StorageID.java
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * @return the string ID concatenating the values of {@link #fragments} by the underline(_).
+     */
+    public String build() {
+        sealed = true;
+        String[] idParts = new String[fragments.size()];
+        for (int i = 0; i < fragments.size(); i++) {
+            idParts[i] = fragments.get(i).toString();
+        }
+        return StringUtil.join('_', idParts);
```
kezhenxu94, line 115:
```suggestion
        return com.google.common.base.Joiner.on("_").join(fragments);
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/StorageID.java
```diff
@@ -0,0 +1,159 @@
...
+         */
+        private final Class<?> type;
+        /**
+         * If true, the field was from {@link #name}, and value is changed by internal rules.
+         * Such as time bucket downsampling, use a day-level time-bucket to build the ID for a minute dimension metric.
+         */
+        private final boolean mutate;
+        private final Object value;
+
+        public Optional getName() {
```
kezhenxu94, line 150:
```suggestion
        public Optional<String> getName() {
```

kezhenxu94 (approved):

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/StorageID.java
wu-sheng: Great, I was searching for this, but failed.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/StorageID.java
```diff
@@ -0,0 +1,157 @@
...
+        /**
+         * Represent the class type of the {@link #value}.
+         */
+        private final Class<?> type;
+        /**
+         * If true, the field was from {@link #name}, and value is changed by internal rules.
+         * Such as time bucket downsampling, use a day-level time-bucket to build the ID for a minute dimension metric.
+         */
+        private final boolean mutate;
+        private final Object value;
```
wu-sheng, line 146: @hanahmily @lujiajing1126 You would need this when you try to build BanyanDB series ID without string ID.

Notice, we should check whether there is conflict(s) between series IDs and fragments.

hanahmily: It looks good to me.

wankai123 (dismissed):
LGTM

mrproliu (dismissed):

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/StorageID.java
```diff
@@ -0,0 +1,190 @@
...
+        /**
+         * The column name of the value, or the original column names of the mutate value.
+         *
+         * The names could be
+         * 1. Always one column if this is not {@link #mutate} and from a certain persistent column.
+         * 2. Be null if {@link #mutate} is true and no relative column, such as the original value is not in
+         * the persistence.
+         * 3. One or multi-values if the value is built through a symmetrical or asymmetrical encoding algorithm.
+         */
+        private final String[] name;
```
wu-sheng, line 144: @lujiajing1126 @hanahmily The name has been changed to an Array. There are 3 conditions for different cases.

lujiajing1126: Just to confirm the rule,

1. if a model uses mutant fragments, it **MUST NOT** use `@SeriesID` annotation on its fields
2. Otherwise, we must define `@SeriesID` in the model

wu-sheng: I think @hanahmily has a more complex proposal. @hanahmily Could you share your ideas here?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/StorageID.java
```diff
@@ -0,0 +1,190 @@
...
+
+            return true;
+        }
+
+        @Override
+        public int hashCode() {
+            int result = Arrays.hashCode(name);
+            result = 31 * result + (value != null ? value.hashCode() : 0);
+            return result;
+        }
```
kezhenxu94, line 188: The reason why you have to write these instead of using Lombok `EqualsAndHashCode(of = {"name", "value"})` is that, you have a `getName` that returns an `Optional<String[]>`, Lombok can compare array correctly but not when arrays are wrapped into `Optional`, but luckily you can change the behaviour of Lombok and not to use the getter, but using this :


```java
    @EqualsAndHashCode(of = {
        "name",
        "value"
    }, doNotUseGetters = true)
```

kezhenxu94 (approved):
A nit that you might want to fix it in next PR, so LGTM for now.

kezhenxu94 (approved):

PR #10159, opened 2022-12-15
Fix some metrics session wouldn't expire after TTL.

wankai123:
Such as service_traffic is not `supported updates`, but the id is not `TimeRelative`.


docs/en/changes/changes.md
```diff
@@ -40,6 +40,7 @@
 * Add global-specific settings used to override global configurations (e.g `segmentIntervalDays`, `blockIntervalHours`) in BanyanDB.
 * Use TTL-driven interval settings for the `measure-default` group in BanyanDB.
 * Fix wrong group of non time-relative metadata in BanyanDB.
+* Fix some metrics session wouldn't expire after TTL.
```
wu-sheng, line 43:
```suggestion
```

We don't need a change log for this, because this bug was brought in this iteration(9.4-dev) only.

wu-sheng (approved):
LGTM. FYI @kezhenxu94

PR #9680, opened 2022-09-25
fix oap and ui image

tsonglew:
change images to OAP_IMAGE=apache/skywalking-oap-server:latest, UI_IMAGE=apache/skywalking-ui:latest as previous images not found

- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>.
- [ ] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/docs/en/changes/changes.md).

PR #10163, opened 2022-12-15
Update .env

myie9:
make docker can run ok


wu-sheng:
This docker file is not for downloading from docker hub.
This is for local build.

kezhenxu94:
Duplicate of https://github.com/apache/skywalking/pull/9680

myie9:
> * [ ]  If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #.
> *

it can not run when local run  docker compose

wu-sheng:
Again, it is only used when you have built the image locally, not using the images on the DockerHub.

PR #10164, opened 2022-12-16
Support multiple component IDs in the service topology level

wu-sheng:

* Support multiple component IDs in the service topology level.
* Add `ElasticSearch.Keyword` annotation to declare the target field type as `keyword`.
* [Breaking Change] Field `component_id` of `service_relation_client_side` and `service_relation_server_side` have been replaced by `component_ids`(keyword).


wu-sheng:
NOTICE, SkyWalking booster UI is not using component IDs in the line(Call in GraphQL). We only show the component ID(one value) for the node.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/TopologyQueryEsDAO.java
```diff
@@ -257,7 +257,7 @@ private List<Call.CallDetail> buildProcessRelation(String serviceInstanceId, Dur
             String entityId = (String) entityBucket.get("key");
             final Map<String, Object> componentTerms =
                 (Map<String, Object>) entityBucket.get(
-                    ServiceRelationServerSideMetrics.COMPONENT_IDS);
+                    ProcessRelationServerSideMetrics.COMPONENT_ID);
```
wu-sheng, line 260: @mrproliu I think this is a bug by using service relation column name.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/manual/relation/service/ServiceRelationClientSideMetrics.java
```diff
@@ -60,8 +62,9 @@ public class ServiceRelationClientSideMetrics extends Metrics {
     private String destServiceId;
     @Setter
     @Getter
-    @Column(columnName = COMPONENT_ID, storageOnly = true)
-    private int componentId;
+    @Column(columnName = COMPONENT_IDS, storageOnly = true)
+    @ElasticSearch.Keyword
+    private IntList componentIds = new IntList(3);
```
wu-sheng, line 67: @kezhenxu94 @wankai123 This is what I referred to, so I added this new annotation to override type. IntList is mapping to text type field.

> The field can be [Keyword](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/keyword.html), [Numeric](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/number.html), [ip](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/ip.html), [boolean](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/boolean.html), or [binary](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/binary.html).

ref https://www.elastic.co/guide/en/elasticsearch/reference/7.17/search-aggregations-bucket-terms-aggregation.html

wu-sheng:
@toffentoffen I think you could check whether this works in your cases.

wu-sheng:
@wankai123 @kezhenxu94 @mrproliu I am thinking to add priority mechanism to the component ID to make sure the node ID is always correct. What do you think?

kezhenxu94:
> @toffentoffen I think you could check whether this works in your cases.

@wu-sheng Can you share what cases this PR tries to solve?

wu-sheng:
> > @toffentoffen I think you could check whether this works in your cases.
> 
> @wu-sheng Can you share what cases this PR tries to solve?

ServA is calling ServB both in TCP and HTTP. In the history, two different RPC components(two HTTP client libs) called the same service endpoint.

kezhenxu94 (approved):

Issue #10000, opened 2022-11-22
[Feature] [Zipkin UI] Bundle Zipkin Lens UI in our distribution as a widget component


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

@wankai123  enhanced our Zipkin support in past months. Ref doc, https://skywalking.apache.org/docs/main/next/en/setup/backend/zipkin-trace/

Now, we are more confident to work as an alternative server for people who wants their Zipkin traces to stay side by side with other metrics/logs in an all-in-one APM.

@Fine0830 The lens UI source codes are here, https://github.com/openzipkin/zipkin/tree/master/zipkin-lens. It is a very lightweight UI component, I am considering we could bundle them in our distribution, and you could wrap it as a widget to let the dashboard host it on a tab page.

### Use case

@wankai123 could help you to set up test env for testing. 

@Fine0830 There is no feature change requirement. You need to decide whether we should copy/paste those codes, which may be helpful if we need to bump up dependencies to fix CVE or expired libs. 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wu-sheng:
Notice, if we are going to copy them, we should keep their license headers, and put the statement in the LICENSE and NOTICE files of source codes and binary, as we are doing re-distribution.

wu-sheng:
@tacigar told me there are several enhancements he wants to land on the code base, 

- https://github.com/openzipkin/zipkin/pull/3348
- https://github.com/openzipkin/zipkin/pull/3349
- https://github.com/openzipkin/zipkin/pull/3350

They are tested and used for a private distribution. We should adopt those too.

wu-sheng:
The one last step is to wait for @Fine0830 to provide a new kind of widget, which could host the Zipkin lens UI on an isolated iFramework. This widget would work separately.

wu-sheng:
@kezhenxu94 Let's leverage the latest UI widget and bring Zipkin Lens UI into the Service Mesh dashboard. And consider opening Zipkin tracing in Istio of Showcase?

kezhenxu94:
@wu-sheng did you expect to put the Zipkin UI here? 
[...]

PR #10167, opened 2022-12-16
Add Zipkin UI to webapp

kezhenxu94:

image

image


dist-material/release-docs/LICENSE
```diff
@@ -340,6 +340,7 @@ The text of each license is the standard Apache 2.0 license.
     https://mvnrepository.com/artifact/io.vavr/vavr/0.10.3 Apache-2.0
     https://mvnrepository.com/artifact/io.vavr/vavr-match/0.10.3 Apache-2.0
     https://mvnrepository.com/artifact/io.zipkin.zipkin2/zipkin/2.23.16 Apache-2.0
+    https://mvnrepository.com/artifact/io.zipkin/zipkin-lens/2.23.16 Apache-2.0
```
wu-sheng, line 343: I think we should resolve lens UI's dependencies, and regenerate the LICENSE file

wu-sheng:
@Fine0830 This would provide webapp integration for Zipkin lens UI. 

Ref https://github.com/apache/skywalking/issues/10000

wu-sheng:
@kezhenxu94 We may consider how to showcase this. How about just opening the Zipkin tracer on Envoy in the mesh? Then we at least could have some spans.

kezhenxu94:
> @kezhenxu94 We may consider how to showcase this. How about just opening the Zipkin tracer on Envoy in the mesh? Then we at least could have some spans.

We already open Envoy SkyWalking tracer in showcase, there is no way to open both Zipkin tracer and SkyWalking tracer at the same time

kezhenxu94:
> > @kezhenxu94 We may consider how to showcase this. How about just opening the Zipkin tracer on Envoy in the mesh? Then we at least could have some spans.
> 
> We already open Envoy SkyWalking tracer in showcase, there is no way to open both Zipkin tracer and SkyWalking tracer at the same time

OK, we didn't open Envoy tracer for now, but I remember @mrproliu planned to do so recently

wu-sheng:
> > @kezhenxu94 We may consider how to showcase this. How about just opening the Zipkin tracer on Envoy in the mesh? Then we at least could have some spans.
> 
> We already open Envoy SkyWalking tracer in showcase, there is no way to open both Zipkin tracer and SkyWalking tracer at the same time

I can't find a trace including the envoy span. I think we don't, otherwise, the topology should be very strange.(the alias name could be strange unless we set the `not-analysis` attribute to true for the segments from envoy)
We just instrument the services, and sidecars forward the `sw8` headers.

Also, we even don't have a trace tab on the mesh dashboard.

oap-server-bom/pom.xml
```diff
@@ -304,6 +304,12 @@
                 <artifactId>zipkin</artifactId>
                 <version>${zipkin.version}</version>
             </dependency>
+            <!-- SkyWalking Web UI -->
+            <dependency>
+                <groupId>io.zipkin</groupId>
+                <artifactId>zipkin-lens</artifactId>
+                <version>${zipkin.version}</version>
```
wu-sheng, line 311: We should bump up the version to `2.23.19`?(maybe)

dist-material/release-docs/LICENSE
kezhenxu94: To regenerate the license file (and make it possible to automatically generate it as what we are doing now), we have to copy the `package.json` and `package-lock.json` from Zipkin Lens project, also Zipkin Lens put all dependencies (including development dependencies) into `dependencies` section so we would also include many un-distributed licenses.

wu-sheng:
> > > @kezhenxu94 We may consider how to showcase this. How about just opening the Zipkin tracer on Envoy in the mesh? Then we at least could have some spans.
> > 
> > 
> > We already open Envoy SkyWalking tracer in showcase, there is no way to open both Zipkin tracer and SkyWalking tracer at the same time
> 
> OK, we didn't open Envoy tracer for now, but I remember @mrproliu planned to do so recently

@mrproliu and you could decide whether we need Zipkin context propagation in our demo service. That is your call.
It is just more fancy if we did.

mrproliu:
> > > @kezhenxu94 We may consider how to showcase this. How about just opening the Zipkin tracer on Envoy in the mesh? Then we at least could have some spans.
> > 
> > 
> > We already open Envoy SkyWalking tracer in showcase, there is no way to open both Zipkin tracer and SkyWalking tracer at the same time
> 
> OK, we didn't open Envoy tracer for now, but I remember @mrproliu planned to do so recently

I just noticed that envoy generates the zipkin format trace context by default when forwarding HTTP protocols. If we want to show that both envoy and skywalking traces are included, we can enhance the demo service to support both skywalking and zipkin traces.

dist-material/release-docs/LICENSE
wu-sheng: OK, this seems a little crazy :(
It looks like we are over declaring more than we should be.

wu-sheng: Let's keep the lib dependencies for now. Copy the package and lock files seem easy to be out-of-date too.

kezhenxu94: > Let's keep the lib dependencies for now. Copy the package and lock files seem easy to be out-of-date too.

I don't think we will update the lens dependency too often, this feature is not in our main path and the lens dependency has only static compiled html/css/js files

dist-material/release-docs/LICENSE
```diff
@@ -195,21 +195,104 @@ The text of each license is also included in licenses/LICENSE-[project].txt.
     https://npmjs.com/package/vis-timeline/v/7.5.1 7.5.1 (Apache-2.0 OR MIT)
     https://npmjs.com/package/vis-util/v/5.0.3 5.0.3 (Apache-2.0 OR MIT)
 
+========================================================================
```
wu-sheng, line 198: It seems you added a new zipkin-License, I think it is good.
Then the following new could be removed?

wu-sheng: We could add a line `Zipkin-lens dependences are attached in `zipkin-LICENSE` file. Apache 2.0`(Or something similar)

wu-sheng (approved):
LGTM

PR #10169, opened 2022-12-16
Enhance service topology to support multiple components of a service

wu-sheng:

* Support `priority` definition in the `component-libraries.yml`. 
* Enhance service topology query. When there are multiple components detected from the server side, 
  the component type of the node would be determined by the priority, which was random in the previous release.

@mrproliu `* Support component IDs' priority in process relation metrics.` is re-implemented by this new mechanism.


wu-sheng:
The component ID in service and process topology are sharing the same mechanism from the kernel level now. 

After this is merged, I would consider removing the component ID from the instance topology. If you have any concerns, please let me know.
My idea is, we never show this, and as an `n pods(ServA) * m pods(ServB)` size topology, there is pointless to show the component of each node.
This would be only a logical level change, I will add a notice to the GraphQL file to declare these cases. And will keep those type/component fields as empty strings only.

mrproliu (approved):

wankai123 (approved):

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/NotGraphQLField.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * NotGraphQLField represents the annotated field is not used by GraphQL protocol.
+ *
+ * Notice, this is a dev-level friendly label, there is no practical use in the runtime.
+ *
+ * @since 9.4.0
+ */
+@Target({ElementType.FIELD})
+public @interface NotGraphQLField {
```
kezhenxu94, line 32: This annotation looks to be useless

wu-sheng: I just don't want comments everywhere. This annotation would not be compiled into the binary.

kezhenxu94 (approved):

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/NotGraphQLField.java
kezhenxu94: > I just don't want comments everywhere. This annotation would not be compiled into the binary.

It WOULD be compiled into binary, if you want to skip compiling it into binary, please annotation with `@Retention(java.lang.annotation.RetentionPolicy.SOURCE)`, the default retention policy is runtime

PR #10170, opened 2022-12-16
Remove component_id from the instance topology

wu-sheng:

* Remove `component_id` from `service_instance_relation_client_side` and `service_instance_relation_server_side`.
* Remove unnecessary columns from topology query in the BanyanDB.


oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBTopologyQueryDAO.java
```diff
@@ -129,8 +129,6 @@ List<Call.CallDetail> queryServiceRelation(Duration duration,
             MeasureQueryResponse resp = query(modelName,
                                               ImmutableSet.of(
                                                   ServiceRelationClientSideMetrics.COMPONENT_IDS,
-                                                  ServiceRelationClientSideMetrics.SOURCE_SERVICE_ID,
-                                                  ServiceRelationClientSideMetrics.DEST_SERVICE_ID,
```
wu-sheng, line 133: @lujiajing1126 @hanahmily Please review this class.

These columns are actually not used in final analysis. Entity ID should be enough for decoding.

BTW, BanyanDB implementation is not using `Aggregation` query, so with 30-60 min, we face a huge dataset of query result. Please recheck.

wankai123 (approved):

PR #10180, opened 2022-12-19
Fix cannot display attached events when using Zipkin Lens UI query traces

mrproliu:
Fixed the issue in Zipkin Lens UI where attached events were not displayed when searching for traces. The case was caused by the trace display not triggering a new query after a search.


oap-server/server-query-plugin/zipkin-query-plugin/src/main/java/org/apache/skywalking/oap/query/zipkin/handler/ZipkinQueryHandler.java
```diff
@@ -304,6 +306,16 @@ private byte[] encodeTraces(List<List<Span>> traces) {
...
 
+    private void appendEventsToTraces(List<List<Span>> spans) throws IOException {
+        for (List<Span> span : spans) {
+            if (CollectionUtils.isEmpty(span)) {
+                continue;
+            }
+
+            appendEvents(span, getSpanAttachedEventQueryDAO().querySpanAttachedEvents(SpanAttachedEventTraceType.ZIPKIN, span.get(0).traceId()));
+        }
+    }
```
wankai123, line 317:
```suggestion
    private void appendEventsToTraces(List<List<Span>> traces) throws IOException {
        for (List<Span> spans : traces) {
            if (CollectionUtils.isEmpty(spans)) {
                continue;
            }

            appendEvents(spans, getSpanAttachedEventQueryDAO().querySpanAttachedEvents(SpanAttachedEventTraceType.ZIPKIN, spans.get(0).traceId()));
        }
    }
```

wankai123 (approved):
Some nits.

oap-server/server-query-plugin/zipkin-query-plugin/src/main/java/org/apache/skywalking/oap/query/zipkin/handler/ZipkinQueryHandler.java
```diff
@@ -304,6 +306,16 @@ private byte[] encodeTraces(List<List<Span>> traces) {
         return buff.array();
     }
 
+    private void appendEventsToTraces(List<List<Span>> traces) throws IOException {
+        for (List<Span> spans : traces) {
+            if (CollectionUtils.isEmpty(spans)) {
+                continue;
+            }
+
+            appendEvents(spans, getSpanAttachedEventQueryDAO().querySpanAttachedEvents(SpanAttachedEventTraceType.ZIPKIN, spans.get(0).traceId()));
```
wu-sheng, line 315: Should consider a list query rather than this.
10-20 iterations would be very slow.

mrproliu: Updated, please re-check.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/stream/BanyanDBSpanAttachedEventQueryDAO.java
```diff
@@ -51,11 +51,11 @@ public BanyanDBSpanAttachedEventQueryDAO(BanyanDBStorageClient client) {
     }
 
     @Override
-    public List<SpanAttachedEventRecord> querySpanAttachedEvents(SpanAttachedEventTraceType type, String traceId) throws IOException {
+    public List<SpanAttachedEventRecord> querySpanAttachedEvents(SpanAttachedEventTraceType type, List<String> traceIds) throws IOException {
         final StreamQueryResponse resp = query(SpanAttachedEventRecord.INDEX_NAME, TAGS, new QueryBuilder<StreamQuery>() {
             @Override
             protected void apply(StreamQuery query) {
-                query.and(eq(SpanAttachedEventRecord.RELATED_TRACE_ID, traceId));
+                query.and(having(SpanAttachedEventRecord.RELATED_TRACE_ID, traceIds));
```
wu-sheng, line 58: @hanahmily @lujiajing1126 Should we use `have` or `or`? Or the latest `in`?

wu-sheng: I think `have` means having all conditions. @mrproliu Please check through e2e or locally

wu-sheng: `in` should be a right term, https://github.com/apache/skywalking/pull/10181

mrproliu: I have tested locally, the `have` only matches the first trace. I think I need to wait for the `in` condition. The `or` condition looks like it needs all the conditions to be `or` to be valid.

wu-sheng: You put one trace ID or all to query? I think if we use multiple trace IDs, having would match nothing.

wu-sheng: Generally, we should use one query to get all events for all traces once.

mrproliu: I put all trace IDs to the query, It matches the first trace(in the ID list) result.

mrproliu: > Generally, we should use one query to get all events for all traces once.

Yes, so I think I need to wait for the `in` condition. Also, I found two issues when storing and querying. 
1. The `sampled_slow_trace_record` cannot store the `latency` field, It throws `invalid ref to tag latency`.
2. The `span_attached_event_record` fields `timestamp` is missing when query.

wu-sheng: @lujiajing1126 Could you follow this?

lujiajing1126: > @lujiajing1126 Could you follow this?

Oh, I didn't notice this PR.

> @hanahmily @lujiajing1126 Should we use `have` or `or`? Or the latest `in`?

From the semantics, it should be `in` here. `having` means subset, usually for `Set`/`List` typed tag.

wu-sheng (approved):

Issue #10166, opened 2022-12-16
[BanyanDB][OAP] Apply "IN" to topology queries


hanahmily:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

`BanyanDBTopologyQueryDAO` uses several sub-queries to fetch relations. Once #10165 merged, the DAO could leverage the "IN" operation to compose a single query to reduce the I/O overhead. 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #10181, opened 2022-12-19
Refactor Topology query for BanyanDB

### LINKED ISSUE #10171: [BanyanDB][OAP] Refactor topology query
There are two issues here:

### Redundant tags
`sourceService` and `destService ` are not used in the final analysis. Entity ID should be enough for decoding.

### Use `groupBy` in the query 

`TopologyDAO` is not using `groupBy` query, so with 30-60 min, we face a huge dataset of the query results.

_Originally posted by @wu-sheng in https://github.com/apache/skywalking/pull/10170#discussion_r1050716464_

lujiajing1126:
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #10171, #10166 .


lujiajing1126:
Ready for review @wu-sheng @hanahmily

oap-server-bom/pom.xml
```diff
@@ -73,7 +73,7 @@
         <awaitility.version>3.0.0</awaitility.version>
         <httpcore.version>4.4.13</httpcore.version>
         <commons-compress.version>1.21</commons-compress.version>
-        <banyandb-java-client.version>0.2.1</banyandb-java-client.version>
+        <banyandb-java-client.version>0.3.0-SNAPSHOT</banyandb-java-client.version>
```
wu-sheng, line 76: A TODO before 9.4.0 release.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBTopologyQueryDAO.java
```diff
@@ -123,72 +118,77 @@ List<Call.CallDetail> queryServiceRelation(Duration duration,
...
-        });
+                List<AbstractCriteria> instanceRelationsQueryConditions = new ArrayList<>(2);
 
-        queryBuilderList.add(new QueryBuilder<MeasureQuery>() {
-            @Override
-            protected void apply(MeasureQuery query) {
-                query.and(eq(ServiceInstanceRelationServerSideMetrics.DEST_SERVICE_ID, clientServiceId))
-                     .and(eq(ServiceInstanceRelationServerSideMetrics.SOURCE_SERVICE_ID, serverServiceId));
+                instanceRelationsQueryConditions.add(
+                        // source_service_id = clientServiceId AND dest_service_id = serverServiceId
```
wu-sheng, line 173: What is this commented code about?

oap-server-bom/pom.xml
lujiajing1126: added

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBTopologyQueryDAO.java
lujiajing1126: The comment describes the query condition below

wu-sheng (approved):

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBTopologyQueryDAO.java
```diff
@@ -303,25 +300,35 @@ List<Call.CallDetail> queryProcessRelation(Duration duration,
...
-                                          Collections.emptySet(), timestampRange, new QueryBuilder<MeasureQuery>() {
-                @Override
-                protected void apply(MeasureQuery query) {
-                    query.and(eq(ProcessRelationServerSideMetrics.SERVICE_INSTANCE_ID, serviceInstanceId));
+                ImmutableSet.of(Metrics.ENTITY_ID, ProcessRelationClientSideMetrics.COMPONENT_ID),
+                Collections.emptySet(), timestampRange, new QueryBuilder<MeasureQuery>() {
+                    @Override
+                    protected void apply(MeasureQuery query) {
+                        query.and(eq(ProcessRelationServerSideMetrics.SERVICE_INSTANCE_ID, serviceInstanceId));
+                        query.groupBy(Sets.newHashSet(Metrics.ENTITY_ID, ProcessRelationServerSideMetrics.COMPONENT_ID));
```
hanahmily, line 311: * The HashSet doesn't maintain the order of tags. Could you opt for a sortable set instead? 
* Please add `component_id` to the measure's entity list. The order of tags in groupBy should match the order of entities. That will help improve the memory overhead

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBTopologyQueryDAO.java
```diff
@@ -74,43 +78,34 @@ public List<Call.CallDetail> loadServiceRelationDetectedAtClientSide(Duration du
...
-            });
-        }
-        return queryBuilderList;
+    private QueryBuilder<MeasureQuery> buildServiceRelationsQuery(List<String> serviceIds) {
+        return new QueryBuilder<MeasureQuery>() {
+            @Override
+            protected void apply(MeasureQuery query) {
+                query.or(in(ServiceRelationServerSideMetrics.SOURCE_SERVICE_ID, serviceIds))
+                        .or(in(ServiceRelationServerSideMetrics.DEST_SERVICE_ID, serviceIds));
+                query.groupBy(Sets.newHashSet(Metrics.ENTITY_ID, ServiceRelationServerSideMetrics.COMPONENT_IDS));
```
hanahmily, line 102: The same as the comment left at line:311

hanahmily (changes requested):

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBTopologyQueryDAO.java
wu-sheng: component id exists in service topology only, now.

lujiajing1126: Fixed

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBTopologyQueryDAO.java
lujiajing1126: You mean mark component_id and entity_id with `@SeriesID`?

hanahmily: Yes. But the component_id doesn't exist in process_relation. From https://github.com/apache/skywalking/pull/10181#discussion_r1051883005, we could remove component_id from groupBy's tag list.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBTopologyQueryDAO.java
hanahmily: Please use `@SeriesID` to mark `entity_id` and `component_id`

hanahmily: Could you upload the service_relation_server_side and service_relation_client_side measure's schema here for double-checking?

wu-sheng: They don't have `SeriesID` for now, please add them. `component_id` don't have to be a part of `SeriesID`, because they have chances to update when service IDs are not changed. 

@hanahmily Is that OK to keep it out of `SeriesID` but query them out as `group by`.

lujiajing1126: I've added missing `@Series` to only `entity_id`.

lujiajing1126:
### ServiceRelationServiceSide

<details>
<summary>Schema</summary>

```protobuf
measure {
  metadata {
    group: "measure-default"
    name: "service_relation_server_side_hour"
  }
  tag_families {
    name: "default"
    tags {
      name: "source_service_id"
      type: TAG_TYPE_STRING
    }
    tags {
      name: "dest_service_id"
      type: TAG_TYPE_STRING
    }
    tags {
      name: "entity_id"
      type: TAG_TYPE_STRING
    }
    tags {
      name: "time_bucket"
      type: TAG_TYPE_INT
    }
    tags {
      name: "id"
      type: TAG_TYPE_ID
    }
  }
  tag_families {
    name: "storage-only"
    tags {
      name: "component_ids"
      type: TAG_TYPE_STRING
    }
  }
  entity {
    tag_names: "entity_id"
  }
  interval: "1h"
}
```
</details>

### ServiceRelationClientSide

<details>
<summary>Schema</summary>

```protobuf
measure {
  metadata {
    group: "measure-default"
    name: "service_relation_client_side_hour"
  }
  tag_families {
    name: "default"
    tags {
      name: "source_service_id"
      type: TAG_TYPE_STRING
    }
    tags {
      name: "dest_service_id"
      type: TAG_TYPE_STRING
    }
    tags {
      name: "entity_id"
      type: TAG_TYPE_STRING
    }
    tags {
      name: "time_bucket"
      type: TAG_TYPE_INT
    }
    tags {
      name: "id"
      type: TAG_TYPE_ID
    }
  }
  tag_families {
    name: "storage-only"
    tags {
      name: "component_ids"
      type: TAG_TYPE_STRING
    }
  }
  entity {
    tag_names: "entity_id"
  }
  interval: "1h"
}
```
</details>

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBTopologyQueryDAO.java
lujiajing1126: > @hanahmily Is that OK to keep it out of `SeriesID` but query them out as `group by`.

Yes. It is possible. But `groupBy` with entityID(s) save much memory according to the current implementation. If groupBy keys are not entityIDs, then hash-based physical plan will be used. Probably we may still need `componentIDs` as a part of the entity? @hanahmily

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBTopologyQueryDAO.java
lujiajing1126: > Yes. But the component_id doesn't exist in process_relation. From [#10181 (comment)](https://github.com/apache/skywalking/pull/10181#discussion_r1051883005), we could remove component_id from groupBy's tag list.

It is `ProcessRelationServerSideMetrics.COMPONENT_ID`. I think we need this in groupBy @wu-sheng, right?

wu-sheng: Oops, sorry.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBTopologyQueryDAO.java
hanahmily: > For update, this could be `ServA->ServB with component 1`, then update to `ServA->ServB with component 1,2,3`. 
> 
> This is the same entity from OAP perspective.

`GroupBy` will output two items in this scenario if the components ids is a part of the group key, whether it belongs to the seriesID or not. In other words, marking component ids as a part of seriesID affects the performance, not the semantics. 

If this is the case, I don't see any side effects to adding component ids to seriesID.

lujiajing1126: Annotations added

hanahmily (approved):

PR #10185, opened 2022-12-20
Bump up Zipkin and Zipkin lens UI dependency to 2.24.0

wu-sheng:

### DIFF
--- a/dist-material/release-docs/LICENSE
+++ b/dist-material/release-docs/LICENSE
@@ -339,8 +339,8 @@ The text of each license is the standard Apache 2.0 license.
     https://mvnrepository.com/artifact/io.swagger/swagger-annotations/1.6.6 Apache-2.0
     https://mvnrepository.com/artifact/io.vavr/vavr/0.10.3 Apache-2.0
     https://mvnrepository.com/artifact/io.vavr/vavr-match/0.10.3 Apache-2.0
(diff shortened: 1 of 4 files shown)


wankai123 (dismissed):

kezhenxu94:
No sure but you might need to regenerate the zipkin-license from the new Lens dependency's package.json and package-lock.json， if they have new dependencies or versions changes in Lens, this is needed.

wu-sheng:
> No sure but you might need to regenerate the zipkin-license from the new Lens dependency's package.json and package-lock.json， if they have new dependencies or versions changes in Lens, this is needed.

Could you provide which command I should run? I thought it is a part of CI check.

wu-sheng:
Do you mean I need to copy the lock file again from the Zipkin repo?

kezhenxu94:
> Do you mean I need to copy the lock file again from the Zipkin repo?

You have to regenerate the zipkin-LICENSE from the package.json and package-lock.json in https://github.com/openzipkin/zipkin/tree/2.24.0/zipkin-lens, this is no part of CI because they are using npm 14 and we are using npm 16. Also, even we copy their package.json and package-lock.json, we have to manually copy those 2 files every time we upgrade the zipkin-lens version. So I'd just check in the generated ziplin-LICENSE file instead of check in the 2 large files

wu-sheng:
@kezhenxu94 Updated.

kezhenxu94 (approved):

PR #10187, opened 2022-12-20
fix ServiceMeshServiceDispatcher overwrite ServiceDispatcher debug file when open SW_OAL_ENGINE_DEBUG.

zhyyu:

The `org.apache.skywalking.oap.server.core.source.oal.rt.dispatcher.ServiceMeshServiceDispatcher` class will overwrite `ServiceDispatcher.class`. Change class file name logic when generate `Dispatcher` class in `openEngineDebug` circumstance.

### DIFF
--- a/oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/OALRuntime.java
+++ b/oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/OALRuntime.java
@@ -50,6 +50,7 @@ import javassist.bytecode.annotation.StringMemberValue;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.JavaVersion;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.SystemUtils;
 import org.apache.skywalking.oal.rt.output.AllDispatcherContext;
 import org.apache.skywalking.oal.rt.output.DispatcherContext;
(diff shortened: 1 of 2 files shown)


wu-sheng:
Could you explain more about how this override happens? They are two classes, why would one will override the other?

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/OALRuntime.java
```diff
@@ -523,6 +524,9 @@ private void writeGeneratedFile(CtClass metricsClass, String className, String t
                 if (!folder.exists()) {
                     folder.mkdirs();
                 }
+                if (StringUtils.equalsIgnoreCase("dispatcher", type)) {
+                    className = metricsClass.getSimpleName();
+                }
```
kezhenxu94, line 529: This looks to be not an elegant solution, maybe we can just simplify this method's signature to `private void writeGeneratedFile(CtClass metricsClass, String type) throws OALCompileException` and use `metricsClass.getSimpleName()` to replace the original argument `className`?

wu-sheng:
Be careful of full class name, it would be an issue for window(too long file path, sometimes show on git clone as well.)

wu-sheng:
As this file is generated sequentially, you could generate `_index` when the name is in conflict. 
In your case, maybe `ServiceDispatcher.class` and `ServiceDispatcher_1.class` are a better choice. After all, this is not a runtime name.

zhyyu:
> As this file is generated sequentially, you could generate `_index` when the name is in conflict. In your case, maybe `ServiceDispatcher.class` and `ServiceDispatcher_1.class` are a better choice. After all, this is not a runtime name.

In my case, in the file content, the class name is `ServiceMeshServiceDispatcher`, but the file name is `ServiceDispatcher.class`. I have used `metricsClass.getSimpleName() `to replace the original argument `className` to fix the problem.

wu-sheng (approved):

Issue #9814, opened 2022-10-19
[Bug] Cluster coordinator should be responsive to instance up/down


kezhenxu94:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar issues.

### Apache SkyWalking Component

OAP server (apache/skywalking)

### What happened

Currently the cluster coordinator is using polling strategy to fetch the OAP instances, and the interval is 5s:

https://github.com/apache/skywalking/blob/34cfafe398e80ca1a7299e1243827937c1a691dd/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/remote/client/RemoteClientManager.java#L96

There is a common case when some of the OAP instances restarted/shutdown, the living ones will still try to connect to those dead instances in the interval, causing error logs like this:

```
2022-10-18 07:33:28,297 - org.apache.skywalking.oap.server.core.remote.client.GRPCRemoteClient -31025 [grpc-default-executor-0] ERROR [] - UNAVAILABLE: io exception
io.grpc.StatusRuntimeException: UNAVAILABLE: io exception
	at io.grpc.Status.asRuntimeException(Status.java:535) ~[grpc-api-1.49.0.jar:1.49.0]
	at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:487) [grpc-stub-1.49.0.jar:1.49.0]
	at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:563) [grpc-core-1.49.0.jar:1.49.0]
	at io.grpc.internal.ClientCallImpl.access$300(ClientCallImpl.java:70) [grpc-core-1.49.0.jar:1.49.0]
	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInternal(ClientCallImpl.java:744) [grpc-core-1.49.0.jar:1.49.0]
	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInContext(ClientCallImpl.java:723) [grpc-core-1.49.0.jar:1.49.0]
	at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37) [grpc-core-1.49.0.jar:1.49.0]
	at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:133) [grpc-core-1.49.0.jar:1.49.0]
[...]

PR #10192, opened 2022-12-22
The cluster coordinator support watch mechanism for notifying `RemoteClientManager` and `ServerStatusService`.

### LINKED ISSUE #10076: [Feature] Support listeners for the `RemoteClients` change.

wankai123:

- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #10076.

also relate to #9814 
I reserve the query timer mechanism and extended the interval, just in case the notification from the third-party client fails due to network or other issues.

### DIFF
--- a/oap-server/server-cluster-plugin/cluster-zookeeper-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/zookeeper/ZookeeperCoordinator.java
+++ b/oap-server/server-cluster-plugin/cluster-zookeeper-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/zookeeper/ZookeeperCoordinator.java
@@ -65,12 +65,12 @@ public class ZookeeperCoordinator extends ClusterCoordinator {
     }
 
     @Override
-    public synchronized void registerRemote(RemoteInstance remoteInstance) throws ServiceRegisterException {
+    public void registerRemote(RemoteInstance remoteInstance) throws ServiceRegisterException {
         try {
             if (needUsingInternalAddr()) {
                 remoteInstance = new RemoteInstance(new Address(config.getInternalComHost(), config.getInternalComPort(), true));
             }
-
+            this.selfAddress = remoteInstance.getAddress();
             ServiceInstance<RemoteInstance> thisInstance = ServiceInstance.<RemoteInstance>builder().name(REMOTE_NAME_PATH)
                                                                                                     .id(UUID.randomUUID()
                                                                                                             .toString())
@@ -84,8 +84,6 @@ public class ZookeeperCoordinator extends ClusterCoordinator {
                                                                                                     .build();
 
             serviceDiscovery.registerService(thisInstance);
-
-            this.selfAddress = remoteInstance.getAddress();
             this.healthChecker.health();
         } catch (Throwable e) {
             this.healthChecker.unHealth(e);
--- a/oap-server/server-cluster-plugin/cluster-zookeeper-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/zookeeper/ZookeeperCoordinator.java
+++ b/oap-server/server-cluster-plugin/cluster-zookeeper-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/zookeeper/ZookeeperCoordinator.java
@@ -67,7 +67,6 @@ public class ZookeeperCoordinator extends ClusterCoordinator {
     @Override
     public synchronized void registerRemote(RemoteInstance remoteInstance) throws ServiceRegisterException {
         try {
-            initHealthChecker();
             if (needUsingInternalAddr()) {
                 remoteInstance = new RemoteInstance(new Address(config.getInternalComHost(), config.getInternalComPort(), true));
             }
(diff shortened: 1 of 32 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleProvider.java
```diff
@@ -355,18 +355,20 @@ public void start() throws ModuleStartException {
 
         Address gRPCServerInstanceAddress = new Address(moduleConfig.getGRPCHost(), moduleConfig.getGRPCPort(), true);
         TelemetryRelatedContext.INSTANCE.setId(gRPCServerInstanceAddress.toString());
+        ClusterCoordinator coordinator = this.getManager()
+                                             .find(ClusterModule.NAME)
+                                             .provider()
+                                             .getService(ClusterCoordinator.class);
+        coordinator.startCoordinator();
+        coordinator.registerWatcher(remoteClientManager);
```
wu-sheng, line 363: Should we do register first before start coodinator?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/cluster/ClusterModule.java
```diff
@@ -32,7 +32,9 @@ public ClusterModule() {
     public Class[] services() {
         return new Class[] {
             ClusterRegister.class,
-            ClusterNodesQuery.class
+            ClusterNodesQuery.class,
+            ClusterWatcherRegister.class,
+            ClusterCoordinator.class
```
wu-sheng, line 37: As `ClusterCoordinator` inherits `ClusterWatcherRegister`, do we need to expose both?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleProvider.java
wankai123: Not necessary, but do register first more clearly. I'll fix it

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/remote/client/RemoteClientManager.java
```diff
@@ -93,14 +94,26 @@ public RemoteClientManager(final ModuleDefineHolder moduleDefineHolder, final in
...
+            this.clusterNodesQuery = moduleDefineHolder.find(ClusterModule.NAME)
+                                                       .provider()
+                                                       .getService(ClusterNodesQuery.class);
+        }
+
+        this.refresh(clusterNodesQuery.queryRemoteNodes());
     }
 
     /**
      * Query OAP server list from the cluster module and create a new connection for the new node. Make the OAP server
```
wu-sheng, line 113: Comments should be updated accordingly.

oap-server/server-cluster-plugin/cluster-zookeeper-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/zookeeper/ZookeeperCoordinator.java
```diff
@@ -128,4 +134,30 @@ private void initHealthChecker() {
             healthChecker = metricCreator.createHealthCheckerGauge("cluster_zookeeper", MetricsTag.EMPTY_KEY, MetricsTag.EMPTY_VALUE);
         }
     }
+
+    @Override
+    protected void start() {
+        initHealthChecker();
+        serviceCache.addListener(new ZookeeperEventListener());
```
kezhenxu94, line 141: Maybe move the line 64 (`this.serviceCache.start();`) here and move this line to the constructor?

oap-server/server-cluster-plugin/cluster-zookeeper-plugin/src/main/java/org/apache/skywalking/oap/server/cluster/plugin/zookeeper/ZookeeperCoordinator.java
```diff
@@ -115,6 +117,10 @@ public List<RemoteInstance> queryRemoteNodes() {
             this.healthChecker.unHealth(e);
             throw new ServiceQueryException(e.getMessage());
         }
+
+        if (log.isDebugEnabled()) {
+            remoteInstances.forEach(instance -> log.debug("Zookeeper cluster instance: {}", instance.toString()));
```
kezhenxu94, line 122:
```suggestion
            remoteInstances.forEach(instance -> log.debug("Zookeeper cluster instance: {}", instance));
```

kezhenxu94 (commented):
Some nits, others LGTM

kezhenxu94 (approved):

wu-sheng (approved):
lgtm. In the future, we may have to decide which are recommended ways to use coordinator mechanism. Query list and watch exist at the same time, could cause confusion.

PR #10195, opened 2022-12-23
Support server status watcher for `MetricsPersistentWorker` to check the metrics whether required initialization.

wankai123:
relate to: https://github.com/apache/skywalking/pull/10085#discussion_r1039067178
and #10076 


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
```diff
@@ -395,6 +397,18 @@ private Metrics requireInitialization(Metrics metrics) {
         return null;
     }
 
+    @Override
+    public void onServerBooted(final BootingStatus bootingStatus) {
+        timeOfLatestStabilitySts = TimeBucket.getMinuteTimeBucket(
```
kezhenxu94, line 402: Please make `timeOfLatestStabilitySts` `volatile`, now it's updated in multiple threads

kezhenxu94: > Please make `timeOfLatestStabilitySts` `volatile`, now it's updated in multiple threads

Or change the type to `AtomicLong` (as well as the one in class `ClusterStatus`)

wu-sheng (approved):

wu-sheng (dismissed):

wu-sheng (approved):

Issue #9883, opened 2022-10-31
[Feature] Support AWS service monitoring


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

In the recent several months, mostly due to @pg-yang @yswdqz hard work in these two iterations. We are now having more and more infrastructure monitoring, e.g. Linux, Kubernetes, Database, Gateway, MQ server(TODO, such as Kafka), Scheduler(TODO, such as Airflow)

Let's begin to do some research for monitoring cloud infra. As we know, this would be long-term work, this issue should be a catalog to list all potential components/services

## List to be completed
- [x] EKS @pg-yang  - https://github.com/apache/skywalking/pull/10199 plan to release in v9.4.0
- [x] Amazon S3 @pg-yang  https://github.com/apache/skywalking/pull/10407 plan to release in v9.4.0
- [x] Amazon DynamoDB @yswdqz  https://github.com/apache/skywalking/pull/10418 plan to release in v9.4.0
- [x] Amazon API Gateway @pg-yang  https://github.com/apache/skywalking/pull/10645  plan to release in v9.5.0

### Use case

We wouldn't trace the cloud infra from the code level, this is unrealistic. But most of good cloud services expose metrics through Prometheus, OpenTelemetry, or some cloud APIs. We could leverage those and integrate these monitoring in SkyWalking.

Making all metrics in one place side by side with their application performance data, and alerting through the same system.

@pg-yang told me, he would work on this after virtual MQ.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

pg-yang:
Yes

pg-yang:
I updated the task list to add EKS and put it first as it's supported directly by https://github.com/aws-observability/aws-otel-collector, which is more clearly and easy, Let's begin with EKS

wu-sheng:
@pg-yang I noticed this, https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/awscloudwatchreceiver/README.md

I think we could use this for reading metrics of S3 and DynamoDB.

pg-yang:
Let's pick up S3 as the second task

wu-sheng:
[...]

PR #10199, opened 2022-12-25
Support monitoring AWS Cloud EKS

pg-yang:

Related issue https://github.com/apache/skywalking/issues/9883
1. UI repo has synced
2. I copied otlp  proto to `java-test-service` for mocking EKS e2e test

## UI screenshot 

image

image

### Node metrics

image

image

### Service metrics 

image


docs/en/setup/backend/backend-aws-eks-monitoring.md
```diff
@@ -0,0 +1,265 @@
+# AWS Cloud EKS monitoring
```
wu-sheng, line 1: @pg-yang I updated the doc a little. Mostly because you put an empty space between the last word and the comma. This is not correct from English grammar perspective.

pg-yang: Thanks, I will be more careful with this mistake.

test/e2e-v2/java-test-service/e2e-mock-sender/src/main/java/org/apache/skywalking/e2e/controller/AwsEKSOtelMetricsSender.java
```diff
@@ -0,0 +1,191 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.apache.skywalking.e2e.E2EConfiguration;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@Slf4j
+@RestController
+@RequestMapping("/aws/otel/eks")
+public class AwsEKSOtelMetricsSender {
```
wu-sheng, line 44: Should we consider moving these mock data into JSON format, and converting from JSON to proto? I know, we have to reset the timestamp again, but the performance doesn't matter in the e2e.

pg-yang: Make sense. It will be a common case API for otlp. Anyone could use this API to mock otlp data by specifying their own JSON data file.

wu-sheng: We may consider a general OTLP mock sender. It could load JSON from different booting JVM properties or system env variables.

oap-server/server-starter/src/main/resources/otel-rules/aws-eks/eks-cluster.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - name: service_count
+    exp: service_number_of_running_pods.sum(['Service','cluster'])
+  - name: net_rx_dropped
+    exp: node_network_rx_dropped.downsampling(SUM)
+  - name: net_rx_error
+    exp: node_network_rx_errors.downsampling(SUM)
+  - name: net_tx_dropped
+    exp: node_network_tx_dropped.downsampling(SUM)
+  - name: net_tx_error
+    exp: node_network_tx_errors.downsampling(SUM)
```
wankai123, line 52: Can you confirm why these metrics use `downsampling(SUM)`, I can see these units are `Count/Second` from the
origin OTEL metrics. `SUM` would plus the value of the metrics each time, should use the default  downsampling 
 `AVG`?

wankai123 (approved):

wankai123 (dismissed):
rollback approve

oap-server/server-starter/src/main/resources/otel-rules/aws-eks/eks-node.yaml
```diff
@@ -0,0 +1,69 @@
...
+  - name: net_rx_bytes
+    exp: node_network_rx_bytes
+  - name: net_rx_error
+    exp: node_network_rx_errors
+  - name: net_tx_bytes
+    exp: node_network_tx_bytes
+  - name: net_tx_error
+    exp: node_network_tx_errors
+  - name: disk_io_write
+    exp: node_diskio_io_service_bytes_write.sum(['device','cluster','NodeName'])
```
kezhenxu94, line 52: Can you confirm that they use  two naming styles in tags like `device` and `NodeName`? From your screenshot the disk-related metrics are all 0’s so I ask

oap-server/server-starter/src/main/resources/otel-rules/aws-eks/eks-cluster.yaml
wankai123: please ignore the approval for now... It's a mistake.

wu-sheng: Sum may be not correct, especially at high dimensiona(hour/day)

oap-server/server-starter/src/main/resources/otel-rules/aws-eks/eks-node.yaml
wu-sheng: 😂😂

test/e2e-v2/java-test-service/e2e-mock-sender/src/main/java/org/apache/skywalking/e2e/controller/AwsEKSOtelMetricsSender.java
pg-yang: I have updated OTLP mock service in `org.apache.skywalking.e2e.controller.OtelMetricsSender`, it will send OTLP data that base on JSON files, one file for one request.

wu-sheng (dismissed):
LGTM. Pending on more confirmations from @kezhenxu94 @wankai123

oap-server/server-starter/src/main/resources/otel-rules/aws-eks/eks-cluster.yaml
wu-sheng: @pg-yang Could you check here? `Network TX Error Count (per second)` is the description, why it could be `downsampling(SUM)`?

pg-yang: Yep, I will confirm these issues by the end of this week at the latest.

wu-sheng: I think, with `SUM` downsampling, it is more likely, we should not need `per second`. Because the value would be error count in minute/hour/day as a SUM.

pg-yang: The unit of `node_network_tx_errors `(original data source) is `Count/Second`, so whether use `SUM` or `AVG`, the result may be incorrect.

wankai123: If we got the metrics values in 1 min for example:
6/s and 10/s
I think the AVG value 8/s is more accuracy than the SUM value 16/s

wu-sheng: It relies on how it is measures. Usually, a thing, if it is calculated count per second, and it is updated with time, then it means there is a window and buffer mechanism in there. We just need the LATEST downsampling policy to save and visualize data.
The key is, is this the case? If there is one error happens in T1, how this metric shows on T1, T1+1s, and T1+1s. 

From what I learned, there is nearly impossible to have count per second always, there has to be a range.

wu-sheng: > If we got the metrics values in 1 min for example:
> 
> 6/s and 10/s
> 
> I think the AVG value 8/s is more accuracy than the SUM value 16/s

If so, latest is the best. Because the latest status matters for APM.

wankai123: > If so, latest is the best. Because the latest status matters for APM.

Agree, I think should to confirm the time window range. If the metric value is calculated from the beginning till current.

wu-sheng: Yes, it is.

docs/en/setup/backend/backend-aws-eks-monitoring.md
```diff
@@ -0,0 +1,265 @@
...
+| Network TX Error Count                | count/s | eks_cluster_node_pod_net_tx_error          | Endpoint   | Network TX error count of the pod that belongs to the service | [AWS Container Insights Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/awscontainerinsightreceiver/README.md) |
+
+### Customizations
+You can customize your own metrics/expression/dashboard panel.
+The metrics definition and expression rules are found in `/config/otel-rules/aws-eks/`.
+The AWS Cloud EKS dashboard panel configurations are found in `/config/ui-initialized-templates/aws_eks`.
+
+### AWS Container Insights Receiver Deploy Example
+
+```yaml
```
kezhenxu94, line 74: This sample yaml is way too long that it doesn't highlight the important part, please remove unrelated content and replace with their official documentation link, an example would be 

```yaml
# ... other resources definitions

# OTEL configuration that specify OAP address("oap-service:11800") as otlp exporter address
apiVersion: v1
kind: ConfigMap
metadata:
  name: otel-agent-conf
  namespace: aws-otel-eks
  labels:
    app: opentelemetry
    component: otel-agent-conf
data:
  otel-agent-config: |
    # ...
    receivers:
      awscontainerinsightreceiver:

    # Make sure to add a job-name attribute
    processors:
      resource/job-name:
        attributes:
        - key: job_name   
          value: aws-cloud-eks-monitoring
          action: insert     

    exporters:
      otlp:
        endpoint: oap-service:11800
        tls:
          insecure: true
      logging:
          loglevel: debug          

    service:
      pipelines:
        metrics:
          receivers: [awscontainerinsightreceiver]
          processors: [resource/job-name]
          exporters: [otlp,logging]

# ... other resources

Comment (kezhenxu94) [reply docs/en/setup/backend/backend-aws-eks-monitoring.md:74]: I meant you should only highlight the important part here and link to the place where users can find the complete yaml, like https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/awscontainerinsightreceiver#sample-configuration-for-container-insights

Comment (kezhenxu94) [reply docs/en/setup/backend/backend-aws-eks-monitoring.md:74]: If you want to give a complete sample, let's create a dedicated yaml file and link to that file, so users can simply use `kubectl apply -f sample.yaml` to set up the sample, without having to copy the large chunk of content.

**Review by kezhenxu94 — DISMISSED** [review]
A nit, otherwise LGTM

File: oap-server/server-starter/src/main/resources/otel-rules/aws-eks/eks-cluster.yaml (hunk 3 above)
Comment (pg-yang) [reply oap-server/server-starter/src/main/resources/otel-rules/aws-eks/eks-cluster.yaml:52]: I couldn't find any documentation that describes the calculation, but I found the source code.
In a word, the calculation logic is `(preValue-currentValue) / (preTimeInSecond-currentTimeInSecond)`.

1. value (`preValue` or `currentValue`) is cumulative count of transmit errors encountered. It's defined in https://github.com/google/cadvisor/blob/master/info/v1/container.go#L434
2. `currentValue`, `currentTimeInSecond` will be stored to cache as `preValue`, `preTimeInSecond` for the next calculation

I put the source code here:
1.  Calculate method: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/951245d4273845de59db3b00f5445bcad9ce4372/internal/aws/metrics/metric_calculator.go#L67
2. The `calculateFunc ` method that is called by `Calculate method`: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/951245d4273845de59db3b00f5445bcad9ce4372/receiver/awscontainerinsightreceiver/internal/cadvisor/extractors/extractor.go#L127

BTW
```
  - name: net_tx_error
    exp: node_network_tx_errors.downsampling(SUM)
```
The above metric is organized from node level to cluster level,so I think `latest ` should be inappropriate.Also,if we use `avg`, the meaning will be net errors per second, per node.

wu-sheng: From your description, I would say avg makes more sense.
Sum for day and hour dimension would be a disaster, right?
We have to keep per node, I am afraid, there is no way we could sum first and then avg, because from data process perspective, there is no difference between two node and two period report of one node, right?

pg-yang: Agree.

docs/en/setup/backend/backend-aws-eks-monitoring.md
wu-sheng: I think a complete OTEL YAML file would be good to host in the doc. Or we could link the file to the incoming file in the showcase.

oap-server/server-starter/src/main/resources/otel-rules/aws-eks/eks-node.yaml
pg-yang: image

I temporarily open the option `as table` to demonstrate the labels are correct.

wu-sheng:
@pg-yang Let's fix the left issues and merge this :)

docs/en/setup/backend/backend-aws-eks-monitoring.md
pg-yang: > I think a complete OTEL YAML file would be good to host in the doc

Agree with it, we don't have showcase for now.

wu-sheng: @pg-yang Are you going to move this to a separate file?

pg-yang: I only provide a OTEL configuration sample in the doc

oap-server/server-starter/src/main/resources/otel-rules/aws-eks/eks-cluster.yaml
pg-yang: image

wu-sheng (approved):
LGTM

wankai123 (approved):
LGTM

PR #10202, opened 2022-12-26
Remove time_bucket tag for BanyanDB impl

### LINKED ISSUE #10196: [BanyanDB][OAP] Remove “time_bucket” from tags
### Search before asking

### Description

The “time_bucket” is transformed into the timestamp in the writing process. We should remove this tag by the `BanyanDBConvert`. 

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

lujiajing1126:
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #10196

1. All usage of `TIME_BUCKET` is removed from query and write.
2. Skip `TIME_BUCKET` registration and writing.

But the read path is different,

### For Stream

It is set to the annotated `@BanyanDB.TimestampColumn` tag.

### For Measure

It is set to the timestamp of the `Measure`

### DIFF
--- a/oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBConverter.java
+++ b/oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBConverter.java
@@ -29,6 +29,9 @@ import org.apache.skywalking.banyandb.v1.client.StreamWrite;
 import org.apache.skywalking.banyandb.v1.client.TagAndValue;
 import org.apache.skywalking.banyandb.v1.client.grpc.exception.BanyanDBException;
 import org.apache.skywalking.banyandb.v1.client.metadata.Serializable;
+import org.apache.skywalking.oap.server.core.analysis.TimeBucket;
+import org.apache.skywalking.oap.server.core.analysis.metrics.Metrics;
+import org.apache.skywalking.oap.server.core.analysis.record.Record;
 import org.apache.skywalking.oap.server.core.storage.type.Convert2Entity;
 import org.apache.skywalking.oap.server.core.storage.type.Convert2Storage;
 import org.apache.skywalking.oap.server.core.storage.type.StorageDataComplexObject;
@@ -48,6 +51,11 @@ public class BanyanDBConverter {
 
         @Override
         public Object get(String fieldName) {
+            if (fieldName.equals(Record.TIME_BUCKET)) {
+                final String timestampColumnName = schema.getTimestampColumn4Stream();
+                long timestampMillis = ((Number) this.get(timestampColumnName)).longValue();
+                return TimeBucket.getTimeBucket(timestampMillis, schema.getMetadata().getDownSampling());
+            }
             MetadataRegistry.ColumnSpec spec = schema.getSpec(fieldName);
             if (double.class.equals(spec.getColumnClass())) {
                 return ByteUtil.bytes2Double(rowEntity.getTagValue(fieldName));
@@ -70,6 +78,9 @@ public class BanyanDBConverter {
 
         @Override
         public void accept(String fieldName, Object fieldValue) {
+            if (fieldName.equals(Record.TIME_BUCKET)) {
+                return;
+            }
             if (fieldName.equals(this.schema.getTimestampColumn4Stream())) {
                 streamWrite.setTimestamp((long) fieldValue);
             }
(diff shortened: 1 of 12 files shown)


oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBConverter.java
```diff
@@ -229,6 +243,9 @@ public StorageToMeasure(MetadataRegistry.Schema schema, DataPoint dataPoint) {
 
         @Override
         public Object get(String fieldName) {
+            if (fieldName.equals(Metrics.TIME_BUCKET)) {
```
wu-sheng, line 246: `Record` and `Metrics` seem to have duplicated **TIME_BUCKET** fields. I think we need a PR to merge these two into `StorageData`, and use it instead of here.

lujiajing1126:
Just one remaining reference in `IProfileTaskQueryDAO`, currently it is possible that `startTimebucket` and `endTimeBucket` are all null @mrproliu @wu-sheng 

image

image

wu-sheng:
> Just one remaining reference in `IProfileTaskQueryDAO`, currently it is possible that `startTimebucket` and `endTimeBucket` are both null @mrproliu @wu-sheng
> 
> image
> 
> image

I think it should not be both null. @mrproliu Please could you confirm the use case?

lujiajing1126:
The latter existence check is contributed by our team member @xzyJavaX, pls check the null `startTimebucket` as well.

mrproliu:
> I think it should not be both null. @mrproliu Please could you confirm the use case?

Only when the UI query the start/end is null, will this cause any problems?

lujiajing1126:
> > I think it should not be both null. @mrproliu Please could you confirm the use case?
> 
> Only when the UI query the start/end is null, will this cause any problems?

We are going to remove `TimeBucket` tag from database and we don't want to scan all time range. It is possible to narrow down the scan time range?

wu-sheng:
We could provide a reasonable range?
The BanyanDB expects a time range for query, I believe we should not let DAO implementation to pick the default range.

mrproliu:
> > > I think it should not be both null. @mrproliu Please could you confirm the use case?
> > 
> > 
> > Only when the UI query the start/end is null, will this cause any problems?
> 
> We are going to remove `TimeBucket` tag from database and we don't want to scan all time range. Is it possible to narrow down the scan time range?

Do we need the start/end time bucket both not empty? or just one?

wu-sheng:
@lujiajing1126 I think when the time range is missing, it is as same as `service_traffic` query. List all available tasks for UI, nothing special. Any tech block?

lujiajing1126:
> @lujiajing1126 I think when the time range is missing, it is as same as `service_traffic` query. List all available tasks for UI, nothing special. Any tech block?

We could use `START_TIME` instead of `TIME_BUCKET`?

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBConverter.java
lujiajing1126: Yes. I think it should be moved to the base class. We may do this later? We have a lot of references to these constants

wu-sheng:
I think the key here is, this query wants all tasks, no filter.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBConverter.java
wu-sheng: Yes, that should be on a separate PR, no matter before or after.

lujiajing1126:
CI pass. PTAL @wu-sheng

wu-sheng:
How is `IProfileTaskQueryDAO` going to be about?

lujiajing1126:
> How is `IProfileTaskQueryDAO` going to be about?

`TIME_BUCKET` and `STAR_TIME` is equivalent since the `start_time` is annotated with `@BanyanDB.TimestampColumn`. So it can be used interchangeably.

wu-sheng (approved):
LGTM

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBConverter.java
wu-sheng: Done through https://github.com/apache/skywalking/pull/10208

PR #10218, opened 2022-12-29
Fix `time_bucket` of `ServiceTraffic` not set correctly in `slowSql`

wu-sheng:

### DIFF
--- a/oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/extractor/ExtractorSpec.java
+++ b/oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/extractor/ExtractorSpec.java
@@ -267,14 +267,8 @@ public class ExtractorSpec extends AbstractSpec {
         DatabaseSlowStatementBuilder builder = new DatabaseSlowStatementBuilder(namingControl);
         builder.setLayer(Layer.nameOf(log.getLayer()));
 
-        long timeBucket = TimeBucket.getTimeBucket(log.getTimestamp(), DownSampling.Minute);
         builder.setServiceName(log.getService());
 
-        ServiceMeta serviceMeta = new ServiceMeta();
-        String serviceName = namingControl.formatServiceName(log.getService());
-        serviceMeta.setName(serviceName);
-        serviceMeta.setLayer(builder.getLayer());
-        serviceMeta.setTimeBucket(builder.getTimeBucket());
         BINDING.get().databaseSlowStatement(builder);
 
         cl.setDelegate(slowSql);
@@ -291,12 +285,14 @@ public class ExtractorSpec extends AbstractSpec {
         builder.setTimeBucket(timeBucketForDB);
         builder.setTimestamp(log.getTimestamp());
 
-        String entityId = serviceMeta.getEntityId();
         builder.prepare();
-        DatabaseSlowStatement databaseSlowStatement = builder.toDatabaseSlowStatement();
-        databaseSlowStatement.setDatabaseServiceId(entityId);
+        sourceReceiver.receive(builder.toDatabaseSlowStatement());
 
-        sourceReceiver.receive(databaseSlowStatement);
+        ServiceMeta serviceMeta = new ServiceMeta();
+        serviceMeta.setName(builder.getServiceName());
+        serviceMeta.setLayer(builder.getLayer());
+        long timeBucket = TimeBucket.getTimeBucket(log.getTimestamp(), DownSampling.Minute);
+        serviceMeta.setTimeBucket(timeBucket);
         sourceReceiver.receive(serviceMeta);
     }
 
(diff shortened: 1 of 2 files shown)


oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/extractor/ExtractorSpec.java
```diff
@@ -291,12 +285,14 @@ public void slowSql(@DelegatesTo(SlowSqlSpec.class) final Closure<?> cl) {
         builder.setTimeBucket(timeBucketForDB);
         builder.setTimestamp(log.getTimestamp());
 
-        String entityId = serviceMeta.getEntityId();
         builder.prepare();
-        DatabaseSlowStatement databaseSlowStatement = builder.toDatabaseSlowStatement();
-        databaseSlowStatement.setDatabaseServiceId(entityId);
```
wu-sheng, line 297: @yswdqz I think `entityId` had been set in `builder.toDatabaseSlowStatement()`. Please confirm why this was required.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/extractor/ExtractorSpec.java
```diff
@@ -267,14 +267,8 @@ public void slowSql(@DelegatesTo(SlowSqlSpec.class) final Closure<?> cl) {
...
         builder.setLayer(Layer.nameOf(log.getLayer()));
 
-        long timeBucket = TimeBucket.getTimeBucket(log.getTimestamp(), DownSampling.Minute);
         builder.setServiceName(log.getService());
 
-        ServiceMeta serviceMeta = new ServiceMeta();
-        String serviceName = namingControl.formatServiceName(log.getService());
-        serviceMeta.setName(serviceName);
-        serviceMeta.setLayer(builder.getLayer());
-        serviceMeta.setTimeBucket(builder.getTimeBucket());
```
wu-sheng, line 277: At here, `builder.getTimeBucket()` is always not initialized. Or the time bucket of ServiceTraffic is always wrong. OAP just fixed this due to this service's traffic also generated through MAL, which provides the correct timestamp.

hanahmily (approved):

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/extractor/ExtractorSpec.java
yswdqz: I think  `builder.getTimeBucket()` should be initialized by MAL such as 
image

wu-sheng: That is `log.getTimestamp()`. Not this timestamp.

wu-sheng: image

Check the code history, you build a timestamp from it, but you never use it. Ths easy fix could be one line

image

I changed others because you did duplicate `naming control` and are unnecessary to set `entity id`.

yswdqz: I found the problem and it was my mistake. Sorry about that.

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/extractor/ExtractorSpec.java
yswdqz: It's right. I was worried that the IDs generated twice would be different before. But it probably doesn't need to be worried about.

yswdqz (approved):

oap-server/analyzer/log-analyzer/src/main/java/org/apache/skywalking/oap/log/analyzer/dsl/spec/extractor/ExtractorSpec.java
wu-sheng: > I found the problem and it was my mistake. Sorry about that.

You don't need to be `sorry`. **Bugs** is a part of engineers' daily life. If SkyWalking needs one to say `sorry`, that would be me. I did the most damage to the project.
So, don't worry. I just want to double-check check I found the right bug, and fix it in a right way.

yswdqz: I get it. Thank you!

PR #10239, opened 2023-01-06
Add Zipkin trace tab in service mesh panel

### LINKED ISSUE #10000: [Feature] [Zipkin UI] Bundle Zipkin Lens UI in our distribution as a widget component

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #10000 .

image

### DIFF
--- a/oap-server/server-starter/src/main/resources/ui-initialized-templates/mesh/mesh-root.json
+++ b/oap-server/server-starter/src/main/resources/ui-initialized-templates/mesh/mesh-root.json
@@ -1,4 +1,4 @@
- /**
+/**
  * Licensed to the Apache Software Foundation (ASF) under one or more
  * contributor license agreements.  See the NOTICE file distributed with
  * this work for additional information regarding copyright ownership.
--- a/oap-server/server-starter/src/main/resources/ui-initialized-templates/mesh/mesh-root.json
+++ b/oap-server/server-starter/src/main/resources/ui-initialized-templates/mesh/mesh-root.json
@@ -1,4 +1,4 @@
-/**
+*
  * Licensed to the Apache Software Foundation (ASF) under one or more
  * contributor license agreements.  See the NOTICE file distributed with
  * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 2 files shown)


oap-server/server-starter/src/main/resources/ui-initialized-templates/mesh/mesh-root.json
```diff
@@ -1,12 +1,12 @@
-/**
+*
```
wu-sheng, line 1: @kezhenxu94 You accidentally changed here.

oap-server/server-starter/src/main/resources/ui-initialized-templates/mesh/mesh-root.json
```diff
@@ -1,12 +1,12 @@
...
  * Licensed to the Apache Software Foundation (ASF) under one or more
  * contributor license agreements.  See the NOTICE file distributed with
  * this work for additional information regarding copyright ownership.
  * The ASF licenses this file to You under the Apache License, Version 2.0
  * (the "License"); you may not use this file except in compliance with
  * the License.  You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *      http: //www.apache.org/licenses/LICENSE-2.0
```
wu-sheng, line 9: This is an unexpected change.

wu-sheng (approved):

PR #10244, opened 2023-01-06
Update ShardingSphere version for compatibility verification

RaigorJiang:
For #9477, Update ShardingSphere to the latest development version for compatibility verification.

Changes:
- Use DistSQL to query table metadata, which is compatible with different database protocols and versions
- Update the configuration format of server.yaml
- The transaction type in e2e is temporarily changed to `LOCAL`, because ShardingSphere's XA transaction manager needs to be improved to adapt to the RDL concurrency scenario.

---


wankai123 (dismissed):
LGTM, thanks!

wu-sheng (approved):

PR #10269, opened 2023-01-12
Add documentation for profiling APIs

mrproliu:


docs/en/api/profiling-protocol.md
```diff
@@ -0,0 +1,277 @@
+# Profiling APIs
+
+SkyWalking offers two types of Profiling, Trace and eBPF, each with its own API.
```
wu-sheng, line 3: Please use concept page naming.
Such as in-process profiling.

https://skywalking.apache.org/docs/main/next/en/concepts-and-designs/profiling/

mrproliu: Done.

wu-sheng (approved):

PR #10282, opened 2023-01-16
Optimize ElasticSearch single trace query performance by custom…

shichaoyuan:
related discussions  https://github.com/apache/skywalking/discussions/10276

- [X] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>.

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -78,6 +78,7 @@
 * Fix TCP service instances are lack of instance properties like `pod` and `namespace`, which causes Pod log not to work for TCP workloads.
 * Add Python HBase happybase module component ID(94).
 * Fix gRPC alarm cannot update settings from dynamic configuration source.
+* [**Breaking Change**] Optimize single trace query performance by customizing routing in ElasticSearch
 
 #### UI
 
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/ElasticSearch.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/ElasticSearch.java
@@ -23,6 +23,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import lombok.Getter;
+import org.apache.skywalking.oap.server.core.analysis.record.Record;
 
 /**
  * ElasticSearch annotation is a holder including all annotations for ElasticSearch storage
@@ -96,5 +97,14 @@ public @interface ElasticSearch {
         @Deprecated
         String columnAlias();
 
+    }
+
+    /**
+     * Routing is to identify which field in {@link Record} is providing the routing for ElasticSearch.
+     */
+    @Target(ElementType.TYPE)
+    @Retention(RetentionPolicy.RUNTIME)
+    @interface Routing {
+        String value();
     }
 }
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ElasticSearchModelExtension.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ElasticSearchModelExtension.java
@@ -13,22 +13,21 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
  */
 
-package org.apache.skywalking.library.elasticsearch.requests;
+package org.apache.skywalking.oap.server.core.storage.model;
 
-import java.util.Map;
-import lombok.Builder;
 import lombok.Getter;
 import lombok.Setter;
+import org.apache.skywalking.oap.server.core.analysis.record.Record;
 
-@Getter
-@Setter
-@Builder
-public final class IndexRequest {
-    private final String index;
-    private final String type;
-    private final String id;
-    private final String routing;
-    private final Map<String, ?> doc;
+public class ElasticSearchModelExtension {
+
+    /**
+     * Routing is to identify which field in {@link Record} is providing the routing for ElasticSearch.
+     */
+    @Getter
+    @Setter
+    private String routing;
 }
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/StorageModels.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/StorageModels.java
@@ -60,6 +60,7 @@ public class StorageModels implements IModelManager, ModelCreator, ModelManipula
         ShardingKeyChecker checker = new ShardingKeyChecker();
         SQLDatabaseModelExtension sqlDBModelExtension = new SQLDatabaseModelExtension();
         BanyanDBModelExtension banyanDBModelExtension = new BanyanDBModelExtension();
+        ElasticSearchModelExtension elasticSearchModelExtension = new ElasticSearchModelExtension();
         retrieval(aClass, storage.getModelName(), modelColumns, scopeId, checker, sqlDBModelExtension, record);
         // Add extra column for additional entities
         if (aClass.isAnnotationPresent(SQLDatabase.ExtraColumn4AdditionalEntity.class)
@@ -97,6 +98,11 @@ public class StorageModels implements IModelManager, ModelCreator, ModelManipula
             banyanDBModelExtension.setStoreIDTag(true);
         }
 
+        if (aClass.isAnnotationPresent(ElasticSearch.Routing.class)) {
+            String routing = aClass.getAnnotation(ElasticSearch.Routing.class).value();
+            elasticSearchModelExtension.setRouting(routing);
+        }
+
         checker.check(storage.getModelName());
 
         Model model = new Model(
@@ -109,7 +115,8 @@ public class StorageModels implements IModelManager, ModelCreator, ModelManipula
             aClass,
             storage.isTimeRelativeID(),
             sqlDBModelExtension,
-            banyanDBModelExtension
+            banyanDBModelExtension,
+            elasticSearchModelExtension
         );
 
         this.followColumnNameRules(model);
--- a/oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/factory/v6/codec/V6IndexRequestSerializer.java
+++ b/oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/factory/v6/codec/V6IndexRequestSerializer.java
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.databind.JsonSerializer;
 import com.fasterxml.jackson.databind.SerializerProvider;
 import java.io.IOException;
 import org.apache.skywalking.library.elasticsearch.requests.IndexRequest;
+import org.apache.skywalking.oap.server.library.util.StringUtil;
 
 final class V6IndexRequestSerializer extends JsonSerializer<IndexRequest> {
     @Override
@@ -38,6 +39,9 @@ final class V6IndexRequestSerializer extends JsonSerializer<IndexRequest> {
                 gen.writeStringField("_index", value.getIndex());
                 gen.writeStringField("_type", value.getType());
                 gen.writeStringField("_id", value.getId());
+                if (StringUtil.isNotBlank(value.getRouting())) {
+                    gen.writeStringField("routing", value.getRouting());
+                }
             }
             gen.writeEndObject();
         }
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/RoutingUtils.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/RoutingUtils.java
@@ -15,20 +15,24 @@
  * limitations under the License.
  */
 
-package org.apache.skywalking.library.elasticsearch.requests;
+package org.apache.skywalking.oap.server.storage.plugin.elasticsearch.base;
+
+import org.apache.skywalking.oap.server.core.storage.model.Model;
+import org.apache.skywalking.oap.server.library.util.StringUtil;
 
 import java.util.Map;
-import lombok.Builder;
-import lombok.Getter;
-import lombok.Setter;
 
-@Getter
-@Setter
-@Builder
-public final class IndexRequest {
-    private final String index;
-    private final String type;
-    private final String id;
-    private final String routing;
-    private final Map<String, ?> doc;
+public class RoutingUtils {
+
+    public static String getRouting(Model model, Map<String, Object> builder) {
+        String routingField = model.getElasticSearchModelExtension().getRouting();
+        if (StringUtil.isBlank(routingField)) {
+            return null;
+        }
+        Object value = builder.get(routingField);
+        if (value == null) {
+            return null;
+        }
+        return value.toString();
+    }
 }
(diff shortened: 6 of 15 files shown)


wankai123 (approved):
LGTM

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/ElasticSearch.java
```diff
@@ -97,4 +98,13 @@ enum AnalyzerType {
         String columnAlias();
 
     }
+
+    /**
+     * Routing is to identify which field in {@link Record} is providing the routing for ElasticSearch.
```
wu-sheng, line 103:
```suggestion
     * Routing defines a filed of {@link Record} to control the sharding policy.
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ElasticSearchModelExtension.java
```diff
@@ -0,0 +1,33 @@
...
+package org.apache.skywalking.oap.server.core.storage.model;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.apache.skywalking.oap.server.core.analysis.record.Record;
+
+public class ElasticSearchModelExtension {
+
+    /**
+     * Routing is to identify which field in {@link Record} is providing the routing for ElasticSearch.
```
wu-sheng, line 28: Please update comments accordingly.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ElasticSearchModelExtension.java
```diff
@@ -0,0 +1,33 @@
...
+     */
+    @Getter
+    @Setter
+    private String routing;
```
wu-sheng, line 32: We should have a setter method to limit this could be set once for a model.

wu-sheng (changes requested):
Mentioned https://github.com/apache/skywalking/discussions/10276#discussioncomment-4693566, more record should be updated. 
- Zipkin Span
- How would we treat log entity? It may have trace-id. What is the policy?

shichaoyuan:
> Mentioned [#10276 (reply in thread)](https://github.com/apache/skywalking/discussions/10276#discussioncomment-4693566), more record should be updated.
> 
> * Zipkin Span
> * How would we treat log entity? It may have trace-id. What is the policy?


The currently supported query scenarios in `LogQueryEsDAO`are rich, not easy to set policy.
If the main usage scenario is to query through traceid, it is appropriate to set routing as traceid.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/zipkin/ZipkinQueryEsDAO.java
```diff
@@ -236,7 +238,9 @@ public List<List<Span>> getTraces(final Set<String> traceIds) {
         BoolQueryBuilder query = Query.bool().must(Query.terms(ZipkinSpanRecord.TRACE_ID, new ArrayList<>(traceIds)));
         SearchBuilder search = Search.builder().query(query).sort(ZipkinSpanRecord.TIMESTAMP_MILLIS, Sort.Order.DESC)
                                      .size(SCROLLING_BATCH_SIZE); //max span size for 1 scroll
-        final SearchParams params = new SearchParams().scroll(SCROLL_CONTEXT_RETENTION);
+        final SearchParams params = new SearchParams()
+                .scroll(SCROLL_CONTEXT_RETENTION)
+                .routing(String.join(",", traceIds));
```
wu-sheng, line 243: @kezhenxu94 Could you check this?

docs/en/changes/changes.md
```diff
@@ -78,6 +78,7 @@
 * Fix TCP service instances are lack of instance properties like `pod` and `namespace`, which causes Pod log not to work for TCP workloads.
 * Add Python HBase happybase module component ID(94).
 * Fix gRPC alarm cannot update settings from dynamic configuration source.
+* [**Breaking Change**] Optimize single trace query performance by customizing routing in ElasticSearch
```
wu-sheng, line 81:
```suggestion
* [**Breaking Change**] Optimize single trace query performance by customizing routing in ElasticSearch. SkyWalking trace segments and Zipkin spans are using trace ID for routing.
```

wu-sheng: So, does this breaking mean, the existing trace segments/spans are not searchable anymore? Due to the routing is changed, it most likely can't hit the records.

shichaoyuan: yes

kezhenxu94, line 81: Correct me if I'm wrong, although you declare this is a breaking change, users can still use the storage from previous version (without any warnings / errors), but they won't be able to search old traces by trace id because the routing field is absent when indexing but the routing field is present when querying from UI, is that the case?

shichaoyuan: yes

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/StorageModels.java
```diff
@@ -98,6 +99,7 @@ public Model add(Class<?> aClass, int scopeId, Storage storage, boolean record)
         }
 
         checker.check(storage.getModelName());
+        elasticSearchModelExtension.setRouting(storage.getModelName(), modelColumns);
```
wu-sheng, line 102:
```suggestion
        // Set routing rules for ElasticSearch
        elasticSearchModelExtension.setRouting(storage.getModelName(), modelColumns);
        
        checker.check(storage.getModelName());
```

The codes should follow existing code style

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/ElasticSearchModelExtension.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ElasticSearchModelExtension {
+
+    /**
+     * Routing defines a field of {@link Record} to control the sharding policy.
+     */
+    @Getter
+    private String routing;
```
wu-sheng, line 34: This may be better as `Optional`. Because mostly this would not have value.

wu-sheng:
> > Mentioned [#10276 (reply in thread)](https://github.com/apache/skywalking/discussions/10276#discussioncomment-4693566), more record should be updated.
> > 
> > * Zipkin Span
> > * How would we treat log entity? It may have trace-id. What is the policy?
> 
> The currently supported query scenarios in `LogQueryEsDAO`are rich, not easy to set policy. If the main usage scenario is to query through traceid, it is appropriate to set routing as traceid.

TraceSegment query has various conditions too. What is the different?

shichaoyuan:
> > > Mentioned [#10276 (reply in thread)](https://github.com/apache/skywalking/discussions/10276#discussioncomment-4693566), more record should be updated.
> > > 
> > > * Zipkin Span
> > > * How would we treat log entity? It may have trace-id. What is the policy?
> > 
> > 
> > The currently supported query scenarios in `LogQueryEsDAO`are rich, not easy to set policy. If the main usage scenario is to query through traceid, it is appropriate to set routing as traceid.
> 
> TraceSegment query has various conditions too. What is the different?

yes.
From a product point of view, trace mainly has two scenarios:
1. scan by service and other filters
2. query by id

But I am not very clear about the scenario of the log

wu-sheng:
The logs would be filtered by services, endpoints, tags, or trace IDs. I feel they are very similar.

@kezhenxu94 WDYT? Should we adopt this in the log part?

kezhenxu94:
> The logs would be filtered by services, endpoints, tags, or trace IDs. I feel they are very similar.
> 
> @kezhenxu94 WDYT? Should we adopt this in the log part?

It's OK but from my understanding, if we use service name as a routing key, the shards might be too unbalanced in some scenarios.

wu-sheng:
> > The logs would be filtered by services, endpoints, tags, or trace IDs. I feel they are very similar.
> > @kezhenxu94 WDYT? Should we adopt this in the log part?
> 
> It's OK but from my understanding, if we use service name as a routing key, the shards might be too unbalanced in some scenarios.

I think we only need to set routing manually when query is slow. If the log query is not slow (yet), we should be OK to leave it for now.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/RoutingUtils.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.skywalking.oap.server.core.storage.model.Model;
+
+import java.util.Map;
+import java.util.Optional;
+
+public class RoutingUtils {
+
+    public static Optional<String> getRouting(final Model model, final Map<String, Object> builder) {
+        Optional<String> routingField = model.getElasticSearchModelExtension().getRouting();
+        return routingField.map(v -> extractRoutingValue(v, builder));
```
wu-sheng, line 29: What does this `map` mean? `Optional` -> `map` -> `Optional` again?

shichaoyuan:
```
if `routingField` is empty:
    just return emtpy
else
    apply process func, either empty or some(value)
```

common patterns of functional programming

wu-sheng: I know what is `map` function. I think what you are trying to process?

shichaoyuan: check if the value of routingFiled is presented.

wu-sheng: OK, I think I am following the logic now.

But this is actually wrong. Check `ElasticSearchConverter#ToStorage#accept`. 

Your change actually works because this only works as the entity don't merge. Trace segment and Zipkin span are not merging with others, but the codes should not be written in this way.

You should ensure `IndexRequest` includes this `routingValue`(Optional). You can't rely on the builder's key as column name as always.

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/IndexRequest.java
```diff
@@ -29,5 +31,7 @@ public final class IndexRequest {
     private final String index;
     private final String type;
     private final String id;
+    @Builder.Default
+    private final Optional<String> routing = Optional.empty();
```
wu-sheng, line 35: This should be the routing value, rather than the routing column name.

Please follow https://github.com/apache/skywalking/pull/10282#discussion_r1071853153, and update this and add comments.

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/IndexRequest.java
```diff
@@ -29,5 +31,10 @@ public final class IndexRequest {
     private final String index;
     private final String type;
     private final String id;
+    /**
+     * the routing value to control the sharding policy
+     */
+    @Builder.Default
+    private final Optional<String> routing = Optional.empty();
```
wu-sheng, line 38:
```suggestion
    /**
     * The routing value of the request.
     */
    @Builder.Default
    private final Optional<String> routing;
```

I think with @Builder.Default, this should have value as always.
https://stackoverflow.com/questions/35739670/what-is-the-default-for-optional-data-members-in-lombok-using-builder

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/factory/v6/codec/V6IndexRequestSerializer.java
```diff
@@ -38,6 +38,9 @@ public void serialize(final IndexRequest value, final JsonGenerator gen,
                 gen.writeStringField("_index", value.getIndex());
                 gen.writeStringField("_type", value.getType());
                 gen.writeStringField("_id", value.getId());
+                if (value.getRouting().isPresent()) {
+                    gen.writeStringField("routing", value.getRouting().get());
```
wu-sheng, line 42: Is the key `routing` or `_routing`? 
Have you tested whether the routing works? 
https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-routing-field.html

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/requests/IndexRequest.java
shichaoyuan: https://projectlombok.org/features/Builder
> If a certain field/parameter is never set during a build session, then it always gets 0 / null / false. If you've put @Builder on a class (and not a method or constructor) you can instead specify the default directly on the field, and annotate the field with @Builder.Default:
@Builder.Default private final long created = System.currentTimeMillis();

as official document, both operations require:
1. specify the default directly on the field
2. and annotate the field with @Builder.Default:

wu-sheng: OK, then please update the comments only.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/zipkin/ZipkinQueryEsDAO.java
wu-sheng: One last thing, should we expose API as Iterable rather than a string?

shichaoyuan: as Iterable, the param logic more cohesive.
I change it

wu-sheng (dismissed):
LGTM. Pending on @kezhenxu94 

FYI @kezhenxu94 We should update demo after merging, and verify this works there.

docs/en/changes/changes.md
kezhenxu94: @wu-sheng read this thread, I believe this would cause user confusions when they upgrade from an older release.

wu-sheng: Yes, I had read so. What do you suggest? Do you suggest we intend to breaking something or somehow raise a warning?
I believe this change is forward positive improvement.

kezhenxu94 (dismissed):
@wu-sheng @shichaoyuan what do you think if we add a configuration to enable/disable the routing field, so users can opt out if they want to upgrade to 9.4.0 but still want to query the old traces, we can enable it by default but leave an option for users to disable it, and consider remove this configuration when we bump up to 10.x in the future and make this permanent enabled.

Others LGTM

wu-sheng:
> @wu-sheng @shichaoyuan what do you think if we add a configuration to enable/disable the routing field, so users can opt out if they want to upgrade to 9.4.0 but still want to query the old traces, we can enable it by default but leave an option for users to disable it, and consider remove this configuration when we bump up to 10.x in the future and make this permanent enabled.
> 
> 
> 
> Others LGTM

Good to me. Let's add one more configuration and update the storage doc for this.
And turn this new config OFF by default, then this could be a feature, rather than breaking change. @shichaoyuan This could be  a general config, enableCustomRouting.

shichaoyuan:
> > @wu-sheng @shichaoyuan what do you think if we add a configuration to enable/disable the routing field, so users can opt out if they want to upgrade to 9.4.0 but still want to query the old traces, we can enable it by default but leave an option for users to disable it, and consider remove this configuration when we bump up to 10.x in the future and make this permanent enabled.
> > Others LGTM
> 
> Good to me. Let's add one more configuration and update the storage doc for this. And turn this new config OFF by default, then this could be a feature, rather than breaking change. @shichaoyuan This could be a general config, enableCustomRouting.

good idea

wu-sheng (approved):
LGTM

kezhenxu94 (approved):

wankai123 (approved):

PR #10287, opened 2023-01-17
[Elasticsearch] Support batchOfBytes configuration to limit size of bulk flush.

chenhaipeng:
### [Elasticsearch] Support batchOfBytes configuration to limit size of bulk flush. 
 Execute the async bulk record data every 5000 requests, sometime the request body is too large, causing the es server to reject it , found in the production environment, so i think the best way is provide byte-sized configuration.


oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -177,9 +185,20 @@ private CompletableFuture<Void> doFlush(final List<Holder> batch) {
         return future;
     }
 
+
+    private byte[] toByteArray (Object obj) throws IOException {
```
wu-sheng, line 189: This seems a code format issue..

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -94,14 +100,16 @@ public CompletableFuture<Void> add(UpdateRequest request) {
     private CompletableFuture<Void> internalAdd(Object request) {
         requireNonNull(request, "request");
         final CompletableFuture<Void> f = new CompletableFuture<>();
+        int len = toByteArray(request).length;
```
wu-sheng, line 103: @kezhenxu94 I have a little concern about the payload of this, which pushed a request to byte[] but only for length.

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/bulk/BulkProcessor.java
chenhaipeng: ok i fix it

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/bulk/BulkProcessor.java
wu-sheng: I am feeling we want this feature, we should somehow work with ElasticSearch client, rather than run serialization twice for a bulk.

wu-sheng: In another word, we should make `sizeOfBytes` a protection mechanism rather than a flush mechanism. Add this into `BulkProcessor#doFlush` to control the final size of the HTTP request.

@chenhaipeng What do you think?

chenhaipeng: i think so , where is it better to control sizeOfBytes？

chenhaipeng: > i think so , where is it better to control sizeOfBytes？

If a protection mechanism is added when sending, then an unpacking may be required, maybe which increase the complexity

wu-sheng: I think `BulkProcessor#doFlush` includes all logic you need. You should have everything you need for. I don't think you need to unpack from there. You only have to do is adding the mechanism to build one request(current) or more.

From my understanding, this mechanism is to resolve too large an HTTP body issue, right?

wu-sheng: Then, changing `BulkProcessor#doFlush` should be good enough and easy enough. Could you try that?

chenhaipeng: I use elasticsearch in the production environment，I solved it simply as above，but this problem may also occur with other storage type

chenhaipeng: > Then, changing `BulkProcessor#doFlush` should be good enough and easy enough. Could you try that?

ok ,i try it

wu-sheng: > I use elasticsearch in the production environment，I solved it simply as above，but this problem may also occur with other storage type

That is fine. We don't need to resolve an issue that doesn't happen(at least in some known scenarios). Let's focus on fixing this on ElasticSearch.

chenhaipeng: > I use elasticsearch in the production environment，I solved it simply as above，but this problem may also occur with other storage type

i fix it ,please check again

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -129,57 +135,79 @@ public void flush() {
...
-
-        final CompletableFuture<Void> flush = doFlush(batch);
-        flush.whenComplete((ignored1, ignored2) -> semaphore.release());
-        flush.join();
-
+        final List<CompletableFuture<Void>> futures = doFlush(batch);
+        for (final CompletableFuture<Void> future : futures) {
+            future.join();
+        }
+        futures.stream().map(future -> future.join());
```
wu-sheng, line 142: Are these duplicated?

wu-sheng:
`configuration-vocabulary.md` is missed to update.

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/bulk/BulkProcessor.java
wu-sheng: > yes , this mechanism is to resolve too large an HTTP body

As you set the default value to 5M, could you share what is the size limitation?

wu-sheng:
> [Elasticsearch] Support batchOfBytes configuration to limit size of bulk flush.

@chenhaipeng I updated the PR title, AFAIK, both record(trace/log) and metrics are using this. Are you aware of this? Your PR title was about `record` only.

chenhaipeng:
> > [Elasticsearch] Support batchOfBytes configuration to limit size of bulk flush.
> 
> @chenhaipeng I updated the PR title, AFAIK, both record(trace/log) and metrics are using this. Are you aware of this? Your PR title was about `record` only.

yes ,both record(trace/log) and metrics are using this!

docs/en/changes/changes.md
```diff
@@ -78,6 +78,8 @@
 * Fix TCP service instances are lack of instance properties like `pod` and `namespace`, which causes Pod log not to work for TCP workloads.
 * Add Python HBase happybase module component ID(94).
 * Fix gRPC alarm cannot update settings from dynamic configuration source.
+* Elasticsearch storage: Support batchOfBytes configuration to set up the batch bytes size of the record data.
```
wu-sheng, line 81:
```suggestion
* Add `batchOfBytes` configuration to limit the size of bulk flush.
```

docs/en/setup/backend/configuration-vocabulary.md
```diff
@@ -106,6 +106,7 @@ The Configuration Vocabulary lists all available configurations provided by `app
...
+| -                       | -              | batchOfBytes                                                                                                                                                             | Async bulk byte size of the record data batch execution，If the batch data is too large, it will be split into the specified size.default is 5M                                                                                                                                                                                                                                                                                                                            | SW_STORAGE_ES_BATCH_OF_BYTES                      | 5242880                                                                                      |
```
wu-sheng, line 109:
```suggestion
| -                       | -              | batchOfBytes                                                                                                                                                             | A threshold to control the max body size of ElasticSearch Bulk flush.                                                                                                                                                                                                                                                                                                                           | SW_STORAGE_ES_BATCH_OF_BYTES                      | 5242880                                                                                      |
```

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java
```diff
@@ -88,6 +88,8 @@ public class StorageModuleElasticsearchConfig extends ModuleConfig {
      * @since 8.7.0 This setting affects all traces/logs/metrics/metadata flush policy.
      */
     private int bulkActions = 5000;
+
+    private int batchOfBytes = 1024 * 1024 * 5;
```
wu-sheng, line 92:
```suggestion
    /**
     * @since 9.4.0 A threshold to control the max body size of ElasticSearch Bulk flush. 
     */
    private int batchOfBytes = 1024 * 1024 * 5;
```

wu-sheng: Please update the `application.yml` comments accordingly.

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/bulk/BulkProcessor.java
chenhaipeng: > > yes , this mechanism is to resolve too large an HTTP body
> 
> As you set the default value to 5M, could you share what is the size limitation?

there is elastic http limit ,Defaults to 100mb, I think it should not exceed 10m. Kafka has a similar limit of 10m.
https://www.elastic.co/guide/en/elasticsearch/reference/7.17/modules-network.html#http-settings

wu-sheng: So, you are facing a 100m limitation? I am feeling 5m is very small, and if there are many tags/logs in the segment/log, we may flush a few data into the storage.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/StorageModuleElasticsearchConfig.java
chenhaipeng: > Please update the `application.yml` comments accordingly.

done

wu-sheng:
One unresolved question, https://github.com/apache/skywalking/pull/10287#discussion_r1072219692

Pending on @kezhenxu94 review, this part is maintained by him.

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -129,57 +135,76 @@ public void flush() {
...
+            Codec codec = es.get().version().get().codec();
+            final List<byte[]> bs = new ArrayList<>();
+            List<CompletableFuture<Void>> futures = new ArrayList<>();
+            List<ByteBuf> byteBufList = new ArrayList<>();
+            for (final Holder holder : batch) {
+                byte[] bytes = codec.encode(holder.request);
+                bs.add(bytes);
+                bs.add("\n".getBytes());
+                bufferOfBytes += bytes.length;
+                if (bufferOfBytes > batchOfBytes) {
```
kezhenxu94, line 160:
```suggestion
                if (bufferOfBytes >= batchOfBytes) {
```

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -129,57 +135,76 @@ public void flush() {
...
+                    final ByteBuf content = Unpooled.wrappedBuffer(bs.toArray(new byte[0][]));
+                    byteBufList.add(content);
+                    bs.clear();
```
kezhenxu94, line 163:
```suggestion
                    bs.clear();
                    bufferOfBytes = 0;
```

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -129,57 +135,76 @@ public void flush() {
...
-        future.whenComplete((ignored, exception) -> {
-            if (exception != null) {
-                batch.stream().map(it -> it.future)
-                     .forEach(it -> it.completeExceptionally(exception));
-                log.error("Failed to execute requests in bulk", exception);
-            } else {
-                log.debug("Succeeded to execute {} requests in bulk", batch.size());
-                batch.stream().map(it -> it.future).forEach(it -> it.complete(null));
+            for (final ByteBuf content : byteBufList) {
+                CompletableFuture future = es.get().version().thenCompose(v -> {
```
kezhenxu94, line 171:
```suggestion
                CompletableFuture<Void> future = es.get().version().thenCompose(v -> {
```

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -129,57 +135,76 @@ public void flush() {
...
         final List<Holder> batch = new ArrayList<>(requests.size());
         requests.drainTo(batch);
-
-        final CompletableFuture<Void> flush = doFlush(batch);
-        flush.whenComplete((ignored1, ignored2) -> semaphore.release());
-        flush.join();
-
+        final List<CompletableFuture<Void>> futures = doFlush(batch);
+        futures.stream().map(future -> future.join());
+        semaphore.release();
```
kezhenxu94, line 140:
```suggestion

        final List<CompletableFuture<Void>> futures = doFlush(batch);
        final CompletableFuture<Void> future = CompletableFuture
            .allOf(futures.toArray(new CompletableFuture[futures.size()]));
        future.whenComplete((v, t) -> semaphore.release());
        future.join();

```

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/bulk/BulkProcessor.java
```diff
@@ -129,57 +135,76 @@ public void flush() {
...
+            int bufferOfBytes = 0;
+            Codec codec = es.get().version().get().codec();
+            final List<byte[]> bs = new ArrayList<>();
+            List<CompletableFuture<Void>> futures = new ArrayList<>();
+            List<ByteBuf> byteBufList = new ArrayList<>();
+            for (final Holder holder : batch) {
+                byte[] bytes = codec.encode(holder.request);
+                bs.add(bytes);
+                bs.add("\n".getBytes());
+                bufferOfBytes += bytes.length;
```
kezhenxu94, line 159: You need to also add the `\n`

```suggestion
                bufferOfBytes += bytes.length + 1;
```

kezhenxu94 (changes requested):

oap-server/server-library/library-elasticsearch-client/src/main/java/org/apache/skywalking/library/elasticsearch/bulk/BulkProcessor.java
chenhaipeng: In a production environment, 5m is about 3000 per dimension. I think in extreme cases, with too many tags, it might cause oversized. It is reasonable to set 10m, https://cloud.tencent.com/document/product/845/56274

wu-sheng: >  5m is about 3000 per dimension

What does this mean?

chenhaipeng: I set the request body to 5mB, about 3000 docs on average

wu-sheng: OK, 3k-10k docs should not be an issue. After all, the bulk mixed metrics, traces and logs. Let's keep the default size of 10m. For today's network performance, a 10m flush should not be a problem.

wu-sheng:
@chenhaipeng Have you been back on this?

chenhaipeng:
> @chenhaipeng Have you been back on this?

yes，i change the default value to 10M

wu-sheng:
> > @chenhaipeng Have you been back on this?
> 
> yes，i change the default value to 10M

How about Zhenxu review?

kezhenxu94:
@chenhaipeng my review comments are not addressed yet :)

chenhaipeng:
> @chenhaipeng my review comments are not addressed yet :)

ok , please code review again

docs/en/setup/backend/configuration-vocabulary.md
```diff
@@ -106,6 +106,7 @@ The Configuration Vocabulary lists all available configurations provided by `app
...
+| -                       | -              | batchOfBytes                                                                                                                                                             | A threshold to control the max body size of ElasticSearch Bulk flush.                                                                                                                                                                                                                                                                                                                           | SW_STORAGE_ES_BATCH_OF_BYTES                      | 5242880                                                                                      |
```
wu-sheng, line 109:
```suggestion
| -                       | -              | batchOfBytes                                                                                                                                                             | A threshold to control the max body size of ElasticSearch Bulk flush.                                                                                                                                                                                                                                                                                                                           | SW_STORAGE_ES_BATCH_OF_BYTES                      | 10485760  (10m)                                                                                    |
```
Do you miss this?

chenhaipeng: ok ，i check again

kezhenxu94 (approved):
LGTM

PR #10296, opened 2023-01-19
Enhance OAP HTTP server to support HTTPS

yswdqz:


oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServer.java
```diff
@@ -70,6 +78,21 @@ public void initialize() {
             })
             .decorator(LoggingService.newDecorator());
 
+        if (config.isEnableTLS()) {
+            sb.https(config.getHttpsPort());
+            if (config.isEnableTlsSelfSigned()) {
+                sb.tlsSelfSigned();
```
kezhenxu94, line 84: It's impossible to use self signed certificate for services that are exposed to external consumer, i.e. aws won't definitely skip verify certificate like you do in the unit tests, so I think this is useless in reality.

wu-sheng:
> I have not found docs about httpserver. Is there a doc I need to edit?

No, there isn't doc for a class.
But there are core and sharing server modules using this, if you are going to expose configurations(currently you seem not), you need to update relative docs.

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServer.java
```diff
@@ -70,6 +78,21 @@ public void initialize() {
...
+            } else {
+                try (InputStream cert =
+                             Files.newInputStream(Paths.get(config.getTlsCertChainPath())
+                                     .toFile().toPath());
+                     InputStream key = PrivateKeyUtil.loadDecryptionKey(config.getTlsKeyPath())) {
+                    sb.tls(cert, key);
```
kezhenxu94, line 90: Https certificates are bound to a domain name so you have to add a configuration item to configure virtual host for that cert.

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServerConfig.java
```diff
@@ -39,4 +39,14 @@ public class HTTPServerConfig {
...
     private int maxRequestHeaderSize = 8192;
+
+    @Builder.Default
+    private boolean enableTLS = false;
+    @Builder.Default
+    private boolean enableTlsSelfSigned = false;
+
+    private int httpsPort;
+    private String tlsKeyPath;
+    private String tlsCertChainPath;
```
kezhenxu94, line 50: All these configurations are domain-name:port specific, i.e. every domain name:port should have the corresponding https port, tlsKeyPath, tlsCertChainPath.

kezhenxu94 (changes requested):
From my experience when users want to enable https they usually need a gateway to configure domain name and the port. Adding this to OAP makes it troublesome, for example, professional gateway can refresh expired cert without reboot but this is not possible in OAP for now as of your implementation.

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServerConfig.java
wu-sheng: One question, today's gRPC seems not setting the domain name?

kezhenxu94: > One question, today's gRPC seems not setting the domain name?

We have no built-in https support in today's gRPC server inside OAP. In another word, users have to setup a gateway to support https for gRPC port 11800.

wu-sheng: image

I mean we have relative/similar configurations in gRPC part.

kezhenxu94: We are actually using the server host here 

https://github.com/apache/skywalking/blob/289427990aa857298e498782ad97a1d8db68ae4a/oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/grpc/GRPCServer.java#L86

After thinking twice I think I was wrong, we don't need to configure the host/domain name, as long as the  CN field in the cert is the same as the hostname that are used to access this endpoint, we are good to go.

kezhenxu94: But still, we need to make it clear, do we support https for all other http server like Prometheus, Zipkin receiver, or do we want to only support that for AWS receiver? If we want to support for all, we need these configurations for each server respectively, otherwise we should move these configurations to that plugin.

wu-sheng: > But still, we need to make it clear, do we support https for all other http server like Prometheus, Zipkin receiver, or do we want to only support that for AWS receiver? If we want to support for all, we need these configurations for each server respectively, otherwise we should move these configurations to that plugin.

For now, we plan to build a new HTTPS receiver inside `aws-firehose-receiver` separate from the existing core and sharing server HTTP server, also separate from Prometheus or Zipkin server.
We don't need to support all in HTTPS, at least not for now.

yswdqz: So I don't need to change configuration item now ?

wu-sheng: We don't need to add a domain name for now. But `enableTlsSelfSigned` seems to need to be removed.

yswdqz:
Is there anything else I should change?

wu-sheng:
@pg-yang Have you tested this?

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServerConfig.java
```diff
@@ -39,4 +39,12 @@ public class HTTPServerConfig {
     private int acceptQueueSize = 0;
     @Builder.Default
     private int maxRequestHeaderSize = 8192;
+
+    @Builder.Default
+    private boolean enableTLS = false;
+
+    private int httpsPort;
```
wu-sheng, line 46: I am wondering are we going to share tls and no tls server together? If not, we could use port, rather than using a new port.
By this, I feel we open an insecure port for http.
It is better we don't expose that.

yswdqz: For now I think use port is enough, because only `aws-receriver` need https, and it only support https.

So I modify now?

wu-sheng: Yes, please.

pg-yang:
> @pg-yang Have you tested this?

I'm using gateway(nginx) to test firehouse https, rather than OAP.
BTW,  @yswdqz  if you need domain,pulibc certificate, I could provide it.

wu-sheng (approved):

kezhenxu94:
> From my experience when users want to enable https they usually need a gateway to configure domain name and the port. Adding this to OAP makes it troublesome, for example, professional gateway can refresh expired cert without reboot but this is not possible in OAP for now as of your implementation.

This is still my main concern, it doesn't look to be a production-ready implementation when you have to restart all the OAP instances when you want to renew the certificate

wu-sheng:
> > From my experience when users want to enable https they usually need a gateway to configure domain name and the port. Adding this to OAP makes it troublesome, for example, professional gateway can refresh expired cert without reboot but this is not possible in OAP for now as of your implementation.
> 
> This is still my main concern, it doesn't look to be a production-ready implementation when you have to restart all the OAP instances when you want to renew the certificate

You are correct. Certification update thing should be considered if you want to take security very seriously. But do we have a good way to do in Java and armeria?
But I think this should not be a block. If users want to use a Gateway(we could document this for firehole-receiver), I am totally fine, and this feature would not block them to do so, right?

In another way, our gRPC service seems to have the same issue(can't update cert in the runtime), right? We would not call that not product-ready.
This reminds me, of when I wrote the security notice, https://skywalking.apache.org/docs/main/next/en/security/readme/. How many people would really put a Gateway before the SkyWalking OAP in the product? AFAIK, very few.

kezhenxu94:
> In another way, our gRPC service seems to have the same issue(can't update cert in the runtime), right? We would not call that not product-ready. 

gRPC can reload the certs in about every 10 seconds.

wu-sheng:
> > In another way, our gRPC service seems to have the same issue(can't update cert in the runtime), right? We would not call that not product-ready.
> 
> gRPC can reload the certs in about every 10 seconds.

How is `armeria`? Could it do that? If so, polish is good. If it still can't, I feel still fine.

kezhenxu94:
> > > In another way, our gRPC service seems to have the same issue(can't update cert in the runtime), right? We would not call that not product-ready.
> > 
> > 
> > gRPC can reload the certs in about every 10 seconds.
> 
> How is `armeria`? Could it do that? If so, polish is good. If it still can't, I feel still fine.

It’s not built in for gRPC, it’s skywalking’s implementation does the reloading work. Armeria doesn’t do that either out of the box either. Feel free to merge

wu-sheng:
> > > > In another way, our gRPC service seems to have the same issue(can't update cert in the runtime), right? We would not call that not product-ready.
> > > 
> > > 
> > > gRPC can reload the certs in about every 10 seconds.
> > 
> > 
> > How is `armeria`? Could it do that? If so, polish is good. If it still can't, I feel still fine.
> 
> It’s not built in for gRPC, it’s skywalking’s implementation does the reloading work. Armeria doesn’t do that out of the box either. Feel free to merge

Got it. Saw `AbstractSslContext`. Yes, it is SkyWalking's implementation.

wu-sheng:
@yswdqz I think we could refer gRPC to do similar in HTTP

> Unfortunately, there is no easy way to update a SSL context at the moment.
One possible workaround is to reconfigure a whole Server  with new certs.
https://github.com/line/armeria/blob/871d87297e4d051241589cb1ae95641cbc83f880/core/src/test/java/com/linecorp/armeria/client/ReconfigurableServerTest.java#L120

This is from Armeria officially. I think we could follow this.

yswdqz:
> @yswdqz I think we could refer gRPC to do similar in HTTP
> 
> > Unfortunately, there is no easy way to update a SSL context at the moment.
> > One possible workaround is to reconfigure a whole Server  with new certs.
> > https://github.com/line/armeria/blob/871d87297e4d051241589cb1ae95641cbc83f880/core/src/test/java/com/linecorp/armeria/client/ReconfigurableServerTest.java#L120
> 
> This is from Armeria officially. I think we could follow this.

Got it.

yswdqz:
@wu-sheng  We also need to be like grpc, reload the certs in every 10 seconds?

wu-sheng:
Like we discussed, it is better to be. Because in real prod env, cert file has expired time, and we should not have to reboot OAP for it, like Zhenxu mentioned.

PR #10302, opened 2023-01-23
Support reload certs in HTTPS server.

yswdqz:

### DIFF
--- a/oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServer.java
+++ b/oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServer.java
@@ -72,7 +72,7 @@ public class HTTPServer implements Server {
     private com.linecorp.armeria.server.Server httpServer;
 
     private ScheduledExecutorService scheduledExecutorService;
-    
+
     public HTTPServer(HTTPServerConfig config) {
         this.config = config;
     }
--- a/oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServer.java
+++ b/oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServer.java
@@ -19,6 +19,7 @@
 package org.apache.skywalking.oap.server.library.server.http;
 
 import com.google.common.collect.Sets;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.linecorp.armeria.common.HttpMethod;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.HttpStatus;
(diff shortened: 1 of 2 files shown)


yswdqz:
I forget add a unit test, I will do it later.

wu-sheng:
I think usually we need a manual test. I don't think tls could be easily tested. 
Wait for @kezhenxu94

kezhenxu94:
> I think usually we need a manual test. I don't think tls could be easily tested. 
> Wait for @kezhenxu94 

To test the reloading mechanism, we can recreate new cert pairs with scripts and update them in test and verify the API still works before and after the update, but I'm totally ok if you don't want to test this case. But we need to add a basic case that verify the https works, the same situation exists in our gRPC agent protocol server in this repo. You can take that as an example.

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServer.java
```diff
@@ -107,15 +149,51 @@ public void addHandler(Object handler, List<HttpMethod> httpMethods) {
...
+            try (InputStream cert = new FileInputStream(tlsCertChainPath.toFile());
+                 InputStream key = PrivateKeyUtil.loadDecryptionKey(tlsKeyPath.toString())) {
+                sb.tls(cert, key);
+            } catch (IOException e) {
+                throw new IllegalArgumentException(e);
+            }
+            normalInitialize(sb);
+            sb.annotatedService()
+              .pathPrefix(config.getContextPath())
+              .build(handlers.toArray());
```
kezhenxu94, line 185: Can you confirm this is needed when we only want to reconfigure the tls cert? I think Armeria will maintain the existing service endpoint if we didn't add new or remove existing service.

kezhenxu94 (changes requested):
According to https://github.com/line/armeria/pull/4637 this PR can't be merged until they release a new version containing the mentioned Armeria PR.

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServer.java
yswdqz: `reconfigure` method will creat a new ServerBuild. So I think it is needed.
image
image

yswdqz: `sb.buildServerConfig(config())` only set the host and port.

[unit test of Armeria](https://github.com/line/armeria/blob/871d87297e4d051241589cb1ae95641cbc83f880/core/src/test/java/com/linecorp/armeria/client/ReconfigurableServerTest.java#L120) can also prove it.

yswdqz:
I have not tested it now. I will test it after they `armeria` release a new version.

kezhenxu94:
Armeria 1.22 is released you can continue this. https://armeria.dev/release-notes/1.22.0/

yswdqz:
Got it , But when I was writing tests for dynamodb, I found that after loading the certificate on skywalking, aws would not successfully send data to the endpoint, while when loading the certificate on nginx and routing it to skywalking, aws could successfully send data to the endpoint. So the existing https server may have some bugs, I will test and finish this PR after the PR about dynamodb is finished

wu-sheng:
By following https://github.com/apache/skywalking/pull/10409, we don't need to support TLS.
Close this too.

PR #10324, opened 2023-01-31
Update core.oal

xiaoziv:
Update core.oal


wu-sheng (approved):
This should be fine but not really meaningful.

PR #10343, opened 2023-02-05
Polish aws-firehose-receiver to adapt existing OTEL proto

pg-yang:

Because AWS only supports [opentelemetry-proto v0.7.0](https://github.com/open-telemetry/opentelemetry-proto/tree/v0.7.0), it's incompatible with [existing opentelemetry-proto in SkyWalking](https://github.com/apache/skywalking/tree/master/oap-server/server-receiver-plugin/receiver-proto/src/main/proto/opentelemetry/proto) as some field names changed.  
To ensure `FirehoseHTTPHandler` works correctly, `aws-firehose-receiver` must parse ExportMetricsServiceRequest through v0.0.7 proto.
So  I copied the metrics part of OTEL v0.7.0 proto to`aws-firehose-receiver`(because aws-firehose-receiver only processes metrics data), and  changed the `java_package` to avoid class conflict
`org.apache.skywalking.oap.server.receiver.aws.firehose.OtelMetricsConvertor` is responsible to convert ExportMetricsServiceRequest of v0.7.0 to  ExportMetricsServiceRequest of [existing opentelemetry-proto in SkyWalking](https://github.com/apache/skywalking/tree/master/oap-server/server-receiver-plugin/receiver-proto/src/main/proto/opentelemetry/proto)


pg-yang:
@yswdqz I recommend adding more test data when you are developing AWS monitoring

oap-server/server-receiver-plugin/aws-firehose-receiver/README.MD
```diff
@@ -0,0 +1,7 @@
+### Notice
+The OTEL proto in this module copy from [opentelemetry-proto v0.7.0](https://github.com/open-telemetry/opentelemetry-proto/tree/v0.7.0) as AWS firehose only support OTEL 0.7, and rewrite follows :
```
kezhenxu94, line 2: > AWS firehose only support OTEL 0.7

What does this mean? Does it mean the https://github.com/apache/skywalking/pull/10300 doesn't work at all or anything else?

pg-yang: OTEL 0.7 has little difference from existing OTEL proto, for example:
```
instrumentationLibraryMetrics (0.7)-> scopeMetrics
doubleSummary, intSummary  (0.7) -> summary
labels (0.7) -> attributes
```

wu-sheng: We should reword, this receiver is using OTEL 0.7 protocol according to AWS doc.

And the copy should be mentioned in the root License file.

pg-yang: Root License has already mentioned `opentelemetry-proto`, I have confirmed 0.7.0 is also `Apache 2.0`

docs/en/setup/backend/aws-firehose-receiver.md
```diff
@@ -18,3 +18,10 @@ CloudWatch metrics with S3 -->  CloudWatch Metric Stream (OpenTelemetry formart)
 
 1. Only OpenTelemetry format is supported (refer to [Metric streams output formats](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-metric-streams-formats.html))
 2. Only HTTPS could be accepted, you could directly enable TLS and set the receiver to listen 443, or put the receiver behind a gateway with HTTPS (refer to [Amazon Kinesis Data Firehose Delivery Stream HTTP Endpoint Delivery Specifications](https://docs.aws.amazon.com/firehose/latest/dev/httpdeliveryrequestresponse.html))
+
+The OTEL proto in this module copy from [opentelemetry-proto v0.7.0](https://github.com/open-telemetry/opentelemetry-proto/tree/v0.7.0) as AWS firehose only support OTEL 0.7, and rewrite follows :
```
wu-sheng, line 22: Update the proto decleared from https://github.com/apache/skywalking/blob/master/LICENSE#L222

docs/en/setup/backend/aws-firehose-receiver.md
```diff
@@ -18,3 +18,10 @@ CloudWatch metrics with S3 -->  CloudWatch Metric Stream (OpenTelemetry formart)
...
+
+1. Only remain metrics part
```
wu-sheng, line 24: What do you mean only metrics? Do you mean tracing and logs excluded?

docs/en/setup/backend/aws-firehose-receiver.md
```diff
@@ -18,3 +18,10 @@ CloudWatch metrics with S3 -->  CloudWatch Metric Stream (OpenTelemetry formart)
...
+2. Change `java_package` to avoid class conflict with existing OTEL proto (oap-server/server-receiver-plugin/receiver-proto)
+
+To reduce change of `otel-receiver-plugin` and reuse MAL, `org.apache.skywalking.oap.server.receiver.aws.firehose.OtelMetricsConvertor` converts OTEL v0.7.0 ExportMetricsServiceRequest to existing OTEL proto (oap-server/server-receiver-plugin/receiver-proto).
```
wu-sheng, line 27:
```suggestion
```

We don't need to mention this(why we wrote codes in this way) in the doc.
If you prefer to be clear, left this in PR description in case of ppl want to know the context.

docs/en/setup/backend/aws-firehose-receiver.md
pg-yang: I only copied metrics part, because aws-firehose-receiver only processes metrics data.
BTW I have updated the description of this PR and will revert the doc later.

docs/en/setup/backend/aws-firehose-receiver.md
wu-sheng: @pg-yang This has to be done. Copy from 3rd party should be decleared.

pg-yang: I have updated the license, but I'm not sure whether it's correct

oap-server/server-receiver-plugin/aws-firehose-receiver/src/main/proto/opentelemetry/proto/common/v1/common.proto
```diff
@@ -0,0 +1,78 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package opentelemetry.proto.common.v1;
```
wu-sheng, line 17: I think we should name package name as v0_7 suffix? This could be more clear. WDYT?

pg-yang: The package is a relative concept, I think it's necessary.

pg-yang: If we change package, the directory have to change as well

wu-sheng: I think it is fine to change. No one has used this.
I want to change because I feel one day, AWS would really pick up v1 or upgrade to a specific version.
Then we could be clear for changing from version to version.

pg-yang: OK, `opentelemetry/v0_7/proto/.....` is better.

image

wu-sheng: Why do we keep v1?

pg-yang: Adding a parent directory is simpler than adding a directory for each module. 
In the future, we could create a  directory named `v1`, and copy the whole OTEL proto files to the directory, rather than change each directory in OTEL

wu-sheng: We don't need to update folder structure, I just make sure Java package including correct version.

wu-sheng (approved):
LGTM. @pg-yang @yswdqz Let's try this.

PR #10349, opened 2023-02-06
 Change `escapedMetricsNameCache` method to support `/`

yswdqz:


yswdqz:
Initially I wanted to convert `/` to `-` but mal doesn't support `-` either.
I also tried to write unit tests, but creating PrometheusMetricConverter object was complicated. So I have tested it locally.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/PrometheusMetricConverter.java
```diff
@@ -71,7 +71,7 @@ public String load(final String name) {
 
     public PrometheusMetricConverter(Rule rule, MeterSystem service) {
         this.convert = new MetricConvert(rule, service);
-        this.metricsNameEscapePattern = Pattern.compile("\\.");
+        this.metricsNameEscapePattern = Pattern.compile("[/.]");
```
wu-sheng, line 74: I am not good at regex, is this testes? Do we have UT for this?

kezhenxu94 (approved):
LGTM

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/prometheus/PrometheusMetricConverter.java
yswdqz: I have tried to write unit tests, but creating PrometheusMetricConverter object was complicated. So I have tested it locally.

Issue #10366, opened 2023-02-09
[Feature][Breaking Change] Enhance SQL-database storage through merging table and managing day-based table rolling.


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Currently, with increasing requirements of using SQL database to support medium-scale deployment, from H2 to MySQL/Postgre, then Sharding MySQL. We face the length limitation for tables and indices to limit our flexibility. 
Here, I want to propose changing the SQL database storage in another way.
1. Merging all tables by referring to the latest ElasticSearch's [`SW_STORAGE_ES_LOGIC_SHARDING=false`](https://skywalking.apache.org/docs/main/next/en/faq/new-elasticsearch-storage-option-explanation-in-9.2.0/)
2. We begin to implement day-based rolling as we had done in ElasticSearch.

### Use case

The primary background of this change is due to the super quickly expanding of SkyWalking's monitoring scope. More metrics are being added for various data sources, such as AWS services(#9883) and Airflow(#10341).
The database eventually would not be able to support the number of tables. Meanwhile, the scale of monitoring targets could be unbalanced, such as the number of VM is much less than the number of pods and services.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #10367, opened 2023-02-09
Refactor `@Column` annotation, swap `Column#name` and `ElasticSearch.Column#columnAlias` and rename `ElasticSearch.Column#columnAlias` to `ElasticSearch.Column#legacyName`

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. This is preparation for https://github.com/apache/skywalking/issues/10366, after that merging tables are the primary case.


wu-sheng:
I think we are better to keep legacy column name as ElasticSearch only?

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/parser/DeepAnalysis.java
```diff
@@ -136,7 +138,10 @@ public AnalysisResult analysis(AnalysisResult result) {
             for (Field field : c.getDeclaredFields()) {
                 Column column = field.getAnnotation(Column.class);
                 if (column != null) {
-                    result.addPersistentField(field.getName(), column.columnName(), field.getType());
+                    result.addPersistentField(
+                        field.getName(),
+                        !Strings.isNullOrEmpty(column.legacyName()) ? column.legacyName() : column.name(),
+                        field.getType());
```
kezhenxu94, line 144: > I think we are better to keep legacy column name as ElasticSearch only?

legacy column is not only used in ElasticSearch storage, it's used in all other storage, moving it to ElasticSearch would make the codes look confusing and hard to understand.

Here is an example, this is in `oal-rt` module, there is no reason we should use `ElasticSearch.Column`, that's really confusing why we should care about storage-specific implementation here.

wu-sheng: I am fine with this change. Meanwhile, legacy column should be  used in ElasticSearch only.
What do I miss?

kezhenxu94: > I am fine with this change. Meanwhile, legacy column should be used in ElasticSearch only.
> What do I miss?

Do we want to break all other storages? I mean do we want to use the new column names (like `datatable_summation` instead of previously `summation`) for all other storages? If true, than the legacy column is only used in ElasticSearch, otherwise, the other storages have to check whether `legacyColumn` is empty, if not use the legacy one so no breaking changes happen.

wu-sheng: There is no other storage, BanyanDB is PoC stage only. All SQL database is going to be broke due to new design proposal.

wu-sheng: We only need to keep consistent for ElasticSearch only.

kezhenxu94: > There is no other storage, BanyanDB is PoC stage only. All SQL database is going to be broke due to new design proposal.

What about influxdb?

kezhenxu94: So you mean we should use the `ElasticSearch.Column` everywhere in the oal-module and maybe other non-storage related modules?

kezhenxu94: Well let me just revise

wu-sheng (approved):

Issue #8551, opened 2022-02-17
[Feature] Migration skywalking from Junit4 to Junit5


GOODBOY008:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

The main motivations are:
- Making full use of Java 8 Lambdas for writing easier to read tests and a
better performing way of composing failure messages.
- Improved test structures with nested and dynamic tests.
- Much better support for parameterized tests to avoid separating
parameterized and non-parameterized parts into different test classes.
- Composable dependencies and better hooks for advanced use cases.
- Better exception verification
- More current infrastructure
- Better parallelizable

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wu-sheng:
If you want to do this, I am fine. But notice, there are several Java repositories, backend and agent are major.

wu-sheng:
But the work will be huge, and this wouldn't be easy to review.

lujiajing1126:
As I understand, if we are using `powermock`, migrating to `Junit5` may not be possible due to the upstream issue https://github.com/powermock/powermock/issues/929.

wu-sheng:
@lujiajing1126 Thank you for updating. And yes, powermock is common in our UTs. I am moving this to discussion for now.

PR #10382, opened 2023-02-12
Migrate tests from junit 4 to junit 5

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #[<issue number>](https://github.com/apache/skywalking/issues/8551).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,10 +151,10 @@
         <maven.compiler.target>1.8</maven.compiler.target>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-        <powermock.version>2.0.7</powermock.version>
+        <powermock.version>2.0.9</powermock.version>
         <checkstyle.version>6.18</checkstyle.version>
-        <junit.version>4.13.1</junit.version>
-        <mockito-core.version>3.5.13</mockito-core.version>
+        <junit.version>5.9.2</junit.version>
+        <mockito-core.version>4.11.0</mockito-core.version>
         <lombok.version>1.18.22</lombok.version>
 
         <!-- core lib dependency -->
@@ -190,27 +190,39 @@
 
         <skipUTs>false</skipUTs>
         <skipITs>true</skipITs>
+        <system-stubs.version>2.0.2</system-stubs.version>
     </properties>
 
     <dependencies>
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-inline</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <scope>test</scope>
+        </dependency>
+
         <dependency>
             <groupId>org.powermock</groupId>
-            <artifactId>powermock-module-junit4</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.powermock</groupId>
-            <artifactId>powermock-api-mockito2</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.objenesis</groupId>
-            <artifactId>objenesis</artifactId>
+            <artifactId>powermock-reflect</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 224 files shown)


wu-sheng:
Is this SkyWalking being able to compile in JDK17?

kezhenxu94:
> Is this SkyWalking being able to compile in JDK17?

Yes

wu-sheng:
Great. Once all CI passed, I am good.

kezhenxu94:
We can compile and test in Java 17, but currently some of the tests will fail in Java 17, as that requires some JVM arguments available only in Java version > 8, like `--add-opens`, this is hard to make it in both Java 8 and Java 17, I'll do in another PR to move the minimum requirement to Java 11 after discussed and voted in mailing list, for now this PR is good to merge.

wu-sheng:
Maven profile could be activated based on JDK version. Does it help your case to inject compiler parameters? Such as through a property value.

```
<profile>
  <id>jdk-versions</id>
  <activation>
    <jdk>(1.4, 8]</jdk>
  </activation>
  ------
</profile> 
```

kezhenxu94:
> Maven profile could be activated based on JDK version. Does it help your case to inject compiler parameters? Such as through a property value.
> 
> ```
> <profile>
>   <id>jdk-versions</id>
>   <activation>
>     <jdk>(1.4, 8]</jdk>
>   </activation>
>   ------
> </profile> 
> ```

I know that, but it is not worth to do this as we don't plan to support both java 8 and java 17.

Using the jdk profile, you have to explicitly specify the other profile such as `-Pdist`, when jdk8 profile is automatically activated, the other profiles are not activated even they have `<activeByDefault>true</activeByDefault>`.

kezhenxu94:
OK I think I can remove some of the original profiles as they are always activated by default, now I can add JDK profile, let's see

wu-sheng:
> I know that, but it is not worth to do this as we don't plan to support both java 8 and java 17, I mean we should have only one recommended development JDK version, right?

If it is not too complex, we could have JDK 8/11/17 all supported. 

> Using the jdk profile, you have to explicitly specify the other profile such as -Pdist, when jdk8 profile is automatically activated, the other profiles are not activated even they have <activeByDefault>true</activeByDefault>.

This is a range for the JDK version, I think we only need one profile for JDK 8 and the other for JDK 9+.
The `<jdk>` is inside `<activation>`, isn't it automatically picked up in different JDK versions? I am guessing, but otherwise, what is the point of this configuration.

kezhenxu94:
> isn't it automatically picked up in different JDK versions? I am guessing, but otherwise, what is the point of this configuration.

It's automatically picked up, but the other profiles are deactivated, even they have `<activeByDefault>true</activeByDefault>`.

wu-sheng:
OK, I see the point now.

pom.xml
```diff
@@ -33,6 +33,10 @@
 
     <modules>
         <module>apm-protocol</module>
+        <module>oap-server</module>
+        <module>oap-server-bom</module>
+        <module>apm-webapp</module>
+        <module>apm-dist</module>
```
wu-sheng, line 39: This is the reason we could close `activeByDefault` for those modules, right?

wu-sheng (approved):
LGTM

pom.xml
kezhenxu94: > This is the reason we could close `activeByDefault` for those modules, right?

Yes, and because these modules are added in the previous default profiles, so I think this is OK we just make it default here

kezhenxu94:
I also remove the `backend`, `ui` profiles, we seldom rely on those two profiles, and if users want to only package backend or UI module, there are a bunch of maven-way to do that, such as `./mvnw -f oap-server/pom.xml package` or `./mvnw -pl :oap-server -am package`, I've also updated the related doc and Makefile.

wu-sheng:
> I also remove the `backend`, `ui` profiles, we seldom rely on those two profiles, and if users want to only package backend or UI module, there are a bunch of maven-way to do that, such as `./mvnw -f oap-server/pom.xml package` or `./mvnw -pl :oap-server -am package`, I've also updated the related doc and Makefile.

I think there is an issue, because dist profile is missed.

kezhenxu94:
> > I also remove the `backend`, `ui` profiles, we seldom rely on those two profiles, and if users want to only package backend or UI module, there are a bunch of maven-way to do that, such as `./mvnw -f oap-server/pom.xml package` or `./mvnw -pl :oap-server -am package`, I've also updated the related doc and Makefile.
> 
> I think there is an issue, because dist profile is missed.

dist profile includes apm-dist module, and apm-dist module is added to the root `<modules>` already, anything else?

wu-sheng:
> > > I also remove the `backend`, `ui` profiles, we seldom rely on those two profiles, and if users want to only package backend or UI module, there are a bunch of maven-way to do that, such as `./mvnw -f oap-server/pom.xml package` or `./mvnw -pl :oap-server -am package`, I've also updated the related doc and Makefile.
> > 
> > 
> > I think there is an issue, because dist profile is missed.
> 
> dist profile includes apm-dist module, and apm-dist module is added to the root `<modules>` already, so I remove `dist` profile, anything else?

The previous `-Pbackend,dist` means using the existing UI and new packaged backend to build dist. Right now, in `./mvnw -f oap-server/pom.xml package` case, how maven could know this? I feel UI would be recompiled again too, as it is on the module list.

kezhenxu94:
> Right now, in ./mvnw -f oap-server/pom.xml package case, how maven could know this? I feel UI would be recompiled again too, as it is on the module list.

Hi, you can translate the command `./mvnw -pl oap-server -pl apm-dist` as:

```xml
    <modules>
        <module>oap-server</module>
        <module>apm-dist</module>
    </modules>
```

UI won't be compiled again

kezhenxu94:
Well, things become complex now, there are profiles inside `apm-dist` module...

kezhenxu94:
I've run out of ideas, just reverted back the changes, currently only Java 8 is supported out of the box.

PR #10383, opened 2023-02-13
Add HTTPX component ID

XinweiLyu:
Adding id for HTTPX plugin in python agent.

 
### DIFF
--- a/oap-server/server-core/src/test/resources/component-libraries.yml
+++ b/oap-server/server-core/src/test/resources/component-libraries.yml
@@ -653,6 +653,12 @@ AsyncPG:
 AIORedis:
   id: 7017
   languages: Python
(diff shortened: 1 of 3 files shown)


oap-server/server-core/src/test/resources/component-libraries.yml
```diff
@@ -653,6 +653,12 @@ AsyncPG:
 AIORedis:
   id: 7017
   languages: Python
+Websockets:
+  id: 7018
+  languages: Python
+HTTPX:
+  id: 7019
+  languages: Python
```
wu-sheng, line 661: You don't need to update the test file, although it is harmless.

wu-sheng (approved):

PR #10387, opened 2023-02-13
Support JDK 11/17 test out of box.

wu-sheng:


wu-sheng:
After this change, I run `./mvnw clean package -Pbackend,dist` both on JDK 8 and 17. Both passed.

PR #10390, opened 2023-02-14
Bump up Java version to 11 and use junit tag to group IT and slow IT

kezhenxu94:

### DIFF
--- a/.github/workflows/skywalking.yaml
+++ b/.github/workflows/skywalking.yaml
@@ -70,7 +70,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: "temurin"
-          java-version: "8"
+          java-version: "11"
           cache: "maven"
       - name: Setup Go
         uses: actions/setup-go@v3
@@ -157,7 +157,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: "temurin"
-          java-version: "8"
+          java-version: "11"
           cache: "maven"
       - name: Build distribution tar
         run: |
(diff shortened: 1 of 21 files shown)


wu-sheng:
So, if there is a test with tag, it won't run by default?

kezhenxu94:
> So, if there is a test with tag, it won't run by default?

Tagging the tests does nothing by default if we don't use that tag, let me elaborate this:

- `./mvnw test ...` by its nature only runs unit tests, whose name is pattern of `*Test`, and it does never runs integration tests, whose name is patter of `IT*` or `*IT`, so in this PR we use `./mvnw clean test ...` to only run unit tests.
- `./mvnw integration-test ...` will run integration tests and unit tests, so we have `skipUTs` to control whether to skip the UT when running ITs, we already had this before.
  - As for ITs, we have two groups, one is normal tests without any `@Tag`s, the other one is slow integration tests, which is annotated with `@Tag("slow")`, so we divided the integration tests into two workflow jobs:
  - `./mvnw -DskipUTs=true clean integration-test -DexcludedGroups=slow ...` run the ITs but don't run the slow ITs and UTs,
  - `./mvnw -DskipUTs=true clean integration-test -Dcheckstyle.skip -Dtest=${{ matrix.test.class }}` run the slow ITs one case at a time, by setting `-Dtest={{ class }}`, not run UTs,

By doing this, we can easily verify all tests, including UTs, ITs, as well as slow ITs, in local machine with a single command `./mvnw verify`, this is helpful when we need to run the all tests locally, just like this PR.

wu-sheng (approved):
LGTM

PR #10401, opened 2023-02-16
Refactor http-based alarm plugins and extract common logic to `HttpAlarmCallback`

kezhenxu94:

### DIFF
--- a/oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/AlarmCore.java
+++ b/oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/AlarmCore.java
@@ -18,18 +18,18 @@
 
 package org.apache.skywalking.oap.server.core.alarm.provider;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
-
 import org.apache.skywalking.oap.server.core.alarm.AlarmCallback;
 import org.apache.skywalking.oap.server.core.alarm.AlarmMessage;
 import org.joda.time.LocalDateTime;
 import org.joda.time.Minutes;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 /**
  * Alarm core includes metrics values in certain time windows based on alarm settings. By using its internal timer
@@ -82,7 +82,9 @@ public class AlarmCore {
                     }
                     List<AlarmMessage> filteredMessages = alarmMessageList.stream().filter(msg -> !msg.isOnlyAsCondition()).collect(Collectors.toList());
                     if (!filteredMessages.isEmpty()) {
-                        allCallbacks.forEach(callback -> callback.doAlarm(filteredMessages));
+                        for (AlarmCallback callback : allCallbacks) {
+                            callback.doAlarm(filteredMessages);
+                        }
                     }
                 }
             } catch (Exception e) {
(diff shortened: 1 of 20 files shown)


oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/AlarmCore.java
```diff
@@ -82,7 +82,9 @@ public void start(List<AlarmCallback> allCallbacks) {
                     }
                     List<AlarmMessage> filteredMessages = alarmMessageList.stream().filter(msg -> !msg.isOnlyAsCondition()).collect(Collectors.toList());
                     if (!filteredMessages.isEmpty()) {
-                        allCallbacks.forEach(callback -> callback.doAlarm(filteredMessages));
+                        for (AlarmCallback callback : allCallbacks) {
+                            callback.doAlarm(filteredMessages);
+                        }
```
wu-sheng, line 87: Why change this?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/alarm/HttpAlarmCallback.java
```diff
@@ -0,0 +1,62 @@
...
+            final URI uri,
+            final String body,
+            final Map<String, String> headers)
+            throws IOException, InterruptedException {
+        final var request = HttpRequest
+                .newBuilder()
+                .uri(uri)
+                .POST(HttpRequest.BodyPublishers.ofString(body))
+                .header("Content-Type", "application/json")
+                .header("Accept", "application/json")
```
wu-sheng, line 42: I think we should expose this. We can't make sure HTTP endpoint using JSON format.

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/AlarmCore.java
kezhenxu94: > Why change this?

To propagate the exception up to the caller, we don't want to catch the exception in every plugin, because there is a place to catch all and print the logs in the alarm core

https://github.com/apache/skywalking/blob/da9fc57804a130d884c8ce0997073f9b32809c71/oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/AlarmCore.java#L80-L93

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/alarm/HttpAlarmCallback.java
kezhenxu94: > I think we should expose this. We can't make sure HTTP endpoint using JSON format.

This is not "make sure", it's a preference here, this means the client (us, OAP) prefers json response, and if the server can provide the response in JSON format, it will send the JSON in high priority, otherwise the server will send whatever format it can. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation

kezhenxu94: you can test this with

```shell
curl -H"Accept: application/xml" https://httpbin.org/json
```

It won't return xml, it will still return json

kezhenxu94:
@wu-sheng mind taking another look?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/alarm/HttpAlarmCallback.java
wu-sheng: I know this is flexible for most time, but I really had seen some notification server has very strict rules to check and parse data.
I just recommend we should have the post method to expose this, and add a postJSON to lock this to be convenient for callbacks.

kezhenxu94: I just remove this header directly... it's never used in our alarm plugins, and it's seldom used in my opinion, also if in the future we need, we can simply pass in the `headers` argument of this method.

wu-sheng (approved):

PR #10407, opened 2023-02-18
Support Amazon Simple Storage Service (AWS S3) metrics monitoring

pg-yang:

Related issue https://github.com/apache/skywalking/issues/9883

### Screenshots

image

image

### DIFF
--- a/docs/en/setup/backend/backend-aws-s3-monitoring.md
+++ b/docs/en/setup/backend/backend-aws-s3-monitoring.md
@@ -1,0 +1,37 @@
+# AWS Cloud S3 monitoring
+Amazon Simple Storage Service (Amazon S3) is an object storage service. SkyWalking leverages [AWS Kinesis Data Firehose receiver](./aws-firehose-receiver.md) to transfer the CloudWatch metrics of s3 to
+[OpenTelemetry receiver](opentelemetry-receiver.md) and into the [Meter System](./../../concepts-and-designs/meter.md).
+
+### Data flow
+1. AWS CloudWatch collect metrics for S3, refer to [S3 monitoring with CloudWatch](https://docs.aws.amazon.com/AmazonS3/latest/userguide/cloudwatch-monitoring.html)
+2. [CloudWatch metric streams](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Metric-Streams.html) stream CloudWatch metrics of S3 to AWS Kinesis Data Firehose
+3. AWS Kinesis Data Firehose delivery metrics to [AWS Kinesis Data Firehose receiver](./aws-firehose-receiver.md) through the HTTP endpoint
+
+### Set up
+1. Create CloudWatch metrics configuration for S3, refer to [S3 metrics configuration](https://docs.aws.amazon.com/AmazonS3/latest/userguide/metrics-configurations.html)
+2. Create an Amazon Kinesis Data Firehose Delivery Stream, and set [AWS Kinesis Data Firehose receiver](./aws-firehose-receiver.md)'s address as HTTP Destination, refer to [Create Delivery Stream](https://docs.aws.amazon.com/firehose/latest/dev/basic-create.html)
+3. Create CloudWatch metric stream, and select the Firehose Delivery Stream which has been created above, set `Select namespaces` to `AWS/S3`, `Select output format` to `OpenTelemetry 0.7`. refer to [CloudWatch Metric Streams](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Metric-Streams.html)
+
+### S3 Monitoring
+
+SkyWalking observes CloudWatch metrics of the S3 bucket, which is cataloged as a `LAYER: AWS_S3` `Service` in the OAP.
+
+#### Supported Metrics
+
+| Monitoring Panel           | Unit  | Metric Name                 | Catalog        | Description                                                                                                                                | Data Source                                                                                                       |
+|----------------------------|-------|-----------------------------|----------------|--------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------|
+| 4xx Errors                 | count | aws_s3_4xx                  | Service        | The number of HTTP 4xx client error status code requests made to the S3 bucket                                                             | [S3 monitoring with CloudWatch](https://docs.aws.amazon.com/AmazonS3/latest/userguide/cloudwatch-monitoring.html) |
+| 5xx Errors                 | count | aws_s3_5xx                  | Service        | The number of HTTP 5xx client error status code requests made to the S3 bucket                                                             | [S3 monitoring with CloudWatch](https://docs.aws.amazon.com/AmazonS3/latest/userguide/cloudwatch-monitoring.html) |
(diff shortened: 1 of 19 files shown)


docs/en/setup/backend/backend-aws-s3-monitoring.md
```diff
@@ -0,0 +1,37 @@
...
+[OpenTelemetry receiver](opentelemetry-receiver.md) and into the [Meter System](./../../concepts-and-designs/meter.md).
+
+### Data flow
+1. AWS CloudWatch collect metrics for S3, refer to [S3 monitoring with CloudWatch](https://docs.aws.amazon.com/AmazonS3/latest/userguide/cloudwatch-monitoring.html)
+2. [CloudWatch metric streams](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Metric-Streams.html) stream CloudWatch metrics of S3 to AWS Kinesis Data Firehose
+3. AWS Kinesis Data Firehose delivery metrics to [AWS Kinesis Data Firehose receiver](./aws-firehose-receiver.md) through the HTTP endpoint
+
+### Set up
+1. Create CloudWatch metrics configuration for S3, refer to [S3 metrics configuration](https://docs.aws.amazon.com/AmazonS3/latest/userguide/metrics-configurations.html)
+2. Create an Amazon Kinesis Data Firehose Delivery Stream, and set [AWS Kinesis Data Firehose receiver](./aws-firehose-receiver.md)'s address as HTTP Destination, refer to [Create Delivery Stream](https://docs.aws.amazon.com/firehose/latest/dev/basic-create.html)
```
wu-sheng, line 12:
```suggestion
2. Create an Amazon Kinesis Data Firehose Delivery Stream, and set [AWS Kinesis Data Firehose receiver](./aws-firehose-receiver.md)'s address as HTTP(s) Destination, refer to [Create Delivery Stream](https://docs.aws.amazon.com/firehose/latest/dev/basic-create.html)
```

docs/en/setup/backend/opentelemetry-receiver.md
```diff
@@ -43,5 +43,6 @@ for identification of the metric data.
 | Metrics of AWS Cloud EKS| otel-rules/aws-eks/eks-cluster.yaml |AWS Container Insights Receiver -> OpenTelemetry Collector -- OC/OTLP exporter --> SkyWalking OAP Server |
 | Metrics of AWS Cloud EKS| otel-rules/aws-eks/eks-service.yaml |AWS Container Insights Receiver -> OpenTelemetry Collector -- OC/OTLP exporter --> SkyWalking OAP Server |
 | Metrics of AWS Cloud EKS| otel-rules/aws-eks/eks-node.yaml |AWS Container Insights Receiver -> OpenTelemetry Collector -- OC/OTLP exporter --> SkyWalking OAP Server |
+| Metrics of AWS Cloud S3 | otel-rules/aws-s3/s3-service.yaml | AWS CloudWatcher Metrics Stream -> AWS Firehose delivery stream -> SkyWalking OAP Server with [AWS Firehose receiver](./aws-firehose-receiver.md) |
```
wu-sheng, line 46: I think we have mentioned this in `firehose receiver`, and should not mention here? At least, they are different protocol until AWS bumping up OTLP version.

pg-yang: Agree.  Actually, I forgot to remove this content

docs/menu.yml
```diff
@@ -189,6 +189,8 @@ catalog:
             catalog:
               - name: "EKS Monitoring"
                 path: "/en/setup/backend/backend-aws-eks-monitoring"
+              - name: "S3 Monitoring"
+                path: "/en/setup/backend/backend-aws-s3-monitoring"
```
wu-sheng, line 193: Just FYI, we plan to add DynamoDB on both AWS and Database menu(UI and docs)

https://github.com/apache/skywalking/pull/10380#discussion_r1103720431

oap-server/server-starter/src/main/resources/otel-rules/aws-s3/s3-service.yaml
```diff
@@ -0,0 +1,55 @@
...
+#    "PT10H"     -- parses as "10 hours" (where an hour is 3600 seconds)
+#    "P2D"       -- parses as "2 days" (where a day is 24 hours or 86400 seconds)
+#    "P2DT3H4M"  -- parses as "2 days, 3 hours and 4 minutes"
+#    "P-6H3M"    -- parses as "-6 hours and +3 minutes"
+#    "-P6H3M"    -- parses as "-6 hours and -3 minutes"
+#    "-P-6H+3M"  -- parses as "+6 hours and -3 minutes"
+# </pre>
+
+filter: "{ tags -> {tags.cloud_provider == 'aws' && tags.Namespace == 'AWS/S3' } }"
+expSuffix: tag({tags -> tags.bucket = 'aws-s3-buckets::' + tags.BucketName}).service(['bucket'], Layer.AWS_S3)
```
wu-sheng, line 33:
```suggestion
expSuffix: tag({tags -> tags.bucket = 'aws-s3::' + tags.BucketName}).service(['bucket'], Layer.AWS_S3)
```

As a group name, how about making it shorter?

oap-server/server-starter/src/main/resources/otel-rules/aws-s3/s3-service.yaml
```diff
@@ -0,0 +1,55 @@
...
+metricPrefix: aws_s3
+metricsRules:
+  - name: 4xx
+    exp: amazonaws_com_AWS_S3_4xxErrors_sum.downsampling(SUM)
```
wu-sheng, line 37: You mentioned this as a count in the document. Are you meaning 4xx requests count in the current time bucket? (minute/hour/day)

wu-sheng:
I think we should polish the dashboard a little.
* Add 4 single values as the first line. 4xx, 5xx, Request Count, Request Latency
* 4xx, 5xx, Request latency, first byte, as line two
* 4 counts for all, get, put and delete as line three
* Download and Uploaded at line four, one for half of the line, also open this new option
image

oap-server/server-starter/src/main/resources/otel-rules/aws-s3/s3-service.yaml
wankai123: From the AWS doc: 
`Sum (reports per period)`, sorry I didn't found the period‘s definition, could you share here?
I think downsampling(SUM) could be a issue.

```
4xxErrors:
The number of HTTP 4xx client error status code requests made to an Amazon S3 bucket with a value of either 0 or 1. The Average statistic shows the error rate, and the Sum statistic shows the count of that type of error, during each period.

Units: Count

Valid statistics: Average (reports per request), Sum (reports per period), Min, Max, Sample Count
```

wu-sheng: I think this is a report period, then, we are doing summary for data in our dimensions.
If it is reported per several seconds, we are still good.

pg-yang: From doc [Monitoring metrics with Amazon CloudWatch](https://docs.aws.amazon.com/pdfs/AmazonS3/latest/userguide/s3-userguide.pdf#cloudwatch-monitoring), `Request Metrics` are available at 1-minute intervals. I think it means the reporting period is 1 minute and I have confirmed it from logs

wu-sheng: I think generally, this should be good.

pg-yang:
Update UI :

image

wu-sheng (approved):
LGTM.

PR #10417, opened 2023-02-20
Sync UI

yswdqz:


docs/en/changes/changes.md
```diff
@@ -130,6 +130,7 @@
 * Add logo for HTTPX.
 * Refactor: optimize the attached events visualization in the trace widget.
 * Update BanyanDB client to 0.3.0.
+* Add AWS DynamoDB menu.
```
wu-sheng, line 133: More change logs when you sync submodule. For this time, from efed817f73fd227244e60efbb3c26b591165d47e to latest.

docs/en/changes/changes.md
```diff
@@ -130,6 +130,9 @@
...
+* Add period.
```
wu-sheng, line 134:
```suggestion
* [Add period.](fix: add auto period to the independent mode for widgets)
```

wu-sheng (approved):

wu-sheng:
SkyWalking eyes seems having a bug when some other PRs break maven cache, and this CI doesn't need to recompile at the same time.

Wait for @kezhenxu94 to fix on skywalking-eyes.

PR #10420, opened 2023-02-20
Add cpm5d aggregation function

### LINKED ISSUE #10419: [Feature] Add a new cpm aggregation function that does not lose precision when downsampled
### Search before asking

### Description

For services with a low rate of requests, for instance 1 per hour, the `cpm` aggregation function loses the requests when it is downsampled to hours or days treating it as effectively 0 requests.
This lose of precision can misleading while being unable to different if there really has been a request or not.

### Use case

Create a new function, for instance `cpm5d` (5 decimals), that doesn't lose precision after downsampling.
We need at least 5 decimals if we want to represent as precise as possible 1 request in a concrete hour of a day, downsampled to days.
The following example is 1 request in a concrete hour or a day. Which means at the end 1 request per day.
1 decimal: 1 / 1440 = 0 // We cannot different anymore
4 decimals: (1 * 10000) / 1440 = 6
5 decimals: (1 * 100000) / 1440 = 69

### Related issues

_No response_

### Code of Conduct

toffentoffen:
`cpm5d` function allows does not lose precision when it is downsampled to hours or days.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #10419.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/CPM5DecimalsMetrics.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/CPM5DecimalsMetrics.java
@@ -25,7 +25,7 @@ public abstract class CPM5DecimalsMetrics extends CPMMetrics implements LongValu
 
     @Override
     public void calculate() {
-        this.setValue((this.getTotal() * 100000) / getDurationInMinute());
+        this.setValue((this.getTotal() * 100_000) / getDurationInMinute());
     }
 }
 
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/CPM5DecimalsMetrics.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/CPM5DecimalsMetrics.java
@@ -1,0 +1,31 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.oap.server.core.analysis.metrics;
+
+import org.apache.skywalking.oap.server.core.analysis.metrics.annotation.MetricsFunction;
+
+@MetricsFunction(functionName = "cpm5d")
+public abstract class CPM5DecimalsMetrics extends CPMMetrics implements LongValueHolder {
+
+    @Override
+    public void calculate() {
+        this.setValue((this.getTotal() * 100000) / getDurationInMinute());
+    }
+}
+
--- a/oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/analysis/metrics/CPM5DecimalsMetricsTest.java
+++ b/oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/analysis/metrics/CPM5DecimalsMetricsTest.java
@@ -1,0 +1,91 @@
+package org.apache.skywalking.oap.server.core.analysis.metrics;
+
+import org.apache.skywalking.oap.server.core.remote.grpc.proto.RemoteData;
+import org.apache.skywalking.oap.server.core.storage.StorageID;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class CPM5DecimalsMetricsTest {
+
+    @Test
+    public void testEntranceCombine() {
+        CPM5DecimalsMetricsTest.CountMetricsImpl impl = new CountMetricsImpl();
+        impl.combine(5);
+        impl.combine(6);
+        impl.combine(7);
+
+        impl.calculate();
+
+        Assertions.assertEquals(1800000, impl.getValue());
+    }
+
+    @Test
+    public void testSelfCombine() {
+        CPM5DecimalsMetricsTest.CountMetricsImpl impl = new CountMetricsImpl();
+        impl.combine(5);
+        impl.combine(6);
+        impl.combine(7);
+
+        CPM5DecimalsMetricsTest.CountMetricsImpl impl2 = new CountMetricsImpl();
+        impl2.combine(5);
+        impl2.combine(6);
+        impl2.combine(7);
+
+        impl.combine(impl2);
+
+        impl.calculate();
+
+        Assertions.assertEquals(3600000, impl.getValue());
+    }
+
+    @Test
+    public void testDownSampleCombine() {
+        CPM5DecimalsMetricsTest.CountMetricsImpl impl = new CountMetricsImpl();
+        impl.setTimeBucket(202302201608L);
+        impl.combine(1);
+
+        impl.calculate();
+
+        Assertions.assertEquals(100000, impl.getValue());
+
+        impl.setTimeBucket(impl.toTimeBucketInHour());
+        impl.calculate();
+
+        impl.setTimeBucket(impl.toTimeBucketInDay());
+        impl.calculate();
+
+        Assertions.assertEquals(69, impl.getValue());
+    }
+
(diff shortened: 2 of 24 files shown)


wu-sheng:
@Fine0830 Could you add `Precision is 5` for this? We need to format `cpm5d` through that.

@toffentoffen We need to update CPM configuration to `CPM5D` in OAL, and settings in general and mesh dashboards, after @Fine0830 's work. Also, please fix your UT.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/CPM5DecimalsMetrics.java
```diff
@@ -0,0 +1,31 @@
...
+package org.apache.skywalking.oap.server.core.analysis.metrics;
+
+import org.apache.skywalking.oap.server.core.analysis.metrics.annotation.MetricsFunction;
+
+@MetricsFunction(functionName = "cpm5d")
+public abstract class CPM5DecimalsMetrics extends CPMMetrics implements LongValueHolder {
+
+    @Override
+    public void calculate() {
+        this.setValue((this.getTotal() * 100000) / getDurationInMinute());
```
wu-sheng, line 28:
```suggestion
        this.setValue((this.getTotal() * 100_000) / getDurationInMinute());
```

oap-server/server-core/src/test/java/org/apache/skywalking/oap/server/core/analysis/metrics/CPM5DecimalsMetricsTest.java
```diff
@@ -0,0 +1,114 @@
...
+package org.apache.skywalking.oap.server.core.analysis.metrics;
+
+import org.apache.skywalking.oap.server.core.remote.grpc.proto.RemoteData;
+import org.apache.skywalking.oap.server.core.storage.StorageID;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class CPM5DecimalsMetricsTest {
+
+    public static final long TIME_BUCKET = 2015_11_07_1130L;
```
toffentoffen, line 28: Does this date look familiar to anyone?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/CPM5DecimalsMetrics.java
toffentoffen: Applied. Thanks for the suggestion.

wu-sheng:
https://github.com/apache/skywalking-booster-ui/pull/239 has merged. Please update the relative dashboards and OAL to adopt this cpm5d function.

toffentoffen:
> [apache/skywalking-booster-ui#239](https://github.com/apache/skywalking-booster-ui/pull/239) has merged. Please update the relative dashboards and OAL to adopt this cpm5d function.

Let's say I have updated all the needed OAL expression to use `cpm5d()` instead of `cmp()`:
For instance `endpoint_cpm` would be:
```
# FROM
# endpoint_cpm = from(Endpoint.*).cpm();
# TO
 endpoint_cpm = from(Endpoint.*).cpm5d(); # easy peasy change XD
```

So now I am trying to update the dashboards. I guess that what I need to do is to change the `calculation`(of the `metricConfig`) of any component in a dashboard that relies on any `*_cpm` metric to use the  `cpm5d`, but I have some doubts about the dashboards. 
Looking at the UI PR, it looks like `cmp5d` is a new calculation https://github.com/apache/skywalking-booster-ui/blob/5cc913a33233d7fb0db151af4e4e444eb3a9a24b/src/hooks/data.ts#L41. 

However, now I am wondering as `cmp5d` is a Calculation, how should I change the following component:
https://github.com/apache/skywalking/blob/1335a48f1c034abc1fe24f6197ee7acfa3118bf0/oap-server/server-starter/src/main/resources/ui-initialized-templates/general/general-service.json#L715-L717
This concrete component is the following endpoints list:
endpoints_list
The first column, is configured as "read all values of the duration"(`readMetricsValues`), and then all the values get averaged. 
But here comes the problem, if I change it to `cpm5d`, the list will be broken.
The think is that we still need calculate the average, but then format it differently. 
So we maybe need to have an `Average5d` calculation, and maybe other `[calculation]5d` calculations, or we have developed the solution in a way that doesn't fit correctly.

wu-sheng:
> But here comes the problem, if I change it to cpm5d, the list will be broken.

Could you check the GraphQL to see what happens? Basically, changing the metric name(keep calculation unchanged) should not break anything.
@Fine0830 Anything different here?

kezhenxu94:
> > But here comes the problem, if I change it to cpm5d, the list will be broken.
> 
> Could you check the GraphQL to see what happens? Basically, changing the metric name(keep calculation unchanged) should not break anything. @Fine0830 Anything different here?

By "broken", I think @toffentoffen means the values are not correct, because currently a calculation "average" is already applied to each cell of the first column, if we choose to `cpm5d`, then the cpm for each service is a list (`readMetricsValues` returns a list).

wu-sheng:
I am not following, the read through GraphQL is reading values, and calculated by UI function.
So, the OAL should not be irrelevant.

wu-sheng:
Do you mean we need a new UI function?

wu-sheng:
@Fine0830 I think we need `cpm5d + AvgPreview` for table list calculation.

Fine0830:
> @Fine0830 I think we need `cpm5d + AvgPreview` for table list calculation.

https://github.com/apache/skywalking/pull/10456 has merged. Please feel free to use `cpm5d + AvgPreview` calculation. @toffentoffen

toffentoffen:
Thank you very much @Fine0830 I think with the new `cmd5dAvg` it will work.
Also, thank you @wu-sheng and @kezhenxu94 for helping here.

wu-sheng:
Marc, have you deployed showcase and verify these pages?

toffentoffen:
Okay I think I understood it now.
So in the general service table for instance:
service_list

For each service and metrics, we fetch all the values because we need them to display the spline graph as well as the table value. But the table value is computed on the client side by averaging all the fetch data points for a given service and metric.
That is the main driving force to have `cpm5dAvg` calculation functions on the UI.

toffentoffen:
> Marc, have you deployed showcase and verify these pages?

I wanted to try it out. Will do it next, will let you know @wu-sheng 
Need to figure out how to change the showcase project and point it to my branch.

wu-sheng:
> > Marc, have you deployed showcase and verify these pages?
> 
> I wanted to try it out. Will do it next, will let you know @wu-sheng Need to figure out how to change the showcase project and point it to my branch.

You just need some images. Everything is here, https://github.com/apache/skywalking-showcase/blob/main/Makefile.in

wu-sheng:
@wankai123 Could you re-check this?

wu-sheng (approved):
LGTM. I noticed there is no `cpm` function doc in the OAL doc. I will update after this gets merged for `cpm` and `cpm5d`.

kezhenxu94:
I have a doubt, now we simply replace the original `cpm` to `cpm5d`, and keep the original metrics name, when users upgrade the OAP but not UI (yet), they will find the related metrics get 100000 times by the real value, this can bring confusion. Also, this is not considered a breaking change, and the metrics data before upgrading and after upgrading will mess up, some of them are original `cpm` and some of them are `x 100000`, the final data make no sense

kezhenxu94 (changes requested):
The doubt should be addressed, thanks.

wu-sheng:
> I have a doubt, now we simply replace the original `cpm` to `cpm5d`, and keep the original metrics name, when users upgrade the OAP but not UI (yet), they will find the related metrics get 100000 times by the real value, this can bring confusion. Also, this is not considered a breaking change, and the metrics data before upgrading and after upgrading will mess up, some of them are original `cpm` and some of them are `x 100000`, the final data make no sense

What do you suggest? Change the metric name in the OAL?

wu-sheng:
BTW, if we change this, our Grafana setup has to follow.

oap-server/server-starter/src/main/resources/oal/core.oal
```diff
@@ -19,15 +19,15 @@
 // Service scope metrics
 service_resp_time = from(Service.latency).longAvg();
 service_sla = from(Service.*).percent(status == true);
-service_cpm = from(Service.*).cpm();
+service_cpm = from(Service.*).cpm5d();
```
wu-sheng, line 22:
```suggestion
service_cpm5d = from(Service.*).cpm5d();
```

Let's follow this to update settings. @wankai123 Once this is merged, Grafana setup should be updated according on showcase.

wu-sheng:
@kezhenxu94 @toffentoffen Another warning, this change(update metric name) could break existing e2e.

wu-sheng:
About the metric name change, `service_cpm` would be about at least cause 66 places changed, including docs and UI setup.

image

wu-sheng:
Close for now. As this is no update for a while, and this will break many metrics from an integration perspective. Let's see whether we really need this.

Issue #10421, opened 2023-02-21
[Feature] Storage-elasticsearch-plugin supports query of selected fields


xzyJavaX:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

In storage-elasticsearch-plugin, query of selected fields is not supported at present, and the entire data must be found every time.Similar functions are already available in other modules,for example:[storage-banyandb-plugin](https://github.com/apache/skywalking/blob/master/oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/measure/BanyanDBMetadataQueryDAO.java#L166-L181)

### Use case

No response

### Related issues

No response

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wu-sheng:
This is a nice-to-have thing. If you want to follow this, welcome.

PR #10423, opened 2023-02-21
Read the name and properties of the instance only in `getInstance` service.

xzyJavaX:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #[10421](https://github.com/apache/skywalking/issues/10421)


wu-sheng:
If you want to add this, please make sure query dao would use this.

docs/en/changes/changes.md
```diff
@@ -104,6 +104,7 @@
 * `Scope` in the Entity of Metrics query v1 protocol is not required and automatical correction. The scope is determined based on the metric itself.
 * Add explicit `ReadTimeout` for ConsulConfigurationWatcher to avoid `IllegalArgumentException: Cache watchInterval=10sec >= networkClientReadTimeout=10000ms`.
 * Fix `DurationUtils.getDurationPoints` exceed, when `startTimeBucket` equals `endTimeBucket`.
+* Storage-elasticsearch-plugin supports query of selected fields.
```
wu-sheng, line 107:
```suggestion
* Read the name and properties of the instance only in `getInstance` service.
```

wu-sheng:
@kezhenxu94 Please take a look.

kezhenxu94 (approved):
Code level changes look good to me, but this may bring some troubles when future contributors use the `getInstance` and need the other fields like instance uuid, or id

wu-sheng:
> Code level changes look good to me, but this may bring some troubles when future contributors use the `getInstance` and need the other fields like instance uuid, or id, they might have no idea why these two fields are null/empty

@xzyJavaX You are better to keep consistent logic in JDBCMetadataQueryDAO`

wu-sheng (changes requested):
I think it is wrong to build only two columns, even for BanyanDB.

@lujiajing1126 It seems BanyanDB implementation doesn't return required fields.

wu-sheng:
Close for now until we have a real use case to use Sources of ElasticSearch accessing.

PR #10432, opened 2023-02-23
[Temp] Exclude gpg mit site from dead link check

wu-sheng:

@kezhenxu94 This link seems dead for days, I just exclude it to make sure other checks working.


Superskyyy:
Yes it's totally dead, I encountered this while adding my gpg key. (I used ubuntu keyserver instead)

Superskyyy (approved):

PR #10448, opened 2023-02-25
Integrate BanyanDB server-side TopN

### LINKED ISSUE #9864: [Feature] [BanyanDB storage] Integrate new TopN and native double/float fields
### Search before asking

### Description

Since TopN calculus is supported by BanyanDB from 0.2, we can make use of this mechanism in OAP core.

And we have to add double/float support in BanyanDB

### Use case

_No response_

### Related issues

_No response_

### Code of Conduct

lujiajing1126:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Partially resolve #9864.

Two upstream related PRs:

- https://github.com/apache/skywalking-banyandb-java-client/pull/37
- https://github.com/apache/skywalking-banyandb/pull/257


oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBAggregationQueryDAO.java
```diff
@@ -51,6 +52,43 @@ public BanyanDBAggregationQueryDAO(BanyanDBStorageClient client) {
...
+        if (schema == null) {
+            throw new IOException("schema is not registered");
+        }
+
+        MetadataRegistry.ColumnSpec spec = schema.getSpec(valueColumnName);
+        if (spec == null) {
+            throw new IOException("field spec is not registered");
+        }
+
+        if (schema.hasTopNAggregation()) {
```
wu-sheng, line 66: In which case, there is no server-side pre-calculated aggregation?

lujiajing1126: > In which case, there is no server-side pre-calculated aggregation?

For those metrics that do not have field.

wu-sheng: Such as? I think metric type has locked those metrics in single value type only.

lujiajing1126: You are right. I was confused. The legacy code is removed

lujiajing1126:
I've tested some cases. It worked,

endpoint_cpm

image

service_cpm

image

Now we have to decide how we can define the parameters? May be in the annotation?

lujiajing1126:
And another issue is to support additional filter in the query. The only property is `serviceID` now?

wu-sheng:
> Now we have to decide how we can define the parameters? May be in the annotation?

I think BanyanDB's specific annotation works. 

> And another issue is to support additional filter in the aggregation query. The only property is serviceID now?

For now, I think only the time range, max candidates for precalculation and service ID.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/BanyanDBModelExtension.java
```diff
@@ -45,4 +46,23 @@ public class BanyanDBModelExtension {
...
 
+    /**
+     * lru_size defines how many snapshots are allowed to be maintained in the memory.
+     * The default value is 2 in the BanyanDB if not set.
+     *
+     * @since 9.4.0
+     */
+    @Getter
+    @Setter
+    private int lruSize = 2;
```
wu-sheng, line 57: Could you explain a little more about lru_size? How does this impact?

wu-sheng: We should have a TopN sub-class. Building this inside `BanyanDBModelExtension` is not very clear. Especially, I think all models have BanyanDBModelExtension, but not all of them have TopN definition.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/MetadataRegistry.java
```diff
@@ -173,24 +172,31 @@ public Measure registerMeasureModel(Model model, BanyanDBStorageConfig config, C
...
+        if (valueColumnOpt.isEmpty()) {
             return null;
         }
-        // TODO: how to configure parameters?
+
+        List<String> groupByTagNames = new ArrayList<>();
+        groupByTagNames.add(Metrics.ENTITY_ID);
+        for (final TagMetadata tagSpec : tagsAndFields.tags) {
+            if (tagSpec.getTagSpec().getTagName().equals(InstanceTraffic.SERVICE_ID)) {
+                groupByTagNames.add(InstanceTraffic.SERVICE_ID);
```
wu-sheng, line 191: How about adding an `Array` in the `TopNAggregation` annotation to define the columns? Then we could verify whether these columns are valid. Only should boot successfully after validation passed.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/BanyanDBModelExtension.java
hanahmily: https://github.com/apache/skywalking-banyandb/blob/ad6a81c726640c8ffb8805adac4f43ffa5b2c423/api/proto/banyandb/database/v1/schema.proto#L138.

It determines how many time_buckets are held. `2` means a data point belonging to the latest 2 time_buckets will be analyzed.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/MetadataRegistry.java
lujiajing1126: If we add an array to `TopNAggregation`, do we need to annotate all Metrics that should have TopN support?

wu-sheng: All single value metrics, I think. We don't support labeled value metrics for topN query.

That is why the manual setup is better than this hidden logic.

BTW, the top-N query support includes two cases, w/ or w/o parent service ID. Usually instance Top-N should have service ID but it is not required from protocol level. Also, service metric has no parent for topN

lujiajing1126: > All single value metrics, I think. We don't support labeled value metrics for topN query.
> 
> That is why the manual setup is better than this hidden logic.
> 
> BTW, the top-N query support includes two cases, w/ or w/o parent service ID. Usually instance Top-N should have service ID but it is not required from protocol level. Also, service metric has no parent for topN

I've added `@TopNAggregation` for

- OAL: dynamically annotate `@TopNAggregation` for single-valued metrics,
- MAL: only annotate `@MeterFunction`s that generate single value

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/model/BanyanDBModelExtension.java
lujiajing1126: Thanks for the explanation. I've reworded a bit for this doc

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/OALRuntime.java
```diff
@@ -315,6 +326,30 @@ private Class generateMetricsClass(AnalysisResult metricsStmt) throws OALCompile
...
         annotationsAttribute.addAnnotation(streamAnnotation);
+
+        // Add @BanyanDB.TopNAggregation if applicable
+        // 1. groupByTagName is not empty
+        // 2. parentClass implements LongValueHolder, IntValueHolder or DoubleValueHolder
+        try {
+            if (Arrays.stream(parentMetricsClass.getInterfaces())
+                    .anyMatch(i -> i.getSimpleName().equals("IntValueHolder") ||
+                            i.getSimpleName().equals("LongValueHolder") ||
+                            i.getSimpleName().equals("DoubleValueHolder"))) {
```
wu-sheng, line 337: I think `isAssignableFrom` is a better way.

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/OALRuntime.java
```diff
@@ -315,6 +326,30 @@ private Class generateMetricsClass(AnalysisResult metricsStmt) throws OALCompile
...
+                    MemberValue[] values = new MemberValue[groupByTagNames.size()];
+                    for (int i = 0; i < groupByTagNames.size(); ++i) {
+                        values[i] = new StringMemberValue(groupByTagNames.get(i), constPool);
+                    }
+                    amv.setValue(values);
+                    topNAnnotation.addMemberValue("groupByTagNames", amv);
+                    annotationsAttribute.addAnnotation(topNAnnotation);
+                }
+            }
+        } catch (NotFoundException ignored) {
```
wu-sheng, line 350: Please don't ignore this important exception.

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/OALRuntime.java
lujiajing1126: For `CtClass` in Javassist, it that possible to use `isAssignableFrom`?

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBAggregationQueryDAO.java
```diff
@@ -85,24 +60,45 @@ protected void apply(MeasureQuery query) {
...
+        TopNQueryResponse resp = null;
+        if (condition.getOrder() == Order.DES) {
+            resp = topN(schema, timestampRange, condition.getTopN(), additionalConditions);
+        } else {
+            resp = bottomN(schema, timestampRange, condition.getTopN(), additionalConditions);
+        }
+
+        if (resp.size() == 0) {
+            return Collections.emptyList();
+        } else if (resp.size() > 1) { // since we have done aggregation, i.e. MEAN
```
wu-sheng, line 76: What do you mean `MEAN` here?

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBAggregationQueryDAO.java
```diff
@@ -85,24 +60,45 @@ protected void apply(MeasureQuery query) {
...
+        if (schema.getTopNSpec() == null) {
+            throw new IOException("TopN spec is registered");
+        }
+
+        TopNQueryResponse resp = null;
+        if (condition.getOrder() == Order.DES) {
+            resp = topN(schema, timestampRange, condition.getTopN(), additionalConditions);
+        } else {
+            resp = bottomN(schema, timestampRange, condition.getTopN(), additionalConditions);
+        }
```
wu-sheng, line 72: `additionalConditions` may contain `service_id` or not. Do we need two TopN annotations for two rules?

wu-sheng: I think w/ and w/o service ID mean different pre-ordered at BanyanDB side for pre-calculation.

wu-sheng: If we need two annotations at the class level, we need another annotation(syntactic sugar of TopN)

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/OALRuntime.java
wu-sheng: Oops, forgot this. Then, use the full class name, such as `LongValueHolder.class.getCanonicalName()`?

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBAggregationQueryDAO.java
lujiajing1126: I think we should always add complete groupByTags to the annotation in order to make filter works.

BanyanDB client can restore those groupBy tags, thus we only need one TopN processor in the database.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBAggregationQueryDAO.java
lujiajing1126: Aggregation by avg

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBAggregationQueryDAO.java
wu-sheng: I mean there are two kinds of topN.
1. Where **service_id = xxx** and time_range in [start, end] Group by entityID
2. Where and time_range in [start, end] Group by entityID

Check the query condition protocol. 
https://github.com/apache/skywalking-query-protocol/blob/450328c6445d4dc50a868a0b797a14755f1d82cc/metrics-v2.graphqls#L74-L75

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBAggregationQueryDAO.java
wu-sheng: We are asking for a TopN(with limit). Why it should be `size==1` only? I am confused.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBAggregationQueryDAO.java
lujiajing1126: > I mean there are two kinds of topN.
> 
> 1. Where **service_id = xxx** and time_range in [start, end] Group by entityID
> 2. Where and time_range in [start, end] Group by entityID
> 
> Check the query condition protocol. https://github.com/apache/skywalking-query-protocol/blob/450328c6445d4dc50a868a0b797a14755f1d82cc/metrics-v2.graphqls#L74-L75

BanyanDB supports post-aggregation as is hinted in the protocol

https://github.com/apache/skywalking-banyandb/blob/main/api/proto/banyandb/measure/v1/topn.proto#L62

I suppose it can serve the first case

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBAggregationQueryDAO.java
lujiajing1126: For a time range query, we will get a series of time windows. For example, when we are querying for the last 30mins, we may get 30 windows provided that the interval of the metric is 1min.

And each time window (1 min) contains a fully ordered list, the maximum number of entries is controlled by the `countersNum` given in the TopNAggregation schema.

So the (final) MEAN aggreation is imposed on all 30 windows, and calculate the average value of the same key (groupByTagValues). After all these operations, we get a single result set.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBAggregationQueryDAO.java
wu-sheng: I think server side should treat topN more complex. Such as, we have conditions(service id) as filter, put pre aggregated list according to service candidate?

Anyway, we need a solution for this. Otherwise, this is a bug. In a distributed system, some services are slower than others.

hanahmily: > > Let's say there are 200 services, each service has 50k endpoints, we need both service-based topN endpoints and global level topN endpoints. How do we make this work? If we only set the rule as top 1000, which are all belonging to several services. This means we can't have topN for other services.
> 
> > 
> 
> > @lujiajing1126 @hanahmily What are your suggestions for this real case?
> 
> 
> 
> Yes. It is possible. Even two TopN(s) or more cannot help in this case.

From the server’s perspective, we can set a global endpoint topn process and several sub-endpoints top with a service id to support this scenario. @lujiajing1126 What’s the gap here?

wu-sheng: How to define that? The service ID is not predictable. We just know service ID should be used as `group by` condition.

wu-sheng: @lujiajing1126 I did a discussion with @hanahmily 
I think the server side is clear, and the thing is you seem mixed query groupBy condition and topN pre-aggregation GroupBy condition.

In the current case, we should have two TopN definition per metric
1. Global TopN 
2. Service-ID oriented groupBy TopN

Notice, the entityID should never be involved pre-aggregation TopN. That ID is the groupBy condition in the query stage only about how to merge data cross bucket. But pre-aggregation is bucket-based.

I mentioned to @hanahmily, in theory <1> could be built through all <2>'s TopN by costing some extra CPU. Considering <1> oriented query is supported but not widely used, we could enhance the server side to build <1> in the query stage only. You could discuss with Gao to see whether we could do this.

wu-sheng: So about the annotation, we could keep  `@BanyanDB.TopNAggregation(groupByTagNames = {"SERVICE_ID"})` for all OAL and MAL function, and let `groupByTagNames` accepts the column not existing. Then we even dont' need to update OALRuntime, but keep the annotation on the function(OAL and MAL) only.
WDYT?

hanahmily:
@lujiajing1126 Would you pls move forward since https://github.com/apache/skywalking-banyandb/pull/257 is merged?

oap-server/oal-rt/src/main/java/org/apache/skywalking/oal/rt/parser/SourceColumn.java
```diff
@@ -34,8 +34,10 @@ public class SourceColumn {
     private int length;
     private String fieldSetter;
     private String fieldGetter;
+    private final boolean groupByColInTopN;
 
-    public SourceColumn(String fieldName, String columnName, Class<?> type, boolean isID, int length) {
+    public SourceColumn(String fieldName, String columnName, Class<?> type, boolean isID, int length,
+                        boolean groupByColInTopN) {
```
wu-sheng, line 40: Is this `Col` meaning `Column`? I think this is duplicated as this class is column. My proposal was `Cond` for `Condition`.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBAggregationQueryDAO.java
```diff
@@ -85,24 +59,41 @@ protected void apply(MeasureQuery query) {
             throw new IOException("field spec is not registered");
         }
 
+        if (schema.getTopNSpec() == null) {
+            throw new IOException("TopN spec is registered");
```
wu-sheng, line 63: @hanahmily You need to test this very carefully. For now, the error could only be thrown when the TopN query is applied. We would not be able to check whether TopN query is set or not.

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBAggregationQueryDAO.java
```diff
@@ -85,24 +59,41 @@ protected void apply(MeasureQuery query) {
...
+
+        if (resp.size() == 0) {
+            return Collections.emptyList();
+        } else if (resp.size() > 1) { // since we have done aggregation, i.e. MEAN
+            throw new IOException("invalid TopN response");
+        }
+
         final List<SelectedRecord> topNList = new ArrayList<>();
-        for (DataPoint dataPoint : resp.getDataPoints()) {
+        for (TopNQueryResponse.Item item : resp.getTopNLists().get(0).getItems()) {
```
wu-sheng, line 80: Could you explain why TopNList is a List(Items) nested inside another List? But meanwhile, TopNLists size is always 1.

lujiajing1126: Answered here https://github.com/apache/skywalking/pull/10448#discussion_r1118953797

wu-sheng:
Based on a discussion with @hanahmily, please open this for `Endpoint` and `BrowserAppPagePerf` sources only. All other sources and MAL functions should skip this as they don't have so many candidates for our use cases.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/BanyanDB.java
```diff
@@ -157,10 +161,26 @@ enum IndexType {
...
+     *
+     * @since 9.4.0
+     */
+    @Target({ElementType.FIELD})
+    @Retention(RetentionPolicy.RUNTIME)
+    @Inherited
+    @interface TopNAggregation {
+        int lruSize() default 2;
+
+        int countersNumber() default 1000;
```
wu-sheng, line 184:
```suggestion
        /**
         * The max size of buffer for the pre-aggregation
         */
        int countersNumber() default 1000;
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/annotation/BanyanDB.java
```diff
@@ -157,10 +161,26 @@ enum IndexType {
...
+     * Generate a TopN Aggregation and use the annotated column as a groupBy tag.
+     * It also contains parameters for TopNAggregation
+     *
+     * @since 9.4.0
+     */
+    @Target({ElementType.FIELD})
+    @Retention(RetentionPolicy.RUNTIME)
+    @Inherited
+    @interface TopNAggregation {
+        int lruSize() default 2;
```
wu-sheng, line 182:
```suggestion
        /**
         * The size of LRU determines the max tolerant time range.
         * The data in [T - lruSize * n, T] would be accepted in the pre-aggregation process.
         * T = the current time in the current dimensionality.
         * n = interval in the current dimensionality.
         */
        int lruSize() default 2;
```

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBAggregationQueryDAO.java
```diff
@@ -51,7 +52,53 @@ public BanyanDBAggregationQueryDAO(BanyanDBStorageClient client) {
...
+        if (schema == null) {
+            throw new IOException("schema is not registered");
+        }
+
+        MetadataRegistry.ColumnSpec spec = schema.getSpec(valueColumnName);
+        if (spec == null) {
+            throw new IOException("field spec is not registered");
+        }
+
+        // BanyanDB server-side TopN support: for metrics aggregated at instance, endpoint levels
```
wu-sheng, line 65:
```suggestion
        // BanyanDB server-side TopN support for metrics pre-aggregation.
```

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBAggregationQueryDAO.java
```diff
@@ -51,7 +52,53 @@ public BanyanDBAggregationQueryDAO(BanyanDBStorageClient client) {
...
+        if (schema.getTopNSpec() != null) {
+            return serverSideTopN(condition, schema, spec, timestampRange, additionalConditions);
```
wu-sheng, line 67: I think we should check the keys(additionalConditions#key) is declared as  `groupByTagNames`.

Otherwise, we may face unexpected query result, right?

lujiajing1126: Fixed

oap-server/server-storage-plugin/storage-banyandb-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/banyandb/BanyanDBAggregationQueryDAO.java
```diff
@@ -64,6 +63,13 @@ public List<SelectedRecord> sortMetrics(TopNCondition condition, String valueCol
 
         // BanyanDB server-side TopN support for metrics pre-aggregation.
         if (schema.getTopNSpec() != null) {
+            // check if additional conditions are registered
+            if (CollectionUtils.isNotEmpty(additionalConditions)) {
+                final Set<String> additionalConditionNames = additionalConditions.stream().map(KeyValue::getKey).collect(Collectors.toSet());
+                if (!ImmutableSet.copyOf(schema.getTopNSpec().getGroupByTagNames()).containsAll(additionalConditionNames)) {
+                    throw new IOException("[" + Strings.join(additionalConditionNames, ',') + "] conditions are not registered as groupBy tags");
```
wu-sheng, line 73: Two questions,
1. I think group-by pre-aggregation should be `equal` or `no condition`, rather than `contains`? I know that we have one condition, so, contains means extractly the same. But I am trying to be precise. 
2. When this pre-aggregation doesn't fit the current query, I think failing back to on-demand group-by still works. We don't have to through exception, right?

lujiajing1126: You're right. Fixed

oap-server-bom/pom.xml
```diff
@@ -72,7 +72,7 @@
         <awaitility.version>3.0.0</awaitility.version>
         <httpcore.version>4.4.13</httpcore.version>
         <commons-compress.version>1.21</commons-compress.version>
-        <banyandb-java-client.version>0.4.0-SNAPSHOT</banyandb-java-client.version>
+        <banyandb-java-client.version>0.4.0-rc0</banyandb-java-client.version>
```
wu-sheng, line 75: Is the rc0 already out on maven central?

lujiajing1126: > Is the rc0 already out on maven central?

Yes. I've released last night.

wu-sheng: Please make sure you sent the mail list update and other process first.

wu-sheng (approved):
LGTM

Issue #10341, opened 2023-02-04
[Feature] Setup dashboard for Airflow monitoring 


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

This is an open issue for new contributors. Apache Airflow is a widely used workflow scheduler. We are encouraging someone new to the community to add a new level catalog(**workflow**) for Airflow.

## Metrics
Airflow exposes metrics through StatsD, https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/logging-monitoring/metrics.html. 
We could use StatsD + OpenTelemetry StatesD(https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/statsdreceiver/README.md) + OpenTelemetry OTEL exporter to ship the metrics to SkyWalking OTEL receiver.
Then use MAL to build metrics as well as a dashboard for those metrics. Notice, a new layer and new UI menu should be added.

## Logging
Airflow supports Fluents to ship metrics, https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/logging-monitoring/logging-architecture.html. SkyWalking already has FluentD setup support, so we should be able to receive and catalog the logs.

Additionally, [Task Logs](https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/logging-monitoring/logging-tasks.html) seems an interesting think. We could use LAL(Log Analysis) to group the logs by task name(or ID) by treating tasks as endpoints(SkyWalking concept).

### Use case

Add more observability for Airflow server. 

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

mufiye:
I have done one issue about the agent and I'm interested in oap. I think I can do this task, could you please assign it to me?

wu-sheng:
Assigned. Good luck.

mufiye:
> Assigned. Good luck.

Ok, I will do my best.

wu-sheng:
@mufiye You could take one step at a time. Make metrics available for airflow first. Then move forward on logs.

mufiye:
> @mufiye You could take one step at a time. Make metrics available for airflow first. Then move forward on logs.

[...]

PR #10449, opened 2023-02-25
Adapt otel exponential histogram data

mufiye:
### <Feature description>
I find the `OpenTelemetryMetricRequestProcessor#adaptMetrics` in otel-receiver-plugin can not process [exponential histogram metric](https://opentelemetry.io/docs/reference/specification/metrics/data-model/#exponentialhistogram). But some service will use this exponentialHistogram instead of histogram metric, such as [statsd receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/statsdreceiver/README.md) in opentelemetry collector. So I do this pull request to adapt this case.
The related discussion about this pull request is in issue #10341.

### DIFF
--- a/oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
+++ b/oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
@@ -163,19 +163,54 @@ public class OpenTelemetryMetricRequestProcessor implements Service {
         return result;
     }
 
+    /**
+     * ExponentialHistogram data points are an alternate representation to the Histogram data point in OpenTelemetry
+     * metric format(https://opentelemetry.io/docs/reference/specification/metrics/data-model/#exponentialhistogram).
+     * It uses scale, offset and bucket index to calculate the bound. Firstly, calculate the base using scale by
+     * formula: base = 2**(2**(-scale)). Then the upperBound of specific bucket can be calculated by formula:
+     * base**(offset+index+1). Above calculation way is about positive buckets. For the negative case, we just
+     * map them by their absolute value into the negative range using the same scale as the positive range. So the
+     * upperBound should be calculated as -base**(offset+index).
+     *
+     * @param positiveOffset       corresponding to positive Buckets' offset in ExponentialHistogramDataPoint
+     * @param positiveBucketCounts corresponding to positive Buckets' bucket_counts in ExponentialHistogramDataPoint
+     * @param negativeOffset       corresponding to negative Buckets' offset in ExponentialHistogramDataPoint
+     * @param negativeBucketCounts corresponding to negative Buckets' bucket_counts in ExponentialHistogramDataPoint
+     * @param scale                corresponding to scale in ExponentialHistogramDataPoint
+     * @return The map is a bucket set for histogram, the key is specific bucket's upperBound, the value is item count
+     * in this bucket lower than or equals to key(upperBound)
+     */
     private static Map<Double, Long> buildBucketsFromExponentialHistogram(
         int positiveOffset, final List<Long> positiveBucketCounts,
         int negativeOffset, final List<Long> negativeBucketCounts, int scale) {
 
         final Map<Double, Long> result = new HashMap<>();
         double base = Math.pow(2.0, Math.pow(2.0, -scale));
+        if (base == Double.POSITIVE_INFINITY) {
+            if (log.isDebugEnabled()) {
+                log.warn("Receive and reject out-of-range ExponentialHistogram data");
+            }
+            return result;
+        }
         double upperBound;
         for (int i = 0; i < negativeBucketCounts.size(); i++) {
             upperBound = -Math.pow(base, negativeOffset + i);
+            if (upperBound == Double.NEGATIVE_INFINITY) {
+                if (log.isDebugEnabled()) {
+                    log.warn("Receive and reject out-of-range ExponentialHistogram data");
+                }
+                break;
+            }
             result.put(upperBound, negativeBucketCounts.get(i));
         }
         for (int i = 0; i < positiveBucketCounts.size() - 1; i++) {
             upperBound = Math.pow(base, positiveOffset + i + 1);
+            if (upperBound == Double.POSITIVE_INFINITY) {
+                if (log.isDebugEnabled()) {
+                    log.warn("Receive and reject out-of-range ExponentialHistogram data");
+                }
+                break;
+            }
             result.put(upperBound, positiveBucketCounts.get(i));
         }
         result.put(Double.POSITIVE_INFINITY, positiveBucketCounts.get(positiveBucketCounts.size() - 1));
--- a/oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
+++ b/oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
@@ -187,18 +187,14 @@ public class OpenTelemetryMetricRequestProcessor implements Service {
         final Map<Double, Long> result = new HashMap<>();
         double base = Math.pow(2.0, Math.pow(2.0, -scale));
         if (base == Double.POSITIVE_INFINITY) {
-            if (log.isWarnEnabled()) {
-                log.warn("Receive and reject out-of-range ExponentialHistogram data");
-            }
+            log.warn("Receive and reject out-of-range ExponentialHistogram data");
             return result;
         }
         double upperBound;
         for (int i = 0; i < negativeBucketCounts.size(); i++) {
             upperBound = -Math.pow(base, negativeOffset + i);
             if (upperBound == Double.NEGATIVE_INFINITY) {
-                if (log.isWarnEnabled()) {
-                    log.warn("Receive and reject out-of-range ExponentialHistogram data");
-                }
+                log.warn("Receive and reject out-of-range ExponentialHistogram data");
                 break;
             }
             result.put(upperBound, negativeBucketCounts.get(i));
@@ -206,9 +202,7 @@ public class OpenTelemetryMetricRequestProcessor implements Service {
         for (int i = 0; i < positiveBucketCounts.size() - 1; i++) {
             upperBound = Math.pow(base, positiveOffset + i + 1);
             if (upperBound == Double.POSITIVE_INFINITY) {
-                if (log.isWarnEnabled()) {
-                    log.warn("Receive and reject out-of-range ExponentialHistogram data");
-                }
+                log.warn("Receive and reject out-of-range ExponentialHistogram data");
                 break;
             }
             result.put(upperBound, positiveBucketCounts.get(i));
--- a/oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
+++ b/oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
@@ -172,8 +172,8 @@ public class OpenTelemetryMetricRequestProcessor implements Service {
      * map them by their absolute value into the negative range using the same scale as the positive range. So the
      * upperBound should be calculated as -base**(offset+index).
      *
-     * We temporarily ignored the zero_count field, because the zero_threshold even could cover over the existing
-     * bucket scopes. The discussion is in pull request #10449.
+     * Ignored the zero_count field temporarily, 
+     * because the zero_threshold even could overlap the existing bucket scopes. 
      *
      * @param positiveOffset       corresponding to positive Buckets' offset in ExponentialHistogramDataPoint
      * @param positiveBucketCounts corresponding to positive Buckets' bucket_counts in ExponentialHistogramDataPoint
--- a/oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
+++ b/oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
@@ -187,7 +187,7 @@ public class OpenTelemetryMetricRequestProcessor implements Service {
         final Map<Double, Long> result = new HashMap<>();
         double base = Math.pow(2.0, Math.pow(2.0, -scale));
         if (base == Double.POSITIVE_INFINITY) {
-            if (log.isDebugEnabled()) {
+            if (log.isWarnEnabled()) {
                 log.warn("Receive and reject out-of-range ExponentialHistogram data");
             }
             return result;
@@ -196,7 +196,7 @@ public class OpenTelemetryMetricRequestProcessor implements Service {
         for (int i = 0; i < negativeBucketCounts.size(); i++) {
             upperBound = -Math.pow(base, negativeOffset + i);
             if (upperBound == Double.NEGATIVE_INFINITY) {
-                if (log.isDebugEnabled()) {
+                if (log.isWarnEnabled()) {
                     log.warn("Receive and reject out-of-range ExponentialHistogram data");
                 }
                 break;
@@ -206,7 +206,7 @@ public class OpenTelemetryMetricRequestProcessor implements Service {
         for (int i = 0; i < positiveBucketCounts.size() - 1; i++) {
             upperBound = Math.pow(base, positiveOffset + i + 1);
             if (upperBound == Double.POSITIVE_INFINITY) {
-                if (log.isDebugEnabled()) {
+                if (log.isWarnEnabled()) {
                     log.warn("Receive and reject out-of-range ExponentialHistogram data");
                 }
                 break;
--- a/oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
+++ b/oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
@@ -24,12 +24,14 @@ import io.opentelemetry.proto.common.v1.KeyValue;
 import io.opentelemetry.proto.metrics.v1.Sum;
 import io.opentelemetry.proto.metrics.v1.SummaryDataPoint;
 import io.vavr.Function1;
+
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Stream;
+
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.skywalking.oap.meter.analyzer.MetricConvert;
@@ -62,13 +64,13 @@ public class OpenTelemetryMetricRequestProcessor implements Service {
     private final OtelMetricReceiverConfig config;
 
     private static final Map<String, String> LABEL_MAPPINGS =
-        ImmutableMap
-            .<String, String>builder()
-            .put("net.host.name", "node_identifier_host_name")
-            .put("host.name", "node_identifier_host_name")
-            .put("job", "job_name")
-            .put("service.name", "job_name")
-            .build();
+            ImmutableMap
+                    .<String, String>builder()
+                    .put("net.host.name", "node_identifier_host_name")
+                    .put("host.name", "node_identifier_host_name")
+                    .put("job", "job_name")
+                    .put("service.name", "job_name")
+                    .build();
     private List<PrometheusMetricConverter> converters;
 
     public void processMetricsRequest(final ExportMetricsServiceRequest requests) {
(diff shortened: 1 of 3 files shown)


wu-sheng:
Please build a test case for the convert, as `otel exponential histogram` is pretty new for many people.

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
```diff
@@ -161,6 +163,24 @@ private static Map<Double, Long> buildBuckets(
         return result;
     }
 
+    private static Map<Double, Long> buildBucketsFromExponentialHistogram(
```
wu-sheng, line 166: Please add detailed and well formatted comments about the converting logic. It is new.

mufiye:
> Please build a test case for the convert, as `otel exponential histogram` is pretty new for many people.

Ok, I get it. I will add the test case into the test dir. I think I need some time to learn how to build my test case because it is different from the plugin-test which I done before.

wu-sheng:
We just need a unit test.

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
```diff
@@ -161,6 +163,24 @@ private static Map<Double, Long> buildBuckets(
...
+        double base = Math.pow(2.0, Math.pow(2.0, -scale));
+        double upperBound;
+        for (int i = 0; i < negativeBucketCounts.size(); i++) {
+            upperBound = -Math.pow(base, negativeOffset + i);
+            result.put(upperBound, negativeBucketCounts.get(i));
+        }
+        for (int i = 0; i < positiveBucketCounts.size(); i++) {
+            upperBound = Math.pow(base, positiveOffset + i + 1);
+            result.put(upperBound, positiveBucketCounts.get(i));
+        }
```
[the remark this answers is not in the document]
mufiye: Ok, I will add it.

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
```diff
@@ -161,6 +163,24 @@ private static Map<Double, Long> buildBuckets(
         return result;
     }
 
+    private static Map<Double, Long> buildBucketsFromExponentialHistogram(
+        int positiveOffset, final List<Long> positiveBucketCounts,
+        int negativeOffset, final List<Long> negativeBucketCounts, int scale) {
+
+        final Map<Double, Long> result = new HashMap<>();
+        double base = Math.pow(2.0, Math.pow(2.0, -scale));
```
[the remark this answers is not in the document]
mufiye: > Please check all power calculations whether they are overflowed and reject the data if so.

Ok, I will add the check.

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
```diff
@@ -187,7 +187,7 @@ private static Map<Double, Long> buildBucketsFromExponentialHistogram(
         final Map<Double, Long> result = new HashMap<>();
         double base = Math.pow(2.0, Math.pow(2.0, -scale));
         if (base == Double.POSITIVE_INFINITY) {
-            if (log.isDebugEnabled()) {
+            if (log.isWarnEnabled()) {
                 log.warn("Receive and reject out-of-range ExponentialHistogram data");
             }
```
wu-sheng, line 192: You don't have extra operation besides the logout, then usually, with or w/o `isWarnEnabled`, they are nearly same.

mufiye: > You don't have extra operation besides the logout, then usually, with or w/o `isWarnEnabled`, they are nearly same.

So should I delete the if judge statement? And for the error, I think it is because I use jdk since 16. I will try to use jdk 11.

wu-sheng: JDK 11 is our minimum JDK requirement. You could set IDE(IntelliJ Language level) to JDK 11 for code APIs.
But JDK16 is not an LTS version, JDK 11 and 17 are a better choice.

wu-sheng: Unless you built more complex logic in the logging, you don't need this for every debug.

wu-sheng:
@mufiye One question, does this histogram always build bucket boundaries as integers?  Could it be a very small double? Because when we calculate percentile from the histogram (usually we do today), the values are bucket boundaries and they should be integer/long only.

mufiye:
> @mufiye One question, does this histogram always build bucket boundaries as integers? Could it be a very small double? 

This histogram always builds bucket boundaries as double. I don't fully understand what you mean.

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
```diff
@@ -161,6 +163,54 @@ private static Map<Double, Long> buildBuckets(
...
+     * @param negativeBucketCounts corresponding to negative Buckets' bucket_counts in ExponentialHistogramDataPoint
+     * @param scale                corresponding to scale in ExponentialHistogramDataPoint
+     * @return The map is a bucket set for histogram, the key is specific bucket's upperBound, the value is item count
+     * in this bucket lower than or equals to key(upperBound)
+     */
+    private static Map<Double, Long> buildBucketsFromExponentialHistogram(
+        int positiveOffset, final List<Long> positiveBucketCounts,
+        int negativeOffset, final List<Long> negativeBucketCounts, int scale) {
+
+        final Map<Double, Long> result = new HashMap<>();
```
kezhenxu94, line 187: Looks like you didn't take into consideration of the `zero_count` field in OTEL exponential histgram data model? Can you elaborate why?

mufiye: I think I miss this part, I would like to assign its upper bound as the lower bound of small positive, WDYT?

wu-sheng: From my understanding, zero means less than a threshold, no matter this is less than which boundaries.

wu-sheng:
> > @mufiye One question, does this histogram always build bucket boundaries as integers? Could it be a very small double? 
> 
> 
> 
> This histogram always builds bucket boundaries as double. I don't fully understand what you mean.
> 
> 

Check hisgram percentile function, which calculate the value. The percentile value is the rank of the bucket. 
Read the calculation will help you to understand the process.

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
kezhenxu94: > I think I miss this part, I would like to assign its upper bound as the lower bound of small positive, WDYT?

> From my understanding, zero means less than a threshold, no matter this is less than which boundaries.

There is clear definition of zeros in OTEL data model, we can tackle the zeros according to the definition, in short, the `zero_count` is the count of bucket `(-threshold, +threshold)`, or count of samples that are approximately `0` if `threshold` is not set.

> The ExponentialHistogram contains a special zero_count bucket and an optional zero_threshold field where zero_count contains the count of values whose absolute value is less than or equal to zero_threshold. The precise value for the zero_threshold is arbitrary and not related to the scale.

> When zero_threshold is unset or 0, this bucket stores values that cannot be expressed using the standard exponential formula as well as values that have been rounded to zero.

wu-sheng: Ignoring the count make sense when do percentile, as it should not be considered as base.
But if we read the count, maybe this helps.
This is just my understanding.

wu-sheng: > There is clear definition of zeros in OTEL data model, we can tackle the zeros according to the definition, in short, the zero_count is the count of bucket (-threshold, +threshold), or count of samples that are approximately 0 if threshold is not set.

Yes, they defined. Last time, I check, it mentioned, the zero threshold even could cover over the existing bucket scopes.

wu-sheng:
Check how `SampleFamily#histogram_percentile` works(Ref Analyzer#init, createMetric).

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
wu-sheng: Check this https://opentelemetry.io/docs/reference/specification/metrics/data-model/#exponentialhistogram

> The ExponentialHistogram contains a special zero_count bucket and an optional zero_threshold field where zero_count contains the count of values whose absolute value is less than or equal to zero_threshold. The precise value for the zero_threshold is arbitrary and not related to the scale.

So, we can't assume the range of the threshold would not overlap with existing buckets. We can't build more buckets from it.

wu-sheng: We could leave some comments here as a TBD about ignoring this, maybe one day, we will know how should we process this. WDYT?

kezhenxu94: > We could leave some comments here as a TBD about ignoring this, maybe one day, we will know how should we process this. WDYT?

Good to me

mufiye: > We could leave some comments here as a TBD about ignoring this, maybe one day, we will know how should we process this. WDYT?

I'm back from class. I read the docs and the new opentelemetry proto file. I think it is a nice idea.

mufiye:
> @mufiye One question, does this histogram always build bucket boundaries as integers? Could it be a very small double? Because when we calculate percentile from the histogram (usually we do today), the values are bucket boundaries and they should be integer/long only.

I have understood this question. I think the small double case maybe happen.

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
mufiye: About comments, should I add them to code?

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
```diff
@@ -161,6 +163,54 @@ private static Map<Double, Long> buildBuckets(
...
     }
 
+    /**
+     * ExponentialHistogram data points are an alternate representation to the Histogram data point in OpenTelemetry
+     * metric format(https://opentelemetry.io/docs/reference/specification/metrics/data-model/#exponentialhistogram).
+     * It uses scale, offset and bucket index to calculate the bound. Firstly, calculate the base using scale by
+     * formula: base = 2**(2**(-scale)). Then the upperBound of specific bucket can be calculated by formula:
+     * base**(offset+index+1). Above calculation way is about positive buckets. For the negative case, we just
+     * map them by their absolute value into the negative range using the same scale as the positive range. So the
+     * upperBound should be calculated as -base**(offset+index).
```
wu-sheng, line 173: @mufiye You could add some comments here about `zero_count`.

wu-sheng:
> > @mufiye One question, does this histogram always build bucket boundaries as integers? Could it be a very small double? Because when we calculate percentile from the histogram (usually we do today), the values are bucket boundaries and they should be integer/long only.
> 
> I have understood this question. I think the small double case maybe happen. But this case also maybe happen for otel histogram data.

If current, the buckets of airflow definition are good, we should be fine. Otherwise, we may need some closure to process the buckets, such as maybe, we need multiple 1000 from microsecond to millisecond precision.

wu-sheng:
> I review the data collected from airflow and find there are some small timer data, such as 1.182709 and 1.348083 in one exponentialHistogram data. Where should I add the closure?

When you write MAL.

wu-sheng:
> > > I review the data collected from airflow and find there are some small timer data, such as 1.182709 and 1.348083 in one exponentialHistogram data. Where should I add the closure?
> > 
> > 
> > When you write MAL.
> 
> I read other services' otel rule cases and meter analysis language but have no idea how to do this. Could you explain it more?

There is no other for these. Because they don't use small doubles as buckets. 

If you want to learn this for now, then MAL is groovy based script. `SampleFamily` is the root class running in MAL engine, and this is the type referred by the metric name of the expression. 
You can see `SampleFamily#filter(Closure<Boolean> filter)`, this is what I mean about closure. It means a code block as an inner class for the metric process.
We don't have a similar mechanism for this, but from what you described these buckets, we seem to need one for the exponential histogram.  For example, we may need `buildBucketsFromExponentialHistogram(..., Closure<Boolean> processor)` as another new method. Then we could write some codes in the MAL expression to do `mutiple 1000`(ns -> ms) mentioned above.

mufiye:
> If you want to learn this for now, then MAL is groovy based script. `SampleFamily` is the root class running in MAL engine, and this is the type referred by the metric name of the expression. You can see `SampleFamily#filter(Closure<Boolean> filter)`, this is what I mean about closure. It means a code block as an inner class for the metric process. We don't have a similar mechanism for this, but from what you described these buckets, we seem to need one for the exponential histogram. For example, we may need `buildBucketsFromExponentialHistogram(..., Closure<Boolean> processor)` as another new method. Then we could write some codes in the MAL expression to do `mutiple 1000`(ns -> ms) mentioned above.

I think I need to fully understand the meter analyzer code first. In my current opinion, the function can not be put into the `OpenTelemetryMetricRequestProcessor#adaptMetrics` part and should be put in the analyzer part. And should we also consider small double data case for "otel histogram data"? Anyway, the difference between histogram and ExponentialHistogram is that the latter does the data compression.

wu-sheng:
>  the function can not be put into the OpenTelemetryMetricRequestProcessor#adaptMetrics part and should be put in the analyzer part. And should we also consider small double data case for "otel histogram data"? 

Yes, it should be on the MAL(SampleFamily) part.

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
```diff
@@ -161,6 +163,57 @@ private static Map<Double, Long> buildBuckets(
...
+     *
+     * We temporarily ignored the zero_count field, because the zero_threshold even could cover over the existing
+     * bucket scopes. The discussion is in pull request #10449.
```
wu-sheng, line 176:
```suggestion
     * Ignored the zero_count field temporarily, 
     * because the zero_threshold even could overlap the existing bucket scopes. 
```

Usually, we don't use `we` in the comments. The codes are we :)

wu-sheng (approved):
LGTM @kezhenxu94 @wankai123 Please take another look.

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
kezhenxu94: > Check this https://opentelemetry.io/docs/reference/specification/metrics/data-model/#exponentialhistogram
> 
> 
> 
> > The ExponentialHistogram contains a special zero_count bucket and an optional zero_threshold field where zero_count contains the count of values whose absolute value is less than or equal to zero_threshold. The precise value for the zero_threshold is arbitrary and not related to the scale.
> 
> 
> 
> So, we can't assume the range of the threshold would not overlap with existing buckets. We can't build more buckets from it.

Actually I am curious where does it indicate the threshold can overlap with existing buckets ?

wu-sheng: > The precise value for the zero_threshold is arbitrary and not related to the scale.

**not related to the scale** means **not related to the bucket**. The buckets are made by (2**2**-scale)

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
```diff
@@ -161,6 +163,57 @@ private static Map<Double, Long> buildBuckets(
...
+        if (base == Double.POSITIVE_INFINITY) {
+            log.warn("Receive and reject out-of-range ExponentialHistogram data");
+            return result;
+        }
+        double upperBound;
+        for (int i = 0; i < negativeBucketCounts.size(); i++) {
+            upperBound = -Math.pow(base, negativeOffset + i);
+            if (upperBound == Double.NEGATIVE_INFINITY) {
+                log.warn("Receive and reject out-of-range ExponentialHistogram data");
+                break;
```
kezhenxu94, line 201: I think you want to `return` here. We don't want to save part of the invalid data

oap-server/server-receiver-plugin/otel-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessor.java
```diff
@@ -161,6 +163,57 @@ private static Map<Double, Long> buildBuckets(
...
+            }
+            result.put(upperBound, negativeBucketCounts.get(i));
+        }
+        for (int i = 0; i < positiveBucketCounts.size() - 1; i++) {
+            upperBound = Math.pow(base, positiveOffset + i + 1);
+            if (upperBound == Double.POSITIVE_INFINITY) {
+                log.warn("Receive and reject out-of-range ExponentialHistogram data");
+                break;
```
kezhenxu94, line 209: Same here, `return`

oap-server/server-receiver-plugin/otel-receiver-plugin/src/test/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessorTest.java
```diff
@@ -0,0 +1,134 @@
...
+                                                                                          .setOffset(10)
+                                                                                          .addBucketCounts(
+                                                                                              1) // (0, 6.72]
+                                                                                          .addBucketCounts(
+                                                                                              1
+                                                                                          ) // (6.72, 8]
+                                                                                          .addBucketCounts(
+                                                                                              1
+                                                                                          ) // (8, 9.51]
+                                                                                          .build();
```
kezhenxu94, line 74: I meant this (I just change the type, please reformat the following lines):

```suggestion
        var var1 = ExponentialHistogramDataPoint.Buckets.newBuilder()
                                                                                          .setOffset(10)
                                                                                          .addBucketCounts(
                                                                                              1) // (0, 6.72]
                                                                                          .addBucketCounts(
                                                                                              1
                                                                                          ) // (6.72, 8]
                                                                                          .addBucketCounts(
                                                                                              1
                                                                                          ) // (8, 9.51]
                                                                                          .build();
```

oap-server/server-receiver-plugin/otel-receiver-plugin/src/test/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessorTest.java
```diff
@@ -0,0 +1,134 @@
...
+                                                                                          ) // (-INFINITY, -19.02]
+                                                                                          .build();
+        ExponentialHistogramDataPoint dataPoint = ExponentialHistogramDataPoint.newBuilder()
+                                                                               .setCount(7)
+                                                                               .setSum(-50)
+                                                                               .setScale(2)
+                                                                               .setPositive(var1)
+                                                                               .setNegative(var2)
+                                                                               .setTimeUnixNano(1000000)
+                                                                               .build();
```
kezhenxu94, line 95:
```suggestion
        var var2 = ExponentialHistogramDataPoint.Buckets.newBuilder()
                                                                                          .setOffset(15)
                                                                                          .addBucketCounts(
                                                                                              1
                                                                                          ) // (-16, -13.45]
                                                                                          .addBucketCounts(
                                                                                              2
                                                                                          ) // (-19.02, -16]
                                                                                          .addBucketCounts(
                                                                                              1
                                                                                          ) // (-INFINITY, -19.02]
                                                                                          .build();
        var dataPoint = ExponentialHistogramDataPoint.newBuilder()
                                                                               .setCount(7)
                                                                               .setSum(-50)
                                                                               .setScale(2)
                                                                               .setPositive(var1)
                                                                               .setNegative(var2)
                                                                               .setTimeUnixNano(1000000)
                                                                               .build();
```

oap-server/server-receiver-plugin/otel-receiver-plugin/src/test/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessorTest.java
```diff
@@ -0,0 +1,134 @@
...
+        // number is 4, 7, 9
+        ExponentialHistogramDataPoint.Buckets var1 = ExponentialHistogramDataPoint.Buckets.newBuilder()
+                                                                                          .setOffset(10)
+                                                                                          .addBucketCounts(
+                                                                                              1) // (0, 6.72]
+                                                                                          .addBucketCounts(
+                                                                                              1
+                                                                                          ) // (6.72, 8]
+                                                                                          .addBucketCounts(
+                                                                                              1
```
kezhenxu94, line 72: Can you change the bucket counts to different values? For example `1, 2, 3`, so we can verify in the test `1, 2, 3` without worrying we mixed each other and the test still passed

oap-server/server-receiver-plugin/otel-receiver-plugin/src/test/java/org/apache/skywalking/oap/server/receiver/otel/otlp/OpenTelemetryMetricRequestProcessorTest.java
```diff
@@ -0,0 +1,134 @@
...
+        ExponentialHistogramDataPoint.Buckets var2 = ExponentialHistogramDataPoint.Buckets.newBuilder()
+                                                                                          .setOffset(15)
+                                                                                          .addBucketCounts(
+                                                                                              1
+                                                                                          ) // (-16, -13.45]
+                                                                                          .addBucketCounts(
+                                                                                              2
+                                                                                          ) // (-19.02, -16]
+                                                                                          .addBucketCounts(
+                                                                                              1
```
kezhenxu94, line 85: Same here, perhaps change to `3`

kezhenxu94 (approved):
looks good

PR #10450, opened 2023-02-26
fix:Avoid accessing static methods of a class through its object ref

4yDX3906:
Avoid accessing static methods of a class through its object reference.


wu-sheng:
@wankai123 I think this is caused by last PromQL support?

wankai123 (approved):

PR #10415, opened 2023-02-20
Support prometheus HTTP API and promQL.

wankai123:
- [ ] If this is non-trivial feature, paste the links/URLs to the design doc.
- [ ] Update the documentation to include this new feature.
- [X] Tests(including UT, IT, E2E) are added to verify the new feature.
- [X] If it's UI related, attach the screenshots below.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #10080 .
- [X] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/docs/en/changes/changes.md).

Support HTTP API list:
- /api/v1/metadata
- /api/v1/labels
- /api/v1/label/{label_name}/values
- /api/v1/series
- /api/v1/query
- /api/v1/query_range

Support basic promQL for query traffic/metrics/labeled_metrics/topN/sampled_records from OAP

A Grafana dashboards demo based on this feature can visit:
http://34.92.90.119:3000/
admin/admin

Include: General Service / Service Mesh / Virtual-Database

PR #10459, opened 2023-02-28
Add PromQL Service doc and how to use in Grafana.

wankai123:
Doc for: https://github.com/apache/skywalking/pull/10415

### DIFF
--- a/docs/en/api/promql-service.md
+++ b/docs/en/api/promql-service.md
@@ -1,0 +1,632 @@
+# PromQL Service
+PromQL([Prometheus Query Language](https://prometheus.io/docs/prometheus/latest/querying/basics/)) Service
+provide Promql-based expression language and HTTP API.
+Third-party systems or visualization platforms that already support PromQL (such as Grafana),
+If metrics data are required for integration, they can be obtained directly from SkyWalking through the PromQL Service.
+
+Since SkyWalking and Prometheus have different mechanisms for metrics including classification, format, storage, etc.
+The PromQL Service supported will be a subset of the full PromQL.
+
+## Details Of Supported Protocol
+The following doc describes the details of the supported protocol and compared it to the PromQL official documentation.
+If not mentioned, it will not be supported by default.
+
+### Time series Selectors
+#### [Instant Vector Selectors](https://prometheus.io/docs/prometheus/latest/querying/basics/#instant-vector-selectors)
+For example: select metric `service_cpm` which the service is `$service` and the layer is `$layer`.
+```shell
+service_cpm{service='$service', layer='$layer'}
+```
+**Note: The label matching operators only support `=` instead of regular expressions.**
+
+#### [Range Vector Selectors](https://prometheus.io/docs/prometheus/latest/querying/basics/#range-vector-selectors)
+For example: select metric `service_cpm` which the service is `$service` and the layer is `$layer` within the last 5 minutes.
+```shell
+service_cpm{service='$service', layer='$layer'}[5m]
+```
+
+#### [Time Durations](https://prometheus.io/docs/prometheus/latest/querying/basics/#time-durations)
+| Unit | Definition   | Support |
+|------|--------------|---------|
+| ms   | milliseconds | yes     |
+| s    | seconds      | yes     |
+| m    | minutes      | yes     |
+| h    | hours        | yes     |
+| d    | days         | yes     |
+| w    | weeks        | yes     |
+| y    | years        | **no**  |
+
+### Binary operators
+#### [Arithmetic binary operators](https://prometheus.io/docs/prometheus/latest/querying/operators/#arithmetic-binary-operators)
+| Operator | Definition           | Support |
+|----------|----------------------|---------|
+| +        | addition             | yes     |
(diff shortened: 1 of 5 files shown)


docs/en/api/promql-service.md
```diff
@@ -0,0 +1,632 @@
...
+
+| Parameter | Definition      | Support | Optional |
+|-----------|-----------------|---------|----------|
+| match[]   | series selector | yes     | no       |
+| start     | start timestamp | **no**  | yes      |
+| end       | end timestamp   | **no**  | yes      |
+
+For example:
+```shell
+/api/v1/label/__name__/values'
```
kezhenxu94, line 311: Look like a typo, and there are the same typos in other places, please check

```suggestion
/api/v1/label/__name__/values
```

docs/en/api/promql-service.md
```diff
@@ -0,0 +1,632 @@
+# PromQL Service
+PromQL([Prometheus Query Language](https://prometheus.io/docs/prometheus/latest/querying/basics/)) Service
+provide Promql-based expression language and HTTP API.
+Third-party systems or visualization platforms that already support PromQL (such as Grafana),
+If metrics data are required for integration, they can be obtained directly from SkyWalking through the PromQL Service.
```
wu-sheng, line 5:
```suggestion
PromQL([Prometheus Query Language](https://prometheus.io/docs/prometheus/latest/querying/basics/)) Service
exposes Prometheus Querying HTTP APIs including the bundled PromQL expression system.
Third-party systems or visualization platforms that already support PromQL (such as Grafana), 
could obtain metrics through PromeQL Service.
```

docs/en/api/promql-service.md
```diff
@@ -0,0 +1,632 @@
...
+
+Since SkyWalking and Prometheus have different mechanisms for metrics including classification, format, storage, etc.
+The PromQL Service supported will be a subset of the full PromQL.
```
wu-sheng, line 8:
```suggestion
As SkyWalking and Prometheus have fundamental differences in metrics classification, format, storage, etc. 
The PromQL Service supported will be a subset of the complete PromQL
```

docs/en/api/promql-service.md
```diff
@@ -0,0 +1,632 @@
...
+### HTTP API
+
+#### Expression queries
+
+##### [Instant queries](https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries)
+```shell
+GET|POST /api/v1/query
+```
+
+| Parameter | Definition                                                                                                                          | Support | Optional   |
```
wu-sheng, line 106: Could you explain what `Optional` mean? The context seems not very clear.

docs/en/api/promql-service.md
```diff
@@ -0,0 +1,632 @@
...
+    ],
+    ...
+  }
+}
+```
+
+## Metrics Type For Query
+
+### Supported Metrics [Scope](../../../oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/enumeration/Scope.java)(Catalog)
+Full scopes are not currently supported, please check the following table:
```
wu-sheng, line 385:
```suggestion
All scopes are not supported completely, please check the following table:
```

docs/en/api/promql-service.md
```diff
@@ -0,0 +1,632 @@
...
+      }
+    ]
+  }
+}
+```
+
+### Sort Metrics
+- Query Labels:
+```shell
+--parent_service: Name of the parent service
```
wu-sheng, line 513: This should be optional? Do we support that? Such as we have a TopN slow service list.
I know we are not recommending that due to Layers mixed, but in theory, we could do this query if users have only one layer, right?

docs/en/setup/backend/ui-grafana.md
```diff
@@ -0,0 +1,50 @@
+# Use Grafana As The UI
+Since 9.4.0, SkyWalking provide [PromQL Service](../../api/promql-service.md). You can choose [Grafana](https://grafana.com/) 
+as the SkyWalking UI. About the installation and how to use please refer to the [official document](https://grafana.com/docs/grafana/v9.3/).
```
wu-sheng, line 3:
```suggestion
as the SkyWalking UI. About the installation and how to use please refer to the [official document](https://grafana.com/docs/grafana/v9.3/).

Notice <1>, Gafana is [AGPL-3.0 license](https://github.com/grafana/grafana/blob/main/LICENSE), which is very different from Apache 2.0.
Please follow AGPL 3.0 license requirements.

Notice <2>, SkyWalking always uses its native UI as first class. All visualization features are only available on native UI.
Grafana UI is an extension on our support of PromQL APIs. We don't maintain or promise the complete Grafana UI dashboard setup.
```

docs/en/api/promql-service.md
wankai123: In the Prometheus official doc, use `Optional` so I think follow this is better.

wu-sheng: OK, got it.

wu-sheng (approved):

PR #10466, opened 2023-02-28
fix:Utility classes should not have public constructors

4yDX3906:
Utility classes, which are collections of static members, are not meant to be instantiated. Even abstract utility classes, which can be extended, should not have public constructors.
Java adds an implicit public constructor to every class which does not define at least one explicitly. Hence, at least one non-public constructor should be defined.


wu-sheng:
These code style things seem not having meaning in practice.
We don't want to be strict for this after all.
We intend to resolve real problems.

PR #10476, opened 2023-03-03
Feature DCS supports global configuration

### LINKED ISSUE #10461: [Feature] DCS supports global configuration
### Search before asking

### Description

Add a configuration item "global-configurations" to the dynamic configuration
Global configuration has lower priority than service level "configurations"

### Use case

image

### Related issues

_No response_

### Code of Conduct

xlq20080808:
###  DCS supports global configuration
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #10461 .

The priority of each service configuration is higher than the global configuration, format like this
```
configurations:
  // NEW added global config
  key1 : value1
  key2: value2
  ...

  //service name
  serviceA:
    // Configurations of service A
    // Key and Value are determined by the agent side.
    // Check the agent setup doc for all available configurations.
    key1: value1
    key2: value2
    ...
  serviceB:
    ...
```


wu-sheng:
Do you need a (default-config) as a hard coded key? Rather than putting things in the root?
I feel the codes are more complex than I thought.

xlq20080808:
I agree, I have modified
 format like this

configurations:
  // NEW added global config
  default-config:
    key1 : value1
    key2: value2
  ...

  //service name
  serviceA:
    // Configurations of service A
    // Key and Value are determined by the agent side.
    // Check the agent setup doc for all available configurations.
    key1: value1
    key2: value2
    ...
  serviceB:
    ...

oap-server/server-receiver-plugin/configuration-discovery-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/recevier/configuration/discovery/AgentConfigurationsReader.java
```diff
@@ -54,6 +54,9 @@ public AgentConfigurationsTable readAgentConfigurationsTable() {
                 if (configurationsData != null) {
                     configurationsData.forEach((k, v) -> {
                         Map map = (Map) v;
+                        if (configurationsData.containsKey("default-config")) {
+                            ((Map) configurationsData.get("default-config")).forEach(map::putIfAbsent);
+                        }
```
wu-sheng, line 59: I am confused about this part. As `default-config` is a reserved key, and it shares the same structure as other services. Why do we have to change the init logic?
Also, we don't build `agentConfigurationsTable` properly(no UUID part).

oap-server/server-receiver-plugin/configuration-discovery-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/recevier/configuration/discovery/AgentConfigurationsWatcher.java
```diff
@@ -73,7 +73,7 @@ public String value() {
     public AgentConfigurations getAgentConfigurations(String service) {
         AgentConfigurations agentConfigurations = agentConfigurationsTable.getAgentConfigurationsCache().get(service);
         if (null == agentConfigurations) {
-            return emptyAgentConfigurations;
+            return agentConfigurationsTable.getAgentConfigurationsCache().getOrDefault("default-config", emptyAgentConfigurations);
```
wu-sheng, line 76: This logic should work directly, isn't it? As from YAML perspective, `default-config` is a `service name` only.

wu-sheng: Please correct me I missed anything.

wu-sheng:
@xlq20080808 Are you following this?

xlq20080808:
@wu-sheng yes i'm following

oap-server/server-receiver-plugin/configuration-discovery-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/recevier/configuration/discovery/AgentConfigurationsReader.java
xlq20080808: We need to add the missing configuration items of each service in default-config to the each service configuration, I think they should be similar to union relationship. And then we will traverse the map to generate UUID for each service

wu-sheng:
OK, I think I got your point. I would suggest you enhance `AgentConfigurations` rather than pushing all configurations into the service-oriented configuration. WDYT?

xlq20080808:
Thank you for your suggestion. I will modify it

xlq20080808:
Try to move the merging of config to ConfigurationDiscoveryServiceHandler

oap-server/server-receiver-plugin/configuration-discovery-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/recevier/configuration/discovery/AgentConfigurations.java
```diff
@@ -37,9 +41,23 @@ public class AgentConfigurations {
...
 
+    private volatile boolean needMerge = true;
+
     public AgentConfigurations(final String service, final Map<String, String> configuration, final String uuid) {
         this.service = service;
         this.configuration = configuration;
         this.uuid = uuid;
     }
+
+    public void mergeAgentConfigurations(Map<String, String> config) {
```
wu-sheng, line 52: Actually, you don't need this. The only thing `ConfigurationDiscoveryServiceHandler#newAgentDynamicConfigCommand` required is `getConfiguration`. You could build an iterator by including both config.

You could do `agentConfigurations.buildConfiguraion(agentConfigurationsWatcher.getAgentConfigurations("default-config"))`.`foreach`.

By this, you don't need `needMerge`, also don't need to really merge two maps.

wu-sheng:
I think you still make the code boundaries not clear, which makes me hard to approve this change to upstream.
Let me try to re-word how I think the codes should be.
1. The default config should be held inside every other AgentConfigurations, rather than read every time.
2. You don't need the AgentConfigurations for the default config as the map is in other AgentConfigurations already.
3. UUID should be built inside the constructor as we need to combine the value of service-specific and default configs. 
4. Build a temp iterator, rather than using streaming to merge those maps, which doesn't provide benefits but payload.

wu-sheng:
Close for now, as this has not been updated for a week. If you want to continuous on this, please polish the codes, the logic is clear, we just need the codes better. Thanks.

StruggleYang:
Is this feature under further development? I found that I needed this configuration recently because I needed to globally exclude some endpoints from capture

PR #10478, opened 2023-03-03
 Fix SQL OTEL document path error.

yswdqz:


wu-sheng:
I think `opentelemetry-receiver.md` should be updated as well.

wu-sheng (approved):

PR #10500, opened 2023-03-08
register FreeRedis as 3018

guochen2:
Register FreeRedis as 3018


wu-sheng:
Please update changes.md according to the PR template(you removed all somehow).

wu-sheng (approved):

PR #10513, opened 2023-03-08
Fix release doc

wu-sheng:


kezhenxu94:
We only need one of this or https://github.com/apache/skywalking/pull/10511

kezhenxu94 (approved):

PR #10518, opened 2023-03-09
Fix wrong layer of metric `userError` in DynamoDB monitoring.

yswdqz:


oap-server/server-starter/src/main/resources/otel-rules/aws-dynamodb/dynamodb-endpoint.yaml
```diff
@@ -78,8 +78,7 @@ metricsRules:
     exp: amazonaws_com_AWS_DynamoDB_SystemErrors.tagMatch('Operation','GetItem|Scan|Query|BatchGetItem|TransactGetItems')
   - name: endpoint_write_system_errors
     exp: amazonaws_com_AWS_DynamoDB_SystemErrors.tagMatch('Operation','PutItem|UpdateItem|DeleteItem|BatchWriteItem|TransactWriteItems')
-  - name: endpoint_user_errors
```
wu-sheng, line 81: I can see `User Errors` in the `backend-aws-dynamodb-monitoring.md` doc, should it update as well?

wu-sheng:
Could you explain what does the `wrong layer` mean?

PR #10522, opened 2023-03-09
Support to bind TLS status as a part of component for service topology.

wu-sheng:

By this, the service topology(source/target components) includes the TLS for the service mesh case. The components could be [http, tls], [http, mtls], etc.
@toffentoffen Would you like to try this?

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -7,6 +7,7 @@
 
 * Fix wrong layer of metric `user error` in DynamoDB monitoring.
 * Support to bind TLS status as a part of component for service topology.
+* Fix component ID priority bug.
 
 #### Documentation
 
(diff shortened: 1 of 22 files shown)


kezhenxu94 (dismissed):

kezhenxu94 (dismissed):

wu-sheng:
@kezhenxu94 One left issue caused by Service Mesh analysis. The TLS_MODE is not set by ALS analyzer, which leads to clientComponent is always `http` only(no tls status).

wu-sheng:
> @kezhenxu94 One left issue caused by Service Mesh analysis. The TLS_MODE is not set by ALS analyzer, which leads to clientComponent is always `http` only(no tls status).

image

We tried to check, the TLS property seems not to be reported from client side by ALS.

wu-sheng:
TODO @kezhenxu94 @mrproliu We don't query source/target component IDs from CLI. We should fix these(I feel this may be caused by these fields being relatively new added).

kezhenxu94 (approved):

PR #10534, opened 2023-03-13
JavaDoc input should contain delombok output and fix javadoc not executed in CI

kezhenxu94:
mvn install doesn’t include the javadoc goal so the CI didn’t find the problem introduced in #10532, which changed the delombok output but didn’t change the javadoc input accordingly

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -192,6 +192,8 @@
         <skipUTs>false</skipUTs>
 
         <system-stubs.version>2.0.2</system-stubs.version>
+
+        <delombok.output.dir>${project.build.directory}/delombok</delombok.output.dir>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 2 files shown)


wu-sheng (approved):

pom.xml
```diff
@@ -513,7 +515,7 @@
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <configuration>
                     <sourcepath>
-                        ${project.build.directory}/generated-sources/delombok;
+                        ${delombok.output.dir};
```
wu-sheng, line 518: This looks like revert the previous change, which could lead IDE reports errors again. Have you tested it?

kezhenxu94:
> > CI / Check JavaDoc (pull_request)
> 
> Oops, this still doesn't work.

turns out that `javadoc:javadoc' goal must be run after a maven install(otherwise it failed to find submodule dependencies), we cannot separate it in an individual job, so I explicitly add `javadoc` goal to the `package` job right after the install goal.

kezhenxu94:
And the following line should have prevented the delombok’ed sources being added as source codes in maven, not sure whether it’s the IDE automatically adds the `generated-sources` as source codes or not. Anyway it’s a good idea to put the delombok’ed codes in a different folder than those protobuf codes that we expect to be added as source codes. 

https://github.com/apache/skywalking/blob/c74d21c96fee0c830ef6f4e27b6f59fb61aafc70/pom.xml#L491

wu-sheng (approved):

PR #10544, opened 2023-03-16
[Breaking Change] Enhance JDBC storage through merging table and managing day-based table rolling

### LINKED ISSUE #10366: [Feature][Breaking Change] Enhance SQL-database storage through merging table and managing day-based table rolling.

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #10366.


wu-sheng:
Should we remove the tidb provider with relative docs(maybe keep a little in MySQL section)? As the rolling and drop table as TTL, the TiDB provider doesn't seem necessary.

wu-sheng:
> > Should we remove the tidb provider with relative docs(maybe keep a little in MySQL section)? As the rolling and drop table as TTL, the TiDB provider doesn't seem necessary.
> 
> Do you want to keep the tests for TiDB?

 I think we are safe to remove. We don't have obligations to test compatibility.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/type/SelectedRecord.java
```diff
@@ -38,7 +38,7 @@ public class SelectedRecord {
     /**
      * Usually an integer value as this is metrics.
      */
-    private String value;
+    private long value;
```
wu-sheng, line 41: I think you missed reverting this back.

wu-sheng:
> [Disable SS related tests](https://github.com/apache/skywalking/pull/10544/commits/e59626b6a51b8f596bff0379aa270e0b797289e6)

Does this mean we will skip SS implementation for this PR?

kezhenxu94:
> > [Disable SS related tests](https://github.com/apache/skywalking/pull/10544/commits/e59626b6a51b8f596bff0379aa270e0b797289e6)
> 
> Does this mean we will skip SS implementation for this PR?

Yes, I have no idea (and I have no interest in) supporting ShardingSphere if I can choose, honestly. I'll manage to make the tests passed in a following PR tho.

wu-sheng:
With the new mechanism, actually, we could work better on the traditional database, and many transparent options to be distributed MySQL/PostgreSQL DBs. I am considering the reasons for keeping or removing it.

oap-server/server-storage-plugin/storage-shardingsphere-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/shardingsphere/ShardingRulesOperator.java
```diff
@@ -157,23 +152,35 @@ private ShardingRule.ShardingRuleBuilder buildShardingRule(ShardingRule.Sharding
...
-            case TIME_RELATIVE_ID_SHARDING_ALGORITHM:
-                buildExpressionTableStrategy(builder, tableName, tableShardingColumn,
-                                             TIME_RELATIVE_ID_SHARDING_EXPRESSION);
-                break;
-            case TIME_BUCKET_SHARDING_ALGORITHM:
-                buildExpressionTableStrategy(builder, tableName, tableShardingColumn, TIME_BUCKET_SHARDING_EXPRESSION);
-                break;
-            default:
-                throw new UnexpectedException("Unsupported sharding algorithm " + shardingAlgorithm);
-        }
```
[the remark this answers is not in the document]
kezhenxu94: > We don't need these anymore? But I don't see you remove them from the top of the Class code.

Now removed.

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/TableHelper.java
```diff
@@ -0,0 +1,141 @@
...
+public class TableHelper {
+    private final ModuleManager moduleManager;
+    private final JDBCClient jdbcClient;
+
+    public static String getTableName(Model model) {
+        final var aggFuncName = FunctionCategory.uniqueFunctionName(model.getStreamClass()).replaceAll("-", "_");
+        return StringUtil.isNotBlank(aggFuncName) ? aggFuncName : model.getName();
+    }
+
+    public static String getTableForWrite(Model model) {
```
wu-sheng, line 56:
```suggestion
    public static String getLatestTableForWrite(Model model) {
```

I feel this should not be an overload method. It only represents for getting the name of latest table.

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/dao/JDBCHistoryDeleteDAO.java
```diff
@@ -18,66 +18,100 @@
...
+                    break;
+                default:
+                    return;
             }
-        } catch (JDBCClientException | SQLException e) {
-            throw new IOException(e.getMessage(), e);
+        }
+        // Delete data in additional tables
+        for (final var additionalTable : model.getSqlDBModelExtension().getAdditionalTables().values()) {
+            SQLBuilder additionalTableDeleteSQL = new SQLBuilder("delete from " + additionalTable.getName() + " where ")
```
wu-sheng, line 104: Why `delete` rather than `drop`? I think the additional table should follow the timeseries process policy.

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/TableHelper.java
```diff
@@ -0,0 +1,141 @@
...
+        timeBucketStart = TimeBucket.getTimeBucket(TimeBucket.getTimestamp(timeBucketStart), DownSampling.Day);
+        timeBucketEnd = TimeBucket.getTimeBucket(TimeBucket.getTimestamp(timeBucketEnd), DownSampling.Day);
+
+        return LongStream
+            .rangeClosed(timeBucketStart, timeBucketEnd)
+            .distinct()
+            .mapToObj(it -> tableName + "_" + it)
+            .filter(table -> {
+                try {
+                    return jdbcClient.tableExists(table);
```
wu-sheng, line 97: I think failing to read from Database in sync may be not worth it. 
1. It could be slow
2. If this kind of query keeps happening, we cost a lot of resources in the database. 

Instead, moving the table cache from `DeleteDao` to JDBC global scope should help. After all, it only read once period.

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/dao/JDBCHistoryDeleteDAO.java
```diff
@@ -18,66 +18,100 @@
...
 
     @Override
-    public void deleteHistory(Model model, String timeBucketColumnName, int ttl) throws IOException {
-        SQLBuilder dataDeleteSQL = new SQLBuilder("delete from " + model.getName() + " where ")
-            .append(timeBucketColumnName).append("<= ? ")
-            .append(" and ")
-            .append(timeBucketColumnName).append(">= ? ");
+    @SneakyThrows
+    public void deleteHistory(Model model, String timeBucketColumnName, int ttl) {
+        final var endTimeBucket = TimeBucket.getTimeBucket(System.currentTimeMillis(), DownSampling.Day);
```
wu-sheng, line 52: `IHistoryDeleteDAO` is driven by DataTTLKeeperTimer for a period of 5 mins. I think we don't need to check the sync at such a high frequency.
Ref `HistoryDeleteEsDAO#indexLatestSuccess`, reduced the load. JDBC was about deleting from tables, so, can't follow this.

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/TableHelper.java
wu-sheng: OK, I noticed this is another misguide by naming. This method is from TTL, so, most likely we should rename to `getExistingTables` or similar.

wu-sheng:
@kezhenxu94 As you are not implementing the ShardingSphere in this PR, please remove all relative codes and tests. If one day, we will add it back, we could pick from 9.4.0 branch.
I don't hope we have confused change logs about this.

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/TableHelper.java
kezhenxu94: > I think failing to read from Database in sync may be not worth it.
> 
> 1. It could be slow
> 2. If this kind of query keeps happening, we cost a lot of resources in the database.
> 
> Instead, moving the table cache from `DeleteDao` to JDBC global scope should help. After all, it only read once period.

Caching these brings other problems, when the tables are deleted in another OAP node, we don't get notified, and still consider this table exists, then we might send queries to an inexistent table, which causes errors.

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/dao/JDBCHistoryDeleteDAO.java
kezhenxu94: Done

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/dao/JDBCHistoryDeleteDAO.java
kezhenxu94: Done

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/TableHelper.java
kezhenxu94: Done

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/TableHelper.java
wankai123: I think we should control the query start time and end time within TTL rather than to query table if exist.

wu-sheng: It is worth to a cache without loader mechanism to deal with this. 
The metadata reading is low performance. We should not include it into query process.

wu-sheng: A cache with timeout load mechanism.

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/TableHelper.java
```diff
@@ -54,6 +58,16 @@ public class TableHelper {
...
 
+    private final LoadingCache<String, Boolean> tableExistence =
+        CacheBuilder.newBuilder()
+                    .expireAfterWrite(10, TimeUnit.MINUTES)
+                    .build(new CacheLoader<>() {
+                        @Override
+                        public @NonNull Boolean load(@NonNull String tableName) throws Exception {
+                            return jdbcClient.tableExists(tableName);
+                        }
+                    });
```
wu-sheng, line 69: How about we have an `expireAfterRead` and do manually remove when `table not found` SQL exception? If possible, this could reduce more payload.

kezhenxu94: > How about we have an `expireAfterRead` and do manually remove when `table not found` SQL exception? If possible, this could reduce more payload.

Hi, `expireAfterWrite` --> `expireAfterRead` sounds reasonable to me. However manually invalidating the cache doesn't make any sense to me, manually invalidating the cache only cause the `LoadingCache` to invoke the `load` method and tries to fill the cache again and thus increase the load, it **doesn't** sounds like "hey, if there is no entry in cache, don't do anything", this is how a loading cache works: load in the cache first, if absent, load from a remote source or load through a heavy computational operations, by manually invalidating the cache you are requesting the heavy computational operations over and over again because the cache entry is invalidated manually.

kezhenxu94: Sounds like you want to capture the "table not found SQL exception", if that's the case, we don't even need a loading cache, we just use a plain cache (HashMap), and query the database for the first time, since then we only remove the cache when captured the exception, and the cache won't never expire unless exception is caught. Capturing such exception requires string manipulation though.

wu-sheng: > Sounds like you want to capture the "table not found SQL exception", if that's the case, we don't even need a loading cache, we just use a plain cache (HashMap), and query the database for the first time, since then we only remove the cache when captured the exception, and the cache won't never expire unless exception is caught. Capturing such exception requires string manipulation though.

Make sense. Map could be easier to archive this than cache.

kezhenxu94: The cache structure is `<String table, Boolean exists>`, so if the `N - 7` tables are absent, the results will be cached for 10 minutes (`expireAfterWrite(10, TimeUnit.MINUTES)`), so within the 10 minutes, no query will be sent to database. It's nothing special for the `N - 7` tables actually.

wu-sheng: Let's just change to `expireAfterRead`. I think it is enough.

kezhenxu94: Done

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/TableHelper.java
```diff
@@ -125,13 +125,10 @@ public List<String> getTablesForRead(String modelName, long timeBucketStart, lon
         timeBucketStart = TimeBucket.getTimeBucket(TimeBucket.getTimestamp(timeBucketStart), DownSampling.Day);
         timeBucketEnd = TimeBucket.getTimeBucket(TimeBucket.getTimestamp(timeBucketEnd), DownSampling.Day);
 
-        final var tablesWithinTTL = new HashSet<>(getTablesWithinTTL(modelName));
-
         return LongStream
             .rangeClosed(timeBucketStart, timeBucketEnd)
             .distinct()
             .mapToObj(it -> tableName + "_" + it)
-            .filter(tablesWithinTTL::contains)
```
wu-sheng: I think this is a good one. Why do you remove this?

wu-sheng:
I think I am mostly done with high-priority things for now.

docs/en/changes/changes.md
```diff
@@ -11,6 +11,7 @@
 * Support to bind TLS status as a part of component for service topology.
 * Fix component ID priority bug.
 * Fix component ID of topology overlap due to storage layer bugs.
+* [Breaking Change] Enhance JDBC storage through merging table and managing day-based table rolling.
```
wu-sheng, line 14:
```suggestion
* [Breaking Change] Enhance JDBC storage through merging tables and managing day-based table rolling.
* [Breaking Change] Sharding-MySQL implementations and tests get removed due to we have the day-based rolling mechanism by default
```

wu-sheng (approved):
LGTM.

## Commit logs
```
* [Breaking Change] Enhance JDBC storage through merging tables and managing day-based table rolling.
* [Breaking Change] Sharding-MySQL implementations and tests get removed due to we have the day-based rolling mechanism by default
```

wu-sheng:
@wankai123 Please take a look. 

FYI @mrproliu I expect this would be merged soon, and your block will be gone.

PR #10557, opened 2023-03-18
modify the service query and query by group and layer and modify the elasticsearch scroll queryMaxSize

xiaozaiyuji:

### DIFF
--- a/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetadataQueryV2.java
+++ b/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetadataQueryV2.java
@@ -67,6 +67,10 @@ public class MetadataQueryV2 implements GraphQLQueryResolver {
         return getMetadataQueryService().listServices(layer, null);
     }
 
+    public List<Service> listGroupServices(final String layer, final String group) throws IOException {
+        return getMetadataQueryService().listServices(layer, group);
+    }
+
     public Service findService(final String serviceName) throws IOException {
         return getMetadataQueryService().getService(IDManager.ServiceID.buildId(serviceName, true));
     }
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
@@ -67,6 +67,8 @@ public class MetadataQueryEsDAO extends EsDAO implements IMetadataQueryDAO {
     private static final Gson GSON = new Gson();
 
     private final int queryMaxSize;
+    private final int resultWindowMaxSize;
+    private final int metadataMaxSize;
     private final int scrollingBatchSize;
     private String endpointTrafficNameAlias;
     private boolean aliasNameInit = false;
@@ -76,9 +78,11 @@ public class MetadataQueryEsDAO extends EsDAO implements IMetadataQueryDAO {
         ElasticSearchClient client,
         StorageModuleElasticsearchConfig config) {
         super(client);
-        this.queryMaxSize = config.getMetadataQueryMaxSize();
+        this.metadataMaxSize = config.getMetadataQueryMaxSize();
+        this.resultWindowMaxSize = config.getResultWindowMaxSize();
         this.scrollingBatchSize = config.getScrollingBatchSize();
         this.layerSize = Layer.values().length;
+        this.queryMaxSize = Math.min(metadataMaxSize, resultWindowMaxSize);
     }
 
     @Override
@@ -86,7 +90,7 @@ public class MetadataQueryEsDAO extends EsDAO implements IMetadataQueryDAO {
         final String index =
             IndexController.LogicIndicesRegister.getPhysicalTableName(ServiceTraffic.INDEX_NAME);
 
-        final int batchSize = Math.min(queryMaxSize, scrollingBatchSize);
+        final int batchSize = Math.min(metadataMaxSize, scrollingBatchSize);
         final BoolQueryBuilder query =
             Query.bool();
         final SearchBuilder search = Search.builder().query(query).size(batchSize);
@@ -118,7 +122,7 @@ public class MetadataQueryEsDAO extends EsDAO implements IMetadataQueryDAO {
                     break;
                 }
                 // We've got enough data
-                if (services.size() >= queryMaxSize) {
+                if (services.size() >= metadataMaxSize) {
                     break;
                 }
                 results = getClient().scroll(SCROLL_CONTEXT_RETENTION, scrollId);
@@ -159,7 +163,7 @@ public class MetadataQueryEsDAO extends EsDAO implements IMetadataQueryDAO {
         if (IndexController.LogicIndicesRegister.isMergedTable(InstanceTraffic.INDEX_NAME)) {
             query.must(Query.term(IndexController.LogicIndicesRegister.METRIC_TABLE_NAME, InstanceTraffic.INDEX_NAME));
         }
-        final int batchSize = Math.min(queryMaxSize, scrollingBatchSize);
+        final int batchSize = Math.min(metadataMaxSize, scrollingBatchSize);
         final SearchBuilder search = Search.builder().query(query).size(batchSize);
 
         final List<ServiceInstance> instances = new ArrayList<>();
@@ -170,7 +174,7 @@ public class MetadataQueryEsDAO extends EsDAO implements IMetadataQueryDAO {
             if (batch.size() < batchSize) {
                 break;
             }
-            if (batch.size() >= queryMaxSize) {
+            if (batch.size() >= metadataMaxSize) {
                 break;
             }
             response = getClient().scroll(SCROLL_CONTEXT_RETENTION, response.getScrollId());


oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetadataQueryV2.java
```diff
@@ -67,6 +67,10 @@ public List<Service> listServices(final String layer) throws IOException {
         return getMetadataQueryService().listServices(layer, null);
     }
 
+    public List<Service> listGroupServices(final String layer, final String group) throws IOException {
```
wu-sheng, line 70: Where is the approval of this new query API? I can't find this.

wu-sheng (changes requested):

wu-sheng:
And what is the reason of ElasticSearch change?
A pull request must include the explanation. One pull request is about one thing.
Please be clear about the change, otherwise, we don't know what do you mean, and how could we follow?

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetadataQueryV2.java
xiaozaiyuji: > And what is the reason of ElasticSearch change? A pull request must include the explanation. One pull request is about one thing. Please be clear about the change, otherwise, we don't know what do you mean, and how could we follow?

Sorry this the first to pull request for open source ,I explain will explain it carefully.

wu-sheng: > > And what is the reason of ElasticSearch change? A pull request must include the explanation. One pull request is about one thing. Please be clear about the change, otherwise, we don't know what do you mean, and how could we follow?
> 
> Sorry this the first to pull request for open source ,I explain will explain it carefully.

You are replying on another context. Please the keep thread clear.

xiaozaiyuji: > Where is the approval of this new query API? I can't find this.

I add the group and layer search condition,because the performance for oap server is very sufficient.In product,I as the user with 10000 agent and use the group classify the search result will be limit by the query max size with min(metadataSize and scrollsize) with 5000,but the scroll query will help the user to search as many as the user want service to to serarch by the agent. In order to let the user to find the service sufficient,I want to modify the code.
1.I give the user to find the service by group and layer when the user have classify the service by group.
2.I give the user to query the service as they want the number they want but not limit by the elasticsearch maxWindowSize.
So I modify the word by change the two place,I think carefully and without to change the other,I think this will very flexible and will more sufficient.

xiaozaiyuji:
> And what is the reason of ElasticSearch change? A pull request must include the explanation. One pull request is about one thing. Please be clear about the change, otherwise, we don't know what do you mean, and how could we follow?

Sorry,I will explain it.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetadataQueryV2.java
xiaozaiyuji: In product,If the user change the metadataSize 、scrollsize over the elasticserch maxWindowSize.  ,It also will be limit by  the elasticserch maxWindowSize and give an exception. So I change the limit logical to use the scrollSize and metadataSize .This change will not cause the exception.

wu-sheng: > .In product,I as the user with 10000 agent and use the group classify the search result will be limit by the query max size with min(metadataSize and scrollsize) with 5000,

I can't follow this. 10k agents only mean 10k instances. Service must be much less than that.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
```diff
@@ -76,17 +78,19 @@ public MetadataQueryEsDAO(
         ElasticSearchClient client,
         StorageModuleElasticsearchConfig config) {
         super(client);
-        this.queryMaxSize = config.getMetadataQueryMaxSize();
+        this.metadataMaxSize = config.getMetadataQueryMaxSize();
+        this.resultWindowMaxSize = config.getResultWindowMaxSize();
         this.scrollingBatchSize = config.getScrollingBatchSize();
         this.layerSize = Layer.values().length;
+        this.queryMaxSize = Math.min(metadataMaxSize, resultWindowMaxSize);
```
wu-sheng, line 85: Why do you have to run `min` between these two configurations? `metadataQueryMaxSize` is already open for your manual setup. I can't see your point of changing this.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetadataQueryV2.java
xiaozaiyuji: In fact the skywalking ui include  many different agent, kubenetes,istio and many, but the user may by just use the java agent service or the python agent, so they will write a ui by they use,the new query API wll helpfull for this.

wu-sheng: > metadataSize 、scrollsize over the elasticserch maxWindowSize

All these are open for you to change. Why don't you change them by your own?

wu-sheng: > In fact the skywalking ui include many different agent, kubenetes,istio and many, but the user may by just use the java agent service or the python agent, so they will write a ui by they use,the new query API wll helpfull for this.

So what? I can't see in any cases, you are limited by this. Set them as the number you want ,

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
xiaozaiyuji: I change this because this is also limit by the elasticsearch resultWindowMaxSize, if the metadataQueryMaxSize  setting by 20000 over the default limit by elasticserch it will cause an exception. The metadataQueryMaxSize is add to help the user to query  with elasticserch sroll query to avoid the limit 10000 of elasticserch.In order to fix this problem,I did this by not affect the other query,but support the  elasticserch sroll query, this will separate the queryMaxSize and metadataSize,when the metadataSize is litte than the resultWindowMaxSize, will query the data as the user setting, when the user setting the metadataSize over the resultWindowMaxSize, the elastic scroll query will give the metadataSize result as the user want, the other query will also limit by the resultWindowMaxSize and will not cause an exception.

wu-sheng: > I change this because this is also limit by the elasticsearch resultWindowMaxSize

Why don't you change the value to fix the expectation? The server setup matters too.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetadataQueryV2.java
xiaozaiyuji: > > .In product,I as the user with 10000 agent and use the group classify the search result will be limit by the query max size with min(metadataSize and scrollsize) with 5000,
> 
> I can't follow this. 10k agents only mean 10k instances. Service must be much less than that.

In product, the metrics-all index will generate the data by day ,the service with the same name is added,does I setting something wrong for oap server.I  query the elasticsearch for index metrics-all and find the data is that.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
wu-sheng: If it is a hardcode case, then we could fix it. But AFAIK, all these are exposed to you. Make them right, rather than adding more complex logic to cause further confusing.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetadataQueryV2.java
xiaozaiyuji: > > metadataSize 、scrollsize over the elasticserch maxWindowSize
> 
> All these are open for you to change. Why don't you change them by your own?

I change the metadataSize 20000 but this param is also use by the other query ,this cause an exception.This is over the elasticserch limit 10000. I also try to change the elasticsearch maxWindowSize to 30000,but this is very low performance for the query,So I think to split the metadataSize  and maxWindowSize to by this.

wu-sheng: > In product, the metrics-all index will generate the data by day ,the service with the same name is added

No, they are not. Service would be only added once. Check the data before guessing. That is not a fact.

xiaozaiyuji: > Where is the approval of this new query API? I can't find this.



> > > And what is the reason of ElasticSearch change? A pull request must include the explanation. One pull request is about one thing. Please be clear about the change, otherwise, we don't know what do you mean, and how could we follow?
> > 
> > 
> > Sorry this the first to pull request for open source ,I explain will explain it carefully.
> 
> You are replying on another context. Please the keep thread clear.

1.This will help the query to service by group which will classify the service by group,which had many services.This will more sufficient and flexible.
2.Split the metadataSize and resultWindowMaxSize and aviod an exception when the user did not change elasticserch max size limit. And will find the query result as many as the user want to search,without reduce performance。

wu-sheng: And your explanation doesn't make the sense even for your codes. 
You added `resultWindowMaxSize`, but only `Math.min(metadataMaxSize, resultWindowMaxSize)`. Why just set a larger `metadataMaxSize`? `resultWindowMaxSize` isn't used in any place of this DAO. 
Still in all other places, `queryMaxSize`(`metadataMaxSize`) is still used. What is the point? What is changed?

wu-sheng: > 2.Split the metadataSize and resultWindowMaxSize and aviod an exception when the user did not change elasticserch max size limit. And will find the query result as many as the user want to search,without reduce performance。

I don't know how you get this conclusion. There is no `resultWindowMaxSize` used in any place other than NetworkAddressAliasEsDAO. 

image

xiaozaiyuji:
> > And what is the reason of ElasticSearch change? A pull request must include the explanation. One pull request is about one thing. Please be clear about the change, otherwise, we don't know what do you mean, and how could we follow?
> 
> Sorry,I will explain it.



> > And what is the reason of ElasticSearch change? A pull request must include the explanation. One pull request is about one thing. Please be clear about the change, otherwise, we don't know what do you mean, and how could we follow?
> 
> Sorry,I will explain it.

The metadataSize will effect queryMaxSize and will effect other query and the sroll query.But when user not change the elasticsearch max result limit size.The sroll query will not cause an exception，but the other query will fail for the exception.
So I split the two size.

wu-sheng:
I can't follow you are randomly replying on different threads. If you want to discuss it, do it in a more formal way.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetadataQueryV2.java
xiaozaiyuji: This param resultWindowMaxSize is correspond the elasticsearch max limit size 10000. What I think is wrong?I am sorry for this.But this name is very confusion.
But the aim of sroll query is to query data not limit by elasticserach  limit 10000, In fact,If I didnot to change this to 20000.The other query will cause an exception. So I split them and can help the scroll query can query 20000 result,and other not scroll query will also query the data they want by the elasticsearch index max result limit.
Did I should define a othe static param?

wu-sheng: Are you kidding? You are contributing to an open-source project, but understand a variable from the name rather than from codes.

I am going to leave this PR, with no further steps. The conclusion is very clear, you even don't read the codes and debug the codes. Just guessing.

I am out.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetadataQueryEsDAO.java
xiaozaiyuji: Thanks for dear wu,I just want to help skywalking to become better and offer an effort.I will remember that not to adding more complex logic and further confusion. Thanks a lot.

xiaozaiyuji:
> I can't follow you are randomly replying on different threads. If you want to discuss it, do it in a more formal w

> I can't follow you are randomly replying on different threads. If you want to discuss it, do it in a more formal way.

I will change the pull request and submit it soon.Sorry for this.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetadataQueryV2.java
xiaozaiyuji: I did not just guessing, I am a coder,I think this problem is obviously to find by a test.I will do an other pull request and give the errorlog and the compare.Sorry for confusion you.

PR #10568, opened 2023-03-21
Fix K8sRetag reads the wrong k8s service

innerpeacez:

Different namespace may contain services with the same name and the same label. 
Only distinguishing through labels will cause the cached service name to be wrong.

for example:
```
in namespace1
my-deployment1 -> labels: my-install
my-service -> selector: my-install

in namespace2 
my-deployment2 -> labels: my-install
my-service -> selector: my-install
```
original cached： 
```
my-service.namespace1
or
my-service.namespace2
```
now cached： 
```
my-service.namespace1
my-service.namespace2
```

- [] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>.


oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/k8s/K8sInfoRegistry.java
```diff
@@ -96,6 +98,8 @@ private K8sInfoRegistry() {
                     .filter(it -> it.getMetadata() != null)
                     .filter(it -> it.getSpec() != null)
                     .filter(it -> requireNonNull(it.getSpec()).getSelector() != null)
+                    .filter(it -> it.getMetadata().getNamespace() != null
+                            && it.getMetadata().getNamespace().equals(pod.get().getMetadata().getNamespace()))
```
wu-sheng, line 102: This should move after the next `filter` because it is also for checking `spec#selector`.

innerpeacez: Do you mean so?
image

wu-sheng: Do L103 - L106 first.

wu-sheng: I think your codes are correct, I just want to keep logic more clear.

docs/en/changes/changes.md
```diff
@@ -16,6 +16,7 @@
 * Fix otel k8s-cluster rule add namespace dimension for MAL aggregation calculation(Deployment Status,Deployment Spec Replicas)
 * Support continuous profiling feature.
 * Support collect process level related metrics.
+* Fix k8s service name cache(podServiceMap) wrong.
```
wu-sheng, line 19:
```suggestion
* Fix K8sRetag reads the wrong k8s service from the cache due to namespace mismatch.
```

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/k8s/K8sInfoRegistry.java
innerpeacez: I think to filter the namespace first, the data hit by the next filter will be less. :)

wu-sheng: Then you should move before `it -> requireNonNull(it.getSpec()).getSelector() != null`.

innerpeacez: got it

wankai123 (approved):

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/k8s/K8sInfoRegistry.java
```diff
@@ -94,6 +96,8 @@ private K8sInfoRegistry() {
                     .list()
                     .stream()
                     .filter(it -> it.getMetadata() != null)
+                    .filter(it -> it.getMetadata().getNamespace() != null
+                                && it.getMetadata().getNamespace().equals(pod.get().getMetadata().getNamespace()))
```
kezhenxu94, line 100:
```suggestion
                    .filter(it -> Objects.equals(it.getMetadata().getNamespace(), pod.get().getMetadata().getNamespace())
```

wu-sheng: I think there is `)` missing.

wu-sheng (approved):

Issue #10373, opened 2023-02-10
Support Java profiling continues when cross thread.


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Currently, the profiling targets the endpoint, and ends in current thread and segment only.
We could enhance this to make the profiling flag continues with Snapshot(capture/continue). Therefore, we could have a multi-thread profiling for one endpoint.

### Use case

Typically with more undertow use case, when it doesn't work with SpringMVC, the profiling feature is limited.
Also, for some in memory scheduleling, the current profiling doesn't fit well.

By implementing this enhancement, we should consider enhancing profiling Snapshot to bring thread or segment ID for UI to distinguish threads and merge them with different options.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

mikechengwei:
Can I try to finish this function?

wu-sheng:
@mrproliu When you and @Fine0830 finished all, and the feature is ready on demo.skywalking.a.o, please close here. 
I will run 8.15.0 agent side release.

One question, what will happen if agent 8.15 reports multiple profiled segments for the older(< 9.5.0) OAP? Is the old UI still working?

mrproliu:
> One question, what will happen if agent 8.15 reports multiple profiled segments for the older(< 9.5.0) OAP? Is the old UI still working?

In the old UI, it still works, but the name displayed in Segment will be different from the configured endpoint name in Profiling because no merging process has been performed on the displayed segment at this time.

wu-sheng:
OK. That should be fine.

wu-sheng:
https://github.com/apache/skywalking/pull/10613 has merged.

@mrproliu Is this feature ready? If so, let's bring the showcase updated to this latest feature and verify again.

PR #10575, opened 2023-03-22
Support cross thread trace profiling

mrproliu:
Following https://github.com/apache/skywalking/issues/10373, support the query of the cross thread trace profiling. 

Also, the `readNullableMetricsValue` query been added with default implementation, this should be update in other PR. 


wu-sheng:
Is this going to break UI temporarily, right?

mrproliu:
> Is this going to break UI temporarily, right?

Yes, It need the UI to adapt with the new query protocol.

wu-sheng:
FYI @apache/skywalking-committers After this gets merged, we are going to hold the demo update until we have UI side adoption. Otherwise, the profiling relative features are broken.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/resolver/MetricQuery.java
```diff
@@ -92,6 +93,14 @@ public IntValues getLinearIntValues(final MetricCondition metrics,
         return metricsValues.getValues();
     }
 
+    public NullableValue readNullableMetricsValue(MetricsCondition condition, Duration duration) {
+        // TODO default implantation
```
wu-sheng, line 97: FYI @wankai123

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/profiling/trace/ProfileTaskQueryService.java
```diff
@@ -161,38 +182,126 @@ public List<ProfileTaskLog> getProfileTaskLogs(final String taskID) throws IOExc
...
+        // build result
+        return instanceTraceWithSegments.values().stream()
+            .flatMap(s -> buildProfiledSegmentsList(s, profiledSegmentIdList).stream())
+            .collect(Collectors.toList());
+    }
 
-        ProfiledSegment profiledSegment = new ProfiledSegment();
-        SegmentObject segmentObject = SegmentObject.parseFrom(segmentRecord.getDataBinary());
-        profiledSegment.getSpans().addAll(buildProfiledSpanList(segmentObject));
+    protected List<ProfiledTraceSegments> buildProfiledSegmentsList(List<SegmentRecord> segmentRecords, List<String> profiledSegmentIdList) {
```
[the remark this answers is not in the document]
mrproliu: fixed.

wu-sheng:
@mrproliu There is a snapshot exporter tool. Do we need some update about that?

mrproliu:
> @mrproliu There is a snapshot exporter tool. Do we need some update about that?

The tools have been updated to support multiple segments. If users want to use them, they need to update them too.

wu-sheng:
Is there anything expected to change for https://skywalking.apache.org/docs/main/next/en/guides/backend-profile-export/?

Notice, the profiling docs should be moved from contribute doc to new profiling doc. https://skywalking.apache.org/docs/main/next/en/guides/readme/#profile
Your next move should be about docs, right?

wu-sheng (dismissed):

mrproliu:
> Is there anything expected to change for https://skywalking.apache.org/docs/main/next/en/guides/backend-profile-export/?

No, this file only descript how to use this export tool, and the usage has no change. 

> Your next move should be about docs, right?

Yes, I will write the documentation for cross-thread profiling and continuous profiling.

docs/en/changes/changes.md
```diff
@@ -17,6 +17,7 @@
 * Support continuous profiling feature.
 * Support collect process level related metrics.
 * Fix K8sRetag reads the wrong k8s service from the cache due to a possible namespace mismatch.
+* Support cross thread trace profiling.
```
wu-sheng, line 20:
```suggestion
* [Breaking Change] Support cross-thread trace profiling. The data structure and query APIs are changed.
```

wu-sheng (approved):

PR #10592, opened 2023-03-25
Fix jdbc storage, delete table by mistake

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. NO.
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/docs/en/changes/changes.md). No need, not released yet.

### DIFF
--- a/oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/dao/JDBCHistoryDeleteDAO.java
+++ b/oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/dao/JDBCHistoryDeleteDAO.java
@@ -75,16 +75,17 @@ public class JDBCHistoryDeleteDAO implements IHistoryDeleteDAO {
 
         final var ttlTables = tableHelper.getTablesForRead(model.getName(), startTimeBucket, endTimeBucket);
         final var tablesToDrop = new HashSet<String>();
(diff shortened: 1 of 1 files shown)


oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/dao/JDBCHistoryDeleteDAO.java
```diff
@@ -75,16 +75,17 @@ public void deleteHistory(Model model, String timeBucketColumnName, int ttl) {
...
-             final var result = conn.getMetaData().getTables(null, null, TableHelper.getTableName(model) + "%", new String[]{"TABLE"})) {
+             final var result = conn.getMetaData().getTables(null, null, tableName + "%", new String[]{"TABLE"})) {
             while (result.next()) {
                 tablesToDrop.add(result.getString("TABLE_NAME"));
             }
         }
 
         ttlTables.forEach(tablesToDrop::remove);
-        tablesToDrop.removeIf(it -> !it.matches(".*_\\d{8}$"));
+        tablesToDrop.removeIf(it -> !it.matches(tableName + "_\\d{8}$"));
```
kezhenxu94, line 88: This fixes the issue when one model name is the prefix of another, the longer one will be deleted by mistake.

Example: when model `meter_sumpermin` is deleting its outdated tables, the `meter_sumperminlabeled` will be deleted by mistake

wu-sheng (approved):

PR #10593, opened 2023-03-25
Fix invalid links from doc analysis

Superskyyy:
Doing some doc automation and some urls are not in correct form (only one is actually broken), fixing and removing them.

### DIFF
--- a/docs/menu.yml
+++ b/docs/menu.yml
@@ -46,8 +46,6 @@ catalog:
             path: "/en/concepts-and-designs/lal"
           - name: "Profiling"
             path: "/en/concepts-and-designs/profiling"
-          - name: "Query in OAP"
-            path: "/en/protocols/readme#query-protocol"
       - name: "Event"
         path: "/en/concepts-and-designs/event/"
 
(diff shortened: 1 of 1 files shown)


docs/menu.yml
```diff
@@ -80,7 +78,7 @@ catalog:
           - name: "Setup External Communication Channels"
             path: "/en/setup/backend/backend-expose"
           - name: "Kafka Fetcher"
-            path: "/en/setup/backend/Kafka-fetcher"
+            path: "/en/setup/backend/kafka-fetcher"
```
wu-sheng, line 81: This is a fix. But you should know actually this isn't an issue as website building fixes this automatically.

Check this menu https://skywalking.apache.org/docs/main/next/en/setup/backend/kafka-fetcher/

wu-sheng (approved):

docs/menu.yml
Superskyyy: Yes, I was mainly intended to fix the actual broken one for protocol, along with small fxies.

wu-sheng: @wankai123 told me this was broken. I am thinking how to write a new one, or actually we leave this part to API docs only.

PR #10594, opened 2023-03-26
Fix additional table is not created for the next day

kezhenxu94:
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/docs/en/changes/changes.md). NO need, not released yet.


oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/JDBCTableInstaller.java
```diff
@@ -86,15 +88,16 @@ public boolean isExists(Model model) {
...
     }
 
     @SneakyThrows
-    public void createTable(Model model, String table) {
+    public void createTable(Model model, long timeBucket) {
+        final var table = TableHelper.getTable(model, timeBucket);
         createOrUpdateTable(table, model.getColumns(), false);
         createOrUpdateTableIndexes(table, model.getColumns(), false);
-        createAdditionalTable(model);
+        createAdditionalTable(model, timeBucket);
```
kezhenxu94, line 100: `createAdditionalTable` always used the current day timebucket to create a new table, when the `JDBCHistoryDeleteDAO` is creating tables for next day, the additional tables won't be created

wu-sheng (approved):

PR #10597, opened 2023-03-27
Retain +1 day tables when deleting history

kezhenxu94:
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/docs/en/changes/changes.md). No need, not released yet.

Currently when deleting outdated tables, the DAO only filters the tables that are in `T - ttl`, causing the newly created table for `T + 1` is deleted, this makes the table schema for `T + 1` incomplete, for example, when 2 (or more) metrics, `metrics_1` and `metrics_2` are the same function type (`metrics_longavg`), when deleting history tables for `metrics_1`, it will delete all tables in the pattern `metrics_longavg_\d{8}` except for the `[T - ttl, T]`, and a new table `metrics_longavg_<T+1>` is created, but at the same time when deleting the same for `metrics_2`, it will delete all tables except for `[T - ttl, T]`, causing the newly created `metrics_longavg_<T+1>` be deleted, and the data columns for `metrics_1` is gone.

### DIFF
--- a/oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/TableHelper.java
+++ b/oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/TableHelper.java
@@ -85,15 +85,6 @@ public class TableHelper {
         return tableName + Const.UNDERSCORE + dayTimeBucket;
     }
 
-    /**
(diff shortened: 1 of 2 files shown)


wu-sheng:
So, after the change, are we going to delete the table older than `NOW - ttl` only. Others are skipped?

kezhenxu94:
> So, after the change, are we going to delete the table older than `NOW - ttl` only. Others are skipped?

Previously, we delete tables that are **NOT** in range `[NOW - ttl, NOW]`, now, we delete tables that are **NOT** in range of `[NOW - ttl, NOW + 1]`, that's all in this patch.

PR #10600, opened 2023-03-28
Fix T+1 table is delete when deleting history tables, in jdbc storage

kezhenxu94:

### DIFF
--- a/oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/TableHelper.java
+++ b/oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/TableHelper.java
@@ -45,6 +45,7 @@ import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.LongStream;
 
+import static java.util.function.Predicate.not;
 import static java.util.stream.Collectors.toList;
 
 /**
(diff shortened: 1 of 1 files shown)


wu-sheng:
Yesterday's fix is not working?

oap-server/server-storage-plugin/storage-jdbc-hikaricp-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/jdbc/common/TableHelper.java
```diff
@@ -124,22 +133,19 @@ public List<String> getTablesForRead(String modelName, long timeBucketStart, lon
...
 
-        final var ttlTimeBucketRange = getTTLTimeBucketRange(model);
-
         return LongStream
             .rangeClosed(timeBucketStart, timeBucketEnd)
             .distinct()
-            .filter(ttlTimeBucketRange::contains)
-            .mapToObj(it -> tableName + "_" + it)
+            .mapToObj(timeBucket -> getTable(rawTableName, timeBucket))
             .collect(toList());
```
kezhenxu94, line 149: `HistoryDeleteDAO` uses this method to filter tables that should not be deleted, it turns out this method still filter out the `T + 1` table because of `.filter(ttlTimeBucketRange::contains)`,

kezhenxu94 (commented):
turns out https://github.com/apache/skywalking/pull/10597 doesn't fix the issue it aimed to fix 🥲 after another day's testing, this patch should fix the issue

wu-sheng:
Do we need to revert https://github.com/apache/skywalking/pull/10597?

kezhenxu94:
> Do we need to revert #10597?

No, that's still needed

wu-sheng (approved):

wu-sheng (approved):

PR #10610, opened 2023-03-29
Support alarm metric OP `!=`

innerpeacez:

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -21,6 +21,7 @@
 * Fix PromQL HTTP API `/api/v1/labels` response missing `service` label.
 * Fix possible NPE when initialize `IntList`.
 * Support parse PromQL expression has empty labels in the braces for metadata query.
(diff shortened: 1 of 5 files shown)


wu-sheng:
What is the use case of `not equal`?

innerpeacez:
> What is the use case of `not equal`?

labeled values metric. like node status

wu-sheng:
> > What is the use case of `not equal`?
> 
> labeled values metric. like node status

Make sense. Please don't change the code style, I can't see it is relevant to this change.

innerpeacez:
> > > What is the use case of `not equal`?
> > 
> > 
> > labeled values metric. like node status
> 
> Make sense. Please don't change the code style, I can't see it is relevant to this change.

I think using abbreviations for operator symbols is more elegant. I feel like the abbreviation and full name are mixed up(like LESS_EQ), and "<" should be translated as "less than".

wu-sheng (approved):

PR #10613, opened 2023-03-30
Fix trace profiling segments build failure when span list not ordering

mrproliu:
When querying Trace Profiling Segments, the failure of segment construction might occur if the first span hasn't been ref. To resolve this problem, we should check the ref in the each span. 


wu-sheng (approved):
LGTM

wu-sheng (approved):

PR #10616, opened 2023-03-30
Support isEmptyValue flag for metrics query.

### LINKED ISSUE #10501: [Feature] Support isEmptyValue flag for metrics query
### Search before asking

### Description

In the current case, we always respond to zero if timestamp slots are absent.

I opened https://github.com/apache/skywalking-query-protocol/pull/110 and https://github.com/apache/skywalking-query-protocol/pull/112 for a protocol update(9.5.0 only), and we could distinguish these two cases. 

### Use case

`no value`(no metric at this time bucket) V.S. `actual zero value`(such as success rate == 0% or the value is less than 1 due to precision)

@wankai123 According to this, let's evaluate whether the default value still makes sense, or we should remove that permanently. 

FYI @Fine0830 After this lands on the backend, UI should do
1. Build a flat line cross the absent(isEmptyValue = true) data point. Rather than dropping the line to zero as a peak. 
2. Skip absent(isEmptyValue = true) data point when the table runs preview avg.

### Related issues

_No response_

### Code of Conduct

wankai123:
And Fix `NPE` when query the not exist series indexes in ElasticSearch storage.

- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #10501.


test/e2e-v2/cases/php/expected/metrics-has-value.yml
```diff
@@ -15,5 +15,11 @@
 
 {{- contains . }}
 - key: {{ notEmpty .key }}
-  value: {{ ge .value 2 }}
+  value:
+    value: 0
+    isemptyvalue: true
```
kezhenxu94, line 20: Why do we need to assert that there is empty value?

wu-sheng: I think the point is, from a test perspective, the initial query should cover the time range from no telenmetry data to having data.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/query/MetricsQueryService.java
```diff
@@ -50,7 +51,7 @@ private IMetricsQueryDAO getMetricQueryDAO() {
     /**
      * Read metrics single value in the duration of required metrics
      */
-    public long readMetricsValue(MetricsCondition condition, Duration duration) throws IOException {
+    public NullableValue readMetricsValue(MetricsCondition condition, Duration duration) throws IOException {
```
wu-sheng, line 54: I think we don't change this API. Somehow we don't tested this?

We just added a new API --> **readNullableMetricsValue**

wu-sheng (approved):
LGTM

PR #10628, opened 2023-03-31
PromQL: Remove empty values from the query result, fix `/api/v1/metad…ata` param `limit` could cause out of bound.

wankai123:
And add e2e for PromQL service


wu-sheng (approved):
Once the e2e passed, I am good.

PR #10632, opened 2023-03-31
Support sumLabeled in MAL

BFergerson:
Added support for the `sumLabeled` function in MAL. The implementation is based off of `SumPerMinLabeledFunction` and the test is based off of `AvgLabeledFunctionTest`.


wu-sheng:
Is there any specific use case for this new function? We only should maintain the function we used. If you need this in a fork, you could re-build the package to involve this new in.

PR #10643, opened 2023-04-04
Fix wrong calculation in history delete dao

kezhenxu94:


wu-sheng (dismissed):
Once e2e passed, this should be good.

wu-sheng:
There are still bugs under investigation.

wu-sheng (approved):

PR #10645, opened 2023-04-04
Support Amazon API Gateway monitoring

pg-yang:

image

image

image

image

Maybe we could close https://github.com/apache/skywalking/issues/9883 ?


wu-sheng:
> Maybe we could close https://github.com/apache/skywalking/issues/9883 ?

Yes, we should.

___

In your screenshot, there is an endpoint called `GET /students`. Is this your route rule?

oap-server/server-starter/src/main/resources/application.yml
```diff
@@ -332,7 +332,7 @@ receiver-otel:
   selector: ${SW_OTEL_RECEIVER:default}
   default:
     enabledHandlers: ${SW_OTEL_RECEIVER_ENABLED_HANDLERS:"oc,otlp"}
-    enabledOtelRules: ${SW_OTEL_RECEIVER_ENABLED_OTEL_RULES:"apisix,k8s/*,istio-controlplane,vm,mysql/*,postgresql/*,oap,aws-eks/*,windows,aws-s3/*,aws-dynamodb/*"}
+    enabledOtelRules: ${SW_OTEL_RECEIVER_ENABLED_OTEL_RULES:"aws-gateway/*"}
```
wu-sheng, line 335: I think this is changed by mistakes, which would cause e2e failing.

oap-server/analyzer/meter-analyzer/src/main/java/org/apache/skywalking/oap/meter/analyzer/dsl/SampleFamily.java
```diff
@@ -464,7 +464,7 @@ public SampleFamily instance(List<String> serviceKeys, List<String> instanceKeys
         return instance(serviceKeys, Const.POINT, instanceKeys, Const.POINT, layer, null);
     }
 
-    public SampleFamily endpoint(List<String> serviceKeys, List<String> endpointKeys, Layer layer) {
+    public SampleFamily endpoint(List<String> serviceKeys, List<String> endpointKeys, String delimiter, Layer layer) {
```
wu-sheng, line 467: Could you explain a little more about this new delimiter parameter? What happens?

pg-yang: As mentioned above,  `router` is made by `Method(route method)` and `Resource(route path)` such as `GET /students`.
`endpoint(['service_name'],['Method','Resource'], Layer.AWS_GATEWAY)`  will generate an endpoint that contains a dot by default, such as `GET./students`. It's better to use space to replace dot to display.

Alternative, ` tags.endpoint_name= tags.Method+' '+tags.Resource` also could generate an endpoint with a space delimiter.

docs/menu.yml
```diff
@@ -205,6 +205,8 @@ catalog:
                 path: "/en/setup/backend/backend-aws-s3-monitoring"
               - name: "AWS DynamoDB"
                 path: "/en/setup/backend/backend-aws-dynamodb-monitoring"
+              - name: "AWS API Gateway"
+                path: "/en/setup/backend/backend-aws-api-gateway-monitoring"
```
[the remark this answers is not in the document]
pg-yang: Done

oap-server/server-starter/src/main/resources/application.yml
pg-yang: Reverted it

wu-sheng:
The CI would be blocked until https://github.com/apache/skywalking/pull/10643 is done.

wu-sheng (approved):
LGTM

kezhenxu94 (approved):

wankai123 (approved):

Issue #10520, opened 2023-03-09
[Feature] Support no-proxy mode for aws-firehose receiver


wu-sheng:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

As upstream https://github.com/line/armeria/issues/4681 has been fixed, we could wait for next armeria release, and update our receiver setup to make proxy optional.

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #10684, opened 2023-04-13
Support no-proxy mode for aws-firehose receiver

pg-yang:

Close https://github.com/apache/skywalking/issues/10520


wu-sheng:
Please fix license versions.

docs/en/setup/backend/aws-firehose-receiver.md
```diff
@@ -32,5 +32,7 @@ The following blogs demonstrate complete setup process for AWS S3 and API Gatewa
 ## Notice
 
 1. Only OpenTelemetry format is supported (refer to [Metric streams output formats](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-metric-streams-formats.html))
-2. A proxy(e.g. Nginx, Envoy) is required in front of OAP's Firehose receiver to accept HTTPS requests from AWS Firehose through port `443` (refer to [Amazon Kinesis Data Firehose Delivery Stream HTTP Endpoint Delivery Specifications](https://docs.aws.amazon.com/firehose/latest/dev/httpdeliveryrequestresponse.html).
+2. According to HTTPS requirement by AWS Firehose(refer to [Amazon Kinesis Data Firehose Delivery Stream HTTP Endpoint Delivery Specifications](https://docs.aws.amazon.com/firehose/latest/dev/httpdeliveryrequestresponse.html), users have two options
+  - A proxy(e.g. Nginx, Envoy) is required in front of OAP's Firehose receiver to accept HTTPS requests from AWS Firehose through port `443`. (Recommended based on the general security policy)
+  - Set `aws-firehose/enableTLS=true` and `acceptProxyRequest=true` at OAP side to accept requests from firehose directly.
```
wu-sheng, line 37: @pg-yang I reworded this, please take a look

wu-sheng:
@pg-yang `.licenserc.yaml` should be updated, as versions of two manually set lib licenses have been changed.

docs/en/setup/backend/aws-firehose-receiver.md
kezhenxu94, line 37: I think we can enable `acceptProxyRequest=true` always and remove this config to minimize the configurations, users don't need to set this extra config. It has no side effect right?

Also for the tls setting, we should also mention the cert/key files settings here.

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServer.java
```diff
@@ -119,4 +123,14 @@ public void addHandler(Object handler, List<HttpMethod> httpMethods) {
     public void start() {
         sb.build().start().join();
     }
+
+    private String transformAbsoluteURI(final String uri) {
+        if (uri.startsWith("https://")) {
+            return uri.substring(uri.indexOf("/", 9));
```
wu-sheng, line 129: Why `9` of `uri.indexOf("/", 9)`? Length of `https://` is 8 only. What do I miss?

docs/en/setup/backend/aws-firehose-receiver.md
wu-sheng: Updated.

wu-sheng: > I think we can enable acceptProxyRequest=true always and remove this config to minimize the configurations, users don't need to set this extra config. It has no side effect right?

About this, I think `acceptProxyRequest=true` is harmless from our use case. @pg-yang Could you recheck about this?

pg-yang: Consider a case, users use a gateway to accept TCP over TLS(participate in TLS handshake), and then transport TCP without TSL to a firehose receiver, the configuration will be `enableTLS =false,acceptProxyRequest=true`. I know the case is uncommon.

wu-sheng: This seems never recommended by us, I think.
This configuration is not straight forward, we knew this only we read source codes of America. I don't think users would know.

oap-server/server-library/library-server/src/main/java/org/apache/skywalking/oap/server/library/server/http/HTTPServer.java
pg-yang: `"https://".length=8`
`uri.indexOf("/", 9)` means that find the index of '/' from `"https://".length+1`

pg-yang: OK, I should find index of next `/` begin with `"https://".lengt` rather than `"https://".lengt+1`. It's a mistake.

docs/en/setup/backend/aws-firehose-receiver.md
pg-yang: acceptProxyRequest will have same value with enableTLS: 

https://github.com/apache/skywalking/pull/10684/commits/07f3de50f18edae8e3c789839a73f4722c435974#diff-8f7b4b497c144f8ccbc29c6ccfa29d06281ee8eeab86ccc03979e6223bba9f6eR78

wu-sheng (approved):

PR #10704, opened 2023-04-19
UI templates: Simplify metric name with the label.

wankai123:
Include: general, mesh, self-observe, browser,k8s, linux 


docs/en/changes/changes.md
```diff
@@ -52,6 +52,7 @@
 * Add elasticsearch menu.
 * [Clean UI templates before upgrade] Set `showSymbol: true`, and make the data point shows on the Line graph.
   Please clean `ui_template` index in elasticsearch storage or table in JDBC storage.
+* UI templates: Simplify metric name with the label.
```
wu-sheng, line 55:
```suggestion
* [Clean UI templates before upgrade] UI templates: Simplify metric name with the label.
```

I think this should follow the same log format as the previous one.

wu-sheng (approved):

wu-sheng (approved):

PR #10801, opened 2023-05-15
Filter out unknown_cluster metric data in `elasticsearch-monitoring`

innerpeacez:

### DIFF
--- a/oap-server/server-starter/src/main/resources/otel-rules/elasticsearch/elasticsearch-cluster.yaml
+++ b/oap-server/server-starter/src/main/resources/otel-rules/elasticsearch/elasticsearch-cluster.yaml
@@ -34,39 +34,39 @@ metricPrefix: meter_elasticsearch_cluster
 metricsRules:
   # cluster health
   - name: health_status
-    exp: elasticsearch_cluster_health_status.valueEqual(1).sum(['cluster' , 'color'])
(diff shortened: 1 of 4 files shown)


oap-server/server-starter/src/main/resources/otel-rules/elasticsearch/elasticsearch-cluster.yaml
```diff
@@ -34,39 +34,39 @@ metricPrefix: meter_elasticsearch_cluster
 metricsRules:
   # cluster health
   - name: health_status
-    exp: elasticsearch_cluster_health_status.valueEqual(1).sum(['cluster' , 'color'])
+    exp: elasticsearch_cluster_health_status.tagNotEqual('cluster','unknown_cluster').valueEqual(1).sum(['cluster' , 'color'])
```
wu-sheng, line 37: Attaching the source codes of the exporter, just FYI to others

image

wu-sheng (approved):

oap-server/server-starter/src/main/resources/otel-rules/elasticsearch/elasticsearch-cluster.yaml
wu-sheng: This `shouldn't` thing keeps happening in our demo env. We have reasons to believe this is still a case.
Adding this to remove illegal data.

PR #10808, opened 2023-05-15
Add OpenTelemetry log protocol support

### LINKED ISSUE #10752: [Feature] Support OTLP Native Logging
### Search before asking

### Description

SkyWalking has OTLP log support through SkyWalking exporter, ref https://skywalking.apache.org/docs/main/next/en/setup/backend/log-otlp/

We should begin to consider whether we could have gRPC/HTTP native OTLP logging support as well.

### Use case

It should work similarly to the MAL engine. Various formats of logging would be transferred as the native format and pushed into the LAL engine.

### Related issues

_No response_

### Code of Conduct

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #10752 .


oap-server/server-starter/src/main/resources/application.yml
```diff
@@ -331,8 +331,8 @@ receiver-meter:
 receiver-otel:
   selector: ${SW_OTEL_RECEIVER:default}
   default:
-    enabledHandlers: ${SW_OTEL_RECEIVER_ENABLED_HANDLERS:"otlp"}
-    enabledOtelRules: ${SW_OTEL_RECEIVER_ENABLED_OTEL_RULES:"apisix,k8s/*,istio-controlplane,vm,mysql/*,postgresql/*,oap,aws-eks/*,windows,aws-s3/*,aws-dynamodb/*,aws-gateway/*,redis/*,elasticsearch/*"}
+    enabledHandlers: ${SW_OTEL_RECEIVER_ENABLED_HANDLERS:"otlp-metrics,otlp-logs"}
+    enabledOtelMetricsRules: ${SW_OTEL_RECEIVER_ENABLED_OTEL_RULES:"apisix,k8s/*,istio-controlplane,vm,mysql/*,postgresql/*,oap,aws-eks/*,windows,aws-s3/*,aws-dynamodb/*,aws-gateway/*,redis/*,elasticsearch/*"}
```
wu-sheng, line 335: Should we update `SW_OTEL_RECEIVER_ENABLED_OTEL_RULES` as well and add breaking change log?

kezhenxu94: Sure will add that

wu-sheng:
`log-otlp.md` should be updated according, especially about 
1. Highlighting the OTLP native logging is supported.
2. OLTP SkyWalking log exporter is optional for now.

docs/en/setup/backend/log-otlp.md
```diff
@@ -1,4 +1,46 @@
 # OpenTelemetry Logging Format
 
-OpenTelemetry Collector supports to transport the logs to SkyWalking OAP.
-Read the doc on [Skywalking Exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/skywalkingexporter) for a detailed guide.
+SkyWalking can receive logs exported from OpenTelemetry collector, the data flow is:
+
+```mermaid
+graph LR
+    B[OpenTelemetry SDK 1]
```
wu-sheng, line 7: I think this doesn't have to be OTEL SDK? OTEL collector has capabilities to transfer the log format accordingly, right?

kezhenxu94: Polished the graph

wu-sheng (approved):
LGTM

docs/en/setup/backend/log-otlp.md
```diff
@@ -1,4 +1,46 @@
...
+the `receiver-otel/default/enabledHandlers` configuration section:
+
+```yaml
+receiver-otel:
+  selector: ${SW_OTEL_RECEIVER:default}
+  default:
+    enabledHandlers: ${SW_OTEL_RECEIVER_ENABLED_HANDLERS:"otlp-metrics,otlp-logs"}
+```
+
+Also, because most of the language SDKs of OpenTelemetry do not support logging feature (yet) or the logging feature is
```
kezhenxu94, line 41: As said, many of the language SDKs don't support logging feature yet or are experimental stage now, we will need more improvements in the future when they support logging

wu-sheng: We don't have to work with OTEL SDKs/Agents anyway. Abou that OTEL, we are always only adopting as much as possible no matter where the source is from and how the format is built.
There are always a group of people who would think the OTEL is somehow a standard or much more meaningful than the other formats. 
But from my personal perspective and experiences, there is nothing new there. The metrics format is from Prometheus, the Trace format is from Zipkin, and logs are easy generally text only.

All the points of this kind of support are, we could leverage some existing integrations, which the OTEL community would do. That is the only benefit.

Don't worry whether their agents are supporting or not. The lessons I learned from OT are, these APIs would have special values when they are stale(not updated anymore). Ironically, but true.

PR #10832, opened 2023-05-19
Fix typo in backend-meter doc

maclong1989:
The word 'Sprign' in document [docs/en/setup/backend/backend-meter.md] is wrong.It is shoud be 'Spring'.so changed  'Sprign' to 'Spring'.


wu-sheng:
Please merge all typo fix into one PR.

PR #10833, opened 2023-05-19
Fix typo in dynamic-config docs

maclong1989:
Fix some typos in dynamic-config docs.the word 'Vaule' is wrong.it is shoud be 'Value'.so fixed it.


wu-sheng (approved):

wu-sheng (changes requested):
Please add https://github.com/apache/skywalking/pull/10832 into this PR as well.

wu-sheng (changes requested):
Please add https://github.com/apache/skywalking/pull/10832 into this PR as well.

wu-sheng (approved):

wu-sheng (approved):

PR #10840, opened 2023-05-21
Support composite ID to replace entity ID in BanyanDB case.

wu-sheng:

* Support composite ID declared from existing columns, rather than a new virtual column, in the BanyanDB.
* In the BanyanDB implementation, metrics built from ServiceInstance and Endpoint sources are not using `entity_id` as ID,
  instead, they use composite ID built by `service_id` and `name` to reduce redundant indices in the database.

@lujiajing1126 This is a request from Gao according to his perf test benchmark. I made most parts of the implementation, and hope the idea is clear enough for both of you to continue. 

I am marking this as a draft PR, as it is not tested. These should be expected
* In all other storages, the column and logic should be unchanged, entity_id is still treated as the major ID.
* BanyanDB storage would not generate entity IDs for metrics built from ServiceInstance and Endpoint.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/StorageCharacter.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/StorageCharacter.java
@@ -1,0 +1,27 @@
+package org.apache.skywalking.oap.server.core.storage;
+
+import org.apache.skywalking.oap.server.core.source.ScopeDefaultColumn;
+import org.apache.skywalking.oap.server.library.module.Service;
+
+/**
+ * StorageCharacter provides core aware characters which make the core could run optimized codes accordingly.
+ *
+ * @since 9.5.0
+ */
+public interface StorageCharacter extends Service {
+    /**
+     * See {@link ScopeDefaultColumn.DefinedByField#idxOfCompositeID()}
+     *
+     * @return true if ID is declared through existing column, but not a virtual column. Typically, there was an entity_id
+     * column to represent a subject(service, endpoint, et.c)
+     */
+    boolean supportCompositeID();
+
+    class Default implements StorageCharacter {
+
+        @Override
+        public boolean supportCompositeID() {
+            return false;
+        }
+    }
+}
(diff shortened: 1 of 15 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/StorageCharacter.java
```diff
@@ -0,0 +1,27 @@
...
+ * @since 9.5.0
+ */
+public interface StorageCharacter extends Service {
+    /**
+     * See {@link ScopeDefaultColumn.DefinedByField#idxOfCompositeID()}
+     *
+     * @return true if ID is declared through existing column, but not a virtual column. Typically, there was an entity_id
+     * column to represent a subject(service, endpoint, et.c)
+     */
+    boolean supportCompositeID();
```
wu-sheng, line 18: This new interface and method are how the core module is aware of storage preference. I am adding this interface to mark as an extension point for further in cases of most DB-specific optimization.

cc @kezhenxu94 @wankai123

wu-sheng:
From the testing, we noticed that removing `entity_id` permanently in the composite ID scenario is very complex.
For now, we are going to support composite IDs by reusing entity_id.

Here is the generated `endpoint_percentile`(from Endpoint source) metric.
```java
@Stream(
    name = "endpoint_percentile",
    scopeId = 3,
    builder = EndpointPercentileMetricsBuilder.class,
    processor = MetricsStreamProcessor.class
)
public class EndpointPercentileMetrics extends PercentileMetrics implements WithMetadata {
    @Column(
        name = "entity_id",
        length = 512
    )
    @SeriesID(
        index = 1
    )
    private String entityId;
    @Column(
        name = "service_id",
        length = 256
    )
    @SeriesID(
        index = 0
    )
    @TopNAggregation
    private String serviced;
```

wu-sheng:
Close for now. We had a discussion, the conclusion is, BanyanDB supports the sparse index for columns like  `service_id`, which link to an entity(without timestamp) rather than linking to a specific record(with timestamp).

Issue #10796, opened 2023-05-12
[Bug] ping msg will recover the property from register msg  in ServiceManagementHandler.java


lubberland:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar issues.

### Apache SkyWalking Component

OAP server (apache/skywalking)

### What happened

ServiceManagementHandler serviceReportProperties(InstanceProperties request)  and keepAlive(InstancePingPkg request)
 both use ServiceInstanceUpdate to update service_instance
but ping msg does not contain property 
so when handled keepAlive(), we lose property from serviceReportProperties()

### What you expected to happen

i do not know how to solve it

### How to reproduce

i do not know how to solve it

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wu-sheng:
AFAIK, no it wouldn't. Instance properties are only overriding, rather than replacing.

lubberland:
InstanceTraffic.java  entity2Storage()
```
        public void entity2Storage(final InstanceTraffic storageData, final Convert2Storage converter) {
            converter.accept(SERVICE_ID, storageData.getServiceId());
            converter.accept(NAME, storageData.getName());
            if (storageData.getProperties() != null) {
                converter.accept(PROPERTIES, GSON.toJson(storageData.getProperties()));
            } else {
                converter.accept(PROPERTIES, Const.EMPTY_STRING);
            }
            converter.accept(LAST_PING_TIME_BUCKET, storageData.getLastPingTimestamp());
            converter.accept(TIME_BUCKET, storageData.getTimeBucket());
        }
```
when ping msg,   
converter.accept(PROPERTIES, Const.EMPTY_STRING);
 i can not find properties in es  😭

wu-sheng:
InstanceTraffic would call `combine(final Metrics metrics)` first, then flushing. Unless your storage has some issues in query, or OAP running in a improperly cluster setup.

lubberland:
java agent sends register msg  every 5 minutes  and sends ping msg every 30 seconds

does combine() work ok for properties?

wu-sheng:
That doesn't matter at all. You could debug codes on your own, I don't think there is black magic there.

lubberland:
when there is no sessionCache for InstanceTraffic
[...]

PR #10841, opened 2023-05-22
keepAlive msg have to load from database, fectch the properties when …

lubberland:
…there is no sessionCache

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/worker/MetricsPersistentWorker.java
@@ -30,6 +30,7 @@ import org.apache.skywalking.oap.server.core.analysis.DownSampling;
 import org.apache.skywalking.oap.server.core.analysis.TimeBucket;
 import org.apache.skywalking.oap.server.core.analysis.data.MergableBufferedData;
 import org.apache.skywalking.oap.server.core.analysis.data.ReadWriteSafeCache;
+import org.apache.skywalking.oap.server.core.analysis.manual.instance.InstanceTraffic;
 import org.apache.skywalking.oap.server.core.analysis.metrics.Metrics;
 import org.apache.skywalking.oap.server.core.exporter.ExportEvent;
 import org.apache.skywalking.oap.server.core.status.BootingStatus;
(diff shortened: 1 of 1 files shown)


wu-sheng:
I am not following. The data should either cached with latest or loaded.

lubberland:
> I am not following. The data should either cached with latest or loaded.

https://github.com/apache/skywalking/issues/10796#issuecomment-1556611342
please see this link

lubberland:
when there is no sessionCache for InstanceTraffic
1 、the skywalking service just started
2、the sessionCache was expired
then get a keepalive metrics from http interface or kafka consumer
```
private Metrics requireInitialization(Metrics metrics) {
        final Metrics cached = sessionCache.get(metrics);

        // All cached metrics, it at least had been written once.
        if (cached != null) {
            return cached;
        }

        // If the metrics do not have time bucket relative ID
        // it is treated as high dimensionality metrics.
        // Policy, always try load from the database when miss in the cache.
        if (!model.isTimeRelativeID()) {
            return null;
        }

        // When
        // (1) the time bucket of the server's latest stability status is provided
        //     1.1 the OAP has booted successfully
        //     1.2 the current dimensionality is in minute
        //     1.3 the OAP cluster is rebalanced due to scaling
        // (2) the metrics are from the time after the timeOfLatestStabilitySts
        // (3) the metrics don't exist in the cache
        // the kernel should NOT try to load it from the database.
        //
        // Notice, about condition (2),
        // for the specific minute of booted successfully, the metrics are expected to load from database when
        // it doesn't exist in the cache.
        if (timeOfLatestStabilitySts > 0 &&
            metrics.getTimeBucket() > timeOfLatestStabilitySts
            && cached == null) {
            // Return metrics as input to avoid reading from database.
            return metrics;
        }

        return null;
    }
```

requireInitialization will return metrics,  and do not load  from storage, do not call combine() ,then call  prepareBatchInsert() ,  then the properties is null

this is the pr https://github.com/apache/skywalking/pull/10841  for simple fix

wu-sheng:
I saw that, but properties are reported periodically too. There may be several minutes latency, but eventually, it would be overrided back to normal.
We don't know there would be properties or nor, we don't want to risk reloading everytime if therw is no property for some agents.

lubberland:
1、in java-agent, the kafka producter send keepalive and register msg , use the different key, so the two msg will be consumed by different consumer.
2、for http interface, we use  a domain name (nginx )for the skywalking cluster,  after the balancer,  the two msg may be handled by different skywalking service.
so the properties is often null
when "it would be overrided back to normal.", after one or two minutes ,  the properties will by null again.

i agree with "We don't know there would be properties or nor, we don't want to risk reloading everytime if therw is no property for some agents."   
 the agent writed by unofficial  is uncontrollable

lubberland:
thank you !   I will check the cluster config . Thank you very much for guidance

PR #10854, opened 2023-05-24
Fix format in es and rabbitmq monitor docs

smartboy37597:
In `backend-elasticsearch-monitoring.md` and `backend-rabbitmq-monitoring.md` document, several format issues have been modified.


wu-sheng:
I don't think these have any issue. Do you mean the blank?

Superskyyy:
> I don't think these have any issue. Do you mean the blank?

I believe it's the blanks and Chinese commas instead of English version.

wu-sheng (approved):

PR #10859, opened 2023-05-26
fix rabbitmq cluster metrics doc

innerpeacez:


wu-sheng (approved):
It seems name changed?IHow this happens?

innerpeacez:
> It seems name changed?IHow this happens?

just forgot to modify :(

wu-sheng:
So, UI and dashboard settings are correct, just docs somehow copied the name from other docs?

PR #10874, opened 2023-06-01
Fix data loss when query continuous profiling task record

mrproliu:
Because Continuous Profiling did not set TimeBucket when reporting the task, the data was deleted by TTL. So I set the TimeBucket field to fix it. 


mrproliu:
Let me close this PR and create a new one with the new design.

Issue #10875, opened 2023-06-02
[Feature] Add the maven plugin maven-compiler-plugin to specify the build version


weihubeats:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

The maven.compiler.source property defined in the properties of the skywalking pom does not seem to be used, we can add the maven-compiler-plugin plugin to specify the compiled version to avoid compilers such as idea reading the default compiled version

![image](https://github.com/apache/skywalking/assets/42484192/37533b35-2c07-4f28-bd16-52f05ff121c5)

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wu-sheng:
What is the issue? I am not following the case.

weihubeats:
Some of our code using jdk 11 syntax, for some local installation of jdk8 and jdk11 people, the project if you do not specify the use of 11 to compile, then idea may use the default jdk8 to compile thus leading to compile errors, if you want to solve this problem may need to manually change the idea in the compiled jdk version, due to many modules so the need to change the module is very much, so we can not consider the introduction of maven-compiler-plugin this plug-in to specify the compiled jdk version, there will be no such problems

Translated with www.DeepL.com/Translator (free version)

wu-sheng:
@kezhenxu94 Don't we have this?

kezhenxu94:
> The maven.compiler.source property defined in the properties of the skywalking pom does not seem to be used,

This is a convention naming in maven and this property will be automatically picked up by the maven-compiler-plugin.

> we can add the maven-compiler-plugin plugin to specify the compiled version to avoid compilers such as idea reading the default compiled version

As said above, you don't need to explicitly add maven-compiler-plugin, it's there, and the property name `maven.compiler.source` is automatically picked by it the plugin.

I remember IntelliJ also automatically change the JDK version based on the property `maven.compiler.source` so you actually don't need to do anything, it just works.

weihubeats:
The source code I downloaded was compiled with 1.8 by default, and when I added `maven-compiler-plugin` it updated to 11 for me. `maven.compiler.sources` didn't seem to work

kezhenxu94:
[...]

PR #10876, opened 2023-06-02
add maven-compiler-plugin

weihubeats:
Add the maven plugin maven-compiler-plugin to specify the build version

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<https://github.com/apache/skywalking/issues/10875>.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -149,6 +149,7 @@
         <!-- Built in properties -->
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
+        <maven-compiler-plugin.version>3.5.1</maven-compiler-plugin.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
         <powermock.version>2.0.9</powermock.version>
@@ -520,6 +521,17 @@
                         ${project.build.directory}/generated-sources/protobuf/grpc-java;
                         ${project.build.directory}/generated-sources/antlr4;;
                     </sourcepath>
+                </configuration>
+            </plugin>
+            <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>${maven-compiler-plugin.version}</version>
+                <configuration>
+                    <source>11</source>
+                    <target>11</target>
+                    <compilerVersion>11</compilerVersion>
+                    <showDeprecation>true</showDeprecation>
+                    <showWarnings>true</showWarnings>
                 </configuration>
             </plugin>
         </plugins>
--- a/pom.xml
+++ b/pom.xml
@@ -149,7 +149,7 @@
         <!-- Built in properties -->
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
-        <maven-compiler-plugin.version>3.5.1</maven-compiler-plugin.version>
+        <maven-compiler-plugin.version>3.11.0</maven-compiler-plugin.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -149,6 +149,7 @@
         <!-- Built in properties -->
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
+        <maven-compiler-plugin.version>3.5.1</maven-compiler-plugin.version>
```
kezhenxu94, line 152: What about using a more recent version? 

```suggestion
        <maven-compiler-plugin.version>3.11.0</maven-compiler-plugin.version>
```

image

pom.xml
```diff
@@ -522,6 +523,17 @@
                     </sourcepath>
                 </configuration>
             </plugin>
+            <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>${maven-compiler-plugin.version}</version>
+                <configuration>
+                    <source>11</source>
+                    <target>11</target>
+                    <compilerVersion>11</compilerVersion>
```
kezhenxu94, line 532: We don't need these, they are defined in the `properties`, is removing these ok in your case?

```suggestion
```

weihubeats: If we use `maven.compiler.source` he will read the jdk version of the maven setting, if we want to define variables we need to define different variable names

kezhenxu94: @weihubeats read this https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#source and https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#target

> <source>
The -source argument for the Java compiler.
Type: java.lang.String
Required: No
User Property: maven.compiler.source // <<< 👈 this one
Default: 1.8


the equivalent property is `maven.compiler.source` and we already defined it in the `<properties>` section, so I mean we don't need to set `<source>/<target>/<compilerVersion>` here

weihubeats: `maven.compiler.source` doesn't work, we need to define

weihubeats: My test defaults to reading the jdk version in maven settings

wu-sheng: At least using `maven.compiler.source` and `maven.compiler.target`, rather than hard coded.

kezhenxu94: > `maven.compiler.source` doesn't work, we need to define

This doesn't make sense to me, it's clearly documented in maven's documentation, the property `maven.compiler.source` is equivalent to the `<source>` in `<configuration`

wu-sheng: We need to verify why it doesn't work before apply any change.

weihubeats: I know, but if we use the `maven.compiler.source` variable name he will still read the jdk version of the maven setting, so we need to define a new variable

wu-sheng: I am afraid we need to follow doc comes first, especially it works on others' env.

weihubeats: We change the variable name `maven.compiler.source` to `skywalking.maven.compiler.source`?

wu-sheng: We need why for any change

weihubeats: Because using `${maven.compiler.source}` he will not read the version defined in the project, but will still read the jdk version in maven setttings.xml, so we need to define a different variable

weihubeats: Or define a new variable for `maven-compiler-plugin` to use

lujiajing1126: > I know, but if we use the `maven.compiler.source` variable name he will still read the jdk version of the maven setting, so we need to define a new variable

Then it would be a bug of JetBrains. If so, you should raise a ticket in their support system rather than manually fix configurations in SkyWalking.

PR #10882, opened 2023-06-06
Bump up zipkin ui dependency

kezhenxu94:


wu-sheng:
query-protocol is reverted by accident. Please fix.

wu-sheng (approved):

PR #10884, opened 2023-06-06
Add `ai-pipeline` module and support HTTP URI formatting

wu-sheng:

* Add `ai-pipeline` module.
* Support HTTP URI formatting through `ai-pipeline` to do pattern reorganization.

@Superskyyy This is what I am going to propose. The sync would run in 30 mins period. 
Several things to highlight.
1. This is a replacement for Regex grouping and is the recommended alternative option of OpenAPI.
2. Any 3rd party AI or algorithm server could be called. I **will** add a gRPC service for shipping this to the remote.
3. The size of per service cache is 3000 URIs.

The context we would provide for the remote AI server is the service name and endpoints with the matched count. If the count of some URIs is more than one, those could be patterns already.


wu-sheng:
@Superskyyy Are you possible to build a gRPC server to do URI pattern recognization in weeks? We don't need complex features or high performance. This is pure async and only 3000 URIs per service, and 30 mins as a period for all services.
I am looking forward to see this as a good start.

oap-server/ai-pipeline/src/main/java/org/apache/skywalking/oap/server/ai/pipeline/services/HttpUriRecognitionService.java
```diff
@@ -0,0 +1,29 @@
...
+
+package org.apache.skywalking.oap.server.ai.pipeline.services;
+
+import java.util.List;
+import org.apache.skywalking.oap.server.ai.pipeline.services.api.HttpUriRecognition;
+
+public class HttpUriRecognitionService implements HttpUriRecognition {
+    @Override
+    public void recognize(final String serviceId, final List<HTTPUri> unrecognizedURIs, final Callback callback) {
+
```
wu-sheng, line 27: This is TODO, to build the connection between OAP and the remote server.

oap-server/ai-pipeline/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+    </parent>
+
+    <groupId>org.apache.skywalking</groupId>
+    <artifactId>ai-pipeline</artifactId>
+
+    <description>
+        AI pipeline module include pipelines connecting the OAP kernel and any 3rd party AI and/or algorithm
+        server. The kernel would call the pipeline in the specific context/scenario with well formatted telemetry data,
+        and expect the remote to respond with the result of the AI/algorithm as quick as possible, but in async way in
+        case of large size of data or lower performance behavior of the remote.
```
wu-sheng, line 38: @jcchavezs This is the responsibility of the pipeline. Bundle the logic of assembling AI/algorithm required data into OAP kernel, and provide a clear path for those data.

docs/en/changes/changes.md
```diff
@@ -51,6 +51,8 @@
 * Support Metrics Query Expression(MQE) and allows users to do simple query-stage calculation through the expression.
 * Deprecated metrics query v2 protocol.
 * Deprecated record query protocol.
+* Add `ai-pipeline` module.
+* Support HTTP URI formatting through `ai-pipeline` to do pattern reorganization.
```
BFergerson, line 55:
```suggestion
* Support HTTP URI formatting through `ai-pipeline` to do pattern recognition.
```

Superskyyy:
Hi! I will offer two algorithm choices for future implementation references: First I will do the tree version as it's easier and quicker, expect to have a working algorithm in a week (will provide a simple web interface to test it).

1. My own tree based algorithm
2. LLM based response using [Langchain](https://github.com/hwchase17/langchain)

Input is like this, output will be structured according to proto.
```
cachedHttpUris: ConcurrentHashMap
|
|-- Service Name (String)
|   |-- URI (String) : Occurrence Count (AtomicInteger)
|   |-- URI (String) : Occurrence Count (AtomicInteger)
|
|-- Service Name (String)
|   |-- URI (String) : Occurrence Count (AtomicInteger)
|

Output:

Something like this

{
	"FormattedPattern1": [URI, URI, URI...],   # FormattedPattern is not regex, but logical endpoint. 
						   # Algorithm generated regex is unreliable, it will introduce ambiguity, so I don't recommend update rules with algorithm, any new uri that unhandled by openapi and user regex should be always sent to algorithm
	"FormattedPattern2": [URI, URI, URI...],
	"FormattedPattern3": [URI, URI, URI...],
}
```

Considerations of the algorithm:

It's going to be stateful, based on incremental trees. Meaning across different batches (at most 3k per 30mins) the algorithm will remember previously grouped uris (assumption is that it will always end up to be a finite number), so results is consistent throughout the service lifecycle. LLM based response is stateless by nature, so previous uri results will be kept in a local cache and passed back to each gpt question in later batches to simulate the stateful nature.


[I have a comment on the regex pattern group added back to ruleset, please see in pr code comment]

Superskyyy:
Since the endpoint matching patterns/results are generated in async, is there a mechanism to update the endpoints after results come back from ai module?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/config/group/EndpointNameGrouping.java
```diff
@@ -69,6 +113,40 @@ private String formatByOpenapi(String serviceName, String endpointName) {
...
+                                                   ))
+                                               .collect(Collectors.toList());
+                                 // Reset the cache once the URIs are sent to the recognition server.
+                                 httpUris.clear();
+                                 httpUriRecognitionSvr
+                                     .recognize(serviceName, uris,
+                                                (service, patterns) -> {
+                                                    StringFormatGroup group = new StringFormatGroup(patterns.size());
+                                                    patterns.forEach(
+                                                        p -> group.addRule(p.getFormattedUri(), p.getPattern()));
```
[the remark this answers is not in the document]
Superskyyy: I'm not sure yet but there might be cases where the regex rule can change (algorithm doesn't know before hand what future uris will look like).

For example: /order/123/refund -> in future there might be something called /order/134b/refund -> I will just assume them to be two patterns to simplify algorithm and avoid updating exsiting rules.

The generated pattern will always start from minimal scope, if theres only INT between / /, it will be only int, not a var. this is to prevent algorithm sort everything into a general regex and nothing makes sense anymore.

wu-sheng:
> Hi! I will wrie two algorithm choices for future implementation references: 
> 
> First I will do the tree version as it's easier and more feasible for users, expect to have a working algorithm in **a week** (will provide a simple web interface to test it).
> 
> 
> 
> 1. My own tree based algorithm (priority)
> 
> 2. LLM based response using [Langchain](https://github.com/hwchase17/langchain) 
> 
> 
> 
> Input is like this, output will be structured according to proto.
> 
> ```
> 
> cachedHttpUris: ConcurrentHashMap
> 
> |
> 
> |-- Service Name (String)
> 
> |   |-- URI (String) : Occurrence Count (AtomicInteger)
> 
> |   |-- URI (String) : Occurrence Count (AtomicInteger)
> 
> |
> 
> |-- Service Name (String)
> 
> |   |-- URI (String) : Occurrence Count (AtomicInteger)
> 
> |
> 
> 
> 
> Output:
> 
> 
> 
> Something like this
> 
> 
> 
> {
> 
> 	"FormattedEndpoint1": {"regex": "pattern", "uris": [URI, URI, URI...]},  # < Regex starts from the smallest scope possible?
> 
> 	"FormattedEndpoint2": {"regex": "pattern", "uris": [URI, URI, URI...]}, 
> 
> 	"FormattedEndpoint3": {"regex": "pattern", "uris": [URI, URI, URI...]}
> 
> }
> 
> ```
> 
> 
> 
> Considerations of the algorithm:
> 
> 
> 
> It's going to be stateful, based on incremental trees. Meaning across different batches (at most 3k per 30mins) the algorithm will remember previously grouped uris (assumption is that it will always end up to be a finite number), so results is consistent throughout the service lifecycle. LLM based response is stateless by nature, so previous uri results will be kept in a local cache and passed back to each gpt question in later batches to simulate the stateful nature.
> 
> 

I am feeling you are not reading codes correctly. OAP will provide you all URIs as INPUT, and expect yo get the regex to format with formatted URI.

wu-sheng:
Your server wouldn't do the match ever, OAP will do that internally.

Superskyyy:
@wu-sheng Thats correct. I think we actually are talking about the same thing. 

I don't think I misunderstood the process, the matching in algorithm is inherent in such clustering problems and is extremely performant. It takes in the endpoints. It outputs patterns and regex to the endpoints given.

oap-server/ai-pipeline/src/main/java/org/apache/skywalking/oap/server/ai/pipeline/services/HttpUriRecognitionService.java
wu-sheng: @Superskyyy This is for the whole process. `serviceId` and `unrecognizedURIs` are the raw data for you. `callback` will return the patterns to the kernel. Nothing more.

I feel you are being held by your old design. I am repeating myself again, runtime matching(including further metrics baseline calculation for alerting) is not an AI part responsibility.

wu-sheng:
@Superskyyy What concerns me is that you seem to respond for all raw data. That seems a little overload. Those are good logs of the AI server side, but it seems not useful for the OAP.

Superskyyy:
After a quick discussion offline, we have arrived at a common understanding. The algorithm output will not include urls in the reponse to OAP since it doesn't bring any actual value. Only pattern and regex is returned.

In the debug logging of the POC it will be still included as optional logging. As well as an fully experimental feature that allow simply pass PatternIDs between oap -> ai algorithm. Only when an update is needed for previous regex a new proto message will be sent back to oap.

oap-server/ai-pipeline/src/main/proto/ai_http_uri_recognition.proto
```diff
@@ -0,0 +1,50 @@
...
+
+syntax = "proto3";
+
+option java_multiple_files = true;
+option java_package = "org.apache.skywalking.oap.server.ai.pipeline.grpc";
+
+service HttpUriRecognitionService {
+    rpc fetchAllPatterns(HttpUriRecognitionSyncRequest) returns (HttpUriRecognitionResponse) {}
+    rpc recognize(stream HttpUriRecognitionRequest) returns (HttpUriRecognitionResponse) {}
+}
```
wu-sheng, line 27: @Superskyyy These two services, 
- `fetchAllPatterns` for syncing existing patterns. This would be called during OAP booting and call periodically every minute. No server-side operation, just asking for the existing patterns.
- `recognize` is for shipping the raw URIs and expecting to get a response soon. 

**To be discussed**, @Superskyyy @wankai123  considering the frequency of fetchAllPatterns is high(every minute). Should we consider two things
1. Remove HttpUriRecognitionResponse of `recognize`, as the next fetch/sync would come soon enough.
2. Would the AI server side provide some kind of pattern candidate dataset ID? The OAP could hold that in the memory,  and only expect the response of updated patterns if the AI server has new identified results. @Superskyyy Is this possible?

Superskyyy: 1 sounds very reasonable, a simple acknowledge will be enough and fully rely on fetchall for the patterns. 

For 2, could you please explain a bit more on what is the "pattern candidate dataset ID" with an example? Is it the PatternID <-> (ActualPattern, ActualPatternRegex) mapping I previously mentioned?

oap-server/ai-pipeline/src/main/proto/ai_http_uri_recognition.proto
```diff
@@ -0,0 +1,53 @@
...
+option java_package = "org.apache.skywalking.oap.server.ai.pipeline.grpc";
+
+service HttpUriRecognitionService {
+    rpc fetchAllPatterns(HttpUriRecognitionSyncRequest) returns (HttpUriRecognitionResponse) {}
+    rpc feedRawData(HttpUriRecognitionRequest) returns (NoResponse) {}
+}
+
+message HttpUriRecognitionSyncRequest {
+    string service = 1;
+}
```
wu-sheng, line 31: If https://github.com/apache/skywalking/pull/10884#discussion_r1221205604 is able to support, then we could add version ID here and in `HttpUriRecognitionResponse`.
Then AI server doesn't need to respond all patterns if the version is the latest.

wu-sheng:
One more thing, `AI/ML extendion` doc would come after the PR is considered ready.

oap-server/ai-pipeline/src/main/proto/ai_http_uri_recognition.proto
Superskyyy: Yes this is possible. A version number can be maintained on algorithm side.

Superskyyy:
Would query parameters like (?q=keyword&page=12) appear in the endpoints?

wu-sheng:
> Would query parameters like (?q=keyword&page=12) appear in the endpoints?

No, it wouldn't. These parameters would be converted as parameters of request automatically according to HTTP standard AFAIK

kezhenxu94 (commented):
just did a quick review and I'm wondering whether the module `ai-pipeline` is suitable for the current architecture.

We have been organize the modules in the perspective of functionality so that users can plug/unplug features easily, but the module `ai-pipeline` seems to be a module that addresses the technology (AI) we use, and from the API I assume if we want to add other AI-powered features (for example, log outlier detection), all of them would be added inside this module too (if not, which module should it be in?), which might mess things up.

this PR adds a URI pattern recognition API and limit the implementations to be AI-powered (at least all the names `AIPipeline....`), but it can actually be any implementation that are nothing related t o AI, e.g. I can even just implement the APIs and perform a naive regular expression match and returns the pattern.


## Summary

From my point of view, I think we can do the following:

1. the new module can be moved/renamed into a feature-related module, such as `uri-grouping-plugin`,
1. we don't have to address the technology (AI) we want to use, we can just provide the API and document that we have an implementation powered by AI. (reminds me the difference between **HAS A** vs. **IS A**)

oap-server/ai-pipeline/src/main/proto/ai_http_uri_recognition.proto
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+syntax = "proto3";
+
+option java_multiple_files = true;
+option java_package = "org.apache.skywalking.oap.server.ai.pipeline.grpc";
+
+service HttpUriRecognitionService {
+    rpc fetchAllPatterns(HttpUriRecognitionSyncRequest) returns (HttpUriRecognitionResponse) {}
+    rpc feedRawData(HttpUriRecognitionRequest) returns (NoResponse) {}
```
kezhenxu94, line 26: Usually, it should be better if this is a `stream`ing request?

oap-server/ai-pipeline/src/main/proto/ai_http_uri_recognition.proto
```diff
@@ -0,0 +1,53 @@
...
+    repeated HTTPRawUri unrecognizedURIs = 2;
+}
+
+message HTTPRawUri {
+    string name = 1;
+    int64 matchedCounter = 2;
+}
+
+message NoResponse {
+}
```
kezhenxu94, line 44: We can simply use `google.protobuf.Empty`, the built-in representation for "empty" gRPC message.

oap-server/ai-pipeline/src/main/proto/ai_http_uri_recognition.proto
wu-sheng: > Usually, it should be better if this is a streaming request?

It could, but I was wondering, as a 3000 - 4000 string in the same cluster, is that complexity really necessary?

oap-server/ai-pipeline/src/main/proto/ai_http_uri_recognition.proto
```diff
@@ -0,0 +1,53 @@
...
+message HttpUriRecognitionSyncRequest {
+    string service = 1;
+}
+
+message HttpUriRecognitionRequest {
+    string service = 1;
+    repeated HTTPRawUri unrecognizedURIs = 2;
+}
+
+message HTTPRawUri {
```
kezhenxu94, line 38: `HTTPRawUri` the name has two styles in it, `HTTP` and `Uri`, maybe `HttpRawUri` or `HTTPRawURI`?

wu-sheng:
> this PR adds a URI pattern recognition API and limit the implementations to be AI-powered (at least all the names AIPipeline....), but it can actually be any implementation that are nothing related t o AI, e.g. I can even just implement the APIs and perform a naive regular expression match and returns the pattern.

First, let's begin with this. AI-powered doesn't mean AI could do things without context. We are not limiting everything, but AFAIK, certain contexts provide better output from every algorithm/model we have known.
The idea had been put in [here](https://github.com/apache/skywalking/discussions/10872), especially **Ideas** part.

____
Then back to your summary

> the new module can be moved/renamed into a feature-related module, such as uri-grouping-plugin,

It is switchable, and it doesn't have to run as the service is only up and running when upstream(server addr) has been set. I just don't want the kernel logic so context. Read **EndpointNameGrouping#L131**, `if (!httpUriRecognitionSvr.isInitialized())`, nothing happens.

> we don't have to address the technology (AI) we want to use, we can just provide the API and document that we have an implementation powered by AI. (reminds me the difference between HAS A vs. IS A)

We don't as I mentioned. And it is definitely pluggable to use different servers to bind with a simple LLM pattern recognition or OpenAPI commercial API.

BFergerson:
I also find the naming of `ai-pipeline` a bit confusing and think this should be renamed to `uri-grouping-plugin`. I'm just as eager to see SkyWalking add AI/GPT technology, but the code currently drafted only exposes endpoint URIs to alternative grouping algorithms. I get that AI/GPT can be one of those grouping algorithms, but that doesn't make this any more of a "AI pipeline".

wu-sheng:
> I also find the naming of `ai-pipeline` a bit confusing and think this should be renamed to `uri-grouping-plugin`. I'm just as eager to see SkyWalking add AI/GPT technology, but the code currently drafted only exposes endpoint URIs to alternative grouping algorithms. I get that AI/GPT can be one of those grouping algorithms, but that doesn't make this any more of a "AI pipeline".

More things would be added to that module. Such as alerting threshold baseline calculation. But let's make one step per time.
The algorithm would run by the server side from @Superskyyy. This PR only gets merged we at least one integration.
Again, AI/ML doesn't have any magic. It just supplies kinds of pattern recognition. URI is as simple, just much more straightforward than other cases.

wu-sheng:
`pipeline` means this is a channel for AI/ML engine. Nothing more, nothing less. Same as the pipeline from oil well to oil refinery.

BFergerson:
Doesn't the config start to look a bit weird when other AI/GPT logic is added (i.e. Alerting Baseline Calculation)?

Wouldn't the config look like this:
```yaml
ai-pipeline:
  selector: ${SW_AI_PIPELINE:default}
  default:
    uriRecognitionServerAddr: ${SW_AI_PIPELINE_URI_RECOGNITION_SERVER_ADDR:0.0.0.0}
    uriRecognitionServerPort: ${SW_AI_PIPELINE_URI_RECOGNITION_SERVER_PORT:17128} 
    alertingBaselineServerAddr: ${SW_AI_PIPELINE_ALERTING_BASELINE_SERVER_ADDR:0.0.0.0}
    alertingBaselineServerPort: ${SW_AI_PIPELINE_ALERTING_BAEELINE_SERVER_PORT:17128} 
```

If so, that config design is different than all the other config designs, isn't it?

wu-sheng:
It would be like this. Because actually, you need that context as input when you put it into GPT modules. Such as for URI recognition, you want to get the pattern. But for alerting, you want to find the peak and reasonable range from historical data.

BFergerson:
The fact that different AI extensions require different configurations suggests a more modularized design is necessary. Isn't it odd you have to configure URI recognition and alerting baseline just to use one or the other?

wu-sheng:
I am not 100% following your concern. Correct me if I misunderstand your point, please.
Every modularized design has to build boundaries. If I follow URI and Alerting are two modules in Maven, it is totally possible. But, is that necessary to keep every less than 10 classes?

The idea of sharing the maven module is that, we could share the gRPC client with various gRPC services because generally, we are not sending the data to real AI engine. An extra layer of adapter is expected.

oap-server/ai-pipeline/src/main/proto/ai_http_uri_recognition.proto
wu-sheng: Fixed.

oap-server/ai-pipeline/src/main/proto/ai_http_uri_recognition.proto
wu-sheng: Fixed.

oap-server/ai-pipeline/src/main/proto/ai_http_uri_recognition.proto
```diff
@@ -0,0 +1,63 @@
...
+    string name = 1;
+    int64 matchedCounter = 2;
+}
+
+message HttpUriRecognitionResponse {
+    repeated Pattern patterns = 1;
+    // The version of pattern dictionary at the AI-server side.
+    // If it is as same as HttpUriRecognitionSyncRequest#version,
+    // patterns could be empty as nothing should be updated.
+    string version = 2;
```
wu-sheng, line 55: This version and HttpUriRecognitionSyncRequest#version were my proposed. @Superskyyy I hope this is good enough for you.

oap-server/ai-pipeline/src/main/proto/ai_http_uri_recognition.proto
```diff
@@ -0,0 +1,63 @@
...
+}
+
+message Pattern {
+    // Quick match URI pattern. {var} represents a variable part in the URI.
+    // /product/{var} is the pattern for /product/1, /product/2
+    // GET:/product/{var}/detail is the pattern for GET:/product/1/detail, GET:/product/2/detail
+    string pattern = 2;
```
wu-sheng, line 62: @Superskyyy This is a quick pattern tree by our new implementation.

oap-server/microbench/src/main/java/org/apache/skywalking/oap/server/microbench/core/config/group/uri/RegexVSQuickMatchBenchmark.java
```diff
@@ -0,0 +1,166 @@
...
+ * # VM invoker: /Users/wusheng/Library/Java/JavaVirtualMachines/temurin-11.0.18/Contents/Home/bin/java
+ * # VM options: -ea --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED -Didea.test.cyclic.buffer.size=1048576 -javaagent:/Applications/IntelliJ IDEA.app/Contents/lib/idea_rt.jar=53714:/Applications/IntelliJ IDEA.app/Contents/bin -Dfile.encoding=UTF-8 -Xmx512m -Xms512m -XX:MaxDirectMemorySize=512m -XX:BiasedLockingStartupDelay=0 -Djmh.executor=CUSTOM -Djmh.executor.class=org.apache.skywalking.oap.server.microbench.base.AbstractMicrobenchmark$JmhThreadExecutor
+ * # Warmup: 1 iterations, 10 s each
+ * # Measurement: 1 iterations, 10 s each
+ * # Timeout: 10 min per iteration
+ * # Threads: 4 threads, will synchronize iterations
+ * # Benchmark mode: Throughput, ops/time
+ * # Benchmark: org.apache.skywalking.oap.server.microbench.core.config.group.uri.RegexVSQuickMatchBenchmark.matchFirstQuickUriGrouping
+ *
+ Benchmark                                                                            Mode  Cnt         Score   Error   Units
```
wu-sheng, line 129: @Superskyyy This is the temp benchmark for new kinds of grouping.

@wankai123 You had known this, do you think it is worth we replace the regex permanently? From a code perspective, it is fine to keep both.
For now, this new mode doesn't work from the config file or/and dynamic configuration. 
It only works for AI/ML responses.

docs/en/changes/changes.md
```diff
@@ -52,6 +52,10 @@
 * Deprecated metrics query v2 protocol.
 * Deprecated record query protocol.
 * Add OpenSearch 2.8.0 to test case.
+* Add `ai-pipeline` module.
+* Support HTTP URI formatting through `ai-pipeline` to do pattern recognition.
+* Add new HTTP URI grouping engine with benchmark. 
+* [Breaking Change] Use the new HTTP URI grouping engine to replace the old regex based mechanism.
```
wu-sheng, line 58: FYI @apache/skywalking-committers breaking change in this PR.

Considering a better performance of endpoint grouping, regex-based implementation has been replaced.

wu-sheng:
@BFergerson @kezhenxu94 I am totally open to a new style of configuration. I am just keeping as simple as possible for now. Changing this config file structure should be easy to reflect. I would flag this feature as `alpha`(for the first time in the SkyWalking doc) to warn the breaking change.

wu-sheng:
@Superskyyy Let me know when you could test and verify. Could you host the codes in SkyAPM? I could create a repository for you. What name do you prefer?

Superskyyy:
Update: Algorithm part is on going, now testing on 100 endpoints with diverse styles shows promising results.
image

Some extra considerations and testing are needed, so I'm gonna say it will be ready in few more days.

Superskyyy (dismissed):
LGTM

wu-sheng:
The pattern recognition would be implemented in https://github.com/SkyAPM/r3 as a quick start and easy integration.

oap-server/ai-pipeline/src/main/proto/ai_http_uri_recognition.proto
kezhenxu94: > This version and HttpUriRecognitionSyncRequest#version were my proposed. @Superskyyy I hope this is good enough for you.

Does it make sense to use `string` as the type of `version`? I'd expect it to be a number so that we can compare (conveniently), tho you might claim we can compare strings too.

wu-sheng: I just don't want to limit the server. Number(int) is always better but requires comparability, may be too much?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/config/group/EndpointNameGrouping.java
```diff
@@ -18,32 +18,91 @@
...
+
+        if (!formattedName._2() && quickUriGroupingRule != null) {
+            formattedName = formatByQuickUriPattern(serviceName, endpointName);
         }
 
-        if (endpointGroupingRule != null) {
-            formattedName = formatByCustom(serviceName, formattedName);
+        if (!formattedName._2()) {
+            // Only URI starts with '/' will be cached and formatted later.
+            if (endpointName.indexOf("/") > -1) {
```
kezhenxu94, line 86: `endpointName.indexOf("/") > -1` doesn't match the comment `Only URI starts with '/' will be cached and formatted later.`

```suggestion
            // Only URI starts with '/' will be cached and formatted later.
            if (endpointName.startsWith("/")) {
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/config/group/EndpointNameGrouping.java
```diff
@@ -69,6 +128,77 @@ private String formatByOpenapi(String serviceName, String endpointName) {
...
+        this.maxHttpUrisNumberPerService = maxEndpointCandidatePerSvr;
+        if (!httpUriRecognitionSvr.isInitialized()) {
+            return;
+        }
+        this.quickUriGroupingRule = new QuickUriGroupingRule();
+        Executors.newSingleThreadScheduledExecutor()
+                 .scheduleWithFixedDelay(
+                     new RunnableWithExceptionProtection(
+                         () -> {
+                             if (aiPipelineExecutionCounter.incrementAndGet() % 25 == 0) {
```
kezhenxu94, line 161:
```suggestion
                             if (aiPipelineExecutionCounter.incrementAndGet() % 30 == 0) {
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/config/group/uri/quickmatch/PatternTree.java
```diff
@@ -0,0 +1,105 @@
...
+                    matched = true;
+                    break;
+                }
+            }
+            if (!matched) {
+                return new StringFormatGroup.FormatResult(false, uri, null);
+            }
+            current = matchedToken.children();
+        }
+        if (matchedToken.isLeaf()) {
```
kezhenxu94, line 99: `matchedToken` might be null

```suggestion
        if (matchedToken != null && matchedToken.isLeaf()) {
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/config/group/uri/quickmatch/PatternTree.java
```diff
@@ -0,0 +1,105 @@
...
+        if (tokens.length == 1) {
+            current.setExpression(pattern);
+            return;
+        }
+
+        for (int i = 1; i < tokens.length; i++) {
+            final String token = tokens[i];
+            PatternToken newToken;
+            if (VarToken.VAR_TOKEN.equals(token)) {
+                newToken = new VarToken();
```
kezhenxu94, line 67: Correct me if I am wrong, looks like we can have a singleton for `VarToken` and don't have to recreate `VarToken` every time.

docs/en/setup/ai-pipeline/README.md
```diff
@@ -0,0 +1,22 @@
...
+Pattern Recognition, Machine Learning(ML) and Artificial Intelligence(AI) are common technology to identify patterns in data. 
+This module provides a way to integrate these technologies in a standardized way about shipping the data from OAP kernel
+to 3rd party.
+
+From the industry practice, Pattern Recognition, Machine Learning(ML) and Artificial Intelligence(AI) are always overestimated,
+they are good at many things but have to run in a clear context.
+
+## How to enable
+
+ai-pipeline module is required to activate manually through `application.yml` or system environment variable `SW_AI_PIPELINE`.
```
kezhenxu94, line 14: This doesn't match the config in `application.yaml`, it's activated there by default

kezhenxu94:
Looks like github is having issue, some review comments I posted are gone after I refresh

docs/en/setup/ai-pipeline/README.md
```diff
@@ -0,0 +1,22 @@
...
+## How to enable
+
+ai-pipeline module is required to activate manually through `application.yml` or system environment variable `SW_AI_PIPELINE`. 
+
+```yaml
+ai-pipeline:
+  selector: ${SW_AI_PIPELINE:-}
+  default:
+    uriRecognitionServerAddr: ${SW_AI_PIPELINE_URI_RECOGNITION_SERVER_ADDR:0.0.0.0}
+    uriRecognitionServerPort: ${SW_AI_PIPELINE_URI_RECOGNITION_SERVER_PORT:17128}
```
kezhenxu94, line 21: What's the rationale of using port `17128`? If there is no specific reason to use this port, I'd suggest using a port near to `11800 ~ 12800`, it's not a good practice to occupy a too wide port range for a system

oap-server/ai-pipeline/src/main/proto/ai_http_uri_recognition.proto
kezhenxu94: > I just don't want to limit the server. Number(int) is always better but requires comparability

Actually I think this is useful in client side, consider this, the client sends `fetchAllPatterns` periodically and, if, for any reasons, the requests are not responded in the same order they are sent, you might update the content to an older version, example:

client send request 1, and 1 minute later, send a request 2, for any reasons, response 2 comes back to client and then later response 1, the client (OAP) will consider response 1 is latest, which is not the case.

This might be a rare case, but I just think it's a low hanging fruit that we can avoid easily by using a number version

docs/en/setup/ai-pipeline/README.md
wu-sheng: It is remote(skyapm r3 server address).

oap-server/ai-pipeline/src/main/proto/ai_http_uri_recognition.proto
wu-sheng: Is this kind of increasinh version number possible in training?
Especially, remote could be distributed deployment Maintaining this is very hard, remember SkyWalking numberic endpoint ID? Failing bad.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/config/group/EndpointNameGrouping.java
wu-sheng: Let me fix codes. start with is wrong. Get and Post are prefix, not /. That was a legacy comment.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/config/group/uri/quickmatch/PatternTree.java
wu-sheng: That works.

wu-sheng: It doesn't. VarToken is stateful too, because of `VarToken#expression`. It carries the expression to represent as a leaf in the tree.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/config/group/EndpointNameGrouping.java
wu-sheng: Fixed

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/config/group/EndpointNameGrouping.java
wu-sheng: Fixed

docs/en/setup/ai-pipeline/README.md
wu-sheng: Fixed

oap-server/ai-pipeline/src/main/proto/ai_http_uri_recognition.proto
Superskyyy: > Is this kind of increasing version number possible in training?
> Especially, remote could be distributed deployment Maintaining this is very hard, remember SkyWalking numberic endpoint ID? Failing bad.

Yes, each service can maintain a version number. Whenever a new pattern is discovered or old pattern changes in the batch, the version number will += 1.

Superskyyy: > Especially, remote could be distributed deployment Maintaining this is very hard, remember SkyWalking numberic endpoint ID? Failing bad.

Are you saying maintaining the version counter in a distributed deployment of the algorithm? If that's the case, the algorithm scales by splitting in units of services, so they will always be only one modifier of each version number concurrently.

wu-sheng: I am just saying, to keep monotone increasing in a distributed service is not easy. And considering the number of patterns, this seems to not have to limit from protocol perspective.
If we want to make this optimization, r3 could do directly. r3 compares and responds when the version is older rather than different.

wu-sheng: Anyway, we don't have to do this for now. Let's keep in MVP and make things running.

wu-sheng:
@Superskyyy with new updates, could you approve again?

Superskyyy (approved):

Superskyyy (approved):

PR #10905, opened 2023-06-08
Consider 4xx as client side error for envoy ALS based metrics

toffentoffen:

When metrics are generated from ALS envoy entries, `ServiceRelation` and `ServiceInstanceRelation` **client side** generated metrics should consider 4xx status codes as error (`status` false).


oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/LogEntry2MetricsAdapter.java
```diff
@@ -81,10 +81,16 @@ public HTTPServiceMeshMetric.Builder adaptToUpstreamMetrics() {
...
+        final HTTPServiceMeshMetric.Builder builder = adaptCommonPart();
+        // For client side call, status needs to be overridden because 4xx http status codes
+        // are considered as errors to.
+        final boolean status = builder.getResponseCode() < 400;
+
         return adaptCommonPart()
             .setStartTime(outboundStartTime)
             .setEndTime(outboundEndTime)
             .setLatency((int) Math.max(1L, outboundEndTime - outboundStartTime))
+            .setStatus(status)
```
wu-sheng, line 93: The status was not set? If so, the current status should be all `false`.
I am feeling there is a later place overriding this?

kezhenxu94: > The status was not set? If so, the current status should be all `false`.
> I am feeling there is a later place overriding this?

It was actually set in `adaptCommonPart`, and here is to "override"

wu-sheng: OK, got it. Then this should be fine.

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/LogEntry2MetricsAdapter.java
```diff
@@ -81,10 +81,16 @@ public HTTPServiceMeshMetric.Builder adaptToUpstreamMetrics() {
         final long outboundStartTime = startTime + formatAsLong(properties.getTimeToFirstUpstreamTxByte());
         final long outboundEndTime = startTime + formatAsLong(properties.getTimeToLastUpstreamRxByte());
 
+        final HTTPServiceMeshMetric.Builder builder = adaptCommonPart();
+        // For client side call, status needs to be overridden because 4xx http status codes
+        // are considered as errors to.
+        final boolean status = builder.getResponseCode() < 400;
+
         return adaptCommonPart()
```
kezhenxu94, line 89: Can we share the two `adaptCommonPart()`s?

```suggestion
        final HTTPServiceMeshMetric.Builder builder = adaptCommonPart();
        // For client side call, status needs to be overridden because 4xx http status codes
        // are considered as errors to.
        final boolean status = builder.getResponseCode() < 400;

        return builder
```

oap-server/server-receiver-plugin/envoy-metrics-receiver-plugin/src/main/java/org/apache/skywalking/oap/server/receiver/envoy/als/LogEntry2MetricsAdapter.java
```diff
@@ -81,10 +81,16 @@ public HTTPServiceMeshMetric.Builder adaptToUpstreamMetrics() {
         final long outboundStartTime = startTime + formatAsLong(properties.getTimeToFirstUpstreamTxByte());
         final long outboundEndTime = startTime + formatAsLong(properties.getTimeToLastUpstreamRxByte());
 
-        return adaptCommonPart()
+        final HTTPServiceMeshMetric.Builder builder = adaptCommonPart();
+        // For client side call, status needs to be overridden because 4xx http status codes
+        // are considered as errors to.
```
kezhenxu94, line 86: nit:

```suggestion
        // are considered as errors too.
```

kezhenxu94 (dismissed):

kezhenxu94 (approved):

wu-sheng (approved):
I just resolved the conflicts.

PR #10913, opened 2023-06-12
Apply MQE on General and Virtual-Database layer UI-templates.

wankai123:
1. Sync UI module https://github.com/apache/skywalking-booster-ui/pull/279
2. When MQE operation single-type result value is null, will return a null value item in the results list rather than return an empty result list.


wu-sheng (approved):
LGTM

PR #10917, opened 2023-06-12
Migrate from kubernetes-client/java to fabric8 client

kezhenxu94:


kezhenxu94:
I'm going to add some improvements in ALS receiver by mapping Istio ServiceEntries IP to service, and found fabric8's Kubernetes client has Istio client extension, but official Kubernetes client has no such client, I don't think we want to re-implement the Istio Java client by ourselves, so this PR migrates from kubernetes official java client to fabric8's.

I also found the APIs in fabrics8 client is fluent and more friendly...

wu-sheng:
> so this PR migrates from kubernetes official java client to fabric8's

Does this mean the k8s java client is going to be removed from our dependency tree?

kezhenxu94:
> > so this PR migrates from kubernetes official java client to fabric8's
> 
> 
> 
> Does this mean the k8s java client is going to be removed from our dependency tree?

Yes

wu-sheng (approved):
FYI @wankai123

PR #10980, opened 2023-06-22
Fix `NPE` in metrics query when the metric is not exist.

wankai123:


wu-sheng (dismissed):
LGTM. Do you plan to fix the dashboard in another PR?

wu-sheng (dismissed):

wu-sheng (approved):

PR #10981, opened 2023-06-23
Clean up tests for components that are too old

kezhenxu94:


wu-sheng:
Please send a warning to the mail list to be achieved and in case of objection officially.

wu-sheng (approved):

PR #11001, opened 2023-06-26
refactor: rename recevier to receiver

BFergerson:
I was expecting the following to work

```gradle
implementation("org.apache.skywalking:skywalking-meter-receiver-plugin:9.5.0")
implementation("org.apache.skywalking:skywalking-jvm-receiver-plugin:9.5.0")
implementation("org.apache.skywalking:skywalking-log-receiver-plugin:9.5.0")
```

But fails to download the `skywalking-log-receiver-plugin` dependencies as the artifact has been incorrectly named `skywalking-log-recevier-plugin`.

This PR also addresses some Java packages with the same misspelling.


wu-sheng (approved):

PR #11005, opened 2023-06-28
Support dynamic UI menu query

mrproliu:
Add the feature of dynamic UI menu query, similar to the UI Template feature. By predefining the UI menu, it's automatically written into the database when the system starts. Associate menu items with Dashboard ID, **when there is data in the Service dimension of the Layer corresponding to the Dashboard, the menu is deemed to need to be displayed.**


docs/en/setup/backend/configuration-vocabulary.md
```diff
@@ -48,6 +48,7 @@ The Configuration Vocabulary lists all available configurations provided by `app
...
+| -                       | -             | uiMenuFetchInterval                                                                                                                                                      | The UI menu should activate fetch interval(in seconds).                                                                                                                                                                                                                                                                                                                                                                                                                    | -                                                     | 20                                                                                           |
```
wu-sheng, line 51:
```suggestion
| -                       | -             | uiMenuRefreshInterval                                                                                                                                                      | The period(in seconds) of refreshing the status of all UI menu items.                                                                                                                                                                                                                                                                                                                                                                                                                    | -                                                     | 20                                                                                           |
```

oap-server/server-starter/src/main/resources/ui-initialized-templates/menu.yaml
```diff
@@ -0,0 +1,147 @@
...
+# menus: the menus of the left navigation bar
+#   - name: the name of the menu item
+#     icon: icon name
+#     path: link path
+#     dashboard: related dashboard name, the menu item should be activated when the dashboard value is empty or dashboard related scope has services
+#     menus: sub menus of the current item
+
+menus:
+  - name: General Service
+    icon: chart
```
wu-sheng, line 25: Where are the definitions of these icons? Should UI implement this part?

oap-server/server-starter/src/main/resources/ui-initialized-templates/menu.yaml
```diff
@@ -0,0 +1,147 @@
...
+    menus:
+      - name: Services
+        path: /general
+        dashboard: General-Root
```
wu-sheng, line 29: Are you going to use the dashboard name? So, the dashboard ID would be irrelevant

oap-server/server-starter/src/main/resources/ui-initialized-templates/menu.yaml
mrproliu: Yes, these value is the current dashboard icon name.

oap-server/server-starter/src/main/resources/ui-initialized-templates/menu.yaml
mrproliu: It's dashboard ID, I needs to using the dashboard ID to query the dashboard configuration.

oap-server/server-starter/src/main/resources/ui-initialized-templates/menu.yaml
wu-sheng: They seem in low readability. Please rename them. @Fine0830 Please reflect the UI icon names to adopt this PR change.

oap-server/server-starter/src/main/resources/ui-initialized-templates/menu.yaml
wu-sheng: If that is the ID, let's name it as ID.

wu-sheng: This doc should be updated as well, https://skywalking.apache.org/docs/main/next/en/ui/readme/. Add menu change introduction.

oap-server/server-starter/src/main/resources/ui-initialized-templates/menu.yaml
mrproliu: So, I need to wait for @Fine0830 updates, then I update the names? If It's not defined there, the UI doesn't know what kind of icon should be showing.

wu-sheng: You should update first. UI follows only.

docs/en/ui/README.md
```diff
@@ -8,6 +8,41 @@ The left side menu lists all available supported stacks with default dashboards.
 
 Follow the `Official Dashboards` menu to explore all default dashboards on their ways to monitor different tech stacks.
 
+## Sidebar Menu and Marketplace
+
+All available feature menu items are only listed in the marketplace(since 9.6.0). They are only visible on the Sidebar Menu when there are relative services
+being observed by various supported observation agents, such as installed language agents, service mesh platform, OTEL integration.
+
+The menu items defined in `ui-initialized-templates/menu.yaml` are the universal marketplace for all default-supported integration.
+The menu definition supports one and two levels items. The leave menu item should have the `path` and `dashboardID` for navigation, the icon of the level one item is defined from UI side.
```
wu-sheng, line 17: @Fine0830 Please check this, whether `path` is necessary?

wu-sheng: Why doesn't the level one(browser) menu need this, but level two does?

wu-sheng:
@mrproliu Please recheck the bot scanning results.

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/management/ui/menu/UIMenu.java
```diff
@@ -0,0 +1,77 @@
...
+@Stream(name = UIMenu.INDEX_NAME, scopeId = UI_MENU, builder = UIMenu.Builder.class, processor = ManagementStreamProcessor.class)
+@EqualsAndHashCode(of = "menuId", callSuper = false)
+public class UIMenu extends ManagementData {
+    public static final String INDEX_NAME = "ui_menu";
+    public static final String MENU_ID = "menu_id";
+    public static final String CONFIGURATION = "configuration";
+    public static final String UPDATE_TIME = "update_time";
+
+    @Column(name = MENU_ID)
+    private String menuId;
```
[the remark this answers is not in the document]
mrproliu: @sonatype-lift ignoreall

docs/en/ui/README.md
Fine0830: It seems that the `name` is necessary, and others fields are not necessary.

docs/en/ui/README.md
```diff
@@ -8,6 +8,41 @@ The left side menu lists all available supported stacks with default dashboards.
...
+menus:
+  - name: General Service
+    icon: general_service
+    menus:
+      - name: Services
+        path: /general
+        dashboardID: General-Root
+      - name: Visual Database
+        path: /database
+        dashboardID: Virtual-Database-Root
```
wu-sheng, line 29: According to the discussion between @Fine0830 and me. We need to change this.
`name`, `icon`, and `layer` are required, others are not.
This is the runtime menu context.
image

path should be generated by UI side directly by `lower case` the `name`, and use `_` replacing the blank between words.

Also, this means the query protocol should be fixed as well, no `path`, but need to add `layer`.
https://github.com/apache/skywalking-query-protocol/blob/1f80cd0af8bc0baf71bb68961a894c110c9d9b79/ui-configuration.graphqls#L32

Fine0830: I think the `name` should not contain special characters, including spaces.

mrproliu: > I think the `name` should not contain special characters, including spaces.

So, in the 'Control Plane' of 'Service Mesh', the menu should be 'ControlPlane' and 'ServiceMesh'? Am I right?

Fine0830: Right.

wu-sheng: > I think the `name` should not contain special characters, including spaces.

Then how the name is showing? We need `Service Mesh` text to display, that is my meaning of the name.

wu-sheng: > When menu item is alerting or dashboard, the value of layer should be empty?

Alerting and all menus are not configurable are excluded from this. Those would be hard coded only from UI side.

mrproliu: > > When menu item is alerting or dashboard, the value of layer should be empty?
> 
> 
> 
> Alerting and all menus are not configurable are excluded from this. Those would be hard coded only from UI side.

Got it. It should be including dashboard settings, alert and settings.

Fine0830: > > I think the `name` should not contain special characters, including spaces.
> 
> Then how the name is showing? We need `Service Mesh` text to display, that is my meaning of the name.

In the graph, the `name` means ID.  `title` could display on the page.

docs/en/ui/README.md
```diff
@@ -8,6 +8,41 @@ The left side menu lists all available supported stacks with default dashboards.
 
 Follow the `Official Dashboards` menu to explore all default dashboards on their ways to monitor different tech stacks.
 
+## Sidebar Menu and Marketplace
+
+All available feature menu items are only listed in the marketplace(since 9.6.0). They are only visible on the Sidebar Menu when there are relative services
+being observed by various supported observation agents, such as installed language agents, service mesh platform, OTEL integration.
+
+The menu items defined in `ui-initialized-templates/menu.yaml` are the universal marketplace for all default-supported integration.
+The menu definition supports one and two levels items. The leave menu item should have the `layer` for navigation.
```
wu-sheng, line 17:
```suggestion
The menu definition supports one and two levels items. The leaf menu item should have the `layer` for navigation.
```

oap-server/server-starter/src/main/resources/ui-initialized-templates/menu.yaml
```diff
@@ -0,0 +1,107 @@
...
+# limitations under the License.
+
+# menus: the menus of the left navigation bar
+#   - title: the title of the menu item
+#     icon: icon name
+#     layer: related layer name, the menu item should be activated when layer has services
+#     menus: sub menus of the current item
+
+menus:
+  - title: GeneralService
```
wu-sheng, line 23: Title is for display, it should not be like this. `General Service` is proper value.

oap-server/server-starter/src/main/resources/ui-initialized-templates/menu.yaml
```diff
@@ -0,0 +1,107 @@
...
+# menus: the menus of the left navigation bar
+#   - title: the title of the menu item
+#     icon: icon name
+#     layer: related layer name, the menu item should be activated when layer has services
+#     menus: sub menus of the current item
+
+menus:
+  - title: General Service
+    icon: general_service
+    menus:
```
wu-sheng, line 25: @Fine0830 You need to covert the title to what you need for path and name.
We don't define it on the OAP side

wu-sheng (approved):

PR #11036, opened 2023-07-03
Support reduce aggregate function in MQE.(#11031)

weixiang1862:
### reduce function
55567bee522b0cdf2e797f546670b27

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #11031 .

### DIFF
--- a/oap-server/server-query-plugin/mqe-grammar/src/main/antlr4/org/apache/skywalking/mqe/rt/grammar/MQELexer.g4
+++ b/oap-server/server-query-plugin/mqe-grammar/src/main/antlr4/org/apache/skywalking/mqe/rt/grammar/MQELexer.g4
@@ -59,6 +59,7 @@ ABS:         'abs';
 CEIL:        'ceil';
 FLOOR:       'floor';
 ROUND:       'round';
+REDUCE:      'reduce';
 
 // TopN
 TOP_N:        'top_n';
@@ -72,6 +73,7 @@ DES options { caseInsensitive=true; }: 'des';
 
 // Literals
 INTEGER: Digit+;
+STRING_PARAM: '\'' NameLetter+ '\'';
 DECIMAL: Digit+ DOT Digit+;
 NAME_STRING: NameLetter+;
 VALUE_STRING: '\'' .*? '\'' | '"' .*? '"';
--- a/oap-server/server-query-plugin/mqe-grammar/src/main/antlr4/org/apache/skywalking/mqe/rt/grammar/MQEParser.g4
+++ b/oap-server/server-query-plugin/mqe-grammar/src/main/antlr4/org/apache/skywalking/mqe/rt/grammar/MQEParser.g4
@@ -27,7 +27,7 @@ expression
     | expression mulDivMod expression  # mulDivModOp
     | expression addSub expression     # addSubOp
     | expression compare expression    # compareOp
-    | aggregation L_PAREN expression R_PAREN # aggregationOp
+    | aggregation L_PAREN expression (COMMA parameter)* R_PAREN # aggregationOp
     | function0 L_PAREN expression R_PAREN #function0OP
     | function1 L_PAREN expression COMMA parameter R_PAREN #function1OP
     | topN L_PAREN metric COMMA parameter COMMA order R_PAREN  #topNOP
@@ -52,7 +52,7 @@ label:           labelName EQ labelValue;
 scalar:   INTEGER | DECIMAL;
 
 aggregation:
-    AVG | COUNT | LATEST | SUM | MAX | MIN | ;
+    AVG | COUNT | LATEST | SUM | MAX | MIN | REDUCE;
 
 // 0 parameter function
 function0:
@@ -65,6 +65,6 @@ topN: TOP_N;
 
 relabels: RELABELS;
 
-parameter:      INTEGER;
+parameter:      INTEGER | STRING_PARAM;
 
 order: ASC | DES;
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -35,6 +35,7 @@
 * Set the `SW_QUERY_MAX_QUERY_COMPLEXITY` default value to 3000(was 1000).
 * Accept `length=4000` parameter value of the event. It was 2000. 
 * Tolerate parameter value in illegal JSON format.
+* Add reduce aggregate function in MQE.
 
 #### UI
 
--- a/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/mqe/rt/MQEVisitor.java
+++ b/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/mqe/rt/MQEVisitor.java
@@ -141,7 +141,7 @@ public class MQEVisitor extends MQEParserBaseVisitor<ExpressionResult> {
             return expResult;
         }
         try {
-            return AggregationOp.doAggregationOp(expResult, opType);
+            return AggregationOp.doAggregationOp(expResult, opType, ctx.parameter());
         } catch (IllegalExpressionException e) {
             ExpressionResult result = new ExpressionResult();
             result.setType(ExpressionResultType.UNKNOWN);
--- a/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/mqe/rt/operation/AggregationOp.java
+++ b/oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/mqe/rt/operation/AggregationOp.java
@@ -20,6 +20,7 @@ package org.apache.skywalking.oap.query.graphql.mqe.rt.operation;
 
 import com.google.common.collect.Streams;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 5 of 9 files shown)


weixiang1862:
@wu-sheng @wankai123  It is not the final code, I want to confirm the following questions:
1. whether `reduce` is a good name?
2. `reduce` should be defined as function or aggregate function?
3. whether reduce avg agorithm should filter null values?
4. I use `(COMMA parameter)*` define aggregate function param list, It means aggregation function can have 0-N params, I'm worried it may effect other agg function，should I define it like 'function0' 'function1'?

wu-sheng:
Reduce seems a strange name, Could you explain the logic?

weixiang1862:
> Reduce seems a strange name, Could you explain the logic?

reduce merge several time serial data into one，in dashboard it looks like several line merged into one line. Essentially, it aggregate labelvalue to singlevalue.

reduce name is reference from java stream reduce. I feel it strange too, so I ask you for help.

wu-sheng:
I think we should use aggregate with function name, with a parameter to match labels, star(*) or a list of known labels.

weixiang1862:
> I think we should use aggregate as function name, with a parameter to match labels, star(*) or a list of known labels.
> 
> And, how would you process the number? avg, max, min? I think it should be a parameter too.

Like this?
```
aggregate(metric_name, by_label, sum|avg|max)
```

weixiang1862:
> I think that makes more sense and clear.

Ok, thanks. I will adjust code like this.

oap-server/server-query-plugin/mqe-grammar/src/main/antlr4/org/apache/skywalking/mqe/rt/grammar/MQEParser.g4
```diff
@@ -27,7 +27,7 @@ expression
     | expression mulDivMod expression  # mulDivModOp
     | expression addSub expression     # addSubOp
     | expression compare expression    # compareOp
-    | aggregation L_PAREN expression R_PAREN # aggregationOp
+    | aggregation L_PAREN expression (COMMA parameter)* R_PAREN # aggregationOp
```
wankai123, line 30: It's better to create a new expression like `top_n` or `relabels`. An expression should have determined parameters to avoid unexpected parsing.

wankai123:
> 1. whether `reduce` is a good name?

aggregate_label or something? I think it's only for the `labeled_value` series.
> 2. `reduce` should be defined as function or aggregate function?

see https://github.com/apache/skywalking/pull/11036#discussion_r1250789220
> 3. whether reduce avg agorithm should filter null values?

I think should reserve NULL if the result is NULL
> 4. I use `(COMMA parameter)*` define aggregate function param list, It means aggregation function can have 0-N params, I'm worried it may effect other agg function，should I define it like 'function0' 'function1'?

please refer to https://github.com/apache/skywalking/pull/11036#discussion_r1250789220

oap-server/server-query-plugin/mqe-grammar/src/main/antlr4/org/apache/skywalking/mqe/rt/grammar/MQELexer.g4
```diff
@@ -72,6 +73,7 @@ DES options { caseInsensitive=true; }: 'des';
 
 // Literals
 INTEGER: Digit+;
+STRING_PARAM: '\'' NameLetter+ '\'';
```
wankai123, line 76: if the params are `sum/max/avg..`, should be better to define them rather than use `String`. Can refer to the `order` https://github.com/apache/skywalking/blob/de3a49ef7b5631f4930d33c0f33b3ce07f461af9/oap-server/server-query-plugin/mqe-grammar/src/main/antlr4/org/apache/skywalking/mqe/rt/grammar/MQEParser.g4#L70

weixiang1862: Ok，thanks. I will fix these.

weixiang1862:
@wu-sheng I think aggregate function does not need a label param to match labels, because metric name expression already have this ability implemented by @wankai123 , the following two expression is equivalent：
```
aggregate_label(mongodb_cluster_connection{label="1,2,3,4"}, SUM)

aggregate_label(mongodb_cluster_connection, {label="1,2,3,4"}, SUM)
```

wu-sheng:
What do you mean? D you mean it is supported or what?

weixiang1862:
> What do you mean? D you mean it is supported or what?

I means `label match` is supported, if we aggregate all label values, we can define MQE like this:
```
aggregate_label(mongodb_cluster_connection, SUM)
```
If we just want aggregate part of labels, we can define MQE like this:
```
aggregate_label(mongodb_cluster_connection{label='1,2,3,4'}, SUM)
```

wu-sheng:
So, no need for a new PR to support?

wankai123:
> @wu-sheng I think aggregate function does not need a label param to match labels, because metric name expression already have this ability implemented by @wankai123 , the following two expression is equivalent：
> 
> ```
> aggregate_label(mongodb_cluster_connection{label="1,2,3,4"}, SUM)
> 
> aggregate_label(mongodb_cluster_connection, {label="1,2,3,4"}, SUM)
> ```

yes, the `aggregate_label` should base on the result of the `labeled_value expression`. Just .ike `avg/min` or other functions.

weixiang1862:
> So, no need for a new PR to support?

Sorry, I didn't explain it clearly.

aggregate_label function is need, and I will perfect this pr according to your requirements we talked yesterday.

But this function does not need a label param to match label because it is already supported by metric_expression.

wu-sheng:
> mongodb_cluster_connection{label="1,2,3,4"}

This style works for me. Question would be, are labels always predicted? If not, do we support all or some sub expression?

wankai123:
> This style works for me. Question would be, are labels always predicted? If not, do we support all or some sub expression?

`mongodb_cluster_connection{label="1,2,3,4"}` is a metric_expression already supported by MQE and returns time series labeled value.  So the labels should be predicted, and the  new `aggregate_label ` func just needs to aggregate the labels in the result of `mongodb_cluster_connection{label="1,2,3,4"}`.

wu-sheng:
OK. Make sense. One tip, the word `label` seems not precious. Could you support `labels` and put that in the docs?
I know it is a breaking change, but a label with multiple values seem not good for the longer term. @wankai123

wankai123:
> OK. Make sense. One tip, the word `label` seems not precious. Could you support `labels` and put that in the docs? I know it is a breaking change, but a label with multiple values seem not good for the longer term. @wankai123

I think it's related to Skywalking's label mechanism, we could discuss it later.

wu-sheng:
Sync discussion conclusion, the default label key would be change to _(under line).

weixiang1862:
I hava completed the code adjustment, add introduce docs and tests about aggregate_label function, please help me review.

oap-server/server-query-plugin/query-graphql-plugin/src/main/java/org/apache/skywalking/oap/query/graphql/mqe/rt/operation/AggregateLabelOp.java
```diff
@@ -0,0 +1,71 @@
...
+            AggregateLabelFunc aggregateLabelFunc = factory.getAggregateLabelFunc();
+            for (MQEValues result : results) {
+                aggregateLabelFunc.combine(result.getValues().get(i).getDoubleValue());
+            }
+            combineTo.get(i).setDoubleValue(aggregateLabelFunc.getResult());
+        }
+
+        MQEValues mqeValues = new MQEValues();
+        mqeValues.setValues(combineTo);
+        expResult.setResults(Collections.singletonList(mqeValues));
```
wankai123, line 68:
```suggestion
        expResult.setResults(Collections.singletonList(mqeValues));
        expResult..setLabeledResult(false);
```

wankai123:
Could you add an `e2e case` to verify the new expression? You can use `service _percentile` metric and refer here: https://github.com/apache/skywalking/blob/8aab7d7ac7386e2a1b0dc8af995ceb1c6715d2ff/test/e2e-v2/cases/mqe/mqe-cases.yaml#L48

weixiang1862:
> Could you add an `e2e case` to verify the new expression? You can use `service_percentile` metric and refer here:
> 
> https://github.com/apache/skywalking/blob/8aab7d7ac7386e2a1b0dc8af995ceb1c6715d2ff/test/e2e-v2/cases/mqe/mqe-cases.yaml#L48

Ok, I will add a e2e case.

docs/en/api/metrics-query-expression.md
```diff
@@ -178,3 +178,28 @@ relabel(service_percentile{label='0,1,2,3,4'}, label='P50,P75,P90,P95,P99')
 
 ### Result Type
 Follow the input expression.
+
+## AggregateLabel Operation
+AggregateLabel Operation takes an expression and performs aggregate calculation on its labeled value results. It aggregates a group of TIME_SERIES_VALUES into a single TIME_SERIES_VALUES.
```
wu-sheng, line 183: The original data format should be labeled value metric? Please pick the term of type SkyWalking defined

wu-sheng (commented):
One thing about the function name, from your current logic, it seems all labels are going to be merged, right?

weixiang1862:
> One thing about the function name, from your current logic, it seems all labels are going to be merged, right?

Yes, all labels will be merged.

wu-sheng:
> > One thing about the function name, from your current logic, it seems all labels are going to be merged, right?
> 
> 
> 
> Yes, all labels will be merged.

Is aggregate_labels better?

weixiang1862:
> > > One thing about the function name, from your current logic, it seems all labels are going to be merged, right?
> > 
> > 
> > Yes, all labels will be merged.
> 
> Is aggregate_labels better?
aggregate_labels beter than aggregate_label, I will change name to this, thanks.

weixiang1862:
I have changed function name to `aggregate_labels` and add an e2e case for this function.

weixiang1862:
Anything I can do? I found there are 4 failings.

wu-sheng:
I rerun, it works.

wankai123 (approved):
lgtm

PR #11068, opened 2023-07-10
Allow skipping UI menu configuration

kezhenxu94:
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/docs/en/changes/changes.md). No need


wu-sheng (approved):

kezhenxu94:
@wu-sheng @wankai123 I'm also wondering whether it's suitable to put UI templates management and UI menu management in the core module, in the current architecture, we allow users to build their own query plugin and UI, but we now force the UI templates/menus configurations that are tightly bound to the booster UI, into the core module.

PR #11070, opened 2023-07-10
Add description and document link into the UI menu query protocol

mrproliu:


oap-server/server-starter/src/main/resources/ui-initialized-templates/menu.yaml
```diff
@@ -18,90 +18,156 @@
...
 
 menus:
   - title: General Service
     icon: general_service
+    description:  Observe services and relative direct dependencies through telemetry data collected from SkyWalking Agents.
     menus:
       - title: Services
         layer: GENERAL
+        description: Observe services through telemetry data collected from SkyWalking Agent.
+        documentLink: https://skywalking.apache.org/docs/main/next/en/setup/service-agent/server-agents/
```
wu-sheng, line 32: @Fine0830 We added `description` and `documentLink` as we discussed. 
- description, represents the comments showing as a description of the catalog, or the leaf feature.
- documentLink, provides an extra link, usually for now, only for the leaf feature. On the marketplace page, we should add another link button for this.

wu-sheng (dismissed):
LGTM

wu-sheng (approved):

PR #11071, opened 2023-07-10
[code refactoring] CLRSourceDispatcher and JVMSourceDispatcher shall reside in one package

lu-xiaoshuang:
[code refactoring] CLRSourceDispatcher and JVMSourceDispatcher shall reside in one package.

### DIFF
--- a/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/clr/CLRSourceDispatcher.java
+++ b/oap-server/analyzer/agent-analyzer/src/main/java/org/apache/skywalking/oap/server/analyzer/provider/clr/CLRSourceDispatcher.java
@@ -16,7 +16,7 @@
  *
  */
 
-package org.apache.skywalking.oap.server.receiver.clr.provider.handler;
+package org.apache.skywalking.oap.server.analyzer.provider.clr;
 
 import org.apache.skywalking.apm.network.common.v3.CPU;
 import org.apache.skywalking.apm.network.language.agent.v3.CLRMetric;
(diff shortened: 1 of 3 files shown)


wu-sheng:
I don't think moving it inside the analyzer is a good idea.

lu-xiaoshuang:
> What is the point of this change? I can't see it.

@wu-sheng 

1. JVMSourceDispatcher.java belongs to oap-server/analyzer/agent-analyzer direcorty.
2. CLRSourceDispatcher.java belongs to oap-server/server-receiver-plugin/skywalking-clr-receiver-plugin directory.

JVMSourceDispatcher and CLRSourceDispatcher are similar things, but they reside in different packages. Does it make any sense?

CC: @peng-yongsheng @kezhenxu94

lu-xiaoshuang:
> Because JVM is native, but clr(.net) is out of the ASF. We keep that receiver as a plugin to add on. Also, we only had Java agent for first 3 years, everything about Java agent having higher prioritized when implementating. That is the reality.

I don't think **clr(.net) is out of the ASF** is a good reason for CLRSourceDispatcher.java has not been placed in oap-server/analyzer/agent-analyzer direcorty.

Key point is `agent-analyzer` (on which `kafka-fetcher-plugin` relies) enables `kafka-fetcher-plugin` to laverage `JVMSourceDispatcher` to handle messages from kafka. If CLRSourceDispatcher does not belong to `agent-analyzer`, it means that all other virtual machines (e.g. dotnet) can not report their data via kafka. This may be a critical defect of the design.

CC: @peng-yongsheng @kezhenxu94

wu-sheng:
CLR and JVM metrics as separate gRPC services are history only. The meter APIs have covered all these. PVM(Python), Go VM(go agent) are all using the new meter APIs.
We don't plan to update anything for the legacy things as it has little value.

PR #11072, opened 2023-07-11
Removed "&" symbols from shell scripts

rabajaj0509:
This commit removes all "&" symbols from the shell script. The "&" symbol is typically used for background execution and concurrent tasks. The refactor simplifies the script's logic and ensures deterministic command execution while running them as systemd services.

### DIFF
--- a/dist-material/bin/oapService.sh
+++ b/dist-material/bin/oapService.sh
@@ -38,12 +38,4 @@ done
 OAP_OPTIONS=" -Doap.logDir=${OAP_LOG_DIR}"
 
 eval exec "\"$_RUNJAVA\" ${JAVA_OPTS} ${OAP_OPTIONS} -classpath $CLASSPATH org.apache.skywalking.oap.server.starter.OAPServerStartUp \
-        2>${OAP_LOG_DIR}/oap.log 1> /dev/null &"
-
-if [ $? -eq 0 ]; then
-    sleep 1
-	echo "SkyWalking OAP started successfully!"
-else
-	echo "SkyWalking OAP started failure!"
-	exit 1
-fi
+        2>${OAP_LOG_DIR}/oap.log 1> /dev/null"
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -39,6 +39,7 @@
 * Support aggregate `Labeled Value Metrics` in MQE.
 * [Breaking change] Change the default label name in MQE from `labe`l to `_`.
 * Bump up grpc version to 1.53.0.
+* [Breaking change] Removed '&' symbols from shell scripts to avoid OAP server process running as a background process.
 
 #### UI
 
@@ -52,6 +53,7 @@
 * Fix the title of instance list and notices in the continue profiling.
 * Add a link to explain the expression metric, add units in the continue profiling widget.
 * Calculate string width to set Tabs name width.
+* [Breaking change] Removed '&' symbols from shell scripts to avoid web application server process running as a background process.
 
 #### Documentation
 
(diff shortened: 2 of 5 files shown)


dist-material/bin/oapService.sh
```diff
@@ -38,7 +38,7 @@ done
 OAP_OPTIONS=" -Doap.logDir=${OAP_LOG_DIR}"
 
 eval exec "\"$_RUNJAVA\" ${JAVA_OPTS} ${OAP_OPTIONS} -classpath $CLASSPATH org.apache.skywalking.oap.server.starter.OAPServerStartUp \
-        2>${OAP_LOG_DIR}/oap.log 1> /dev/null &"
```
kezhenxu94, line 41: We can also remove all commands below

kezhenxu94 (commented):
After removing the `&`, all commands afterwards are not necessary anymore, as it blocks until the progress exits, so those `if [ $? -eq 0]; then` can be removed too;

wu-sheng:
So, are we going to make all processes run in blocking mode?

kezhenxu94 (dismissed):

kezhenxu94:
> So, are we going to make all processes run in blocking mode?

Yes, users can simply add `&` to the script if they want to run in background

wu-sheng:
If that is the case, please update the change log at the project level. If(only if) users are using the script at some kind of pipeline, the pipeline is going to be blocked due to this change. We should set a warning for this.
Another check, our docker file is using a different script, right?

kezhenxu94:
> If that is the case, please update the change log at the project level. If(only if) users are using the script at some kind of pipeline, the pipeline is going to be blocked due to this change. We should set a warning for this.
> Another check, our docker file is using a different script, right?

@rabajaj0509 please add the change log, though I don't believe anyone would use these scripts in reality.

wu-sheng:
> please add the change log, though I don't believe anyone would use these scripts in reality.

`Not using`? I don't think so. There are questions about this periodically in the discussion channels. 
Docker/K8s may be daily work for us, but far away from everyone, I am afraid.

kezhenxu94 (dismissed):

docs/en/changes/changes.md
```diff
@@ -39,6 +39,7 @@
 * Support aggregate `Labeled Value Metrics` in MQE.
 * [Breaking change] Change the default label name in MQE from `labe`l to `_`.
 * Bump up grpc version to 1.53.0.
+* Removed '&' symbols from shell scripts to avoid OAP server process running as a background process.
```
wu-sheng, line 42:
```suggestion
* [Breaking change] Removed '&' symbols from shell scripts to avoid OAP server process running as a background process.
```

docs/en/changes/changes.md
```diff
@@ -52,6 +53,7 @@
 * Fix the title of instance list and notices in the continue profiling.
 * Add a link to explain the expression metric, add units in the continue profiling widget.
 * Calculate string width to set Tabs name width.
+* Removed '&' symbols from shell scripts to avoid web application server process running as a background process.
```
wu-sheng, line 56:
```suggestion
* [Breaking change] Removed '&' symbols from shell scripts to avoid web application server process running as a background process.
```

kezhenxu94:
@rabajaj0509 thanks for the changes! As we have systemd unit files in the terraform project, I think users can benefit from the unit files if we add them to this repo too, if you agree, we can move those unit files to this project and, don't have to replicate the unit files in terraform project. Users can benefit from these even if they don't use our terraform module

rabajaj0509:
@kezhenxu94 perfect! I will move both the unit files to this folder itself. Would we need to write a script that places the unit files in the correct location as well?

kezhenxu94:
> @kezhenxu94 perfect! I will move both the unit files to this folder itself. Would we need to write a script that places the unit files in the correct location as well?

That would be a bonus! I would expect a script like `install.sh` that put the unit files to correct place and afterwards users can just `systemctl enable skywalking-oap` and `systemctl start skywalking-oap`

PR #11078, opened 2023-07-12
restrict github actions from running on all PRs

rabajaj0509:
Modified GitHub Actions workflow to only trigger the CodeQL analysis job when changes are made to Java files. This ensures that the analysis focuses specifically on Java code. Enhanced the Dead Link Checker GitHub Actions workflow to only run when changes are made in the 'docs' directory. This allows us to focus the link checking specifically on the documentation files.


rabajaj0509:
Currently, the CI jobs run on each and every PR which is time-consuming, resource inefficient, and frustrating as well. If maintainers and contributors think that this effort is worth it, then I will go ahead and make similar changes in the upcoming PRs.

kezhenxu94 (approved):

kezhenxu94:
> Currently, the CI jobs run on each and every PR which is time-consuming, resource inefficient, and frustrating as well. If maintainers and contributors think that this effort is worth it, then I will go ahead and make similar changes in the upcoming PRs.

CI job is always executed(in order to make the merge button green) but the tasks inside it are not all executed all the time. Check the path filter in the workflow file.

PR #11080, opened 2023-07-12
[feature] add kafka topic skywalking-clr-metrics

lu-xiaoshuang:
Current kafka-fetch-plugin only supports JVM metrics. It would be great if skywalking backend can handles other virtual machines' data from kafka.

This pull request **SHALL** be merged after [11071](https://github.com/apache/skywalking/issues/11071).


wu-sheng:
Merge or not, it is maintainers' call. Not you. I can't see others stand out to support this.
So, it is kept as it was to be. No change to be made.

PR #11086, opened 2023-07-13
Cache all service entity in memory for query

mrproliu:
Querying all service entities and storing them in the cache periodically. All service-level queries should be modified to retrieve data from the cache.


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/CoreModuleConfig.java
```diff
@@ -217,6 +217,20 @@ public class CoreModuleConfig extends ModuleConfig {
...
+    @Setter
+    @Getter
+    private int serviceCacheRefreshInterval = 10;
+
+    /**
+     * The max size of service cache, default 10000
+     */
+    @Setter
+    @Getter
+    private int serviceCacheMaxSize = 10_000;
```
wu-sheng, line 232:
```suggestion
    private int serviceCacheMaxSize = 100_000;
```

As it is only a name cache,  a little more larger should be fine.

wu-sheng: @kezhenxu94 As scrolling is there, this size should be fine, right?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/management/ui/menu/UIMenuManagementService.java
```diff
@@ -48,7 +49,7 @@ public class UIMenuManagementService implements Service, Runnable {
     private ModuleManager moduleManager;
     private CompletableFuture<List<MenuItem>> menuItemFuture;
```
wu-sheng, line 50: Future is going to be changed through another PR, right?

mrproliu: Yes.

kezhenxu94 (dismissed):

wu-sheng (dismissed):

wu-sheng (dismissed):

mrproliu:
Based on the log prompt, it seems that there is a problem displaying all Kubernetes resource logs. I suspect that this issue might be related to the Kubernetes resources, so I need to debug the container. Let me switch to draft mode for now.

mrproliu:
The previous E2E error occurred due to insufficient storage space in GHA, resulting in the inability to write ES data. To resolve this, the data write threshold configuration for ES was updated, leading to the recovery of GHA.

I think now it's ready to review. @wu-sheng @kezhenxu94

kezhenxu94 (approved):

PR #11089, opened 2023-07-14
Revert part of #10616 to fix the unexpected changes

kezhenxu94:


wu-sheng (approved):

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/query/MetricsQueryEsDAO.java
```diff
@@ -192,7 +192,7 @@ public List<MetricsValues> readLabeledMetricsValues(final MetricsCondition condi
             }
         }
         return Util.sortValues(
-            Util.composeLabelValue(condition, labels, idMap),
+            Util.composeLabelValue(condition, labels, ids, idMap),
             ids,
```
wankai123, line 196: I think here already filling all ids with the default value 0? https://github.com/apache/skywalking/blob/1d9a4304f3b8d51fd5419346adf39e21baa3646a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetricsQueryDAO.java#L83

kezhenxu94: No, the length of origins is already 0 here so the for loop is not reached

https://github.com/apache/skywalking/blob/1d9a4304f3b8d51fd5419346adf39e21baa3646a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/storage/query/IMetricsQueryDAO.java#L81-L85

wankai123 (approved):

wankai123 (approved):

PR #11095, opened 2023-07-15
Increase the default memory size to avoid OOM

rabajaj0509:
The changes aim to adjust the Java heap size to prevent memory exhaustion. By increasing the available heap space, we can provide sufficient memory to handle the application's requirements.


kezhenxu94 (approved):
512M is way small indeed

kezhenxu94 (approved):

PR #11099, opened 2023-07-16
Using cache reference for query UI menu

mrproliu:

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/management/ui/menu/UIMenuManagementService.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/management/ui/menu/UIMenuManagementService.java
@@ -18,10 +18,15 @@
 
 package org.apache.skywalking.oap.server.core.management.ui.menu;
 
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import com.google.common.reflect.TypeToken;
 import com.google.gson.Gson;
+import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.skywalking.oap.server.core.CoreModule;
+import org.apache.skywalking.oap.server.core.CoreModuleConfig;
 import org.apache.skywalking.oap.server.core.query.MetadataQueryService;
 import org.apache.skywalking.oap.server.core.query.type.MenuItem;
 import org.apache.skywalking.oap.server.core.storage.StorageModule;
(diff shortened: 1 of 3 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/management/ui/menu/UIMenuManagementService.java
```diff
@@ -35,25 +40,29 @@
...
-    private boolean isMenuItemsBeenFetched = false;
     private MetadataQueryService metadataQueryService;
 
-    public UIMenuManagementService(ModuleManager moduleManager) {
+    public UIMenuManagementService(ModuleManager moduleManager, CoreModuleConfig moduleConfig) {
         this.moduleManager = moduleManager;
-        this.menuItemFuture = new CompletableFuture<>();
+        this.menuItemCache = CacheBuilder.newBuilder()
+            .maximumSize(1)
+            .refreshAfterWrite(moduleConfig.getUiMenuRefreshInterval(), TimeUnit.SECONDS)
```
wu-sheng, line 59: RefreshAfterWrite? Where would we write this cache?
Do I misread the codes?

mrproliu: The cache would be read from the database and written into the cache when accessing the menu for the first time. After the writing is completed, a timer starts counting, and when it reaches a certain point, the data would be automatically rewritten.

wu-sheng: Why push first time late? This kind of mechanism usually is used for cold data, but menu is super hot. I think the timer should be up and running after initialized(if needed).

mrproliu: updated, please re-check

wu-sheng (dismissed):
@kezhenxu94 Another round check?

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/management/ui/menu/UIMenuManagementService.java
```diff
@@ -83,46 +92,23 @@ public void saveMenuAndStartFetch(List<UIMenuItemSetting> menuItems, int fetchIn
         menu.setConfigurationJson(GSON.toJson(menuItems));
 
         getMenuDAO().saveMenu(menu);
-        startingFetchMenu(fetchInterval);
-    }
-
-    private void startingFetchMenu(int fetchIntervalSecond) {
-        Executors.newSingleThreadScheduledExecutor()
-            .scheduleWithFixedDelay(this, 0, fetchIntervalSecond, java.util.concurrent.TimeUnit.SECONDS);
+        this.menuItemCache.put(true, fetchMenuItems());
```
kezhenxu94, line 95: There is no point to call cache `fetchMenuItems`, just call cache.get(true) and the loading cache finds it absent and will call the fetchMenuItems inside the load method

```suggestion
        this.menuItemCache.get(true);
```

wu-sheng (approved):

PR #11118, opened 2023-07-20
Disable dependabot pull requests

kezhenxu94:
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/docs/en/changes/changes.md). NO NEED


kezhenxu94:
We never merge those pull requests and they become annoying

wu-sheng (approved):

PR #11119, opened 2023-07-20
Fix Elasticsearch storage TTL timer, move created the latest index before retrieval indexes by aliases to avoid the 404 exception.

wankai123:

When there is no index under the related es-template (There's no clear reason, but it's possible, such as someone manually deleted it), query indexes by aliases will cause 404 and prevents the following TTL process from executing. And will stop deleting and creating related indexes.

I move created the latest index logic before retrieving indexes by aliases to ensure there is at least has one index.

### DIFF
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -50,6 +50,7 @@
 * Apply MQE on Virtual-Cache layer UI-templates
 * Add Echo component ID(5015) language: Golang.
 * Fix `index out of bounds exception` in `aggregate_labels` MQE function.
+* Fix Elasticsearch storage TTL timer, move created the latest index before retrieval indexes by aliases to avoid the 404 exception. 
 
 #### UI
 
--- a/docs/en/changes/changes.md
+++ b/docs/en/changes/changes.md
@@ -50,7 +50,7 @@
 * Apply MQE on Virtual-Cache layer UI-templates
 * Add Echo component ID(5015) language: Golang.
 * Fix `index out of bounds exception` in `aggregate_labels` MQE function.
-* Fix Elasticsearch storage TTL timer, move created the latest index before retrieval indexes by aliases to avoid the 404 exception. 
+* Move created the latest index before retrieval indexes by aliases to avoid the 404 exception. This just prevents some interference from manual operations.
 
 #### UI
 
--- a/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/HistoryDeleteEsDAO.java
+++ b/oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/HistoryDeleteEsDAO.java
@@ -24,6 +24,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.skywalking.library.elasticsearch.exception.ResponseException;
 import org.apache.skywalking.oap.server.core.analysis.DownSampling;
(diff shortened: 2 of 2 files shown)


docs/en/changes/changes.md
```diff
@@ -50,6 +50,7 @@
 * Apply MQE on Virtual-Cache layer UI-templates
 * Add Echo component ID(5015) language: Golang.
 * Fix `index out of bounds exception` in `aggregate_labels` MQE function.
+* Fix Elasticsearch storage TTL timer, move created the latest index before retrieval indexes by aliases to avoid the 404 exception.
```
wu-sheng, line 53:
```suggestion
* Move created the latest index before retrieval indexes by aliases to avoid the 404 exception. This just prevents some interference from manual operations.
```

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/HistoryDeleteEsDAO.java
```diff
@@ -63,20 +62,26 @@ public void deleteHistory(Model model, String timeBucketColumnName, int ttl) {
...
             return;
         }
+
+        String latestIndex = TimeSeriesUtils.latestWriteIndexName(model);
+        if (!client.isExistsIndex(latestIndex)) {
+            client.createIndex(latestIndex);
+            if (log.isDebugEnabled()) {
+                log.debug("Latest index = {} is not exist, create.", latestIndex);
+            }
+        }
```
kezhenxu94, line 72: Please revert the `try-catch` back, the index might be created by by ES when the data is received and if you remove this `try-catch` exception might be thrown.

Even if you check `if (!client.isExistsIndex(latestIndex)) {` but data might be sent to ES after the `if` statement before the `client.createIndex(latestIndex);`

docs/en/changes/changes.md
wu-sheng: First of all, this is unexpected from us that users manually change indices.

Then, @kezhenxu94 retrieve indices by alias without index could cause 404, is this an expected ES client behavior?

wankai123: And another possible cause: TTL deletes all indexes and fails to create the latest index (no chance to create again)

kezhenxu94: > And another possible cause: TTL deletes all indexes and fails to create the latest index (no chance to create again)

In what situation TTL can delete all indices but fail to create new index? TTL delete only one outdated index per day, if all the indices are deleted it might indicate that it always fails when creating index and thus indicates the ES server is broken, why it succeeds to create new index if you move the creation to earlier?

wankai123: > > And another possible cause: TTL deletes all indexes and fails to create the latest index (no chance to create again)
> 
> In what situation TTL can delete all indices but fail to create new index? TTL delete only one outdated index per day, if all the indices are deleted it might indicate that it always fails when creating index and thus indicates the ES server is broken, why it succeeds to create new index if you move the creation to earlier?

yes, it just exists in theory, network or other issues that cause success to delete but fails to create.

I move create index earlier to give a chance to recreate index.

oap-server/server-storage-plugin/storage-elasticsearch-plugin/src/main/java/org/apache/skywalking/oap/server/storage/plugin/elasticsearch/base/HistoryDeleteEsDAO.java
toffentoffen: >  the index might be created by ES when the data is received

Although this should never happen when the TTL job is working fine, as it will create the index for the next day in advance, this a scenario to take into consideration.

wu-sheng (approved):

PR #11120, opened 2023-07-20
Do not print sensitive properties' values in logs

kezhenxu94:

### DIFF
--- a/oap-server/server-starter/src/main/java/org/apache/skywalking/oap/server/starter/config/ApplicationConfigLoader.java
+++ b/oap-server/server-starter/src/main/java/org/apache/skywalking/oap/server/starter/config/ApplicationConfigLoader.java
@@ -18,22 +18,20 @@
 
 package org.apache.skywalking.oap.server.starter.config;
 
+import lombok.extern.slf4j.Slf4j;
+import org.apache.skywalking.oap.server.library.module.ApplicationConfiguration;
+import org.apache.skywalking.oap.server.library.module.ProviderNotFoundException;
+import org.apache.skywalking.oap.server.library.util.CollectionUtils;
+import org.apache.skywalking.oap.server.library.util.PropertyPlaceholderHelper;
+import org.apache.skywalking.oap.server.library.util.ResourceUtils;
+import org.yaml.snakeyaml.Yaml;
+
 import java.io.FileNotFoundException;
 import java.io.Reader;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Properties;
-import java.util.Set;
-
-import com.google.common.collect.ImmutableSet;
-import lombok.extern.slf4j.Slf4j;
-import org.apache.skywalking.oap.server.library.util.PropertyPlaceholderHelper;
-import org.apache.skywalking.oap.server.library.module.ApplicationConfiguration;
-import org.apache.skywalking.oap.server.library.module.ProviderNotFoundException;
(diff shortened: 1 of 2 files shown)


oap-server/server-starter/src/main/java/org/apache/skywalking/oap/server/starter/config/ApplicationConfigLoader.java
```diff
@@ -40,6 +43,15 @@
  */
 @Slf4j
 public class ApplicationConfigLoader implements ConfigLoader<ApplicationConfiguration> {
+    static final Set<String> SENSITIVE_PROPERTY_NAMES =
+        ImmutableSet.<String>builder()
+                    .add("password")
+                    .add("authentication")
+                    .add("accessKey")
+                    .add("secretKey")
```
toffentoffen, line 51: NIT: Check against https://skywalking.apache.org/docs/main/next/en/setup/backend/configuration-vocabulary/ and I think 1 is missing.
```suggestion
                    .add("secretKey")
                    .add("firehoseAccessKey")
```

toffentoffen (approved):
LGTM. Just a small suggestion to add 1 more sensitive property.

oap-server/server-starter/src/main/java/org/apache/skywalking/oap/server/starter/config/ApplicationConfigLoader.java
kezhenxu94: @toffentoffen thank you!!

wankai123 (dismissed):

wu-sheng:
I can get the idea to.prevent data link, but the implementation seems not elegant.

Coukd you add an annotatio(SecurityConfig) for the module config field? And leverage that flag to output ******(if set) or empty

wu-sheng (changes requested):
Hard code would not be a good practice, as OAP is high modularization and easy to add new module as extension.

kezhenxu94:
> Hard code would not be a good practice, as OAP is high modularization and easy to add new module as extension.

What do you suggest?

wu-sheng:
> I can get the idea to.prevent data link, but the implementation seems not elegant.
> 
> Coukd you add an annotatio(SecurityConfig) for the module config field? And leverage that flag to output ******(if set) or empty

I mean this.
Every module config should take their own responsibilities to declare this.
And the core only hide when needed.

kezhenxu94:
We'd either not to log the configurations at all if you insist hardcode is not an option

kezhenxu94:
> I can get the idea to prevent data leak, but the implementation seems not elegant.
> 
> Coukd you add an annotatio(SecurityConfig) for the module config field? And leverage that flag to output ******(if set) or empty

That's impossible, when printing the logs the configurations are still in `Map` type, there is no `ModuleConfig` type involved at this time, not to say a new annotation

kezhenxu94:
At the same time I think it's unnecessary to do the complex work just for printing logs, I just removed the logs to print configurations.

kezhenxu94:
> As a quick fix, the current implementation will work. However, I am totally with @wu-sheng suggestion to "modularize" it and wanted to suggest more or less the same. But as a follow up that could be tackled in the near future.
> 
> So, the final solution is to not print any configuration properties, until we implement in a standarized way, isn't it?

The logs are printed too early as soon as they are read, before they are deserialized into the configuration beans, so there is no way to add an annotation to those module config and leverage that annotation to determine whether or not to print the logs, if the logs are really needed they must be moved to somewhere after the configs are deserialized into the configuration beans

PR #11125, opened 2023-07-21
sync ui

weixiang1862:


weixiang1862:
mongodb menu i18n

wu-sheng:
You need update change log according to changes you synced. Please.

wu-sheng (approved):
As i18n only.

PR #11137, opened 2023-07-27
Support multiple configs for each hook types and specifying the hooks in the alarm rule.

wankai123:

### DIFF
--- a/docs/en/setup/backend/backend-alarm.md
+++ b/docs/en/setup/backend/backend-alarm.md
@@ -167,14 +167,15 @@ Submit an issue or a pull request if you want to support any other scopes in Ala
 
 ## Hooks
 Hooks are a way to send alarm messages to the outside world. SkyWalking supports multiple hooks of the same type, each hook can support different configurations. 
-For example, you can configure two Slack hooks, one named `default` and set `is-global: true` means this hook will apply on all `Alarm Rules` **without config** `hooks`.
+For example, you can configure two Slack hooks, one named `default` and set `is-default: true` means this hook will apply on all `Alarm Rules` **without config** `hooks`.
 Another named `custom1` will only apply on the `Alarm Rules` which **with config** `hooks` and include the name `slack.custom1`.
 
 ```yaml
 hooks:
   slack:
     default:
-      is-global: true # if true, this hook will apply on all rules, unless a rule has its own specific hook.
+      # If true, this hook will apply on all rules, unless a rule has its own specific hook. Could have more than one default hooks in the same hook type.
+      is-default: true 
       text-template: |-
         {
           "type": "section",
@@ -278,7 +279,7 @@ The alarm message will be sent through HTTP post by `application/json` content t
 ```yml
 slack:
   default:
-    is-global: true
+    is-default: true
     text-template: |-
       {
         "type": "section",
@@ -297,7 +298,7 @@ The alarm message will be sent through HTTP post by `application/json` content t
 ```yml
 wechat:
   default:
-    is-global: true
+    is-default: true
     text-template: |-
       {
         "msgtype": "text",
@@ -316,7 +317,7 @@ The alarm message will be sent through HTTP post by `application/json` content t
 ```yml
 dingtalk:
   default:
-    is-global: true
+    is-default: true
     text-template: |-
       {
         "msgtype": "text",
@@ -337,7 +338,7 @@ The alarm message will be sent through HTTP post by `application/json` content t
 ```yml
 feishu:
   default:
-    is-global: true
+    is-default: true
     text-template: |-
       {
         "msg_type": "text",
@@ -357,7 +358,7 @@ The alarm message will be sent through HTTP post by `application/json` content t
 ```yml
 welink:
   default:
-    is-global: true
+    is-default: true
     text-template: "Apache SkyWalking Alarm: \n %s."
     webhooks:
     # you may find your own client_id and client_secret in your app, below are dummy, need to change.
@@ -381,7 +382,7 @@ Then configure as follows:
 ```yml
 pagerduty:
   default:
-    is-global: true
+    is-default: true
     text-template: "Apache SkyWalking Alarm: \n %s."
     integration-keys:
     - 5c6d805c9dcf4e03d09dfa81e8789ba1
@@ -396,7 +397,7 @@ Then configure as follows:
 ```yml
 discord:
   default:
-    is-global: true
+    is-default: true
     text-template: "Apache SkyWalking Alarm: \n %s."
     webhooks:
     - url: https://discordapp.com/api/webhooks/1008166889777414645/8e0Am4Zb-YGbBqqbiiq0jSHPTEEaHa4j1vIC-zSSm231T8ewGxgY0_XUYpY-k1nN4HBl
--- a/docs/en/setup/backend/backend-alarm.md
+++ b/docs/en/setup/backend/backend-alarm.md
@@ -299,12 +299,12 @@ wechatHooks:
   default:
     isGlobal: true
     textTemplate: |-
-    {
-      "msgtype": "text",
-      "text": {
-        "content": "Apache SkyWalking Alarm: \n %s."
+      {
+        "msgtype": "text",
+        "text": {
+          "content": "Apache SkyWalking Alarm: \n %s."
+        }
       }
-    }
     webhooks:
     - https://qyapi.weixin.qq.com/cgi-bin/webhook/send?key=dummy_key
 ```
--- a/docs/en/setup/backend/backend-alarm.md
+++ b/docs/en/setup/backend/backend-alarm.md
@@ -46,8 +46,8 @@ For example, in **percentile**, `value1` is the threshold of P50, and `-, -, val
 - **Period**. The size of metrics cache in minutes for checking the alarm conditions. This is a time window that corresponds to the backend deployment env time.
 - **Count**. Within a period window, if the number of times which **value** goes over the threshold (based on OP) reaches `count`, then an alarm will be sent.
 - **Only as condition**. Indicates if the rule can send notifications or if it simply serves as a condition of the composite rule.
-- **Specific Hooks**. Binding the specific names of the hooks when the alarm is triggered.
-  The name format is `{hookType}.{hookName}` (slackHooks.custom1 e.g.) and must be defined in the `hooks` section of the `alarm-settings.yml` file.
+- **Hooks**. Binding the specific names of the hooks when the alarm is triggered.
+  The name format is `{hookType}.{hookName}` (slack.custom1 e.g.) and must be defined in the `hooks` section of the `alarm-settings.yml` file.
   If the hook name is not specified, the global hook will be used.
 - **Silence period**. After the alarm is triggered at Time-N (TN), there will be silence during the **TN -> TN + period**.
 By default, it works in the same manner as **period**. The same Alarm (having the same ID in the same metrics name) may only be triggered once within a period. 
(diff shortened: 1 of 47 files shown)


docs/en/setup/backend/backend-alarm.md
```diff
@@ -223,85 +271,95 @@ message KeyStringValuePair {
...
+  default:
+    isGlobal: true
+    textTemplate: |-
     {
       "type": "section",
       "text": {
         "type": "mrkdwn",
         "text": ":alarm_clock: *Apache Skywalking Alarm* \n **%s**."
       }
     }
```
kezhenxu94, line 289:
```suggestion
      {
        "type": "section",
        "text": {
          "type": "mrkdwn",
          "text": ":alarm_clock: *Apache Skywalking Alarm* \n **%s**."
        }
      }
```

docs/en/setup/backend/backend-alarm.md
```diff
@@ -223,85 +271,95 @@ message KeyStringValuePair {
...
-  textTemplate: |-
+  default:
+    isGlobal: true
+    textTemplate: |-
     {
       "msgtype": "text",
       "text": {
         "content": "Apache SkyWalking Alarm: \n %s."
       }
     }
```
kezhenxu94, line 307:
```suggestion
      {
        "msgtype": "text",
        "text": {
          "content": "Apache SkyWalking Alarm: \n %s."
        }
      }
```

docs/en/setup/backend/backend-alarm.md
```diff
@@ -223,85 +271,95 @@ message KeyStringValuePair {
...
```
kezhenxu94, line 326:
```suggestion
      {
        "msgtype": "text",
        "text": {
          "content": "Apache SkyWalking Alarm: \n %s."
        }
      }
```

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/AlarmHookSettings.java
```diff
@@ -0,0 +1,40 @@
...
+    @Getter
+    private final AlarmHooksType type;
+    @Getter
+    private final boolean isGlobal;
+
+   public AlarmHookSettings(String name, AlarmHooksType type, boolean isGlobal) {
+        this.name = name;
+        this.type = type;
+        this.isGlobal = isGlobal;
+    }
```
kezhenxu94, line 35: Replace with `@RequiredArgsConstructor`

docs/en/setup/backend/backend-alarm.md
```diff
@@ -123,13 +129,20 @@ rules:
     silence-period: 5
     message: The request number of entity {name} non-200 status is more than expected.
     only-as-condition: false
+    specific-hooks:
```
wu-sheng, line 132:
```suggestion
   hooks:
```

I think hooks is a good name already.

docs/en/setup/backend/backend-alarm.md
```diff
@@ -152,7 +165,42 @@ Currently, metrics from the **Service**, **Service Instance**, **Endpoint**, **S
...
+## Hooks
+Hooks are a way to send alarm messages to the outside world. SkyWalking supports multiple hooks of the same type, each hook can support different configurations. 
+For example, you can configure two Slack hooks, one named `default` and set `isGlobal: true` means this hook will apply on all `Alarm Rules` **without config** `specific-hooks`.
+Another named `custom1` will only apply on the `Alarm Rules` which **with config** `specific-hooks` and include the name `slackHooks.custom1`.
+
+```yaml
+hooks:
+  slackHooks:
+    default:
+      isGlobal: true # if true, this hook will apply on all rules, unless a rule has its own specific hook.
```
wu-sheng, line 177: Will mutiple global rules conflict? Or we will pick one randomly?

docs/en/setup/backend/backend-alarm.md
```diff
@@ -368,4 +430,5 @@ the sliding window will be destroyed and re-created, causing the Alarm of this s
 | Count                | count                      | int            |                    |
 | Only as condition    | only-as-condition          | boolean        |                    |
 | Silence period       | silence-period             | int            |                    |
-| Message              | message                    | string         |                    |
\ No newline at end of file
+| Message              | message                    | string         |                    |
+| Specific Hooks       | specific-hooks             | string array   |                    |
```
wu-sheng, line 434: I think we don't specific hook name, as users are set the server side as known.

docs/en/setup/backend/backend-alarm.md
wankai123: support multiple global hooks in one type, if they have different names

wu-sheng: So, how would we react when conflicts? Are we still forwarding all as users say so?

oap-server/server-starter/src/main/resources/alarm-settings.yml
```diff
@@ -41,81 +41,128 @@ rules:
...
 
-pagerDutyHooks:
-  textTemplate: "Apache SkyWalking Alarm: \n %s."
-  integrationKeys:
-#    # you can find your integration key(s) on the Events API V2 integration page for your PagerDuty service(s).
-#    # (you may need to create an Events API V2 integration for your PagerDuty service if you don't have one yet)
-#    # below are dummy keys that should be replaced with your own integration keys.
-#    - dummy_key
-#    - dummy_key2
+hooks:
```
kezhenxu94, line 54: Why do we want to add a section level `hooks` here? The following sections all have a `**hooks` suffix already and this looks pretty verbose

docs/en/setup/backend/backend-alarm.md
wankai123: If use the `yaml` settings, we can't config the same name under one item. such as we can't define 2 `default` name under slackHooks

wankai123: Syntactic restriction

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/discord/DiscordHookCallback.java
```diff
@@ -43,17 +44,28 @@ public class DiscordHookCallback extends HttpAlarmCallback {
...
+
+        Map<String, List<AlarmMessage>> groupedMessages = groupMessagesByHook(alarmMessages);
+        for (Map.Entry<String, List<AlarmMessage>> entry : groupedMessages.entrySet()) {
+            var hookName = entry.getKey();
+            var messages = entry.getValue();
+            var setting = settingsMap.get(hookName);
+            if (setting == null || CollectionUtils.isEmpty(setting.getWebhooks()) || CollectionUtils.isEmpty(
+                messages)) {
+                continue;
+            }
```
kezhenxu94, line 60: You can't simply skip if `settings == null`, `settings` being null has high possibility that the users configure a wrong name that doesn't exist at all (e.g. typo), you should check the hook names exist when reading the configs and bail out as a startup exception.

docs/en/setup/backend/backend-alarm.md
wankai123: The `default` is just a name and has no special meaning, users can define it themself.

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/webhook/WebhookCallback.java
```diff
@@ -39,13 +41,24 @@ public class WebhookCallback extends HttpAlarmCallback {
...
             return;
         }
 
-        for (String url : alarmRulesWatcher.getWebHooks()) {
-            post(URI.create(url), gson.toJson(alarmMessage), Map.of());
+        Map<String, List<AlarmMessage>> groupedMessages = groupMessagesByHook(alarmMessages);
+        for (Map.Entry<String, List<AlarmMessage>> entry : groupedMessages.entrySet()) {
+            var hookName = entry.getKey();
+            var messages = entry.getValue();
+            var setting = settingsMap.get(hookName);
```
kezhenxu94, line 54: The `hookName` from the `groupMessagesByHook(alarmMessages);` is something like `slackHooks.custom1`, but the keys in `settingsMap` is something like `custom1`, how can you get the setting with `settingsMap.get(hookName);`? Did you process the hook name somewhere I missed?

And the tests only covers global webhooks, please test the rule-level web hooks

kezhenxu94 (changes requested):

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/webhook/WebhookCallback.java
wankai123: The full name is `{hookType}.{hookName}`,  define and get from `AlarmHookSettings.java`

oap-server/server-alarm-plugin/src/main/java/org/apache/skywalking/oap/server/core/alarm/provider/discord/DiscordHookCallback.java
wankai123: Added

docs/en/setup/backend/backend-alarm.md
```diff
@@ -152,7 +165,42 @@ Currently, metrics from the **Service**, **Service Instance**, **Endpoint**, **S
...
 
-## Webhook
+## Hooks
+Hooks are a way to send alarm messages to the outside world. SkyWalking supports multiple hooks of the same type, each hook can support different configurations. 
+For example, you can configure two Slack hooks, one named `default` and set `is-global: true` means this hook will apply on all `Alarm Rules` **without config** `hooks`.
+Another named `custom1` will only apply on the `Alarm Rules` which **with config** `hooks` and include the name `slack.custom1`.
+
+```yaml
+hooks:
+  slackHooks:
```
kezhenxu94, line 175:
```suggestion
  slack:
```

docs/en/setup/backend/backend-alarm.md
```diff
@@ -223,120 +271,134 @@ message KeyStringValuePair {
...
+  default:
+    is-global: true
+    text-template: |-
     {
       "msg_type": "text",
       "content": {
         "text": "Apache SkyWalking Alarm: \n %s."
       },
       "ats":"feishu_user_id_1,feishu_user_id_2"
     }
```
kezhenxu94, line 348:
```suggestion
      {
        "msg_type": "text",
        "content": {
          "text": "Apache SkyWalking Alarm: \n %s."
        },
        "ats":"feishu_user_id_1,feishu_user_id_2"
      }
```

kezhenxu94 (approved):

kezhenxu94 (dismissed):

docs/en/setup/backend/backend-alarm.md
```diff
@@ -152,7 +165,42 @@ Currently, metrics from the **Service**, **Service Instance**, **Endpoint**, **S
...
+## Hooks
+Hooks are a way to send alarm messages to the outside world. SkyWalking supports multiple hooks of the same type, each hook can support different configurations. 
+For example, you can configure two Slack hooks, one named `default` and set `is-global: true` means this hook will apply on all `Alarm Rules` **without config** `hooks`.
+Another named `custom1` will only apply on the `Alarm Rules` which **with config** `hooks` and include the name `slack.custom1`.
+
+```yaml
+hooks:
+  slack:
+    default:
+      is-global: true # if true, this hook will apply on all rules, unless a rule has its own specific hook.
```
wu-sheng, line 177: > unless a rule has its own specific hook.

So, if there are `hooks` specified in the rules, the `is-global` is not working.
Am I right? If so, the name `global` seems not proper, it is more like a default rule. We should consider not to set `default` as name in the demo, and change this to `is-default : true`.

wankai123: > So, if there are hooks specified in the rules, the is-global is not working. Am I right?

yes

> If so, the name global seems not proper, it is more like a default rule. We should consider not to set default as name in the demo, and change this to is-default : true

I think `is-global` means you can define more than one global hook in one type. `is-default` mostly means only one?

wu-sheng: But `global` seems not to be able to be overridden.

wankai123: make sense, changed

docs/en/setup/backend/backend-alarm.md
```diff
@@ -152,7 +165,43 @@ Currently, metrics from the **Service**, **Service Instance**, **Endpoint**, **S
...
-## Webhook
+## Hooks
+Hooks are a way to send alarm messages to the outside world. SkyWalking supports multiple hooks of the same type, each hook can support different configurations. 
+For example, you can configure two Slack hooks, one named `default` and set `is-default: true` means this hook will apply on all `Alarm Rules` **without config** `hooks`.
+Another named `custom1` will only apply on the `Alarm Rules` which **with config** `hooks` and include the name `slack.custom1`.
+
+```yaml
+hooks:
+  slack:
+    default:
```
wu-sheng, line 176:
```suggestion
    # default here is just a name, set the field default: true if this notification hook is expected to be default globally.
    default:
```

wu-sheng (approved):

PR #11157, opened 2023-07-31
Fix BooleanMatch and BooleanNotEqualMatch doing Boolean comparison.

### LINKED ISSUE #11156: [Bug] Boolean match bug
### Search before asking

### What happened

The class of BooleanMath and BooleanNotEqualMatch，the first compare method.
It's almost always a mistake to compare two instances of java.lang.String or boxed types like java.lang.Integer using reference equality == or !=, because it is not comparing actual value but locations in memory.

use equals() not == or !=

### How to reproduce

Update methods, using the equals api

### Anything else

_No response_

### Code of Conduct

wu-sheng:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #11156.

I believe codes can't run into this part, but fix it anyway as reported.

### DIFF
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/expression/BooleanMatch.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/expression/BooleanMatch.java
@@ -23,7 +23,7 @@ import org.apache.skywalking.oap.server.core.analysis.metrics.annotation.Boolean
 @BooleanValueFilterMatcher
 public class BooleanMatch {
     public boolean match(Boolean left, Boolean right) {
-        return left == right;
+        return left.equals(right);
     }
 
     public boolean match(boolean left, boolean right) {
--- a/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/expression/BooleanNotEqualMatch.java
+++ b/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/expression/BooleanNotEqualMatch.java
@@ -23,7 +23,7 @@ import org.apache.skywalking.oap.server.core.analysis.metrics.annotation.Boolean
 @BooleanValueFilterMatcher
 public class BooleanNotEqualMatch {
     public boolean match(Boolean left, Boolean right) {
-        return left != right;
+        return !left.equals(right);
     }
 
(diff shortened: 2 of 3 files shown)


oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/expression/BooleanMatch.java
```diff
@@ -23,7 +23,7 @@
 @BooleanValueFilterMatcher
 public class BooleanMatch {
     public boolean match(Boolean left, Boolean right) {
-        return left == right;
+        return left.equals(right);
```
kezhenxu94, line 26: When `left == null`, NPE is thrown

```suggestion
        return Objects.equals(left, right);
```

oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/metrics/expression/BooleanNotEqualMatch.java
```diff
@@ -23,7 +23,7 @@
 @BooleanValueFilterMatcher
 public class BooleanNotEqualMatch {
     public boolean match(Boolean left, Boolean right) {
-        return left != right;
+        return !left.equals(right);
```
kezhenxu94, line 26:
```suggestion
        return !Objects.equals(left, right);
```

kezhenxu94 (approved):
you might need to add `import` after applying the suggestions

wankai123 (approved):

wu-sheng:
I don't think it should be null? Actually, I don't know how codes land here.

kezhenxu94:
> I don't think it should be null? Actually, I don't know how codes land here.

`Boolean` can be null, we have a method `match(boolean left, boolean right)` that `left` and `right` must not be null, and the one you modified `match(Boolean left, Boolean right)` is for those parameters that might be null

wu-sheng:
I mean for OAL context, it seems can't execute here. This is just an in-case method.

wu-sheng:
The only reason I changed, it is from an issue from code scaning.

kezhenxu94:
> The only reason I changed, it is from an issue from code scaning.

Code scanning absolutely will report potential NPE after your change.

wu-sheng:
I don't care about that. This fix is only because the codes are wrong. Potential npe doesn't mean wrong.

PR #11177, opened 2023-08-04
fix incorrect expression in screenshot ref by ui-grafana.md

weixiang1862:
### Fix screenshot


docs/en/changes/changes.md
```diff
@@ -61,6 +61,7 @@
 * Fix BooleanMatch and BooleanNotEqualMatch doing Boolean comparison.
 * Support LogQL HTTP query APIs.
 * Add Mux Server component ID(5017) language: Golang.
+* Fix incorrect LogQL expression in screenshot ref by ui-grafana.md.
```
wu-sheng, line 64:
```suggestion
```

No change log is required, as the error version is never released.

wu-sheng (approved):

PR #11179, opened 2023-08-05
Polish ElasticSearch relative docs and bump up tested versions.

wu-sheng:

* Remove ElasticSearch 6.3.2 from our client lib tests.
* Bump up ElasticSearch server 8.8.1 to 8.9.0 for latest e2e testing. 8.1.0, 7.16.3 and 7.17.10 are still tested.
* Add OpenSearch 2.8.0 to our client lib tests.

### DIFF
--- a/docs/en/setup/backend/backend-storage.md
+++ b/docs/en/setup/backend/backend-storage.md
@@ -10,7 +10,7 @@ storage:
 Natively supported storage:
 - H2
 - OpenSearch
-- ElasticSearch 7, 8, ElasticSearch 6 might also work but is end of life and not supported.
+- ElasticSearch 7 and 8. 
 - MySQL
 - PostgreSQL
 - BanyanDB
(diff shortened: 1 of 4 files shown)


docs/en/setup/backend/backend-storage.md
```diff
@@ -65,9 +67,11 @@ SkyWalking rebuilds the ElasticSearch client on top of ElasticSearch REST API an
 correct request formats according to the server-side version, hence you don't need to download different binaries
 and don't need to configure different storage selectors for different ElasticSearch server-side versions anymore.
 
-For now, SkyWalking supports ElasticSearch 6.x, ElasticSearch 7.x, ElasticSearch 8.x, and OpenSearch 1.x, their
+For now, SkyWalking supports ElasticSearch 7.x, ElasticSearch 8.x, and OpenSearch 1.x, their
 configurations are as follows:
 
+_Notice, ElasticSearch 6 worked and is not promised due to end of life officially._
```
kezhenxu94, line 73: "worked" sounds like it is not working NOW

kezhenxu94 (approved):

docs/en/setup/backend/backend-storage.md
wu-sheng: From my understanding of English(I could be wrong as I am not a native speaker)
- `may work` means there are possibilities of working, but not sure
- `might work` means less possible compared with `may work`
- `worked` means we tried, and it worked before. 

This is the reason I choose `worked` as the new statement.

PR #11194, opened 2023-08-08
Add `view_as_seq` function in MQE for listing metrics in the given prioritized sequence 

mrproliu:


oap-server/mqe-grammar/src/main/antlr4/org/apache/skywalking/mqe/rt/grammar/MQEParser.g4
```diff
@@ -33,6 +33,11 @@ expression
     | topN L_PAREN metric COMMA parameter COMMA order R_PAREN  #topNOP
     | relabels L_PAREN expression COMMA label R_PAREN #relablesOP
     | aggregateLabels L_PAREN expression COMMA aggregateLabelsFunc R_PAREN #aggregateLabelsOp
+    | viewAsSeq L_PAREN expressionList R_PAREN #viewAsSeqOp
```
wu-sheng, line 36: @wankai123 Do we need a kind of function for this? I don't think `viewAsSeq` is so special, as generally, it is just a kind of in-memory logic operations.

wu-sheng: One suggesting group name is `logical_operator`.
Meanwhile, could you rename `function0` and `function1` to `mathematical_operator0`/`1`?

@mrproliu @wankai123 WDYT?

docs/en/changes/changes.md
```diff
@@ -67,7 +67,7 @@
 * Add OpenSearch 2.8.0 to our client lib tests.
 * Apply MQE on RabbitMQ Dashboards.
 * Use listening mode for apollo implementation of dynamic configuration.
-
+* Add `view_as_seq` function in MQE for filtering the not empty metrics.
```
wu-sheng, line 70:
```suggestion
* Add `view_as_seq` function in MQE  for listing metrics in the given prioritized sequence.
```

test/e2e-v2/cases/mqe/mqe-cases.yaml
```diff
@@ -58,4 +58,8 @@ cases:
   - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql metrics exec --expression="aggregate_labels(service_percentile,max)" --service-name=e2e-service-provider
     expected: expected/aggregateLabels-OP.yml
   - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql metrics exec --expression="aggregate_labels(service_percentile,min)" --service-name=e2e-service-provider
-    expected: expected/aggregateLabels-OP.yml
\ No newline at end of file
+    expected: expected/aggregateLabels-OP.yml
+
+  # viewAsSeq-OP
+  - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql metrics exec --expression="view_as_seq(not_existing,service_sla)" --service-name=e2e-service-provider
```
wu-sheng, line 64: What would be the result if both given metrics are not existing?

docs/en/api/metrics-query-expression.md
```diff
@@ -206,3 +206,22 @@ aggregate_labels(total_commands_rate, SUM)
 
 ### Result Type
 The ExpressionResultType of the aggregateLabels operation is TIME_SERIES_VALUES.
+
+## ViewAsSequence Operation
+ViewAsSequence operation searches for expressions in the given list from start to end and returns the expression that not empty.
```
kezhenxu94, line 211: I'm curious what does it mean `viewAsSequence`, the name doesn't look any relevant to the logic behind it: filter first non empty expression.

docs/en/api/metrics-query-expression.md
```diff
@@ -206,3 +206,22 @@ aggregate_labels(total_commands_rate, SUM)
...
+```text
+view_as_seq([<expression_1>, <expression_2>, ...])
+```
+
+For example:
+if the first expression value is empty but the second one not empty, it would return the result from the second expression. 
+The following example would return the content of the **service_sla** metric.
+
+```text
+view_as_seq(not_existing, service_cpm)
```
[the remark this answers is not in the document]
wu-sheng: I think it is a typo

docs/en/api/metrics-query-expression.md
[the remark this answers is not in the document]
wu-sheng: `short-circuit` as the function explanation is good.
But,  this statement may be not correct. 
> the result is returned as soon as the first non empty expression is found and the remaining expressions will not be evaluated

The return is not as soon as the first not-empty is found. I believe the actual execution is getting all expression results first and then check according to the seq.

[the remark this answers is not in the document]
wu-sheng: We don't have this kind of execution plan optimization or similar mechanism. So, if `short-circuit` is just a feeling from result perspective, but actually it isn't.

I was thinking about the name like this, but I am concerned about this misunderstanding.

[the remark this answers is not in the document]
kezhenxu94: So it's purely my misunderstanding of this. My thought was just to add that whether all expressions will be evaluated.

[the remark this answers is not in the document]
wu-sheng: `short-circuit` is still a good name in explanation, I just follow this https://github.com/apache/skywalking/pull/11194#discussion_r1287313101. But, from the current mechanism, we don't have dynamic/intelligent execution plan to archive the short-circuit.

oap-server/mqe-grammar/src/main/antlr4/org/apache/skywalking/mqe/rt/grammar/MQEParser.g4
wankai123: make sense, if there are similar expressions, but the logical processing is different, it is should to distinguish by different group names.

wankai123: `mathematical_operator0/1` is more clear

wu-sheng: @mrproliu Could you update the existing codes as well by following the new proposed names?

test/e2e-v2/cases/mqe/mqe-cases.yaml
mrproliu: It will return the value from the first expression.

oap-server/mqe-grammar/src/main/antlr4/org/apache/skywalking/mqe/rt/grammar/MQEParser.g4
mrproliu: Updated, please re-check.

mrproliu:
> @mrproliu Your e2e tested metric not existing, have you tested the metric is there but has no value? I think that is a more important case.

Got it, I have changed the no-value metric.

docs/en/api/metrics-query-expression.md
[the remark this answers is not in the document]
mrproliu: Yes, It's a typo, and has been fixed.

docs/en/api/metrics-query-expression.md
kezhenxu94: short-circuit in logical operators means, if the left expression already satisfies the condition, the right expressions won't be evaluated at all, e.g., `list != null && list.isEmpty()`, if list is `null`, then the `list.isEmpty()` won't be evaluated at all, thus no NullPointerException, but from what you said 

> The return is not as soon as the first not-empty is found. I believe the actual execution is getting all expression results first and then check according to the seq.

IT's totally opposite to the case, so I think we should say the new function is not short-circuit, all expressions are evaluated no matter how

docs/en/api/metrics-query-expression.md
```diff
@@ -206,3 +206,22 @@ aggregate_labels(total_commands_rate, SUM)
 
 ### Result Type
 The ExpressionResultType of the aggregateLabels operation is TIME_SERIES_VALUES.
+
+## ViewAsSequence Operation
+ViewAsSequence operation represents the first not-null metric from the listing metrics in the given prioritized sequence(left to right). It could also be considered as a `short-circuit` of given metrics for the first value existing metric.
```
kezhenxu94, line 211: Suggestion: make it clear that it does not only check `non-null`, but also check the `isEmptyValue`

wu-sheng: `isEmptyValue` actually `non-null`, because we were using `0` to represent null, which had been changed. So, this is `isEmptyValue` flag.

wu-sheng (approved):
I just polished the docs a little to make it more clear and aligning with our naming changes.

wankai123 (approved):
LGTM

PR #11200, opened 2023-08-09
Fix the wrong default value of `k8sServiceNameRule` if it's not explicitly set

kezhenxu94:


wu-sheng (approved):
LGTM

PR #11204, opened 2023-08-10
Make builds reproducible for automatic releases CI

kezhenxu94:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,7 +44,7 @@
         <url>https://github.com/apache/skywalking</url>
         <connection>scm:git:https://github.com/apache/skywalking.git</connection>
         <developerConnection>scm:git:https://github.com/apache/skywalking.git</developerConnection>
-        <tag>v9.4.0</tag>
+        <tag>HEAD</tag>
     </scm>
 
     <issueManagement>
@@ -150,6 +150,8 @@
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <!-- `project.build.outputTimestamp` is required for reproducible builds: https://maven.apache.org/guides/mini/guide-reproducible-builds.html -->
+        <project.build.outputTimestamp>0</project.build.outputTimestamp>
 
         <powermock.version>2.0.9</powermock.version>
         <checkstyle.version>6.18</checkstyle.version>
@@ -372,7 +374,7 @@
                     <artifactId>antlr4-maven-plugin</artifactId>
                     <version>${antlr.version}</version>
                 </plugin>
-                
+
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-release-plugin</artifactId>
--- a/.github/workflows/skywalking.yaml
+++ b/.github/workflows/skywalking.yaml
@@ -175,9 +175,11 @@ jobs:
           cache: "maven"
       - name: Build distribution tar
         run: |
-          ./mvnw clean install javadoc:javadoc -B -q \
+          ./mvnw clean install javadoc:javadoc -B -q -Pall \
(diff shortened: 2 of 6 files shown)


wu-sheng:
What do you mean `reproducible for automatic releases`? Are you going to add another CI to build release candidate on ASF maven staging?

pom.xml
```diff
@@ -44,7 +44,7 @@
         <url>https://github.com/apache/skywalking</url>
         <connection>scm:git:https://github.com/apache/skywalking.git</connection>
         <developerConnection>scm:git:https://github.com/apache/skywalking.git</developerConnection>
-        <tag>v9.4.0</tag>
+        <tag>HEAD</tag>
```
wu-sheng, line 47: See #10512 did this(to specific tag) automatically. Wonder what we should put here.

kezhenxu94: > See #10512 did this(to specific tag) automatically. Wonder what we should put here.

So #10512 accidentally submit this, here it should be `HEAD` and it will only be changed when doing a release, but should not in master branch

wu-sheng:
Could you explain `reproducible`? I am not following the context.

kezhenxu94:
> Could you explain `reproducible`? I am not following the context.

I'm going to automate (most parts of) the release process by building and publishing the artifacts (including maven packages and the source tar, binary tar) to maven repository in the CI, this requires that the builds are reproducible bit by bit, meaning the binaries built by CI should be exactly the same as what anyone builds locally in their machine, from the same source revision. This is to make it possible for PMC members to verify the binaries before voting on a release, if the builds are not reproducible, you have no idea what's the difference between the binaries built by CI and the one built by you. This is also a mandatory requirement to request infra team to add the GPG keys to GitHub Actions.

pom.xml
wu-sheng: > So https://github.com/apache/skywalking/pull/10512 accidentally submit this, here it should be HEAD and it will only be changed when doing a release, but should not be in the master branch

But every time, when I do the release scripts, it is changed automatically. And then, we merge that release branch into the master branch. 
What could we do to avoid this?

wu-sheng:
Then how to do with https://github.com/apache/skywalking/pull/11204#discussion_r1290262420? That thing is going to change if we apply a new version during `./mvnw release:prepare -DautoVersionSubmodules=true -Darguments='-Dmaven.test.skip' -Pall`. 
I never changed anything manually when doing release.

pom.xml
wu-sheng: Have you tried that? This is what I did before, but it is not changed back. As I mentioned, I never manually change this, but from the commit log, it is from my 9.4 release version bumping up.

kezhenxu94: > Have you tried that? This is what I did before, but it is not changed back. 


I tried that. 


> As I mentioned, I never manually change this, but from the commit log, it is from my 9.4 release version bumping up.

It's from 9.4 release, that's the problem. If it's always like that, the main branch should be 9.5.0(previous version) now, but it's 9.4.0, and I assume it will be 9.4.0 forever even when you release 9.6.0. This is the issue, if it is not HEAD, it assumes you have already set the tag manually and it won't change it, I didn't verify this point though.

kezhenxu94: You can simply check all previous releases tag and compare the  commits before and after that tag, you can easily find how the tag is modified

wu-sheng: Meanwhile, I want to say, I did `9.3.0` release, #10077, with the same script, it was correct. 

image

But the version of apache parent was changed from 25 to 29.
```
<parent>
        <groupId>org.apache</groupId>
        <artifactId>apache</artifactId>
        <version>29</version>
        <relativePath />
    </parent>
```

So, I am wondering whether this is an existing bug, and we may need to fix this as well.

wu-sheng: @kezhenxu94 Please check this, https://github.com/apache/skywalking/pull/11209

wu-sheng (approved):

PR #11210, opened 2023-08-11
Improve PromQL to allow for multiple metric operations within a single query

mrproliu:
When performing multi-metric operations in PromQL, only the labels need to be verified for similarity, and there is no need to verify the names.

### DIFF
--- a/test/e2e-v2/cases/promql/promql-cases.yaml
+++ b/test/e2e-v2/cases/promql/promql-cases.yaml
@@ -61,6 +61,8 @@ cases:
   ## multiple metrics operation query
   - query: curl -X GET http://${oap_host}:${oap_9090}/api/v1/query -d 'query=service_sla{service="e2e-service-consumer", layer="GENERAL"} + service_cpm{service="e2e-service-consumer", layer="GENERAL"}'
     expected: expected/service-metric-vector.yml
+  - query: curl -X GET http://${oap_host}:${oap_9090}/api/v1/query -d 'query=service_percentile{service="e2e-service-consumer", layer="GENERAL", labels="4", relabels="p_value"} - service_percentile{service="e2e-service-consumer", layer="GENERAL", labels="3", relabels="p_value"}'
+    expected: expected/service-metric-labeled-compare-matrix.yml
   ## query_range
   - query: curl -X GET http://${oap_host}:${oap_9090}/api/v1/query_range -d 'query=service_sla{service="e2e-service-consumer", layer="GENERAL"}&start='$(($(date +%s)-1800))'&end='$(date +%s)
     expected: expected/service-metric-matrix.yml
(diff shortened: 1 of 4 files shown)


test/e2e-v2/cases/promql/promql-cases.yaml
```diff
@@ -58,6 +58,9 @@ cases:
     expected: expected/service-metric-sort-vector.yml
   - query: curl -X GET http://${oap_host}:${oap_9090}/api/v1/query -d 'query=service_cpm{layer="GENERAL",top_n="10", order="DES"}[30m]'
     expected: expected/service-metric-sort-matrix.yml
+  ## multiple metrics operation query
+  - query: curl -X GET http://${oap_host}:${oap_9090}/api/v1/query -d 'query=service_sla{service="e2e-service-consumer", layer="GENERAL"} + service_cpm{service="e2e-service-consumer", layer="GENERAL"}'
+    expected: expected/service-metric-vector.yml
```
wankai123, line 63: I think we need to add a test case with labeled metrics.

wu-sheng: @mrproliu Could you add one for percentile with a specific label? Such as `p99`?

wu-sheng: > get specific label from PromQL

What do you mean get specific label?

mrproliu: I thought that you want to make a judgment by only reading a specific label from a certain metric? For example, if there are {label1: "1", label2: "2"}, you would like to compare only the metrics with "label1" and ignore "label2".

wankai123: `service_percentile{service="e2e-service-consumer", layer="GENERAL", labels="0,1,2", relabels="P50,P75,P90"}` is select `0,1,2` and relabel them.

wu-sheng: I mean you could do `service_percentile{service="e2e-service-consumer", layer="GENERAL", labels="4"} - service_percentile{service="e2e-service-consumer", layer="GENERAL", labels="3"}` which mean the gap between p99 and p95 to see whether the long tail of the metric exists or not?

mrproliu: I understand now. I was trying to determine if they have the same labels but different name metrics, so I misunderstood your intention.

wu-sheng: Sorry my fault, it should be 

> service_percentile{service="e2e-service-consumer", layer="GENERAL", labels="4", relabels"p_value"} - service_percentile{service="e2e-service-consumer", layer="GENERAL", labels="3", relabels"p_value"}

The label should be aligned manually.

wankai123: `service_percentile{service="e2e-service-consumer", layer="GENERAL", labels="4", relabels="P99_P55"} - service_percentile{service="e2e-service-consumer", layer="GENERAL", labels="3", relabels="P99_P55"}`

wu-sheng (approved):

PR #11217, opened 2023-08-14
Fix MQE Binary Operation between labeled metrics and other type of value result.

wankai123:


wu-sheng:
Could you update the MQE docs about how to apply the operation between labeled values with no-label values, as well as other combinations?

kezhenxu94:
> Could you update the MQE docs about how to apply the operation between labeled values with no-label values, as well as other combinations?

Do we have real use case for "operation between labeled values with no-label values, and other combinations"?

In the doc, I only see this is valuable:

> If we want to transform the percentile value unit from `ms` to `s` the expression is:
> ```text
> service_percentile{_='0,1'} / 1000
> ```

All the other operations, such as `avg(service_percentile{_='0,1'})`, `service_percentile{_='0,1'} - avg(service_percentile{_='0,1'})` don't make any sense to me in reality. Can we come out with some more useful examples in the doc?

wu-sheng:
> Do we have real use case for "operation between labeled values with no-label values, and other combinations"?

AFAIK, I can't say the current cases are perfect, but it could be useful if we are going to measure the long tail case. 

This may not be used in the metric query, but MQE could use this in the alerting rule settings.

wankai123:
> Can we come out with some more useful examples in the doc?

`avg(service_percentile{_='0,1'})` calculate the avg resp_time in `P50,P70`
`service_percentile{_='0,1'} - avg(service_percentile{_='0,1'}`  calculate the deviation from avg resp_time in `P50,P70`
`service_resp_time - service_percentile{_='0,1'}` calculate the difference from all request avg resp_time between `P50,P70`

I think these calculations make sense in statistics.

wu-sheng (approved):

PR #11221, opened 2023-08-15
Add Spring Runner plugin component ID

lsq27:

https://github.com/apache/skywalking-java/pull/594


wu-sheng:
ID 150 was just registered, please use 151 instead, thanks.

wu-sheng (changes requested):
Pending on https://github.com/apache/skywalking-java/pull/594. @kezhenxu94 and I have doubts about universality of this plugin.

PR #11223, opened 2023-08-16
update changes log about add nacos icon

peachisai:


wu-sheng:
You dont' need to submit this PR separately. The change logs should be updated when UI submodule is being synced.

PR #11226, opened 2023-08-16
Support `Compare Operation` in MQE

wankai123:

### DIFF
--- a/docs/en/api/metrics-query-expression.md
+++ b/docs/en/api/metrics-query-expression.md
@@ -48,7 +48,7 @@ If we want to rename the label values to `P50,P75,P90,P95,P99`, see [Relabel Ope
 The `ExpressionResultType` of the expression is `TIME_SERIES_VALUES` and with labels.
 
 ## Binary Operation 
-The binary Operation is an operation that takes two expressions and performs a calculation on their results.
+The Binary Operation is an operation that takes two expressions and performs a calculation on their results.
 The following table lists the binary operations supported by MQE.
 
 Expression:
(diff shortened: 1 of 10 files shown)


docs/en/api/metrics-query-expression.md
```diff
@@ -87,6 +87,75 @@ The expression could be on the left or right side of the operator.
 | TIME_SERIES_VALUES      | SORTED_LIST/RECORD_LIST   | no     |                          |
 | SORTED_LIST/RECORD_LIST | SORTED_LIST/RECORD_LIST   | no     |                          |
 
+## Compare Operation
+Compare Operation takes two expressions and compares their results.
+The following table lists the compare operations supported by MQE.
```
wu-sheng, line 92: How does the comparison work for a labeled value?

wankai123: same as the binary op and the result is 0 or 1

wu-sheng: OK, so it would be values (0/1) with labels. How would you consider these labeled 0/1 fitting alerting?

wankai123: if the result has multiple labels, select the max matched times one
for example in 5 minutes, the percentil{_'P50,P75'} compare result is:
 P50(0,1,1,1,0), means match 3 times
 P75(0,1,0,1,0), means match 2 times
then the `matchCount` is 3

wu-sheng: Then when we implement `matchCount`, do we consider requiring a single time-series value or accepting one label?

wu-sheng: I see. So, we would still have multiple expected values to match multiple labels as before?

wankai123: no， the max matched label will be set to `matchCount`

PR #11230, opened 2023-08-16
Rename Nacos component to Nacos-client 

peachisai:


wu-sheng:
You need to add a nacos as server, and update mapping according.
Then please test locally first, to see whether logo shows correctly.

peachisai:
> according

